PR #266, opened 2012-02-15
Watch option does not reload the actual changed file

bjoerge:
Running mocha with the --watch option will only reload the tests sources, and never reload changes in the application's source files. It means that it is actually impossible to fix the offending application code while running in watch mode. I assume this is not the desired behavior :-)

Attached commit fixes this by removing the changed file from the require-cache before re-running the suite.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -207,10 +207,11 @@ if (program.watch) {
 
   play(frames);
 
-  utils.watch(utils.files(cwd), function(){
-    stop()
+  utils.watch(utils.files(cwd), function(file){
+    stop();
     suite = suite.clone();
     ui = interfaces[program.ui](suite);
+    delete require.cache[file];
     load(files, function(){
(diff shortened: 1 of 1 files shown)


jareware:
Same here, my project is split into several `node_modules` and anything in another module won't ever get reloaded.

boneskull:
@jareware you will want to investigate other options. 

@mochajs/mocha what if we threw a "DEPRECATED" label on `--watch` in the docs?  This would make me happy.

jareware:
@boneskull yeah I get that feeling reading any issues related to watching here... :) The sad part is when you're e.g. transpiling a large codebase with Babel, the `--watch` method is **a lot** faster than an external watch tool (I've been using https://github.com/kimmobrunfeldt/chokidar-cli for that), as it only recompiles changed modules. Whereas the external one will just start from scratch every time.

PR #601, opened 2012-10-05
Add option to run mocha only in the master cluster process

inspector-ambitious:


Boelensman1:
I would really like this option, its really useful when testing servers and making sure your clustering setup works. Is it still too obscure?

PR #632, opened 2012-10-30
Revert grep to original functionality with strings converted to regex's....

nathanalderson:
...  Add Mocha#find() which escapes the string.

We have a bunch of grep-strings which we pass via URL using a script.  With commit b00b6f54, the Mocha#grep() behavior changed so that the grep strings, which were previously treated as regular expressions, are now escaped.  This broke our grep-patterns.

To me, the word "grep" implies regular expressions, and it is occasionally useful to use regular expressions when selecting tests/suites.  I've reverted Mocha#grep() to the original functionality and implemented the new functionality in Mocha#find(), adding support for a new URL variable "find" which uses the new function.

You may not want this behavior in the main tree, but I thought I'd pass it along.  Feel free to reject.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ lib-cov:
 
 test: test-unit
 
-test-all: test-bdd test-tdd test-qunit test-exports test-unit test-grep test-jsapi test-compilers
+test-all: test-bdd test-tdd test-qunit test-exports test-unit test-only test-jsapi test-compilers
 
 test-jsapi:
 	@node test/jsapi
@@ -72,16 +72,16 @@ test-exports:
 		--ui exports \
 		test/acceptance/interfaces/exports
 
-test-grep:
+test-only:
 	@./bin/mocha \
 	  --reporter $(REPORTER) \
-	  --grep fast \
+	  --only fast \
 	  test/acceptance/misc/grep
 
 test-invert:
 	@./bin/mocha \
 	  --reporter $(REPORTER) \
-	  --grep slow \
+	  --only slow \
 	  --invert \
 	  test/acceptance/misc/grep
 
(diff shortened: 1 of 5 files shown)


danielstjules:
@travisjeffery Thanks for merging https://github.com/mochajs/mocha/pull/1579 ! This can probably be closed too now

nathanalderson:
Wow, thanks for picking this up!  Does #1579 support URL variables for grep and fgrep? We were using mocha from the browser, so we needed URL support.

danielstjules:
@nathanalderson Not sure what you mean by url variables. Does that mean query strings? The PR that was merged is almost exactly your changes applied to master, in addition to @jbnicolai's recommended naming, so behavior should be the same as what's here.

For example, if understood correctly:
``` javascript
$ cat example.js
it('http://www.example.com/?foo=bar', function() {});

it('http://www.example.com/?foo=false', function() {});

$ ./bin/mocha --fgrep '?foo=bar' example.js

  ․

  1 passing (5ms)
```

nathanalderson:
As described in http://mochajs.org/#browser-support, you can use a query
string (`?grep=pattern`) to grep for tests when running tests in a browser.
I was wondering if this PR added support for `?fgrep=string`. Sorry I was
unclear.

danielstjules:
Sure does!

```html
<html>
<head>
  <meta charset="utf-8">
  <title>Mocha Tests</title>
  <link rel="stylesheet" href="mocha.css" />
</head>
<body>
  <div id="mocha"></div>
  <script src="mocha.js"></script>
  <script>mocha.setup('bdd')</script>
  <script>
    it('grep()', function() {});
    it('greps', function() {});
  </script>
  <script>
    mocha.checkLeaks();
    mocha.run();
  </script>
</body>
</html>
```

```
example.html?fgrep=grep()
=> grep()

example.html?fgrep=grep
=> grep(), greps

example.html?grep=grep()
=> grep(), greps

example.html?grep=grep\(\)
=> grep()
```

PR #641, opened 2012-11-04
Fix broken symbols when using node.js on Windows with default terminal fonts

jsalonen:
This pull request fixes - imho - a major annoyance with Mocha when running with node.js on Windows.

Several reporters (`dot`, `list` and `spec`) use special unicode symbols for visualising check marks (`✓`), error marks (`✖`) and small dots (`․`).

Sadly, default terminal fonts for Windows don't support these symbols. While mocha itself works just fine, it is somewhat annoying that your console gets filled with odd placeholder characters. This problem can be worked around by installing a custom font, such as DeJavu Sans Mono (http://stackoverflow.com/questions/9817926/how-can-i-get-mochas-unicode-output-to-display-properly-in-a-windows-console). Given this requires extra work, I'm pretty sure many developers just live with the broken symbols.

This pull request introduces a fix to this problem as follows:
- A new, common symbol map containing symbols for check mark, error mark and small dots has been added to `reporters/base.js`
- In case Node.js/Windows are detected, the default symbol map is modified to fall back to use symbols that are readily available on default console fonts
- Reporters (`dot`, `list`, `spec`) have been modified to use special symbols from the symbol map

### DIFF
--- a/lib/reporters/progress.js
+++ b/lib/reporters/progress.js
@@ -41,7 +41,7 @@ function Progress(runner, options) {
   // default chars
   options.open = options.open || '[';
   options.complete = options.complete || '▬';
-  options.incomplete = options.incomplete || '⋅';
+  options.incomplete = options.incomplete || Base.symbols.dot;
   options.close = options.close || ']';
   options.verbose = false;
 


silkentrance:
Great, now we have to include two sets of test fixtures, one for win32 and one for the rest of the platforms that are doing it correctly 😁

silkentrance:
chcp 65001 anyone before running mocha? or will this not work?

PR #765, opened 2013-03-06
Add when() to BDD

dave-irvine:
This is useful if you want to maintain a group of tests, only some of which
need to run in certain circumstances. For example, testing your shimmed
Array.forEach, but only on browsers that don't support it natively.

Rather than including/excluding specific test files depending on which
platform you are testing, simply wrap the questionable test suite in a
when() block.


sergeyzam:
(myCondition ? xit : it)('should', () => test);

PR #795, opened 2013-04-05
Expose current test information to beforeEach/afterEach, fixes #794, #797

### LINKED ISSUE #794: Expose current test information to beforeEach
Currently only the reporters have access to find out which is the current running test using `runner.on('test', fn)`.

It would be useful if this information was exposed to the `beforeEach` method in some way in order to setup perform some custom actions based on the current test being run. This would allow us to write something like:

``` javascript
// TDD interface
suite('foo', function(){
    setup(function(){
        // this.currentTest would be the Test class for the test being executed
        this.processed = processTitle(this.currentTest.title);
    });
    test('bar', function(){
        assert.ok(doStuff(this.processed));
    });
    test('ham', function(){
        assert.ok(doStuff(this.processed));
    });
});
```

avaly:
Feature description: https://github.com/visionmedia/mocha/issues/794 , https://github.com/visionmedia/mocha/issues/797

### DIFF
--- a/test/hook.async.js
+++ b/test/hook.async.js
@@ -16,18 +16,21 @@ describe('async', function(){
       , 'before test one'
       , 'one'
       , 'after'
+      , 'after test one passed'
       , 'parent after'
       , 'parent before'
       , 'before'
       , 'before test two'
       , 'two'
       , 'after'
+      , 'after test two passed'
       , 'parent after'
       , 'parent before'
       , 'before'
       , 'before test three'
       , 'three'
       , 'after'
+      , 'after test three passed'
       , 'parent after'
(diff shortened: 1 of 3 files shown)


ivolucien:
For posterity, this is a great help in identifying which test fails when a test's output does not uniquely identify the test in question. Having the file and title exposed is exactly what we needed to immediately find the problem test. By far easier than process of elimination.

xbill82:
Hi, was this feature documented? Can't find it in docs...

lotosotol:
hmmm... and how do I retrieve the status of the test suite before closing the browser or before calling the .quite() method?!

PR #918, opened 2013-07-02
Programmatically change title of test case

juzerali:
Right now it has to be done by using [this hack](https://coderwall.com/p/riklkg)

Provide a `this.title()` method inside a test case such that it returns the present test case's title if called without any argument. If called with an argument set it as the title of that test case. 
### Use Case

So that we can do something like this. Displaying the input along with the title in report.

``` javascript

describe("My Test", function(){
    it("should pass with input ", function(){
        var input = "This will be appended at the end of the title ";
        this.title( this.title() + input );
        assert(typeof input === 'string');
    });
});
```

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -67,3 +67,18 @@ Context.prototype.inspect = function(){
     return val;
   }, 2);
 };
+
+/**
+ * Programmatically obtain or set the title of test case.
+ *
+ * @return String
+ * @api public
+ */
+
+Context.prototype.title = function(text){
+  if (!text) {
+    return this._runnable.title;
+  };
+
+  this._runnable.title = text;
+};


FFdhorkin:
> @travisjeffery For you and for anyone else who happens to see this. I do actually utilize this occasionally for data driven tests, and for reducing test repetition. For example, say we had 3 objects that we are expecting to have the same interface.
> 
> ```js
> describe("My Test", function(){
>     var methods_to_test = [
>       { "name": "method1", "method": object1 },
>       { "name": "method2", "method": object2 },
>       { "name": "method3", "method": object3 }
>     ];
>     function iMethodTest(methodDef){
>       it(methodDef.name + " should pass iMethod test ", function(){
>           expect(methodDef.method).to.exist;
>       });
>   });
>   methods_to_test.each(iMethodTest);
> });
> ```

I know this is a super old PR, but... this is not a viable option if you can't know what your test case title will be in advance. Trivial example:

```js
describe('My Test', () => {
    let foo;

    before(() => {
        foo = 'bar';
    });

    function iMethodTest() {
        it(`${foo} should pass iMethod test `, () => {
            expect(foo).to.equal('bar');
        });
    }

    iMethodTest();
});
```

The generated test case passes, but it has a title of "Undefined should pass iMethod test" because before() doesn't execute until after iMethodTest.

Issue #581, opened 2012-09-13
an error in a beforeEach hook causes the test run to terminate prematurely


jfirebaugh:
Testcase: https://gist.github.com/3711350

As you can see from the output, only one example was executed. I would have expected the first example to fail, and then the second two to execute and pass.

Apollon0701:
Uncaught TypeError: vue_router__WEBPACK_IMPORTED_MODULE_2__.default.beforeEach is not a function
what does this mean?
code is as follow
Router.beforeEach((to, from, next) => {
  // redirect to login page if not logged in and trying to access a restricted page
  const publicPages = ['/pages/login','/pages/register'];
  const authRequired = !publicPages.includes(to.path);
  const loggedIn = localStorage.getItem('auth_user');

  if (authRequired && !loggedIn) {
    return next('/login');
  }

  next();
})

shridharkalagi:
i'm still seeing the same issue? any update on this issue?

Issue #516, opened 2012-07-19
after is not getting called when the test process is interrupted by hitting Ctrl + C


c4milo:
@visionmedia what do you think about catching SIGTERM, SIGINT and SIGHUP, running the `after` callback function to clean things up in the test?

jbnicolai:
Issue has been inactive for years, and I feel @tj was completely right in his doubt about adding the feature. Closing for now, happy to re-open if someone feels like this is still a problem.

normanzb:
If looks like this issue still exist in mocha 6.
My use case for this is to make sure puppeteer launched chromium process gets shutted down correctly.
Now, because after() is not called, the chromium process is left in zombie state if ctrl+c is press.

slavistan:
I'm launching docker containers inside `before()` which provide the environment for integration tests. `after()` is supposed to shutdown the ephemeral containers afterwards. This all works fine unless I terminate the tests prematurely via SIGINT, for example when something is stuck inside a loop or if I accidentally ran the wrong test suite. I then have to do the cleanup manully.

Being able to jump into `after()` on SIGINT would be a great feature.

PR #1043, opened 2013-11-19
Hook error handling (#534) -- continue testing even if some hooks are failing

### LINKED ISSUE #534: after() not called on error
describe 'Test', ->
      after -> console.error 'cleanup is not called'
    
      describe 'When I do something', ->
        before -> throw new Error 'BAM' # and there is a problem
        it -> assert.ok true

Rulikkk:
This closes #534 and also affects #973, #975 and maybe #516, #581, #664, etc.

I have implemented hook-error behaviour in a try-finally manner, which means that for the failing hook, its counterpart will be executed (if exists), but the tests will be ignored, and the execution will skip to the next sibling suite.

In detail:
- If before hook fails, tests in a suite (where the hook is) and sub-suites are not executed. If after hook exists, it is called.
- If before each hook fails, remaining tests in that suite (where the hook is) or sub-suites are not executed. The remaining before hooks are also not executed. If after each hooks exist in that suite, or parent suites -- they get executed. The suite's after hooks are also executed.
- If after hook fails, no additional hooks are called.
- If after each hook fails, remaining tests in that suite (where the hook is) and sub-suites are not executed. The remaining after hooks (from parent suites) are executed. After hooks are executed 'up' starting from the suite of the last executed test, finishing with a suite where the hook is (inclusively).

After any hook error, the next sibling suite (relative to the suite with hook error) is executed.

This allows writing test hooks in a symmetrical manner:
- before (each): allocate all resources for this and child test suites or tests
- after (each): free all resources, allocated by before (each). Make sure, this will work correctly, even if some resource allocation failed -- it will still be called.

### DIFF
--- a/test/hook.err.js
+++ b/test/hook.err.js
@@ -1,0 +1,296 @@
+describe('hook error handling', function(){
+  // Lines in this test should be uncommented to see actual behavior
+  // You will also see errors in hooks
+  describe('before hook error', function() {
+    var calls = [];
+    describe('spec 1', function () {
+      describe('spec 1 nested', function () {
+        it('should not be called, because hook error was in a parent suite', function() {
+          calls.push('test nested');
+        })
+      })
+      before(function(){
+        calls.push('before');
+        // throw new Error('before hook error');
+      })
+      after(function(){
+        calls.push('after');
+      })
+      it('should not be called because of error in before hook', function() {
+        calls.push('test');
+      })
+    })
+    describe('spec 2', function () {
+      before(function(){
+        calls.push('before 2');
+      })
+      after(function(){
+        calls.push('after 2');
+      })
+      it('should be called, because hook error was in a sibling suite', function() {
+        calls.push('test 2');
+      })
+    })
+    after(function () {
+      // calls.should.eql(['before', 'after', 'before 2', 'test 2', 'after 2']);
+    })
+  })
+
+  describe('before each hook error', function() {
+    var calls = [];
+    describe('spec 1', function () {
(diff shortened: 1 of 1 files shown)


jest:
@BryanDonovan here's what should do it (I use it this to retrieve remote resources in `before`). However, if you expect the exception to be thrown also in `nextTick`, that should be wrapped in try-catch also.

```javascript

describe("before-and-after", function() {
    before(function(done) {
        process.nextTick(function() {
            try {
                assert.ok(false);
            } catch (e) {
                done(e);
            }
            done();
        });
    });

    after(function() {
        console.log("\nin after()");
    });

    it("the test", function() {
        assert.ok(true);
    });
});```

BryanDonovan:
Yeah, that would work, but it's very cumbersome to put a try/catch in every before() function.

ajaykodali:
I have a question regarding this fix. So, this issues seems to be resolved and is working as expected when using the hooks in sync mode. When I attempt to use the hooks in async mode (with done callback), sibling suites do not get executed.

 If I have two sibling suites with async beforeEach, afterEach and tests - an async error thrown in the afterEach of the first suite will fail all tests in that suite (as expected) and also will stop execution of the second suite which does not seem expected.

ajaykodali:
@BryanDonovan , I think your issue is similar to what I am seeing above. I have tried fixing this in a branch - https://github.com/ajaykodali/mocha/commit/65eba1c533bc1feebc69413fee24a5e176f52a9e
You could try this fix and see if that works for you too.

PR #1070, opened 2013-12-14
--dry-run option

joshski:
Adds a --dry-run option, which makes all tests pass without actually running any.

This is sometimes useful as a developer, but also allows other tools to list tests without running them.

### DIFF
--- a/test/acceptance/misc/dry-run.js
+++ b/test/acceptance/misc/dry-run.js
@@ -2,6 +2,8 @@ var spawn = require('child_process').spawn;
 
 describe('dry-run', function(done){
   it('should pass each test without running any tests', function(done){
+    this.timeout(300);
+
     var run = spawn('./bin/mocha', [__filename, '--dry-run', '--reporter', 'json']);
     var stdout = '';
     var stderr = '';


Blackrush:
This needs to be merged.

clickthisnick:
Useful for listing the amount of tests should be run, then if you run tests in parallel across multiple containers assert all the tests did in fact run

raistlin:
Why is this feature still not merged?

lukewis:
Another use case for this feature:
I split my tests into two categories "unit" tests (which run with the CI build) and "proto" or "integration" tests which can be run on demand.  Typically proto or integration **tests are run one at a time** and often require database interaction or other configuration.  We use them mostly to test our data access logic.  Due to the nature of these tests, **running the wrong test can be a real problem** (manipulating a database unexpectedly, connecting to the wrong environment, etc).

I have two npm scripts to handle this scenario:
```
"scripts": {
    "proto": "mocha dist/tests/proto/**/*.proto.js",
    "test": "mocha dist/tests/**/*.spec.js"
}
```


I use mocha's "only" feature to cherry pick specific tests to run.  However, I often forget to remove the "only" flag when I'm done.  Finding which tests have specified "only" can be rather tediuous.  I'm always very hesitant to hit the "go" button until I'm CERTAIN that I'm only running ONE SPECIFIC TEST.  **The "--dry-run" flag would give me confidence that I'm only going to run the test I intend to run.**

lukewis:
Very interesting, @Droogans , thanks for sharing (especially appreciate the code sample)!  I hadn't used any git hooks before, so researching them has been quite eye-opening.  It feels like they are "really close" to being super useful!  I wish it was **easier** to do things like
- Commit your hooks as part of the repository (so that an entire team shares the same hooks)
- Create cross-platform hooks (in nodejs or similar) so Windows users aren't left behind

I know there are solutions to these problems, but it feels like it adds a moderate amount of complexity to the repository (git is already fairly complex by itself!).  In the end, I'm still on the fence if hooks are a good fit to solve this problem (though I really appreciate the creativity!).  I do love the "simplicity" of just having a --dry-run option, and I too can appreciate your suggested use case of creating a "report" of the active tests.

JamieMcNaught:
Sorry, but this would be really bloody useful.

We've currently got mocha seriously misbehaving following an upgrade and the ability to do a dry run to list all the tests to make sure we've not got require issues or some other issue with the test setup rather than the tests themselves would be a godsend.
Would also be useful for easily getting a list of all our tests - no, grepping the codebase wont do as a lot of our tests are generated programatically and grep'ing just won't provide a true list.

sh3raawii:
Can we re-open this PR since it's indeed extremely useful ?!

sh3raawii:
Guys i created a package for this, go check it out
https://github.com/sh3raawii/mocha-decaf

digeff:
👍

gabssnake:
Still relevant

PR #1084, opened 2013-12-24
Update Jade dependency and update template

mazerte:
I update Jade to 1.0.0 and change template for HTMLCov.

PS: add quote in glob test because my project path have some space

### DIFF
--- a/lib/reporters/templates/coverage.jade
+++ b/lib/reporters/templates/coverage.jade
@@ -1,4 +1,4 @@
-!!! 5
+doctype html
 html
   head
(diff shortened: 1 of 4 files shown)


josephpage:
:+1:
Now jade 1.11.0. You should update.

boneskull:
yeah, we should probably update, but it breaks stuff.  I'm going to reopen this.  PRs accepted, since I don't think anyone wants to go out of their way to fix the issues.

boneskull:
oh wait, this is a PR.

PR #1092, opened 2014-01-01
Fixed Broken Nyan Reporter

FredKSchott:
The Nyan reporter is the only reporter that modified the reporter
prototype. This prototype is overwritten in the mocha.js compilation,
causing the reporter to fail when used in mocha.js. Normally mocha.js
is only used in browsers, where this reporter wouldn't work anyway. But
there are many CLIs that use mocha & phantomJS to run front-end tests
on the command line. Because they rely on mocha.js, Nyan wasn't working.

This commit removes the prototype modifications and moves those functions off
of the NyanCat instance, instead binding them to be run with the correct
this context. This way the prototype can be properly overwritten in mocha.js
while still allowing the reporter to work as expencted.

See: 
https://github.com/jdcataldo/grunt-mocha-phantomjs/issues/14
https://github.com/metaskills/mocha-phantomjs/issues/71#issuecomment-28753156

### DIFF
--- a/lib/reporters/nyan.js
+++ b/lib/reporters/nyan.js
@@ -3,7 +3,8 @@
  */
 
 var Base = require('./base')
-  , color = Base.color;
+  , color = Base.color
+  , rainbowColors = generateColors();
 
 /**
  * Expose `Dot`.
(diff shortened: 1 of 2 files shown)


ryanberckmans:
@tj @boneskull 14 months later, nyan reporter is still broken for mocha-phantomjs users. Please merge :hamster:

PR #1105, opened 2014-01-09
Pass the --use_strict flag through to mocha

b0ri5:
Passes the --use_strict flag through to support harmony/ES6.

Adds a test to ensure that it works (only verifies a usage of 'let' but that should be enough).
The test is goofy because should.js isn't --use_strict compatible so the rule needs to remove test/mocha.opts for its run and return it after completion.

Requires mocha be clean with respect to --use_strict; added two vars to do this. The test will ensure it stays this way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,16 +140,10 @@ test-sort:
 		--sort \
 		test/acceptance/sort
 
-# The test/mocha.opts requires should.js which is not use_strict compliant,
-# so it is necessary to move it before testing harmony. The harmony file uses
-# extension js6 so that it is not included in the unit tests.
 test-harmony:
-	@mv test/mocha.opts test/mocha.opts.backup && \
-	./bin/mocha \
+	@./bin/mocha \
 		--harmony \
-		--use_strict \
-		test/acceptance/harmony.js6; \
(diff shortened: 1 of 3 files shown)


blitzprog:
Hi, it's 2015 now and I don't see a reason why I should add a boilerplate in every single test if mocha could just support strict mode via --use-strict. There are enough modules that can be compiled in strict mode, the argument of "forcing every module to be strict" holds no water anymore. I haven't found that option in `mocha --help`, excuse me if I missed it somewhere. Could you consider adding `--use_strict` again?

garrett-gottlieb:
+1

underoot:
+1

dhoffman34:
+1

stefankendall:
+1

Jank1310:
+1

victorgama:
+!

jholster:
+1

danielstjules:
This was completed via https://github.com/mochajs/mocha/pull/1999 and https://github.com/mochajs/mocha/commit/5722c2ba4b6a49ff9539fff652b0c55c5af9846c

PR #1232, opened 2014-05-29
fixes for strict mode

csantero:
A couple fixes to allow running with --harmony and --use-strict


dcousens:
Is this still the case?

PR #1247, opened 2014-06-23
Allows --watch to monitor additional, user-specified file extensions

lakmeer:
I created this modification after I failed to get Mocha to watch files from my Livescript-based project. It adds a value to the --watch flag to allow the user to specify extra extensions to monitor.
##### Description

The `--watch` flag will now take an optional value which is a comma-separated list of file extensions to monitor in addition to the defaults.

With no value (`mocha --watch ...`) it will monitor `js`, `coffee`, `litcoffee` and `coffee.md` as normal.

With a value (`mocha --watch ls,jsx ...`) it will monitor all of the above, plus `ls` and `jsx`

Thanks,
Lak


KarolAdamiec:
Yeah, just lost some time because of that....

PR #1287, opened 2014-07-27
ammended for deprecated customFds option in child_process

taddei:
I found myself having to patch this up and I thought it might help others too.

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -44,7 +44,7 @@ process.argv.slice(2).forEach(function(arg){
   }
 });
 
-var proc = spawn(process.argv[0], args, { customFds: [0,1,2] });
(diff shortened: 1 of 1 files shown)


rfink:
+1 needed

(also https://www.npmjs.org/package/compare-version is no longer maintained)

Cheap plug for https://www.npmjs.com/package/version-compare.js

marcooliveira:
@boneskull 0.12.0 has been released, and we're getting `child_process: customFds option is deprecated, use stdio instead.`. Maybe it's time for that upgrade?

dasilvacontin:
And we are not supporting node 0.4.x anymore, which was the holdoff.

jbnicolai:
:+1: afaik

boneskull:
Yup.

PR #1323, opened 2014-08-25
Don't report timeouts while enableTimeouts=false

Rob--W:
Timeouts that occcur while enableTimeouts = false should not be
reported when timeouts are enabled again before calling the callback.


jbnicolai:
Hi @Rob--W,

I agree with @boneskull that I don't really see the necessity. Should others chime in and ask for the same, I'd be happy to reopen the corresponding issue. Note that the built of this PR also reported errors. so it would need some attention before being potentially merged anyway.

Thanks for the effort though!

PR #1349, opened 2014-09-09
HTML reporter: Allow the error to contain a formatted HTML error message

papandreou:
For the next version of the [Unexpected](https://github.com/sunesimonsen/unexpected/tree/v5) assertion library we're focusing on adding colored object diffs to error messages in "deep equal" and related cases. We have reached the limits of mocha's current approach with serializing `err.actual` and `err.expected` as JSON and then doing a line-based diff. It imposes too many restrictions on the quality of the output, especially when attempting to shoehorn other types of nested diffs (eg. binary `Buffer` or `Uint8Array`) into the resulting output.

Also, we have implemented a [lightweight type system](https://github.com/sunesimonsen/unexpected/blob/v5/lib/types.js) that allows "userland" to introduce new types and control exactly how they they equal, inspect and diff. It's awkward to make that interoperate with the test framework rendering the diffs.

When running in node.js nothing is preventing us from omitting `err.actual` and `err.expected` and including an ansi-colored diff in `err.message`. We have that working fine already. We also want this to work in the browser, though, where mocha currently doesn't render diffs at all for some reason. We found that the simplest way of accomplishing that in mocha-land is for the HTML reporter to look for an HTML-formatted message on the error object and use it if available.

What do you think, is this something you would consider merging in?

Alternatively, the type system and all formatting of output could be adopted by mocha and maybe other test runners, but we're afraid that would be a premature standardization that would make it too cumbersome to iterate the concept further.

Here's a showcase of some of the diffs we have implemented so far (we temporarily removed the scrollbars for a better overview):

unexpectedobjectdiff

unexpectedhexdiff

unexpectedfunction

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -157,7 +157,13 @@ function HTML(runner) {
 
       stackString = stackString || '';
 
-      el.appendChild(fragment('<pre class="error">%e%e</pre>', message, stackString));
+      if (test.err.htmlMessage && stackString) {
+        el.appendChild(fragment('<div class="html-error">%s\n<pre class="error">%e</pre></div>', test.err.htmlMessage, stackString));
+      } else if (test.err.htmlMessage) {
+        el.appendChild(fragment('<div class="html-error">%s</div>', test.err.htmlMessage));
+      } else {
+        el.appendChild(fragment('<pre class="error">%e%e</pre>', message, stackString));
+      }
     }
 
     // toggle code
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -136,23 +136,28 @@ function HTML(runner) {
       var el = fragment('<li class="test pass pending"><h2>%e</h2></li>', test.title);
     } else {
       var el = fragment('<li class="test fail"><h2>%e <a href="%e" class="replay">‣</a></h2></li>', test.title, self.testURL(test));
-      var str = test.err.stack || test.err.toString();
-
-      // FF / Opera do not add the message
-      if (!~str.indexOf(test.err.message)) {
-        str = test.err.message + '\n' + str;
-      }
+      var stackString, // Note: Includes leading newline
+          message = test.err.toString();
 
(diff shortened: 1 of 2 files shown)


papandreou:
We're still eagerly awaiting feedback on this. At this point this issue should be thought of as a place to discuss how to add colored messages and diffs in the browser. The fact that it includes a PR is just meant to demonstrate our position on how to accomplish it.

By the way -- I'm not sure how to add a test for this behavior in html reporter as it currently doesn't have any.

In the mean time we've published a mocha fork with these patches it in (mocha-papandreou), and Unexpected 5 with support for `htmlMessage` went out out of beta, so if anyone's interested in getting really nice object diffs that are also colored in the browser, try installing mocha-papandreou@2.1.0-patch2 or above and unexpected 5.2.0 or above.

jbnicolai:
I'm still quite excited about this feature.

// CC @mochajs/mocha

sunesimonsen:
A lot of stuff has happened since we opened this feature request, Unexpected v7 is the running version and v8 is just around the corner. This feature request is more valid than ever as we attract more and more users and the html reporter is really lacking behind when you are using Unexpected. 

We created a documentation site that uses the html serializer from Unexpected to format the samples, that will provide you with more examples of why we want this supported by the Mocha html reporter:
https://unexpectedjs.github.io/

This is such a tiny change that is completely backwards compatible, please don't make us fork the html reporter. 

Kind regards
Sune

jbnicolai:
@sunesimonsen I completely agree.

Spend some time reviewing the changes and getting a local demo running. All looks great. Rereading the discussion in this thread, it seems no one was opposed to this change - but it was proposed at a bad time (me and @boneskull just started as maintainers and were more than swamped).

Apologies for the long wait.. and I'd be happy to accept your offer to help maintaining this functionality, @papandreou.

Thanks for the hard work!

sunesimonsen:
@jbnicolai thank you so much, this will be a huge improvement for Unexpected. I know you had a lot going on at the time, no worries at all :-)

PR #1400, opened 2014-10-21
Use fancy diffs for node's assert.deepEqual

glenjamin:
Is there a reason why only opted-in 3rd-party libraries get the fancy coloured/unified diff feature?

I wasn't sure, so here's a PR that detects node core's `deepEqual` and lets it get the nice diff behaviour. :tada:

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -176,15 +176,17 @@ exports.list = function(failures){
       msg = 'Uncaught ' + msg;
     }
 
-    // explicitly show diff
-    if (err.showDiff && sameType(actual, expected)) {
+    var diff = (err.showDiff || err.operator == 'deepEqual');
+
+    // explicitly show diff (or node deepEqual)
+    if (diff && sameType(actual, expected)) {
       escape = false;
       err.actual = actual = utils.stringify(actual);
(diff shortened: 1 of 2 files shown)


jbnicolai:
@boneskull @glenjamin is there anything holding of a merge?

jbnicolai:
This PR is currently no longer mergable, and with the inclusion of [unexpectedjs](https://github.com/unexpectedjs/unexpected) I think it's no longer relevant either.

Feel free to update if you think it still adds value though, and thanks for the effort!

glenjamin:
Could you point me towards the PR for the unexpected change you're referring to? I tried to search but couldn't seem to find anything.

The specific aim here was to improve formatting when using Node core's `assert.deepEqual`, I'm happy to rebase the PR.

jbnicolai:
This one (which was indeed tricky to find): #1349

But now that I'm looking at it, it seems this was only for the HTML reporter. If you'd like to rebase your changes I'd be happy to have an other look at merging them :)

glenjamin:
I looked into this a bit more, the change is no longer needed due to #1626 which changes the default behaviour to always show diffs.

jbnicolai:
@glenjamin :+1: thanks

PR #1412, opened 2014-11-02
Add --disallow-only option

jamestalmage:
Using `describe.only(...)`, `test.only(...)`, etc. can be useful in your local test environment, but you almost never want that filter checked in to source control. This patch adds a `--disallow-only` flag which will create a test failure if that syntax is used. The intended use would be to set the flag for builds on your CI server, but not in your local development environment (except for possibly a pre-commit hook).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ test-disallow-only:
 		--ui tdd \
 		test/acceptance/misc/only/tdd > /dev/null 2>&1 ; \
 		failures="$$?" ; \
-		if [ "$$failures" == '0' ] ; then \
+		if [ "$$failures" = '0' ] ; then \
 			echo 'test-failing:' ; \
 			echo "  expected tdd interface to disallow only but tests passed" ; \
 			exit 1 ; \
@@ -177,7 +177,7 @@ test-disallow-only:
 		--ui bdd \
 		test/acceptance/misc/only/bdd > /dev/null 2>&1 ; \
 		failures="$$?" ; \
-		if [ "$$failures" == '0' ] ; then \
+		if [ "$$failures" = '0' ] ; then \
 			echo 'test-failing:' ; \
 			echo "  expected bdd interface to disallow only but tests passed" ; \
 			exit 1 ; \
@@ -189,7 +189,7 @@ test-disallow-only:
 		--ui qunit \
 		test/acceptance/misc/only/qunit > /dev/null 2>&1 ; \
 		failures="$$?" ; \
-		if [ "$$failures" == '0' ] ; then \
+		if [ "$$failures" = '0' ] ; then \
 			echo 'test-failing:' ; \
 			echo "  expected qunit interface to disallow only but tests passed" ; \
 			exit 1 ; \


hurrymaplelad:
:+1: 

I'd love to see support for this in core mocha.  Accidentally committing a .only will silently bypass CI. Not cool.  We've been bitten by this several times.

There are at least [two](https://github.com/auth0/mocha-only-detector) [scripts](https://github.com/aarongraham/mocha-dot-only) floating around in the node world.  Both recommending installing as a git hook, but the [git-hooks project](https://github.com/icefox/git-hooks#install) notes that  hooks should not be enabled by default.  They're a convenience, not a guarantee.

A git server hook could give us a better guarantee, but [github webhooks](https://developer.github.com/v3/repos/hooks/) don't appear to support rejecting commits like an old fashioned `post-receive`, leaving Travis CI builds vulnerable.

I agree with @jamestalmage that any regular expression based solution will likely yield false-positives and false-negatives.

With such a severe failure mode, I'd really like a guarantee that no `.only`s are running in CI.  I'm about to hack a solution that monkey patches `.only` to throw, but I'd much rather see this in core mocha.

hurrymaplelad:
Perhaps the [plugin API](https://github.com/mochajs/mocha/issues/1457) will support building this more elegantly than a regex or a monkey patch.

jamestalmage:
@hurrymaplelad 
>  I'm about to hack a solution that monkey patches .only to throw, but I'd much rather see this in core mocha.

post a gist/repo here when you do please.

hurrymaplelad:
[Here's the hack](https://github.com/hurrymaplelad/never-say-only).  [Published to npm](https://www.npmjs.com/package/never-say-only) as `never-say-only`.

danielstjules:
@hurrymaplelad Thanks for contributing that solution! :) Hopefully that helps others that needed this functionality.

As for a potential RegExp, this would detect a majority of cases in the wild with the default reporters and ui: `\b(describe|it|suite|test)\s*\.\s*only\s*\(`

``` javascript
// ------------------------------------------
// Typical style
// ------------------------------------------

describe.only('suite', function() { // match
  it.only('test', function() {}); //match
});

// ------------------------------------------
// Ridiculous examples it handles
// ------------------------------------------

describe    .   only(); // match

it.only('test', // match
  function() {

  }
);

it. only('test'); // match

it.
only('test'); //match

it.    only   ('test'); // match

it

  .only('test'); //match

// ------------------------------------------
// What it doesn't handle
// ------------------------------------------

// Comments littered somewhere around the call expression
it.only // not sure why someone would this
('test')

// Alias, which is just silly
// To handle this, you either need to parse the AST
// or bake it into mocha
var alias = it.only;
alias('test');
```

On its own, it'll have false positives for instances found in comments. But you could combine it with a naive approach of filtering out comments:

``` javascript
// it.only()

/* it.only() */

// foobar it.only()

  // it.only()

  // foobar it.only()

/**
 * it.only()
 */

  /**
   * it.only()
   */

  /**
   * foobar it.only()
   */

it.only() // match
```

```
$ grep -E -v "^(\s*//|\s*/\*|\s*\*)" test.js | grep -E "\b(describe|it|suite|test)\s*\.\s*only\s*\("
it.only() // match
```

Though it's possible to come up with contrived counter-examples for any RegExp one might come up with, this might handle 99% of your expected cases. I understand there's exceptions, but keep in mind that mocha's cli already has 45 options... it's a big list.

PR #1439, opened 2014-11-21
add ability to delay root suite; closes #362, closes #1124

### LINKED ISSUE #362: Deferred execution
Hey TJ,

Spent the afternoon improving the unit tests for DocPad, but spent about 6 hours tumbling down the rabbit hole of weirdness.

The first issue I came across, is without line 14 in my [plugins.test.coffee](https://github.com/bevry/docpad/blob/v4.0/test/plugins.test.coffee#L14) file, mocha exits absolutely instantly, not good.

The second issue is a bit more complicated, let's now start doing the test, in the same file we hit [line 33](https://github.com/bevry/docpad/blob/v4.0/test/plugins.test.coffee#L33) which fires another describe for the `pluginName`, then we continue to call the [plugin tester's test function](https://github.com/bevry/docpad/blob/v4.0/lib/testers.coffee#L164) which is all fine and good, however the describes and tests for `testLoaded` and `testGeneration` are no longer grouped under the `pluginName`, although the initial `taskCreation` is grouped under the `pluginName` - go figure?!?!

It seems to me that mocha tries to be a lot smarter than it actually is, in terms of predicting when our tests or test suites have actually finished - and it is failing quite miserably in my use cases.

It seems the only solution to this is to create an interface such as:

``` javascript
// Using Total
// tests fire instantly, and when total is hit, then we are done
// the total is set manually at the start by the user
describe('my plugin',function(desc){
	desc.total = 2;
	desc.it('should know 5*5',function(test){
		assert.equal(25,5*5);
		test.done();
	});
	desc.it('should know 5*10',function(test){
		assert.equal(50,5*10);
		test.done();
	});
});

// Using Push
// tests are appended to a queue, and then fired at the end via a sync or async call
// the total is calculated automatically by the queue length
describe('my plugin',function(desc){
	desc.it('should know 5*5',function(test){
		assert.equal(25,5*5);
		test.done();
	});
	desc.it('should know 5*10',function(test){
		assert.equal(50,5*10);
		test.done();
	});
	desc.async();
	// or for sync: desc.sync();
});
```

that way you always have the correct hierarchy, and will always know exactly when all the children have finished. In DocPad and my other work, we use this paradigm all the time with [balUtil's Groups](https://github.com/balupton/bal-util.npm/blob/master/lib/flow.coffee#L41).

What are your thoughts?

Update: Not sure if this will ever make it into Mocha, so I created [Joe](https://github.com/bevry/joe), which works really really well.

boneskull:
This allows delayed root suite execution with flag `--delay`.

While I was in there I refactored some to reduce LOC (see `lib/interfaces/common.js`).

It works like this:

- if you specify `--delay`, the Runner will wait until the root suite has emitted a `run` event
- in addition, `--delay` will put a function `run()` in the global context
- once you have done your async business and generated your suite(s), call `run()`.
- this emits the event on the root suite, and it is executed.

See the [unit test](/boneskull/mocha/blob/be02d437cfaea67c177f7422e9bde97bbd2c66dd/test/delay/delay.js) for an example.

Also rebuilt `/mocha.js`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,12 @@ test-reporters:
 		--reporter $(REPORTER) \
 		test/reporters/*.js
 
+test-delay:
+	@./bin/mocha \
+		--reporter $(REPORTER) \
+		--delay \
+		test/delay/*.js
+
 test-only:
 	@./bin/mocha \
 		--reporter $(REPORTER) \
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -96,6 +96,7 @@ program
   .option('--trace', 'trace function calls')
   .option('--trace-deprecation', 'show stack traces on deprecations')
   .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
+  .option('--delay', 'wait for async suite definition')
 
 program.name = 'mocha';
 
@@ -280,6 +281,10 @@ if (program.growl) mocha.growl();
 // --async-only
 
 if (program.asyncOnly) mocha.asyncOnly();
+
+// --delay
+
+if (program.delay) mocha.delay();
 
 // --globals
 
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -29,38 +29,13 @@ module.exports = function(suite){
 
   suite.on('pre-require', function(context, file, mocha){
 
-    /**
-     * Execute before running tests.
-     */
+    var common = require('./common')(suites, context);
 
-    context.before = function(name, fn){
-      suites[0].beforeAll(name, fn);
-    };
-
-    /**
-     * Execute after running tests.
-     */
-
-    context.after = function(name, fn){
-      suites[0].afterAll(name, fn);
-    };
-
-    /**
-     * Execute before each test case.
-     */
-
-    context.beforeEach = function(name, fn){
-      suites[0].beforeEach(name, fn);
-    };
-
-    /**
-     * Execute after each test case.
-     */
-
-    context.afterEach = function(name, fn){
-      suites[0].afterEach(name, fn);
-    };
-
+    context.before = common.before;
+    context.after = common.after;
+    context.beforeEach = common.beforeEach;
+    context.afterEach = common.afterEach;
+    context.run = mocha.options.delay && common.runWithSuite(suite);
     /**
      * Describe a "suite" with the given `title`
      * and callback `fn` containing nested suites
@@ -135,5 +110,6 @@ module.exports = function(suite){
     context.it.skip = function(title){
       context.it(title);
     };
+
   });
 };
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -1,0 +1,58 @@
+/**
+ * Functions common to more than one interface
+ * @module lib/interfaces/common
+ */
+
+'use strict';
+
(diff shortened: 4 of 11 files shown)


ainthek:
is this finished ? Please.
I also do not understand why simply not making describe(function(done)) signature to signal, that 
builder of suite is async and needs to wait();

boneskull:
@ainthek It's not finished yet.  I'm not sure what you're suggesting here.

ainthek:
boneskull: I do not get the idea with delay() and run() and wanted to test it.
When looking at the proposal of API I would suggest 

describe(function(done){
asyncData(unroll);
function unroll(err,data){
data.formEach(function(){
it();
});
done();
}
});

simple and symmetric with it, before and others supporting done and async nature

ainthek:
Another trick to solve 'async unroll' until some reasonable support is available:

var assert=require("assert");
describe("Async unroll is possible", function(){
	
	function unrollDataAsyncTask(done){
		// fs, request whatever async here, callback or promise
		setTimeout(function(){
			done(null,[1,2,3,4]);
		}, 1000);
	}

	var addTest=this.addTest.bind(this);
	before(function(done){
		this.timeout(2000);
		// call asynct task and generate tests
		unrollDataAsyncTask(function(err,datas){
			datas.map(_it).forEach(addTest);
			done();	
		});
	});
	// make at least on emethod, so before gets called
	it("Unroll Data shall exist", function(done){
		unrollDataAsyncTask(function(err,datas){
			assert(datas.length>0);
			done();
		});
	});

	// then this will become many it() registered in before();
	function _it(data){
		return it("test for "+data, function(){
			// ...
		});
	}
});

boneskull:
@ainthek I think you misunderstand.  What you wrote is something Mocha already supports.  I'm also not sure what you mean by "unroll".

What Mocha does *not* support is delaying execution of the initial `describe` block.  If you attempt this without the change, currently your tests will not run.

See [this unit test](https://github.com/boneskull/mocha/blob/issue/362/test/delay/delay.js) for an example (the unit test needs some tweaks, but that's the general idea).

ainthek:
why would you need to delay execution of decribe ? one reason is building the list of tests dynamically (based on some asyncly retrieved data, which is called unroll in other fmwks) or waiting for something else before executing function in describe, both can be currently achieved by before. yes i posted what can be already achieved. what and why you are trying to achieve here ? waiting before describe itself is called not before function in describe is called ? why ?  really not clear nor from discussion nor from test case. will have look again.

Sent from my iPad

On 10.2.2015, at 6:24, Christopher Hiller <notifications@github.com<mailto:notifications@github.com>> wrote:


@ainthek<https://github.com/ainthek> I think you misunderstand. What you wrote is something Mocha already supports. I'm also not sure what you mean by "unroll".

What Mocha does not support is delaying execution of the initial describe block. If you attempt this without the change, currently your tests will not run.

See this unit test<https://github.com/boneskull/mocha/blob/issue/362/test/delay/delay.js> for an example (the unit test needs some tweaks, but that's the general idea).

�

dasilvacontin:
@ainthek, please check the relevant threads mentioned in the title: #362 and #1124.

foxx:
I'm so happy to have found this feature, it's a shame it took 2 years to get included but it's here now and that's all that matters. Thank you @boneskull !

Issue #333, opened 2012-03-19
Output gets cut off if piped into another application


joewhite:
If I run my Mocha tests directly from the command line, they run fine, and the output displays with no problems. But if I try to pipe Mocha's output to another program, the output doesn't all get there; only the beginning of the output ever makes it into the pipe.

I've seen this in several manifestations, but it's easy to reproduce just from the command line. I've reproduced this on two different computers, both running 64-bit Windows 7, the latest version of Node.js installed from the Web site, and the latest Mocha installed with `npm install -g mocha`.

If I create a file with just one test in it:

    it('works', function() {});

and run it from the command line, I see the expected output:

    C:\Temp>mocha test.js

      .

      ✔ 1 tests complete (2ms)

    C:\Temp>

But if I pipe its output into any other tool, even something as simple as `more`, then Mocha's output is cut off. I only see the line of dots -- sometimes not even that. The summary line isn't shown.

    C:\Temp>mocha test.js | more

      .

    C:\Temp>

If there are failing tests, and you run STDERR through a pipe, then the test-failure information will be cut off:

    C:\Temp>mocha fail.js 2>&1 | more

      .  Γ£û 1 of 1 tests failed:

    C:\Temp>

I've also seen this behavior when running Mocha from my text editor (Sublime Text 2) and capturing the output into an output window (the output doesn't all appear); and when writing Node.js code that shells out to Mocha (using `child_process.exec`) and captures its STDOUT and STDERR (the output doesn't all get sent). I haven't seen any other app cutting off its output like this in any of these circumstances, but Mocha does it every time its output is a pipe.

am11:
Hello guys, over at node-sass, we are also experiencing the same issue. Any updates? 

Related: https://github.com/sass/node-sass/issues/506 and http://help.appveyor.com/discussions/problems/1091-npm-test-always-fails. 

Note: we tried @stephenmathieson's suggestion and used `--no-exit`, but then it doesn't exit the test process even after passing all the testes (until the CI [times out](https://ci.appveyor.com/project/sass/node-sass/build/250/job/0r75512tu1rbb9hf) and reports failure).

dougwilson:
@am11 solution's code: https://github.com/mochajs/mocha/pull/1440/files

dougwilson:
@am11 it was merged here, but just hasn't yet been released to npm.

am11:
@dougwilson, thanks for fixing this issue in mocha.  :+1: 

[...]

PR #1440, opened 2014-11-21
Fix test running output truncation on async STDIO

dougwilson:
For issue https://github.com/mochajs/mocha/issues/333

Note: this does not address the issues with `mocha interfaces` and other non-test commands, because if we actually waited for the output to get flushed, the default `mocha` action (running tests) would start to occur. Refactoring needs to happen to fix this.

But this solves the output being cut when the STDIO is async, which is the main use-case.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -377,10 +377,30 @@ if (program.watch) {
 // load
 
 mocha.files = files;
-runner = mocha.run(program.exit ? process.exit : exitLater);
+runner = mocha.run(program.exit ? exit : exitLater);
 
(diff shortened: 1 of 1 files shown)


dougwilson:
*ping*

boneskull:
Yes, sorry.  It's on my list.

dougwilson:
It's no problem :) I was just going through all the forks in my account yester to clean them out, haha :D

jonathandelgado:
+1

travisjeffery:
Thanks

Issue #1330, opened 2014-08-30
audit utils.js


boneskull:
There may be some code in `utils.js` that would be better handled by 3p modules, or fully eliminated.

Depending where you stand, 3p deps may be good or bad.

An argument "for" would be that we don't have to maintain code.  An argument "against" is that we have to live with the decisions of whoever maintains the code.

Identify functionality that may be replaced by (better) 3p modules, and raise issues for each.

boneskull:
well, this is cleaned up enough now.

PR #1445, opened 2014-11-24
Support for --tags to include/exclude tests based on some optional tagging

rprieto:
:warning:  **Before you read**
_the suggested tagging API has changed since the initial proposal, see new comments further down_

---

As discussed in #928, this is a tentative PR for tagging support in the `bdd` interface. I hope this is not against the guidelines (PR for a new feature), but it will be easier to discuss what's feasible with some visibility on the code changes.
- you can optionally pass an array of tags as the first argument to `describe` and `it`
- by default, all tests always run
- you can specify `--tags` to run a subset of the tests
- tests that don't match the filter are marked as pending

``` js
describe('my service', function() {
  it('test 1', function() {});
  describe(['integration'], 'backend connectivity', function() {
    describe(['fast'], 'quick tests', function() {
      it('test 2', function() {});
    });
    it(['slow'], 'test 3', function() {});
  });
});
```
- `mocha` runs everything
- `mocha --tags "not:integration"` runs test `1`
- `mocha --tags "is:integration"` runs tests `2` and `3`
- `mocha --tags "is:integration not:slow"` runs test `2`

You can also programmatically update the filter with 

``` js
if (/* out of business hours */) {
  require('mocha').options.tags.add('not:backend');
}
```

If this looks good, happy to look into how to unit test & document it.
Cheers

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/1445)

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -98,6 +98,8 @@ program
   .option('--trace', 'trace function calls')
   .option('--trace-deprecation', 'show stack traces on deprecations')
   .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
+  .option('--tags <tags>', 'only include tests with the given tags', list, [])
+  .option('--skip-tags <tags>', 'exclude all tests with the given tags', list, [])
 
 program.name = 'mocha';
 
@@ -254,6 +256,14 @@ if (program.grep) mocha.grep(new RegExp(program.grep));
 // --invert
 
 if (program.invert) mocha.invert();
+
+// --tags
+
+if (program.tags) mocha.tags(program.tags);
+
+// --skip-tags
+
+if (program.skipTags) mocha.skipTags(program.skipTags);
 
 // --check-leaks
 
--- a/lib/filter.js
+++ b/lib/filter.js
@@ -1,0 +1,36 @@
+var utils = require('./utils');
+
+module.exports = Filter;
+
+function Filter() {
+  this.grep = /.*/;
+  this.invertGrep = false;
+  this.tags = [];
+  this.skipTags = [];
+}
+
+Filter.prototype.count = function(suite) {
+  var self = this;
+  var total = 0;
+  suite.eachTest(function(test){
+    if (self.shouldRun(test)) total++;
+  });
+  return total;
+};
+
+Filter.prototype.shouldRun = function(test) {
+  // --grep
+  var grepMatch = this.grep.test(test.fullTitle());
+  if (this.invertGrep) grepMatch = !grepMatch;
+  // --tags --skip-tags
+  var include = (!this.tags.length) || matchTags(test.ctx._tags, this.tags);
+  var exclude = this.skipTags.length && matchTags(test.ctx._tags, this.skipTags);
+  // final decision
+  return grepMatch && include && !exclude;
+};
+
+function matchTags(actualTags, against) {
+  return utils.some(against, function(tag) {
+    return utils.indexOf(actualTags, tag) !== -1;
+  });
+}
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -5,6 +5,7 @@
 var EventEmitter = require('events').EventEmitter
   , debug = require('debug')('mocha:runner')
   , Test = require('./test')
+  , Filter = require('./filter')
   , utils = require('./utils')
   , filter = utils.filter
   , keys = utils.keys;
@@ -54,12 +55,12 @@ function Runner(suite) {
   var self = this;
   this._globals = [];
   this._abort = false;
+  this._filter = new Filter();
   this.suite = suite;
   this.total = suite.total();
   this.failures = 0;
   this.on('test end', function(test){ self.checkGlobals(test); });
   this.on('hook end', function(hook){ self.checkGlobals(hook); });
-  this.grep(/.*/);
   this.globals(this.globalProps().concat(extraGlobals()));
 }
 
@@ -88,34 +89,42 @@ Runner.prototype.__proto__ = EventEmitter.prototype;
  * @api public
  */
 
-Runner.prototype.grep = function(re, invert){
-  debug('grep %s', re);
-  this._grep = re;
-  this._invert = invert;
-  this.total = this.grepTotal(this.suite);
+Runner.prototype.grep = function(re, invert) {
+  debug('grep %s (%s)', re, invert);
+  this._filter.grep = re;
+  this._filter.invertGrep = invert;
+  this.total = this._filter.count(this.suite);
   return this;
 };
 
 /**
- * Returns the number of tests matching the grep search for the
- * given suite.
+ * Only run tests matching the given tags
  *
- * @param {Suite} suite
- * @return {Number}
+ * @param {Array} tags
+ * @return {Runner} for chaining
  * @api public
  */
 
-Runner.prototype.grepTotal = function(suite) {
-  var self = this;
-  var total = 0;
+Runner.prototype.tags = function(tags){
+  debug('include tags: %s', tags);
+  this._filter.tags = tags;
+  this.total = this._filter.count(this.suite);
+  return this;
+};
 
-  suite.eachTest(function(test){
-    var match = self._grep.test(test.fullTitle());
-    if (self._invert) match = !match;
-    if (match) total++;
-  });
+/**
+ * Exclude tests matching the given tags
+ *
+ * @param {Array} tags
+ * @return {Runner} for chaining
+ * @api public
+ */
 
-  return total;
+Runner.prototype.skipTags = function(tags){
+  debug('skip tags: %s', tags);
+  this._filter.skipTags = tags;
+  this.total = this._filter.count(this.suite);
+  return this;
 };
 
 /**
@@ -430,10 +439,8 @@ Runner.prototype.runTests = function(suite, fn){
     // all done
     if (!test) return fn();
 
-    // grep
-    var match = self._grep.test(test.fullTitle());
-    if (self._invert) match = !match;
-    if (!match) return next();
+    // filter
+    if (!self._filter.shouldRun(test)) return next();
 
     // pending
     if (test.pending) {
@@ -480,7 +487,7 @@ Runner.prototype.runTests = function(suite, fn){
  */
 
 Runner.prototype.runSuite = function(suite, fn){
-  var total = this.grepTotal(suite)
+  var total = this._filter.count(suite)
     , self = this
     , i = 0;
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -118,6 +118,21 @@ exports.filter = function(arr, fn){
 };
 
 /**
+ * Array#some
+ *
+ * @param {Array} array
+ * @param {Function} fn
+ * @api private
+ */
+
+exports.some = function(arr, fn) {
+  for (var i = 0, l = arr.length; i < l; ++i) {
+    if (fn(arr[i], i, arr)) return true;
+  }
+  return false;
+};
+
+/**
  * Object.keys (<=IE8)
  *
  * @param {Object} obj
--- a/mocha.js
+++ b/mocha.js
@@ -223,7 +223,22 @@ var JsDiff = (function() {
 
   var LineDiff = new Diff();
   LineDiff.tokenize = function(value) {
-    return value.split(/^/m);
+    var retLines = [],
+        lines = value.split(/^/m);
+
+    for(var i = 0; i < lines.length; i++) {
+      var line = lines[i],
+          lastLine = lines[i - 1];
+
+      // Merge lines that may contain windows new lines
+      if (line == '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
+        retLines[retLines.length - 1] += '\n';
+      } else if (line) {
+        retLines.push(line);
+      }
+    }
+
+    return retLines;
   };
 
   return {
@@ -779,7 +794,9 @@ module.exports = Context;
  * @api private
  */
 
-function Context(){}
+function Context(){
+  this._tags = [];
+}
 
 /**
  * Set or get the context `Runnable` to `runnable`.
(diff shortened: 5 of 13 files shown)


rprieto:
Hi, any news about this?

It also ties in with other PRs in progress (e.g. #1591) because it puts all the filtering logic together, which is nice IMHO, but might need rework since it's been ~5 months.

rprieto:
Hi @boneskull, what's your thoughts? I think this PR will help, refactoring all the filtering to be done in 1 place. Also just keen to get some feedback to know if I should get my hopes up about tagging tests :smile: (which I still believe adds a lot of value on large projects that need to reliably run selected tests on different environments).

noorus:
I'm very interested to see where this goes and would like to see it merged.
I'm implementing a REST testing framework which invokes mocha programmatically under node/io, and I was going to inject my own runner/interface to implement real tagging in mocha anyway, but now that the work has already been done for me and has hopes of becoming official, I'll wait to see what happens.

--grep is definitely not enough in massive corporate environments, where my work is happening.

rprieto:
Hi @noorus, curious to hear about your REST framework when it's available!

Since you also looked at tagging, does the current suggestion of `--tags "unit,integration" --skip-tags "redis,ie8,slow"` make sense to you? Or do you see a need for the more complex `+` syntax described in #1475? Both suggestions would be exposed on the `runner.filter` object anyway, so you should be able to use them programmatically.

noorus:
I don't really have strong feelings about the CLI in that regard, but I *kinda* want to think that if you find yourself needing the complex `+-,` syntax just to run the right tests, then you might have been overcomplicating your tags in the first place :stuck_out_tongue:
So I think I'd personally go with the simpler syntax which is more intuitive, and might even encourage good tagging practices as such.

rprieto:
Good point :+1: I'm happy to rebase this on master, if there's a chance of it being merged.

noorus:
I really hope the maintainers choose to do the right thing and adopt it soon, since the neato `this.tags` syntax won't even introduce any breaking changes. :+1:

bin/_mocha
```diff
@@ -98,6 +98,8 @@ program
   .option('--trace', 'trace function calls')
   .option('--trace-deprecation', 'show stack traces on deprecations')
   .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
+  .option('--tags <tags>', 'only include tests with the given tags', list, [])
```
jbnicolai: This should probably mention the `a+b` and `a,b` syntax.

lib/utils.js
```diff
@@ -118,6 +118,21 @@ exports.filter = function(arr, fn){
 };
 
 /**
+ * Array#some
```
jbnicolai: Looks like we discussed this here: https://github.com/mochajs/mocha/issues/1330

Want to revisit and get a small lodash build going?

jbnicolai:
@boneskull @travisjeffery did a quick review, and afaik am concerned this would make a nice feature. I'd understand if you consider the bloat and overlap with `grep` to be a reason not to merge though, but let's at least make decision after all the work @rprieto put in ;)

bin/_mocha
rprieto: This version doesn't actually have the `+` syntax, that was just a suggestion in the `gh-pages` branch. But :+1: to the `,`, I'll amend that.

rprieto:
Thanks @jbnicolai :)

noorus:
Sooo.. About this lil' PR here, how about we just slip it into ol' master trunk over there? Any naysayers?

boneskull:
I'll need to look at this again, but I'm gonna catch some zzz's

noorus:
Hey guys, still very much hoping for this merge to happen in master!
Is there anything I could do to further this particular development?

mcquackers:
I'm a big supporter of this, but I was wondering:  Will the reporters output the tags in the desired space?

For example, with the XUnit reporter, in the <testcase> element, having an attribute of "flags='x,y,z'"

rprieto:
I don't see why not :) The tags are a property of a `Suite`'s context so reporters should be able to query/process/report them.

jcrben:
:+1: for unambiguously splitting tests up with tags

rprieto:
Quick update to say I ran into this: http://engineering.sharethrough.com/blog/2013/08/10/greater-test-control-with-rspecs-tag-filters/. It seems RSpec uses a very similar system to what's being proposed here, for what it's worth :)

davidpaulhunt:
@rprieto Did you finish implementing the tagging for all interfaces? Or do you need help cleaning up the merge conflicts, etc?

I'm really interested in making this inherent. Let me know if I can help at all.

rprieto:
Hi @davidpaulhunt! In this PR tagging is implemented as part of the core, mostly in `mocha.js`, `context.js` and a new `filter.js`. I've only done the BDD interface, but it's really just a few lines of code to call into the core, so I believe other interfaces would be very easy.

However you're right that it's been over a year and it would need a massive rebasing. I'm happy to put in the work to rebase/cleanup, if there's an indication it will be merged. Hope that doesn't come across badly, I just think we can already see the API / usage / implementation before going through any rebase. Maybe an update from @boneskull / @danielstjules / @jbnicolai?

davidpaulhunt:
@rprieto Thanks for responding. I'm also willing to put in any work that's needed of me. It'd be great to have this feature out of the box and not as a hack locally.

@danielstjules seems to be more interested if the functionality is successfully implemented across all interfaces. I think at that point, it's safe to assume we could get complete consideration for merging. Of course, I don't want to speak for him, that's just my impression from his response to my PR.

If we get an update or at least a couple of us committed to the feature, we should do as you suggested and work from #1830 via @noorus (again thanks!). I have my fingers crossed we can get momentum!

klamping:
Any hope of this going through? I'd be happy to help test if needed. We're at a point in our project where the ability to run specific subsets of tests will be very valuable (e.g. has X tag but not Y tag)

LikeJasper:
+1
I just spent ages trying to work out why seemingly random extra tests were being run outside a `.only` before I discovered #1481. Tags would completely sidestep that issue.

munkyjunky:
+1, getting this in would be a huge benefit to my testing suite!

hjalet:
+1

jdmarshall:
Do we have a way in this configuration to run all the tests without a tag and all the tests with a particular tag?  

I could run multiple passes of Mocha but that could potentially change the test behavior (it shouldn't, but that's very different from "it won't")

rprieto:
>  to run all the tests without a tag 

If you mean tests without a particular tag, then yes you can do `--tags this --skip-tags that`. If you mean tests without *any* tags, then unfortunately the PR as it stands doesn't support it. There's no way to refer to untagged tests specifically for the moment.

jdmarshall:
Besides things like 'slow', 'fast', 'phantomjs-broken', the next logical place to need tags is for feature toggles.  

Feature toggles are not particularly interesting until you have at least 1 of them, and to have one toggle you need to mark your tests as unaffected by the toggle, only working when disabled, or only working when enabled.  

To run a full tests suite you need to run all of the unaffected tests and the affected tests that apply to the current configuration.  That would be group A and B, or A and C depending on which way it's configured at present.  

Now add a second flag, and you see the --skip-tags flag doesn't scale very well without the catchall.  In order to run all of the unaffected tests you have to update your --skip-tags arguments to include all feature toggles in the project.

If there were a special case for 'none' (empty string perhaps?) then you could run mocha twice to get any combination of tests to run.  That may be good enough, but as hard as we try, tests do tend to behave differently when executed in a different order.

MatthewRalston:
I'd love to see this.

jcrben:
@MatthewRalston please use the thumbs up emoji reaction rather than spamming people with a message like that

rprieto:
Hi @jcrben, any thoughts about this PR? Or more specifically, what do you think is missing so there can be a merge/close decision?

jcrben:
@rprieto I'm not a team member. It looks like @dasilvacontin who is a member gave me a reaction, tho, so maybe we should ask him 😉 

looks like he was waiting on word from @boneskull who never got around to it. @travisjeffery wanted to stick with grep but grep really doesn't handle this in a truly clean way

lib/utils.js
dasilvacontin: This would be a fine workaround until we get around the shims issue, imo.

mocha.js
```diff
@@ -223,7 +223,22 @@ var JsDiff = (function() {
...
+
+      // Merge lines that may contain windows new lines
+      if (line == '\n' && lastLine && lastLine[lastLine.length - 1] === '\r') {
+        retLines[retLines.length - 1] += '\n';
+      } else if (line) {
+        retLines.push(line);
+      }
+    }
+
+    return retLines;
```
dasilvacontin: Don't build `mocha.js`. That's done for releases, it's confusing, and it clutters the PR / commit. No worries though!

lib/runner.js
```diff
@@ -88,34 +89,42 @@ Runner.prototype.__proto__ = EventEmitter.prototype;
...
+ * @param {Array} tags
+ * @return {Runner} for chaining
+ * @api public
+ */
 
-  return total;
+Runner.prototype.skipTags = function(tags){
+  debug('skip tags: %s', tags);
+  this._filter.skipTags = tags;
+  this.total = this._filter.count(this.suite);
```
rprieto: Good point thanks, I'll change extract & for a more descriptive name.

lib/filter.js
```diff
@@ -0,0 +1,36 @@
...
+
+Filter.prototype.shouldRun = function(test) {
+  // --grep
+  var grepMatch = this.grep.test(test.fullTitle());
+  if (this.invertGrep) grepMatch = !grepMatch;
+  // --tags --skip-tags
+  var include = (!this.tags.length) || matchTags(test.ctx._tags, this.tags);
+  var exclude = this.skipTags.length && matchTags(test.ctx._tags, this.skipTags);
+  // final decision
+  return grepMatch && include && !exclude;
```
rprieto: That's true. I thought `grepMatch && include && !exclude` was simple and easy to understand, and chose that over performance. Do you think performance matters a lot here?

dasilvacontin: From my testing, the performance gain is negligible, and this reads easier. LGTM.

dasilvacontin: Sharing so that someone points out my test is bad or wrong:

```js
const filterTags = ['ci']
const testTags = ['slow', 'vendor', 'thing', 'ci']

function matchTags (actualTags, against) {
  return against.some((tag) => {
    return actualTags.indexOf(tag) !== -1
  })
}

console.time('shouldRun')

let tests = 100 * 1000
while (--tests) matchTags(testTags, filterTags)

console.timeEnd('shouldRun')
```

dasilvacontin:
Please rebase since we added a linter to CI recently. 👌

dasilvacontin:
Also, since we are now under the JS Foundation, you'll need to sign the JS Foundation CLA before this gets merged. Just pointing it out, in case it's missed between the other CI checks. :)

aseemk:
Hey folks,

I'm just a regular Mocha user, and I don't feel strongly one way or the other about this feature.

I just wanted to understand: it seems all the mentions of `--grep` here are talking about fuzzy searching over titles, which I agree is imprecise. But what we do — which we learned straight from the [Mocha wiki](https://github.com/mochajs/mocha/wiki/Tagging) — is include tags (by convention) in the title, and search for exactly those tags.

E.g. examples from our real-world code:

```js
// This test suite relies on hitting third-party services,
// so skip it if we're running offline.
describe('Foo bar #skipoffline', function () {

    // This individual test can be pretty slow,
    // because we have to wait for yada yada,
    // so skip it if we're running fast tests only.
    it('lorem ipsum #skipfast', ...);

});
```

Then we can do e.g. `--grep '#skipoffline' --invert`, and combine multiple tags just like normal regex, e.g. `--grep '#skipoffline|#skipfast' --invert`. We have ~5 such tags, including `#skipci` (modeled after Travis's `[skip ci]` commit message syntax).

Is the difference between `--grep` like this, and this PR, that the filtering is more powerful? E.g. I can run tests tagged "X and Y", or "X or Y", or "X but not Y"? I _think_ `--grep` can only do the middle one.

Thanks!

aseemk:
Funny story: I added a feature today where I found it much easier to use a conditional everywhere (with `this.pending = true` if the conditional was true) rather than any `#skipfoo` tag.

Specifically, I could have done `--grep '#skipfoo' --invert`, but this is a cross-cutting concern, and doing this in a generic way, while still allowing helpers/wrappers around Mocha for business concerns, would have been challenging and error-prone.

So +1 to this feature now. =) Thanks!

noahsilas:
I wonder if there is an opportunity to do something more along the lines of RSpec's "User-defined metadata" tooling: https://www.relishapp.com/rspec/rspec-core/docs/metadata/user-defined-metadata. It seems like tags provide a fair amount of power, but have some limitations:

- Tags seem to be additive only - that is, a nested context can add a tag, but can't remove one that was applied by its parent. While this may seem convoluted, I can imagine scenarios where (for instance) shared behaviors may want to remove a "slow" tag.

- Allowing more structured metadata opens up new extension options. As an example, writing tests for an API requiring an API key to authenticate could make use of something like this:

  ```
  // common test suite setup
  beforeEach(function () {
    let auth = this.test.meta.authenticatedWith || {};
    if (auth.apiKey) {
      httpClient.defaults({ headers: {'Authorization': 'APIKEY ' + auth.apiKey }});
    }
  });
  
  // tests:
  describe("POST /widgets", function () {
    it("fails(401 unauthorized)", ...);
    context("logged in", { authenticatedWith: { apiKey: '123' }}, function () {
      it('succeeds', ...)
    });
  });
  ```

boneskull:
that could be useful, but rspec doesn't seem to have any filtering functionality w the metadata (though i only skimmed). "tags" are absolutely for filtering, and expanding that to user-defined metadata isn't too bad--except we'd need to get clever on the CLI end.

boneskull:
wrong button. anyway this is on hold until we hammer down the roadmap. we will have a forum for discussion about what goes in the roadmap.

alexjv89:
+1

sn1ckers:
this would be a really nice feature to have, we are in need to run tests based on inverted form factor '@not_mobile', '@not_tablet' for test that don't support all form factors and also other tags, like slow/fast etc.

I have tried to use grep but can't find any good solution to achieve what I wan't to do above.

PR #1447, opened 2014-11-25
Decorate failed hook titles with test title

### LINKED ISSUE #1230: More descriptive beforeEach/afterEach messages
I would like a failure in an `beforeEach` or `afterEach` to output which test it ran afterwards. For example:

```
describe('My Object', function() {
  var myObj;
  beforeEach(function() {
    myObj = new MyObject();
  });
  afterEach(function() {
    myObj.ensureClean(); // May throw an error
  });
  it('should do something', function() {
    myObj.doSomething();
  });
  it('should do something else', function() {
    myObj.doSomethingElse();
  });
});
```

If `ensureClean` throws an error, I don't know which test corrupted the object, since the error message gives me something generic:

```
  1) My Object "after each" hook:
     Error: The object was not clean
     <stack>
```

It would be great if the message said something like this:

```
  1) My Object "after each" hook after "should do something"
     Error: The object was not clean
     <stack>
```

I frequently put global checks into the afterEach, maybe that's not the right way to go about this?

duncanbeevers:
Fixes #1230

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -245,6 +245,11 @@ Runner.prototype.fail = function(test, err) {
  * @param {Error} err
  */
 Runner.prototype.failHook = function(hook, err) {
+  if (hook.ctx && hook.ctx.currentTest) {
+    hook.originalTitle = hook.originalTitle || hook.title;
+    hook.title = hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
+  }
+
   this.fail(hook, err);
   if (this.suite.bail()) {
     this.emit('end');
--- a/test/runner.js
+++ b/test/runner.js
@@ -275,6 +275,19 @@ describe('Runner', function(){
       runner.failures.should.equal(2);
     })
 
+    it('should augment hook title with current test title', function(){
+      var hook = {
+        title: '"before each" hook',
+        ctx: { currentTest: new Test('should behave') }
+      };
+      runner.failHook(hook, {});
+      hook.title.should.equal('"before each" hook for "should behave"');
+
+      hook.ctx.currentTest = new Test('should obey');
+      runner.failHook(hook, {});
+      hook.title.should.equal('"before each" hook for "should obey"');
+    })
+
     it('should emit "fail"', function(done){
       var hook = {}, err = {};
       runner.on('fail', function(hook, err){


danielstjules:
While https://github.com/mochajs/mocha/pull/1448 avoids throwing more logic in runner.js, I much prefer the simplicity of this PR. It doesn't switch from using a property to a getter, let alone 3 additional getters spread across runnable/hook/suite. It also means looser coupling between hooks and specs, as it doesn't need to be aware of its title - the runner handles that presentation.

@boneskull Since it's a "short title", it only includes the the spec name for which it was hopefully associated with, and not the full title (suite name and all).

``` javascript
$ cat example.js
describe('example', function() {
  beforeEach(function(done) {
    setTimeout(done, 1000);
  });

  it('test', function(done) {
    setTimeout(done, 0);
  });
});

$ ./bin/mocha example.js

  ․

  0 passing (204ms)
  1 failing

  1) example "before each" hook for "test":
     Error: timeout of 200ms exceeded
      at null.<anonymous> (/Users/danielstjules/git/mocha/lib/runnable.js:158:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
```

jbnicolai:
@danielstjules you'd prefer this one over #1448?

askhogan:
Added https://github.com/mochajs/mocha/pull/1745 to handle the context loss of the filename.  I am using this pull request 1447 and my pull request in a project with success.  Seems like a simple way to solve this issue.  +1

danielstjules:
@jbnicolai I'm happy with either solution. It would be nice if we didn't require a getter in https://github.com/mochajs/mocha/pull/1448, and the title was just another property set by the constructor. But that would mean changing the [constructor's signature and injecting the ctx](https://github.com/mochajs/mocha/blob/7493bca76662318183e55294e906a4107433e20e/lib/suite.js#L171-L179), which isn't an option for BC as part of this major.

I think 1447, as-is, keeps things a bit simpler. 1448 results in 3 name related properties/methods: title, shortTitle, fullTitle. I feel like that's a bit more confusing.

danielstjules:
@boneskull Given [the example](https://github.com/mochajs/mocha/pull/1447#issuecomment-80090072), what's your take on these two proposed solutions?

jbnicolai:
@mochajs/mocha I say we merge this one then :)

boneskull:
good

boneskull:
@duncanbeevers thanks

PR #1448, opened 2014-11-25
Runnable short title

### LINKED ISSUE #1230: More descriptive beforeEach/afterEach messages

duncanbeevers:
Fixes #1230

In this approach, rather than modifying the hook titles at failure time, a run-time `shortTitle` method is exposed on `Suite`, `Runnable`, and `Hook`.

In the case of a `Suite` and a `Runnable`, this method returns the current value of `title` attribute.
In the case of a `Hook` running in the context of a test, the `shortTitle` will reflect both the hook's own title and the title of the currently-executing test.


jbnicolai:
Is there anyhting holding of a merge, @boneskull @danielstjules? LGTM :)

@duncanbeevers thanks for the effort!

jbnicolai:
Closing in favor of #1447

PR #1450, opened 2014-11-25
Async errors from sync tests

duncanbeevers:
We recently encountered an issue where a test was written synchronously (no `done` argument) but failed to return a promise.

The test in question was failing an assertion, but after its success was already reported. The failure was tallied, and so we saw one failure reported in the UI, but had no connection between that number and the source of the failure.

This change first exposes a `wasAlreadyDone` property on the test instance, and second, re-displays the failing test in the html reporter when the runner emits a `fail` event.

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -114,7 +114,7 @@ function HTML(runner) {
   });
 
   runner.on('fail', function(test, err){
-    if ('hook' == test.type) runner.emit('test end', test);
+    if ('hook' == test.type || test.wasAlreadyDone) runner.emit('test end', test);
   });
 
   runner.on('test end', function(test){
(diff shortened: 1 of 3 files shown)


danielstjules:
Appreciate you proposing this fix! I'm gonna close this for now though, as I believe the latest release handles things as you intended :)

```javascript
$ cat example.js
it('should be true', function () {
  setTimeout(function () {
    throw new Error('oops');
  }, 10);
});

it('delay...', function(done) {
  setTimeout(done, 100);
});

$ ./bin/mocha example.js

  ․․

  1 passing (26ms)
  1 failing

  1)  delay...:
     Uncaught Error: oops
      at null._onTimeout (example.js:3:11)
```

screen shot 2015-03-13 at 11 19 42 pm

Of course, it gets tied to the wrong test in this example, though the stack trace tells you where to really look. Fixing that is probably outside the scope of this PR.

Thanks again!

duncanbeevers:
Thanks for following up @danielstjules

danielstjules:
No problem :) Glad it works

PR #1484, opened 2014-12-31
fix(LICENSE): update copyright year

a8m:


jbnicolai:
Thanks for the PR. To be honest, from a legal perspective updating this is useless-to-detrimental, see for instance the discussion [here](https://news.ycombinator.com/item?id=8808907). On the other hand, people use if as a heuristic when judging wether an open-source project is still 'alive', so might as well update it.

PR #1488, opened 2015-01-04
utils.stringify should handle objects without an Object prototype

BenoitZugmeyer:
I use a lot of prototype-less objects (`Object.create(null)`) in my code. Sadly, mocha fails to display the diff when an error occurs.


travisjeffery:
Thanks

Issue #1481, opened 2014-12-23
describe.only matches more that one describe test suite


DinisCruz:
If I have these two test suites (just showing the first one)

```coffee
describe 'Boolean',->

  it 'is_True',->
    expect(true.is_True    ).to.be.an('Function')
    expect(false.is_True   ).to.be.an('Function')
    expect(true.is_True()  ).to.be.true
    expect(false.is_True() ).to.be.false
    expect((1==1).is_True()).to.be.true
    expect((1==2).is_True()).to.be.false
    expect((1!=1).is_True()).to.be.false
``` 

and

```coffee
describe 'Assert | Boolean',->
  it 'assert_Is_True', ->
    true.assert_Is_True.assert_Is_Function()
    true.assert_Is_True().assert_Is_True()
    (-> false.assert_Is_True()).assert_Throws()
```

If I add **.only** to the first one : 
```coffee
describe.only 'Boolean',->
```

both will now be executed:

![image](https://cloud.githubusercontent.com/assets/656739/5537286/e7824a9e-8a96-11e4-9b96-b1643bd4ae93.png)

![image](https://cloud.githubusercontent.com/assets/656739/5537292/06cfd97a-8a97-11e4-9b18-7521030aab68.png)

and I would expect only the first one to execute

I assuming that mocha behind the scenes is not keeping a pointer to the actual tests to execute, it is just keeping the name/string, which is then searched on all tests/describes loaded (is that correct?)

tylerhjones:
https://github.com/mochajs/mocha/pull/1492

tylerhjones:
Try that.

twolfson:
@tylerhjones That solution will fit most cases but not if there are 2 cases with the same `describe` block and I use a `.only` on one of them.

```js
// We should only run this block
describe.only('A server', function () {
  describe('receiving an HTTP request', function () {
    it('replies successfully', function () {
    });
  });
});

describe('A server', function () { /* Similar setup but different test */ });
```

We need to solve it from file and position within the file. That is a valid unique identifer for each block.

danielstjules:
So far we've had 3 candidate fixes for this issue: https://github.com/mochajs/mocha/pull/1492 https://github.com/mochajs/mocha/pull/1595 and https://github.com/mochajs/mocha/pull/1591

If no one else has jumped on this already, I'll be reviewing all 3 tomorrow morning to hopefully get this closed up :)

demisx:
+1

[...]

PR #1492, opened 2015-01-07
.only patch

tylerhjones:
patch for describe . only in response to https://github.com/mochajs/mocha/issues/1481

### DIFF
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -96,7 +96,8 @@ module.exports = function(suite){
 
     context.describe.only = function(title, fn){
       var suite = context.describe(title, fn);
-      mocha.grep(suite.fullTitle());
+      var reString = '^' + escapeRe(suite.fullTitle()) + '$';
+      mocha.grep(new RegExp(reString));
       return suite;
     };
 
--- a/lib/interfaces/qunit.js
+++ b/lib/interfaces/qunit.js
@@ -87,7 +87,8 @@ module.exports = function(suite){
 
     context.suite.only = function(title, fn){
       var suite = context.suite(title, fn);
-      mocha.grep(suite.fullTitle());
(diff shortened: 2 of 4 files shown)


dasilvacontin:
+1. Patch is an improvement until we develop the full fix.

bjornstar:
LGTM, I think that it's a good idea to make `describe.only` match the behavior of `it.only`.

dasilvacontin:
This patch doesn't work (yay our tests). You are only patching one of the interfaces and you are doing it in the *compiled* mocha file. The compiled mocha file is auto-generated from the lib for the releases.

You have to patch the following files:
* `interfaces/bdd.js`
* `interfaces/qunit.js`
* `interfaces/tdd.js`

dasilvacontin:
ping @tylerhjones

tylerhjones:
thanks for the info, ill work on it later tonight MST

dasilvacontin:
Cool! I was just pinging in case you missed my message.

tylerhjones:
This currently passes 'npm test', but I have not yet added a separate test to assert that this .only bug does not appear again.

dasilvacontin:
The `mocha.js` file at the root of the repo shouldn't be modified; it is compiled from the source whenever a release is prepared.

Regarding test location, I should suggest adding tests for each of the interfaces at `test/acceptance/misc/only`.

Please squash your commits into a single one after the changes. Thanks!

danielstjules:
Thanks for the PR! For issue #1481, it looks like the team has decided to go with https://github.com/mochajs/mocha/pull/1591 which fixes some related `only` bugs due to the use of RegExps. Appreciate you taking the time to work on this though!

Vanuan:
I don't think that fixing multiple issues with one pull request is a good idea.
This is a good PR, it's just missing some tests.

Vanuan:
#1591 is closed

Issue #1229, opened 2014-05-28
Diff Viewer Verbosity with -0


skeggse:
As per chaijs/chai#223, it might make sense to display `-0` in mocha's diff viewer if the value is actually `-0` and not `+0`.

Thoughts?

a8m:
fixed! see: #1493
/cc @boneskull @jbnicolai

jbnicolai:
@a8m fantastic!

Issue #1132, opened 2014-02-12
Diffs for buffers


vkurchatkin:
Diffs  for buffers seem to be broken:

``` javascript
require('should')

describe('comparing buffers', function () {
  it('should produce correct diff', function () {
    Buffer([0x01]).should.eql(Buffer([0x02]))
  })
})
```

This code produces the following:

```
  + expected - actual

       {
      +  "0": 2,
      -  "0": 1,
         "length": 1,
      +  "offset": 5752,
      -  "offset": 5744,
         "parent": {
           "0": 47,
           "1": 42,
           "2": 33,
```

It's almost ok, except I can see `offset` and `parent`, but all I really care is data.
But this example shows the worst scenario:

``` javascript
require('should')

describe('comparing buffers', function () {
  it('should produce correct diff', function () {
    Buffer([0x01]).should.eql(Buffer(0))
  })
})
```

Not only this diff is veeery long (it shows the whole `parent`, which is the pool, I assume), but also it takes significant time to be generated.

So, the question is, shouldn't diffs for `Buffer`s be treated specially, to make them really useful?

PR #1493, opened 2015-01-07
[Fix]: Diff viewer(#1229, #1132, ...)

a8m:
This PR should fix some issues in the __Diff viewer__ that depends on the `stringify` function
for example: 
* #1229:
```js
expect(-0).eql(0);  // assert with empty representation
expect({ n: -0 }).eql({ n: 0 });  // same with the canonical behavior
expect(-0).eql(1);
/* representation:
  +expected: 1
  -actual: 0
```
* buffer diff #1132 :
```js
expect(new Buffer([0x01])).eql(new Buffer([0x02])); // asset with empty representation
```
* Infinity,NaN:
```js
expect({ infi: Infinity, nan: NaN }).eql({});
/* representation:
-{
-  "infi": null
-  "nan": null
-}
this happens because JSON.stringify(NaN/Infinity) returns `null`(could fixed with `replacer` function)
```
* Date, RegExp, Function and Circular represents as a string:
```js
var actual = {
  string: "string",
  regex: /^[a-z]/,
  func: function() {},
  date: new Date(0)
};
actual.circular =  actual;
expect(actual).eql({});
/* should represent without the `"` sign(that indicating to string representation)
 -{
 -  "circular": "[Circular]"
 -  "date": "[Date: 1970-01-01T00:00:00.000Z]"
 -  "func": "[Function]"
 -  "regex": "/^[a-z]/"
 -  "string": "string"
 -}

- This PR fixed that and the result will be like so:
 -{
 -  "circular": [Circular]
 -  "date": [Date: 1970-01-01T00:00:00.000Z]
 -  "func": [Function]
 -  "regex": /^[a-z]/
 -  "string": "string"
 -}
```
I changed the current specs, created more tests to get full coverage and did some DRY refactoring

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -175,7 +175,6 @@ exports.list = function(failures){
     if (err.uncaught) {
       msg = 'Uncaught ' + msg;
     }
-
     // explicitly show diff
     if (err.showDiff && sameType(actual, expected)) {
 
@@ -386,7 +385,7 @@ function unifiedDiff(err, escape) {
   function notBlank(line) {
     return line != null;
   }
-  msg = diff.createPatch('string', err.actual, err.expected);
+  var msg = diff.createPatch('string', err.actual, err.expected);
   var lines = msg.split('\n').splice(4);
   return '\n      '
          + colorLines('diff added',   '+ expected') + ' '
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -127,7 +127,7 @@ exports.filter = function(arr, fn){
 
 exports.keys = Object.keys || function(obj) {
   var keys = []
-    , has = Object.prototype.hasOwnProperty // for `window` on <=IE8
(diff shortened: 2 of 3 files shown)


travisjeffery:
broken tests?

a8m:
@travisjeffery since __v0.11__ node changes the `.toJSON()` implementation, I fixed that in 91c899400d85d8a99c13b6507ed8dcfbe303120c.
thanks

a8m:
@travisjeffery the tests was failed on **0.8** and **0.11** versions,  I fixed that in 4011032a01bdf89a85bf49875f0d78382cdec3a4.
thanks

a8m:
//cc @boneskull

travisjeffery:
@a8m can you squash these commits? thanks

a8m:
@travisjeffery done.

a8m:
Awesome @travisjeffery 
Thx

PR #1497, opened 2015-01-14
Only use colors if the $TERM has been set to allow for colors

parkr:
This environment variable has been used historically to tell programs whether the
program may write escape sequences which encode color to stdout and stderr.

TTY detection only determines whether the file descriptor is interactive. It says nothing about
whether it's fit for colors.

Calling `mocha` from within vim, for example, causes some problems:

screen shot 2015-01-13 at 4 16 52 pm

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -33,7 +33,7 @@ exports = module.exports = Base;
  * Enable coloring by default.
  */
 
-exports.useColors = isatty || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = (isatty && /color/.test(process.env['TERM'])) || (process.env.MOCHA_COLORS !== undefined);
 
 /**
  * Inline diffs instead of +/-


fixe:
👍

dasilvacontin:
Hi @parkr, thanks for the PR. I'd like to test/reproduce this before merging. What is your vim config or plugins to split with a shell inside vim? Also, could you add test(s)?

parkr:
@dasilvacontin That's just [`!npm test`](https://github.com/parkr/dotfiles/blob/master/vim/vimrc.symlink#L193), essentially.

parkr:
`useColors` isn't tested at all.

parkr:
Furthermore, it cannot be easily tested in the same runtime, because it's not re-evaluated when `TERM` changes.

dasilvacontin:
@parkr I know about the interactive shell, it's just that I saw your vim split, and I thought you were using a different method. 

Shouldn't we parse the output of `infocmp` instead of checking the `TERM` var directly? e.g. `xterm` supports 8 colors, but doesn't have `color` in its name.

Okay, I'll let the tests rest for now..

parkr:
> Shouldn't we parse the output of infocmp instead of checking the TERM var directly? e.g. xterm supports 8 colors, but doesn't have color in its name.

Hm. Is `infocmp` available in every env, or is it *nix-only? Does it shut off support for Windows?

jbnicolai:
How about we start using [`supports-color`](https://github.com/sindresorhus/supports-color)?

parkr:
> How about we start using supports-color?

:100: . I've done the (minor) legwork for you: https://github.com/mochajs/mocha/pull/1545

dasilvacontin:
We close this PR then?

jbnicolai:
@dasilvacontin I'm still curious if you and others agree about adding `supports-color`. As @parkr says, it's light-weight and does exactly what we need, so I for one am in favor :smile:

dasilvacontin:
IMO I prefer delegating general functionality to other packages. Specially in this case, I could enumerate quite a few pros about adding `supports-color` as a dependency rather than 'hosting' our own solution. So a big :+1: from me. Maybe @boneskull or @travisjeffery change my mind though. :)

PR #1505, opened 2015-01-16
Fixed the display of the mocha stats bar

LaurentGoderre:
Before:
mocha_before

After:
mocha_after

### DIFF
--- a/mocha.css
+++ b/mocha.css
@@ -220,12 +220,16 @@ body {
   font-size: 12px;
   margin: 0;
   color: #888;
+  background-color: #FFF;
+  border: 1px solid #000;
+  box-shadow: 0 0 10px #333;
   z-index: 1;
 }
 
 #mocha-stats .progress {
   float: right;
   padding-top: 0;
+  height: auto;
 }
 
 #mocha-stats em {


dasilvacontin:
What is this PR fixing? Can you provide a simple .html/.css file where this PR is fixing visual issues? Are you sure the visual issues are not caused by your own custom css?

LaurentGoderre:
Maybe partially but it's not very safe to not specify a background color and relying that the banner will be a white background element

mocha.css
```diff
@@ -220,12 +220,16 @@ body {
...
+  background-color: #FFF;
+  border: 1px solid #000;
+  box-shadow: 0 0 10px #333;
   z-index: 1;
 }
 
 #mocha-stats .progress {
   float: right;
   padding-top: 0;
+  height: auto;
```
LaurentGoderre: This solve the display when bootstrap is used on the test page

mocha.css
```diff
@@ -220,12 +220,16 @@ body {
   font-size: 12px;
   margin: 0;
   color: #888;
+  background-color: #FFF;
+  border: 1px solid #000;
+  box-shadow: 0 0 10px #333;
```
LaurentGoderre: This ensures that the report block always has a white color otherwise, the text can become unreadable

boneskull:
Thanks, but we can't accept this kind of PR, as it is unintended use.  This change may break other people's styles in weird ways too..

LaurentGoderre:
Instead of `height:auto` can we at least add `overflow: visible;` here: https://github.com/mochajs/mocha/blob/master/mocha.css#L248?

boneskull:
@LaurentGoderre What bug does that fix?

LaurentGoderre:
Well the issue specifically is the collision in the class progress between Mocha and Bootstrap. Bootstrap sets the progress class with a height of 23px which truncates the canvas element which is 40px of height.

boneskull:
What's preventing you from adding a custom style to your page which adds to Mocha's styles?  I don't get it.

dasilvacontin:
@boneskull, it would be nice for it to be plug and play, right? Maybe we should wait to see the number of people requesting this, and if considerable, we could make the `.progress` class more specific to avoid Bootstrap messing with it.

LaurentGoderre:
@boneskull It's actually the way I am doing it right now but it would be nice for other people using Bootstrap to not need to add this fix

boneskull:
Sorry, I don't understand why you need Bootstrap CSS in a unit test.  There's no bug here, so there's no need to modify our CSS.

@dasilvacontin The number of people is one (1).  

This is a slippery slope.  Think of it this way: would Bootstrap accept a PR that modifies their CSS to work with `mocha.css`?  What would they say?

dasilvacontin:
Don't people also use mocha for functional or end-to-end testing?

@boneskull, I said 'wait' since I don't have any usage data.

dasilvacontin:
Bootstrap styles all `progress` elements as if they were progress bars. You could blame Mocha for not using a very concrete class declaration, or you could blame Bootstrap for making such assumption on such a vague class declaration. Or you could blame both.

LaurentGoderre:
Isn't it reasonable to assume that if people want to use Mocha in the browser, that they might want to use it with Bootstrap to test their project?

PR #1514, opened 2015-01-24
test(utils|ms|*): test existing units

a8m:
Add tests to existing units(that untested).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -58,7 +58,7 @@ exports.forEach = function(arr, fn, scope){
 exports.map = function(arr, fn, scope){
   var result = [];
   for (var i = 0, l = arr.length; i < l; i++)
-    result.push(fn.call(scope, arr[i], i));
+    result.push(fn.call(scope, arr[i], i, arr));
   return result;
 };
 
--- a/test/ms.js
+++ b/test/ms.js
@@ -1,0 +1,82 @@
+'use strict';
+var ms = require('../lib/ms');
+
+describe('.ms()', function() {
+  // Helpers
+  var time = {
+    minutes: function(n) { return n * 60 * 1000; },
+    hours: function (n) { return n * this.minutes(60); },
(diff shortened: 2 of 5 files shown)


lib/utils.js
```diff
@@ -58,7 +58,7 @@ exports.forEach = function(arr, fn, scope){
 exports.map = function(arr, fn, scope){
   var result = [];
   for (var i = 0, l = arr.length; i < l; i++)
-    result.push(fn.call(scope, arr[i], i));
+    result.push(fn.call(scope, arr[i], i, arr));
```
boneskull: What's this for?

travisjeffery: looks like this is just being done so the test has access to the original array? if so, just make a var for the array in the test and get rid of this

a8m: @travisjeffery, @boneskull because `.map()` used as [IE8 polyfill](https://github.com/mochajs/mocha/blob/master/lib/utils.js#L50), It's make sense to align it with the [spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).
i.e: callback should call with 3 arguments: elmement, index and the array was called upon.
Thx.

a8m: @boneskull @travisjeffery PTAL

jbnicolai:
LGTM :+1: 

@boneskull, @travisjeffery any reason to hold off on merging?

boneskull:
@jbnicolai lgtm.  What I'd really prefer though is not to hand-roll our own polyfills :stuck_out_tongue_closed_eyes:

jbnicolai:
@boneskull fully agree, but that should be a separate issue I suppose? #1330 perhaps

a8m:
Awesome.
thx @jbnicolai @boneskull

jbnicolai:
To the contrary, thanks for contributing @a8m!

Issue #1532, opened 2015-02-04
Rejecting a promise with an object gives a blank failure


jakecraige:
If a promise is rejected with a normal javascript object the test still fails, but with a blank message. This confused me for the longest time.. It would be great if we could have it at least `JSON.stringify` the object or alert the user that they should return an `Error` object like it does if you resolve with a string.

Test case to reproduce plus show current behavior using RSVP promise library:

```js
var RSVP = require('rsvp');

it('gives helpful message when rejected with string', function() {
  return RSVP.reject('an error message');
});
// => Error: the string "an error message" was thrown, throw an Error :) ... stack

it('gives helpful message when rejected with error', function() {
  return RSVP.reject(new Error('an error message'));
});
// => Error: an error message ... stack

it('gives blank message and no stack when rejected with object', function() {
  return RSVP.reject({});
});
// => nothing
```

dasilvacontin:
Confirmed.

dasilvacontin:
Closed via #1537.

boneskull:
@dasilvacontin @jakecraige 

Sorry I didn't see this earlier.  Please see my comment [here](https://github.com/mochajs/mocha/pull/1537/files#r24287807).

PR #1537, opened 2015-02-06
Improve error output when a test fails with a non-error

jakecraige:
closes issue #1532

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -8,7 +8,9 @@ var EventEmitter = require('events').EventEmitter
   , Test = require('./test')
   , utils = require('./utils')
   , filter = utils.filter
-  , keys = utils.keys;
+  , keys = utils.keys
+  , type = utils.type
+  , stringify = utils.stringify;
 
 /**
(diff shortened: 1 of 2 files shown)


dasilvacontin:
Thanks!

lib/runner.js
```diff
@@ -197,6 +199,8 @@ Runner.prototype.fail = function(test, err){
 
   if ('string' == typeof err) {
     err = new Error('the string "' + err + '" was thrown, throw an Error :)');
+  } else if (!(err instanceof Error)) {
+    err = new Error('the ' + type(err) + ' ' + stringify(err) + ' was thrown, throw an Error :)');
```
boneskull: Why is there a second branch of the conditional here?  Why not just remove the check for string, and simply display this message if `err` is not an `Error`?

dasilvacontin: True.

jakecraige: Thanks @boneskull PR #1532 submitted

PR #1545, opened 2015-02-10
reporters/base: use supports-color to detect colorable term

parkr:
This replaces https://github.com/mochajs/mocha/pull/1497.

Detecting whether a terminal is TTY is not enough
to ensure that it can also support color.
This module, supports-color, is a wholistic solution
which captures the greatest range of possible terminals
with the greatest accuracy. The module is extremely
lightweight and does not represent a burdern to Mocha
users.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -5,7 +5,8 @@
 var tty = require('tty')
   , diff = require('diff')
   , ms = require('../ms')
-  , utils = require('../utils');
+  , utils = require('../utils')
+  , supportsColor = require('supports-color');
 
 /**
  * Save timer references to avoid Sinon interfering (see GH-237).
@@ -33,7 +34,7 @@ exports = module.exports = Base;
  * Enable coloring by default.
  */
 
-exports.useColors = isatty || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = supportsColor || (process.env.MOCHA_COLORS !== undefined);
 
 /**
  * Inline diffs instead of +/-
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "glob": "3.2.3",
     "growl": "1.8.1",
     "jade": "0.26.3",
-    "mkdirp": "0.5.0"
+    "mkdirp": "0.5.0",
+    "supports-color": "~1.2.0"
   },
   "devDependencies": {
     "coffee-script": "~1.8.0",


dasilvacontin:
LGTM if we fix the build errors.

jbnicolai:
Looks like the problem is [this line](https://github.com/sindresorhus/supports-color/blob/master/package.json#L16). Seems `supports-color` chose to only support `node >= 10.0`, probably just to encourage adoption of a more modern node in the ecosystem.

I'll see if I can have it support `node >= 8.0.0`.

parkr:
I think @dasilvacontin is right – the `^` character is crazy. I remember being able to use `~` in older versions of node. My guess is this new CL will work.

dasilvacontin:
I would prefer using the caret character. Can't we just force a higher version of npm? Kinda what we ended up doing at mustache.js: https://github.com/janl/mustache.js/pull/421

parkr:
> I would prefer using the caret character.

What do you prefer about the caret character?

> Can't we just force a higher version of npm? Kinda what we ended up doing at mustache.js: janl/mustache.js#421

Force-install a newer version of NPM? That's doable as long as it's compatible with that version of node. When was the caret character introduced?

dasilvacontin:
https://github.com/rlidwka/sinopia/issues/49#issuecomment-36779628

With the caret character, we will be getting both minor releases and patches/bugfixes. With the tilde, as soon as they do either a major or a minor release we will stop automatically getting patches and bugfixes.

On a second look, it seems like the philosophy in mocha is manually updating dependencies, since a lot of people rely on this package for their tests.

lib/reporters/base.js
```diff
@@ -33,7 +34,7 @@ exports = module.exports = Base;
  * Enable coloring by default.
  */
 
-exports.useColors = isatty || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = (isatty && supportsColor) || (process.env.MOCHA_COLORS !== undefined);
```
sindresorhus: `supportsColor` already checks if stdout is TTY, so I think the `isatty` check could be dropped.

parkr: Corrected.

sindresorhus:
Any reason you're not just using [`chalk`](https://github.com/sindresorhus/chalk)?

jbnicolai:
@sindresorhus discussed in #1330 and #1325. I'm still all in favor.

boneskull:
Re: fuzzy dependencies (`^` vs `~`) --fine for `devDependencies`.  We should keep them exact for production dependencies, however.

Would prefer to use `chalk` or something.  Someone just needs to implement it

parkr:
Ok, this is starting to feel like a wild :bird: chase. Chalk looks like a pretty big lift – completely reworking the way writing to the console happens. Am I correct in this assumption?

travisjeffery:
@parkr yep, gonna merge this and we can talk about chalk somewhere else if necessary

parkr:
Thank you!

jonathandelgado:
There should really be a CLI flag for this too. This breaks colors for piping all together. Environment flags are great, but if you set it to true, you are back to square one of this PR for all the non-pipe stuff.

PR #1547, opened 2015-02-10
travis: use docker containers

parkr:
The docker containers are far more abundant and
offer vastly reduced boot time and some increased
CPU power. As long as you don't require 'sudo',
this is the way to go.

http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
 - '0.11'
 - '0.10'
 - '0.8'
+sudo: false
 notifications:
   slack:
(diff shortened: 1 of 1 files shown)


boneskull:
Cool idea.

dasilvacontin:
Goes. :shipit:

boneskull:
Unfortunately...this doesn't seem to be noticeably faster.

parkr:
The boot time isn't recorded, but is much much shorter. Added CPU won't make a huge difference with already-fast tests.

boneskull:
Well, I'll take your word for it.

PR #1550, opened 2015-02-14
index: Use custom Gitter badge from shields.io

TimothyGu:
Prettier, and more consistent with the general color scheme.

Old:
<a href="https://gitter.im/mochajs/mocha" alt="Gitter" class="badge"></a>

New:
<a href="https://gitter.im/mochajs/mocha" alt="Gitter" class="badge"></a></p>


dasilvacontin:
Hmm, to be honest, I like the "old" one better. It has Gitter's logo, the badge is consistent across other repos, ie easier to identify, and even though it doesn't match our color scheme, the badge looks better to me with the default colors.

TimothyGu:
@dasilvacontin I guess it's just personal preference, but for me at least right now the badge on mochajs.org looks utterly weird and out of place. I agree however that it might be harder to identify spontaneously, so closing.

dasilvacontin:
I see what you mean. But if we add more badges in the future, some of them are going to be green anyways, since it's the 'passing' color.

TimothyGu:
@dasilvacontin imo the default badge green is more friendly towards the mocha color on the website, but Gitter's green contains more blue so that's why it looks kind of weird.

PR #1553, opened 2015-02-15
Windows trayballoon support

ayhankuru:
Windows trayballon example
screenshot


boneskull:
Thanks for the PR!

Please read `CONTRIBUTING.md`. You would have seen that the best course of action for a feature PR is to open an issue first for a discussion, as this may save everyone time. 

We could have explained two things:

- this functionality is better implemented as a reporter
- we will likely not be accepting more reporters into core

I'm going to close this PR, but it's a good idea. I encourage you to implement this as a reporter in a fork, but keep in mind the proposed plugin API would likely break compatibility when it is delivered. 

As an alternative something like `grunt` with `grunt-notify` may be a good bet.

ayhankuru:
You are right about the reasons thanks for the reply :+1:

Issue #1359, opened 2014-09-13
coding style


boneskull:
I'm unable to really get a handle on it, and what with all the PRs over the years, the codebase seems pretty inconsistent.

We have leading commas in some places, not in others, semi-colons here, no semi-colons there, blocks here and no blocks there.  

I'm proposing: *let's clean it up!*

We could go whatever direction, but in the interest of full disclosure *my coding style is different than TJ's*.  My style aligns pretty closely with [Felix's Style Guide](https://github.com/felixge/node-style-guide), but I'm more in the "Crockford" camp than he is.

> Some differences, *none of which I am willing to go to war over*: 
- I declare all my `var`s at the top of each function, as one statement, each separated by newlines + commas.
- I tend to use `foo_bar` for a variable name, and `fooBar` for a function name
- I tend to write ternary statements on one line unless they are hairy.
- I use JSDoc docstrings where appropriate, instead of always using slashes
- I will use setters.

However...

I would *strongly prefer* to always use semi-colons,  because it is more friendly to developers who don't know how to *not* use semi-colons.  (We get a lot of PRs.)

I would *strongly prefer* using curly braces in conditionals and loops, because I don't trust myself to write a complex "if/else if/else if/else" conditional without them.

I *strongly prefer* not to using leading commas, because it is untoward and impure.

We can set up `.jshintrc` and go with that; add a `jshint` target to the `Makefile`.  Usually I have one for production code, and another for test code (which will have different global vars/functions).  We can specify `browser: true` and `node: true` and solve the issue of having Mocha run in dual contexts.

Also an [`.editorconfig`](http://editorconfig.org/) would be helpful.

As far as actually reformatting things, I use PyCharm which is pretty smart about it, and could turn this around pretty quickly.  

So my questions are:

1.  Should we enforce a style?
2.  If so, should we use [Felix's Style Guide](https://github.com/felixge/node-style-guide)?
3.  Some other guide?
4.  Anything style-related you feel strongly about?

dasilvacontin:
Bump. I'd vote to go with JSHint, and start deciding the code style rules. :)

ainthek:
With JSHint, defaults (empty .jshintrc)

267  Bad line breaking before ','.
  29  Comma warnings can be turned off with 'laxcomma'.
  23  Bad line breaking before '+'.
  16  Use '===' to compare with '0'.
  15  Missing semicolon.
  15  Bad line breaking before '?'.
  13  The '__proto__' property is deprecated.
[...]

PR #1555, opened 2015-02-16
[JSHint] lib/mocha.js with less errors contributes to #1359

ainthek:
I gave up to fix this in one round,
I will post incremental fixes, If you accept each and we will,end up with clean "own sources" 
then we can automate the process and set up the rules, before then, it would be the show stopper.

So PLEASE accept partial, pull requests, and one day we may get cleaner code base.

Thanx

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -21,7 +21,7 @@ function Context(){}
  */
 
 Context.prototype.runnable = function(runnable){
-  if (0 == arguments.length) return this._runnable;
+  if (! arguments.length) return this._runnable;
   this.test = this._runnable = runnable;
   return this;
 };
@@ -35,7 +35,7 @@ Context.prototype.runnable = function(runnable){
  */
 
 Context.prototype.timeout = function(ms){
-  if (arguments.length === 0) return this.runnable().timeout();
+  if (! arguments.length ) return this.runnable().timeout();
   this.runnable().timeout(ms);
   return this;
 };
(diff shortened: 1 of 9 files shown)


travisjeffery:
Thanks, but personally don't like some of the config here. will add something to fix this shortly.

danielstjules:
Closing for now as the PR has been inactive for almost a month, and Travis expressed an interest in enforcing different conventions. Appreciate your work on this though! I'm sure the team is in agreement that it would be nice to converge on an accepted style throughout the codebase, and for future work. :)

boneskull:
Fwiw most if not all of the code changes look fine to me

danielstjules:
The second change stuck out:
``` javascript
-  if (arguments.length === 0) return this.runnable().timeout();
+  if (! arguments.length ) return this.runnable().timeout();
```

Adding a space before the closing brace of that conditional expression. As well as after the logical not. I personally like:

``` javascript
-  if (arguments.length === 0) return this.runnable().timeout();
+  if (!arguments.length) return this.runnable().timeout();
```

Though you may want to keep spaces after not operators. But that's totally up to @travisjeffery and yourself, as owners/maintainers of the project. :)

If you feel like the changes are fine, we can def reopen this. Would just take some rebasing, as there's a few conflicts. E.g. https://github.com/mochajs/mocha/pull/1555/files#diff-aa849a970cef551664c12f04a4209f6fL167, and https://github.com/mochajs/mocha/pull/1555/files#diff-037110b12e45b56f6f5c48b9cf213ab6R226 which was added in https://github.com/mochajs/mocha/commit/00ef54cd41d46b1c29358942dc0874e2ae7da0ff#diff-037110b12e45b56f6f5c48b9cf213ab6L226

Otherwise, another solution would be to open a PR that discusses/establishes the .jshintrc configuration we'd like to use.

Issue #1558, opened 2015-02-18
Uncaught ReferenceError: Buffer is not defined [mocha.js:5833]


dustinbrownman:
Getting this error with the latest version of the `mocha.js` file. I'm using the browser interface. Looks like the line that causes this error was added in eab389f3f106290e8386520a2f3b8206c18c486e. I set up a [dummy app](http://github.com/dustinbrownman/test-mochajs) demonstrate the error. Just open up `spec-runner.html` and open up the developer tools. Any thoughts on how to fix this? (Thanks to @clemf for finding this)

boneskull:
Thanks.  If we don't start building with browserify soon I'm gonna jump off a cliff

callmevlad:
Seeing the same thing. Needs to be `if (typeof Buffer !== 'undefined' && ...)` instead of `if (Buffer && ...)`, though I'm still trying to track down another error (`failed to require "supports-color"`) that's downstream from that fix.

a8m:
> Needs to be `if (typeof Buffer !== 'undefined' && ...)` instead of `if (Buffer && ...)`

why it's matter @callmevlad ?

callmevlad:
@a8m Browsers handle this differently than V8 in Node.js - for example, try the following in your Chrome console:

Breaks:
`if (Bob) { console.info("Bob exists."); }` -> ` Uncaught ReferenceError: Bob is not defined`

Works:
`if (typeof Bob !== "undefined") { console.info("Bob exists."); }`

dustinbrownman:
Confirmed that this is still an issue. Updated dummy app with new `mocha.js` file and it's the same error, but on a different line. And @callmevlad is right, this is a browser issue. I've tested it in Chrome and Firefox and gotten the same error each time.

a8m:
@callmevlad makes sense, thx.

callmevlad:
Just added a PR for this: https://github.com/mochajs/mocha/pull/1560

PR #1559, opened 2015-02-18
fix(utils): issue #1558 + make

a8m:
fix issue #1558 and run build for browser version (`mocha.js`)


boneskull:
Thanks.  If we don't start building with browserify soon I'm gonna suicide

dasilvacontin:
Because 'jump off a cliff' wasn't expressive enough.

boneskull:
@dasilvacontin I don't like to beat around the bush

PR #1560, opened 2015-02-18
Fixing errors that prevent mocha.js from loading in the browser - fixes #1558

### LINKED ISSUE #1558: Uncaught ReferenceError: Buffer is not defined [mocha.js:5833]

callmevlad:
Fixing a few issues caused by recent commits that break when mocha.js is loaded in the browser:

https://github.com/mochajs/mocha/commit/eab389f3f106290e8386520a2f3b8206c18c486e introduced `require('supports-color')`, which is not available in the browser.

https://github.com/mochajs/mocha/commit/00ef54cd41d46b1c29358942dc0874e2ae7da0ff added `if(Buffer && ...)` usage, which doesn't work in browsers.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -6,7 +6,7 @@ var tty = require('tty')
   , diff = require('diff')
   , ms = require('../ms')
   , utils = require('../utils')
-  , supportsColor = require('supports-color');
+  , supportsColor = process.env ? require('supports-color') : null;
 
 /**
  * Save timer references to avoid Sinon interfering (see GH-237).
@@ -31,10 +31,12 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 exports = module.exports = Base;
 
 /**
- * Enable coloring by default.
+ * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors = supportsColor || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = process.env
+  ? (supportsColor || (process.env.MOCHA_COLORS !== undefined))
+  : false;
 
 /**
  * Inline diffs instead of +/-
--- a/mocha.js
+++ b/mocha.js
@@ -1974,7 +1974,7 @@ var tty = require('browser/tty')
   , diff = require('browser/diff')
   , ms = require('../ms')
   , utils = require('../utils')
-  , supportsColor = require('supports-color');
+  , supportsColor = process.env ? require('supports-color') : null;
 
 /**
  * Save timer references to avoid Sinon interfering (see GH-237).
@@ -1999,10 +1999,12 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 exports = module.exports = Base;
 
 /**
- * Enable coloring by default.
+ * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors = supportsColor || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = process.env
+  ? (supportsColor || (process.env.MOCHA_COLORS !== undefined))
+  : false;
 
 /**
  * Inline diffs instead of +/-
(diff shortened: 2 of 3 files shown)


dasilvacontin:
:+1:

mocha.js
```diff
@@ -1999,10 +1999,10 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 exports = module.exports = Base;
 
 /**
- * Enable coloring by default.
+ * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors = supportsColor || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = process.env ? (supportsColor || (process.env.MOCHA_COLORS !== undefined)) : false;
```
a8m: **style issue:** this line too long(102 chars), imo you should stay with same code-style
e.g: [1](https://github.com/mochajs/mocha/blob/master/lib/reporters/base.js#L111-L114), [2](https://github.com/mochajs/mocha/blob/master/lib/utils.js#L442-L444)

mocha.js
```diff
@@ -5845,7 +5845,7 @@ var isArray = Array.isArray || function (obj) {
  * Buffer.prototype.toJSON polyfill
  * @type {Function}
  */
-if(Buffer && Buffer.prototype) {
+if(typeof Buffer !== 'undefined' && Buffer.prototype) {
```
a8m: LGTM, see #1558

travisjeffery:
fyi to all new collaborators, changes always need to happen outside of mocha.js. the mocha.js is compiled and the next time someone built it these changes would get blown away

boneskull:
@callmevlad That means you'll need to change this PR to update the source files, not `mocha.js` itself.

mocha.js
travisjeffery: i don't think you need to check this either since supportsColor will be null/undefined anyway

travisjeffery: this logic could also be done as 

`process.env && (supportsColor || (process.env.MOCHA_COLORS !== undefined)`

callmevlad: Thanks! Just a quick question - is `process.env` the correct way to detect whether we're in a non-browser environment? Not sure if `isatty` is what I should be keying off of...

mocha.js
callmevlad: Doh, rookie mistake, will fix.

mocha.js
a8m: [browser/isatty](https://github.com/mochajs/mocha/blob/master/lib/browser/tty.js#L2) no.
mm.. I'll take a look

travisjeffery: no, and i don't think you need to change this check. cause supportsColor will be null/undefined and process.env.MOCHA_COLORS will definitely be undefined.

callmevlad:
@travisjeffery Apologies for the rookie mistake :-\

@boneskull Just pushed updated code with recommended changes.

travisjeffery:
@callmevlad np! that was more for the other new collaborators who've been working on mocha for longer

lib/reporters/base.js
```diff
@@ -31,10 +31,12 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 exports = module.exports = Base;
 
 /**
- * Enable coloring by default.
+ * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors = supportsColor || (process.env.MOCHA_COLORS !== undefined);
+exports.useColors = process.env
```
travisjeffery: i don't think you need this check here since by your check at line 9, `supportsColor` will be undefined/null

callmevlad: If `supportsColor` is false, it falls through to the next check, and in the browser `process.env` is undefined so `process.env.MOCHA_COLORS` fails.

dasilvacontin:
Yeah, like me. I need to be more careful reviewing PRs.

dasilvacontin:
Bear in mind that, in the browser, `process.env` throws `ReferenceError: process is not defined`.

dasilvacontin:
Never mind, I didn't know about the node shims at https://github.com/mochajs/mocha/blob/master/mocha.js#L6328-L6330

PR #1561, opened 2015-02-20
optionally set suite name in xunit reporter

davebrown:
My use case is that I am running the same tests in different browsers, and so it would be useful for reporting in our CI system to have the suites named "ie_11" and "firefox_33", etc.


jbnicolai:
Could you add documentation on `suiteName` (`git grep suiteName` returns nothing) and a testcase?

hollomancer:
Closed for inactivity.

PR #1562, opened 2015-02-21
Defensive code for Buffer

kamn:
Was trying to build and get working in browser and ran into an issue here. Followed the style that is elsewhere with Buffer existance checking.
```javascript
typeof Buffer !== 'undefined'
```

Also the support for `supports-colors` is the error I get next after this fix in the browser. Or maybe I am not makeing the project correctly?

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -173,7 +173,7 @@ var isArray = Array.isArray || function (obj) {
  * Buffer.prototype.toJSON polyfill
  * @type {Function}
  */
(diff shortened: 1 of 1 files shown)


travisjeffery:
Hey there, I just merged in a pr which dups your fix here and also fixes that supports-color issue. Thanks though!

kamn:
Ops, should of checked the issue list before I submitted this. Thanks for the quick response!

Issue #545, opened 2012-08-22
Shorten stack traces


haraldrudell:
80% of the AssertionError stack trace is uninteresting mocha internal code

Suggestion:
suppress all frames from Test.Runnable.run and previous

Suggestion:
1. Reformat the stack trace so the first text is base-file:line:column, then Object.function, then directory.
2. Directory can be shortened to be relative to the deploy folder (ie. where app.js is)

---

I am using node assert, ui exports and spec reporter

toolbear:
@rstacruz :clap:

aronwoost:
So awesome!

dasilvacontin:
Any news about merging `mocha-clean` into `mocha`?

toolbear:
@dasilvacontin Fork it, merge it, create PR, advertise it, use it, fix it, then either everyone's on your fork or it gets merged in. Enjoy your burrito.

dasilvacontin:
I want a burrito. Let's see.

toolbear:
I think a lot of us stopped at @rstacruz's fix and enjoyed our burritos then. I'd happily try out your fork and enjoy a burrito deluxe.

a8m:
I think we could close this issue @travisjeffery.

ambodi:
This is not fixed at all.

module.js:339
    throw err;
    ^

```bash
Error: Cannot find module '../../lib/camelize'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/ara/dev/iteam/dm-crawler/tests/unit/lib/camelize.js:5:16)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at /Users/ara/dev/iteam/dm-crawler/node_modules/mocha/lib/mocha.js:216:27
    at Array.forEach (native)
    at Mocha.loadFiles (/Users/ara/dev/iteam/dm-crawler/node_modules/mocha/lib/mocha.js:213:14)
    at Mocha.run (/Users/ara/dev/iteam/dm-crawler/node_modules/mocha/lib/mocha.js:453:10)
[...]

PR #1564, opened 2015-02-22
feat(runner/stack-trace): solve issue #545 + test

a8m:
following #545 discussion.
I added a stack-trace filter(based on @rstacruz `mocha-clean` module), ran `make` and added tests too.
If for some reason someone want to stay with full stack-trace, just run mocha with the `--full-trace` flag

### DIFF
--- a/mocha.js
+++ b/mocha.js
@@ -1486,7 +1486,7 @@ function Mocha(options) {
   this.bail(options.bail);
   this.reporter(options.reporter, options.reporterOptions);
   if (null != options.timeout) this.timeout(options.timeout);
-  this.useColors(options.useColors)
+  this.useColors(options.useColors);
   if (options.enableTimeouts !== null) this.enableTimeouts(options.enableTimeouts);
   if (options.slow) this.slow(options.slow);
 
@@ -1545,8 +1545,8 @@ Mocha.prototype.reporter = function(reporter, reporterOptions){
   } else {
     reporter = reporter || 'spec';
     var _reporter;
-    try { _reporter = require('./reporters/' + reporter); } catch (err) {};
-    if (!_reporter) try { _reporter = require(reporter); } catch (err) {};
+    try { _reporter = require('./reporters/' + reporter); } catch (err) {}
+    if (!_reporter) try { _reporter = require(reporter); } catch (err) {}
     if (!_reporter && reporter === 'teamcity')
       console.warn('The Teamcity reporter was moved to a package named ' +
         'mocha-teamcity-reporter ' +
@@ -1568,7 +1568,7 @@ Mocha.prototype.reporter = function(reporter, reporterOptions){
 Mocha.prototype.ui = function(name){
   name = name || 'bdd';
   this._ui = exports.interfaces[name];
-  if (!this._ui) try { this._ui = require(name); } catch (err) {};
+  if (!this._ui) try { this._ui = require(name); } catch (err) {}
   if (!this._ui) throw new Error('invalid interface "' + name + '"');
   this._ui = this._ui(this.suite);
   return this;
@@ -5249,23 +5249,23 @@ function filterLeaks(ok, globals) {
  * @api private
  */
 
- function extraGlobals() {
-  if (typeof(process) === 'object' &&
-      typeof(process.version) === 'string') {
+function extraGlobals() {
+ if (typeof(process) === 'object' &&
+     typeof(process.version) === 'string') {
 
-    var nodeVersion = process.version.split('.').reduce(function(a, v) {
-      return a << 8 | v;
-    });
+   var nodeVersion = process.version.split('.').reduce(function(a, v) {
+     return a << 8 | v;
+   });
 
-    // 'errno' was renamed to process._errno in v0.9.11.
+   // 'errno' was renamed to process._errno in v0.9.11.
 
-    if (nodeVersion < 0x00090B) {
-      return ['errno'];
-    }
-  }
-
-  return [];
+   if (nodeVersion < 0x00090B) {
+     return ['errno'];
+   }
  }
+
+ return [];
+}
 
 }); // module: runner.js
 
--- a/mocha.js
+++ b/mocha.js
@@ -1470,6 +1470,7 @@ function image(name) {
  *   - `bail` bail on the first test failure
  *   - `slow` milliseconds to wait before considering a test slow
  *   - `ignoreLeaks` ignore global leaks
+ *   - `fullTrace` display the full stack-trace on failing
  *   - `grep` string or regexp to filter tests with
  *
  * @param {Object} options
@@ -1670,6 +1671,18 @@ Mocha.prototype.checkLeaks = function(){
 };
 
 /**
+ * Display long stack-trace on failing
+ *
+ * @return {Mocha}
+ * @api public
+ */
+
+Mocha.prototype.fullTrace = function() {
+  this.options.fullStackTrace = true;
+  return this;
+};
+
+/**
  * Enable growl support.
  *
  * @return {Mocha}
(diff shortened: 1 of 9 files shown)


jancarloviray:
awesome. please merge this :) :+1: thanks @rstacruz.

danielstjules:
@a8m That's pretty sweet!
``` javascript
$ cat example.js
it('test', function(done) {
  done(new Error('Something went wrong!'));
});

$ ./bin/mocha example.js


  ․

  0 passing (4ms)
  1 failing

  1)  test:
     Error: Something went wrong!
      at Context.<anonymous> (/Users/danielstjules/git/mocha/example.js:2:8)
      at Test.Runnable.run (/Users/danielstjules/git/mocha/lib/runnable.js:233:15)
      at Runner.runTest (/Users/danielstjules/git/mocha/lib/runner.js:387:10)
      at /Users/danielstjules/git/mocha/lib/runner.js:470:12
      at next (/Users/danielstjules/git/mocha/lib/runner.js:312:14)
      at /Users/danielstjules/git/mocha/lib/runner.js:322:7
      at next (/Users/danielstjules/git/mocha/lib/runner.js:257:23)
      at Immediate._onImmediate (/Users/danielstjules/git/mocha/lib/runner.js:289:5)
      at processImmediate [as _immediateCallback] (timers.js:358:17)



$ git checkout stack-trace
Switched to branch 'stack-trace'
danielstjules:~/git/mocha (stack-trace)
$ ./bin/mocha example.js


  ․

  0 passing (7ms)
  1 failing

  1)  test:
     Error: Something went wrong!
      at Context.<anonymous> (example.js:2:8)
      at Test.Runnable.run (lib/runnable.js:229:15)
      at Runner.runTest (lib/runner.js:390:10)
      at lib/runner.js:473:12
      at next (lib/runner.js:315:14)
      at lib/runner.js:325:7
      at next (lib/runner.js:260:23)
      at Immediate._onImmediate (lib/runner.js:292:5)
```

In addition to the work here, I wonder if there'd be interest in removing parts of the trace that propagated from mocha itself. For example, with the test case above, I think the following output would be a lot nicer:

```
  1)  test:
     Error: Something went wrong!
      at Context.<anonymous> (example.js:2:8)
```

You'd just have to traverse the trace, bottom up, in search of the first line that wasn't a file in mocha. Once found, just drop the lines below. This way you're not hiding any information related to the user's code. I remember doing something similar with https://github.com/danielstjules/pho a while back.

a8m:
@danielstjules you get this trace because you run it from mocha itself (it's assume that is your code, and you care of it).
PTAL on the [mocha](https://github.com/a8m/mocha/blob/master/lib/utils.js#L656-L660) filter.

danielstjules:
Exactly what I hoped for. :) +1

```
$ mocha example.js

  ․

  0 passing (5ms)
  1 failing

  1)  test:
     Error: Something went wrong!
      at Context.<anonymous> (example.js:2:8)
```

mocha.js
```diff
@@ -5249,24 +5266,24 @@ function filterLeaks(ok, globals) {
...
+   });
 
-    if (nodeVersion < 0x00090B) {
-      return ['errno'];
-    }
-  }
+   // 'errno' was renamed to process._errno in v0.9.11.
 
-  return [];
+   if (nodeVersion < 0x00090B) {
```
rstacruz: perhaps [semver](npmjs.org/package/semver) would be a better fit here? (`semver.satisfies(process.version, '>= 0.9.11')`)

rstacruz:
If I may just chime in with API design input, perhaps `--stack-trace` is not the best choice for a name. It seems to imply that stack traces will not be shown unless it's turned on. Maybe it would be better as `--full-stack-trace`.

Good work!

mocha.js
a8m: maybe, but it's not part of this PR. see: [commit](https://github.com/mochajs/mocha/commit/736c743d8b2794cdb9823731afb937e2e3bf7174)
(I just fixed the indentation)

a8m:
Good point @rstacruz, 
`--full-stack-trace` is too long imo, maybe `--full-trace` could be better ?.

a8m:
**update:** renaming flag to: `--full-trace`

a8m:
update: renaming flag to: `--full-trace`

danielstjules:
Would you be willing to run `git reset HEAD~` and `git add -p` to keep your feature-related changes in a separate commit from style fixes? It'll keep blame a lot cleaner. And makes the PR quicker to review. I think @rstacruz's comment on `if (nodeVersion < 0x00090B) {` is an example of not being sure at a glance what's feature-related vs clean up.

a8m:
sure @danielstjules

a8m:
ping @danielstjules

danielstjules:
Sorry to nitpick, for some reason https://github.com/a8m/mocha/commit/8d90e59697b857f7b4e8d5f27a1404bc6a45a335 still contains changes to extraGlobals, along with a bunch of other (and totally valid) style changes.

Would you mind force-pushing https://github.com/danielstjules/mocha/commits/stack-trace onto your branch? I cleaned up that second commit a bit (went from 400 line changes, to 270)

a8m:
PTAL @danielstjules

danielstjules:
LGTM :) +1

cc @travisjeffery @boneskull @dasilvacontin

a8m:
@travisjeffery can you please take a look ?

travisjeffery:
i tried this and it didn't work for me. lemme try again...

travisjeffery:
ah nvm there we go, yup lgtm

a8m:
That's awesome, thx @travisjeffery

rstacruz:
Is this something that will land in 2.3.0?

I'll update [mocha-clean](https://github.com/rstacruz/mocha-clean) (the module this was based off of) to throw a warning for newer versions of mocha.

dasilvacontin:
@rstacruz supposedly. I'm very excited to see this published soon. :smile:

johannes-photobox:
:+1:

`mocha-clean/brief` is so much more readable!

myndzi:
This is actually super UNhelpful in certain cases (mainly async things); I'm running Mocha programmatically and can't seem to work out how to disable this. Specifying `fullTrace: true` or `fullStackTrace: true` to the Mocha constructor or calling `mocha.fullTrace()` don't seem to work. Halp?

myndzi:
Seems to be using `this` here: https://github.com/mochajs/mocha/blob/master/lib/runner.js#L209 -- not sure why, since the `this` object does not contain that option when it is enabled on mocha.

dasilvacontin:
>Seems to be using this here: https://github.com/mochajs/mocha/blob/master/lib/runner.js#L209 -- not sure why, since the this object does not contain that option when it is enabled on mocha.

It's set here: https://github.com/mochajs/mocha/blob/master/lib/mocha.js#L429

dasilvacontin:
`fullStackTrace: true` or `mocha.fullTrace()` should do the trick. Else, something is going wrong.

myndzi:
Indeed, something is going wrong. It doesn't do the trick.

dasilvacontin:
Can you share a Gist with sample code that reproduces the issue?

myndzi:
Not easily, but it may be indirectly my fault; I'll have to dig deeper. I don't see why the 'this' argument would be wrong and mocha still run at all, but I'll find out. It seemed such an obvious disconnect, that I figured at first that whoever submitted the patch just didn't test the opposite case thoroughly and made a logic error.

myndzi:
Yep, I got it sorted. It's weird to set the options and then just copy them over, seems very non-DRY. I had to monkey patch mocha.run to emit an extra event; looks like there's support in the code that will let me not have to do that anymore now. Sorry for the confusion and thanks for the prompt response.

In the opposite direction of my original comment, internal node filtering doesn't catch _stream_readable.js or net.js among other things; not sure if that is intentional. I'm pretty sure that all internal modules do not have paths, so that function could probably be simplified and made more forward-compatible too.

PR #1572, opened 2015-03-01
Prevent default browser behavior for failure/pass links

jschilli:
Changed href in HTML reporter to prevent default behavior. 

This can be an issue when using mocha in integration testing with a booted application (ember in my case)


jbnicolai:
@jbryson3 I suppose the reason is that the css styling is different when you leave the `href` (no cursor change on hover, for instance).

@jschilli merged, but changed the noop to `javascript:void(0);`

jschilli:
@jbnicolai - thanks - much appreciated

PR #1575, opened 2015-03-04
Add Promise & Coroutine support to Runnable

danielstjules:
Hi! Sorry in advance for not creating an issue to discuss this idea. I only just read the contribution guidelines. Would love to hear what people think about this though! I'm finding it pretty useful already. :)

So quick overview. This adds coroutine (via co) and promise support to runnables (specs and hooks). I really only wanted to take advantage of coroutines for my test suite, but promises were a freebie.

``` javascript
it('Example Coroutine', function* () {
  // Coroutines are free of callbacks!
  yield somePromise("It's beautiful");
});

it('A longer example', function* () {
  // Get user
  var user = yield getUserAsync(1);
  // Get all their posts
  var posts = yield user.getPostsAsync();

  // Get a unique list of category names to which
  // the posts belong
  var categories = posts.map(function() {
    return yield post.getCategoryName();
  }).sort().filter(function(name, pos) {
    return !pos || name != a[pos - 1];
  });

  categories.should.equal(['general', 'programming']);
});

it('Example Promise', new Promise(function(resolve, reject) {
  // Promises get double the callbacks
  console.log("I don't know why anyone would use promises for this");
  resolve();
}).then(function() {
  return Promise.resolve();
}));
```

Here's a bit more of a demo:

``` javascript

describe('This Pull Request', function() {
  it('enables the use of GeneratorFunctions for coroutines', function* () {
    var str = yield new Promise(function(res) {
      process.nextTick(function() {res('foo'); });
    });

    assert.equal(str, 'foo');
  });

  it('allowing us to avoid some nesting in specs', function* () {
    // We can use synchronous-looking logic!
    var str1 = yield Promise.resolve('foo');
    var str2 = yield Promise.resolve('foo');
    var str3 = yield new Promise(function(res) {
      process.nextTick(function() {res('bar'); });
    });

    // Showing an error
    assert.equal(str1, str2);
    assert.equal(str2, str3, 'Intentional error');
  });

  it('also enables the use of promises', new Promise(function(resolve, reject) {
    resolve();
  }));

  it('which is less important', new Promise(function(resolve, reject) {
    reject(new Error('Intentional error'));
  }));
});
```

```
$ ./mocha/bin/mocha exampleSpec.js


  This Pull Request
    ✓ enables the use of GeneratorFunctions for coroutines
    1) allowing us to avoid some nesting in specs
    ✓ also enables the use of promises
    2) which is less important


  2 passing (12ms)
  2 failing

  1) This Pull Request allowing us to avoid some nesting in specs:
     AssertionError: Intentional error
      at /Users/danielstjules/git/exampleSpec.js:22:12
      at GeneratorFunctionPrototype.next (native)
      at onFulfilled (/Users/danielstjules/git/mocha/node_modules/co/index.js:61:19)
      at runMicrotasksCallback (node.js:331:7)
      at process._tickCallback (node.js:350:11)

  2) This Pull Request which is less important:
     Error: Intentional error
      at /Users/danielstjules/git/exampleSpec.js:30:12
      at Suite.<anonymous> (/Users/danielstjules/git/exampleSpec.js:29:33)
      at context.describe.context.context (/Users/danielstjules/git/mocha/lib/interfaces/bdd.js:49:10)
      at Object.<anonymous> (/Users/danielstjules/git/exampleSpec.js:3:1)
      at Module._compile (module.js:444:26)
      at Object.Module._extensions..js (module.js:462:10)
      at Module.load (module.js:339:32)
      at Function.Module._load (module.js:294:12)
      at Module.require (module.js:349:17)
      at require (module.js:368:17)
      at /Users/danielstjules/git/mocha/lib/mocha.js:186:27
      at Array.forEach (native)
      at Mocha.loadFiles (/Users/danielstjules/git/mocha/lib/mocha.js:183:14)
      at Mocha.run (/Users/danielstjules/git/mocha/lib/mocha.js:404:31)
      at Object.<anonymous> (/Users/danielstjules/git/mocha/bin/_mocha:400:16)
      at Module._compile (module.js:444:26)
      at Object.Module._extensions..js (module.js:462:10)
      at Module.load (module.js:339:32)
      at Function.Module._load (module.js:294:12)
      at Function.Module.runMain (module.js:485:10)
      at startup (node.js:112:16)
      at node.js:863:3
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 script: travis_retry npm test
 node_js:
+- 'iojs'
 - '0.12'
 - '0.11'
 - '0.10'
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -5,6 +5,7 @@
 var EventEmitter = require('events').EventEmitter
   , debug = require('debug')('mocha:runnable')
   , Pending = require('./pending')
+  , co = require('co')
   , milliseconds = require('./ms')
   , utils = require('./utils');
 
@@ -40,8 +41,8 @@ module.exports = Runnable;
 
 function Runnable(title, fn) {
   this.title = title;
-  this.fn = fn;
-  this.async = fn && fn.length;
+  this.fn = this.normalize(fn);
+  this.async = this.fn && (this.fn.then || this.fn.length);
   this.sync = ! this.async;
   this._timeout = 2000;
   this._slow = 75;
@@ -55,6 +56,35 @@ function Runnable(title, fn) {
  */
 
 Runnable.prototype.__proto__ = EventEmitter.prototype;
+
+/**
+ * Wraps any passed Promise or GeneratorFunction in a thunk.
+ *
+ * @param {Function} fn
+ * @return {Function}
+ * @api private
+ */
+Runnable.prototype.normalize = function(fn) {
+  if (!fn) return fn;
+
+  // Promises need to be wrapped to avoid errors from use of fn.call
+  // with a non-promise scope
+  if (typeof fn.then === 'function') {
+    return function(done) {
+      fn.then(function() {
+        done();
+      }, function(err) {
+        done(err || new Error('Promise.reject called without error'));
+      });
+    };
+  } else if (fn.constructor && fn.constructor.name === 'GeneratorFunction') {
+    return function(done) {
+      co(fn).then(done, done);
+    };
+  }
+
+  return fn;
+};
 
 /**
  * Set & get timeout `ms`.
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "test": "make test-all"
   },
   "dependencies": {
+    "co": "~4.4.0",
     "commander": "2.3.0",
     "debug": "2.0.0",
     "diff": "1.0.8",
--- a/test/harmony/runnable.generators.js
+++ b/test/harmony/runnable.generators.js
@@ -1,0 +1,53 @@
+var mocha = require('../../')
+  , utils = mocha.utils
+  , Runnable = mocha.Runnable
+  , EventEmitter = require('events').EventEmitter
+  , should = require('should');
+
+module.exports = function() {
+  describe('when fn is a GeneratorFunction', function() {
+    it('should invoke the callback', function(done) {
+      var invoked = false;
+      var test = new Runnable('foo', function* () {
+        yield new Promise(function(resolve, reject) {
+          var cb = function() {
+            invoked = true;
+            resolve();
+          };
+          process.nextTick(cb);
+        });
+      });
+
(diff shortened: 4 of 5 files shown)


a8m:
@danielstjules since mocha supports node `>=0.8` but `co` require this [engines support](https://github.com/tj/co/blob/master/package.json#L27-L30)  this build will never pass.

a8m:
my false @danielstjules 
great

danielstjules:
@a8m Just looks like npm hit a wall on the one job for 0.8, complaining the version of co didn't exist. I guessed it may have been that the old version of npm didn't support version numbers like "^4.4.0", so went with "~4.4.0". Can't remember when that was introduced. Or it was just a fluke. :)

tj:
I'd wait for async/await personally, since that will invalidate all current hacks of similar things

travisjeffery:
+1

danielstjules:
Thanks for the input! I agree that async/await would eliminate the need for this, but isn't that a bit far off? The [proposal](https://github.com/lukehoban/ecmascript-asyncawait) is currently only in stage 1 of the 4-step TC39 process, along with [a few others](http://kangax.github.io/compat-table/es7/).

Historically speaking, harmony's taken a while for adoption (no real traction until 2011), though things have certainly picked up more recently. I've been assuming we won't see async/await for a year or more. 
If that's the case, this ~30 line change would allow devs to write code in a similar fashion in the meantime. Then when async/await support is announced, it's almost as easy as `s/function*/async function` & `s/yield/await` through their specs. 

With the increased adoption of coroutines and promises in new node projects, I'm seeing more and more devs dropping callbacks. The only place left is the test suite, where you'd have to resort to some additional nesting:

``` javascript
it(function() {
  co(function* () {
    yield fnAsync();
  });
});
```

And that works since promises have already been given special consideration in runnables: https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L261-L275 So why not coroutines? :)

Hope I didn't come across as pushy! I'll close this PR if that wasn't convincing at all haha.

travisjeffery:
i kinda hear you since in pretty much every project i drop in co-mocha, but at the same time it's super easy to just use co-mocha. there's also a diff between promises which are built-in to the language and co which is a 3rd party lib.

danielstjules:
Fair enough. :) I appreciate the explanation.

lygstate:
@travisjeffery co-mocha doesn't support for browsers.
That's why making co support in mocha is necessary.

dschinkel:
so...seems like especially since Node 4.0 is out this would be in even higher demand.  I too want this support

danielstjules:
Using co-mocha is easy enough. I think we should probably document some of these things on the website though, as they'll certainly become a common request

PR #1580, opened 2015-03-06
Site readme: update instructions for 2015

rstacruz:
A few notes:

* homebrew doesn't have the `make` package anymore
* homebrew requires Xcode CLI tools (it even installs it for you!), so having homebrew implies you have make
* Ubuntu doesn't come with make by default unless you install `build-essential`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,49 +11,23 @@ There's two:
 
 ### Mac OS X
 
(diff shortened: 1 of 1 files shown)


rstacruz:
...btw, a jekyll rewrite might be a good idea, since it's already got markdown :)

danielstjules:
Looks good :) Don't think anyone will mind if I merge this in. Thanks!

PR #1581, opened 2015-03-07
docs(website): update all relevant flags

a8m:
Since @travisjeffery releasing a new version, it's time to update the docs.


travisjeffery:
Sweet thanks

PR #1586, opened 2015-03-09
[Fix] Exclude not own properties when looping on options

anis:
I've been using [mocha-phantomjs][1] along with chai to unit test my Javascript libraries and it appears that Mocha ends up in error when one of those libraries adds a property to Object's prototype.

For instance, when adding a function myMethod to the prototype :
```javascript
Object.prototype.myMethod = function () { // ... };
```

running mocha-phantomjs outputs :
```sh
$ mocha-phantomjs my-runner.html
Unable to open file 'spec'
```

Excluding extra properties from the setup loop seems to solve the problem.

[1]: https://github.com/metaskills/mocha-phantomjs

### DIFF
--- a/mocha.js
+++ b/mocha.js
@@ -6437,7 +6437,11 @@ mocha.ui = function(ui){
 
 mocha.setup = function(opts){
   if ('string' == typeof opts) opts = { ui: opts };
-  for (var opt in opts) this[opt](opts[opt]);
+  for (var opt in opts) {
+    if (opts.hasOwnProperty(opt)) {
+        this[opt](opts[opt]);
+    }
+  }
   return this;
 };
 


a8m:
I'm trying to test it, but I'm not getting this error, can you provide a full example @anis ?

anis:
First of all, thanks for the quick reply! :)

### TL;DR
[Instructions for reproducing the error here][1]

### Reproducing the error
After digging into the issue a little deeper I got a better understanding of it: the actual issue is that the property we add to Object's prototype ___is called with a single argument, being itself!___

That means you'll get that "Unable to open file 'spec'" error if the property you add to the prototype:
* is not callable (_basically not a Function_)
* is a function that expects more than one argument
* is a function that processes its first argument in a way not adapted to a function

If you tried to add an empty function to the prototype, it will definitely not cause any visible error:
```javascript
// I'm not causing any trouble
Object.prototype.myMethod = function () {};
```

But go ahead and try these:

```javascript
// I'm supposed to be a function...
Object.prototype.myMethod = false;
```

```javascript
// I'm not supposed to get a string as an argument...
Object.prototype.myMethod = function (arg1) {
    return arg1.toLowerCase();
};
```

I took a little time to set-up an environment for reproducing the error. If you feel lazy or just can't manage to reproduce the problem yourself, feel free to get a look [at my branch][1] and follow the instructions in the README.

### Understanding the error
The basics:
* __mocha.setup()__ expects, as an input, a list of options (__opts__): each option being a pair of key/value.
* for each option key, __mocha__ should have a corresponding method (for the option "ui", there is a method mocha.ui())
* __mocha.setup()__ loops over each option, calls the corresponding method with a single argument: the option value

By adding a property to Object's prototype (let's say "myMethod"), you actually provide it to both __mocha__ and __opts__. Therefore, __mocha.setup()__ will consider "myMethod" as an option and end up executing this:

```javascript
mocha.myMethod(opts.myMethod)
```

This is an undesired behaviour that leads, most of the time, to the error "Unable to load file 'spec'".

[1]: https://github.com/anis/mocha/tree/fix-useOwnPropertyOnSetupOptions-demo/demo

a8m:
Thanks for this expanded explanation @anis :).
I get the point, and this can actually be solved with this patch.
But I'm not sure if we can mark it as a bug(in mocha at least), because adding enumerable properties to Object.prototype breaks all `for ... in` loops that don't have a `hasOwnProperty` check in them.

I'm curious to know what others think about this issue, for now I'll suggest you to use the [definedProperty](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) function.

anis:
Well, thanks for the advice but I am afraid I will have to stick with the Object.prototype inheritance strategy, for two main reasons essentially:
* my library is supposed to run on televisions, a very heterogeneous environment with a lot of custom/old/buggy browsers, and not all of them correctly support `defineProperty`
* it would cost (_really_) a lot, to redesign (_even slightly_) our library: it is already in production and consumed by over 20 applications

Anyway, not performing the `hasOwnProperty` check is considered a bad practice: most of the time it simply is semantically incorrect, just like in `mocha.setup()`.

boneskull:
> Anyway, not performing the hasOwnProperty check is considered a bad practice: most of the time it simply is semantically incorrect, just like in mocha.setup().

I agree with this.  The check must be done unless you specifically mean *not* to, unfortunately.  

I'll merge this.

Thanks!

boneskull:
Ergh, well, I'll give it a go, since it needs to be rebased.

boneskull:
merged via 8a607872c1ea63ac745f3782a1f08cd0885d06c6

Thanks!!

PR #1590, opened 2015-03-09
docs(README): update repo summary

a8m:
* 19 new contributes
* more than 85~ new commits 
* 11 new files

This is a good reason to update it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,213 +8,232 @@
 
 ```
  project  : mocha
- repo age : 3 years, 4 months
- active   : 509 days
- commits  : 1575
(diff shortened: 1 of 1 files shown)


danielstjules:
Looks good :)

One thing though, I wonder if it's worth keeping this list? With this update, it hits 221 contributors. Moving/removing them would make the resources at the bottom of the page much more visible (which could also be updated).

travisjeffery:
ya i'd rather get rid of it

a8m:
Fair enough, thx.

PR #1591, opened 2015-03-10
feat(runner): fix '.only()' exclusive feature, #1481

### LINKED ISSUE #1481: describe.only matches more that one describe test suite

a8m:
This PR fix #1481, and also extends the `.only()` behaviour.
e.g: 
* it's not use `grep` anymore
* support suite, test-case or both.
* add the ability to run multiple `.only()` (on many files)

There's a tests for [bdd](https://github.com/a8m/mocha/blob/master/test/acceptance/misc/only/bdd.js), [tdd](https://github.com/a8m/mocha/blob/master/test/acceptance/misc/only/tdd.js) and [qunit](https://github.com/a8m/mocha/blob/master/test/acceptance/misc/only/qunit.js) interfaces that illustrate the point more clearly. so please take a look on it, and let me now what do you think.

### DIFF
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -69,9 +69,10 @@ module.exports = function(suite){
      * Exclusive suite.
      */
 
-    context.describe.only = function(title, fn){
+    context.describe.only = function(title, fn) {
       var suite = context.describe(title, fn);
-      mocha.grep(suite.fullTitle());
+      mocha.options.hasOnly = true;
+      suite.isOnly = true;
       return suite;
     };
 
@@ -95,9 +96,11 @@ module.exports = function(suite){
      */
 
     context.it.only = function(title, fn){
-      var test = context.it(title, fn);
-      var reString = '^' + escapeRe(test.fullTitle()) + '$';
-      mocha.grep(new RegExp(reString));
+      var test = context.it(title, fn)
+        , suite = test.parent;
+      mocha.options.hasOnly = true;
+      suite.isOnly = true;
+      suite.onlyTests = (suite.onlyTests || []).concat(test);
       return test;
     };
 
--- a/lib/interfaces/tdd.js
+++ b/lib/interfaces/tdd.js
@@ -76,7 +76,9 @@ module.exports = function(suite){
 
     context.suite.only = function(title, fn){
       var suite = context.suite(title, fn);
-      mocha.grep(suite.fullTitle());
+      mocha.options.hasOnly = true;
+      suite.isOnly = true;
+      return suite;
     };
 
     /**
@@ -99,9 +101,12 @@ module.exports = function(suite){
      */
 
     context.test.only = function(title, fn){
-      var test = context.test(title, fn);
-      var reString = '^' + escapeRe(test.fullTitle()) + '$';
-      mocha.grep(new RegExp(reString));
+      var test = context.test(title, fn)
+        , suite = test.parent;
+      mocha.options.hasOnly = true;
+      suite.isOnly = true;
+      suite.onlyTests = (suite.onlyTests || []).concat(test);
+      return test;
     };
 
     context.test.skip = common.test.skip;
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -514,6 +514,7 @@ Runner.prototype.runSuite = function(suite, fn){
     , self = this
     , i = 0;
 
+
   debug('run suite %s', suite.fullTitle());
 
   if (!total) return fn();
@@ -606,6 +607,9 @@ Runner.prototype.run = function(fn){
   var self = this,
     rootSuite = this.suite;
 
+  // If there is an `only` filter
+  if (this.hasOnly) filterOnly(rootSuite);
+
   fn = fn || function(){};
 
   function uncaught(err){
@@ -655,6 +659,24 @@ Runner.prototype.abort = function(){
   debug('aborting');
   this._abort = true;
 };
+
+/**
+ * Filter suites based on `isOnly` logic.
+ *
+ * @param {Array} suite
+ * @returns {Boolean}
+ * @api private
+ */
+function filterOnly(suite) {
+  // If it has `only` tests, run only those
+  if (suite.onlyTests) suite.tests = suite.onlyTests;
+  // Filter the nested suites
+  suite.suites = filter(suite.suites, filterOnly);
+  // Don't run tests from suites that are not marked as `only`
+  suite.tests = suite.isOnly ? suite.tests : [];
+  // Keep the suite only if there is something to run
+  return suite.suites.length || suite.tests.length;
+}
 
 /**
  * Filter leaks with the given globals flagged as `ok`.
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ lib-cov:
 
 test: test-unit
 
-test-all: test-bdd test-tdd test-qunit test-exports test-unit test-grep test-jsapi test-compilers test-sort test-glob test-requires test-reporters test-only test-failing test-regression
+test-all: test-bdd test-tdd test-qunit test-exports test-unit test-grep test-jsapi test-compilers test-sort test-glob test-requires test-reporters test-global-only test-only test-failing test-regression
 
 test-jsapi:
 	@node test/jsapi
@@ -163,6 +163,22 @@ test-only:
 		--reporter $(REPORTER) \
 		--ui qunit \
 		test/acceptance/misc/only/qunit
+
+test-global-only:
+	@./bin/mocha \
+		--reporter $(REPORTER) \
+		--ui tdd \
+		test/acceptance/misc/global-only/tdd
+
+	@./bin/mocha \
+		--reporter $(REPORTER) \
+		--ui bdd \
+		test/acceptance/misc/global-only/bdd
+
+	@./bin/mocha \
+		--reporter $(REPORTER) \
+		--ui qunit \
+		test/acceptance/misc/global-only/qunit
 
 test-sort:
 	@./bin/mocha \
--- a/lib/interfaces/qunit.js
+++ b/lib/interfaces/qunit.js
@@ -62,7 +62,9 @@ module.exports = function(suite){
 
     context.suite.only = function(title, fn){
       var suite = context.suite(title, fn);
-      mocha.grep(suite.fullTitle());
+      mocha.options.hasOnly = true;
+      suite.isOnly = true;
+      return suite;
     };
 
     /**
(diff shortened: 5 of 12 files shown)


boneskull:
@a8m Thanks!  Hopefully someone will check this out soon--I'm going to be swamped for next few wks still.  I'm thinking might be good for `3.0.0` release

a8m:
Great @boneskull 
/cc @travisjeffery @dasilvacontin

erezesk:
@a8m looks great, tyvm!
+1

dasilvacontin:
Very busy lately, I'll try to dedicate some time to mocha this weekend. Anyways, thanks for the amazing work you've been pushing, @a8m! :)

travisjeffery:
looks pretty good, i've looked it over quickly a couple of times and got the feeling some parts could be simpler/clearer. i should get sometime tomorrow to look more in-depth

a8m:
Great @travisjeffery 
thx

danielstjules:
Looks pretty good! Works with almost all my tests, though I think I found a couple edge cases. The first is with the root suite I think:

``` javascript
describe.only('suite', function() {
  it.only('test', function() {});

  it('test2', function() {});
});

it('should not be ran', function() {});

it.only('another test', function() {});
```

```
danielstjules:~/git/mocha (pr1591)
$ ./bin/mocha --reporter spec example.js


  ✓ should not be ran
  ✓ another test
  suite
    ✓ test


  3 passing (10ms)
```

On master, however, I get the following:

```
$ ./bin/mocha --reporter spec example.js


  ✓ another test

  1 passing (6ms)
```

I think it was decided in the issue that spec-related `only` just take precedence over suite level, so with that in mind, should the correct output of this suite be:

```
  ✓ another test
  suite
    ✓ test


  ` passing (11ms)
```

danielstjules:
The second issue I hit was with the following test:

``` javascript
describe.only('suite', function() {
  it.only('test', function() {});

  it('test2', function() {});

  describe('suite2', function() {
    it('test3', function() {});
  });
});
```

```
danielstjules:~/git/mocha (pr1591)
$ ./bin/mocha --reporter spec example.js


  0 passing (3ms)
```

a8m:
haha, I forgot to test the root suites, it should be fix [here](https://github.com/a8m/mocha/blob/master/lib/runner.js#L611). I'll do it.
about the second example,  I'll play with it and see what happens
thx @danielstjules

a8m:
Fixed! 
PTAL @danielstjules

danielstjules:
Those 2 look good, awesome!

Got another example, though I don't know if this is expected behavior or not. So this is probably not a bug, but just wanted to clarify :)

``` javascript
describe.only('suite', function() {
    it.only('test', function() {});

    it('should not run', function() {});
  });

  it('should not run either', function() {});
});
```

```
dstjules:~/git/mocha (pr1591)
$ ./bin/mocha --reporter spec example.js


  suite
    ✓ should not run either
    inner
      ✓ test


  2 passing (8ms)
```

Is that expected output? If so, then the rest looks good to me!

a8m:
This is the expected behaviour, PTAL [bdd-test](https://github.com/a8m/mocha/blob/master/test/acceptance/misc/only/bdd.js#L79)
Thanks for a great CR @danielstjules

a8m:
This is the expected behaviour, PTAL [bdd-test](https://github.com/a8m/mocha/blob/master/test/acceptance/misc/only/bdd.js#L79)
Thanks for a agreat CR @danielstjules

a8m:
rebased!

mocha.js
```diff
@@ -5230,6 +5247,30 @@ Runner.prototype.abort = function(){
...
+ * @param {Array} suite
+ * @returns {Array}
+ * @api private
+ */
+function filterOnly(suite) {
+  // If it has `only` tests
+  if(suite.hasOnly) suite.tests = suite.hasOnly;
+  // If it doesn't has a suites make it running
+  // only if it mark as an `only`
+  if(!suite.suites.length) return suite.isOnly;
```
dasilvacontin: `suite.isOnly` is a boolean right? The function description say it returns `Array`.

mocha.js
```diff
@@ -5230,6 +5247,30 @@ Runner.prototype.abort = function(){
...
+ * Filter suites based on `isOnly` logic.
+ *
+ * @param {Array} suite
+ * @returns {Array}
+ * @api private
+ */
+function filterOnly(suite) {
+  // If it has `only` tests
+  if(suite.hasOnly) suite.tests = suite.hasOnly;
+  // If it doesn't has a suites make it running
```
dasilvacontin: This comment has a couple of typos.

lib/runner.js
```diff
@@ -657,6 +661,30 @@ Runner.prototype.abort = function(){
...
+ * Filter suites based on `isOnly` logic.
+ *
+ * @param {Array} suite
+ * @returns {Array}
+ * @api private
+ */
+function filterOnly(suite) {
+  // If it has `only` tests
+  if(suite.hasOnly) suite.tests = suite.hasOnly;
+  // If it doesn't has a suites make it running
```
dasilvacontin: This comment has a few typos.

lib/runner.js
```diff
@@ -657,6 +661,30 @@ Runner.prototype.abort = function(){
...
+  // only if it mark as an `only`
+  if(!suite.suites.length) return suite.isOnly;
```
dasilvacontin: `suite.isOnly` is `Boolean` type, right? The function description says it returns `Array`.

It makes me wonder if it works correctly when this return is used, due to the following line:
```
if(this.hasOnly) rootSuite = filterOnly(rootSuite);
````

File: lib/runner.js [hunk 5]
```diff
@@ -657,6 +661,30 @@ Runner.prototype.abort = function(){
...
+  // Filter the nested suites
+  suite.suites = filter(suite.suites, filterOnly);
+  // `Unrunning` suites that not mark as an `only`
```
dasilvacontin: Maybe:
```javascript
//  Don't run suites that are not marked with `only`
```
Though if we are commenting the following line, maybe it would me more adequate:
```javascript
// Don't run tests from suites that are not marked as `only`
```

lib/runner.js
a8m: It's used also as an iterator(recursively).

But, you right, it's a bug. I'll fix it quickly.
thx

lib/interfaces/tdd.js
```diff
@@ -99,9 +101,12 @@ module.exports = function(suite){
...
 
     context.test.only = function(title, fn){
-      var test = context.test(title, fn);
-      var reString = '^' + escapeRe(test.fullTitle()) + '$';
-      mocha.grep(new RegExp(reString));
+      var test = context.test(title, fn)
+        , suite = test.parent;
+      mocha.options.hasOnly = true;
+      suite.isOnly = true;
+      suite.hasOnly = (suite.hasOnly || []).concat(test);
```
dasilvacontin: Isn't a bit confusing that you are using `hasOnly` as a `Boolean` in both `mocha.options` and `Runner`, and then you use it as an `Array` in `Suite`?

Besides that, the variable name `hasOnly` seems appropriate for a `Boolean`, but not so much for an `Array`.

e.g. `suite.tests` is the `Array` with the tests, `suite.onlyTests` could be the `Array` with the tests that 'are only'. It describes much better the content of the property.

lib/runner.js
dasilvacontin: Okay, as expected, it does not work with the following test file (or is this not a valid test file?):
```javascript
it('should not be executed', function() { })
it.only('is the only test that should be executed', function() { })
it('shouldn\'t be executed either', function() { })
```

a8m: it's valid.
I'm on it.

lib/runner.js
```diff
@@ -606,6 +607,9 @@ Runner.prototype.run = function(fn){
   var self = this,
     rootSuite = this.suite;
 
+  // If there is an `only` filter
+  if(this.hasOnly) rootSuite = filterOnly(rootSuite);
```
dasilvacontin: Okay, I see, you are also using`filterOnly` as a filter function. You only return `Boolean` or `suite`.
Why not change the line to:
```javascript
if(this.hasOnly) filterOnly(rootSuite);
```
You were always intending to return `rootSuite` here, right?

lib/runner.js
```diff
@@ -657,6 +661,30 @@ Runner.prototype.abort = function(){
...
 
 /**
+ * Filter suites based on `isOnly` logic.
+ *
+ * @param {Array} suite
+ * @returns {Array}
+ * @api private
+ */
+function filterOnly(suite) {
+  // If it has `only` tests
```
dasilvacontin: ```// If it has `only` tests, we should only run those```

lib/runner.js
dasilvacontin: Cool. I should reload more often, I wasn't seeing your comments. :)

dasilvacontin:
@a8m, in general, you are writing `if(val)` instead of `if (val)`.

lib/runner.js
```diff
@@ -657,6 +661,28 @@ Runner.prototype.abort = function(){
 };
 
 /**
+ * Filter suites based on `isOnly` logic.
+ *
+ * @param {Array} suite
+ * @returns {*}
```
dasilvacontin: Maybe we can set this now to `Boolean`?

dasilvacontin:
LGTM :thumbsup:.

@travisjeffery, were you going to suggest any code improvements?

a8m:
ping @mochajs/mocha

dasilvacontin:
Well, goes.

dasilvacontin:
We'll have to wait for the v3.0.0 release tho.

boneskull:
@a8m I haven't gotten a chance to look at this yet, but please let me before this gets merged.

cc @mochajs/mocha

lib/interfaces/bdd.js
```diff
@@ -95,9 +96,11 @@ module.exports = function(suite){
      */
 
     context.it.only = function(title, fn){
-      var test = context.it(title, fn);
-      var reString = '^' + escapeRe(test.fullTitle()) + '$';
-      mocha.grep(new RegExp(reString));
+      var test = context.it(title, fn)
```
boneskull: various places in this PR will need to be linted

a8m:
Replaced with #1807(rebased on top of v3.0.0)

Vanuan:
WTF? Are you trying to fix all bugs with one PR?

boneskull:
@Vanuan I understand you are probably frustrated with the `.only()` bug(s), but please try to be constructive.  Thanks!

PR #1595, opened 2015-03-13
Stricter .only for suites

peterflynn:
Suites whose names are a superset of the .only suite shouldn't wind up also running.

Similar to PR #941, an earlier fix for .only on tests.

Unlike 941, this doesn't _fully_ fix the bug for suite names -- because test names are concatenated to suite names, the grep can't be as strict. When the name of the .only suite + `" "` is a prefix of another suite's name, that other suite will still run even with this fix.  But it's still an improvement, and fixes the specific unexpected behavior I was running into (when suites are named after the classes they test, and base classes / interfaces are often a prefix of the names of other classes).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ lib-cov:
 
 test: test-unit
 
-test-all: test-bdd test-tdd test-qunit test-exports test-unit test-grep test-jsapi test-compilers test-sort test-glob test-requires test-reporters test-only test-failing test-regression
+test-all: test-bdd test-tdd test-qunit test-exports test-unit test-grep test-jsapi test-compilers test-sort test-glob test-requires test-reporters test-only test-suite-only test-failing test-regression
 
 test-jsapi:
 	@node test/jsapi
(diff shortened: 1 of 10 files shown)


peterflynn:
Ah, unfortunately just saw PR #1591 after submitting this -- it fixes the same issue more thoroughly via a more complex patch.  If #1591 gets merged I'll close this one.

travisjeffery:
Yeah probably gonna go with #1481, thanks though!

davidham:
Sorry if this was addressed elsewhere, but I was just able to reproduce this with two describe blocks in the same file. I have two describe blocks, where the name of one is a superset of the other, and 'only' on the first, but they both execute. Like so:

```
describe.only('Foo', function() {
            it('should be true', function() {
                expect(true).to.be.true;
            });
});

describe('Foo2', function() {
            it('should be true', function() {
                expect(true).to.be.true;
            });
});
```

I expected only the first to run. 

Is this the desired behavior? If so, could a note about this be added to the documentation? Sorry if I am duplicating a known issue but I couldn't tell from the above referenced issues. Thanks!

PR #1601, opened 2015-03-15
Fix #1582, adding global object save/restore capabilities

### LINKED ISSUE #1582: get the default `process.stdout.write` back after tests are finished
I have a test which tests stdout so I need to change the default behaviour of `process.stdout`. 

```js
process.stdout.write = (function(write) {
    return function(string, encoding, fd) {
      var args = _.toArray(arguments);
      callback.call(callback, string);
    };
  }(process.stdout.write));
```

mocha's report, however, cannot be seen any more if I use this. It would be good if mocha remembers the default `process.stdout.write` at the beginning and change it back after tests are finished.

JohnnyEstilles:
This is a fix for issue #1582, which reports that functions in the `global` object that were poisoned during a test (e.g. stubbed or otherwise altered) were not restored upon test completion.

The PR creates a module that provides for saving/caching the `global` object (during the first call to `GlobalObject.save()`), as well as restoring it from cache when needed (by calling `GlobalObject.restore()`).

`Runnable` now includes a call to `GlobalObject.save()` just before kicking off a test, and a call to `GlobalObject.restore()` on the callback invoked upon test completion (whether successful or not).

#### Example
In the example below, every test stubs the `console.log()` method, rendering it useless, unless it is explicitly restored.
```js
$ cat example.js
  before(function() {
    console.log('before: should work');
  });
  it('example 1', function() {
    console.log('example1: should work');
    console.log = function() {};
    console.log('example1: should not work');
  });
  it('example 2', function() {
    console.log('example2: should work');
    console.log = function() {};
    console.log('example2: should not work');
  });
  after(function() {
    console.log('after: should work');
  });
});
```

#### Before Fix
When `console.log()` is stubbed in test `example 1`, it remains so, even after the test is complete. As a result, output from successive tests, the `after()` hook, and the reporter are all suppressed.
```bash
$ ./bin/mocha -R spec example.js

  suite
before: should work
example1: should work
```

#### After Fix
I corrected this by restoring the global object, upon test completion. In my example, now `console.log()` functionality is restored on every successive test, as well as the `after()` hook. Reporter output is, likewise, no longer suppresed.
```bash
$ ./bin/mocha -R spec example.js

  suite
before: should work
example1: should work
    ✓ example 1 
example2: should work
    ✓ example 2 
after: should work


  2 passing (35ms)
```

### DIFF
--- a/lib/global-object.js
+++ b/lib/global-object.js
@@ -1,0 +1,94 @@
+'use strict';
+
+/**
+ * Module dependencies.
+ */
+
+var utils = require('./utils');
+
+/**
+ * Expose `GlobalObject`.
+ */
+
+module.exports = GlobalObject;
+
+/**
+ * Initialize `GlobalObject`.
+ *
+ */
+function GlobalObject() {
+  if (!(this instanceof GlobalObject)) {
+    return new GlobalObject();
+  }
+  this._cache = null;
+}
+
+/**
+ * Cache all functions in the `global object.
+ *
+ * @api public
+ */
+
+GlobalObject.prototype.save = function(){
+  if (!this._cache) {
+    this._cache = saveGlobalObject(global);
+  }
+  return this._cache;
+};
+
+/**
+ * Restore cached functions back to the global object.
+ *
+ * @api public
+ */
+
+GlobalObject.prototype.restore = function(){
+  if (this._cache) {
+    restoreGlobalObject(this._cache, global);
+  }
+}
+
+function saveGlobalObject(object, checklist) {
+  var checklist = checklist ? checklist : [];
+  var val = {};
+
+  if (utils.type(object) === 'global' && arguments.length > 1) {
+    return;
+  }
+
+  for(var i=0; i<checklist.length; i++) {
+    if (checklist === checklist[i]) {
+      return;
+    }
+  }
+  checklist.push(object);
+
+  for(var prop in object) {
+    if(utils.type(object[prop]) === 'object' || utils.type(object[prop]) === 'process') {
+      val[prop] = saveGlobalObject(object[prop], checklist);
+    }
+
+    if(utils.type(object[prop]) === 'function') {
+      val[prop] = object[prop];
+    }
+  }
+
+  return val;
+}
+
+
+function restoreGlobalObject(source, target) {
+  for(var prop in source) {
+    if(utils.type(source[prop]) === 'object' || utils.type(source[prop]) === 'process') {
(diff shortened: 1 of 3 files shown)


a8m:
I think this is part of your program design and not part of Mocha.
IMO, by testing your "unit" you may want to test its own behaviour, and log to stdout/console it's not part of it (i.e: we might want to test if it's called, but not call it directly, because this not the responsibility of our "unit").
You can simply win it by loose coupling and using DI for example.
e.g: your code may look like this:
```js
function MyService(logger) {
  // getting logger service, used by this service
  // it could be `console`, `winston` or stub for testing
  this.logger = logger;
}

MyService.prototype.myVerySmartFn = function(val) {
  // ...
  // logger used by this service
  this.logger(val);
  // ...
}
```
and your test may be something like that: 
```js
describe('MyService', function() {
  var ms = new MyService(stubLogger);
  // ...
  describe('Call `myVerySmartFn` with invalid argument', function() {
    it('should log error', function() {
      ms.myVerySmartFn(null);
      stubLogger.error.called.should.be.true;
      // ...
    });
  });
  // ...
});
```
This is how I see it, and I'd love to hear your opinion @JohnnyEstilles

travisjeffery:
+1 @a8m, i'm not too keen on this change

JohnnyEstilles:
@a8m, the `console.log()` statements in my example are merely for illustration purposes only. It's simple to demonstrate that if a test stubs/changes the behavior of methods in the global object and doesn't clean up after itself, it may affect the rest of the tests, or like in this case, reporter output. Unfortunately, performing this type of cleanup in an `after()` or `afterEach()` hook doesn't really work, because other Mocha code (including the reporter), may run before them.

I understand if there's no interest. I was just responding to the apparent interest expressed in #1582.

Closing ...

stevemao:
My test is not to testing `console.log` but `stdout`. I have to change the behaviour of `process.stdout.write` to achieve this. 

Each test should be sandboxed in mocha and users should be able to do whatever dodge things inside each test. Users can change `process.stdout.write` but they should still be able to see the report (although the report writes to `process.stdout`, this shouldn't concern users).

JohnnyEstilles:
@stevemao your argument is perfectly valid. I was just using `console.log()` to illustrate the point. Maybe I should have used `process.stdout.write()` instead. Either way, I was just trying to should how "not" restoring the state of the global object after a test can affect Mocha's execution (like in this case, suppressing `Reporter` output).

PR #1602, opened 2015-03-15
fix issue where menu overlaps content.

smussell:
There are more improvements that could be made, for example it'd be nice to be able to collapse the menu all together, but at least the fixes the current issue I'm having which is, if my file names are long, the menu covers up the content and I am not able to scroll so my content is completely hidden.

### DIFF
--- a/lib/reporters/templates/style.html
+++ b/lib/reporters/templates/style.html
@@ -8,7 +8,7 @@ body {
 }
 
 #coverage {
-  padding: 60px;
+  padding: 60px 400px 60px 60px;
 }
 
 h1 a {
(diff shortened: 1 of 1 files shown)


a8m:
Could you provide please some "before and after" screenshots? so I can test it more properly.
thx

smussell:
## Before

You can see the menu is obstructing the page content.

screen shot 2015-03-15 at 4 04 37 pm

## After

Now the menu allows scrolling, but has max-width so it isn't overlapping the content.

screen shot 2015-03-15 at 4 07 16 pm


There are certainly some more improvements that could be made, but I don't have much time to spend on it now, and this fixes the issue for me.

danielstjules:
@smussell Looks good! Mind bumping that to 400px though? I think this is a good fix until we start extracting common paths from each file path.

smussell:
Sure, I made the change to bump the size to 400, thanks.

a8m:
thanks @smussell. please squash and we'll merge it.

danielstjules:
@a8m @smussell Just doing some quick cross-browser tests (older IEs) and will merge if all looks good :)

a8m:
go for it @danielstjules :)

danielstjules:
Found some quirks that are def unrelated to this PR - so looks good!

PR #1603, opened 2015-03-16
support --es_staging flag for iojs

klaemo:
As per https://iojs.org/en/es6.html this is a (prefered) synonym for --harmony.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -79,6 +79,7 @@ program
   .option('--debug-brk', "enable node's debugger breaking on the first line")
   .option('--globals <names>', 'allow the given comma-delimited global [names]', list, [])
   .option('--harmony', 'enable all harmony features (except typeof)')
+  .option('--es_staging', 'enable all staged features')
   .option('--harmony-collections', 'enable harmony collections (sets, maps, and weak maps)')
   .option('--harmony-generators', 'enable harmony generators')
   .option('--harmony-proxies', 'enable harmony proxies')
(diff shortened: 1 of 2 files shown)


a8m:
LGTM
thx @klaemo

danielstjules:
@klaemo Looks good! Btw, can you point to where it suggests this is a preferred alternative to `--harmony`? I skimmed https://iojs.org/en/es6.html but didn't notice anything. Just curious, and want to understand why that's the case.

Furthermore, can you update the string as it currently indicates that it's only available in iojs, while it's also available in node 0.12: https://github.com/joyent/node/blob/885f72105bf0b18007100550f131c85ca56536e0/deps/v8/src/flag-definitions.h#L151

'enable all staged features' is probably good :)
Thanks!

a8m:
> 'enable all staged features' is probably good :)

I agree with this.

klaemo:
@danielstjules I just assumed by the wording that `--es_staging` is prefered. The headline is "No more --harmony flag" after all. Also: 
> "These are the "use at your own risk" type of features that now require a runtime flag: --es_staging (or its synonym, --harmony)."

Seems like `--harmony` is there for backwards compat.

I updated my commit according to your comment. I didn't know node 0.12 supported this flag, too. :)

a8m:
Thanks @klaemo

danielstjules:
@klaemo I think the "No more --harmony flag" is in reference to
> With io.js@1.x (V8 4.1+), all that complexity goes away. All harmony features are now logically split into three groups for shipping, staged and in progress features: All shipping features, the ones that V8 has considered stable, like generators, templates, new string methods and many others are turned on by default on io.js and do NOT require any kind of runtime flag.

Meaning a majority of harmony features are accessible without it. So no more --harmony/--es_staging required :) But I see your point as well - def up for interpretation.

PR #1610, opened 2015-03-19
Mocha should now exit active domains after the completion of a test

owenallenaz:
Due to the way the callbacks chain in Node, if a domain was entered within a test, no matter how deep, then that domain would propagate to other tests later in the suite and cause lots of confusing conditions.

In order to correct this I simply check for an active domain and exit. Pull request includes a test for this condition.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -3,7 +3,7 @@
  */
 
 var EventEmitter = require('events').EventEmitter
-  , domain = require("domain")
+  , domain = require('domain')
   , debug = require('debug')('mocha:runnable')
   , Pending = require('./pending')
   , milliseconds = require('./ms')
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -209,7 +209,7 @@ Runnable.prototype.run = function(fn){
 
   // finished
   function done(err) {
-    // if (domain.active) { domain.active.exit(); }
+    if (domain.active) { domain.active.exit(); }
     
     var ms = self.timeout();
     if (self.timedOut) return;
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -3,6 +3,7 @@
  */
 
 var EventEmitter = require('events').EventEmitter
+  , domain = require("domain")
   , debug = require('debug')('mocha:runnable')
   , Pending = require('./pending')
   , milliseconds = require('./ms')
@@ -208,6 +209,8 @@ Runnable.prototype.run = function(fn){
 
   // finished
   function done(err) {
+    // if (domain.active) { domain.active.exit(); }
+    
     var ms = self.timeout();
     if (self.timedOut) return;
     if (finished) return multiple(err || self._trace);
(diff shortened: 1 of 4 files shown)


lib/runnable.js
```diff
@@ -208,6 +209,8 @@ Runnable.prototype.run = function(fn){
 
   // finished
   function done(err) {
+    if (domain.active) { domain.active.exit(); }
```
dasilvacontin: Aren't those braces unnecessary?

owenallenaz: Entirely, it's my personal pref, but I can remove and re-submit if that'll match the style of the repo.

boneskull:
@owenallenaz We need domain support.  Unfortunately, we can't just throw `require('domain')` in `runner.js`, because there's no such thing in the browser. 

We have no browser-based tests (we should), so the TravisCI build passes.

Perhaps best thing to do at the moment would be to wrap the `require('domain')` in a try/catch?

boneskull:
also, please rebase your changes on top of branch `v2.3.0`

owenallenaz:
I was worried about the browser situation. I'm not super familiar with the Mocha codebase so please correct me if I'm wrong, but from looking it I see there is a `compile.js` that is basically responsible for shimming the Node specific require statements to browser require statements.

Since that's the pathway being used for other Node-only things like `EventEmitter` then it seems we'd do something similar for `domain`. With that, my plan would be add a `browser/domain.js` which would basically consist of return { active : false }. Since that is all the shimming we need for this specific use-case. And if we needed to make that more robust we'd be updating our domain shim in one place in the future.

It should only take a few to rebase and alter it that way, so you're OK with that approach let me know and I'll take care of it.

boneskull:
@owenallenaz sounds good.  that stuff needs to be replaced w/ browserify anyway

owenallenaz:
I've rebased this branch and added the shim for the domain in the browser. I wasn't exactly sure where to go to run the official browser tests, but I was able to run `make` and the file that it spat back loaded properly in a browser environment.

owenallenaz:
While testing some code using domains in an unrelated project I've come to learn more about the domain mechanism. This pull request isn't fully complete as if we exit we need to be `process.nextTick` or `setImmediate` to bounce off the event loop or the domain isn't "fully" exited per http://stackoverflow.com/questions/27192917/using-a-domain-to-test-for-an-error-thrown-deep-in-the-call-stack-in-node . I'll update the pull request with that change.

owenallenaz:
Closing pull request. From testing and evaluation it's not the kind of thing that mocha should be doing. The fact is if a user is using domains in their logic then they need to be properly entering and exiting the domains within their modules. Mocha as it stands properly tests domain related code the only time it goes south is when there are mistakes made within the modules themselves (such as not exiting the domain after a caught error).

For that reason I see can see value in it being an error state if a domain is left open after a test, similar to ensuring that done wasn't called twice. Such a thing is different than this request and should be handled accordingly.

boneskull:
@owenallenaz This is good to know, thanks.

Issue #1585, opened 2015-03-08
Coverage script failing


am11:
Redirected from: https://github.com/travis-ci/travis-ci/issues/3347

Hi, we have a [coverage script](https://github.com/sass/node-sass/blob/beffe92723c2a1848eef248b27c79dac2bcf6446/scripts/coverage.js) in node-sass for TravisCI. It was last updated on Nov 4 2014. It was reporting coverage [to covealls](https://coveralls.io/r/sass/node-sass) till yesterday when I bumped mocha version https://github.com/sass/node-sass/commit/c9ed9300640cf746fb99204c10fc3aa8046bee6b. Now we get permission denied on TravisCI: https://travis-ci.org/sass/node-sass/jobs/53495198#L1320 on running the coverage script (after passing all tests with mocha).

>\> node-sass@2.0.1 coverage /home/travis/build/sass/node-sass
 node scripts/coverage.js
>
/home/travis/build/sass/node-sass/node_modules/.bin/_mocha: 1: /home/travis/build/sass/node-sass/node_modules/.bin/_mocha: /bin: Permission denied

If I downgrade mocha to 2.1.0, the result is: https://travis-ci.org/am11/node-sass/jobs/53533514#L1310:

> $ npm run-script coverage
> node-sass@2.0.1 coverage /home/travis/build/am11/node-sass
> node scripts/coverage.js
>
Done. Your build exited with 0

Is it related to https://github.com/mochajs/mocha/pull/1567?

danielstjules:
I think you might be right! Because the permissions are certainly correct:
```
$ ls -al . ../mocha/bin
.:
total 16
drwxr-xr-x  4 danielstjules  staff  136 Mar  8 21:14 .
drwxr-xr-x  4 danielstjules  staff  136 Mar  8 21:14 ..
lrwxr-xr-x  1 danielstjules  staff   19 Mar  8 21:14 _mocha -> ../mocha/bin/_mocha
lrwxr-xr-x  1 danielstjules  staff   18 Mar  8 21:14 mocha -> ../mocha/bin/mocha

../mocha/bin:
total 32
drwxr-xr-x   4 danielstjules  staff    136 Mar  8 21:14 .
drwxr-xr-x  12 danielstjules  staff    408 Mar  8 21:14 ..
-rwxr-xr-x   1 danielstjules  staff  11947 Mar  5 22:56 _mocha
-rwxr-xr-x   1 danielstjules  staff   1722 Feb 18 11:41 mocha
[...]

PR #1616, opened 2015-03-22
Fix 1585: make _mocha executable again

danielstjules:
This is for https://github.com/mochajs/mocha/issues/1585

After merging, I think it'd be a good idea to push a new patch release, since both 2.2.0/2.2.1 broke backwards compatibility for those relying on `_mocha` being executable.

To show that it's compatible with https://github.com/mochajs/mocha/commit/1430c2b5102e9d355119a643e9b2be2e76e91d1f

``` javascript
$ cat example.js
it('test', function() {
  var gen = function*() {};
});

$ node --version
v0.12.0

$ ./bin/mocha --harmony-generators example.js

  ․

  1 passing (4ms)

$ nvm use 0.10
Now using node v0.10.20

# Env has node 0.10, but using node 0.12
$ /Users/danielstjules/.nvm/versions/node/v0.12.0/bin/node ./bin/mocha --harmony-generators example.js

  ․

  1 passing (5ms)

# Env has node 0.10, using node 0.10
$ ./bin/mocha --harmony-generators example.js
Error: unrecognized flag --harmony-generators
Try --help for options

/Users/danielstjules/git/mocha/example.js:2
  var gen = function*() {};
```

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 /**
  * Module dependencies.
  */


a8m:
LGTM.

parshap:
Could this change get published to npm?

Issue #1604, opened 2015-03-16
bug with dynamic skip functionality


ghughal:
I'm getting error with using dynamic skip() functionality with asyn function. Is this a bug or I'm doing something wrong?

``` javascript
    it('should skip after callback', function(done){
        var self = this;
        var asyncFunc = function(cb) {
            setTimeout(cb, 1000);
        };

        asyncFunc(function() {
            self.skip();
            throw new Error('never thrown');
        });
    })
```

Here's error I'm getting:

```
  1) skip in test should skip after callback:
     Error: the object {
  "message": [undefined]
  "uncaught": true
} was thrown, throw an Error :)
      at Runner.fail (/usr/local/lib/node_modules/mocha/lib/runner.js:207:11)
      at Runner.uncaught (/usr/local/lib/node_modules/mocha/lib/runner.js:580:8)
      at process.uncaught (/usr/local/lib/node_modules/mocha/lib/runner.js:609:10)
      at process.EventEmitter.emit (events.js:95:17)
      at process._fatalException (node.js:272:26)
```

a8m:
It's seems like a bug.
I started to play with it, and I'll update later.
Thx @ghughal

ghughal:
Thanks @a8m

danielstjules:
@ghughal @a8m I don't think this behavior's ever been supported, but I submitted a possible solution for allowing it: https://github.com/mochajs/mocha/pull/1618

ghughal:
Thank you @danielstjules

ScottFreeCode:
This should be fixed by #2335; can anyone confirm?

maggiesavovska:
I'm getting the following error: 
"Converting circular structure to JSON"

`it('should skip async test', function(done){
            var self = this;
            return driver.url('https://www.google.com').then(function(result){
                self.skip();

            }).catch(function(err){console.log('error', err.message)});

        });`

ScottFreeCode:
I just checked with the various versions of Mocha and confirmed that this issue from 2.x was fixed in 3.0. I also checked the PR and it added regression tests for this specific case, so this shouldn't come up again either. 🎆 

[...]

Issue #1480, opened 2014-12-22
"Warn" ability, instead of fail


aseemk:
We at @FiftyThree :heart: Mocha. We use it to test [Mix](https://mix.fiftythree.com/) and more.

As our test codebase has grown, there are a few test suites we'd like to run like usual, but not fail overall if they fail. We just want to be warned, sort of like pending tests.

I emailed the group about this a year ago, so starting a formal issue for this now:

https://groups.google.com/d/topic/mochajs/zpCTMFD5EQ8/discussion

Strawman proposal: `describe.warn('foo')` / `it.warn('should bar')`. Or, if you like cuteness, `it.ideally('should bar')`. =)

WDYT? Thanks for the consideration!

kpheng:
@travisjeffery 

Retry support would be awesome

dasilvacontin:
Just throwing out some ideas:
```
try(3).it('should ...', fn);

retry(2).it('should ...', fn);

describe('Suite', function () {

  this.retry(3);

  it('should ...', fn);

  it('should ...', function (done) {
    this.retry(2);
    ...
  })

  it('should ...', function (done) {
    this.timeout(700).retry(5);
    ...
  })

})
```

danielstjules:
So what sort of status code should the process exit with if a spec failed, but was only a "warning"? How would you notice it? If it's a non-zero code, it's going to cause the build in your CI system to fail. If it just exits with zero, then you're expected to look at the test results and notice the warning?

If this is in the context of CI, then one solution is to split up your test suite. E.g. test/unit, test/functional, test/allowed-failures. You would run the first two directories normally. But for the third, you'd break it into a separate command: `mocha test/allowed-failures || true`. The suite will run as normally, but will not fail your build on error.

boneskull:
For reference, here's the text of the forum post:

> Longtime user and fan of Mocha. Really nice work and thanks as always for all the amazing contributions to the Node world.

> I'm running into a tricky timing bug that causes a few of my tests (just 1-2%) to fail randomly. This breaks my CI, which is cumbersome since there are many tests and we haven't optimized CI and test performance.

> So for now, I'd love to be able to mark those test cases as "okay to fail". I know I can skip them, but I'd like to still run them, so I stay aware of whether (and how often) they're still failing. But, I wouldn't like Mocha to return a non-zero exit code if only those tests fail. Sort of like "warnings" rather than "failures".

> Is that possible? If not, is that a feature you'd be open to adding?

[...]

PR #1618, opened 2015-03-23
Fix 1604: Add support for async skip()

danielstjules:
This is for https://github.com/mochajs/mocha/issues/1604

``` javascript
it('should skip', function(done) {
  var test = this;

  setTimeout(function() {
    test.skip();
  }, 0);
});

it('should run', function() {
  // do nothing
});
```

```
dstjules:~/git/mocha (master=)
$ ./bin/mocha example.js

  ․․

  1 passing (14ms)
  1 failing

  1)  should skip:
     Error: the object {
  "message": [undefined]
  "uncaught": true
} was thrown, throw an Error :)
      at Runner.fail (lib/runner.js:206:11)
      at Runner.uncaught (lib/runner.js:583:8)
      at process.uncaught (lib/runner.js:612:10)

dstjules:~/git/mocha (master=)
$ git checkout issue-1604
Switched to branch 'issue-1604'

dstjules:~/git/mocha (issue-1604)
$ ./bin/mocha example.js

  ․․

  1 passing (6ms)
  1 pending
```

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/1618)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -659,7 +659,15 @@ Runner.prototype.runSuite = function(suite, fn) {
  * @api private
  */
 Runner.prototype.uncaught = function(err) {
-  if (err) {
+  var runnable = this.currentRunnable;
+  var pending;
+
+  if (err instanceof Pending) {
+    if (runnable.type === 'test') {
+      runnable.pending = true;
+    }
+    pending = true;
+  } else if (err) {
     debug('uncaught exception %s', err !== function() {
       return this;
     }.call(err) ? err : (err.message || err));
@@ -668,8 +676,6 @@ Runner.prototype.uncaught = function(err) {
     err = undefinedError();
   }
   err.uncaught = true;
-
-  var runnable = this.currentRunnable;
 
   if (!runnable) {
     runnable = new Runnable('Uncaught error outside test suite');
@@ -693,7 +699,16 @@ Runner.prototype.uncaught = function(err) {
   if (runnable.state) {
     return;
   }
-  this.fail(runnable, err);
+
+  if (pending && runnable.type === 'test') {
+    // Async skip() call from test
+    this.emit('pending', runnable);
+  } else if (pending) {
+    // Async skip() call from hook
+    runnable.callback(err);
+  } else {
+    this.fail(runnable, err);
+  }
 
   // recover from test
   if (runnable.type === 'test') {
@@ -717,8 +732,10 @@ Runner.prototype.uncaught = function(err) {
     return this.nextSuite(errSuite);
   }
 
-  // bail
-  this.emit('end');
+  // bail on hooks if not skipped
+  if (!pending) {
+    this.emit('end');
+  }
 };
 
 /**
--- a/test/integration/fixtures/pending/skip.async.before.js
+++ b/test/integration/fixtures/pending/skip.async.before.js
@@ -1,0 +1,17 @@
+describe('skip in asynchronous before', function() {
+  before(function(done){
+    var hook = this;
+    setTimeout(function() {
+      hook.skip();
+      throw new Error('never thrown');
+    }, 0);
+  });
+
+  it('should never run this test', function() {
+    throw new Error('never thrown');
+  });
+
+  it('should never run this test', function() {
+    throw new Error('never thrown');
+  });
+});
--- a/test/integration/fixtures/pending/skip.async.beforeEach.js
+++ b/test/integration/fixtures/pending/skip.async.beforeEach.js
@@ -1,0 +1,17 @@
+describe('skip in asynchronous beforeEach', function() {
+  beforeEach(function(done){
+    var hook = this;
+    setTimeout(function() {
+      hook.skip();
+      throw new Error('never thrown');
+    }, 0);
+  });
+
+  it('should never run this test', function() {
+    throw new Error('never thrown');
+  });
+
+  it('should never run this test', function() {
+    throw new Error('never thrown');
+  });
+});
--- a/test/integration/fixtures/pending/skip.async.spec.js
+++ b/test/integration/fixtures/pending/skip.async.spec.js
@@ -1,0 +1,11 @@
+describe('skip in asynchronous test', function() {
+  it('should skip when called from a test', function(done) {
+    var test = this;
+    setTimeout(function() {
+      test.skip();
+      throw new Error('never thrown');
+    }, 0);
+  });
+
+  it('should run other tests in the suite', function() {});
+});
--- a/test/integration/pending.js
+++ b/test/integration/pending.js
@@ -60,4 +60,47 @@ describe('pending', function() {
       });
     });
   });
+
+  describe('asynchronous skip()', function() {
+    this.timeout(1000);
+
+    describe('in spec', function() {
+      it('should immediately skip the spec and run all others', function(done) {
+        run('pending/skip.async.spec.js', args, function(err, res) {
+          assert(!err);
+          assert.equal(res.stats.pending, 1);
+          assert.equal(res.stats.passes, 1);
+          assert.equal(res.stats.failures, 0);
+          assert.equal(res.code, 0);
+          done();
+        });
+      });
+    });
+
+    describe('in before', function() {
+      it('should skip all suite specs', function(done) {
+        run('pending/skip.async.before.js', args, function(err, res) {
+          assert(!err);
+          assert.equal(res.stats.pending, 2);
+          assert.equal(res.stats.passes, 0);
+          assert.equal(res.stats.failures, 0);
+          assert.equal(res.code, 0);
+          done();
+        });
+      });
+    });
+
+    describe('in beforeEach', function() {
+      it('should skip all suite specs', function(done) {
+        run('pending/skip.async.beforeEach.js', args, function(err, res) {
+          assert(!err);
+          assert.equal(res.stats.pending, 2);
+          assert.equal(res.stats.passes, 0);
+          assert.equal(res.stats.failures, 0);
+          assert.equal(res.code, 0);
+          done();
+        });
+      });
+    });
+  });
 });


danielstjules:
@a8m @dasilvacontin :)

a8m:
Thanks Dan, Looks good, but there's an issue with the global suite.
for this one: 
```js
beforeEach(function() {
   this.skip()
})
it('should never run this test', function() {
  throw new Error('never thrown')
})

describe('Should run this suite', function() {
  it('run this test case', function() {
    // ..
  })
})
```
I get the following:
```sh
 - should never run this test
  Should run this suite
    ✓ run this test case


  1 passing (10ms)
  1 pending
```
but with this one:
```js
beforeEach(function(done) {
  var hook = this
  setTimeout(function() {
    hook.skip()
  })
})
it('should never run this test', function() {
  throw new Error('never thrown')
})

describe('Should run this suite', function() {
  it('run this test case', function() {
    // ..
  })
})
```
I get the following:
```sh
  - "before each" hook

  0 passing (9ms)
  1 pending
```
let me know if something isn't clear.

danielstjules:
Looks like you found a bug with the existing skip() behavior with your first example: `run this test case` should be skipped as well. It's not unique to the root suite, either:

``` javascript
  beforeEach(function() {
    console.log('hook');
    this.skip();
  });
  it('should never run this test', function() {
    throw new Error('never thrown');
  });
  describe('Should run this suite', function() {
    it('should never run this test', function() {
      throw new Error('never thrown');
    });
  });
});

// Result
$ mocha test.js

  hook

  ․hook
․

  0 passing (6ms)
  1 pending
  1 failing
```

As for the second example, I'll get a fix ready for that, thanks! :)

danielstjules:
@a8m Fixed!

``` javascript
// test.js
beforeEach(function(done) {
  var hook = this
  setTimeout(function() {
    hook.skip()
  })
})
it('should never run this test', function() {
  throw new Error('never thrown')
})
describe('Should run this suite', function() {
  it('run this test case', function() {
    // ..
  })
});

// output
$ ./bin/mocha --reporter spec test.js


  - should never run this test
  Should run this suite
    ✓ run this test case


  1 passing (10ms)
  1 pending
```

It's probably worth discussing whether or not asynchronously skipping a test is functionality we'd like to support. Could be a rabbit hole. :) What do you think @boneskull ?

dasilvacontin:
What's the benefit of async skipping a test vs delayed suite definition? Besides maybe cleaner?

danielstjules:
That may be the only reason - it can look cleaner. The same applies to using skip() from a synchronous spec.

``` javascript
  var someCondition = getConditionValue();

  // In conditional block, which isn't ideal as it
  // won't show up as pending
  if (someCondition) {
    it('test', function() {
      assert(true);
    });
  }

  // Handling the spec so that it'll show as pending
  var test = (someCondition) ? it : it.skip;
  test('test', function() {
    assert(true);
  });

  // Using skip()
  it('test', function() {
    if (!someCondition) return this.skip();
    assert(true);
  });
});
```

dasilvacontin:
If it's well tested, I wouldn't mind using it.

Btw, not picky concretely with this case, as there are many more existing tests in this style, but I personally prefer making tests which must all pass. e.g., for this case, ideally I would run mocha inside mocha, asserting the quantity of skipped tests or which tests were skipped. The problem with this is the `uncaughtException` listener, that messes all up, since the exceptions of the mocha instance you are testing are caught by the mocha instance you are using to test.

I don't really like having tests that are expected to be skipped, since if they suddenly start passing one does not notice unless he/she reads carefully the `npm test` output line by line.

danielstjules:
Completely agree! Full integration tests for some cases would be nice. You could execute mocha and check it's output, kinda like https://github.com/danielstjules/buddy.js/blob/master/spec/integration/buddySpec.js If we were to pursue running mocha in mocha, it would be better to also rely on the last known good release of mocha, rather than using `bin/mocha`. Otherwise you end up with the same problem.

danielstjules:
> If it's well tested, I wouldn't mind using it.

I think the existing behavior of `skip()` (non-async) is likely inconsistent. See https://github.com/mochajs/mocha/pull/1618#issuecomment-85062257 This async implementation mimics that output. Both could be updated if others agree.

I guess the real reason I'm not sure about async skip is that you're relying on there not being other issues, outside the currently running spec. E.g. if an exception is thrown from some random timeout and wrongly attributed to a spec that would be skipped. It becomes more difficult to reason about. I haven't experienced this scenario, but it seems common among those having trouble with their tests (or codebase).

``` javascript
$ cat test.js
setTimeout(function() {
  throw new Error('ohno!');
}, 100);

it('test', function(done) {
  setTimeout(function() {
    this.skip();
  }.bind(this), 100);
});

danielstjules:~/git/mocha (issue-1604 =)
$ ./bin/mocha --reporter spec test.js


  1) test

  0 passing (105ms)
  1 failing

  1)  test:
     Uncaught Error: ohno!
      at null._onTimeout (test.js:2:9)
```

dasilvacontin:
True. Actually, I started creating tests for async errors ending up in other suites, but the `.uncaughtError` thing was a barrier.

a8m:
Tested. works good @danielstjules .

> I think the existing behavior of skip() (non-async) is likely inconsistent.

Agree, but at least 'async' aligned to behave the same way as 'non-async'.

danielstjules:
Thanks @a8m!

@dasilvacontin I'm going to add some "integration" tests for this, under a new test directory.

danielstjules:
@dasilvacontin Got some quick integration tests running.

@travisjeffery @boneskull Mind taking a look? :)

travisjeffery:
lgtm

a8m:
LGTM

travisjeffery:
before we merge it i just wanna think about whether we should add it, or just say used a delayed suite. considering pretty much no one in the thread has said they'd actually use this

danielstjules:
Though I don't personally use `skip()`, I think that if it's available for sync code, it might makes sense to offer for async specs/hooks as well? The alternative doesn't seem as clean to me: https://github.com/mochajs/mocha/pull/1618#issuecomment-85255282

danielstjules:
Since the integration tests are a bit slower, and the acceptance tests are ran against the jsapi, I've moved them into their own directory. This way `make test` will still only run faster unit tests, but `make test-integration`, `npm test` or `make test-all` will include all integration tests.

danielstjules:
@mochajs/mocha Any thoughts? :)

dasilvacontin:
Couldn't look at it, but damn, those tests are sexy.

I'd rather let this be merged by either @travisjeffery or @boneskull too. Besides, @travisjeffery said he wanted to taco about it.

a8m:
> but damn, those tests are sexy.

I fully agree with you!
Great work @danielstjules

dasilvacontin:
Even if this doesn't get merged, I could use the helper for the tests I was doing for `BaseReporter`.

boneskull:
@danielstjules Does this make `--delay` redundant?

danielstjules:
I don't think so, though there's some overlap in the problems that they can solve.

danielstjules:
@boneskull To achieve dynamic skip like functionality using defer, I believe you end up having to resort to roughly the same logic as I mentioned in https://github.com/mochajs/mocha/pull/1618#issuecomment-85255282

danielstjules:
@boneskull were you looking for additional details for this PR?

boneskull:
@danielstjules nope.

jbnicolai:
@boneskull any reason to hold of on merging?

danielstjules:
I'll have to update the PR. And while the PR targeted master, maybe I should change it to 2.3.0?

danielstjules:
@jbnicolai @boneskull Just rebased in master

danielstjules:
@mochajs/mocha If there's interest in bringing this into 2.3.0, I'll do another quick rebase. It would offer some additional flexibility, solving the problem discussed in https://github.com/mochajs/mocha/issues/1480#issuecomment-118866370

jbnicolai:
@danielstjules big :+1: as far as I'm concerned.

boneskull:
@jbnicolai if we do release the major, we need to create a new milestone and retarget everything targeted to v3.0.0 to v4.0.0.  then we can rebase branch v3.0.0 into master.

anyway, it looks like this needs rebasing (again), so whoever does that can merge it

danielstjules:
@boneskull Rebased. Should I target 2.3.0, or is master fine?

boneskull:
@danielstjules sorry, didn't see this one.  we can release it in 2.3.1 if you wish

danielstjules:
@boneskull Did you mean 2.3.0? If so, I can rebase again and get things ready to merge :) Doesn't look like 2.3.0 has been tagged yet https://www.npmjs.com/package/mocha

boneskull:
@danielstjules I'm about to tag v2.3.0

bcoe:
I'm jumping through hoops to get functionality similar to async skip() in the node_redis test-suite, would love this feature.

danielstjules:
I'll pick this back up - should be pretty easy to get it working with latest master. However, I'm not sure when a realistic date would be for another minor release.

laughinghan:
What's blocking this from merging? This doesn't appear to have made it to 2.3 nor 2.4.

> before we merge it i just wanna think about whether we should add it, or just say used a delayed suite. 

What is a "delayed suite"? The only mention of "delay" I can find in the docs is delaying the root suite, which, I can't imagine any use cases for async skipping tests that would be solved by that.

And yes, my use case does require skipping an async test, I'm testing an interaction between my code and `.select()` and `.blur()` and the resulting `focus` and `blur` events on a DOM textarea. The browser behavior that used to cause a bug in my library only happens when the browser window is focused, so this particular test is incorrectly failing when I run my test suite in a background tab, but why should I have to manually grep to exclude/include this test when I decide whether I want to run the tests in a foreground or background browser tab? The code is perfectly aware of whether the browser window is focused without me manually passing in a tag with grep, and should be perfectly able to report whether the test passed, failed, or was skipped because the browser window wasn't focused.

Is there any help I can offer to get this merged?

boneskull:
I think @danielstjules just needs to clean it up a bit, but I'm not sure.  perhaps he can comment.

boneskull:
(the tests for this PR aren't actually passing, so something is broken)

danielstjules:
This solution doesn't work anymore after a patch release changing how some error handling was done, haven't been able to come up with an alternative yet.

PR #1621, opened 2015-03-24
Pass `--allow-natives-syntax` flag to node.

keithamus:
This allows the passing of the `--allow-natives-syntax` flag into the node binary.

[Some literature about allow-natives-syntax](http://dailyjs.com/2014/06/26/optimization-killers/)


dasilvacontin:
Thanks!

Issue #1614, opened 2015-03-21
String diffs not shown anymore when using node assert.equal


mantoni:
I created a gist that reproduces the issue for me on OS X with node 0.10.32:
https://gist.github.com/mantoni/e2927ced1484f01b1fb8

Thanks for looking into it!

a8m:
I'm not sure what do you mean by saying "anymore", in which version it works @mantoni ?
as far as I know, using `assert` never was showing diff. for [this](https://github.com/mochajs/mocha/blob/master/lib/reporters/base.js#L182) reason.

mantoni:
Can't say exactly when it stopped working, but I took this screenshot once: https://www.npmjs.com/package/consolify

mantoni:
Thanks for the pointer @a8m, I removed the `err.showDiff &&` part from the conditional in the line you mentioned and then it works perfectly fine.

a8m:
kk, I'm on it.
trying to detect in which version it's stop to work.

a8m:
this commit: https://github.com/a8m/mocha/commit/a5d7240a83f934b16b98b07f170c2d6288d58d7d

a8m:
Thanks for pointing that out @mantoni.

a8m:
Is this something that we want to fix @mochajs/mocha ? if so, I'll label it as a "bug", and start to working on it.

dasilvacontin:
So it depends on whether the asserting library sets `showDiff` to true or not? Like, as an user, could you do something to show the diff, without modifying the library?

mantoni:
It's a feature that is advertised on the mocha page, but it only works in specific setups. So if this is bot going to considered a bug, it could be a new flag to opt in and it would be nice to pit some notes on how to get it working on the web page.

a8m:
Yes, but it's worked before(`<1.21.5`) @dasilvacontin . that's the reason I see it as a regression.

dasilvacontin:
Could you please answer my doubts/questions?

Regarding if it's a regression or not, we are making people waste time trying to figure out why the diff is not showing, and how to make it show it again. That's something I would only do in a major version. We need to get this fixed this asap.

a8m:
> So it depends on whether the asserting library sets showDiff to true or not, right?

Yes.

> Like, as an user, could you do something to show the diff, without modifying the library?

If you're using `assert`, I don't see any other way except this:
```js
try {
  assert.equal("foo", "bar")
} catch(e) {
[...]

Issue #810, opened 2013-04-13
Cannot disable diffs - err.showDiff not working


vjpr:
When `err.showDiff` is `false` it does not prevent a diff being shown.

Here is the only place where `showDiff` is being used in Mocha:

```
    // msg
    var err = test.err
      , message = err.message || ''
      , stack = err.stack || message
      , index = stack.indexOf(message) + message.length
      , msg = stack.slice(0, index)
      , actual = err.actual
      , expected = err.expected
      , escape = true;

    // explicitly show diff
    if (err.showDiff) {
      escape = false;
      err.actual = actual = JSON.stringify(actual, null, 2);
      err.expected = expected = JSON.stringify(expected, null, 2);
    }
```

I think its just been overlooked.

PR #1626, opened 2015-03-25
fix(reporter/base): explicitly ignore showDiff #1614

a8m:
Ignore **showDiff** only if it's explicitly set to `false`.
see #1614 , #810

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -179,7 +179,7 @@ exports.list = function(failures){
       msg = 'Uncaught ' + msg;
     }
     // explicitly show diff
-    if (err.showDiff && sameType(actual, expected)) {
+    if (err.showDiff !== false && sameType(actual, expected)) {
 
       if ('string' !== typeof actual) {
         escape = false;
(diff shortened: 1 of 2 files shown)


danielstjules:
:tada: :shipit: 

Also, those specs are a perfect example of why we should fix https://github.com/mochajs/mocha/issues/1582 If `Base.list` failed, `process.stdout.write` wouldn't be restored. And we can't just do it from `afterEach` because then the reporter won't print the spec at all. I'm gonna tackle that tonight. :)

mantoni:
Thanks @a8m :+1:

dasilvacontin:
LGTM.

dasilvacontin:
thanks guys!

a8m:
Thanks @dasilvacontin

dominicbarnes:
Awesome! Was just wondering about this and glad to see it's been addressed. How long until a release?

dasilvacontin:
Euh, best way to force a sooner release is to make the **releasers** aware of the need. :smile:

/cc @travisjeffery @boneskull

PR #1631, opened 2015-03-27
fix(reporter/base): don't assume error message is first line of stack

mantoni:
The current implementation assumes that the first line of the err.stack is the error message. This is not the case for Safari. When using Mocha with a different HTML generator (e.g. Consolify) this produces weirdly formatted output with no error message.

I don't know how to come up with a test case for this. Please advise if there is anything I can provide.
The attached screenshot was produced with `mochify --consolify runner.html`.

stacktrace

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -168,11 +168,18 @@ exports.list = function(failures){
     var err = test.err
       , message = err.message || ''
       , stack = err.stack || message
-      , index = stack.indexOf(message) + message.length
-      , msg = stack.slice(0, index)
+      , index = stack.indexOf(message)
       , actual = err.actual
       , expected = err.expected
       , escape = true;
+    if (index === -1) {
+      msg = message;
+    } else {
+      index += message.length;
+      msg = stack.slice(0, index);
+      // remove msg from stack
+      stack = stack.slice(index + 1);
+    }
 
     // uncaught
     if (err.uncaught) {
@@ -198,9 +205,8 @@ exports.list = function(failures){
       }
     }
 
-    // indent stack trace without msg
-    stack = stack.slice(index ? index + 1 : index)
-      .replace(/^/gm, '  ');
+    // indent stack trace
+    stack = stack.replace(/^/gm, '  ');
 
     console.log(fmt, (i + 1), test.fullTitle(), msg, stack);
   });
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -186,7 +186,8 @@ exports.list = function(failures){
       msg = 'Uncaught ' + msg;
     }
     // explicitly show diff
-    if (err.showDiff !== false && sameType(actual, expected)) {
+    if (err.showDiff !== false && sameType(actual, expected)
+        && expected !== undefined) {
 
       if ('string' !== typeof actual) {
         escape = false;
--- a/test/reporters/base.js
+++ b/test/reporters/base.js
@@ -40,8 +40,8 @@ describe('Base reporter', function () {
       Base.list([test]);
 
       errOut = stdout.join('\n');
-      errOut.should.match(/actual/);
-      errOut.should.match(/expected/);
+      errOut.should.match(/\- actual/);
+      errOut.should.match(/\+ expected/);
     });
 
     it('should show diffs if property set to `true`', function () {
@@ -55,8 +55,8 @@ describe('Base reporter', function () {
       Base.list([test]);
 
       errOut = stdout.join('\n');
-      errOut.should.match(/actual/);
-      errOut.should.match(/expected/);
+      errOut.should.match(/\- actual/);
+      errOut.should.match(/\+ expected/);
     });
 
     it('should not show diffs when showDiff property set to `false`', function () {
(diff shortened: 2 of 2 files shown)


dasilvacontin:
Thanks for reporting! I guess we can detect this case-scenario by checking if there is a line number at the end of the string.

dasilvacontin:
Oh sorry, I didn't notice this was a PR already. :smile:

Okay, let me check one thing, because line 178 looks odd to me.

lib/reporters/base.js
```diff
@@ -168,11 +168,17 @@ exports.list = function(failures){
...
+      , index = stack.indexOf(message)
       , actual = err.actual
       , expected = err.expected
       , escape = true;
+    if (index === -1) {
+      msg = message;
+    } else {
+      index += message.length;
+      msg = stack.slice(0, index);
+      stack = stack.slice(index + 1)
```
dasilvacontin: This is unnecessary and it eating the stack since the code is already doing it at line `208`: https://github.com/mochajs/mocha/pull/1631/files#diff-5c8a5788ecd4b4af2a29bc27ac4ae986R208

dasilvacontin:
LGTM if done the required change.

lib/reporters/base.js
mantoni: Oh yes, thank you. I removed the `stack.slice` call at line 208, added a missing `;` and fixed the code doc. Please have another look.

dasilvacontin:
squash commits and I'll gladly merge :)

mantoni:
Done.

a8m:
Thanks @mantoni 
Can you please add tests for this ?

dasilvacontin:
Yeah. In fact, it's a bit alarming the fact that you were eating the stacktrace and the build was passing anyways.

My code review would have been unnecessary if we had proper tests. :smile:
(and what's worse, if I had failed to notice the mistake during the code review, I would have merged an annoying bug)

mantoni:
I agree. That is why I asked for advice on the PR about where and how to put tests for this. If you can give me quick pointers, maybe an example of a similar test, I'm happy to write some.

dasilvacontin:
Oh sorry, I missed the tests part. I guess I was in a big rush, because I even missed the fact that it was a PR in the beginning... :smile:

I guess they should go in `test/reporters/base.js`.

a8m:
> I guess they should go in test/reporters/base.js.

Yes.

Feel free to ping me if you need some help, @mantoni

mantoni:
Ok, I've made the reporter/base implementation less repetitive now in a first commit and the fix for the issue now comes after that and contains two test cases. Let me know if there is anything else I can improve.

mantoni:
Fixed another issue with diff titles being shown for thrown errors. This could have gone into a separate pull request, but I already depend on the changes made in this pull request for the test cases.

mantoni:
This brought an issue with node 0.8 to the light because stack traces contain the word `actual` in `at Context.actual`. This means that a couple of asserts where always true in node 0.8 regardless of whether  the diffs where shown. Therefore I changed the affected asserts to verify against the more precise `- actual` and `+ expected`.

dasilvacontin:
Awesome, thanks @mantoni!

a8m:
You should squash those 3 commits before you merged them. @dasilvacontin 
Thanks @mantoni

dasilvacontin:
I was going to ask for squashing, but they seemed like three different changes to me. Sorry if it's not the case.

lib/reporters/base.js
```diff
@@ -168,18 +168,26 @@ exports.list = function(failures){
...
+    }
 
     // uncaught
     if (err.uncaught) {
       msg = 'Uncaught ' + msg;
     }
     // explicitly show diff
-    if (err.showDiff !== false && sameType(actual, expected)) {
+    if (err.showDiff !== false && sameType(actual, expected)
+        && expected !== undefined) {
```
a8m: Can you explain me this case? @mantoni

mantoni: In case an ordinary Error is thrown, there is no `expected` on the error object. In this case the diff header was shown, but no diff since both values are undefined. I've also added a test case for this.

mantoni:
Thanks for the merge!

I actually took some time to rebase things into good shape. The first commit is a test case cleanup / refactoring, the second and third are independent bugfixes. Sorry if things got confusing.

lib/reporters/base.js
a8m: > In this case the diff header was shown, but no diff since both values are undefined.

Sorry, I don't get how this is possible(to get an error if the two values are equals).

mantoni: `actual` and `expected` are both `undefined` in the sense of "not there". They are both equal in this case. Remove this check and run the test cases and you will see :)

a8m: Sure, you throwing an error with impossible fields. 
`actual` and `expected`, they both `undefined`(there are equals).

mantoni:
Why is that impossible? `throw new Error()` has this behavior. This fixes an issue that is very easy to reproduce and I provided a test case for it. What else can I do to convince you that this condition is required? Shall I put a gist together that demonstrates the issue?

a8m:
> Why is that impossible? 

Because Mocha as an error convention(maybe the other collaborators fix me if I wrong).

>  Shall I put a gist together that demonstrates the issue.

yes, sense one

Thanks :) @mantoni

mantoni:
Here is a gist to demonstrate the issue fixed in https://github.com/mantoni/mocha/commit/18d2e05253f0cba3c7396ffb0727ec96327cedbb: https://gist.github.com/mantoni/7ecebd3316a6a851e782

a8m:
haha, I wrote a "sense one".
of course it's possible, but it makes no sense.

a8m:
let's leave it right now.
Thanks @mantoni

dasilvacontin:
:shipit:

Issue #1632, opened 2015-03-27
a more descriptive/user-friendly error when describe title is not provided


dasilvacontin:
```js
describe(function () {
    before(function () {
        this.credentials = {id: 123};
    })

    it('should work', function () {
        expect(this.credentials.id).to.equal(123);
    })
})
```

```bash
/usr/local/lib/node_modules/mocha/lib/interfaces/bdd.js:49
      fn.call(suite);
         ^
TypeError: Cannot call method 'call' of undefined
    at context.describe.context.context (/usr/local/lib/node_modules/mocha/lib/interfaces/bdd.js:49:10)
    at Object.<anonymous> (/Users/dasilvacontin/temp/test/kek.js:1:63)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /usr/local/lib/node_modules/mocha/lib/mocha.js:187:27
    at Array.forEach (native)
    at Mocha.loadFiles (/usr/local/lib/node_modules/mocha/lib/mocha.js:184:14)
    at Mocha.run (/usr/local/lib/node_modules/mocha/lib/mocha.js:405:31)
    at Object.<anonymous> (/usr/local/lib/node_modules/mocha/bin/_mocha:405:16)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

dasilvacontin:
And if the `it` title is not set... well, it thinks that it's a pending test and the title is the function:
```js
  setting vars inside before
    - function () {
        expect(this.credentials.id).to.equal(123);
    }


  0 passing (5ms)
  1 pending
```

1999:
@boneskull probably it's time to close this issue now?

dasilvacontin:
@1999 thanks for pointing it out! Closing.

PR #1634, opened 2015-03-30
feat(interfaces/*): support untitled suites/tests #1632

a8m:
Further to issue #1632, I created this draft.
let's discuss about it.


travisjeffery:
the other option would be to throw an error with a good message, that would probably result in better practices/readable tests imo

a8m:
So there are two options:
1. use a fallback title - `Untitled suite/test`
2. throw an error with a usage message (it means to failing, and not run at all).
Please correct me if I get you wrong @travisjeffery 

I prefer the second option, but the first one is more forgiving.
/cc @dasilvacontin

travisjeffery:
yeah let's go with throwing an error. i'd rather be opinionated on having people do the right thing.

dasilvacontin:
Yeah, I also prefer throwing an error. As @travisjeffery, said, leads to more readable code and enforces good practice, and it's probably what you wanted to do anyways.

a8m:
ping

danielstjules:
I like the solution. It supports the possibility of creating interfaces that accept empty titles for either suites, tests, or both as well.

a8m:
Thanks @danielstjules 
:smile:

a8m:
I started to implement it on the interfaces side, but then I got the feeling of repeating myself.

danielstjules:
Yea, updating it at the suite/test level is good, since any interfaces can just use an empty string for the name if they'd like to accept "anonymous specs" :)

dasilvacontin:
Was is possible to use an object as a title before? That was later being converted to a string inside our code.

Because if that was possible before, this should be considered breaking. Or we should check for a `.toString()` or something.

a8m:
Using object as suite/test-case title, returning this result (is just `toString` it):
```sh
suite
    ✓ [object Object] 
```
>  this should be considered breaking.

fair enough, I added it to `v3.0.0` milestone

danielstjules:
``` javascript
$ cat example.js
it({foo: 'bar'}, function() {});

$ mocha example.js


  ✓ [object Object]

  1 passing (8ms)
```

``` javascript
$ cat example.js
it(12, function() {});

$ mocha example.js


  ✓ 12

  1 passing (8ms)
```

dasilvacontin:
But what if you accept objects that have a `.toString()` or are not a function?

danielstjules:
@dasilvacontin Because mocha currently works with functions, too.
``` javascript
$ cat example.js
it(function() {}, function() {});

$ mocha example.js


  ✓ function () {}

  1 passing (9ms)
```

I don't personally see this PR as breaking the API and requiring a new major release. Minor would be fine, and would probably be a good fit for 2.3.0 or 2.4.0. It's unspecified behavior. We shouldn't be concerned if someone is relying on passing objects or functions as the spec or suite titles, rather than strings. The API documents that the title should be a string: https://github.com/mochajs/mocha/blob/master/lib/suite.js#L25 https://github.com/mochajs/mocha/blob/master/lib/suite.js#L43 https://github.com/mochajs/mocha/blob/master/lib/test.js#L16

dasilvacontin:
I agree that this does not break the API. But if you make people waste time, you are doing the same harm; you don't want someone's build to suddenly stop working and make them waste time tracking down the issue to fix it.

a8m:
> you don't want someone's build to suddenly stop working and make them waste time tracking down the issue to fix it.

So that's the reason we added this user-friendly message :)

dasilvacontin:
> that's the reason we added this user-friendly message :)

User-friendly message that breaks your build and makes you correct hundreds of tests in the worst case scenario.*

a8m:
We are not breaking the API, this is just a weakness that maybe users take advantage of it.

/cc @travisjeffery

danielstjules:
http://symbolhound.com/?q=it%28%22 / http://symbolhound.com/?q=it%28%27 vs http://symbolhound.com/?q=it%28%7B / http://symbolhound.com/?q=it%28function (the only actual match is a typo) I wish I could find a search engine with better support for symbols lol Any suggestions?

So for:
``` javascript
it(function() {}, function() {
});
it({foo: bar}, function() {
});
it(-1, function() {
});
```
it's still undefined behavior, since the public API docblocks (https://github.com/mochajs/mocha/pull/1634#issuecomment-87801546) specify that title should be a string, and nowhere on http://mochajs.org/ do we mention non-string titles are supported.

@dasilvacontin I hear what you're saying about preserving BC, and agree for the most part, but the issue you highlighted in mustache.js seems to be of a different scope. That said, if a 3.0.0 release is in the works for the coming months, I'm not worried either way. :)

dasilvacontin:
Those queries can't come up with the edge-cases I mentioned.

Still, I agree that the number of people who could be affected by this is very very low. And afaik, since we have some *serious* bugs in mocha (eg #534), I wouldn't bump to v3 solely for this reason, since not getting more bugfixes in v2 would probably out-harm the problems merging this PR could cause.

In the case we were soon bumping to v3, I'd include this for the next major.

boneskull:
imo, If it throws exceptions where it didn't throw exceptions before, it's not backwards-compatible and needs to wait until a major.

a8m:
> it's not backwards-compatible and needs to wait until a major.

fair enough.

boneskull:
@a8m this needs rebasing against v3.0.0 as well

a8m:
Replaced with #1809(rebased on top of v3.0.0)

PR #1647, opened 2015-04-08
Reverse multi-line diffs to standard order

nylen:
Display multi-line diffs in the same standard order as `diff`, `git`, etc:

```diff
- actual / removed
+ expected / added
```

The current order is unintuitive which makes it harder to read at a glance:

```diff
+ expected / added
- actual / removed
```

Pending resolution of https://github.com/kpdecker/jsdiff/issues/14 this works by reversing the `added` and `removed` strings passed to `diff.createPatch`, and then reversing the `+` and `-` status of each change.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -384,8 +384,8 @@ function unifiedDiff(err, escape) {
     if (escape) {
       line = escapeInvisibles(line);
     }
-    if (line[0] === '+') return indent + colorLines('diff added', line);
-    if (line[0] === '-') return indent + colorLines('diff removed', line);
+    if (line[0] === '+') return indent + '-' + colorLines('diff removed', line.substring(1));
+    if (line[0] === '-') return indent + '+' + colorLines('diff added', line.substring(1));
     if (line.match(/\@\@/)) return null;
     if (line.match(/\\ No newline/)) return null;
     else return indent + line;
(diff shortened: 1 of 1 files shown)


boneskull:
@nylen I don't understand what this has to do with jsdiff?

boneskull:
@nylen does this depend on #1648 at all?

nylen:
This is really a jsdiff issue but we can and should work around it here, since there doesn't seem to be a clean way to tell jsdiff to do this.

This and #1648 are both usability improvements to diffs, but they are independent changes.

nylen:
@boneskull this PR and #1648 are big wins for diff usability, can they be merged please?

before:
screen shot 2015-04-15 at 2 12 17 pm

after:
screen shot 2015-04-15 at 2 13 29 pm

nylen:
@a8m @boneskull @danielstjules @dasilvacontin @jbnicolai @pkozlowski-opensource @travisjeffery review please?  I'd really like to get this (or something like it) merged.  Same goes for #1648.

nylen:
ONE MONTH LATER

Now that kpdecker/jsdiff#14 is fixed, this PR is obsolete.  The correct fix is to upgrade to the newest version of `jsdiff`.

danielstjules:
Awesome! And sorry for the delay! Glad it was fixed upstream, too, since I use it on other projects :)

PR #1648, opened 2015-04-08
Fix diff colors

### LINKED ISSUE #802: Much of the color output is invisible with solarized terminal theme
If you use the Solarized Dark terminal theme (very popular), then you can't see the majority of the terminal output for success or error. Bower had a similar problem with a certain shade of grey effectively being invisible.

### LINKED ISSUE #1200: Please change color scheme
I've read through https://github.com/visionmedia/mocha/issues/802 and https://github.com/visionmedia/mocha/issues/590

I believe I've found where the issue lies:
Snippet from base.js
https://github.com/visionmedia/mocha/blob/master/lib/reporters/base.js

```
exports.colors = {
    'pass': 90
  , 'fail': 31
  , 'bright pass': 92
  , 'bright fail': 91
  , 'bright yellow': 93
  , 'pending': 36
  , 'suite': 0
  , 'error title': 0
  , 'error message': 31
  , 'error stack': 90
  , 'checkmark': 32
  , 'fast': 90
  , 'medium': 33
  , 'slow': 31
  , 'green': 32
  , 'light': 90
  , 'diff gutter': 90
  , 'diff added': 42
  , 'diff removed': 41
};

var color = exports.color = function(type, str) {
  if (!exports.useColors) return str;
  return '\u001b[' + exports.colors[type] + 'm' + str + '\u001b[0m';
};
```

So the color function takes in a type and uses the conversion table above to get the ANSI escape color.
A chart can be found here:
http://en.wikipedia.org/wiki/ANSI_escape_code#CSI_codes

As you can see:
screenshot from 2014-04-25 03 09 43

Seems to be a closed-source terminal from IBM AIX.
I believe nobody is using mocha on it. (Correct me if I'm wrong)

Furthermore, I found the ECMA specification:
http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-048.pdf
Here is page 62:
mocha2

Color 90-99 are not in the specification. Most terminals wouldn't care implementing these colors correctly and god knows what color they'll fall back into.
And not everyone uses iTerm2, which has a contrast setting.
Please use color 30-37, if you want increased intensity then ```\u001b[30;1m``` would use the intense variant. 
I think users are only responsible for configuring their terminals so the 16 colors look great on their bg/fg colors. So please don't just say *"It's the theme's fault"* and close the issue. It's mocha's problem not using color from the standard palette.

nylen:
This fixes diffs being unnecessarily hard to read:

Before
screen shot 2015-04-08 at 2 19 29 pm

After
screen shot 2015-04-08 at 2 22 03 pm

Others are experiencing the same issue with standard terminal color schemes too:  https://github.com/mochajs/mocha/issues/1200#issuecomment-59562618

- Fixes #802
- Fixes #1200
- Fixes #1212 for diff output

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -66,8 +66,8 @@ exports.colors = {
   , 'green': 32
   , 'light': 90
   , 'diff gutter': 90
-  , 'diff added': 42
-  , 'diff removed': 41
+  , 'diff added': 32
+  , 'diff removed': 31
 };
 
 /**


vyp:
This might be my own fault, but would be it possible to keep the background colors untouched? As in, have `32` instead of `'32;40'` and `31` instead of `'31;40'`? Would that look ugly for you though?

Because for me, the first command's output is with values `'32;40'` and `'31;40'`, but the second is with values of just `32` and `31`, and clearly the second output looks better: http://i.imgur.com/rwnImmI.png

vyp:
Actually, if you want to explicity set the background, then it should be 49 I think, not 40. (i.e. `'32;49'` and `'31;49'`) How does 49 look for you? I think in theory it should look the same for your after screenshot. For me it looks normal, white background as in the second output of my screenshot above.

nylen:
I'm fine with trying out additional variations if someone from @mochajs/mocha can weigh in first.  For now, it seems like no one is watching the PRs here.

cc @a8m @boneskull @danielstjules @dasilvacontin @jbnicolai @pkozlowski-opensource @travisjeffery

vyp:
Okay but 49 I think is the "default background color", so I think it should be that, if it indeed doesn't look ugly that is. (And 40 is "black".)

danielstjules:
I think the contrast will vary quite a bit depending on your terminal color profile. However, I do prefer the changes applied by this PR as it bring it closer to standard diff tools (svn diff, git diff)

vyp:
It will obviously, but 40 being "black" (which is probably a dark color for most themes) doesn't play nice with light-background themes, although it does with dark-background themes. But 49 I think should play nice with both? Have you checked if 49 looks different than 40 for your own case @danielstjules?

vyp:
Source for that: https://web.archive.org/web/20150319145101/http://misc.flogisoft.com/bash/tip_colors_and_formatting#background

Although I'm not sure how credible that page is (but it ranks highly in google search results if that counts at all).

nylen:
I don't care much about explicitly setting the background color.  It looks nicer for me that way but that's just OS X Terminal.app being weird, and I don't want to break things for people with light terminal themes.

Updated the PR accordingly - @danielstjules I think this is good to go now.  Also, could you take a look at #1647?  It's another PR intended to make mocha's diff output consistent with standard tools.

vyp:
@nylen No worries. :+1:

nylen:
@vyp @danielstjules is this library still maintained??

danielstjules:
I'm happy with this. +1 from me. Just wanna hear from one more contributor.

@mochajs/mocha

boneskull:
I'll take a gander at this later.

boneskull:
@nylen thanks!

FelikZ:
This does not Fixes #802 for me. Because `90` is a bg color in Solarized theme and it still used on titles:

    exports.colors = {
        'pass': 90
        , 'error stack': 90
        , 'fast': 90
        , 'light': 90
        , 'diff gutter': 90
    }; 

I suggest to replace them to `94` which is looks really ok both in Solarazied and default terminal color scheme.

Solarazed:
solarized

Normal:
normal_colors

legomind:
Any progress on this?
When using solarized, error stacks are invisible.

screenshot12

nmschulte-aviture:
+1 for this issue -- I'm using the popular [solarized](https://github.com/altercation/solarized) scheme with [rxvt-unicode](http://software.schmorp.de/pkg/rxvt-unicode.html), via [solarized/xresources](https://github.com/solarized/xresources).  My results are the same as @legomind's: unusable.

dasilvacontin:
Same here. I'm interested in seeing this fixed, but I don't have much time lately...

nmschulte-aviture:
So, 1) I just realized this isn't an issue, but a pull-request.  2) I updated to use the latest version and the error does not exist.  @dasilvacontin, @legomind

Issue #808, opened 2013-04-13
make --grep case insensitive?


eldargab:

erezesk:
Hey,

Since grep is now regexp based, I tried sending ```/pattern/i``` but didn't seem to work
Is there a different way of getting the Grep to be case insensitive? 

Thanks :)

boneskull:
Closed via 6126e38566abe2735491deb8b22b4ef88d50226c

shanebdavis:
What was the conclusion on this issue? Case sensitivity plagues me daily - I miss tests when I'm greping because of case miss-match. Would a patch be accepted that adds a case-insensitive option for web-mocha and command-line mocha? --grep-ignore-case perhaps?

boneskull:
@shanebdavis `--grep` accepts a regular expression, including a flag.  `--grep '/foo/i'` would find Foo and Foo and FOO etc.

holm:
I would vote to change the default grep to be case-insensitive, when the grepped string is not "regex-like". This could be done in the next major version.

I think the vast majority would find case-insensitive more valuable, and for the edge-cases where that is not desired, they can specify a regex to get exactly what they want.

boneskull:
Is `--fgrep` case-insensitive?  If not, that may be a better candidate.

boneskull:
(`--fgrep` is a string-only search; if you want case-sensitive string searches, use `--grep <not-a-regex-string>`)

PR #1650, opened 2015-04-10
feature(mocha/grep): improve grep - issue #808

a8m:
Add the ability to pass `grep` as a regexp-like string. (query in the browser or flag in the cli).
This improvement gives you pass a regexp that including flags, and solved #808 as well.

e.g: 
```sh
$ mocha test/ --grep /^Suite/i
```


danielstjules:
LGTM! :)

Some might argue that it breaks BC, and would require a major bump. e.g.:
``` javascript
it('/test/i', function() {});
it('test', function() {});
```
with something like `mocha --grep /test/i test.js`. Before this PR, you'll run one spec. After, you'll run two. But it's certainly an edge case, as with most BC discussions.

dasilvacontin:
I couldn't take a look at it yet.

Def break BC, but since we have other changes that do so too we can just ship them together in the next major.

boneskull:
@a8m possible to rebase against branch `v3.0.0`?

a8m:
Sure, l'II do it later.
same with #1634, #1591

thx

a8m:
Rebased. but it seems to be causing an error on Travis.
Is there anyone who understand what the problem in [this build](https://travis-ci.org/mochajs/mocha/builds/59005927)?
/cc @mochajs/mocha

dasilvacontin:
old npm version carret issue

jbnicolai:
Yep, not an actual issue.

@a8m great feature! Let's merge and ship asap :)

boneskull:
@a8m this still needs to be rebased... not sure what's happening here.

a8m:
Replaced with #1808(rebased on top of v3.0.0)

PR #1652, opened 2015-04-13
Throw an exception when timeout too large.

### LINKED ISSUE #1644: throw an exception when using a big value for timeout
The exception should remind about the behavior of `setTimeout` when the timeout value is greater than 2^31. Related: #1643.

This would be a breaking change, and it should be added on a major version.

callumacrae:
closes #1644

I'm not sure if the error should be thrown or passed to something—let me know if it needs changing.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -65,6 +65,7 @@ Runnable.prototype.__proto__ = EventEmitter.prototype;
 Runnable.prototype.timeout = function(ms){
   if (0 == arguments.length) return this._timeout;
   if (ms === 0) this._enableTimeouts = false;
+  if (ms > Math.pow(2, 31)) throw new RangeError('Timeout too large, must be less than 2^31');
   if ('string' == typeof ms) ms = milliseconds(ms);
   debug('timeout %d', ms);
   this._timeout = ms;
--- a/test/runnable.js
+++ b/test/runnable.js
@@ -40,6 +40,13 @@ describe('Runnable(title, fn)', function(){
     })
   })
 
+  describe('#timeout(ms) when ms>2^31', function(){
+    it('should throw an error', function () {
+      var run = new Runnable;
+      run.timeout.bind(run, 1e10).should.throw(/Timeout too large/);
+    });
+  });
+
   describe('#enableTimeouts(enabled)', function(){
     it('should set enabled', function(){
       var run = new Runnable;


dasilvacontin:
First, thanks for the PR!

I wouldn't call this an 'interval'. I think it would be better to mention `timeout` in the error, since it's the easier way for an user to understand what's happening.

Something more like `Timeout value is too big (> 2^31)`.

callumacrae:
Yeah, not sure why I did that. 6PM commits ¯\_(ツ)_/¯

Fixed!

dasilvacontin:
LGTM!

Since this breaks backwards compatibility, we'll have to wait until a major version is released.

boneskull:
@dasilvacontin You can manually merge this into `v3.0.0` branch

callumacrae:
I sent the PR to the v3.0.0 branch, it doesn't need a manual merge.

dasilvacontin:
Oh true, I thought we only had a `v2.3.0` branch, thanks for pointing that out @boneskull! :)

Ouch, and for pointing that out too @callumacrae!

callumacrae:
:smile:

dasilvacontin:
:shipit:

boneskull:
@dasilvacontin Also, you may want to keep it up-to-date with `v2.3.0`

dasilvacontin:
Will do sir. Can you check Gitter?

danielstjules:
Did this need to go into the next major, rather than minor? Isn't it BC? It can be seen as simply providing a better error message.

Before:
``` javascript
it('test', function(done) {
  this.timeout(Number.MAX_VALUE);
});
```

danielstjules:
Woops, somehow hit comment before I was done writing my reply. :) Here's my actual comment haha

Did this need to go into a new major, rather than minor? It seems to me that it's not breaking BC. It's just offering a more descriptive error message than what was already being thrown.

Before PR:

``` javascript
it('test', function(done) {
  this.timeout(Number.MAX_VALUE);
});

$ mocha example.js


  1) test

  0 passing (14ms)
  1 failing

  1)  test:
     Error: timeout of 1.7976931348623157e+308ms exceeded. Ensure the done() callback is being called in this test.
```

After PR:

``` javascript
$ ./bin/mocha workflow.js


  ․

  0 passing (12ms)
  1 failing

  1)  test:
     RangeError: Timeout too large, must be less than 2^31
      at Test.Runnable.timeout (lib/runnable.js:70:33)
      at Context.timeout (lib/context.js:39:19)
      at Context.<anonymous> (workflow.js:2:8)
      at Test.Runnable.run (lib/runnable.js:234:15)
      at Runner.runTest (lib/runner.js:390:10)
      at lib/runner.js:473:12
      at next (lib/runner.js:315:14)
      at lib/runner.js:325:7
      at next (lib/runner.js:260:23)
      at Immediate._onImmediate (lib/runner.js:292:5)
```

In both instances, whether or applied to suites or specs, the result is all tests failing.

boneskull:
@danielstjules Users may be trying to catch a `RangeError` I guess. 

The output is different; even though it's *unlikely* to cause problems, it's not a chance we need to take.

dasilvacontin:
Good catch @boneskull.

Besides, I think the `Error: timeout of 1.7976931348623157e+308ms exceeded. Ensure the do...` error is not thrown by all invalid timeout values.

danielstjules:
We're going from a general error (Error) to a higher specificity error (RangeError), which still inherits from Error. Isn't this a prime example of a change maintaining BC?

A side effect, though, is that you can now catch the exception. In master, you're not able to from within a spec using the public API:
``` javascript
$ cat example.js
it('test', function(done) {
  try {
    this.timeout(Number.MAX_VALUE);
  } catch (err) {
  }
  setTimeout(done, 100);
});
$ mocha example.js

  ․

  0 passing (6ms)
  1 failing

  1)  test:
     Error: timeout of 1.7976931348623157e+308ms exceeded. Ensure the done() callback is being called in this test.
      at null.<anonymous> (/Users/danielstjules/.nvm/v0.10.20/lib/node_modules/mocha/lib/runnable.js:170:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
```
That side effect isn't a breaking change though since it's adding the possibility to catch an exception where that option previously didn't exist. But if that was seen as breaking, it could be avoided by moving the exception to `Runnable.prototype.resetTimeout`, from which the original was thrown.

While I'm a proponent of semver, I still don't understand how this would require a major? I get the sense that I'm alone on this, but I feel like we might be stretching the BC discussion a bit? I've always been a bit enthusiastic about getting small enhancements/fixes into earlier releases for projects, but I didn't think that was clouding my judgement. Hope I don't come across as confrontational - just looking for clarification.

boneskull:
@danielstjules Sorry, I was reading this ticket on my phone.  I thought we were going from a `RangeError` to an `Error`.  This is not a BC.

boneskull:
If someone wants to cherry-pick this on top of v2.3.0 that's fine.

dasilvacontin:
Hmm, I missed the fact that we were already throwing an error before. But what about:
```js
describe('...', function () {
  it('...', function () {
    try {
      this.timeout(tooBig)
      somethingThatThrowsAnError()
    } catch (e) {
      ...// it's no longer an error from `somethingThatThrowsAnError`
    }
  }
})
```

danielstjules:
> But if that was seen as breaking, it could be avoided by moving the exception to Runnable.prototype.resetTimeout, from which the original was thrown.

In which case we move it :) The example you highlighted is exactly what I mentioned in my last reply. But that's still not a breaking change. The test will still pass.

dasilvacontin:
> The test will still pass.

Or not.

```js
describe('...', function () {
  it('...', function () {
    try {
      this.timeout(tooBig)
      somethingThatThrowsAnError()
    } catch (e) {
      // checking if `somethingThatThrowsAnError` throws error with custom property
      e.someProperty.should.equal(someValue)
    }
  }
})
```

I know it's very unlikely, but I'd rather have zero breaking changes.

dasilvacontin:
Okay, I think we got confused for a moment. This is totally a breaking change.

```bash
➜  temp  cat test/timeout.js
it('test', function(done) {
  this.timeout(4294967296)
  done()
})
```

```bash
➜  temp  mocha -V
2.2.1
➜  temp  mocha


  ✓ test

  1 passing (6ms)

```

```bash
➜  temp  cd ./node_modules/mocha
➜  mocha git:(v3.0.0) git log --pretty=oneline -n 5
cc3f3f94697077d1f6cec9ea8896415f17598f70 Merge remote-tracking branch 'origin/v2.3.0' into v3.0.0
450ecbfb41b6bc6e1a55ba555a8a7395ee1b9b6e Merge pull request #1652 from callumacrae/timeout-exception
ace073d943d0e97d3946811a74cda99015adb025 Throw an exception when timeout too large.
7be4891412b18863bff1fd5874851fd7dd3780f0 Merge commit '7657cb11d960cf2cd8407b256019b2e34dc93328' into v2.3.0
d9a8192cc3f877b69942480d2f6acbe28a82b45c Merge pull request #1631 from mantoni/fix-safari-consolify
➜  mocha git:(v3.0.0) cd ~/temp
➜  temp  ./node_modules/mocha/bin/mocha


  1) test

  0 passing (8ms)
  1 failing

  1)  test:
     RangeError: Timeout too large, must be less than 2^31
      at Context.<anonymous> (test/timeout.js:2:8)

```

danielstjules:
@dasilvacontin I think it's safe to say we disagree, so no worries - I'm not going to push for backporting it to a minor release. But in regards to your last two points.

I don't consider this a breaking change, since it's a bug:
``` javascript
it('test', function(done) {
  this.timeout(4294967296);
  done();
});
```

`this.timeout` is being passed an invalid argument and it's silently ignoring it in this one case. In all other instances, it results in an exception being thrown. E.g. these two both fail with `Error: timeout of 4294967296ms exceeded. Ensure the done() callback is being called in this test`

``` javascript
it('test', function(done) {
  this.timeout(4294967296);
  return new Promise(function(resolve, reject) {
    resolve();
  });
});

it('test', function(done) {
  this.timeout(4294967296);
  setTimeout(done, 100000);
});
```

As for your other example, it goes from failing before the PR, to still failing (not breaking). Of course, they fail for diff reasons :)

``` javascript

function somethingThatThrowsAnError() {
  var err = new Error('foobar');
  err.somePoperty = 'foobar';
}

it('test', function(done) {
  try {
    this.timeout(4294967296);
    somethingThatThrowsAnError();
  } catch (err) {
    // checking if `somethingThatThrowsAnError` throws error with custom property
    assert.equal(err.someProperty, 'foobar');
    done();
  }
});

// Before PR =>
  0 passing (13ms)
  1 failing

  1)  test:
     Error: timeout of 4294967296ms exceeded. Ensure the done() callback is being called in this test.

// After PR =>

  0 passing (14ms)
  1 failing

  1)  test:
     AssertionError: undefined == 'foobar'
      at Context.<anonymous> (test.js:14:12)
      at Test.Runnable.run (lib/runnable.js:234:15)
      at Runner.runTest (lib/runner.js:390:10)
      at lib/runner.js:473:12
      at next (lib/runner.js:315:14)
      at lib/runner.js:325:7
      at next (lib/runner.js:260:23)
      at Immediate._onImmediate (lib/runner.js:292:5)
```

dasilvacontin:
@danielstjules True about the second one. Sorry, I was busy with other stuff and didn't give it enough thought.

danielstjules:
What version of node are you using? And I'm not running it against the 3.0.0 branch since there's lots of other changes in there. I'm running it against this PR applied directly to master (hence if it were pulled down into a minor release)

danielstjules:
Basically, it boils down to whether or not you think this is correct behavior or should be fixed:

```
// throws Error: timeout of 4294967296ms exceeded. Ensure the done() callback is being called in this test
it('test', function(done) {
  this.timeout(4294967296);
  return new Promise(function(resolve, reject) {
    resolve();
  });
});

// throws Error: timeout of 4294967296ms exceeded. Ensure the done() callback is being called in this test
it('test', function(done) {
  this.timeout(4294967296);
  setTimeout(done, 100000);
});

// Does not throw
it('test', function(done) {
  this.timeout(4294967296);
  done();
});
```

dasilvacontin:
I think it should be fixed.

> The example you're using is still taking advantage of what I consider a bug. That's my mistake. Juggling too many snippets for this thread. :)

There's not even documentation about the accepted range for `timeout` on mocha's documentation. We could actually make `timeout` accept any positive value, and behave like most people currently think it works (hence the issue that made this PR happen). You could even see this as an API change that adds a restriction/range to `timeout`'s argument.

The following valid test should pass given mocha's current timeout API, while with this API change it doesn't pass. I'm specially thinking of functions that can sometimes be sync or async, or functions that used to be async and became sync.

```js
function usedToBeAsync (cb) {
  cb()
}

it('test', function(done) {
  this.timeout(4294967296);
  usedToBeAsync(done)
});
```

imo: Fixing `timeout` to accept any positive value would be the bugfix. Adding a new restriction that makes valid tests fail is a breaking change.

I don't mind discussing, it's more interesting than agreeing straight away. :)

callumacrae:
Could always just do this. It's mathematically wrong, but nobody will ever notice:

```js
if (timeout > Math.pow(2, 31)) {
  timeout = Math.pow(2, 31) - 1;
}
```

Issue #1241, opened 2014-06-18
err.showDiff = true breaks string diffs


felixrabe:
Observed:

    2)  should compare strings (raw):

        Error: bad stuff
        + expected - actual

        +"a\nb\nc\nd"
        -"a\nb\n"

Wanted:

    2)  should compare strings (raw):

        Error: bad stuff
        + expected - actual

         a
         b
        +c
        +d

Example code: (see https://github.com/felixrabe/mocha-chai-string-diff/blob/master/test/test.js for the complete example code, run with `mocha`)

    it('should compare strings (raw)', function() {
      var err = new Error('bad stuff');
      err.expected = 'a\nb\nc\nd';
      err.actual = 'a\nb\n';
      err.showDiff = true;  // breaks string diffs
      // err.showDiff = false;  // string diffs are fine
      throw err;
    });

With Chai:

    var expect = require('chai').expect;

    it('should compare strings (chai)', function() {
      try {
        expect('a\nb\n').to.equal('a\nb\nc\nd');
      } catch (err) {
        // string diffs broken because err.showDiff === true
        // err.showDiff = false;  // string diffs are fine
        throw err;
      }
    });

If you uncomment the `// err.showDiff = false;` lines (ie. if you set `showDiff` to `false`), string diffs show up fine.

slang800:
tried out v2.1.0 & this is still broken

a8m:
Fixed in **v2.2.3**

caseywebdev:
Is this really fixed? If anything it looks like [the specs were changed to expect the wrong thing](https://github.com/a8m/mocha/blob/73e49676cc9200292fd5eabd80616130d7a9cf68/test/reporters/base.js#L105). Is anyone else still having this issue in 2.2.4?

a8m:
> Is this really fixed?

yes, you can play it yourself @caseywebdev
thx

caseywebdev:
I don't think it's working correctly (v2.2.4).

Given this test

```js
it('displays line diffs', function () {
  var err = new Error('The Error Message');
  err.expected = 'foo\nbar\nbuz\nbaz';
  err.actual = 'foo\nbar\n';
  throw err;
});
```

the reporter yields

```
  1) displays line diffs

  0 passing (11ms)
  1 failing

  1)  displays line diffs:

      Error: The Error Message
      + expected - actual

      +"foo\nbar\nbuz\nbaz"
      -"foo\nbar\n"

[...]

PR #1655, opened 2015-04-14
fix(reporter/base): issue #1241

a8m:
Follow [1241 discussion](https://github.com/mochajs/mocha/issues/1241#issuecomment-92979416), this PR fix the regression of "strings diff", and solved #1241 as well.
/cc @caseywebdev @domenic


caseywebdev:
I pulled down this commit and can confirm it fixed the issue :+1: Thanks @a8m

Issue #1660, opened 2015-04-18
Unable to assert invalid dates


meandmycode:
When asserting something that may return an invalid date mocha will error trying to report the invalid date due to calling ```toISOString ``` on an invalid date.

Example (using should):

```js
describe('mocha', function() {
  it('should not error when comparing invalid dates', function() {
    new Date('').should.equal(new Date(''));
  })
})
```

I can submit a PR if it helps, I think the stringify utility just needs to call ```toString()``` in the event that the date is invalid, which will result in "[Date: Invalid Date]"

a8m:
Thanks @meandmycode 
PTAL on #1661

meandmycode:
Much appreciated, speedy work!

boneskull:
closed via #1661

PR #1661, opened 2015-04-18
fix(utils/stringify): fix issue #1660

a8m:
Use `toString` for invalid date object


boneskull:
@a8m good idea

a8m:
:squirrel:

PR #1659, opened 2015-04-18
Added option to allow unhandled exceptions to propagate in the browser

amsul:
Related to #553

This adds the `mocha.allowUncaught` method.

When the method **isn’t** invoked, the console is empty:

When the method **is** invoked, the console shows the unhandled exception:

Issue #553, opened 2012-08-23
Option to let exceptions propagate


joliss:
Mocha's browser runner always catches exceptions. I wish it could let them propagate so I can see the stack trace in the JS console.

The line numbers Mocha displays are useful when only (1) you have the file open in the editor, and (2) it's not compiled from CoffeeScript. When either of these conditions isn't met, they turn pretty useless, and you'd really prefer to have a linked stack trace like the console in Chrome or Firebug gives you.

See e.g. http://i.imgur.com/CQV37.png - that trace is not very usable, because now I have to go hunting for the offending line.

berkerpeksag:
This is already fixed in https://github.com/mochajs/mocha/commit/23939a92849cf14d19aaa0a45670106cc41dbbc8 (see #1662) and can now be closed.

danielstjules:
Good call, thanks! :)

PR #1662, opened 2015-04-18
Added option to allow unhandled exceptions to propagate in the browser (rel #1659)

amsul:
- added allowUncaught option (#553)
- allows unhandled exceptions to propagate in the browser
- added tests for allowUncaught option
- global error handler prints to dom with allowUncaught

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -402,6 +402,18 @@ Mocha.prototype.noHighlighting = function() {
 };
 
 /**
+ * Enable uncaught errors to propagate (in browser).
+ *
+ * @return {Mocha}
+ * @api public
+ */
+
+Mocha.prototype.allowUncaught = function(){
+  this.options.allowUncaught = true;
+  return this;
+};
+
+/**
  * Delay root suite execution.
  * @returns {Mocha}
  * @api public
@@ -428,6 +440,7 @@ Mocha.prototype.run = function(fn){
   runner.ignoreLeaks = false !== options.ignoreLeaks;
   runner.fullStackTrace = options.fullStackTrace;
   runner.asyncOnly = options.asyncOnly;
+  runner.allowUncaught = options.allowUncaught;
   if (options.grep) runner.grep(options.grep, options.invert);
   if (options.globals) runner.globals(options.globals);
   if (options.growl) this._growl(runner, reporter);
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -229,21 +229,19 @@ Runnable.prototype.run = function(fn){
   if (this.async) {
     this.resetTimeout();
 
-    try {
-      this.fn.call(ctx, function(err){
-        if (err instanceof Error || toString.call(err) === "[object Error]") return done(err);
-        if (null != err) {
-          if (Object.prototype.toString.call(err) === '[object Object]') {
-            return done(new Error('done() invoked with non-Error: ' + JSON.stringify(err)));
-          } else {
-            return done(new Error('done() invoked with non-Error: ' + err));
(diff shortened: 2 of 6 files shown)


amsul:
@boneskull continuing the PR from #1659

Cheers!

lib/mocha.js
```diff
@@ -402,6 +402,18 @@ Mocha.prototype.noHighlighting = function() {
...
 
 /**
+ * Enable uncaught errors to propagate (in browser).
+ *
+ * @return {Mocha}
+ * @api public
+ */
+
+Mocha.prototype.allowUncaught = function(){
+  this.options.allowUncaught = global.mocha.allowUncaught = true;
```
boneskull: Hmm, taking a closer look, I don't like this.  We should be handling this option the same way we handle others.

```js
Mocha.prototype.allowUncaught = function(){
  this.options.allowUncaught = true;
  return this;
};
```

Below, when you need to read the setting, you'll need to get it a different way.

amsul: Yes, in fact `global.mocha.allowUncaught` doesn’t need to be set at all. The global error handler will still work fine this way in the browser.

amsul:
@boneskull is there anything else that concerns you in this PR?

lib/runnable.js
```diff
@@ -277,4 +275,18 @@ Runnable.prototype.run = function(fn){
       done();
     }
   }
+
+  function callFnAsync(fn) {
+    fn.call(ctx, function(err){
+      if (err instanceof Error || toString.call(err) === "[object Error]") return done(err);
```
boneskull: got a `toString` here and an `Object.prototype.toString` below?

lib/runnable.js
```diff
@@ -277,4 +275,18 @@ Runnable.prototype.run = function(fn){
...
+      if (null != err) {
+        if (Object.prototype.toString.call(err) === '[object Object]') {
```
boneskull: this can be a ternary if you wish

support/tail.js
```diff
@@ -53,7 +53,7 @@ process.on = function(e, fn){
   if ('uncaughtException' == e) {
     global.onerror = function(err, url, line){
       fn(new Error(err + ' (' + url + ':' + line + ')'));
-      return true;
+      return !mocha.allowUncaught;
```
boneskull: isn't this in reference to the global you removed?

amsul: Nope - this is now referring to the `mocha` exposed below on [line 67](https://github.com/mochajs/mocha/pull/1662/files#diff-393a1102677dadf535f6062dbca02af9R67)

lib/runnable.js
amsul: I didn't write this.. the lines were just moved into the method named `callFnAsync`.

I can make the changes if you like.

lib/runnable.js
amsul: Same here.. the lines were just moved. But I can make the changes if you like.

lib/runnable.js
boneskull: don't worry about it

boneskull:
@amsul Everything looks good here, thanks.  Can you open another pr against the `gh-pages` branch to note the available option in the browser?  There should be a (small) section for browser-only options.

dasilvacontin:
Nice, thanks guys! :)

amsul:
@boneskull, @dasilvacontin added #1740 :)

lolmaus:
How do I use it in browser? I tried `mocha.allowUncaught()` in the console but it doesn't change anything.

PR #1666, opened 2015-04-21
Add {request,cancel}AnimationFrame to ignored globals

redforks:
Like setImmediate/clearImmediate,
requestAnimationFrame/cancelAnimationFrame is another time like global
javascript function. Some tests need to mock them out.

### DIFF
--- a/mocha.js
+++ b/mocha.js
@@ -4607,7 +4607,9 @@ var globals = [
   'setImmediate',
   'clearImmediate',
   'requestAnimationFrame',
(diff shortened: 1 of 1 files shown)


boneskull:
couple things:

- `mocha.js` is generated; do not modify it directly
- please provide a use case

jbnicolai:
Been over a month. Closing as I'm not sure this is solving an actual issue.

@redforks you're more than welcome to drop a comment to explain the need for this change, and I'lll happily reconsider

PR #1678, opened 2015-04-30
fix Runner#fail logic to seamlessly support Error-like objects. Closes #1677

### LINKED ISSUE #1677: Runner#fail should work with Error-like objects
**Problem**
The check `if (!(err instanceof Error))` at https://github.com/mochajs/mocha/blob/master/lib/runner.js#L205 causes a problem when writing tests for an Angular app using the `angular-mocks` module. All failed assertions and/or uncaught errors from within an `inject`-ed function are displayed as
```
the object {...} was thrown, throw an Error :)
```
JSFiddle: http://jsfiddle.net/tybx9zwp/

**Reason**
The `angular.mock.inject` function catches errors, wraps them in an `Error`-like object (which does not inherit from `Error`), transforms the `stack` property, and re-throws the wrapper object. See https://github.com/angular/angular.js/blob/master/src/ngMock/angular-mocks.js#L2416. The commit which introduced this mechanism is angular/angular.js@7e916455b36dc9ca4d4afc1e44cade90006d00e3. The given justification is
```
Recent browsers, particularly PhantomJS 1.9.2 and Safari 7.0
treat the stack property as non-configurable and unwritable.
```
which seems reasonable.

**Solution**
This can of course be fixed quickly by replacing the `instanceof` check in `Runner#fail` with something like
```js
if (!err || typeof err.stack !== 'string') {
  err = new Error('the ' + type(err) + ' ' + stringify(err) + ' was thrown, throw something Error-like :)');
}
```
but it is probably also a good idea to follow Angular's example and wrap `err` in an `Error`-like object before modifying its `stack` on https://github.com/mochajs/mocha/blob/master/lib/runner.js#L209.

I would be happy to submit a PR to address this, please let me know if the above approach is acceptable.

igorraush:


lib/runner.js
```diff
@@ -202,10 +203,13 @@ Runner.prototype.fail = function(test, err) {
   ++this.failures;
   test.state = 'failed';
 
-  if (!(err instanceof Error)) {
+  var isError = err instanceof Error;
+  var isErrorLike = err && isString(err.message) && isString(err.name);
```
boneskull: this will throw an exception if err happens to be `null`.  add additional checks.  

I would also probably avoid creating the variables if we are not going to use them elsewhere.

lib/runner.js
```diff
@@ -714,3 +718,19 @@ function extraGlobals() {
 
  return [];
 }
+
```
boneskull: @igorraush Can you point me to the "error-like object" that AngularJS creates?  I'd like to see how they are doing it.

I'm not sure why a constructor is necessary here.  Something similar to:

```js
err = Object.assign({}, err);
```

Ought to be sufficient (though I'm not sure what kind of flimsy `Object.assign()` shim we have)

igorraush: @boneskull Their constructor is defined here
https://github.com/angular/angular.js/blob/master/src/ngMock/angular-mocks.js#L2371
and used here
https://github.com/angular/angular.js/blob/master/src/ngMock/angular-mocks.js#L2417
It serves a dual purpose for them: to extend the error stack of the original error, and to reuse `toString`.

lib/runner.js
igorraush: @boneskull Sorry, I fail to see how it will throw an exception when `err` is `null` :( This test passes for me:
```js
it('should emit a helpful message when failed with null', function(done){
  var test = {}, err = null;
  runner.on('fail', function(test,err){
    err.message.should.equal('the null [null] was thrown, throw an Error :)');
    done();
  });
  runner.fail(test, err);
})
```
The message could definitely be more helpful, but no crash.

boneskull: sorry, n/m, I think I was reading the example in the original issue instead

lib/runner.js
boneskull: ty

PR #1698, opened 2015-05-13
Escape test/suite title for re in html reporter

### LINKED ISSUE #1687: html reporter link to specific test is broken on some special characters
If the test title contains some special characters (for instance $ or () ) then its link in the HTML report does not work.

IMO the fix should be escaping regex [here](https://github.com/mochajs/mocha/blob/908d8dbe15b5ef5f4a5658f6cc870bf563e80717/lib/reporters/html.js#L192) before calling encodeURIComponent(s)

benvinegar:
Fixes #1687

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -5,6 +5,7 @@
 var Base = require('./base')
   , utils = require('../utils')
   , Progress = require('../browser/progress')
+  , escapeRe = require('../browser/escape-string-regexp')
   , escape = utils.escape;
 
 /**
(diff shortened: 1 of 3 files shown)


dasilvacontin:
Hi @benvinegar! Could you provide tests that assert this change fixes the issue? Thanks!

benvinegar:
@dasilvacontin Where do you recommend adding that? I could introduce a new /reporter test for the Html reporter (there isn't one yet), and verify that `makeUrl` escapes the `grep` component.

benvinegar:
Or I could add a new test to test/browser/, similar to test/browser/grep.html, which clicks through generated grep links to verify they work (instead of manipulating `window.location` directly).

jbnicolai:
@benvinegar sorry for the silence on our side. Great fix! I'd recommend the later (simply because it seems easier), but both would prove the fix so pick whichever you'd prefer.

benvinegar:
Okay. I've added a test that demonstrates that clicking through links w/ regex chars works correctly.

But, I'm not sure how to run this test as part of CI. I'm not sure if `test/browser/grep.*` runs as part of CI either.

Please advise.

boneskull:
> But, I'm not sure how to run this test as part of CI. I'm not sure if test/browser/grep.* runs as part of CI either.

`test/browser` is omitted from CI.  I don't have much experience with them, but looking over them briefly I don't think they have much value in the first place--especially since the only way you can assert they work is to eyeball them.

boneskull:
@benvinegar thanks again!

PR #1578, opened 2015-03-06
Ignore asynchronous failures after spec resolution

danielstjules:
This improves on https://github.com/mochajs/mocha/pull/1540 From that thread:

Given the following test, which only has one spec:

``` javascript
it('fails exactly once when a global error is thrown first', function(done) {
  setTimeout(function() {
    throw new Error('global error');

    setTimeout(function() {
      done(new Error('test error'));
    }, 0);
  }, 0);
});
```

You get two two failures:

``` javascript
$ ./bin/mocha pr1540.js

  ․․

  0 passing (6ms)
  2 failing

  1)  fails exactly once:
     Error: test error
      at null._onTimeout (/Users/danielstjules/git/mocha/pr1540.js:10:12)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
      at timers.js:117:18
      at process._tickCallback (node.js:415:13)
      at process._tickFromSpinner (node.js:390:15)

  2)  fails exactly once:
     Error: test error
      at null._onTimeout (/Users/danielstjules/git/mocha/pr1540.js:10:12)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
      at timers.js:117:18
      at process._tickCallback (node.js:415:13)
      at process._tickFromSpinner (node.js:390:15)
```

But after this PR, it correctly only shows one failure, and highlights the error thrown first ("global error"):

``` javascript
$ ./bin/mocha pr1540.js

  ․

  0 passing (6ms)
  1 failing

  1)  fails exactly once:
     Uncaught Error: global error
      at null._onTimeout (/Users/danielstjules/git/mocha/pr1540.js:12:11)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
```

In addition to the above fix, this PR improves on #1540 by also correctly handling the following test case:

``` javascript
it('fails exactly once when a global error is thrown second', function(done) {
  setTimeout(function() {
    done(new Error('test error'));
  }, 0);

  setTimeout(function() {
    throw new Error('global error');
  }, 0);
});
```

It now also only reports a single failing spec, rather than two.

PR #1540, opened 2015-02-08
Ignore asynchronous errors after global failure

jugglinmike:
`Runner#uncaught` labels the currently-executing `Runnable` instance as
"failed" in response to uncaught exceptions. If the the `Runnable`
instance later signals failure through Mocha's asynchronous interface,
the `Runner` counts this as a distinct test failure. Consequently,
reporters generate inaccurate reports (and in some cases, this triggers
uncaught exceptions causing the process itself to crash).

When a runnabe completes, explicitly check that it has not already been
labeled as failing before reporting any additional errors.

PR #1701, opened 2015-05-15
Updates mocha.js to address #1700

### LINKED ISSUE #1700: "PhantomJS timed out, possibly due to a missing Mocha run() call" regression
I'm getting a
```
Warning: PhantomJS timed out, possibly due to a missing Mocha run() call. Use --force to continue.
```
after upgrading mocha from 2.1.0 to 2.2.0 (or any of the 2.2.x series, including the most recent 2.2.5 release).

I've managed to track it to 521b7e443e3a0feed3091c0378973c4faab42160, specifically, lines 4485-4488:
```js
    // Discard the resolution if this test has already failed asynchronously
    if (self.state) return;
```
Commenting out `if (self.state) return;` eliminates the error (I've tested against mocha 2.2.0 and 2.2.5).

# Steps to reproduce:
Disclaimer: I never used mocha, grunt, yeoman, bower, or npm prior to two weeks ago, so I fully accept that the problem lies solely with how I'm testing rather than a problem with mocha itself. That said, I noticed this problem using Yeoman's generator-webapp after updating mocha in bower.json from 1.14.0 to 2.2.4. Dropping to v2.1.0 fixed the issue.
I've pushed a stripped-down, minimal version of an "app" that reproduces the problem: https://github.com/eddies/mocha-test-case

```
git clone git@github.com:eddies/mocha-test-case.git
cd mocha-test-case
npm install && bower install
grunt test
```
Comment out line 4512 of `bower_components/mocha/mocha.js` and re-run `grunt test` for the test to now pass.

eddies:
Fixes #1700

### DIFF
--- a/mocha.js
+++ b/mocha.js
@@ -4508,9 +4508,6 @@ Runnable.prototype.run = function(fn){
     if (self.timedOut) return;
     if (finished) return multiple(err || self._trace);
 
-    // Discard the resolution if this test has already failed asynchronously
-    if (self.state) return;
-
     self.clearTimeout();
     self.duration = new Date - start;
     finished = true;


dasilvacontin:
Hi @eddies! Could you provide tests that pass thanks to this change? Cheers!

eddies:
@dasilvacontin Honestly, I'm not sure how to provide a unit-like test for this. I was only triggering this issue as part of running a test w/ PhantomJS and Grunt and and various other dependencies. I did create https://github.com/eddies/mocha-test-case as a minimal example, which I believe definitively isolates the error to the specific one-line commit to mocha I highlighted.

If someone could take a look at the mocha-test-case repo and provide some suggestions as to how that might be translated into a test for the mochajs repo, I could try running with it from there.

danielstjules:
@eddies Please make the change against runnable.js, rather than the compiled/disted mocha.js. https://github.com/mochajs/mocha/blob/49d81aa17751e47cec26a22f0607076c98c4a1fc/lib/runnable.js#L215

Beyond that, how does this impact https://github.com/mochajs/mocha/pull/1578 and https://github.com/mochajs/mocha/pull/1540 ? After updating runnable.js, I wonder if the build fails.

jbnicolai:
Picked up @danielstjules's comment and merged in a separate branch.

PR #1702, opened 2015-05-17
add more friendly output for dot reporter in colorless mode

pcdevil:
Currently when using dot reporter with `no-colors` option the _dot-matrix_ output has no use.

To resolve that I want to introduce an extra parameter to the base reporter's `color` function which is a fallback string if colors are disabled. If this parameter is given the function will return that instead of the original string.

Using this parameter the dot reporter can easily signal the difference between tests and it will output the following:
 - the default dot if the test passes
 - `P` if the test is pending
 - `S` if the test is slow
 - `F` on failure

### DIFF
--- a/lib/reporters/dot.js
+++ b/lib/reporters/dot.js
@@ -32,13 +32,13 @@ function Dot(runner) {
 
   runner.on('pending', function(test){
     if (++n % width == 0) process.stdout.write('\n  ');
-    process.stdout.write(color('pending', Base.symbols.dot));
+    process.stdout.write(color('pending', Base.symbols.dot, 'P'));
   });
 
   runner.on('pass', function(test){
     if (++n % width == 0) process.stdout.write('\n  ');
     if ('slow' == test.speed) {
-      process.stdout.write(color('bright yellow', Base.symbols.dot));
+      process.stdout.write(color('bright yellow', Base.symbols.dot, 'S'));
     } else {
       process.stdout.write(color(test.speed, Base.symbols.dot));
     }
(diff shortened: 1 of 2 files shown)


jbnicolai:
I like the suggestion, but there's no way we can merge this without breaking backwards compatibility. 

On the other hand, in what bizarre use-case would you use the output of the dot reporter in a script?

@boneskull what do you think? I'd say merge ;-)

pcdevil:
I'm not sure why you are saying it breaks compatibility because now mocha with `no-colors` only prints plain dots to passed and failed tests too and you can't distinguish them. With this modification you can ;)   
But I'm not gonna argue about it, if you want I can duplicate the dot reporter (with _dot-detailed_ or something similar name) and implement this feature in that one - in this way compatibility will be intact and everything will be shiny :) 

The notion behind this is get a minimal report but get it meaningfully. Like if you running the test with a build script and redirect the output into a file you have a nice little output of it.

lib/reporters/dot.js
```diff
@@ -32,13 +32,13 @@ function Dot(runner) {
...
+    process.stdout.write(color('pending', Base.symbols.dot, 'P'));
   });
 
   runner.on('pass', function(test){
     if (++n % width == 0) process.stdout.write('\n  ');
     if ('slow' == test.speed) {
-      process.stdout.write(color('bright yellow', Base.symbols.dot));
+      process.stdout.write(color('bright yellow', Base.symbols.dot, 'S'));
     } else {
       process.stdout.write(color(test.speed, Base.symbols.dot));
```
boneskull: @pcdevil this one too?

boneskull:
@pcdevil This is a breaking change because the output of the dot reporter now *differs* if no TTY is present (for example).  Granted, I don't *expect* anyone is trying to use a machine to parse the output of the `dot` reporter, but it's just not safe to put this feature into anything other than a major release.  If it were any other project, I'd say "go for it", but previously we've altered the *output* (not the "API") of Mocha in patch or minor releases and it sparked a shitstorm, so no can do.  Please rebase against branch `v3.0.0`.

hollomancer:
Handled in #2000.

PR #1707, opened 2015-05-21
fix emit "pending" when running pending test

### LINKED ISSUE #1706: missing `pending` event
`Runner` doesn't throw a `pending` event when a test is marked as `pending` after firing its `test` event.

`Runner` checks for `.pending` at https://github.com/mochajs/mocha/blob/master/lib/runner.js#L465. If the test is marked as `pending` after that, it will get detected at https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L256-L257 (afaik it's the only way to get to that code branch).

However, it is executing `done` without arguments. `Runner` is expecting a `Pending` error for tests ran that are either pending or skipped. https://github.com/mochajs/mocha/blob/master/lib/runner.js#L477

dasilvacontin:
Closes #1706.

`Runnable.run(fn)` wasn't passing an instance of `Pending` to the
callback when a `pending` test was ran. Therefore, `Runner` was not
firing a "pending" event.

Added a test which currently fails due to the `skippedTest`:
`[TypeError: Object #<Object> has no method 'skip']`. How does
one skip the current running test synchronously?

Also: why are we throwing a `Pending` instance? Couldn't we set `pending = true` in `.skip()`, and then check `test.pending` again at https://github.com/mochajs/mocha/blob/master/lib/runner.js#L475 ?

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -196,6 +196,10 @@ Runnable.prototype.run = function(fn){
     , finished
     , emitted;
 
+  // Runner expects to catch an instance of Pending if a pending test is ran
+  if (this.pending)
+    throw new Pending();
+
   // Some times the ctx exists but it is not runnable
   if (ctx && ctx.runnable) ctx.runnable(this);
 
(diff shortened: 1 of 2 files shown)


dasilvacontin:
I forgot there was work pending on this one. Adding to my to-do list.

hollomancer:
This is a valid issue - however, this PR has been inactive for a very long time. Let's revisit if we see a need.

PR #1719, opened 2015-06-02
Update base.js

askhogan:
Remove unneeded console.log() - creating lines with undefined output


jbnicolai:
@askhogan these log statements are to create an empty line before the series of log lines created by the `foreach` below. It seems these have always been there, so I'd be hesitant to remove them without need.

Feel free to update the PR with a motivation though :)

PR #1723, opened 2015-06-04
Fix missing variable 'done'

methuz:
Fix running `mocha test.js` fail because variable `done` is not defined


dasilvacontin:
@danielstjules, can you confirm you accidentally pushed this file?

jbnicolai:
Yeah, that file must've been committed by accident, see #1724.

Issue #1457, opened 2014-11-30
🚀 Feature: Plugin API


boneskull:
We need plugin API(s). 

More to follow.

glenjamin:
Some thoughts on the above proposal:

* While possibly cool, auto-detection of plugins isn't very node-like. Using `--plugin <X>` and requiring the user to list them in `mocha.opts` will be far easier to implement and work basically the same way.
* Perhaps combine with a `--disable-plugin <X>` to allow overriding defaults on the CLI
* I think the module signature should just be a single export function, which receives the mocha plugin API as an argument. This is pretty straightforward and provides a simple way to expose additional functionality.
```js
// my-cool-plugin.js
module.exports = function(mocha) {
  mocha.registerReporter();
  mocha.addCliOption();
  // etc
}
```

I would suggest that core-module-replacement and inter-plugin-communication not be included in an initial release, as they're quite tricky and it would be more important to get the basic mechanisms in first.

A possible initial featureset could be:
* Custom reporters
* Custom interfaces
* Access to the current instance options
* Access to the `Runner` instance (can attach events to do stuff)
* Add extra functions to existing interfaces?

One way to allow plugins to specify extra options but also be parsed from options is to do one pass over argv to get the plugin list, then load them, then do another pass to get the full options list.

boneskull:
@glenjamin Thanks for the input.  I'm going to fiddle with this a bit.

jamestalmage:
First, 
This is a **great** idea.

I've always like chai's plugin architecture:
```javascript
chai.use(require('some-chai-plugin'));
```
That should not preclude functions like `mocha.registerReporter()`, etc. But rather those functions should be called by the plugin author, not the end user. This allows plugin authors to hide ugly implementation details from the user (i.e. my `super-awesome-reporter` may need to install a global `beforeEach`, `afterEach`, etc. You don't want to know that. You just want to to use`super-awesome-reporter`).

So spec should be something like this:
 1. user calls `mocha.use(require('plugin-module'))`

     or optionally `mocha.use(require('plugin-module')(opts))`. 
 2. convention dictates that a plugin module returns a callback.
 3. mocha will validate that callback has never been called before. If it has, silently ignore (mocha should ensure plugins are not added twice - should not be an error).
 4. mocha will call that callback, passing a `registration object` as the first parameter.
 5. registration object will expose all the allowed "hooks"

boneskull:
[...]

Issue #1193, opened 2014-04-17
rename browser require() implementation


jonathanong:
to something like `__require__` so stuff like https://github.com/visionmedia/node-requires and browserify don't try to parse it. wuttaya think

Issue #880, opened 2013-05-30
browserify breaks at require('./lib-cov/mocha')


darvin:
`mocha/index.js`:

```
module.exports = process.env.COV
  ? require('./lib-cov/mocha')
  : require('./lib/mocha');
```

```
Error: module "./lib-cov/mocha" not found from "/Volumes/work/Projects/yaas/node_modules/mocha/index.js"
```

ksikka:
I'm still unable to browserify a file that require's mocha. Error:
> Error: Cannot find module './lib-cov/mocha' from '/Users/ksikka/test_mochify/node_modules/mocha'

Anyone else in the same boat?

Tried it with node 0.10, 0.12, the latest browserify in npm (11.2.0), the latest mocha in npm (2.3.3).

markmenox:
Same here with node 4
even if i have:
"browser": {
    "debug": "./lib/browser/debug.js",
    "events": "./lib/browser/events.js",
    "tty": "./lib/browser/tty.js",
    "mocha": "./mocha.js"
  },

in package.json of mocha module

ksikka:
@tj reopen?

markmenox:
Yes it should be defenitively. Right now it is broken with browserify and the only solution i have for the moment is to keep mocha library (with all my js files that use it), out of browserify process

boneskull:
I'm reopening this, but tentatively.  

Maybe [mochify](https://npmjs.com/package/mochify) will help you guys.

boneskull:
don't need whatever ancient coverage stuff is baked in.  we should use something else.

SonofNun15:
Webpack has the same issue, by the way. Why is mocha conditionally requiring a file that does not exist? Is this file generated under some conditions? What triggers this file to be generated?

kuraga:
Yes, why?

kuraga:
Should be `"browser": "./support/browser-entry"`, no?

Sigfried:
Same issue here with webpack.

tswaters:
> Why is mocha conditionally requiring a file that does not exist? Is this file generated under some conditions? What triggers this file to be generated?

@SonofNun15 It's all in the makefile.  when running `test-cov` it uses the jscoverage tool to instrument the code into `lib-cov` - then mocha runs with the COV environment variable set to report out coverage for the instrumented code.... The initial discussion is in #5 and d4666c2 is the initial commit.

[...]

PR #1727, opened 2015-06-06
Build client-side Mocha using Browserify

ndhoule:
paging @boneskull, @travisjeffery 

Still needs some work before merging but wanted to put it out there early since it seems like a better build process is eating up time and making stuff like the [plugin API](https://github.com/mochajs/mocha/issues/1457) and [breaking reporters out into npm-land](https://github.com/mochajs/mocha/issues/509) difficult.

Outstanding issues:

- Still some cleanup to do, current transition is a little messy. And I bet we can remove a few more shims like `lib/browser/events`.
- Did super cursory testing in the browser and even more cursory at the command line; would want to stress test this a bit more before merging.
- The built Browserify file is still a lot larger than the current build process (~148k currently, 209k under Browserify). Haven't looked into this too deeply yet.
- Browserified file includes all the reporters, which [@tj said](https://github.com/mochajs/mocha/issues/509#issuecomment-48804936) shouldn't happen anyway. I think, though, that removing them would break [mocha-phantomjs](https://github.com/metaskills/mocha-phantomjs); @metaskills, is that right? Is that something to be concerned about?

Will follow up on these in the next day or so, when I get a bit more time.

----------

Related issues:
- https://github.com/mochajs/mocha/issues/1193
- https://github.com/mochajs/mocha/pull/1260
- https://github.com/mochajs/mocha/pull/1559
- https://github.com/mochajs/mocha/issues/880 (this PR skips Jade in Browserify via `--ignore jade`, which I think was causing @jonathanong's issue back in #1260)

Tangentially related:
- https://github.com/mochajs/mocha/issues/1457


lib/mocha.js
```diff
@@ -143,7 +144,9 @@ Mocha.prototype.reporter = function(reporter, reporterOptions){
   } else {
     reporter = reporter || 'spec';
     var _reporter;
-    try { _reporter = require('./reporters/' + reporter); } catch (err) {}
+    // Try to load a built-in reporter.
+    if (reporters[reporter]) _reporter = reporters[reporter];
```
ndhoule: Worth pointing out here is that Browserify [doesn't do dynamic requires](https://github.com/substack/node-browserify/issues/377), so we have to switch to a static approach here. I assumed that `reporter` is always in lowercase, not sure if that's a safe assumption or if we want to implement a more robust approach to grabbing built-in reporters here (spitballing here: e.g. always normalize to `kebab-case`)?

lib/mocha.js
```diff
@@ -143,7 +144,9 @@ Mocha.prototype.reporter = function(reporter, reporterOptions){
...
+    // Try to load reporters from process.cwd() and node_modules
```
ndhoule: So this becomes a Node-only feature. Any workflows that'll break?

lib/reporters/index.js
```diff
@@ -1,17 +1,19 @@
...
-exports.Nyan = require('./nyan');
-exports.XUnit = require('./xunit');
-exports.Markdown = require('./markdown');
-exports.Progress = require('./progress');
-exports.Landing = require('./landing');
-exports.JSONCov = require('./json-cov');
-exports.HTMLCov = require('./html-cov');
-exports.JSONStream = require('./json-stream');
+// Alias exports to a their normalized format Mocha#reporter to prevent a need
+// for dynamic (try/catch) requires, which Browserify doesn't handle.
```
ndhoule: Mildly gross/hacky fix for the no-dynamic-requires problem noted above.

package.json
```diff
@@ -66,10 +68,15 @@
     "mocha.js",
     "LICENSE"
   ],
+  "browser": {
+    "debug": "./lib/browser/debug.js",
```
ndhoule: These (along with) replace the rewrites we were doing in [`support/compile.js`](https://github.com/mochajs/mocha/pull/1727/files#diff-4519ae56dd3583387155ed89f8a99ba1L45)


For more details see https://github.com/substack/node-browserify#browser-field

jbnicolai:
Note that the failing TravisCI build is only on node `v0.8` due to the caret-notation used on the browserify dependency, something older npm versions can't handle.

Makefile
```diff
@@ -1,4 +1,5 @@
 
+BROWSERIFY := node_modules/.bin/browserify
```
jbnicolai: I guess we could do something like `PATH := node_modules/.bin:$(PATH)` to simplify the code (slightly). Seeing as you'd also prefer a local dep. over a global one anyway.

lib/mocha.js
boneskull: what's the node-only feature, exactly?

lib/mocha.js
boneskull: dunno.  I wanted to eliminate them from the core except for `spec`, so whatever's clever for now.

lib/mocha.js
jbnicolai: Don't think so.

Makefile
boneskull: :-1:  Makefiles

boneskull:
let's just keep the browserify dependency exact.  given that you can't build the browser version of Mocha w/o it, I don't want to take any chances

lib/reporters/base.js
```diff
@@ -34,9 +34,9 @@ exports = module.exports = Base;
  * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors = process.env
-  ? (supportsColor || (process.env.MOCHA_COLORS !== undefined))
-  : false;
+exports.useColors = process.browser
+  ? false
+  : (supportsColor || (process.env.MOCHA_COLORS !== undefined));
```
jbnicolai: A ternary where one of the statements is a boolean constant always seems odd to me. How about

```javascript
!process.browser && (supportsColor || process.env.MOCHA_COLORS);
```

boneskull:
> Browserified file includes all the reporters, which @tj said shouldn't happen anyway. I think, though, that removing them would break mocha-phantomjs; @metaskills, is that right? Is that something to be concerned about?

there's only one reporter in the browser...

boneskull:
> The built Browserify file is still a lot larger than the current build process (~148k currently, 209k under Browserify). Haven't looked into this too deeply yet.

I don't care about this *at all*

boneskull:
> Still some cleanup to do, current transition is a little messy. And I bet we can remove a few more shims like lib/browser/events

yes please.  thanks for the tremendous amount of work here.

package.json
jbnicolai: Thanks for the comment!

lib/reporters/base.js
boneskull: I'm in agreement here fwiw

support/browser-entry.js
```diff
@@ -158,3 +145,10 @@ mocha.run = function(fn){
  */
 
 Mocha.process = process;
+
+/**
+ * Expose mocha.
+ */
+
+window.Mocha = Mocha;
+window.mocha = mocha;
```
jbnicolai: Accidental duplicate line?

lib/mocha.js
jbnicolai: Couldn't agree more with @boneskull.

ndhoule:
@boneskull 

> there's only one reporter in the browser...

https://github.com/metaskills/mocha-phantomjs#standard-out
I think mocha-phantomjs relies on all the reporters being present in the browser (which is why [`Mocha.process` is exposed](https://github.com/mochajs/mocha/pull/916)).

Makefile
jbnicolai: @boneskull why? To me this still seems more straightforward than most alternatives (which admittedly says a lot about the alternatives as well..)

lib/reporters/base.js
ndhoule: Ah yeah, I just kept it that way because `process.browser` is `undefined` in Node, but I'll stop being lazy

jbnicolai:
> thanks for the tremendous amount of work here

yes!

> I don't care about this at all

Same. Make it work first, optimize later. ;)

>  I think, though, that removing them would break mocha-phantomjs; @metaskills, is that right? Is that something to be concerned about?

I wouldn't worry about this. The other reporters should be broken-out and included separately and explicitly by `mocha-phantomjs`

ndhoule:
Bleh, looks like Browserify's deps have carets in them so they hose Node 0.8. Could just add a Make target to install it before building `mocha.js` I guess, and note in the docs that Mocha works, but is unbuildable on, 0.8.

jbnicolai:
Yes.. the issue is not even node 0.8 but the bundled npm version. Any production environment stuck due to whatever complexities on an older version would still easily be able to update npm, so I wouldn't consider it an issue at all.

We could add a line to the .travis.yml like:

```yml
before_install:
- npm install -g npm@latest
```

Makefile
ndhoule: Yeah, I prefer local dependencies. Gives you everything you need in an `npm install` and prevents "what version of Browserify are you running, oh, update/downgrade it"-type issues.

I generally just throw all my binaries at the top of the file because I've has issues with the `PATH` solution in the past (can't remember what those were), I'm open to whatever though

lib/mocha.js
ndhoule: My thought was that Browserify's static requires would prevent hot loads here in the case that someone would want to load a custom reporter in the browser, but now that I'm thinking about it that might not be a problem? Not sure, haven't used Browserify for this specific scenario.

Mostly grasping at straws trying to think of a breaking use case because I haven't ever loaded a custom reporter in the browser, but that doesn't necessarily mean other people haven't.

lib/mocha.js
ndhoule: :+1: works for me.

jbnicolai:
> > The built Browserify file is still a lot larger than the current build process (~148k currently, 209k under Browserify). Haven't looked into this too deeply yet.

> I don't care about this at all

If we'd like we could get this down by about 3x using Google's closure compiler.

```bash
make
for f in mocha mocha.browserified; do 
  curl --data-urlencode "js_code@$f.js" \ 
    -d "output_format=text&output_info=compiled_code&compilation_level=SIMPLE_OPTIMIZATIONS&language=ECMASCRIPT5" \ 
    http://closure-compiler.appspot.com/compile > "$f.min.js"; 
done
du -hs | gshort -h
```
>  88K    mocha.browserified.min.js
 88K    mocha.min.js
208K    mocha.browserified.js
208K    mocha.js

boneskull:
> I wouldn't worry about this. The other reporters should be broken-out and included separately and explicitly by mocha-phantomjs

I don't think `karma-mocha` is reliant on the reporters either.  It has its own.  I'm not sure where it hooks in, though.

boneskull:
> I don't think karma-mocha is reliant on the reporters either. It has its own. I'm not sure where it hooks in, though.

(you can use phantomjs as the browser)

jbnicolai:
Played with it a bit more, I'm all for this PR :+1:

Makefile
jbnicolai: Nah, the number of paths explicitly stated is small enough that it doesn't matter, although I'm curious what your issues were.

ndhoule:
@jbnicolai 

> Yes.. the issue is not even node 0.8 but the bundled npm version. Any production environment stuck due to whatever complexities on an older version would still easily be able to update npm, so I wouldn't consider it an issue at all.
>
> We could add a line to the .travis.yml like:

Haha good point, not sure why I didn't think about that. Changed

lib/mocha.js
jbnicolai: Good point - I have no experience with this either.

package.json
ndhoule: Whoops, meant to say "along with the `--ignores` we pass to `browserify`

support/rewriteInheritance.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Replace all instances of `__proto__` in a string with a cross-browser
+ * compatible alternative.
+ *
+ * @param {string} js
+ * @return {string}
+ */
+function parseInheritance(js) {
+  return js.replace(/^ *(\w+)\.prototype\.__proto__ * = *(\w+)\.prototype *;?/gm, function(_, child, parent){
```
jbnicolai: :/ I appreciate the effort, but we really should remove all usages of `__proto_` from the codebase. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto for instance.

Couldn't we apply this rewrite rule verbatim to our codebase, rather than at build-time?

```bash
git grep __proto lib
```
> lib/hook.js:Hook.prototype.__proto__ = Runnable.prototype;
lib/reporters/dot.js:Dot.prototype.__proto__ = Base.prototype;
lib/reporters/landing.js:Landing.prototype.__proto__ = Base.prototype;
lib/reporters/list.js:List.prototype.__proto__ = Base.prototype;
lib/reporters/min.js:Min.prototype.__proto__ = Base.prototype;
lib/reporters/nyan.js:NyanCat.prototype.__proto__ = Base.prototype;
lib/reporters/progress.js:Progress.prototype.__proto__ = Base.prototype;
lib/reporters/spec.js:Spec.prototype.__proto__ = Base.prototype;
lib/reporters/xunit.js:XUnit.prototype.__proto__ = Base.prototype;
lib/runnable.js:Runnable.prototype.__proto__ = EventEmitter.prototype;
lib/runner.js:Runner.prototype.__proto__ = EventEmitter.prototype;
lib/suite.js:Suite.prototype.__proto__ = EventEmitter.prototype;
lib/test.js:Test.prototype.__proto__ = Runnable.prototype;

It looks like the transform below would change
```javascript
Hook.prototype.__proto__ = Runnable.prototype;
```
into
```javascript
function F(){};
F.prototype = Runnable.prototype;
Hook.prototype = new F;
Hook.prototype.constructor = Hook;
```

Is there any reason to not just apply this to the entire codebase and remove this rewrite rule?

Makefile
ndhoule: No clue, I can repro it in my other local repos, but not in my local Mocha repo. Weird.

I also prefer the explicit path just because I keep global installations of some binaries (e.g. ESLint) installed for editor integrations, etc. and like to be sure there's nothing funky going on, like a missing local dependency triggering errors because it's falling back on a global installation, which is a heinous thing to debug. More verbose for fewer terrible things to debug, but maybe that's crossed over into paranoia territory.

Anyway, it's working locally so I don't mind either way

support/rewriteInheritance.js
ndhoule: Haha, no real effort involved, I just ported it from here:

https://github.com/mochajs/mocha/blob/master/support/compile.js#L59

But yeah, we should definitely fix it throughout. I just figured we could handle that as a separate issue and keep this PR's scope strictly to switching to Browserify, since it already runs the risk of breaking so many things

jbnicolai: Completely forgot about that, thought you dug it up somewhere else ;) (heh, my name is even in the edit-list of said file.. oops).

Let's leave this for now :) although a future improvement would definitely be getting rid of these.

ndhoule: https://github.com/mochajs/mocha/issues/1728

ndhoule:
Is there an established/easy way to do cross-browser testing against the built lib right now? I can just do some informal Sauce tests if not

jbnicolai:
@ndhoule there's not. It might be about time to add some selenium based tests.

support/rewriteInheritance.js
boneskull: errr... yeah.  we need to get rid of all of these hinky shims and use something well-supported.  if we're targeting environments that don't support `Object.create()` then we should just do something like use [lodash.create](https://www.npmjs.com/package/lodash.create)

boneskull:
@jbnicolai go for it dude, you're on a roll :wink:

ndhoule:
Had a chance to do some IE < 9 browser testing; glad I did, exposed a few breakages. Fixed and rebased. I think this is about ready to merge.

Adding CI would be a good thing because it's easy to introduce an ES3-incompatible shim with Browserify. Started an issue over here: https://github.com/mochajs/mocha/issues/1732

Going to punt on the reporter removal (https://github.com/mochajs/mocha/issues/1731, feel free to close as a dupe/consolidate issues/etc.) just because I'm concerned it'll break mocha-phantomjs, thinking we should give them some heads up / provide a solution before we remove it. Make sense?

nathanboktae:
Yes removing the reporters will break mocha-phantomjs and [mocha-casperjs](https://github.com/nathanboktae/mocha-casperjs) - both use the compiled `mocha.js` and change the reporter as needed. I'll provide more details in #1731

rstacruz:
how about this to use the environment's default `require()` function:

```diff
-     if (!_reporter) try { _reporter = require(reporter); } catch (err) {
+     if (!_reporter) try { _reporter = global.require(reporter); } catch (err) {
        err.message.indexOf('Cannot find module') !== -1
          ? console.warn('"' + reporter + '" reporter not found')
```

ndhoule:
@rstacruz Unless I'm missing something here—Browserify doesn't globalize its `require` mechanism, and you'd need to somehow inject it into its built cache so that [Mocha knows how to `require` it](https://github.com/mochajs/mocha/blob/master/lib/reporters/index.js).

ndhoule:
@boneskull @jbnicolai Light poke—anything outstanding on this PR that still needs to be addressed?

jbnicolai:
@ndhoule, @boneskull apologies, but I'm on the other side of the world this week for work, so the moments I've got both time and reliable Wi-Fi are rare. I'm still all for this PR, but think we should release it in at least a minor version bump due to the issues @nathanboktae mentions.

@boneskull do you have time to integrate this? Otherwise I'll get to it as soon as possible, but can't guarantee that that'll be before the start of next week

rstacruz:
@ndhoule — yes, that was kind of the point — so to be backwards-compatible with those that rely on node.js's require() mechanism (like mocha-phantomjs).

ndhoule:
@rstacruz Can you describe a currently-working use case that would be broken unless we change that line to `global.require`?

mocha-phantomjs [runs third-party reporters in the browser](https://github.com/metaskills/mocha-phantomjs#third-party-reporters) by assigning them to [`Mocha.reporters.Custom`](https://github.com/metaskills/mocha-phantomjs/blob/master/lib/mocha-phantomjs.coffee#L120), so as far as I can tell, that should make no difference. Would be curious to see otherwise.

> something like require(reporter) will not work with browserify anyway (it will expect a string literal passed onto require).

Not if whatever `reporter` evaluates to is already in Browserify's dependency graph. Not really relevant in this case anyway; doesn't look like Browserify's `require` has a registration API that's available at runtime.

ndhoule:
@jbnicolai @boneskull nbd, risk of merge conflicts is pretty low.

Messed around with mocha-phantomjs and realized that Browserify's `process` shim excludes `process.stdout`. Amended to patch that; all of mocha-phantomjs's tests are now passing, including the ones involving custom runners. Also tested it out briefly in IE 6/7/8 and nothing went boom.

ndhoule:
Should probably wait on https://github.com/mochajs/mocha/pull/1750 to merge this, since there'll be conflicts—once that's merged I'll rebase this.

wejrowski:
Will this allow me to build mocha in a single file with browserify?

ndhoule:
@wejrowski yes (though if you're just looking for a prebuilt version, there's `mocha.js` at the root of this repository)

wejrowski:
@ndhoule Hmm it looks like that one still has dependencies. I need one that has everything in one script. So no dependencies or requires. Just a stand alone script with everything I need.

ndhoule:
@wejendorp I'm not following—that file is totally self-contained. Can you explain more about how you're trying to use it? Are you trying to include it in your Browserify build? (Also, you'll probably get better help if you start a new issue rather than piggyback off this one.)

wejrowski:
I saw this issue and was curious if it would help what I'm trying to do. I have a Nashorn environment/service that can load scripts from a db and call functions. I need to store mocha and all its dependencies basically as one file. It's not a normal browser or node environment which is why I need everything zipped into one file. I'm attempting this with browserify but I'm getting a lot of problems.

ndhoule:
Yeah, open a separate issue. This might solve it but they're not the same issue, so better to track them separately

FWIW you can probably just wrap Mocha and your built stuff in an IIFE and concat the files together.

wejrowski:
Using browserify? How would the IIFE solve that? Not totally tracking.

And thanks. I will play with it some more and then open a ticket. Would love to solve this. I'm using Jasmine right now, which is great because it is just one file, but I'm having test issues with it.

ndhoule:
@wejrowski basically something like:

```js
// head.js
(function() {
```

```js
// tail.js
}).call(this);
```

Then just glue these fragments, mocha, and your browserified bundle together:

`cat head.js mocha.js browserified-stuff.js tail.js > build.js`

Or something. You'll probably have to play with it a bit to get it working right.

jbnicolai:
@ndhoule apologies for the delay, I'm finally back home and ready to put some time in mocha again ;)

It seems the (fantastic!) #1750 has now been merged. Could you rebase this PR to reflect those changes?

jbnicolai:
@wejendorp the built mocha file is just single dependency-free file.. please open a separate issue if you'd like to discuss this further.

ndhoule:
@jbnicolai Rebased.

ndhoule:
Removed the `__proto__` transform, since https://github.com/mochajs/mocha/pull/1753 makes it unnecessary

jbnicolai:
Ah, was just doing this locally as well. Thanks :+1:

jbnicolai:
Other than that I think it's good to merge :)

jbnicolai:
Oh, could you squash that last commit?

jbnicolai:
Nevermind, I'll do it manually. 

Thanks for all the work!

ndhoule:
> Oh, could you squash that last commit?

Done

ndhoule:
Awesome, glad to see it go in :ship: :ship: :ship:

Thanks!

boneskull:
@ndhoule @jbnicolai awesome! thanks much!

jbnicolai:
All credits go to @ndhoule, thanks for the (major!) effort, and we'll be sure to mention it in the release notes :)

Great step towards making mocha a more modern and maintainable project!

boneskull:
@jbnicolai until someone makes a PR building mocha with jspm...

PR #1730, opened 2015-06-07
Adding additional iojs flags

ryedog:
also removing duplicate flags

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -36,17 +36,22 @@ process.argv.slice(2).forEach(function(arg){
     case '--gc-global':
     case '--harmony':
     case '--es_staging':
-    case '--harmony-proxies':
-    case '--harmony_proxies':
+    case '--harmony_arrays':
+    case '--harmony_array_includes':
+    case '--harmony_arrow_functions':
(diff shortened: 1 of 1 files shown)


jbnicolai:
> also removing duplicate flags

And sorting them alphabetically, I see - explaining the removals of non-duplicate lines in the diff.

:+1: LGTM

boneskull:
noooo

boneskull:
ok ok.  it's just dangerous to reorder flags, but it looks like this won't break anything.  

it's still something we shouldn't be doing, but that'll have to wait for v3

jbnicolai:
But.. they're all part of the same fallthrough case :/

boneskull:
@jbnicolai I was referring to having two executables, which we shouldn't be doing.  As far as reordering flags, we've had PRs break stuff because of it.  I didn't look at the commit before I freaked out, which is something I unfortunately tend to do

ryedog:
@boneskull totally agree but in this case it didn't matter so figured alphabetizing would reduce the chance of other dupes making it back in. thanks for merging!

PR #1735, opened 2015-06-09
Ban mixing of promise and async APIs in `it` handler.

### LINKED ISSUE #1734: mixed use of promise and async APIs is hard to debug
I was converting some Mocha tests I'd written to use promises instead of async callbacks, because the underlying API I was testing had changed to use promises.

Given a `testSomething` function that returns a promise, I had a test that looked something like

```
it('should', function(done) {
  testSomething().then(function(result) {
    // check result
    done();
  });
});
```

and I converted it to

```
it('should', function(done) {
  return testSomething();
});
```

can you spot the bug that cost me some head-scratching?

It's probably pretty obvious in the context of this bug report and with such simple test code, but with much more complicated test code, I investigated a bunch of dead ends (e.g. is my promise actually resolving) before I realized that the promise I was returning was being ignored entirely by Mocha because of the function signature of my `it` handler.

I suggest that Mocha should consider it a coding error to have an `it` handler accept a `done` argument and then return a promise.

metamatt:
If the `it` handler accepts a `done` argument, it is using the async
API, and must not return a promise.

Closes #1734.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -227,7 +227,7 @@ Runnable.prototype.run = function(fn){
     this.resetTimeout();
 
     try {
-      this.fn.call(ctx, function(err){
+      var ret = this.fn.call(ctx, function(err){
         if (err instanceof Error || toString.call(err) === "[object Error]") return done(err);
         if (null != err) {
           if (Object.prototype.toString.call(err) === '[object Object]') {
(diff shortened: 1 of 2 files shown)


lib/runnable.js
```diff
@@ -238,6 +238,9 @@ Runnable.prototype.run = function(fn){
         }
         done();
       });
+      if (ret && typeof ret.then === 'function') {
+        done(new Error('runnable accepted "done" but returned promise'));
```
boneskull: this needs different language.

boneskull: how about `Supply a "done" parameter or return a Promise, but not both`

boneskull:
@metamatt 

[ ] rebase against v3.0.0
[ ] change wording of error
[ ] ensure this error *also* applies to hooks

then I think we're good to merge

hollomancer:
This was handled in #1320, closing this out.

PR #1740, opened 2015-06-10
Add note on allowUncaught method in the browser

amsul:
This refers to the `allowUncaught` method added in #1662.


jbnicolai:
Looks good to me :)

PR #1746, opened 2015-06-15
Add style directive

dfcreative:
Make able to include style via postcss-import

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "scripts": {
     "test": "make test-all"
   },
+  "style": "mocha.css",
   "dependencies": {
     "commander": "2.3.0",
     "debug": "2.0.0",


danielstjules:
This doesn't seem to be a valid property? http://wiki.commonjs.org/wiki/Packages/1.0 https://docs.npmjs.com/files/package.json

jbnicolai:
Can't find it documented anywhere either.

@dfcreative feel free to comment with an explanation and I'll consider reopening.

dfcreative:
@jbnicolai I believe it is not in the standard, but it is considered a good practice in some libs. For example, many [components](https://github.com/component/) use that, as well as [swiper](https://github.com/nolimits4web/Swiper) etc. It is just convenient to delegate styles building to postcss-import instead of declaring obscure `import` paths hierarchies. 
Same as [browser](https://github.com/substack/browserify-handbook#browser-field) field for browserify aliases.
So what do we pursue by strict following the standard?

boneskull:
Don't care if it isn't "standard". What I do want to know is just wtf it's for. :)

dfcreative:
Say, I have mocha browser test page and I need to attach mocha styles. With postcss-import i can just put `import 'mocha';` in my stylesheet. Without `style` directive i have to write `import '../node_modules/mocha/mocha.css'`. Which is not a big deal, actually)

PR #1750, opened 2015-06-16
Add linter

ndhoule:
@jbnicolai @boneskull 

Addresses https://github.com/mochajs/mocha/issues/1359

The Make task uses eslint's `--reset` flag to disable all default ESLint settings because [ESLint is planning on making that the default in the future](https://github.com/eslint/eslint/issues/2100#issuecomment-98030925) and I'd rather not have to go back through and change all the settings when that happens. I know there are a lot of settings here, but would be good to get some feedback on what's here. Most of the rules are pretty uncontroversial and we already follow them, but there are a few sections (labeled in `.eslintrc`) worth looking at.

We should also add more directories (e.g. `bin` and `test`), but I'm thinking we can do that in a separate PR to speed this one along to minimize merge conflicts. (Happy to handle that too.)


.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+    "quotes": [2, "single", "avoid-escape"],
+    "semi": [2, "always"],
+    "semi-spacing": [2, { "before": false, "after": true }],
+    "space-in-parens": [2, "never"],
+    "space-infix-ops": 2,
+    "space-return-throw-case": 2,
+    "space-unary-ops": [2, { "words": true, "nonwords": false }],
+    "use-isnan": 2,
+    "wrap-iife": 2
+  }
```
ndhoule: Kinda screwed up here and just deleted any rule I wanted to disable; in reality we should leave them in the config. `--reset` is not yet the default, so most editors' linting plugins will re-enable those rules.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+    "object-curly-spacing": [1, "always"],
+    "quote-props": [1, "as-needed"],
+    "radix": 1,
+
+    // Done, but up for discussion
+    "space-after-keywords": [1, "always"],
+    "space-before-blocks": [1, "always"],
+    "space-before-function-paren": [1, "never"],
+
+    // Pretty standard
```
ndhoule: Vast majority of these are recommendations by ESLint. They're all pretty safe but not attached to any of them.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+    "no-native-reassign": 1,
+    "no-negated-in-lhs": 1,
+    "no-nested-ternary": 1,
+    "no-new-func": 1,
+    "no-proto": 1,
+    "object-curly-spacing": [1, "always"],
+    "quote-props": [1, "as-needed"],
+    "radix": 1,
+
+    // Done, but up for discussion
```
ndhoule: Normalized all these in 857c74ed71b1fc13d4c0be81c064465251556d27.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+  "rules": {
+    // Maybe enable?
+    "one-var": 0,
+    "vars-on-top": 0,
+    "eqeqeq": 0,
+    "yoda": [0, "never"],
+    "no-use-before-define": 0,
+    "no-mixed-requires": [0, false],
+
+    // Discuss?
```
ndhoule: I like all of these settings, strong opinions weakly held.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+  "env": {
+    "browser": true,
+    "node": true
+  },
+
+  "rules": {
+    // Maybe enable?
+    "one-var": 0,
+    "vars-on-top": 0,
+    "eqeqeq": 0,
```
ndhoule: This one I'm slightly afraid of, haha

.eslintrc
```diff
@@ -0,0 +1,113 @@
+{
+  "env": {
+    "browser": true,
+    "node": true
+  },
+
+  "rules": {
+    // Maybe enable?
```
ndhoule: Some of these were suggested over in https://github.com/mochajs/mocha/issues/1359, some are more controversial

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+    "one-var": 0,
+    "vars-on-top": 0,
+    "eqeqeq": 0,
+    "yoda": [0, "never"],
```
ndhoule: Personally can't stand `0 == arr.length` but don't care either way

.eslintrc
```diff
@@ -0,0 +1,113 @@
+{
+  "env": {
+    "browser": true,
+    "node": true
+  },
+
+  "rules": {
+    // Maybe enable?
+    "one-var": 0,
```
ndhoule: Personally strongly prefer multiple vars for reduced diff noise/easier refactoring, but don't care much

ndhoule:
once we settle on a final rule list I'll remove those comments in `.eslintrc`

.eslintrc
```diff
@@ -0,0 +1,113 @@
+{
+  "env": {
+    "browser": true,
```
boneskull: if we're building with browserify, we can just use the `node` env

.eslintrc
boneskull: I use one `var` b/c of less code, but don't feel strongly about it.

.eslintrc
```diff
@@ -0,0 +1,113 @@
...
+    "node": true
+  },
+
+  "rules": {
+    // Maybe enable?
+    "one-var": 0,
+    "vars-on-top": 0,
```
boneskull: would prefer this be `[2]`.  reduces potential bugs

.eslintrc
ndhoule: Mostly to pick up on `document`, `window` etc. in the reporters, `lib/utils.js`, etc. We should probably remove it eventually though

.eslintrc
boneskull: why would we disable this?

.eslintrc
boneskull: this is fine.  it can reduce bugs where assignments were supposed to be conditionals, but that is pretty rare

.eslintrc
ndhoule: Only because there's a lot of `==`s in the codebase right now, and wasn't sure if that was a stylistic choice

.eslintrc
boneskull: `0` means don't check, right?  I prefer both of these to be enabled.

.eslintrc
boneskull: `1` means warning, yes?  my preference is to make nothing a warning and everything an error :smile:

.eslintrc
ndhoule: Yeah totally, I generally enable the jshint/eslint rules that prohibit assignments inside of `if`s so I've never been bitten by this

.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: node_js
+before_install:
+- npm install -g npm@latest
```
boneskull: this might conflict with the iojs target, since iojs needs a special npm, right?

.eslintrc
ndhoule: Right, I'd like to enable all of these, just didn't want to presume to dictate style

.travis.yml
ndhoule: First I've heard of that, but I don't think I've updated npm in any of my io.js projects. will investigate

.eslintrc
ndhoule: I'd rather have them all be errors, too. My thinking was, make things that are irritating/time consuming to fix, or need some research ([example](https://github.com/mochajs/mocha/pull/1750/files#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR566)), and then enable them once those issues are sorted. I'm down to fix everything now, though

.eslintrc
travisjeffery: +1 multiple vars

.eslintrc
travisjeffery: +1 to leave them as warnings for now to try it out

.eslintrc
boneskull: well, somebody has to :smile:

I didn't know about ESLint dropping the defaults.  so I'm not sure which the defaults actually are.  but... this is what I currently work with (plus defaults):

```yaml
---
env:
  node: true
rules:
  brace-style: 0
  camelcase: 0
  comma-style:
    - 2
    - last
  consistent-this:
    - 2
    - self
  consistent-return: 0
  func-style:
    - 2
    - declaration
  max-len:
    - 2
    - 80
    - 4
  max-params:
    - 1
    - 4
  no-shadow: 0
  no-cond-assign:
    - 2
    - except-parens
  no-else-return: 2
  no-floating-decimal: 2
  no-multiple-empty-lines:
    - 2
    - max: 2
  no-nested-ternary: 2
  no-reserved-keys: 2
  no-return-assign: 0
  no-self-compare: 2
  no-underscore-dangle: 2
  no-use-before-define: 0
  no-void: 2
  operator-assignment:
    - 2
    - always
  quotes:
    - 2
    - single
  radix: 2
  space-after-keywords:
    - 2
    - always
  space-before-blocks:
    - 2
    - always
  spaced-line-comment: 0
  valid-jsdoc:
    - 1
    - requireReturn: false
  vars-on-top: 2
  wrap-iife: 2
```

so this is the direction I'd like to go in.

(and yes, I'd prefer yaml)

.travis.yml
boneskull: yeah I was having some weird issues if I did something like install iojs w/ installer then `npm install -g npm`, wherein `node-gyp` was broken.  but I may have done something else weird.

.eslintrc
boneskull: in those cases we could do a 

```js
/* eslint-env browser */
// ^ top of file
```

.eslintrc
boneskull: fine w/ me

.eslintrc
boneskull: if it was, it wasn't mine :wink:

would prefer `==` is never used

.eslintrc
boneskull: sure

boneskull:
@ndhoule thx for all the work on this

.eslintrc
ndhoule: :+1:

everything in `{lib,test}/browser` should have its env set to browser, maybe a few reporters. I think the references to browser stuff elsewhere is feature detection. was being lazy but I'll just fix those

.travis.yml
ndhoule: I can't repro or find any reference to npm versioning/iojs, and the tests seem to be passing. If anyone knows better I'm all ears

ndhoule:
@boneskull Think I addressed all your comments (added your config, converted `.eslintrc`s to YAML, addressed notes like removing `browser: true`, etc). Disabled some of the rules that were making linting output particularly noisy; I think it's important to minimize linting noise because people will end up ignoring the linter otherwise. Will take care of some of those later (or other folks are welcome to tackle them :P).

Just rebased, should be good to go—let me know if you want to see other changes

.travis.yml
rstacruz: This might better illustrate the need for a special npm for iojs:

https://github.com/aredridel/homebrew-iojs/blob/master/Formula/iojs.rb

ndhoule:
Only point to updating npm was to fix 0.8 CI builds, so amended it PR to only update npm when necessary. Also added npm 1.3.7 (introduction of caret) to Mocha's `engines` field.

ndhoule:
just pinging here. I think I've addressed all issues, let me know if that's not the case

boneskull:
taking another look now.  I'm probably going to merge this unless someone objects.

boneskull:
@ndhoule The only thing I see that would prevent me from merging as-is is the npm change.  Let me check on `package.json` really quicklike

boneskull:
@ndhoule OK, so *that's* necessary because why?  If we changed the caret in the `eslint` version to a tilde, would the 0.8 Travis run work?

boneskull:
If we can avoid that, it'd be great, otherwise this would have to be rebased into branch `v3.0.0`

dasilvacontin:
`~` matches only new patch versions. If we plan on manually updating the dependencies, I'd say it's fine leaving the tilde. Otherwise we might miss some bugfixes due to some new feature release.

boneskull:
well, I know that, I just wasn't sure if the caret was there for a reason

dasilvacontin:
yeah, sorry, too early in the morning here :x

ndhoule:
@boneskull 

> @ndhoule OK, so that's necessary because why? If we changed the caret in the eslint version to a tilde, would the 0.8 Travis run work?

We can change the caret to tildes, which should leave Mocha installable on npm > 1.3.7, so long as users aren't trying to install it with its dev dependencies.

ESLint has carets in its dev dependencies, so including it will make npm 1.3.7+ a hard dev dependency for Mocha going forward—so those lines in `.travis.yml` will still be necessary.

ndhoule:
@boneskull rebased from master and fixed the caret issue.

boneskull:
ty. That's fine with the dev deps.

ndhoule:
@boneskull Cool. Removed `npm` from the engines field, since removing the caret made that not a req. This should be good to go, then.

boneskull:
gonna try to merge this.

boneskull:
just waiting on CI.  oh boy oh boy oh boy

boneskull:
@ndhoule thanks a ton!

PR #1752, opened 2015-06-17
Stop runner sending SIGINT to itself from within SIGINT handler

gsilk:
This prevents an infinite loop when the child process (the target of the runner, ie your tests) installs a SIGINT / SIGTERM handler. You might want to do this, for example, if you need to ensure clean shutdown of your tests when the user presses Ctrl+C. I ran into this problem when spawning a subprocess from the setup portion of my tests, and trying to capture SIGINT in the test suite.


jbnicolai:
Hmm, fair enough. I've done some testing, and all seems to work. Odd that this was originally included in the first place..

boneskull:
@gsilk good catch thx. My bad

gsilk:
@boneskull np

Issue #1728, opened 2015-06-07
Replace __proto__ throughout with setPrototypeOf shim


ndhoule:
See https://github.com/mochajs/mocha/pull/1727#discussion_r31872646 for details

jbnicolai:
:+1:

boneskull:
I'm just gonna have to ask why we need to do this in the first place.  If we're subclassing, we can use `util.inherits()`.  If we're modifying the prototype of an existing object, well, we probably shouldn't do that.

ndhoule:
`util.inherits` is fine in Node, but not in the browser, and Browserify's [`util` shim](https://github.com/defunctzombie/node-util) isn't ES3-compatible. We could come up with an ES3-compatible `util` shim, but at that point, unless we're relying on other functions from `util` we might as well just use a standalone lib.

ndhoule:
(Sorry, partially my fault here for naming this issue poorly.)

PR #1753, opened 2015-06-17
Replace __proto__ references with inherits module

ndhoule:
cc @boneskull @jbnicolai 

Resolves https://github.com/mochajs/mocha/issues/1728.

Depends on merging https://github.com/mochajs/mocha/pull/1727 because it pulls in an npm dependency; uses style from https://github.com/mochajs/mocha/pull/1750

### DIFF
--- a/lib/hook.js
+++ b/lib/hook.js
@@ -3,6 +3,7 @@
  */
 
 var Runnable = require('./runnable');
+var create = require('lodash.create');
 
 /**
  * Expose `Hook`.
@@ -25,7 +26,10 @@ function Hook(title, fn) {
 /**
  * Inherit from `Runnable.prototype`.
  */
-Hook.prototype.__proto__ = Runnable.prototype;
+
+Hook.prototype = create(Runnable.prototype, {
+  constructor: Hook
+});
 
 /**
  * Get or set the test `err`.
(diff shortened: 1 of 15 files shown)


boneskull:
I think I'd rather go with LoDash here, since there's going to be a ton of other shims we can replace it with.  @mochajs/mocha thoughts?

ndhoule:
Which lodash function would you use here? `_.create`?

ndhoule:
@boneskull Refactored this to use `lodash.create`, which can easily be swapped out for a full `lodash` when necessary. I think this version is less expressive than the `inherits` version but meh

boneskull:
@ndhoule why do you feel it's less expressive?

jbnicolai:
@nathanboktae I agree
@boneskull `create(..., {constructor: ...})` seems less expressive than `inherit(..., ...)`.

On the other hand, since we're intending to go with lodash anyways, I say we merge this version :)

Perhaps to make this a bit nicer we could refactor it into a utility method in `Base`

```javascript
function inherit(prototype, constructor) {
  return create(prototype, {
    constructor: constructor
  });
```

But no strong feelings about that.

ndhoule:
haha pretty much exactly what @jbnicolai said, but really I don't care—I say merge it as is, it's certainly better than the `__proto__` version

jbnicolai:
Agree, still a huge improvement :+1:

PR #1754, opened 2015-06-18
_mocha: Allow boolean --reporter-options

papandreou:
Previously "mocha --reporter-options foo" would blow up because an equal sign is required. This is quite unhandy for reporters that need to accept boolean (flag-ish) options.

This commit changes that so that a value-less reporter option will be interpreted as having a value of true.

This means that --reporter-options foo=bar,quux will turn into { foo: 'bar', quux: true }


jbnicolai:
Am I correct in understanding that this is a non-breaking change?

Looks good :)

papandreou:
Yes, you are correct -- and thanks :)

PR #1758, opened 2015-06-19
Add cross-frame compatible Error checking for fail

outdooricon:
When passing errors through between frames, AssertionErrors and the like will fail an instanceof check against Error. We need a compatible check. According to [Mozilla Error Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message), Error prototype will always have message defined.

All tests pass: https://travis-ci.org/outdooricon/mocha/builds/67555897

It's worth noting that this currently prevents us from upgrading the Konacha project to use latest Mocha.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -202,7 +202,7 @@ Runner.prototype.fail = function(test, err) {
   ++this.failures;
   test.state = 'failed';
 
-  if (!(err instanceof Error)) {
+  if (!(err instanceof Error || err && typeof err.message == 'string')) {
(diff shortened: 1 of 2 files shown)


jbenner:
:+1: this has been an issue for me as well

danielstjules:
Looks good, thanks!

outdooricon:
@danielstjules Do you know when the next patch version of Mocha is scheduled for release? Just wondering for when I can fix Konacha with this...

PR #1761, opened 2015-06-22
Explicitly check `null` and `undefined` values in exports.type

chromakode:
Some older JS engines, such as PhantomJS, return the `window` object instead of "[object Null]", as would be expected here, crashing the test runner.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -372,7 +372,11 @@ var emptyRepresentation = function emptyRepresentation(value, type) {
  * @returns {string}
  */
 exports.type = function type(value) {
-  if (typeof Buffer !== 'undefined' && Buffer.isBuffer(value)) {
+  if (value === undefined) {
+    return 'undefined';
+  } else if (value === null) {
+    return 'null';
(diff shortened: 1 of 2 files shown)


boneskull:
@chromakode thanks, but *where* does the test runner crash?  can you please add a test as well?

chromakode:
Thanks for the review :) I believe the test runner crashes in the process of formatting object diff output. For a test like:

    chai.assert.deepEqual({a: null}, {a: 'test'})

I get a relatively inscrutable output like:

    TypeError: 'null' is not an object (evaluating 'value.toString')

With an unrelated traceback.

I considered writing a test case that shimmed `Object.prototype.toString`, but wasn't sure this would be a valid test. I'll add it and update this PR.

chromakode:
Ok @boneskull , added a test to replicate the behavior of PhantomJS and rebased. I also added a test for `type(undefined)` in the general case for consistency.

jbnicolai:
Thanks for the work, @chromakode, and including a testcase!

chromakode:
Great! Thanks for merging it. :)

Issue #1764, opened 2015-06-25
Type check window instead of document to detect environment


limdauto:
I think it's generally accepted that to check whether an environment is node or browser, we should type check `window`, instead of `document`, because sometimes document is mocked for testing purpose.

jbnicolai:
I'd rather use `process.browser`, see https://github.com/mochajs/mocha/blob/master/lib/utils.js#L690

If you'd like to open a PR for this, please do!

boneskull:
`process.browser` please. I tend to stub `window` when unit testing my browserified code on the server.

ndhoule:
Yeah, `process.browser` is the way to go. https://github.com/mochajs/mocha/pull/1727 changed a bunch of these env checks to use `process.browser`, but I pretty much did the bare minimum to get it working so I'm sure there are more.

anmuel:
I'd like to contribute here :)
Was the mentioned TODO the only place to look at? (I doubt it)

boneskull:
@anmuel Thanks.  You could grep around for `window` or `document` and see if there's anything else that needs addressing, but that's about all I know of right now.

PR #1765, opened 2015-06-25
Type check window for env detection instead of document

limdauto:
Issue: https://github.com/mochajs/mocha/issues/1764


jbnicolai:
See discussion in #1764

PR #1771, opened 2015-06-30
update Runner to check failures w/ lodash.isError(); closes #1770

### LINKED ISSUE #1770: errors thrown from a vm aren't recognized as errors
As you can read in the comments on commit 02d421b1b548c55ce8c0bd12aae3e94f635aa4ba, it seems that if code under test uses `vm.runInNewContext()` and throws an `Error`:

```js
try {
  vm.runInNewContext('throw new TypeError()');
} catch (e) {
  e instanceof Error; // false
}
```

I think we can safeguard against this by duck-typing instead of an `instanceof` check.

boneskull:
travis seems to be failing on `v3.0.0`, but I don't think it's related to this

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -2,16 +2,17 @@
  * Module dependencies.
  */
 
-var EventEmitter = require('events').EventEmitter
-  , debug = require('debug')('mocha:runner')
-  , Pending = require('./pending')
-  , Test = require('./test')
-  , utils = require('./utils')
-  , filter = utils.filter
-  , keys = utils.keys
-  , type = utils.type
-  , stringify = utils.stringify
-  , stackFilter = utils.stackTraceFilter();
+var EventEmitter = require('events').EventEmitter,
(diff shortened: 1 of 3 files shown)


danielstjules:
Why target 3.0.0? Aren't you relaxing the restrictions? And don't we want to keep the spec for "should emit the error when failed with an Error-like object"?

Also, see https://github.com/mochajs/mocha/issues/1770#issuecomment-117243799

boneskull:
@danielstjules because it's a breaking change.  I'm trying to make it *more* strict.  We want people to throw `Error` instances, or subclasses thereof.  We want stack traces.

boneskull:
(and no, I don't want to keep the spec, because I don't want to allow any object with a string `message` property to be considered an "error")

boneskull:
@jbnicolai after discussion with @danielstjules I think I'm gonna drop this one.

PR #1681, opened 2015-05-04
Implement support for multiple reporters

benvinegar:
Squashed commit of #1360 (previously #930), against v3.0.0.

Necessary for completion:

  - [ ] support [distinct output streams for each reporter](https://github.com/mochajs/mocha/pull/1360#issuecomment-55513864) (e.g. 'spec' → stdout while 'json' → file)
  - [x] `--reporter-options` needs to be capable of [configuring each reporter](https://github.com/mochajs/mocha/pull/1360#issuecomment-72354839)

PR #1360, opened 2014-09-14
Multiple reporters support (continuation of #930)

benvinegar:
This is a continuation of @misterdjules PR #930 – multiple reporters support.

On top of @misterdjules work, it has:
* Merged up w/ master (original patch was > 1 year old)
* Restores 'spec' as the default reporter
* Fixed unit tests
* An additional unit test for multiple reporters (tests/reporters/multiple.js)
* Renames variables to be closer to original mocha.js code (e.g. `_reporter` -> `_reporters`)
* Keeps original `Mocha.prototype.reporter` interface (vs the plural `reporters`) for API compatibility. IMHO anybody using Mocha directly shouldn't have to change their code to remain compatible.

It does not implement Growl as a separate reporter, as was discussed [here](https://github.com/visionmedia/mocha/pull/930#issuecomment-48971366). I think it could be done as a follow-up patch just fine, but if you'd prefer that was done before accepting then that's cool too.

PR #1772, opened 2015-06-30
Implement support for multiple reporters

benvinegar:
Continuation of #1681, targeted against 2.x (master) instead of 3.x.

Necessary for completion:

  - [ ] support [distinct output streams for each reporter](https://github.com/mochajs/mocha/pull/1360#issuecomment-55513864) (e.g. 'spec' → stdout while 'json' → file)
  - [x] `--reporter-options` needs to be capable of [configuring each reporter](https://github.com/mochajs/mocha/pull/1360#issuecomment-72354839)

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/1772)

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -131,32 +131,63 @@ Mocha.prototype.addFile = function(file){
 };
 
 /**
- * Set reporter to `reporter`, defaults to "spec".
+ * Set reporters to `reporters`, defaults to "spec".
  *
- * @param {String|Function} reporter name or constructor
- * @param {Object} reporterOptions optional options
+ * @param {String|Function|Array of strings|Array of functions} reporter name as string,
+ * reporter constructor, or array of constructors or reporter names as strings.
+ * @param {Object} reporterOptions optional options TODO FIXME
  * @api public
  */
-Mocha.prototype.reporter = function(reporter, reporterOptions){
-  if ('function' == typeof reporter) {
-    this._reporter = reporter;
-  } else {
-    reporter = reporter || 'spec';
-    var _reporter;
-    try { _reporter = require('./reporters/' + reporter); } catch (err) {}
-    if (!_reporter) try { _reporter = require(reporter); } catch (err) {
-      err.message.indexOf('Cannot find module') !== -1
-        ? console.warn('"' + reporter + '" reporter not found')
-        : console.warn('"' + reporter + '" reporter blew up with error:\n' + err.stack);
+
+Mocha.prototype.reporter = function(reporters, reporterOptions){
+  // if no reporter is given as input, default to spec reporter
+  reporters = reporters || ['spec'];
+  reporterOptions = reporterOptions || {};
+
+  // If reporters argument is not a list, turn it into a list of reporter names
+  // or constructors that we'll iterate on right after to initialize them
+  if (!Array.isArray(reporters)) {
+    if (('string' == typeof reporters) ||
+        ('function' == typeof reporters)) {
+      reporters = [reporters];
     }
-    if (!_reporter && reporter === 'teamcity')
-      console.warn('The Teamcity reporter was moved to a package named ' +
-        'mocha-teamcity-reporter ' +
-        '(https://npmjs.org/package/mocha-teamcity-reporter).');
-    if (!_reporter) throw new Error('invalid reporter "' + reporter + '"');
-    this._reporter = _reporter;
   }
-  this.options.reporterOptions = reporterOptions;
+
+  // Load each reporter
+  this._reporters = reporters.map(function (reporterConfig) {
+    if ('function' == typeof reporterConfig) {
+      return {
+        fn: reporterConfig,
+        options: reporterOptions
+      };
+    } else {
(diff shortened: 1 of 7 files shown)


lib/mocha.js
```diff
@@ -424,23 +455,45 @@ Mocha.prototype.run = function(fn){
...
   if (options.useColors !== undefined) {
     exports.reporters.Base.useColors = options.useColors;
   }
+
   exports.reporters.Base.inlineDiffs = options.useInlineDiffs;
 
-  function done(failures) {
+  function runnerDone(failures) {
+    var remain = reporters.length
+      , reporterDone = function(failures) {
```
boneskull: fyi: with the linting work that's being done, we're not going to be using leading commas.  also, the below conditional will need curlies

lib/reporters/base.js
```diff
@@ -225,11 +226,21 @@ exports.list = function(failures){
...
+function Base(runner, options, path) {
   var self = this
     , stats = this.stats = { suites: 0, tests: 0, passes: 0, pending: 0, failures: 0 }
     , failures = this.failures = [];
 
+  this.options = options;
+  this.path = path;
+
+  if (path) {
+    if (!fs.createWriteStream) {
```
boneskull: I was thinking about this while working on my branch for plugin support.  

It seems what'd be even more useful is a generic writable stream.  at the risk of ballooning the inheritance tree, I went with keeping the base reporter completely generic.  then, browser-based reporters would use it, and node-based ones would use a subclass `StreamReporter` or something.  they'd just write to their internal stream--whatever that happened to be.  so what if we passed in a writable stream instance instead of a path?  that'd probably eliminate the conditionals below.

you don't have to do this, since my code will conflict with just about everything anyway, but something to think about.

boneskull: that gives me an idea for a server-side reporter which launches the HTML reporter and uses websockets to stream the results. :smile:

lib/mocha.js
benvinegar: Roger.

lib/reporters/base.js
benvinegar: Yeah, the idea of a generic writable stream hit me too ... but I'm hesitant to do much refactoring though, given how long this pull request has been floating around (almost a year). If there's a way to get it out as-is, and follow-up with improvements, I'm more keen to that.

boneskull: understood

boneskull:
@benvinegar 

>  support distinct output streams for each reporter (e.g. 'spec' → stdout while 'json' → file)

Is this in-progress or...?

danielstjules:
@benvinegar Awesome work on this btw :)

> Is this in-progress or...?

@boneskull It might be difficult without a writeable stream being passed to each reporter, since they all invoke either process.stdout.write or console.log directly. Ideally, mocha would pass process.stdout to any reporter writing to stdout, and an fs.WriteStream when writing to a file (or just some buffer that gets written at the end of exec). At that point you'd probably end up with something similar to what you mentioned in https://github.com/mochajs/mocha/pull/1772#discussion_r33631763 

@benvinegar Did you have an easier solution in mind for getting the distinct output streams working? I've got some cycles this week too if you wanna collaborate.

ajwhite:
This would be great to have :)

lostthetrail:
This is a super useful feature. It appears the development has stalled on it?

benvinegar:
This is stalled. Sorry everyone. Ultimately I was driven on this project because it was important for a project I was working on, but I've since moved on and it's no longer in my sights.

@danielstjules – This is a super late reply, but if we could get together to push through on this, I'm game.

Otherwise I can help someone us take over.

antoinebrault:
+1 really need this for CI.

lpinca:
I would also love to have this for the same reason mentioned by @antoinebrault.

boneskull:
Anyway, a prototype of #1969 will have this built in

machineghost:
Please, someone finish merging this; having to run your entire suite multiple times just to get xunit and coverage reports is lame.

boneskull:
@antoinebrault @machineghost @lpinca you may want to take this PR over and see if you can get it finished.  my focus is elsewhere atm

santiagoaguiar:
@boneskull @benvinegar  I'll be integrating the current work on latest and add the missing pieces, any things to keep in mind? 

I'm working over master, should I do it on 3.0 instead?

santiagoaguiar:
First question: most of the reporters in mocha do a direct `process.stdout.write` or `console.log`, but to ensure they can be used while passing `--reporter reporter:outputfilename` they should instead use the inherited `write` method from the `Base` reporter that was added on this PR, which does a `process.stdout.write` (which, by the way,  should probably fallback to `console.log` for browser environments?). I guess this is what was being discussed on https://github.com/mochajs/mocha/pull/1772#discussion_r33631763.

Any reason not to do such change?.

Since this requires updating every reporter, I'm guessing we should do that on an independent PR?.

benvinegar:
See #2184

PR #1774, opened 2015-07-01
support escaped spaces in cli options

adamgruber:
The current parsing of options does not allow spaces within an option so this fixes that. This is particularly useful in the case of reporter options where a custom reporter may want to accept a string option that contains a space.

See https://github.com/adamgruber/mochawesome/issues/20 for an example.

This patch would require any spaces to be escaped like `--option option\ with\ spaces`

### DIFF
--- a/bin/options.js
+++ b/bin/options.js
@@ -21,8 +21,12 @@ function getOptions() {
 
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
+      .replace(/\\\s/g, '%20')
       .split(/\s/)
-      .filter(Boolean);
+      .filter(Boolean)
+      .map(function(value) {
+        return value.replace(/%20/g, ' ');
+      });
 
     process.argv = process.argv
       .slice(0, 2)


boneskull:
I'm unclear why you can't just use double-quotes?  Anyway, this looks ok to me.

bin/options.js
```diff
@@ -22,9 +22,10 @@ function getOptions() {
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
           .trim()
+          .replace(/\\\s+/, '%20')
```
jbnicolai: - The `+` would turn an escaped space followed by one or multiple unescaped spaces into a single `'%20'`.
- The lack of a `//g` flag would make this regex only transform the first escaped whitespace and leave all others

E.g.:

```javascript
"foo bar\\      \\ baz\\ boo".replace(/\\\s+/, '%20')
"foo bar%20\ baz\ boo"
```

bin/options.js
```diff
@@ -22,9 +22,10 @@ function getOptions() {
...
           .split(/\s+/)
-          .filter(function(value) {
-            return value ? true : false;
```
jbnicolai: Why did you remove the filter?

bin/options.js
```diff
@@ -22,9 +22,10 @@ function getOptions() {
...
+          .map(function(value) {
+            return value ? value.replace('%20', ' ') : false;
           });
```
jbnicolai: Why are falsey values mapped to `'false'`?

jbnicolai:
> I'm unclear why you can't just use double-quotes?

I agree, I would always go for quoting as well. On the other hand, supporting escaped whitespaces is pretty standard as well, so why not ;)

> Anyway, this looks ok to me.

It looks pretty broken to me :stuck_out_tongue:

bin/options.js
```diff
@@ -22,9 +22,10 @@ function getOptions() {
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
           .trim()
```
jbnicolai: If the string ends with `"\\ "` trim will remove the trailing space. Edge case, I admit, but these two lines should be swapped.

adamgruber:
And this is why I shouldn't code late at night. :stuck_out_tongue_winking_eye: Thanks for the feedback, I'll fix it.

bin/options.js
adamgruber: I was thinking (incorrectly) that the `map` would do the same as the `filter` so it wasn't needed. 

That said, I'm not sure how useful `filter` is here. We've already trimmed everything and split against multiple spaces so I don't see a case where `value` would be `false` or `undefined`. What am I missing?

jbnicolai: I agree. I've changed these lines a bit to simplify them, see https://github.com/mochajs/mocha/compare/630156d35a19bf5dfe722899084c3f7edaee764a...9a340570dfc5d56d7e8cda1a6046a8f169e85e31

adamgruber:
Made the changes and cleaned up the history.

jbnicolai:
:+1:  thanks!

Issue #1743, opened 2015-06-12
Should.js on site points to the old should.js repo


cloud-walker:
![screenshot from 2015-06-12 12 30 05](https://cloud.githubusercontent.com/assets/1144075/8128421/0b0c58f8-10ff-11e5-9ad1-6085d7e53278.png)

The link on the mochajs.org for should.js points to the wrong repo

Issue #1738, opened 2015-06-10
Idea: mocha for your browser instructions


rstacruz:
Today, to get mocha running on your browser, you have a snippet on the mochajs.org website that looks like this:

```js
<html>
<head>
  <meta charset="utf-8">
  <title>Mocha Tests</title>
  <link rel="stylesheet" href="mocha.css" />
</head>
<body>
  <div id="mocha"></div>
  <script src="jquery.js"></script>
  <script src="expect.js"></script>
  (snip...)
```

This isn't ideal because:

1. there's no indication where `mocha.css` and `expect.js` (et al) come from.
2. it doesn't even mention the `mocha init` command.

Instead, I propose to replace it with something more useful like so:

```html
<!doctype html>
<html>
<head>
  <title>Mocha</title>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <link href="https://cdn.rawgit.com/mochajs/mocha/2.2.4/mocha.css" rel="stylesheet" />
</head>
<body>
  <div id="mocha"></div>
  <script src="https://cdn.rawgit.com/mochajs/mocha/2.2.4/mocha.js"></script>
  <script src="https://cdn.rawgit.com/chaijs/chai/2.3.0/chai.js"></script>
  <script>mocha.setup('bdd')</script
  <script src="tests.js"></script>
  <script>mocha.run()</script>
</body>
</html>
```

rstacruz:
would be happy to PR this to the gh-pages branch if anyone agrees.

danielstjules:
Big +1 from me :) I think it'd be super helpful for newcomers.

travisjeffery:
Sounds cool

rstacruz:
probably no PR until #1729 #1726 #1640 get resolved :)

PR #1729, opened 2015-06-07
gh-pages refactor

rstacruz:
Does stuff outlined in #1714.

No content changes, just reorganizing things.

PR #1726, opened 2015-06-05
build mochajs.org with Jekyll; closes #1640

boneskull:
[Demo](http://boneskull.com/mocha)

Lots of stuff here, but the main things:

- use Jekyll
- convert `index.md` to *actual GFM Markdown*
- a lot of edits of the documentation, including, but not limited to:
  - reorganization
  - grammar fixes
  - code style fixes
  - rewording confusing sentences
  - add a couple more examples
  - dynamically generate TOC
  - use CDNs for JS, though there's basically none

I've taken every effort to make the site look as close to the current version as it does now, and keep it compatible with old browsers.  However, I was not able to reuse the syntax highlighting scheme, as highlighting is now done by Pygments.

screenshot 1

screenshot 2

PR #1777, opened 2015-07-05
Build gh pages with jekyll

jbnicolai:
This is an attempt to merge the pull requests by @boneskull and @rstacruz in #1729 and #1726. Discussed in #1640 and #1657.

I think this is the best of both worlds :smile: but please let me know what you guys think.


jbnicolai:
I've cherry-picked and rewrote all commits that occured on the `gh-pages` branch but were not included in the branch off point where @boneskull started work on this branch. A clean merge into `gh-pages` should now be possible, without loss of any work.

jbnicolai:
Also resolved issue #1743

jbnicolai:
And resolved #1738

jbnicolai:
See https://jbnicolai.github.io/mocha for a demo.

_includes/footer.html
```diff
@@ -0,0 +1,3 @@
+<footer>
+  <span>&copy; 2011-2015 TJ Holowaychuk, Travis Jeffery, Christopher Hiller & Joshua Boy Nicolai Appelman. All rights reserved.  Licensed <a href="http://opensource.org/licenses/MIT" target="_blank">MIT</a></span>
```
boneskull: this is a bit ambiguous.  but at any rate, I don't think it's appropriate to release this code under the MIT license.

I believe because it's under MIT, and others besides TJ would own the copyright to this particular incarnation, we're allowed to relicense the site.

As such, a more suitable license would be [CC](http://creativecommons.org).  Let's change this text to the following, if you don't mind:

```
<a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/"><img alt="Creative Commons License" style="border-width:0" src="https://i.creativecommons.org/l/by-sa/4.0/80x15.png" /></a><br /><span xmlns:dct="http://purl.org/dc/terms/" href="http://purl.org/dc/dcmitype/InteractiveResource" property="dct:title" rel="dct:type">mochajs.org</span> is licensed under a <a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/">Creative Commons Attribution-ShareAlike 4.0 International License</a>.<br />Based on a work at <a xmlns:dct="http://purl.org/dc/terms/" href="https://github.com/mochajs/mocha/tree/gh-pages" rel="dct:source">https://github.com/mochajs/mocha/tree/gh-pages</a>.
```

Then we may want to update `README.md` to note that while the *source* is licensed MIT, a portion of the contents of the `_site/` directory are in fact licensed with CC.

travisjeffery:
-1 on changing the license there. i think that says more about the project as a whole than that particular page.

boneskull:
> -1 on changing the license there. i think that says more about the project as a whole than that particular page.

I'm not sure what you mean.  The license is for the site...

travisjeffery:
yeah updated my comment, just mention that the project is mit somewhere on it

boneskull:
@travisjeffery OK, fair enough.

jbnicolai:
> yeah updated my comment, just mention that the project is mit somewhere on it

Agree with you there :)

boneskull:
@jbnicolai let me/us know if that doesn't make sense.  The CC says "mochajs.org is licensed CC..." but then thereafter we should mention Mocha is licensed MIT.

jbnicolai:
@boneskull yep, was in the middle of a too long response on copyright nuances, but I think we all agree. Let me update that line.

travisjeffery:
imo we could just get rid of that copyright for the page at the bottom, looking at other open source projects and they don't have something similar.

jbnicolai:
@travisjeffery I liked @boneskull's proposal to go with CC, which would mean people could use the source for other projects.

boneskull:
@travisjeffery If there's no license on the site, the site could be construed to not be released under any license, which means while it's "open source", you wouldn't be allowed to use it or copy it in any way.  IMO, it's better to be explicit.

_includes/footer.html
jbnicolai: I've changed it to precisely that, but it doesn't come out looking all that good, see http://blog.jbnicolai.com/mocha/#more-information

travisjeffery:
afaik we can't really re-license something like this. since this page now has many different authors, we'd have to get each contributor to the page's consent to do so. here's a podcast with postsharp talking about when they did it: http://hanselminutes.com/286/startup-series-postsharp-from-open-source-to-commercial-business-with-gael-fraiteur

jbnicolai:
@travisjeffery we do the same relicensing for every pull-request we accept. The work someone presents is technically under their copyright, and we never ask them or make them sign anything when we accept it. Then we present the entire body of work under our original MIT licence.

Of course, I think this is a non-issue - although organisations like https://github.com/google do make you explicitly sign their contributers licence agreement for just this reason.

jbnicolai:
Actually, how about I revert the chances I made to the footer and we discuss this in an other issue?

That way we can move on merging this :)

boneskull:
We need to leave the "license" as-is; it's not actually licensed.

> © 2011 TJ Holowaychuk. All rights reserved.

boneskull:
@jbnicolai after we revert the license change(s) let's merge this plz

jbnicolai:
@boneskull that line has already been reverted. Merge-time!

Issue #1635, opened 2015-03-31
afterEach halts suite


glenjamin:
This is a common pattern with tools like sinon or nock - to have afterEach do some standard verification at the end of each test.

I'm fairly sure this used to work - not sure what's changed.

``` js
var assert = require('assert');

describe("afterEach", function() {
    var verificationFlag;
    beforeEach(function() {
        verificationFlag = 0;
    });
    afterEach(function() {
        assert.equal(verificationFlag, 1);
    });

    it("0", function() { })
    it("1", function() { verificationFlag = 1 })
    it("2", function() { verificationFlag = 2 })
});
```

Expected output: 3 tests, 3 failures
Actual output: 1 test, 1 pass, 1 fail (from hook)

m-reiniger:
I might be off with that, but I think you just forgot about done();

danielstjules:
Looks like it's been the same behavior for a while. Mocha assumes that cleanup couldn't be done properly if an exception was thrown from the afterEach hook.

``` bash
# --------------------------------
# 2.2.1
# --------------------------------
dstjules:~/git/mocha ((2.2.1))
$ ./bin/mocha test.js


  ․․

  1 passing (8ms)
  1 failing

  1) afterEach "after each" hook:
     AssertionError: 0 == 1
      at Context.<anonymous> (/Users/dstjules/git/mocha/test.js:9:16)
      at callFn (/Users/dstjules/git/mocha/lib/runnable.js:266:21)
      at Hook.Runnable.run (/Users/dstjules/git/mocha/lib/runnable.js:259:7)
      at next (/Users/dstjules/git/mocha/lib/runner.js:268:10)
      at Immediate._onImmediate (/Users/dstjules/git/mocha/lib/runner.js:289:5)
      at processImmediate [as _immediateCallback] (timers.js:357:17)

# --------------------------------
# 2.1.0
# --------------------------------
dstjules:~/git/mocha ((2.1.0))
$ ./bin/mocha test.js


  ․․

  1 passing (11ms)
  1 failing

  1) afterEach "after each" hook:
     AssertionError: 0 == 1
      at Context.<anonymous> (/Users/dstjules/git/mocha/test.js:9:16)
      at callFn (/Users/dstjules/git/mocha/lib/runnable.js:251:21)
      at Hook.Runnable.run (/Users/dstjules/git/mocha/lib/runnable.js:244:7)
      at next (/Users/dstjules/git/mocha/lib/runner.js:259:10)
[...]

Issue #1528, opened 2015-02-03
fix tests that need to be uncommented


dasilvacontin:
Some tests ask to be uncommented to see actual behaviour, because they are designed to fail.

Instead of doing that, another instance of mocha should be run and then test on that instance if the before/after/tests are run/failed/etc. Kinda of what I've done at `test/mocha.js`.

dasilvacontin:
Thanks to @geoguide we've also discovered that async errors are reported as 'Uncaught'. All the tests that 'need to be uncommented to check behaviour' should be improved, IMO.

jbnicolai:
Closed thanks to the great work of @glenjamin in #1781

PR #1781, opened 2015-07-05
Rework hook error tests to actually assert

glenjamin:
The current `hook.err` tests don't actually assert anything, to check that they work you have to uncomment a bunch of lines, and then see if the failing tests are the ones which are supposed to fail.

In particular, this makes it difficult to safely implement #1635

I have converted these tests to a new approach which means they do now actually assert things.

I originally began using the `runMocha` helper along with fixtures to do this, but I found that the tests were extremely opaque as they asserted on the details of a complex out-of-band fixture. To combat this I've done something a bit wacky - defined the fixture inline as a function, and then used `toString` to send that function over the filesystem boundary to generate a fixture on the fly. 

I think the implementation wackiness is worth the expressive benefit here, but I wanted to get some feedback before I add any more tests for the new functionality described in #1635 

As mentioned in the commit message, git isn't smart enough to detect the rewrite of this file, but you can force it to see that the diff doesn't change what is being asserted in any way.
```sh
git diff -w --word-diff HEAD^1:test/hook.err.js HEAD:test/integration/hook.err.js
```


jbnicolai:
> As mentioned in the commit message, git isn't smart enough to detect the rewrite of this file, but you can force it to see that the diff doesn't change what is being asserted in any way.

Yeah, this is a common issue. One way around it is to separate things out into two commits, one with a `git mv` and the other with the actual changes.

jbnicolai:
:tada: this commit is awesome!

glenjamin:
I've split the move and the rewrite into separate commits, that's much neater.

Also related to #1528

test/integration/helpers.js
```diff
@@ -36,6 +37,47 @@ module.exports = {
...
+   *
+   * @param {function} fixture
+   * @param {array}    args
+   * @param {function} fn
+   */
+  runMochaFunction: function(fixture, args, fn) {
+    var path = resolveFixturePath(fixture.name + '.js' || 'tempfile.js');
+    args = args || [];
+
+    var fixtureContent = 'var fn = ' + fixture.toString() + '; fn()';
```
jbnicolai: That's pretty damn (beautifully) hacky ;)

jbnicolai:
> I think the implementation wackiness is worth the expressive benefit here, but I wanted to get some feedback before I add any more tests for the new functionality described in #1635

Agree with both statements here. Definitely worth it though! :smile:

Issue #1732, opened 2015-06-07
Add browser testing to CI


ndhoule:
This probably involves some test refactoring work, but moving to Browserify means it's easy to add a Node shim without really thinking about it, and some of Browserify's shims rely on an ES5 environment. Would be easy to break IE6 without noticing.

boneskull:
I'm more than ready to drop IE6 support

boneskull:
@mochajs/mocha We need automated tests for Mocha which run in the browser.

Anyone have any opinions on the best way to get this done?

When I need to run client-side tests, I reach for [Karma](https://www.npmjs.com/package/karma).  If I need to debug something, I tell it to use a real browser, but otherwise I run headlessly with PhantomJS.

If there are any services out there that would automate the tests against real browsers *for free*, I'd like to know about them.

ndhoule:
I generally go for Sauce Labs to run cross-browser tests. Free for OSS projects.

Karma's fine for most cases but its Sauce runner has [a long-standing concurrency issue](https://github.com/karma-runner/karma-sauce-launcher/issues/40) that has made it difficult for me to run tests against multiple browsers in the past. Might be able to get around this by only running one browser at a time through Karma. I prefer [Zuul](https://github.com/defunctzombie/zuul) these days; some annoying edges but it doesn't involve the same level of configuration Karma does and works OOTB with Browserify.

dasilvacontin:
Dupe of #2079.

PR #1788, opened 2015-07-06
Fix: html-runner crashing because of missing process.stdout shim

sunesimonsen:
Fixes: #1789

When running mocha in the browser you get a crash in the following code:

```js
if (isatty) {
  exports.window.width = process.stdout.getWindowSize
      ? process.stdout.getWindowSize(1)[0]
      : tty.getWindowSize()[1];
}
```

That is because we need to shim the process.stdout before requiring
mocha from browser-entry.js, I fixed that by moving it to the top of the
file.

### DIFF
--- a/support/browser-entry.js
+++ b/support/browser-entry.js
@@ -1,3 +1,9 @@
+/**
+ * Shim process.stdout.
+ */
+
+process.stdout = require('browser-stdout')();
+
 var Mocha = require('../');
 
 /**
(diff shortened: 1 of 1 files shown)


jbnicolai:
Thank you very much. This again stresses the importance of #1732 and manual HTML reporter testing until then.

sunesimonsen:
@jbnicolai happy I could help :-)

Issue #1799, opened 2015-07-14
Using delay in setup() depends on object property ordering


jRiest:
When trying to set `{delay: true}` in a `mocha.setup()` call, it only works if the `delay` property is set before the `ui` property.

Here's some reproduction steps for Chrome.

``` html
<!doctype html>
<html>
  <head>
    <link rel="stylesheet" href="mocha.css" />
  </head>
  <body>
    <div id="mocha"></div>
    <script src="mocha.js"></script>
    <script>
      window.onload = function() {
        // This works
        mocha.setup({delay: true, ui: 'bdd'});

        // This doesn't
        // mocha.setup({ui: 'bdd', delay: true});


        setTimeout(function() {
          describe('a test', function() {
            it('works', function() {});
          });
          run();
        }, 500);
        mocha.run();
      };
    </script>
  </body>
</html>
```

In the case where it doesn't work, the global `run()` method is undefined.

The issue is that [mocha.setup()](https://github.com/mochajs/mocha/blob/a81e5558f086808b682b6d196b80893d2c8926f8/mocha.js#L13143) just iterates over the properties and calls their corresponding methods

``` js
for (var opt in opts) this[opt](opts[opt]);
```

And in the `ui` method, the `pre-require` event is emitted, and it's in those event handlers that the global "run" method is supposed to be registered:

``` js
context.run = mocha.options.delay && common.runWithSuite(suite);
```

However, if the `ui` method is called before the `delay` method, then `mocha.options.delay` will not be set and the global `run` method will not be defined.

jRiest:
@hollomancer I think #1800 should fix this issue and all comments were addressed, but the `pr-needs-work` label was never removed. If there is something else that needs to be done, please let me know and I'll be happy to make the changes.

boneskull:
PR needs review again

jljorgenson18:
Any updates on this? This is still an issue for karma-mocha and it should be an easy fix.

plroebuck:
```js
mocha.setup = function(opts) {
  if (typeof opts === 'string') {
    opts = {ui: opts};
  }

  var optsKeys = Object.keys(opts);
  if (optsKeys.length > 1) {
    var _opts = {};
    // Sort object by keys
    optsKeys.sort().forEach(function(key) {
      _opts[key] = opts[key];
    });
    optsKeys = null;
    opts = _opts;
  }

  for (var opt in opts) {
    if (opts.hasOwnProperty(opt)) {
[...]

PR #1800, opened 2015-07-14
Fix delay option in browser

jRiest:
This PR addresses #1799. It feels a bit weird to special-case the delay option like this, but it looks like that's the only option that's read from `mocha.options` prior to `mocha.run()` being called.

Totally open to better suggestions though.

### DIFF
--- a/support/browser-entry.js
+++ b/support/browser-entry.js
@@ -119,6 +119,9 @@ mocha.ui = function(ui){
 
 mocha.setup = function(opts){
   if ('string' == typeof opts) opts = { ui: opts };
+  if ('delay' in opts) {
+    this.delay(opts.delay);
+  }
   for (var opt in opts) this[opt](opts[opt]);
   return this;
 };
--- a/test/browser/delay.js
+++ b/test/browser/delay.js
@@ -1,0 +1,25 @@
+var delay = 500;
+var start = new Date().getTime();
+
+describe('delayed execution', function() {
+  it('should define a global run function', function() {
+    assert(typeof run === 'function', 'run function is not defined');
+  });
+  it('should have waited ' + delay + 'ms to run this suite', function() {
+    assert(new Date().getTime() - delay >= start, 'did not delay');
+  });
+});
+
+setTimeout(function() {
+  describe('delayed execution', function() {
+    it('should be able to define a suite asynchronously', function() {
+      assert(true);
+    });
+  });
+
+  if (typeof run === 'function') {
+    run();
+  } else {
+    mocha.suite.run();
+  }
+}, delay);
--- a/test/browser/delay.html
+++ b/test/browser/delay.html
@@ -1,0 +1,29 @@
+<html>
+<head>
+    <title>Mocha</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="../../mocha.css" />
+    <script src="../../mocha.js"></script>
+    <script>
+      mocha.setup({
+        ui: 'bdd',
+        delay: true
+      })
+    </script>
+    <script>
+        function assert(expr, msg) {
+            if (!expr) throw new Error(msg);
+        }
+    </script>
+    <script src="delay.js"></script>
+    <script>
+        onload = function() {
+            mocha.run();
+        };
+    </script>
+</head>
+<body>
+<div id="mocha"></div>
+</body>
+</html>


support/browser-entry.js
```diff
@@ -119,6 +119,9 @@ mocha.ui = function(ui){
 
 mocha.setup = function(opts){
   if ('string' == typeof opts) opts = { ui: opts };
+  if (Object.prototype.hasOwnProperty.call(opts, 'delay')) {
```
danielstjules: maybe just ```if ('delay' in opts)``` ?

danielstjules:
Thanks for the PR, and appreciate the specs! :) While it might be odd to have a special case, it's pretty understandable that some order could be implied in these invocations. That `for` loop just seems like a convenient and terse shortcut.

test/browser/delay.js
```diff
@@ -0,0 +1,18 @@
...
+setTimeout(function() {
+  describe('delayed execution', function() {
+    it('should have waited ' + delay + 'ms to run this suite', function() {
+      assert(new Date().getTime() - delay >= start);
+    });
+
+    it('should have no effect if attempted twice in the same suite', function() {
+      assert(true);
+      run();
+      assert(true);
```
danielstjules: Maybe this spec could be updated? Rather than the two ```assert(true)```, we could have another assertion on the time, ensuring that some number less than delay has elapsed since (start + delay)

jRiest:
Cool. Made those changes. Do the browser tests get run as part of `npm test`? It doesn't seem like it, but I might be missing something. I tried just opening the delay.html file in a browser though and the tests pass.

danielstjules:
Not right now. I've got a local branch I started that will run them with phantomjs, so what you have is perfect!

danielstjules:
@jRiest is it possible that the test file broke after the change? I'm getting
```
passes: 0failures: 0duration: 0s
```

jRiest:
Hmm, that's strange. It seems to still be working for me when I just open up __delay.html__ in my browser. I did have to run `make all` in order to re-build the __mocha.js__ in the root of the repo, since it looks like all of the web tests use that file.

I assumed I wasn't supposed to commit updates to the built file, is that correct?

jRiest:
Although I guess your comment raises a concern that if the delay functionality did break, the tests may just never be run and we may never know it broke.

I suppose if I move the "should have waited..." test outside of the setTimeout that would catch that error. I'll update the PR.

danielstjules:
> I assumed I wasn't supposed to commit updates to the built file, is that correct?

That's right. My mistake - I must have checked out HEAD after rebuilding mocha.js without paying attention.



> I suppose if I move the "should have waited..." test outside of the setTimeout that would catch that error. I'll update the PR.

That would be perfect :) It would be great if the test failed when the delay didn't correctly run,

natlibfi-arlehiko:
@jRiest what's the status of updating this PR? We also need this fix. I can look into it too, if needed.

jRiest:
Sorry for dropping this. We ended up not needing it so it slipped my mind. Just pushed a change that should make the tests fail without this fix.

It's a little strange since the original issue was that the `run` function to start the tests wasn't defined so the tests didn't run at all if the delay option was enabled. In order to make sure the tests ran and failed appropriately without this fix, I just manually invoke `mocha.suite.run()` if the run function isn't defined. Totally open to suggestions if there's a better option though.

natlibfi-arlehiko:
@danielstjules, is this mergeable now?

hollomancer:
This is a valid issue. @natlibfi-arlehiko, @jRiest - closing this for now due to age, added pr-please to #1799 so this can be updated.

jRiest:
@hollomancer I think this PR has addressed all of the comments. Is there something else that needs to be done before it can be merged?

boneskull:
@jRiest Thanks.  It wasn't your fault this was closed; sorry about that.

I have some comments if you are willing to work on this further:

- `Mocha.prototype.delay()` accepts no parameters; if it is called, the option is enabled, so,
- calling `mocha.setup({delay: false})` here would actually *enable* the delay
- `test/browser/**` is not run by CI--it's unclear if we can even do that--but tests in this directory are of marginal value as they stand.

A good test for this would actually run Mocha (via [karma-mocha](https://www.npmjs.com/package/karma-mocha)) with a `delay` option, which would then attempt to execute a test. The test run should fail if a test was not executed.  You'd want to add probably a new target in the `Makefile` which would run Karma with an environment variable present; then `karma.conf.js` would need to pick this up, set the appropriate option(s) and execute (only) the test file.

So I think, perhaps, if you're unable or unwilling to embark on that, then modifying the code in `browser-entry.js` to assert truthiness of the `delay` option would be good enough to merge, as this is a low-risk change.

ScottFreeCode:
> `test/browser/**` is not run by CI--it's unclear if we can even do that--but tests in this directory are of marginal value as they stand.

Tests in `test/browser-unit/` should be run once #2900 is merged, but in the case of using specific `mocha.setup` options I'm not sure that will cut it -- I'm pretty sure we would need to come up with a way to run browser integration tests (i.e. do the `mocha.setup` and `mocha.run` type stuff, run each file in a separate instance of Karma if using Karma for them)...

PR #1814, opened 2015-07-26
Fixes tests in windows (fixed #1813)

### LINKED ISSUE #1813: Tests are broken in Windows
Having read the contributing guidelines, I realized that on a windows machine I could not confirm whether any changes I committed failed tests or not because many of the tests are failing in windows.

Summary of issues and potential fixes: 

- I wasn't able to get the windows make to work properly at all.  using cygwin seems to works though.

- many references to '/tmp' inside acceptance fs/utils - these can be updated to use os.tmpdir()

- passing `bin/mocha` to exec fails,  "bin is not recognized as an internal or external command" -- updating it to `node bin/mocha` makes it work.

- using `spawn` in integration/helpers doesn't work at all.  Updating to use `exec`, passing args delimited by a space works. (also need to preface with `node`, similar to above)

The trickiest one relates to the dot symbol used by windows --

Integration/hook tests takes the output of test results and splits by new lines and the dot symbol.  Thing is, windows doesn't use the dot symbol (\u2024) - [it uses a full stop](https://github.com/mochajs/mocha/blob/master/lib/reporters/base.js#L88), so the tests fail because the split doesn't work, and a full stop winds up in some of the titles.  Updating the [regexp](https://github.com/mochajs/mocha/blob/master/test/integration/hook.err.js#L340) to use a dot in addition to \u2024 makes it fail in a different way (similar to how the tests would fail if the test titles included \u2024).

I think the best way to fix this is to first, change the test to use whatever base reporter exports for the dot symbol, and second, change the windows dot to use \u00B7.   \u2024 would work, but using the default cmd font causes a beep every time it is used (which gets exceedingly annoying) - \u00B7 looks pretty close to the dot character, is supported by cmd and likely wouldn't be used in test titles.

I have fixes ready for these and will send a PR...  I'm looking for confirmation that these changes (a) are welcome, (b) wouldn't break the tests when run in a unix environment (unsure if spawn vs. exec produces different results and whether prefacing the command with `node` causes issues)

tswaters:


jbnicolai:
:+1:  thanks, I'll review asap :)

TimothyGu:
I fixed a few additional things and rebased the commit in #2270.

boneskull:
@TimothyGu thanks for this!  we will need to rebase again tho

TimothyGu:
@boneskull, I checked my PR, and GitHub says that there are no conflicts. But nevertheless I rebased and force pushed just to make sure everything works.

dasilvacontin:
Closing in favour of #2270 – Thanks!

PR #1827, opened 2015-08-11
Set safe CSS for progress indicator which does not conflict with Bootstrap

irnc:
This is not the best solution for a problem, mocha should not know anything about Bootstrap.

Proposed change would result in less conflicts. However it is made just to get the ball rolling.

In order to prevent such conflicts from happening, it is a good idea to avoid nesting in mochas CSS, e.g. by adapting BEM naming convention (see https://css-tricks.com/bem-101/).

Giving mocha elements classes named `mocha-stats-progress` will eliminate conflicts with other libraries entirely.

### DIFF
--- a/mocha.css
+++ b/mocha.css
@@ -261,6 +261,15 @@ body {
 #mocha-stats .progress {
   float: right;
   padding-top: 0;
+  
+  /**
+   * Set safe initial values, so mochas .progress does not inherit these
+   * properties from Bootstrap .progress (which causes .progress height to
+   * equal line height set in Bootstrap).
+   */
+  height: auto;
+  box-shadow: none;
+  background-color: initial;
 }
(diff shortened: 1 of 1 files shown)


dasilvacontin:
LGTM. Yeah, we should really better scope those stylings, thanks for pointing it out.

danielstjules:
> Giving mocha elements classes named mocha-stats-progress will eliminate conflicts with other libraries entirely.

Might be worth adopting BEM then? Rather than applying small fixes to individual selectors.

Turbo87:
any news on this? I just ran into the exact same problem...

danielstjules:
Will merge for now, but eventually all mocha CSS should just adopt BEM or a similar convention to avoid these issues.

PR #1829, opened 2015-08-12
Support --max-old-space-size node argument

gigadude:
This allows setting the heap size for tests which require it.


boneskull:
@gigadude thanks

pretend I complained about manually managing these flags

PR #1852, opened 2015-09-02
Fix: Bail flag causes before() hooks to be run even after a failure

### LINKED ISSUE #1812: Bail flag causes before() hooks to be run even after a failure
Example file:

```javascript
describe('one', function(){
  it('fails', function(){
    throw Error("failed");
  });
});

var slowHook = function(next) {
  console.log("hook");
  setTimeout(next, 1000);
};

describe('two', function(){
  before(slowHook);
  it('never runs', function(){ });
});

describe('three', function(){
  before(slowHook);
  it('never runs', function(){ });
});

describe('four', function(){
  before(slowHook);
  it('never runs', function(){ });
});
```

So what I'm seeing is that when you run this with `--bail` and the first test fails, it keeps going and runs all the other hooks before actually quitting. The output looks like this:

```
$ mocha --version
2.2.5
$ mocha timeout.js -b


  one
    1) fails

  two
hook

  three
hook

  four
hook


  0 passing (3s)
  1 failing

  1) one fails:
     Error: failed
      at Error (<anonymous>)
      at Context.<anonymous> (timeout.js:3:11)
```

In our actual code base we have hundreds of test suites each with sometimes lengthy hooks that set up and tear down fixtures in the db. When you run the whole suite with `-b`, you might see one red dot, and then it appears to hang for over a minute while it is running all the other hooks. Finally it reports with the failure. I can't think of why this would be desirable, so my best guess is that it is a bug.

aaroncrows:
Fixes #1812

Added a bail check to the runSuite method in runner.js. There was already a commented out check for a subsequent before block in the bail fixture so I just uncommented that for the test.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -569,7 +569,7 @@ Runner.prototype.runSuite = function(suite, fn) {
 
   debug('run suite %s', suite.fullTitle());
 
-  if (!total) {
+  if (!total || (self.failures && suite._bail)) {
     return fn();
   }
 
--- a/test/integration/fixtures/options/bail.js
+++ b/test/integration/fixtures/options/bail.js
@@ -11,13 +11,9 @@ describe('suite1', function() {
 });
 
 describe('suite2', function() {
-  // TODO: When uncommented, the hook below is ran and throws an exception
-  // despite the previous failure with the bail flag. This is a bug. Uncomment
-  // once resolved
-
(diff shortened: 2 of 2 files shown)


danielstjules:
Good catch :) That's been an outstanding `TODO` for a while now

dasilvacontin:
Thanks! LGTM

PR #1862, opened 2015-09-03
Fix stack trace display in FireFox and Safari

mbovel:
Hi!

I encountered three problems using mocha in the browser with FireFox 40 and Safari 8 on Mac OS X:
1. There is no line break before the stack trace.
2. The track trace is not indented.
3. More subtle: the stack trace is wrongly trimmed if the error message is contained in the stack trace.

My solution looks like solving these three problems in FireFox 40 and Safari 8 and also works in Chrome 44, but I haven't tested it on any other plateforme nor did I test it in complexe situations (htmlMessage for exemple, see d71a5c6e0f17ab5b530388edf5b8cfb71ed75d33).

@jbnicolai Are there cases where `test.err.message` isn't at the beginning of the stack trace and where using `indexOf` is really useful?

Screenshot before:

Screenshot before

Screenshot after:

Screenshot afte

Best regards,

Matt

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -166,11 +166,10 @@ function HTML(runner) {
       }
 
       if (test.err.stack) {
-        var indexOfMessage = test.err.stack.indexOf(test.err.message);
-        if (indexOfMessage === -1) {
-          stackString = test.err.stack;
+        if (test.err.stack.substr(0, message.length) === message) {
+          stackString = test.err.stack.substring(message.length, test.err.stack.length);
         } else {
-          stackString = test.err.stack.substr(test.err.message.length + indexOfMessage);
+          stackString = '\n' + test.err.stack.replace(/^/gm, '    ');
         }
       } else if (test.err.sourceURL && test.err.line !== undefined) {
         // Safari doesn't give you a stack. Let's at least provide a source line.


danielstjules:
Mind squashing your commits? :) Taking a look right now

danielstjules:
> Are there cases where test.err.message isn't at the beginning of the stack trace and where using indexOf is really useful?

Some frameworks throw custom objects with their own `stack` property, which isn't necessarily consistent with the format you'd expect from an Error. It's entirely possible that the err.message would not be at the start of the stack trace. It's an edge case, but we sure do hit a lot of those :) Guaranteed someone's done it haha Imagine something like:

``` javascript
try {
  fn();
} catch (e) {
  throw {
    message: e.message,
    stack: 'Framework name internal error\n' + e.stack
  }
}

mbovel:
Ok, I see.

Using `indexOf`, I found no way to solve the third problem I was talking about (stack trace wrongly trimmed when the error message is contained in the stack trace), but I guess it's maybe not that important.

So, what about just changing `test.err.stack` to `stackString = '\n' + test.err.stack.replace(/^/gm, '    ')` intead of the changes I proposed? That would at least fix the line break and the indentation problems.

DarthCharles:
Nice work around, seems to be working fine. +1

boneskull:
too old, no tests

PR #1871, opened 2015-09-08
IE<=8 no [].reduce, so use 'utils.reduce' instead

wsw0108:


danielstjules:
Merged via https://github.com/mochajs/mocha/commit/0565fac49e4153cdc65a6431553d1aae80eb7008 We usually only compile mocha.js as part of releases. Thanks!

PR #1767, opened 2015-06-28
Fix 1766: stackfilter should not ignore node_modules

danielstjules:
Fix for https://github.com/mochajs/mocha/issues/1766

Issue #1766, opened 2015-06-27
Remove 3rd party function calls from stack trace only when earlier than user code


moll:
Hey,

Shortening of stack traces is great, but it shouldn't suppress the exception's source from 3rd party (stuff in `node_modules`) function calls that come _after_ user code. The ones leading up to user code are definitely useless (like two dozens of Mocha internal calls), but hiding the the path from user code to the possibly faulty library makes it a little difficult to discern the problem.

Case in point, I had a problem in a module:
```
      at normalizeValue (/Users/user/node_modules/whatwg-fetch/fetch.js:20:20)
      at Headers.append (/Users/user/node_modules/whatwg-fetch/fetch.js:45:13)
      at /Users/user/node_modules/whatwg-fetch/fetch.js:38:14
      at Array.forEach (native)
      at new Headers (/Users/user/node_modules/whatwg-fetch/fetch.js:37:43)
      at new Request (/Users/user/node_modules/whatwg-fetch/fetch.js:211:20)
      at self.fetch (/Users/user/node_modules/whatwg-fetch/fetch.js:276:17)
      at /Users/user/fetch.js:9:10
      ...
```

With short stack traces on, the only lines visible were:
```
      at Array.forEach (native)
      at /Users/user/fetch.js:9:10
```
Impossible to decipher. ;-)

danielstjules:
I think that's a good point! Easy fix is to drop https://github.com/mochajs/mocha/blob/7493bca76662318183e55294e906a4107433e20e/lib/utils.js#L695 from that bool exp

jbnicolai:
I think I agree. @mochajs/mocha anyone else who has a strong oppinion on this?

boneskull:
:+1:

dasilvacontin:
:+1: too.

PR #1880, opened 2015-09-09
Stackfilter fix: Don't remove modules/components from stack trace in the browser

danielstjules:
Followup to https://github.com/mochajs/mocha/issues/1766 and https://github.com/mochajs/mocha/pull/1767

Makes the behavior consistent in the browser. We now only filter out mocha's own function calls, and don't touch calls from dependencies. Given how modularity is embraced in the node community, I think this is ideal.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -689,7 +689,7 @@ exports.getError = function(err) {
  * @description
  * When invoking this function you get a filter function that get the Error.stack as an input,
  * and return a prettify output.
- * (i.e: strip Mocha, node_modules, bower and componentJS from stack trace).
+ * (i.e: strip Mocha and internal node functions from stack trace).
  * @returns {Function}
  */
 exports.stackTraceFilter = function() {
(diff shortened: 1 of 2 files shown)


aaronjensen:
This is a good default, though it'd be nice if there was an option that came w/ this. I can imagine chai, other matchers, react, etc all muddying up stack traces.

danielstjules:
Agreed - assertion libraries will get noisy as well. :) But this just makes things consistent across node/browser envs.

As for configuration, that's where things like github.com/rstacruz/mocha-clean can come in handy. It doesn't _need_ be in mocha's core. Though I wouldn't be opposed to it.

aaronjensen:
Oh, didn't know there was a mocha-clean. I'm good w/ it being external, though maybe a link in the readme/docs if it's not there already

ArtemGovorov:
Didn't know about mocha-clean either, thanks for the hint. + 1 to @aaronjensen - a link in readme would be very nice to have.

danielstjules:
I guess it could be added to the wiki :) https://github.com/mochajs/mocha/wiki

danielstjules:
+1? @boneskull

dasilvacontin:
Sweet, thanks @danielstjules!

PR #1882, opened 2015-09-10
Fix fragile xunit reporter spec

danielstjules:
After a few dozen jobs on travis, I noticed a single failure from this test. Looks like the root cause is that the time can exceed 0.


danielstjules:
Was able to reproduce :)

``` javascript
<testsuite name="Mocha Tests" tests="2" failures="0" errors="0" skipped="0" timestamp="Fri, 11 Sep 2015 04:50:33 GMT" time="0.004">
<testcase classname="suite" name="test1" time="0.001"/>
<testcase classname="suite" name="test2" time="0"/>
</testsuite>
```

Merging this in to avoid any additional failures. Sorry!

PR #1889, opened 2015-09-13
Fix #1798: Correctly attribute mutiple done err with hooks

### LINKED ISSUE #1798: mocha loses the right title when done called multiple times
I just finished trouble-shooting a wierd "done called multiple times" message.... phewee.. here's a tricky one:

I had a before-all hook that looked as following:
```
{ "lib/rest" : 
  { beforeAll:
    function(done) {
        this.slow("1.1s");
        rest = require('../lib/rest')(sockets, eventsHandler, done);
        setTimeout(done, 1000)
    }
  , 
```
Where evidently, the done is called both from `../lib/rest` and the `setTimeout`.
Anyway, the `../lib/rest` was not calling the callback for a long time, and evidently, making it call the callback as it should - was actually what broke the build.

(I'm using [mocha-ui-exports](https://github.com/osher/mocha-ui-exports), that's why it looks like that)

However, mocha could not direct me to the right place - probably because the `setTimeout` handler that was fired after 1s - was called in the midst of some other suite that followed - and mocha was pointing me to the suite and the test that during it's run the timer was fired...

looking there obviously brought nothing, and pending the tests that were reported as failing just moved the failure to another test that was not pended - and that's where I started suspecting that mocha reports the wrong place, and expanded my search.

After fixing the issue (removing the `setTimeout` from the `beforeAll` hook) - the suite passes.

danielstjules:
``` javascript
$ cat test.js
  before(function(done) {
    setTimeout(done, 30);
    setTimeout(done, 60);
  });

  it('test', function(done) {
    setTimeout(done, 100);
  });
});

// -------------------------------------------------------------
// Before PR, the error is incorrectly attributed to the spec,
// rather than the before hook. Furthermore, the error was
// uncaught, so that suite would stop
// -------------------------------------------------------------
$ mocha test.js


  suite
    1) test


  0 passing (67ms)
  1 failing

  1) suite test:
     Uncaught Error: done() called multiple times

// -------------------------------------------------------------
// After the PR, the error is correctly attributed to the hook
// -------------------------------------------------------------
$ mocha test.js


  ․․

  1 passing (141ms)
  1 failing

  1) suite "before all" hook:
```

### DIFF
--- a/test/integration/multiple.done.js
+++ b/test/integration/multiple.done.js
@@ -4,25 +4,95 @@ var args   = [];
 
 describe('multiple calls to done()', function() {
   var res;
-
   this.timeout(1000);
 
-  before(function(done) {
-    run('multiple.done.js', args, function(err, result) {
-      res = result;
-      done(err);
+  describe('from a spec', function() {
+    before(function(done) {
+      run('multiple.done.js', args, function(err, result) {
+        res = result;
+        done(err);
+      });
+    });
+
+    it('results in failures', function() {
+      assert.equal(res.stats.pending, 0);
+      assert.equal(res.stats.passes, 1);
+      assert.equal(res.stats.failures, 1);
+      assert.equal(res.code, 1);
+    });
+
+    it('throws a descriptive error', function() {
+      assert.equal(res.failures[0].err.message,
+        'done() called multiple times');
     });
   });
 
-  it('results in failures', function() {
-    assert.equal(res.stats.pending, 0);
-    assert.equal(res.stats.passes, 1);
-    assert.equal(res.stats.failures, 1);
-    assert.equal(res.code, 1);
+  describe('with multiple specs', function() {
+    before(function(done) {
+      run('multiple.done.specs.js', args, function(err, result) {
+        res = result;
+        done(err);
+      });
+    });
+
+    it('results in a failure', function() {
+      assert.equal(res.stats.pending, 0);
+      assert.equal(res.stats.passes, 2);
+      assert.equal(res.stats.failures, 1);
+      assert.equal(res.code, 1);
+    });
+
+    it('correctly attributes the error', function() {
+      assert.equal(res.failures[0].fullTitle, 'suite test1');
+      assert.equal(res.failures[0].err.message,
+        'done() called multiple times');
+    });
   });
 
-  it('throws a descriptive error', function() {
-    assert.equal(res.failures[0].err.message,
-      'done() called multiple times');
+  describe('from a before hook', function() {
+    before(function(done) {
+      run('multiple.done.before.js', args, function(err, result) {
+        res = result;
+        done(err);
+      });
+    });
+
+    it('results in a failure', function() {
+      assert.equal(res.stats.pending, 0);
+      assert.equal(res.stats.passes, 1);
+      assert.equal(res.stats.failures, 1);
+      assert.equal(res.code, 1);
+    });
+
+    it('correctly attributes the error', function() {
+      assert.equal(res.failures[0].fullTitle, 'suite "before all" hook');
+      assert.equal(res.failures[0].err.message,
+        'done() called multiple times');
+    });
+  });
+
+  describe('from a beforeEach hook', function() {
+    before(function(done) {
+      run('multiple.done.beforeEach.js', args, function(err, result) {
+        res = result;
+        done(err);
+      });
+    });
+
+    it('results in a failure', function() {
+      assert.equal(res.stats.pending, 0);
+      assert.equal(res.stats.passes, 2);
+      assert.equal(res.stats.failures, 2);
+      assert.equal(res.code, 2);
+    });
+
+    it('correctly attributes the errors', function() {
+      assert.equal(res.failures.length, 2);
+      res.failures.forEach(function(failure) {
+        assert.equal(failure.fullTitle, 'suite "before each" hook');
+        assert.equal(failure.err.message,
+          'done() called multiple times');
+      });
+    });
   });
 });
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -288,12 +288,13 @@ Runner.prototype.hook = function(name, fn) {
 
     self.emit('hook', hook);
 
-    hook.on('error', function(err) {
-      self.failHook(hook, err);
-    });
+    if (!hook.listeners('error').length) {
+      hook.on('error', function(err) {
+        self.failHook(hook, err);
+      });
+    }
 
     hook.run(function(err) {
-      hook.removeAllListeners('error');
       var testError = hook.error();
       if (testError) {
         self.fail(self.test, testError);
--- a/test/integration/fixtures/multiple.done.before.js
+++ b/test/integration/fixtures/multiple.done.before.js
@@ -1,0 +1,10 @@
+describe('suite', function() {
+  before(function(done) {
+    setTimeout(done, 10);
+    setTimeout(done, 30);
+  });
+
+  it('test1', function(done) {
+    setTimeout(done, 50);
+  });
+});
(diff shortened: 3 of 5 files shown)


danielstjules:
setMaxListeners is to avoid some noise:
```
(node) warning: possible EventEmitter memory leak detected. 11 error listeners added. Use emitter.setMaxListeners() to increase limit.
```

I think this is preferable to the original solution: https://github.com/danielstjules/mocha/commit/e673963057eb11cd0ec95935b1d163a69820ceda

dasilvacontin:
So hooks will end up having as many 'error' listeners as tests are run*? (in case of a 'beforeEach)

I don't even get why we set up the 'error' listener once each time before we run the hook. Why not set them up before running the suites and remove them after the suites are done/finished? The callback is always the same, afaik.

dasilvacontin:
Yep, I just checked, and the "multiple done" error is the only 'error' event Runnable emits.

danielstjules:
Just realized I could instead move `hook.removeAllListeners('error');` which does a better job with beforeEach

danielstjules:
> Why not set them up before running the suites and remove them after the suites are done/finished?

Because they could still throw after the suite finished, during a different suite's execution

danielstjules:
> So hooks will end up having as many 'error' listeners as tests are run*? (in case of a 'beforeEach)

Updated code will have the same number of listeners as what's currently in master, except it won't remove the listeners once hook execution has completed (preventing the uncaught error issue)

dasilvacontin:
Weird, locally, tests are stopping after 'acceptance/timeout.js'. Works fine on Travis.

danielstjules:
Does that happen on master for you as well? Or only this branch?

danielstjules:
+1? :)

dasilvacontin:
It does happen in master as well.. weird.
```
➜  mocha git:(master) node -v
v0.12.7
```

test/integration/multiple.done.js
```diff
@@ -4,25 +4,95 @@ var args   = [];
...
+    before(function(done) {
+      run('multiple.done.before.js', args, function(err, result) {
+        res = result;
+        done(err);
+      });
+    });
+
+    it('results in a failure', function() {
+      assert.equal(res.stats.pending, 0);
+      assert.equal(res.stats.passes, 1);
```
dasilvacontin: Why is `passes` 1?

```
➜  mocha git:(danielstjules-issue-1798) ✗ mocha test/integration/fixtures/multiple.done.before.js          

  ․

  0 passing (41ms)
  1 failing

  1) suite test1:
     Uncaught Error: done() called multiple times
      at Suite.<anonymous> (test/integration/fixtures/multiple.done.before.js:2:3)
      at Object.<anonymous> (test/integration/fixtures/multiple.done.before.js:1:63)
      at Array.forEach (native)
      at node.js:814:3
```
(same thing for most other tests)

dasilvacontin: More clear even:
```
➜  mocha git:(danielstjules-issue-1798) ✗ mocha test/integration/fixtures/multiple.done.before.js --reporter JSON
{
  "stats": {
    "suites": 1,
    "tests": 1,
    "passes": 0,
    "pending": 0,
    "failures": 1,
    "start": "2015-09-14T11:33:37.794Z",
    "end": "2015-09-14T11:33:37.834Z",
    "duration": 40
  },
```

dasilvacontin:
> Because they could still emit an error after the suite finished, during a different suite's execution. Without leaving those listeners active, that'd result in an uncaught error

Okay, then setting them up once and not removing them? How does that sound?

As a proof of concept, this passes your tests:
```
diff --git a/lib/runner.js b/lib/runner.js
index 1c91712..09cd414 100644
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -285,13 +285,14 @@ Runner.prototype.hook = function(name, fn) {
     self.currentRunnable = hook;
 
     hook.ctx.currentTest = self.test;
-    hook.removeAllListeners('error');
 
     self.emit('hook', hook);
 
-    hook.on('error', function(err) {
-      self.failHook(hook, err);
-    });
+    if (EventEmitter.listenerCount(hook, 'error') === 0) {
+      hook.on('error', function(err) {
+        self.failHook(hook, err);
+      });
+    }
 
     hook.run(function(err) {
       var testError = hook.error();
```

test/integration/multiple.done.js
danielstjules: > Why is passes 1?

You're using the wrong bin. Use the project's bin, rather than global mocha.
``` javascript
danielstjules:~/git/mocha (issue-1798)
$ ./bin/mocha test/integration/fixtures/multiple.done.before.js --reporter JSON
{
  "stats": {
    "suites": 1,
    "tests": 1,
    "passes": 1,
    "pending": 0,
    "failures": 1,
    "start": "2015-09-14T16:28:29.589Z",
    "end": "2015-09-14T16:28:29.658Z",
    "duration": 69
  },
```

danielstjules:
Sure, that works :)

danielstjules:
Updated the branch with that change

test/integration/multiple.done.js
danielstjules: Since those hook errors were previously uncaught, the suite would exit. Now that they can be handled by the error listener, a given suite will continue to run. That's why we go from 0 to 1 passes.

danielstjules:
Used hook.listeners for node 0.8 support https://nodejs.org/docs/v0.8.0/api/events.html#events_emitter_listeners_event

test/integration/multiple.done.js
dasilvacontin: I see. Since (at least in this case) we know that the before hook has failed before the test has finished running, couldn't we ignore the outcome of the test?

dasilvacontin: I mean, we obviously could ignore it, but what do you think about doing so?

danielstjules: > I mean, we obviously could ignore it, but what do you think about doing so?

I don't think it's ideal. I think that leads to the same confusing output we see in scenarios like:

``` javascript
$ cat test.js
beforeEach(function(done) {
  done(new Error('error'));
});

it('test1', function(done) {
  setTimeout(done, 20);
});

it('test1', function(done) {
  setTimeout(done, 20);
});

$ mocha test.js


  1) "before each" hook for "test1"

  0 passing (11ms)
  1 failing

  1)  "before each" hook for "test1":
     Error: error
      at Context.<anonymous> (test.js:2:8)
```

danielstjules: I'd also rather keep the scope of this PR small and push for a patch release, rather than make any major changes to runner behavior

danielstjules:
@dasilvacontin @boneskull anymore feedback? :) Would love to merge this in

test/integration/multiple.done.js
dasilvacontin: I agree to keep the scope of the PR small, I'd like to discuss about it in a issue then. :)

dasilvacontin:
> Used hook.listeners for node 0.8 support https://nodejs.org/docs/v0.8.0/api/events.html#events_emitter_listeners_event

Cool, I hadn't seen that one.

Well, that was just for showing it worked. What about setting up the hook listeners in a loop before running the suite? It'd be cleaner imo. Something similar to this, but extracted in a nice documented function. (this also passes tests)

```diff
@@ -630,6 +624,17 @@ Runner.prototype.runSuite = function(suite, fn) {
 
   this.nextSuite = next;
 
+  var hookNames = ['beforeAll', 'beforeEach', 'afterEach', 'afterAll']
+  hookNames.forEach(function (hookName) {
+    var hooks = self.suite['_' + hookName]
+    if (!hooks.length) return
+    hooks.forEach(function (hook) {
+      hook.on('error', function (err) {
+        self.failHook(hook, err)
+      })
+    })
+  })
+
   this.hook('beforeAll', function(err) {
     if (err) {
       return done();
```

danielstjules:
Sounds like a later PR could address that if it's just refactoring :)

dasilvacontin:
Just refactoring. :)

LGTM then, good job!

PR #1897, opened 2015-09-18
Remove unused packages

woodyrew:
As per: https://www.bithound.io/github/mochajs/mocha/master

The following packages are unused:
- commander
- mkdirp

### DIFF
--- a/package.json
+++ b/package.json
@@ -274,14 +274,12 @@
     "test": "make test-all"
   },
   "dependencies": {
-    "commander": "2.3.0",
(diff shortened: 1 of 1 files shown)


danielstjules:
That's not true. It's ignoring the bin dir:
https://github.com/mochajs/mocha/blob/master/bin/_mocha#L7
https://github.com/mochajs/mocha/blob/4747437b042427fdeb56848608e3f1ec3e4b3586/bin/_mocha#L109

Issue #1846, opened 2015-08-26
Harmony flags


danielstjules:
Rather than exposing individual node/iojs harmony flags, what if we were to expose only a single `--harmony` flag that enabled all harmony features for that environment?

jkresner:
This is a bad idea. The harmony flag has different behaviors across node versions.

jkresner:
What about a feature that folks and do --node-anything and mocha attempts to append anything when invoking node?

danielstjules:
At the very least, we could be smart about harmony and es_staging. Either way, all these options shouldn't have to be handled by mocha anymore

jkresner:
Should have some sort of interim solution soon though. Gonna have a swarm of people moving off traceur and babel and their code won't run on mocha.

danielstjules:
> Should have some sort of interim solution soon though.

Agreed. I don't think we'll be releasing a patch within the next day though, so happy to discuss an alternative :)

danielstjules:
Automatically passing along any arguments that start with `--harmony` would be a good start I think

jkresner:
Since mocha runs on more than just node, I think it's not it's responsibility to intercept and interpret. My gut says, let people just node args in some format and their own environment do the work.

danielstjules:
As long as we're handling unrecognized flags, I'd be happy with just passing things along. Looks like could just piggyback off node's own handling for node > 0.10 :) Looks like with node 0.10 and 0.8, it'll still run and exit with a 0 error code, which isn't what we'd want.

```
$ cat test.js
console.log('testing');

$ nvm use 0.12
Now using node v0.12.0

# Positive error code, perfect
$ node --invalid test.js
node: bad option: --invalid

$ nvm use 0.10
Now using node v0.10.20

# 0 error code, still runs :(
$ node --invalid test.js
Error: unrecognized flag --invalid
Try --help for options
testing
```

danielstjules:
A more complex solution would be to parse `node --help` and `node --v8-options` to ensure that it's a valid flag. That would work for node 0.10 as well. But maybe we're fine with the behavior I highlighted in https://github.com/mochajs/mocha/issues/1846#issuecomment-139113704

jkresner:
I just don't think you want to be responsible for maintaining logic outside the scope of what a libraries real purpose is.

I wouldn't let people pass anything to mocha, but something like

````
mocha index.js --node-harmony_destructuring
````

[...]

PR #1900, opened 2015-09-21
lazy/nolazy v8 switch

moander:


danielstjules:
I'd like to avoid adding options like this. Otherwise we risk the same problem we were having with the harmony flags. With node v4, I see over 300 v8 options. Related: https://github.com/mochajs/mocha/issues/1846  (though it only mentioned harmony flags)

moander:
It's related to the exposed debug-brk / profiler flags, so I think nolazy should be exposed as well.

danielstjules:
I agree, would just prefer a solution that accommodates all v8 options rather than slowly building up to a 300+ line long switch.

moander:
Would it be an acceptable solution to pass all unknown options down to node?

danielstjules:
https://github.com/mochajs/mocha/issues/1846#issuecomment-139113704 I think that'd work as a temporary solution :)

PR #1905, opened 2015-09-25
emit require/post-require hooks in browser

jasonkarns:
In node, the `Mocha.prototype.loadFiles` method [emits callback events](https://github.com/mochajs/mocha/blob/c4393c456839d6bf2cbb4abb1cd177010ee06458/lib/mocha.js#L215-L217) (`pre-require`,`require`, and `post-require`) around the require of each test file.

These callbacks are the hooks by which custom UIs register their APIs and do their magic. In the browser, no such files are loaded. The browser entry file overrides `Mocha.prototype.ui` to ensure the `pre-require` hook is [still invoked](https://github.com/mochajs/mocha/blob/c4393c456839d6bf2cbb4abb1cd177010ee06458/support/browser-entry.js#L112).

Thus, when a UI is set, the custom UI has a chance to register its API. Up until now, however, the `require` and `post-require` hooks haven't been invoked at all.

In node, the 3 hooks are invoked [right after one another](https://github.com/mochajs/mocha/blob/c4393c456839d6bf2cbb4abb1cd177010ee06458/lib/mocha.js#L453) inside the `loadFiles` method which is invoked by the `run` method.

In node, the `require`/`post-require` hooks are invoked after the file is required as part of a `run`. In the browser, `run` is already guaranteed to be invoked _after_ the tests have been loaded, thus the constraint that the tests be loaded prior to `require`/`post-require` hooks still holds. And they are invoked as part of the `run` flow which is the same as under node.

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -136,6 +136,10 @@ mocha.run = function(fn){
   if (query.fgrep) mocha.grep(query.fgrep);
   if (query.invert) mocha.invert();
 
+  // require/post-require events usually emitted in Mocha.prototype.loadFiles
+  this.suite.emit('require', global, null, this);
+  this.suite.emit('post-require', global, null, this);
+
   return Mocha.prototype.run.call(mocha, function(err){
     // The DOM Document is not available in Web Workers.
     var document = global.document;


boneskull:
I'm not sure I understand this. Is it simply to make things work the same in both environments, or is there a use case here that I missed?

jasonkarns:
Yes, it's to make it work the same in both environments. Specifically, the mocha-gwt UI does **not** work in the browser precisely because the `require` and `post-require` hooks aren't invoked in the browser. The only UIs that *do* work in the browser, do so because they only register the `pre-require` hook.

jasonkarns:
bump.

still blocked using custom UIs in both node and browser because of this.

boneskull:
@jasonkarns Can you resolve conflicts please?

jasonkarns:
resolved

boneskull:
too old; reopen if still interested

jasonkarns:
This is still an issue and needs reopened. The lack of these hooks makes it difficult or impossible for some kinds of UIs to work in browser at all.

jasonkarns:
Branch has been rebased to latest master.

boneskull:
OK. this will need some sort of test.  open to suggestions on best way to do that, as we don't have much in the way of integration tests which run in the browser.

boneskull:
I can't reopen this, so it might necessitate a new PR. "The browser-post-require branch was force-pushed or recreated"

jasonkarns:
@boneskull I've restored the old branch pre-rebase so the PR can be reopened. Once reopened, I'll switch back to the current on (which has been rebased on latest master).

PR #1911, opened 2015-09-30
Use process.browser in stackTraceFilter

### LINKED ISSUE #1764: Type check window instead of document to detect environment

berkerpeksag:
Closes #1764

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -693,10 +693,8 @@ exports.getError = function(err) {
  * @returns {Function}
  */
 exports.stackTraceFilter = function() {
-  // TODO: Replace with `process.browser`
   var slash = '/';
-  var is = typeof document === 'undefined' ? { node: true } : { browser: true };
-  var cwd = is.node
+  var cwd = !process.browser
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Ping. Is there anything I can do to help move this forward?

boneskull:
If you can provide some kind of test output that would be great. 

I'm concerned this may be a breaking change. 

@mantoni I want to ensure this doesn't break your stuff, as my main concerns here are around how PhantomJS may react. We may want to test this branch against mochify before merging...

mantoni:
Thanks for pinging me. At least browserify sets `process.browser = true` (see https://github.com/defunctzombie/node-process/blob/master/browser.js), so this shouldn't make a difference. In case of Mochify with PhantomJS I expect the path to result in `/`. I'm not entirely sure what WebDriver does with the location though ...

georgecrawford:
Is there any movement on this one?

Issue #1417, opened 2014-11-05
Error in "done() called multiple times" assumptions


Qard:
This code here will hide the true error of certain failed async tests:

https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L216-L230

To reproduce, try this:
```
it('should work', function (done) {
  setImmediate(done)
  throw new Error('hidden')
})
```

You will never see the error with the "hidden" message. It will just tell you done was called twice, which is terribly misleading, as the user code only called it once.

LeonFedotov:
:+1:

victorbadila:
+1 got the same error in a very similar case. tested this particular case and it happened just as described, I can confirm the behaviour.

davija:
I am seeing the same behavior.  Has this been researched yet?  This is almost a year old now.

kenr:
I'm also experiencing the same problem. Lost about an hour of work figuring out this was a bug in mocha and not my tests.

PR #1919, opened 2015-10-12
#1417 - displaying actual error instead of incorrect done() called mu…

davija:
Patch for issue 1417 - Error in "done() called multiple times" assumptions.

The error was hidden by not passing in the exiting error from the context which led to an incorrect error stating done() has been called multiple times.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -314,7 +314,11 @@ Runnable.prototype.run = function(fn) {
         }
         return done(new Error('done() invoked with non-Error: ' + err));
       }
-      done();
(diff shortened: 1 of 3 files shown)


danielstjules:
Thanks for the PR! Could you write a regression test that would fail prior to this PR? E.g. in https://github.com/mochajs/mocha/blob/master/test/integration/regression.js

davija:
regression test has been added

hollomancer:
@davija - Please rebase, and we'll look at it again.

boneskull:
Closing as old, thanks though!

PR #1923, opened 2015-10-15
Bug: fix `describe.only` to work

yorkie:
The `describe.only` seems not to be working currently. After researching something codes from the repository, I learn `only` is based on the awesome option `grep`. But the set to grep from `only` describe does after `runner.run(done)`(Essentially it's `grepTotal`), but the current source code attach grep at https://github.com/yorkie/mocha/blob/b0ce87c78e84d32008dcbc93fa32fd0cd3129845/lib/mocha.js#L465, so the grep from `describe.only` doesn't apply at this case.

I added a trigger `beforeSuite` and an `once` listen for doing re-grep. Now this change works for me :-)

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -475,6 +475,12 @@ Mocha.prototype.run = function(fn) {
   }
   exports.reporters.Base.inlineDiffs = options.useInlineDiffs;
 
+  runner.on('beforeSuite', function() {
+    if (options.grep) {
+      runner.grep(options.grep, options.invert);
+    }
+  });
+
   function done(failures) {
     if (reporter.done) {
       reporter.done(failures, fn);
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -569,6 +569,7 @@ Runner.prototype.runSuite = function(suite, fn) {
   var total = this.grepTotal(suite);
   var afterAllHookCalled = false;
 
+  this.emit('beforeSuite');
   debug('run suite %s', suite.fullTitle());
 
   if (!total || (self.failures && suite._bail)) {


yorkie:
Anyone of maintainers can take a look at this one? Thanks

yorkie:
Ping @dasilvacontin @travisjeffery?

travisjeffery:
Hmm it'd be nice if we could find another way, cause that event will get ran a lot.

yorkie:
> cause that event will get ran a lot

This would cause what problems and why you think this way is not good, thanks :-)

travisjeffery:
Primarily performance problems, maybe it's not too big a deal. I'll try it on some different test suites tomorrow and see how big of a difference it'll be.

yorkie:
Ok, thank you @travisjeffery

boneskull:
@travisjeffery If you have some secret benchmarking sauce you're hiding from us, please share

travisjeffery:
@yorkie do you have a small reproducible example of your issue? Just tried with

``` js

describe('suite not run', function () {
  it('should not run', function () {
    assert(fail);
  });
});

describe.only('suite to run', function () {
  it('should run', function () {
    assert(true);
  });
});

describe('suite not run 2', function () {
  it('should not run', function () {
    assert(fail);
  });
});
```

and it worked.

yorkie:
I guess you should use nested `describe` to reproduce :-(

saschagehlich:
@travisjeffery Try it like this:

```js
describe.only('suite to run', function () {
  it('should run', function () {
    assert(true);
  });
});

describe('not a suite to run', function () {
  it('should run', function () {
    assert(true);
  });
});
```

This runs both blocks even though it should only be running the first one...

travisjeffery:
@saschagehlich that particular case is expected because /suite to run/ ~= /not a suite to run/

saschagehlich:
Is it expected? The docs read "The exclusivity feature allows you to run only the specified suite or test-case by appending .only() to the function.". For me, this means that only the block I flagged with `.only` should be run.

yorkie:
@saschagehlich thanks for pointing out the issue, but it seems to be another issue that I thought about, which is relevant to the regular expression to filter cases.

travisjeffery:
Yeah since .only uses .grep internally, so the docs could be improved there.

hollomancer:
This functionality changed in v3.0.0 - closing for inactivity.

PR #1928, opened 2015-10-17
Add ability to specify ignore patterns

kossnocorp:
When the modular code structure approach is used (module container it's a dir, entry point it's `index.js`, and `test.js` lies next to the source code):

```
.
├── index.js
├── test.js
├── module-a
│   ├── index.js
│   └── test.js
├── module-b
│   ├── index.js
│   └── test.js
└── node_modules
│   ├── ...
```

... it's not possible to use `mocha "**/test.js"` because `node_modules` might also contain `test.js` files.

I've added `--ignore` option that sends ignore patters to `glob` function:

```
mocha "**/test.js" --ignore "node_modules/**/*"
```

It's also possible to ignore multiply patterns:

```
mocha "**/test.js" --ignore "node_modules/**/*","bower_components/**/*"
```

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -313,7 +313,10 @@ var args = program.args;
 if (!args.length) args.push('test');
 
 args.forEach(function(arg){
-  files = files.concat(utils.lookupFiles(arg, extensions, program.recursive));
+  files = files.concat(utils.lookupFiles(arg, {
+    extensions: extensions,
+    recursive: program.recursive
+  }));
 });
 
 // resolve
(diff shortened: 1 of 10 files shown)


boneskull:
I'm not a fan--don't need more command line options. This can be accomplished with `find` as a workaround. 

I *would* be in favor of a `.mochaignore` file, but I would like input from other maintainers on that. 

cc @mochajs/mocha ^

Closing for now.

kossnocorp:
cc @mochajs/mocha :wink:

danielstjules:
I'd be happy with either :)

kossnocorp:
@boneskull then I'll re-implement it with `.mochaignore`?

PR #1933, opened 2015-10-20
Typings for npm packages

falsandtru:
Resolve typings of node modules.
https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages


boneskull:
Thanks for the PR. 

I don't have any interest in whatever this is or maintaining it, however. 

If another maintainer does, please pipe up. 

cc @mochajs/mocha

danielstjules:
Could the type definitions not be moved to https://github.com/borisyankov/DefinitelyTyped? It doesn't make sense to add these types of files in Mocha, since they're language-specific (TypeScript, Flow, anybody else who decides to static analysis in JS) :)

falsandtru:
Thanks for your response. I changed to maintenance-free model. You can update from DefinitelyTyped by `$ dtsm update --save`. However, I should update type definition files of DefinitelyTyped  before merge this PR.

falsandtru:
This is useful type management model. Please include type info, if possible.

danielstjules:
I still don't understand why this needs to be shipped with mocha at all?

falsandtru:
These files are for statically typed programmings. We should move to source repositories from DefinitelyTyped. Here is a final destination of type definition files.

danielstjules:
I realize that, but my point is that they shouldn't need to be added to this repo. If you'd like to use mocha with a new typescript project, I'd just:
```
npm install mocha --save-dev
tsd query mocha --save --action install
```

Then write your tests as you normally would:
``` typescript
<reference path="../typings/mocha/mocha.d.ts" />

  //...
});
```

If we were to make an exception for typescript, we'd need to do the same for flow, and any other language that tries to improve the type system.

falsandtru:
Okay, thanks for your response.

danielstjules:
Sorry to sound harsh. I hope the proposed solution makes sense though!

boneskull:
@danielstjules hey, at least you have a solution!

Issue #1759, opened 2015-06-20
Don't allow duplicate function descriptions in the same describe block


rightaway:
There should be an error or at least a warning if multiple `it` functions have the same description string within the same `describe` block. (Or at least it should be configurable if you want it not to give a warning.) Many times a copy and paste happens and you might forget to change the string.

Forgetting to change it from the default copy and paste description means the string doesn't represent what's actually being tested, so this error or warning serves as a good reminder to update the text to match the code.

I found one situation where duplicate description strings cause problems. Marking one of the functions below as `it.only` doesn't work, all 3 get run because they have the same string.

```
var assert = require("assert")
describe('Array', function(){
  describe('#indexOf()', function(){
    it('should return -1 when the value is not present', function(){
      assert.equal(-1, [1,2,3].indexOf(5));
      assert.equal(-1, [1,2,3].indexOf(0));
    })
    it.only('should return -1 when the value is not present', function(){
      assert.equal(-1, [1,2,3].indexOf(5));
      assert.equal(-1, [1,2,3].indexOf(0));
    })
    it('should return -1 when the value is not present', function(){
      assert.equal(-1, [1,2,3].indexOf(5));
      assert.equal(-1, [1,2,3].indexOf(0));
    })
  })
})
```

jbnicolai:
Sounds good. Feel free to propose a pull request.

aaroncrows:
I have a solution for this but I have a couple questions. What I came up with involves failing a test if a suite sees the same test title more than once. Should that be the default behavior or should this be a flag thing? If it's to be a warning rather than a failure is there an example of that elsewhere in Mocha? I'm not sure what that would look like.

danielstjules:
> What I came up with involves failing a test if a suite sees the same test title more than once.

That would be a major breaking change and would unfortunately require a major release :( If it's behind a flag, it would work. Though I'm not sure how the rest of the team feels about introducing another flag for a small feature. The alternative would be to target the 3.0.0 branch.

aaroncrows:
I could do that. I could also try and make it some sort of warning, but like I said I couldn't find precedence for that elsewhere.

danielstjules:
I think a warning would be good :) Maybe restrict it to specs with the same name in the same suite?

aaroncrows:
[...]

PR #1935, opened 2015-10-21
#1818: implement --production flag

aaroncrows:
Addresses #1818 and #1759.

Adds a --production flag that fails tests that are: pending, marked as skipped, marked as only, or have a duplicate title.

Since only is determined by grep the only way I could think to catch them is to fail a test if it sees a grep. Since this is, in a way, a linter meant to be run over all the tests before a commit or the like I figured that wouldn't be an issue. At the moment it only catches at the test level, so it won't catch suites marked as skipped or only.

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/1935)

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -97,6 +97,7 @@ program
   .option('--trace-deprecation', 'show stack traces on deprecations')
   .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
   .option('--delay', 'wait for async suite definition')
+  .option('--production', 'causes pending tests, tests marked only, and tests with duplicate names to fail')
 
 program.name = 'mocha';
 
@@ -283,6 +284,9 @@ if (program.delay) mocha.delay();
 // --globals
 
 mocha.globals(globals);
+
+// --production
+if (program.production) mocha.productionMode();
 
 // custom compiler support
 
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -441,6 +441,15 @@ Mocha.prototype.delay = function delay() {
   return this;
 };
 
+/*
+* Fail skipped, pending, duplicate title and tests marked only
+* @returns {Mocha}
+*/
+Mocha.prototype.productionMode = function() {
+  this.options.productionMode = true;
+  return this;
+};
+
 /**
  * Run tests and invoke `fn()` when complete.
  *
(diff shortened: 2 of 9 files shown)


aaroncrows:
Any thoughts on this?

danielstjules:
Thanks for the PR! :)

@boneskull what do you think?

glenjamin:
My 2¢ - I'd want to fail on `only` or duplicate titles, but not on pending or skipped.

Only is something that should never be committed, if it's present then it must be a mistake (like leaving in a `debugger` statement).

Pending or skipped I'll sometimes use to allow merging code which will later need updates - I can configure CI tools to count these and show builds as unstable if needed.

dasilvacontin:
> My 2¢ - I'd want to fail on only or duplicate titles, but not on pending or skipped.

Same here.

> Only is something that should never be committed, if it's present then it must be a mistake (like leaving in a debugger statement).

If I commit a `.only` or a `debugger` statement is for debugging purposes, or because it's a WIP commit – I want CI to fail the build.

I'm not fond of the flag's name, `production`, since that's not the condition under which I want `.only`s to fail the build. I also want them to fail the build in PRs, in the `dev` branch, in the `feature` branch, etc. Probably everywhere but locally. So it would always make CI fail.

I wonder if there's anyone who doesn't want CI to fail when tests have `.only`, and why aren't they using pending tests instead. Maybe to make a CI build shorter when they are only interested on the output of certain test? dunno

glenjamin:
> I'm not fond of the flag's name, production, since that's not the condition under which I want .onlys to fail the build. I also want them to fail the build in PRs, in the dev branch, in the feature branch, etc. Probably everywhere but locally. So it would always make CI fail.

Perhaps `--strict` ?

aaroncrows:
IIRC 'strict' was the original idea for the flag but 'production' was suggested in order to avoid similarities to 'strict mode' which it isn't really related to. Taking pending and skip filtering out of this PR would be easy enough. I could see the case for leaving pending in. Not so sure about skips, though.

dasilvacontin:
IMO skip is for temporally (days, weeks, hopefully not more) disabling some tests, because they are no longer correct, they have to be reworked, the code they are covering is undergoing changes and I don't care about it being covered or not, whatever. So I want those to be committed, and not to break the build.

Sadly, I still don't have a better naming proposal, but I'll continue thinking about it.

boneskull:
:-1: @aaroncrows is [this](https://www.npmjs.com/package/eslint-plugin-mocha-only) suitable for the use case?

boneskull:
A feature like this *may* be helpful, but since there's a (arguably better) workaround via ESLint, I'm not in a hurry to merge this...

boneskull:
it doesn't cover `skip` (I think), but could be trivially extended to do so

hollomancer:
This can be closed, due to changes in 3.0.0 in how the internals of it work. The original issue of #1818 needs a new PR.

PR #1944, opened 2015-10-23
(re-)enable after each hooks to fail tests.

jjoos:
In version 1.2.1, support was added to fail test in the after each hook:

> - Added `this.test.error(err)` support to after each hooks. Closes #287

While migrating from teaspoon-mocha to karma-mocha, we discovered that
in the current mocha version this isn't supported anymore. That behaviour has been changed in: ebc6aee50dd1fdca9ab940b43ab79db72f0a4f93 without any documentation change.

If this was a conscious decision adding some documentation about dropping this functionality would then be nice. The old behaviour is documented in: https://github.com/mochajs/mocha/wiki/Conditionally-failing-tests-in-afterEach()-hooks .

### DIFF
--- a/test/integration/multiple.done.js
+++ b/test/integration/multiple.done.js
@@ -16,7 +16,7 @@ describe('multiple calls to done()', function() {
 
     it('results in failures', function() {
       assert.equal(res.stats.pending, 0);
-      assert.equal(res.stats.passes, 1);
+      assert.equal(res.stats.passes, 0);
       assert.equal(res.stats.failures, 1);
       assert.equal(res.code, 1);
     });
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -532,7 +532,10 @@ Runner.prototype.runTests = function(suite, fn) {
           if (err instanceof Pending) {
             test.pending = true;
             self.emit('pending', test);
-          } else if (retry < test.retries()) {
+            return next();
+          }
+
+          if (retry < test.retries()) {
             var clonedTest = test.clone();
             clonedTest.currentRetry(retry + 1);
             tests.unshift(clonedTest);
@@ -540,22 +543,22 @@ Runner.prototype.runTests = function(suite, fn) {
             // Early return + hook trigger so that it doesn't
             // increment the count wrong
             return self.hookUp('afterEach', next);
-          } else {
-            self.fail(test, err);
+          }
+
+          self.fail(test, err);
+        } else {
+          test.state = 'passed';
+        }
+        // Run the actual emit that the test passed after the forEach hooks,
+        // so they can still fail a test. And it's not already reported as
+        // passed.
+        self.hookUp('afterEach', function(err, errSuite) {
+          if (test.state === 'passed') {
+            self.emit('pass', test);
           }
           self.emit('test end', test);
-
-          if (err instanceof Pending) {
-            return next();
-          }
-
-          return self.hookUp('afterEach', next);
-        }
-
-        test.state = 'passed';
-        self.emit('pass', test);
-        self.emit('test end', test);
-        self.hookUp('afterEach', next);
+          next(err, errSuite);
+        });
       });
     });
   }
--- a/test/integration/fixtures/hooks/afterEach.hook.failure.js
+++ b/test/integration/fixtures/hooks/afterEach.hook.failure.js
@@ -1,0 +1,16 @@
+describe('spec 1', function () {
+  afterEach(function () {
+    this.test.error(new Error('something went wrong'));
+  });
+  it('should be set to failure in the afterEach', function () {
+    console.log('test 1');
+  });
+  it('should be set to failure in the afterEach', function () {
+    console.log('test 2');
+  });
+});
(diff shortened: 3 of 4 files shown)


test/integration/multiple.done.js
```diff
@@ -16,7 +16,7 @@ describe('multiple calls to done()', function() {
 
     it('results in failures', function() {
       assert.equal(res.stats.pending, 0);
-      assert.equal(res.stats.passes, 1);
+      assert.equal(res.stats.passes, 0);
```
jjoos: As far as I can tell this is a bug that has been fixed by my changes, this is the content of the test fixture:
```javascript
describe('Multiple Done calls', function(){
  it('should report an error if done was called more than once', function(done){
    done();
    done();
  })

  it('should report an error if an exception happened async after done was called', function (done) {
    done();
    setTimeout(done, 50);
  })

  it('should report an error if an exception happened after done was called', function(done){
    done();
    throw new Error("thrown error");
  })
})
```

danielstjules:
Related: https://github.com/mochajs/mocha/issues/1635

danielstjules:
Mind rebasing on top of master? :)

danielstjules:
Heh, anytime we touch the runner code, it's a lot to QA :sweat_smile:

test/integration/multiple.done.js
danielstjules: Are you sure that code corresponds to the right test? Running that with `mocha --reporter json` on master yields:
```
  "stats": {
    "suites": 1,
    "tests": 3,
    "passes": 3,
    "pending": 0,
    "failures": 2,
```

danielstjules: Nvm, the fixture highlighted above is the wrong one. It's actually: https://github.com/jjoos/mocha/blob/master/test/integration/fixtures/multiple.done.js Which makes sense then :) Since its contents are:
``` javascript
it('should fail in a test-case', function(done) {
  process.nextTick(function(){
    done();
    done();
  });
});
```

In that case, your code is an improvement! I'd be curious if you fixed the issue here as well: https://github.com/jjoos/mocha/blob/master/test/integration/fixtures/multiple.done.js#L4-L11

jjoos:
@danielstjules Ah missed #1635, this pr will fix that issue. Just rebased this commit on master.

test/integration/multiple.done.js
jjoos: Ah sorry for quoting the wrong fixture.

I actually looked at the commented out section of that fixture, but didn't know what the the correct behaviour is.

If I uncomment that fixture, the change to the test to make it pass looks like this:
```patch
     it('results in failures', function() {
       assert.equal(res.stats.pending, 0);
-      assert.equal(res.stats.passes, 0);
-      assert.equal(res.stats.failures, 1);
-      assert.equal(res.code, 1);
+      assert.equal(res.stats.passes, 1);
+      assert.equal(res.stats.failures, 2);
+      assert.equal(res.code, 2);
     });
```

Which seems strange since the `beforeEach` error should cancel the execution of the test, but doesn't do that (I confirmed that the test is executed) But this could be an acceptable limitation of the async beforeEach implementation. If you clarify if this is acceptable I can uncomment it in my PR.

jjoos:
> Ah missed #1635, this pr will fix that issue.

I spoke too soon, it won't fix that issue, but it will give them a reasonable workaround.

Fixing #1635 would mean a far bigger overhaul to consolidate the way the assertion errors are caught. And then apply that to a subset of the hooks.

test/integration/multiple.done.js
danielstjules: Nvm, I'd leave that be then :)

danielstjules:
> spoke too soon, it won't fix that issue, but it will give them a reasonable workaround.

Exactly what I'm thinking :)

glenjamin:
I think if this gets merged it'd be reasonable to close #1635 as the main reason I requested it would be to do exactly what this PR enables.

j-funk:
Was very excited when I found the  wiki article - alas it is not working - can we get this soon?

https://github.com/mochajs/mocha/wiki/Conditionally-failing-tests-in-afterEach()-hooks

j-funk:
@jjoos any chance you're up for fixing the merge conflicts?  This would be a huge help if/when it gets merged!

glenjamin:
I'm happy to adopt this if it's likely to be merged afterwards.

There was a time when it was mergeable but wasn't integrated - so I'm not sure whether there was some other reason it wasn't accepted.

j-funk:
@danielstjules sry to push on this so much, are you a potential gatekeeper for this?

hrimhari:
Any news about this?

arikon:
Anybody?

Rush:
Any updates?

Rush:
FYI, pasted a workaround here https://github.com/mochajs/mocha/issues/1635#issuecomment-323542440

boneskull:
can't use this unless the CLA is signed; please reopen if interested.

Issue #1936, opened 2015-10-21
this.skip causes: Cannot read property 'toString' of undefined


ryanshawty:
Error line in mocha.js
```js
    var el;
    if (test.state === 'passed') {
      var url = self.testURL(test);
      el = fragment('<li class="test pass %e"><h2>%e<span class="duration">%ems</span> <a href="%s" class="replay">‣</a></h2></li>', test.speed, test.title, test.duration, url);
    } else if (test.pending) {
      el = fragment('<li class="test pass pending"><h2>%e</h2></li>', test.title);
    } else {
      el = fragment('<li class="test fail"><h2>%e <a href="%e" class="replay">‣</a></h2></li>', test.title, self.testURL(test));
      var stackString; // Note: Includes leading newline
      var message = test.err.toString(); // mocha.js:2427
````

My test cases are
````js
describe('mod', function(){
  describe('mod1', function(){
    before(function(){
       if(!modLoaded) this.skip(); // causes the error
    });

    it('should do stuff', function(){
       // stuff
    });
  });
});
````

Not sure why ``this.skip()`` is causing an error...

PR #946, opened 2013-08-06
Allow skip from test context for #332

cmbuckley:
Opening a PR to get some feedback for this implementation.

PR #1945, opened 2015-10-23
Correctly skip tests when skipping in a suites before()

ryanshawty:
Tests didn't get skipped when calling ``this.skip()`` in the ``before`` function of a suite.
e.g.

````js
describe('Suite', function(){
  before(function(){
    this.skip();
  });

  it('this is still ran', function(){
    //code
  });
});
````

I don't know if this is the best way of fixing the issue, or even if this is the intended behaviour. Suggestions welcome to improve this PR. :boom:

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/1945)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -507,8 +507,12 @@ Runner.prototype.runTests = function(suite, fn) {
       return;
     }
 
+    function parentPending(suite) {
+      return suite.pending || (suite.parent && parentPending(suite.parent));
+    }
+
     // pending
-    if (test.pending) {
+    if (test.pending || parentPending(test.parent)) {
       self.emit('pending', test);
       self.emit('test end', test);
       return next();

### CODE REVIEW

**Comment by danielstjules** [comment]
Thanks for the PR! Could you undo your changes to mocha.js (we only build it before releases to keep minimal changesets) and add an integration test to https://github.com/mochajs/mocha/blob/master/test/integration/pending.js

**Comment by danielstjules** [comment]
When done, could you also rebase/squash to a single commit? Thanks! :)

**Comment by ryanshawty** [comment]
Will do, only had time yesterday to revert not make test.

File: lib/runner.js [hunk 1]
```diff
@@ -506,8 +506,19 @@ Runner.prototype.runTests = function(suite, fn) {
       return;
     }
 
+    function parentPending(parent) {
+      if (parent.pending) {
+        return true;
+      }
+
+      if (parent.parent) {
+        return parentPending(parent.parent);
```
mislav: Implementation of the whole method could be simply:

```js
function parentPending(suite) {
  return suite.pending || (suite.parent && parentPending(suite.parent)
}
```

mislav:
I was bitten by this bug too and am glad that there is this PR to solve it.

This references #946 (/cc @cmbuckley) and might fix #1936 as reported by @ryanshawty.

mislav:
@ryanshawty When updating your PR, if you have time, can you pull in my related fix from https://github.com/mochajs/mocha/compare/master...mislav:fix-skip? I have confirmed that my change fixes the below simple case, but haven't got time to decipher mocha's integration suite right now and figure out how to write tests for this.

```js
describe('skipping', function() {
  it('amazes', function() {
    this.skip()
    assert(true)
  })
})

// expected: "amazes" test shows as pending within test results
// actual outcome: "TypeError: Cannot read property 'toString' of undefined"
```

danielstjules:
Sorry, I realize this fix is for the browser reporters. Please ignore my previous request for an integration test, as this behavior works correctly from node.

dasilvacontin:
I can pick this up on Thursday if @danielstjules or someone doesn't beat me to it. :)

ryanshawty:
@danielstjules I see! I wasn't sure what you wanted there so haven't got around to fixing updating it yet!

@mislav Yes, I shall try that soon.

puzrin:
Any update on this? I could make new PR if that's acceptable. As far as i understand, this is the only PR blocking new release.

ryanshawty:
My bad! Been busy!

Just rebased and should be good to go.

puzrin:
@danielstjules @dasilvacontin ok to merge now?

lib/runner.js
```diff
@@ -507,8 +507,12 @@ Runner.prototype.runTests = function(suite, fn) {
       return;
     }
 
+    function parentPending(suite) {
+      return suite.pending || (suite.parent && parentPending(suite.parent));
+    }
+
     // pending
-    if (test.pending) {
+    if (test.pending || parentPending(test.parent)) {
```
mislav: This condition could be written simply as

```js
if (parentPending(test)) {
```

puzrin: Not obvious. "parentPending" name means we testing parent, not item itself. I'd leave as is - more readable.

mislav: :+1:

dasilvacontin: > Not obvious. "parentPending" name means we testing parent, not item itself. I'd leave as is - more readable.

Due to the naming, `parentPending` feels like it will only check the ancestors, but it will also check the `Runnable` passed as argument.

What about `(test.pending || parentPending(test))`? (with the necessary changes in the function)

mislav: I think the function and the condition is perfectly fine as-is right now. The code could always be expressed in different ways, but what's important right now is to get this merged.

puzrin: :+1: Let's merge it and move forward. 3 months passed...

PR #1947, opened 2015-10-27
Build from latest source

ralf57:
This PR brings the latest build so that the latest code can be actually used.


danielstjules:
Thanks for the PR! We generally leave this task to one of the maintainers, usually just before a new version is tagged.

ralf57:
I understand but it wouldn't have hurt to have it already in.
Now the v3.0.0. branch has a mocha.js build which is not in sync with the source, which is even worse, IMHO.

PR #1954, opened 2015-11-02
Update debug dependency to 2.2.0

joeycozza:
debug@2.0.0 has a vulnerability in its dependency on ms@0.6.2.

debug@2.2.0 has updated its ms dependency to a version without this vulnerability.

screen shot 2015-11-02 at 11 35 56 am


travisjeffery:
Thanks

PR #1958, opened 2015-11-10
Theme for Browser

embrilliant:
Just an idea. Browser based tests could be more interesting.


danielstjules:
Thanks for the PR! :) I think this would be a great idea for a separate repo/module. We could add a link to it from https://github.com/mochajs/mocha/wiki. I'm not sure that it's suitable for the core installation though.

For reference, here's a screenshot of the theme:

screen shot 2015-12-25 at 3 30 07 pm

embrilliant:
Hey, thanks! Thank you for your comment :)

PR #1965, opened 2015-11-12
Dont double install bdd ui

### LINKED ISSUE #1964: require('mocha').beforeEach does not run outer lifecycle hooks when running inner testcases marked by `it.only`
This code associated with this writeup is available here: https://gist.github.com/cowboyd/f5611829c7a3ad084642

given the following mocha code:

``` javascript
var beforeEach = require('mocha').beforeEach;
describe("the outer context", function () {
  beforeEach(function() {
    this.didRunBeforeEach = true;
  });
  describe("the inner context", function() {
    it("runs its parent beforeEach", function() {
      expect(this.didRunBeforeEach).to.equal(true);
    });
  }) ;
});
```

because `"the inner context"` is nested inside `"the outer context"`,
it should run the outer context's `beforeEach` hook. It does do this
under normal circumstances:

``` javascript
~/C/E/mocha-demo ❯❯❯ npm t

> mocha-demo@1.0.0 test /Users/cowboyd/Code/Ember/mocha-demo
> mocha index.js



  the outer context
    the inner context
      ✓ runs its parent beforeEach


  1 passing (7ms)
```

But if we configure the inner context to
use `it.only` to select only the single test-case:

``` javascript
var beforeEach = require('mocha').beforeEach;
describe("the outer context", function () {
  beforeEach(function() {
    this.didRunBeforeEach = true;
  });
  describe("the inner context", function() {
    it.only("runs its parent beforeEach", function() {
      expect(this.didRunBeforeEach).to.equal(true);
    });
  }) ;
});
```

Then it fails with the following error:

```
~/C/E/mocha-demo ❯❯❯ npm t                                                                                                                                                                                                                  ⏎

> mocha-demo@1.0.0 test /Users/cowboyd/Code/Ember/mocha-demo
> mocha index.js


  1) runs its parent beforeEach

  0 passing (9ms)
  1 failing

  1)  runs its parent beforeEach:
     AssertionError: expected undefined to equal true
      at Context.<anonymous> (index.js:15:40)



```

Which indicates that the hook is either not being run.
¯

cowboyd:
Resolves #1964 

the bdd UI was being called twice, once from within the `Mocha` constructor, and also from the `_mocha` command line. This prevents a UI from being called twice, and it's `pre-require` hooks being called twice so that it partially clobbers itself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,11 @@ test-only:
 	@./bin/mocha \
 		--reporter $(REPORTER) \
 		--ui qunit \
+		test/acceptance/misc/only/bdd-require
+
+	@./bin/mocha \
+		--reporter $(REPORTER) \
+		--ui qunit \
 		test/acceptance/misc/only/qunit
 
 test-mocha:
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -77,7 +77,7 @@ module.exports = function(suite) {
      * acting as a thunk.
      */
 
-    context.it = context.specify = function(title, fn) {
+    var it = context.it = context.specify = function(title, fn) {
       var suite = suites[0];
       if (suite.pending) {
         fn = null;
@@ -93,7 +93,7 @@ module.exports = function(suite) {
      */
 
     context.it.only = function(title, fn) {
-      var test = context.it(title, fn);
+      var test = it(title, fn);
       var reString = '^' + escapeRe(test.fullTitle()) + '$';
       mocha.grep(new RegExp(reString));
       return test;
--- a/test/acceptance/misc/only/bdd-require.js
+++ b/test/acceptance/misc/only/bdd-require.js
@@ -1,0 +1,18 @@
+/*jshint node: true */
+
+var mocha = require('../../../../lib/mocha');
+
+var beforeEach = mocha.beforeEach;
+var it = mocha.it;
+var describe = mocha.describe;
+
+describe('it.only via require("mocha")', function() {
+  beforeEach(function() {
+    this.didRunBeforeEach = true;
+  });
+  describe("nested within a describe/context", function() {
+    it.only('should run all enclosing beforeEach hooks', function() {
+      require('assert').equal(this.didRunBeforeEach, true);
+    });
+  });
+});


cowboyd:
Any chance one of the maintainers could have a look at this? It's a pretty simple fix and removes a pretty big paper-cut.

If this isn't the right place to fix it, would be happy to move it somewhere else.

boneskull:
@cowboyd how does this address #1964?  please squash

cowboyd:
@boneskull It addresses #1964 because here https://github.com/mochajs/mocha/blob/3f78f10ca40dba63be10454c8e17dc9586416e10/lib/interfaces/bdd.js#L96 `it.only` does a dynamic lookup on context to find `context.it` 

However, if the hook is run twice, then `context.it` is redefined, and so `it.only` is pointing at the wrong `context.it`. I can also fix it by capturing a reference to the correct `it` function and calling it directly.

cowboyd:
Squashed. It does not address the problem of the BDD interface getting installed twice, but the fix is more explicity and perhaps more robust.

cowboyd:
@boneskull Have you had a chance to look into this? The fix is now both obvious and minimal, and the commits have been squashed.

cowboyd:
/ping @boneskull @travisjeffery @danielstjules 

I'd really like to land this, or make whatever chances are necessary to get it landed. It's a paper-cut that bites me at least a couple times a week.

danielstjules:
Hi @cowboyd, could you add a small test to https://github.com/mochajs/mocha/tree/master/test/integration rather than under acceptance? If not, I can jump on it

Aside from that, fix looks good - makes sense :)

cowboyd:
@danielstjules done, thanks!

cowboyd:
@danielstjules Did you have a chance to have a look?

danielstjules:
I did! Since the integration test is using `only` though, it's preventing the rest of the suite from running. You can compare by running `make test-integration` on this branch vs master. Could you either update the test to use a similar approach as other tests in that dir (e.g. https://github.com/mochajs/mocha/blob/master/test/integration/timeout.js) or update https://github.com/mochajs/mocha/blob/master/test/acceptance/misc/only/bdd.js since test-only will run it with its own instance of mocha https://github.com/mochajs/mocha/blob/master/Makefile#L104-L118

cowboyd:
@danielstjules I had moved it out of acceptance and back into integration, but under a separate file.

danielstjules:
Looks good - thanks for your work! :) And sorry for maybe misdirecting this PR a bit.

PR #1967, opened 2015-11-17
Silence Bluebird js warnings

krisr:
Bluebird js now prints a warning message and stack trace when it discovers Promises that create other Promises but do not return them. Mocha trips this warning when the user chains together multiple hooks, and this PR eliminates the warning by explicitly returning null instead of undefined which signals that the warning should not be emitted.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -289,6 +289,9 @@ Runnable.prototype.run = function(fn) {
       result
         .then(function() {
           done();
+          // Return null so libraries like bluebird do not warn about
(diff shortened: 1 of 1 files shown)


boneskull:
@krisr This fails a lint check; please fix, otherwise LGTM.

krisr:
@boneskull, thanks. I fixed the whitespace issue.

boneskull:
tyvm!

simonratner:
Any chance of publishing a new release? Just ran into this myself.

PR #1972, opened 2015-11-19
Fixes #1794

### LINKED ISSUE #1794: Can't --require custom UI and use it
In the [Third party UIs](https://github.com/mochajs/mocha/wiki/Third-party-UIs) wiki page, it is said that one can `--require` a module which enriches the `mocha.interfaces` object bag with a custom UI and directly use it with the `--ui` option.

This doesn't seem to work.

Looking at the latest [`bin/_mocha`](https://github.com/mochajs/mocha/blob/952dca73d1c24132b0fb8e30e3300b5346f0f1c1/bin/_mocha), I can see why: [the UI lookup](https://github.com/mochajs/mocha/blob/952dca73d1c24132b0fb8e30e3300b5346f0f1c1/bin/_mocha#L214) happens before [all the requires are required](https://github.com/mochajs/mocha/blob/952dca73d1c24132b0fb8e30e3300b5346f0f1c1/bin/_mocha#L311).

Is this a regression?

jgkim:
Moves Mocha interface allocation to after require option has been processed to allow for use of custom Mocha UI's.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -201,10 +201,6 @@ if (program.reporterOptions !== undefined) {
 
 mocha.reporter(program.reporter, reporterOptions);
 
-// interface
-
-mocha.ui(program.ui);
-
 // load reporter
 
 try {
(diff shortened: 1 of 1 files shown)


danielstjules:
Fixed via https://github.com/mochajs/mocha/commit/a61b1f3287b436522545e5458d071d7e791c15d0 thanks!

PR #1981, opened 2015-11-26
Fix HTML reporting display to show errors if done is called multiple times, or if an exception is thrown after done is called.

Standard8:
In either of the test cases below, when using the HTML reporter, the failure details are not currently shown. The failure count is increased, but there's no specific details.

This patch adjusts the reporter so that failures after the initial done() are displayed. This doesn't seem to cause any regressions in the manual browser tests.

The output does show the test as passing, and then shows it as failing, but I think that's reasonable given what is happening and is better than not showing any errors.

Test cases:

```
describe('Multiple Done calls', function(){
  it('should report an error if done was called more than once', function(done){
    done();
    done();
  })

  it('should report an error if an exception happened after done was called', function(done){
    done();
    throw new Error("thrown error");
  })
})
```

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/1981)

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -129,7 +129,10 @@ function HTML(runner) {
   });
 
   runner.on('fail', function(test) {
-    if (test.type === 'hook') {
+    // For type = 'test' its possible that the test failed due to multiple
+    // done() calls. So report the issue here.
+    if (test.type === 'hook'
+      || test.type === 'test') {
       runner.emit('test end', test);
     }
   });
--- a/test/browser/index.html
+++ b/test/browser/index.html
@@ -14,6 +14,7 @@
     <script src="array.js"></script>
     <script src="../acceptance/duration.js"></script>
     <script src="../acceptance/timeout.js"></script>
+    <script src="multiple-done.js"></script>
     <script>
       onload = function(){
         mocha.checkLeaks();
--- a/test/browser/multiple-done.js
+++ b/test/browser/multiple-done.js
@@ -1,0 +1,16 @@
+describe('Multiple Done calls', function(){
+  it('should report an error if done was called more than once', function(done){
+    done();
+    done();
+  })
+
+  it('should report an error if an exception happened async after done was called', function (done) {
+    done();
+    setTimeout(done, 50);
+  })
+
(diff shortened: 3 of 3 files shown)


dasilvacontin:
It's definitely an improvement. `failures` are correctly updated, `passes` don't decrease, we could improve that.

It would be nice to add the following test – it's rendered outside the suite, since the suite has "ended" already, though. At least the error is not reported as belonging to a different test, but that depends on mocha's internals.

```diff
diff --git a/test/browser/multiple-done.js b/test/browser/multiple-done.js
index 7ac3ff3..37b2703 100644
--- a/test/browser/multiple-done.js
+++ b/test/browser/multiple-done.js
@@ -1,10 +1,15 @@
describe('Multiple Done calls', function(){
   it('should report an error if done was called more than once', function(done){
     done();
     done();
   })
 
-  it('should report an error if an exception happened after done was called', function(done){
+  it('should report an error if an exception happened async after done was called', function (done) {
+    done()
+    setTimeout(done, 50)
+  })
+
+  it('should report an error if an exception happened sync after done was called', function(done){
     done();
     throw new Error("thrown error");
   })

```

Standard8:
@dasilvacontin Sorry for the delay, I've now added the extra test.

danielstjules:
Nice PR! :) Looks like this does a good job of aligning behavior with the node reporters:

``` javascript
$ mocha test.js

  ․․․․

  2 passing (9ms)
  2 failing

  1) Multiple Done calls should report an error if done was called more than once:
      at Suite.<anonymous> (test.js:2:3)
      at Object.<anonymous> (test.js:1:63)
      at Array.forEach (native)
      at node.js:961:3

  2) Multiple Done calls should report an error if an exception happened after done was called:
     Error: thrown error
      at Context.<anonymous> (test.js:9:11)
```

danielstjules:
@dasilvacontin I don't think it's necessary to have passes decrease, since we don't do this for any of the other reporters. Furthermore, any async exceptions will be reported as `Uncaught Error`, and we can't correctly attribute them to a specific test anyway.

danielstjules:
Thanks again! :)

PR #1995, opened 2015-12-04
make sure the xunit output dir exists before trying to write to it

ianwremmel:


danielstjules:
Can you take a look at the linting errors?

danielstjules:
Can you move that logic to `_mocha`, since it's already responsible for dir creation with browser tests (grep mkdirp to see what I mean)

ianwremmel:
I encountered the problem with grunt-mocha-test. Won't putting the fix in _mocha only help with command line usage while still leaving it broken for programmatic usage?

danielstjules:
Ah, you're right. Your current approach makes most sense anyway. Thanks!

PR #2001, opened 2015-12-10
Should not throw error when files.length == 0

mariodu:
Directory structure likes below:

```
tests/
    units/
```

In `units`, there isn't any file. When I use mocha to run this path, it throws `can't resolve path`, but `glob(path)` returns `[]`. Maybe, it shouldn't throw error.

```
tests/
    units/
        lib-test.js
```

If directory structure likes above, it works well.

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2001)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -621,8 +621,10 @@ exports.lookupFiles = function lookupFiles(path, extensions, recursive) {
     if (exists(path + '.js')) {
       path += '.js';
     } else {
-      files = glob.sync(path);
-      if (!files.length) {
+      try {
+        files = glob.sync(path);
+      } catch (e) {
+        debug(e);
         throw new Error("cannot resolve path (or pattern) '" + path + "'");
       }
       return files;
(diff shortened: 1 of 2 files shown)


danielstjules:
Thanks for the PR! How are you invoking mocha given the dir structure above to trigger the exception?

mariodu:
like:
```
mocha --reporter spec --recursive --growl -t 0 ./test/units/*-test.js
```

dasilvacontin:
This also happens when you invoke mocha without specifying folder / files, and there's no `test` folder, or there are no files under the folder. It's a common case when starting a new project using a boilerplate, and you don't want to bother with tests just yet because you're experimenting and changing how things work very fast.

knownasilya:
Can this be merged?

Munter:
I think #2450 already covers this case

knownasilya:
Is that in the latest version?

Munter:
I don't think a release has been made yet. It just got merged into master

boneskull:
@Munter should this be closed?  if so, please do

Munter:
I think it should. If I'm wrong and this error case is not covered by #2450 then please reopen

Issue #220, opened 2012-01-18
Would you be open to adding sinon.js to the project?


adomokos:
I need to spy or stub on some of the functions I am trying to test. Can I add [sinon.js](https://github.com/cjohansen/Sinon.JS) as a dependent project to mocha?

PR #2005, opened 2015-12-14
XUnit Reporter Writes to stdout, falls back to console.log

### LINKED ISSUE #1674: XUnit reporter should use process.stdout if available... fallback to console.log
Hello

I'm running Mocha on PhantomJS using https://github.com/metaskills/mocha-phantomjs This means Mocha is not run in Node and thus has no access to fs.

Mocha-PhantomJS provides the option of dumping the reporter output to a file by redirecting process.stdout.write. Unfortunately, I'm having issues with the xunit reporter since it is using console.log.

I understand this is to maintain compatibility with browsers, but would it be desirable to check the existence of process.stdout and call it if available?

It would be a trivial PR... I would gladly prepare it but I would like to ask for your insight first:
```javascript
/**
 * Write out the given line
 */
XUnit.prototype.write = function(line) {
    if (this.fileStream) { 
        // Node JS environment
        this.fileStream.write(line + '\n');
    } else if (process && process.stdout) {
        // Browser with process.stdout (as in mocha-phantomjs)
        process.stdout.write(line + '\n');
    } else {
        // Browser
        console.log(line);
    }
};
```

jonnyreeves:
Fixes #1674; provides support *much* needed for [mocha-phantomjs](https://github.com/nathanboktae/mocha-phantomjs) (see [#133](https://github.com/nathanboktae/mocha-phantomjs/issues/133), [#220](https://github.com/nathanboktae/mocha-phantomjs/issues/220), etc).  Maintains support for running the XUnit reporter in a browser (unlike the previously reverted PR #1068).

ping @alemangui, @nathanboktae

Thanks all.

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2005)

### DIFF
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -109,6 +109,8 @@ XUnit.prototype.done = function(failures, fn) {
 XUnit.prototype.write = function(line) {
   if (this.fileStream) {
     this.fileStream.write(line + '\n');
+  } else if (typeof process === 'object' && typeof process.stdout === 'object') {
+    process.stdout.write(line + '\n');
   } else {
     console.log(line);
   }
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -109,7 +109,7 @@ XUnit.prototype.done = function(failures, fn) {
 XUnit.prototype.write = function(line) {
   if (this.fileStream) {
     this.fileStream.write(line + '\n');
-  } else if (typeof process === 'object' && typeof process.stdout === 'object') {
+  } else if (typeof process === 'object' && process.stdout) {
     process.stdout.write(line + '\n');
   } else {
     console.log(line);


lib/reporters/xunit.js
```diff
@@ -109,6 +109,8 @@ XUnit.prototype.done = function(failures, fn) {
 XUnit.prototype.write = function(line) {
   if (this.fileStream) {
     this.fileStream.write(line + '\n');
+  } else if (typeof process === 'object' && typeof process.stdout === 'object') {
```
danielstjules: This would be better written as:
``` javascript
} else if (typeof process === 'object' && process.stdout) {
```

danielstjules:
Comment aside, LGTM. Ideally, reporters wouldn't invoke console.log directly anyway, but instead be passed a writable stream to use.

jonnyreeves:
Updated based on feedback; merging this would be joy to ~~the world~~many! :christmas_tree:

danielstjules:
:+1: :santa:

nathanboktae:
> Ideally, reporters wouldn't invoke console.log directly anyway, but instead be passed a writable stream to use.

Yes this would be great and solve alot of magic done by plugins like `mocha-multi`, mocha-phantomjs`, `mocha-casperjs`, etc. Thanks all!  :clap: :santa:

PR #1830, opened 2015-08-13
rprieto's tagging support for current master

noorus:
See pull #1445

PR #2011, opened 2015-12-17
Add the ability to tag test files when using the exports interface.

davidpaulhunt:
This is something we implemented in my current shop and it has been very useful. Our applications and their services have grown. This makes being able to test particular functionality out of context very useful (or in context if you so choose). To make this accessible to all, I've added a change to the exports ui that looks for an array of tags in options.

Example:
```js
// test/foo.js

var Taggable = require('mocha').Taggable;

module.exports = Taggable(['foo', 'bar'], {
  Foo: {
    '#new Foo()': {
      'foo should be of Foo': function() {
        var foo = foo();
        foo.should.be.instanceOf(Foo);
      }
    }
}
```

In your task runner you would do something like:
```js
// test/run

var args = process.argv.slice(2);
// some other things like get filenames
var opts = { tags: args };
var mocha = new Mocha(opts).ui('exports');
filenames.forEach(function(file) {
  mocha.addFile(path.join('test',file)).run(failureHandler);
});
```

And to run your tests:
`test/run foo bar other thing` which would result in all files tagged with foo, bar, other, or thing to be run. This doesn't affect tests when being run by `npm test` or `mocha` directly unless they call on your test runner `test/run`.

### DIFF
--- a/lib/interfaces/exports.js
+++ b/lib/interfaces/exports.js
@@ -25,7 +25,25 @@ var Test = require('../test');
 module.exports = function(suite) {
   var suites = [suite];
 
-  suite.on('require', visit);
+  suite.on('require', function(obj, file) {
+    if (suite.tags) {
+      if (obj.tags) {
+        var found = false;
+        for (var i = 0; i < suite.tags.length; i++) {
+          if (obj.tags.indexOf(suite.tags[i]) !== -1) {
+            found = true;
+            break;
+          }
+        }
+
+        if (found) {
+          visit(obj, file);
+        }
+      }
+    } else {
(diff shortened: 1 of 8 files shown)


danielstjules:
Thanks for the PR! I don't think we'd want to introduce the functionality for a single interface. It'd be great to expand on https://github.com/mochajs/mocha/pull/1830 to include exports as well.

davidpaulhunt:
Woah, I should have searched for comments on tags. Really cool to see other people have had the same use case and thoughts on the subject. I plan on digging into #1830 and #1445 in a bit. Thanks for the heads up!

PR #2032, opened 2015-12-30
adds tree/dynamic mocha.opts loading getoptions

ivoputzer:
to me the way `mocha.opts` files are loaded should be more flexible and traverse the entire tree instead of pointing to `./test/mocha.opts` directly (or the specified file using the `--opts` flag). it sometimes happens, at least to me, that i test single files and folders with different configurations instead of an entire suite. so, as mentioned in a few issues the idea is to load `mocha.opts` files directly from where the tests get invoked without breaking backward compatibility nor api! 

running both files and folders would produce an option-lookup like below and the first file found would stop the chain and get loaded into `process.argv` (just like before):

mocha path/to/my/tests
mocha path/to/my/tests/spec.js

  > /path/to/my/tests/mocha.opts
  > /path/to/my/mocha.opts
  > /path/to/mocha.opts
  > /path/mocha.opts
  > /mocha.opts
  > /test/mocha.opts # not to break bakward compatibility

the code is quite straight forward but in order to add tests here i had to change the actual source a bit and didn't want those changes to be part of this pr; let me know if there's anything else i can do :)

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2032)

### DIFF
--- a/bin/options.js
+++ b/bin/options.js
@@ -16,18 +16,19 @@ module.exports = getOptions;
  */
 
 function getOptions() {
-  var lookupOptsPaths = ['mocha.opts', 'test/mocha.opts']
+  var lookupOptsPaths = ['mocha.opts', 'test/mocha.opts'];
 
   process.argv.slice(2)
     .forEach(function (arg, i, argv){
       var regex = /^\-\-?/
-        , isOptionArg = argv[i - 1] && regex.test(argv[i - 1]) || regex.test(arg)
+        , isOptionArg = argv[i - 1] && regex.test(argv[i - 1]) || regex.test(arg);
       if (isOptionArg) return;
       Array.prototype.unshift.apply(lookupOptsPaths, getOptionsLookupTree(arg));
     })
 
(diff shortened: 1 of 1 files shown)


TrejGun:
can i suggest to use
https://github.com/zerkalica/node-config-loader

ivoputzer:
@TrejGun i didn't want to add yet another dependency, but if it's actually ok, using the config-loader could be a great alternative.

TrejGun:
if `escape-string-regexp` is a dependency this can be too, because it does only one thing

danielstjules:
I'd rather not add a dependency on a project with so few stars or little visibility. I'd rather avoid the risk of a project that may be unstable or break semver. Not saying it will, but that it makes me nervous. :)

boneskull:
@TrejGun thanks for the PR.  we're not going to be modifying `mocha.opts` behavior right now, as it's likely to be deprecated in lieu of a proper configuration file.

PR #2033, opened 2016-01-01
Update tests for newer versions of should.js

tomhughes:
[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2033)

### DIFF
--- a/package.json
+++ b/package.json
@@ -290,7 +290,7 @@
     "browserify": "10.2.4",
     "coffee-script": "~1.8.0",
     "eslint": "^1.2.1",
-    "should": "~4.0.0",
+    "should": "~8.0.0",
     "through2": "~0.6.5"
   },
   "files": [
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -140,7 +140,7 @@ module.exports = {
 
     // Ignore empty lines before/after diff
     return diffs.map(function(diff) {
-      return diff.slice(1, -2).join('\n');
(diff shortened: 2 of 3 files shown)


danielstjules:
Looks like the jobs failed with several errors. You can run the full test suite yourself using `make test-all` or `npm test`

tomhughes:
Think I've figured it out - there is an extra line in the stack traces now. Instead of:

```
      AssertionError: expected 'foo rar baz' to be 'foo bar baz'
      + expected - actual

      -foo rar baz
      +foo bar baz
      
      at Assertion.prop.(anonymous function) (node_modules/should/lib/should.js:61:14)
      at Context.<anonymous> (test/integration/fixtures/diffs/diffs.js:12:19)
      at callFn (lib/runnable.js:312:21)
 ```

we get:

```
      AssertionError: expected 'foo rar baz' to be 'foo bar baz'
      + expected - actual

      -foo rar baz
      +foo bar baz
      
      at Assertion.fail (node_modules/should/lib/assertion.js:91:17)
      at Assertion.Object.defineProperty.value (node_modules/should/lib/assertion.js:163:19)
      at Context.<anonymous> (test/integration/fixtures/diffs/diffs.js:12:19)
      at callFn (lib/runnable.js:312:21)
```
so we need to discard one more line when extracting the diff.

I've pushed an updated version that now passes.

danielstjules:
Thanks! It's unfortunate that the trace is a bit noisier, but at least we'll be using a recent release :)

Issue #1991, opened 2015-12-02
Declarations do not get cleaned up unless you set them to `null`


rodoabad:
@boneskull Just wondering if we should be actively cleaning up out declarations per describe block.
### Leaking

![Mocha Memory Leak](https://github.com/rodoabad/mocha-memory-leak/raw/master/img/mocha-memory-leak-1.JPG)

``` javascript
(function () {

  'use strict';

  function MemoryLeak() {
  }

  for (var i = 0; i < 1000; i += 1) {

    describe('Suite #' + i, function () {

      var suite;

      beforeEach(function () {

        suite = new MemoryLeak();

      });

      it('should do something', function () {

        suite.isLeaking = true;

      });

    });

  }

})();
```
### Not Leaking

![Mocha Memory Leak](https://github.com/rodoabad/mocha-memory-leak/raw/master/img/mocha-memory-leak-2.JPG)

```
(function () {

  'use strict';

  function MemoryLeak() {
  }

  for (var i = 0; i < 1000; i += 1) {

    describe('Suite #' + i, function () {

      var suite;

      beforeEach(function () {

        suite = new MemoryLeak();

      });

      afterEach(function () {

        suite = null;

      });

      it('should do something', function () {

        suite.isLeaking = true;

      });

    });

  }

})();
```

More here - https://github.com/rodoabad/mocha-memory-leak

boneskull:
how do we know we're not just waiting on the garbage collector?

rodoabad:
Shouldn't gc run after each test?

tinganho:
I have also experienced this. Mocha is storing the callbacks and it won't clean up because the runtimes doesn't know if you will call those "callbacks" or not. And the memory footprint can be really substantial if your test suite is large. So I agree that it should clean up after each test.

boneskull:
feel free to try and fix this; I'm probably going to be heads-down on a rewrite for awhile.

gumatias:
Anyone know a solution to prevent this other than manually cleaning up after each test? I created a related issue here https://github.com/mochajs/mocha/issues/2030 thanks!

danielstjules:
> Shouldn't gc run after each test?

Nope, just like your node server won't run gc between each request. Here's an example:
``` javascript
// node --trace-gc --max-old-space-size=10 testing.js

var fn;
var obj;
var i = 0;

setInterval(function() {
  fn = function() {
    // Create some objects
    obj = [];
    for (var j = 0; j < 100; j++) {
[...]

PR #2037, opened 2016-01-03
Fix for memory leak caused by referenced to deferred test functions

bd82:
(it/before[Each]/after[Each]).

related to #1991

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2037)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -15,6 +15,7 @@ var stackFilter = utils.stackTraceFilter();
 var stringify = utils.stringify;
 var type = utils.type;
 var undefinedError = utils.undefinedError;
+var isArray = utils.isArray;
 
 /**
  * Non-enumerable globals.
@@ -720,6 +721,49 @@ Runner.prototype.uncaught = function(err) {
 };
 
 /**
+ * Cleans up the reference to the test's deferred function.
+ * @see cleanSuiteReferences for details.
+ * @param {Test} test
+ */
+function cleanTestReferences(test) {
+  delete test.fn;
+}
+
+/**
+ * Cleans up the references to all the deferred functions
+ * (before/after/beforeEach/afterEach) of a Suite.
+ * These must be deleted otherwise a memory leak can happen,
+ * as those functions may reference variables from closures,
+ * thus those variables can never be garbage collected as long
+ * as the deferred functions exist.
+ *
+ * @param {Suite} suite
+ */
+function cleanSuiteReferences(suite) {
+  function cleanArrReferences(arr) {
+    for (var i = 0; i < arr.length; i++) {
+      delete arr[i].fn;
+    }
+  }
+
+  if (isArray(suite._beforeAll)) {
+    cleanArrReferences(suite._beforeAll);
+  }
+
+  if (isArray(suite._beforeEach)) {
+    cleanArrReferences(suite._beforeEach);
+  }
+
+  if (isArray(suite._afterAll)) {
+    cleanArrReferences(suite._afterAll);
+  }
+
+  if (isArray(suite._afterEach)) {
+    cleanArrReferences(suite._afterEach);
+  }
+}
+
+/**
  * Run the root suite and invoke `fn(failures)`
  * on completion.
  *
@@ -749,6 +793,10 @@ Runner.prototype.run = function(fn) {
   }
 
   debug('start');
+
+  // references cleanup to avoid memory leaks
+  this.on('test end', cleanTestReferences);
+  this.on('suite end', cleanSuiteReferences);
 
   // callback
   this.on('end', function() {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -187,6 +187,8 @@ var isArray = typeof Array.isArray === 'function' ? Array.isArray : function(obj
   return Object.prototype.toString.call(obj) === '[object Array]';
 };
 
+exports.isArray = isArray;
+
 /**
  * Buffer.prototype.toJSON polyfill.
  *
--- a/test/integration/fixtures/regression/issue-1991.js
+++ b/test/integration/fixtures/regression/issue-1991.js
@@ -1,0 +1,47 @@
+function MemoryLeak() {
+  this.myArr = [];
+  for (var i = 0; i < 1000000; i++) {
+    this.myArr.push(i)
+  }
+}
+
+var numOfTests = 300;
+for (var i = 0; i < numOfTests; i += 1) {
+  /*
+   * This Test suite will crash V8 due to:
+   * 'FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory'
+   * if all the deferred functions references have not been cleared
+   */
+  describe('Memory Leak Suite #' + i, function () {
+
+    // The <closureVar> variable will be accessed by the test below.
+    // As long as those test's functions are
+    // referenced in memory, the closure variable may not be garbage collected
+    // as it is still referenced.
+    // * In a chrome heap snapshot it will appear under "system / Context" (a scope)
+    var closureVar;
+
+    before(function () {
+      var x = closureVar ? 1 : 2
+    });
+
+    after(function () {
+      var x = closureVar[0]
+    });
+
+    beforeEach(function () {
+      var x = closureVar ? 1 : 2
+    });
+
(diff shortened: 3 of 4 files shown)


test/integration/memory.leak.js
```diff
@@ -0,0 +1,47 @@
...
+
+    afterEach(function () {
+      var x = closureVar ? 1 : 2
+    });
+
+    it('access a variable via a closure', function () {
+      // in node.js versions 0.11 and 0.12 it seems that large array allocation is very inefficient
+      // thus causing the respective Travis-CI jobs to run much longer and causing this test
+      // to sometimes fail due to timeout.
+      this.timeout(1000);
```
danielstjules: `this.timeout` won't actually timeout the test after 1 second, since MemoryLeak's for loop is a blocking operation, and mocha just uses `setTimeout`: https://github.com/mochajs/mocha/blob/d811eb9614e5b459dabb4f2496e69f961decec90/lib/runnable.js#L211-L217

danielstjules:
Thanks for the PR! Going over it now!

Could you move `test/integration/memory.leak.js` to `test/integration/fixtures/memory.leak.js`? Integration test fixtures are ran by separate tests and have their output verified. Running the file directly using `make test-integration` results in quite a bit of noise. If ran as a fixture, you can simply check for a 0 exit code, and the lack of "Allocation failed" in the output :)

lib/runner.js
```diff
@@ -720,6 +721,47 @@ Runner.prototype.uncaught = function(err) {
...
+ */
+function cleanTestReferences(test) {
+  delete test.fn;
+}
+
+/**
+ * Cleans up the references to all the deferred functions (before/after/beforeEach/afterEach)
+ * of a Suite. These must be deleted otherwise it may create a memory leak as those functions may
+ * reference variables from closures, thus those variables can never be garbage collected as long
+ * as the deferred functions exist.
```
danielstjules: The file appears to mostly follow a 80 char ruler - could you update this doc comment?

test/integration/memory.leak.js
danielstjules: This example might help clarify:
``` javascript
$ cat example.js
setTimeout(function() {
  console.log(new Date(), 'Timeout!');
}, 1000);

console.log(new Date());
var array = [];
for (var i = 0; i < 20000000; i++) {
  array.push({val: i});
}

// $ time node example.js
// Sun Jan 03 2016 13:31:21 GMT-0500 (EST)
// Sun Jan 03 2016 13:31:27 GMT-0500 (EST) 'Timeout!'
//
// real	0m6.837s
// user	0m6.444s
// sys	0m0.425s
```

danielstjules:
Actually, rather than an integration test (since it takes 40s+ to run on my laptop) could you add a unit test that verified runnable fns were deleted from a suite? It could be added to `test/runner.js` I'm imagining something like

``` javascript
it('deletes runnable fn when no longer needed', function(done) {
  var test = new Test('test', function() {});
  suite.addTest(test);
  suite.before(function() {});
  suite.beforeEach(function() {});
  suite.afterEach(function() {});
  suite.after(function() {});

  var runner = new Runner(suite);
  runner.runTests(function(err) {
    // perform assertions here
    done();
  });
});
```

test/integration/memory.leak.js
bd82: Thanks for the clarification.
The weaker condition of waiting **at least** 1000ms is good enough in this use case.
The timeout is there because on node.js versions 0.11 and 0.12 on Travis-CI 
sometimes take slightly longer than 200ms. (the default timeout) and thus the change
failed there.

see
 * https://travis-ci.org/mochajs/mocha/builds/99955445
 * https://travis-ci.org/mochajs/mocha/builds/99954248

danielstjules: Ah, my mistake! Makes sense :)

lib/runner.js
bd82: Will do.

bd82:
> Actually, rather than an integration test (since it takes 40s+ to run on my laptop) could you add a unit test that verified runnable fns were deleted from a suite? It could be added to test/runner.js I'm imagining something like

From which century is that laptop? :) anyhow I see your point as the Travis CI containers are also
fairly weak. 

The problem is that verification of simple property deletion is a much weaker check as it will not detect
new memory leaks created by changes that copy those <fn>  references somewhere else.

I think I can do something simple that scans the object graph starting from the Suite and detects 
properties that meet the conditions:
1. are of Function type
2. toString() on them contains the the name of the variable ("closureVar")

That check will be more dynamic (no fixed property names) and better detect regressions in the future.

danielstjules:
> From which century is that laptop? :) anyhow I see your point as the Travis CI containers are also
> fairly weak.

haha that's fair, the 40s is with node v0.12. With Node v4, I'm seeing 8s. 

> The problem is that verification of simple property deletion is a much weaker check as it will not detect
> new memory leaks created by changes that copy those references somewhere else.

A regression that introduced a smaller memory leak might still go undetected. But your integration test does a good job of catching most cases - e.g. beforeEach fn's magically not being deleted. Others, I'm less worried about. Some reporters (markdown, doc, html) perform simple operations on `test.fn`. One could think of contrived examples in which they're copied, causing a memory leak unique to a given reporter. Though I'm sure neither of us would advocate running that integration test for each reporter to achieve that kind of branch coverage.

You make a good point, and I'm happy with the existing integration test if it could be moved under fixtures, as mentioned here: https://github.com/mochajs/mocha/pull/2037#issuecomment-168526913 A single spec could then be added to `test/integration/regression.js`

bd82:
Hi @danielstjules.

I've replaced the test that reproduced the memory leak, but was slow and created a lot of noise in the test report.

The new one is:
* very fast
* will detect even 'minor' memory leaks without needing to crash V8.

I've kept it in integration tests as:
* The report is small and meaningful (prints path to undeleted 'runnable fn').
* It was written as a user Test, without programmatically creating the suite / runner (directly).

What do you think?

danielstjules:
Looking at your latest commit, imagine the following diff:
``` diff
diff --git a/lib/runnable.js b/lib/runnable.js
index 5e1ab76..70ec96d 100644
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -42,9 +42,11 @@ module.exports = Runnable;
  * @param {string} title
  * @param {Function} fn
  */
+var fns = [];
 function Runnable(title, fn) {
   this.title = title;
   this.fn = fn;
+  fns.push(fn);
   this.async = fn && fn.length;
   this.sync = !this.async;
   this._timeout = 2000;
```
Running `./bin/mocha test/integration/memory.leak.js` on your current branch would result in all tests passing - it would not detect the leak. Your previous integration test, however, correctly identifies the leak.

Would you mind reverting and using the previous integration test, along with my suggestion of placing it in fixtures? A single spec in `test/integration/regression.js`, something like:
`it('#1991: does not leak memory by keeping reference to runnable fns')` would clearly convey the intent, as opposed to:
```
* deletion of runnable fn to avoid memory leaks via closures
  * inner suite
    * access a variable via a closure
```

gumatias:
@bd82 this is awesome! thanks a lot for making a PR for this. really excited for this! hopefully it gets pulled in soon :D

bd82:
@danielstjules: 
I agree that it is better to actually reproduce the V8 out of memory crash.
especially because its harder to argue with 'out of memory' vs some strange test failure
after a new change. :)
And my alternative test indeed will not catch references outside the Suite.

**I'll revert to the original test and put it in a fixture.**

np @gumatias, it was an interesting bug giving a chance to learn some memory leak debugging in javascript. :)

danielstjules:
Thanks again @bd82 Appreciate all your work on this PR :)

bd82:
@danielstjules pleasure to help. :)
Also thanks to @shahar-h who helped debug it.

I've updated the PR to use a fixture with clear description/error messages.
And a large timeout as I don't know how slow it will run on a Travis-CI virtual container + node.js 0.12.
Hopefully this PR is now done :)

danielstjules:
Looks good! Reporters that use test.fn still work, so that's perfect. :)

Thanks again!

danielstjules:
Related issue here: https://github.com/mochajs/mocha/pull/2066

PR #2052, opened 2016-01-14
enable timeouts on a per-test basis if given timeout > 0

### LINKED ISSUE #2051: timeout override
Hi,

When using `this.timeout(0)` on a suite, it disables all timeouts even if an inner test case overrides it.

For example, the following snippet will **pass** and not timeout.

``` javascript
describe('timeout test', function() {
    this.timeout(0);

    it('should timeout', function(done) {
        this.timeout(100);

        setTimeout(function() {
            done();
        }, 1000);
    });
});
```

This will timeout:

``` javascript
describe('timeout test', function() {
    this.timeout(999999);

    it('should timeout', function(done) {
        this.timeout(100);

        setTimeout(function() {
            done();
        }, 1000);
    });
});
```

I think the issue is that there when calling `this.timeout()` with a positive number is doesn't reenable timeouts.

shaharmor:
fixes #2051

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2052)

### DIFF
--- a/test/acceptance/timeout.js
+++ b/test/acceptance/timeout.js
@@ -18,10 +18,21 @@ describe('timeouts', function(){
     }, 300);
   })
 
+  describe('enabling', function() {
+    this.timeout(0);
+
+    it('should allow overriding per-test', function(done) {
+      // uncomment
+      // this.timeout(1);
+      // setTimeout(done, 2);
+      done();
+    });
+  });
+
   describe('disabling', function(){
     it('should allow overriding per-test', function(done){
-      this.enableTimeouts(false);
       this.timeout(1);
+      this.enableTimeouts(false);
       setTimeout(done, 2);
     });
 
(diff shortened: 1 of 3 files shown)


test/acceptance/timeout.js
```diff
@@ -18,10 +18,21 @@ describe('timeouts', function(){
     }, 300);
   })
 
+  describe('enabling', function() {
+    this.timeout(0);
+
+    it('should allow overriding per-test', function(done) {
+      // uncomment
```
ryanshawty: Could you remove these comments?

shaharmor: But then the test will fail because the timeout will throw.

I used the same technique from here: https://github.com/shaharmor/mocha/blob/fix/2051-runnable-timeout/test/acceptance/timeout.js#L3

boneskull: this test should be in `test/runnable.js` and use the strategy there

PR #2059, opened 2016-01-20
Fix #1417: Show actual error, not 'multiple calls to done()'

### LINKED ISSUE #1417: Error in "done() called multiple times" assumptions

frankleonrose:
[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2059)

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -289,6 +289,7 @@ Runnable.prototype.run = function(fn) {
     try {
       callFnAsync(this.fn);
     } catch (err) {
+      emitted = true;
       done(utils.getError(err));
     }
     return;
@@ -308,6 +309,7 @@ Runnable.prototype.run = function(fn) {
       callFn(this.fn);
     }
   } catch (err) {
+    emitted = true;
     done(utils.getError(err));
   }
 
--- a/test/integration/fixtures/regression/issue-1417.js
+++ b/test/integration/fixtures/regression/issue-1417.js
@@ -1,0 +1,22 @@
+'use strict';
+
+/**
+ * This file should generate only one failure per spec for the thrown error.
+ * It should not report misleading 'multiple calls to done()'.
+ */
+
+it('fails exactly once when a global error is thrown synchronously and done errors', function(done) {
+  setTimeout(function() {
+    done(new Error('test error'));
+  }, 1); // Not 0 - it will 'succeed', but won't test the breaking condition
+
+  throw new Error('sync error');
+});
+
+it('fails exactly once when a global error is thrown synchronously and done completes', function(done) {
+  setTimeout(function() {
+    done();
+  }, 1); // Not 0 - it will 'succeed', but won't test the breaking condition
+
+  throw new Error('sync error');
+});
--- a/test/integration/regression.js
+++ b/test/integration/regression.js
@@ -2,6 +2,7 @@ var assert = require('assert');
 var fs     = require('fs');
 var path   = require('path');
 var run    = require('./helpers').runMocha;
+var runJSON    = require('./helpers').runMochaJSON;
 
 describe('regressions', function() {
   this.timeout(1000);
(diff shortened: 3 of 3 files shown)


danielstjules:
This PR seems to enforce what I'd imagine as correct behavior, but want to hear from @mochajs/mocha

danielstjules:
**Before PR**
```
$ ./bin/mocha test/integration/fixtures/regression/issue-1417.js


  ․․․․

  0 passing (14ms)
  4 failing

  1)  fails exactly once when a global error is thrown synchronously and done errors:
     Error: test error
      at null._onTimeout (test/integration/fixtures/regression/issue-1417.js:10:10)

  2)  fails exactly once when a global error is thrown synchronously and done errors:
     Error: test error
      at null._onTimeout (test/integration/fixtures/regression/issue-1417.js:10:10)

  3)  fails exactly once when a global error is thrown synchronously and done completes:
      at Test.Runnable (lib/runnable.js:54:17)
      at new Test (lib/test.js:22:12)
      at context.it.context.specify (lib/interfaces/bdd.js:85:18)
      at Object.<anonymous> (test/integration/fixtures/regression/issue-1417.js:16:1)
      at Array.forEach (native)
      at Object.<anonymous> (bin/_mocha:403:18)
      at node.js:868:3

  4)  fails exactly once when a global error is thrown synchronously and done completes:
      at Test.Runnable (lib/runnable.js:54:17)
      at new Test (lib/test.js:22:12)
      at context.it.context.specify (lib/interfaces/bdd.js:85:18)
      at Object.<anonymous> (test/integration/fixtures/regression/issue-1417.js:16:1)
      at Array.forEach (native)
      at Object.<anonymous> (bin/_mocha:403:18)
      at node.js:868:3
```

**After PR**
```
$ ./bin/mocha test/integration/fixtures/regression/issue-1417.js


  ․․

  0 passing (12ms)
  2 failing

  1)  fails exactly once when a global error is thrown synchronously and done errors:
     Error: sync error
      at Context.<anonymous> (test/integration/fixtures/regression/issue-1417.js:13:9)
      at callFnAsync (lib/runnable.js:340:8)
      at Test.Runnable.run (lib/runnable.js:290:7)
      at Runner.runTest (lib/runner.js:422:10)
      at lib/runner.js:529:12
      at next (lib/runner.js:342:14)
      at lib/runner.js:352:7
      at next (lib/runner.js:284:14)
      at Immediate._onImmediate (lib/runner.js:320:5)

  2)  fails exactly once when a global error is thrown synchronously and done completes:
     Error: sync error
      at Context.<anonymous> (test/integration/fixtures/regression/issue-1417.js:21:9)
      at callFnAsync (lib/runnable.js:340:8)
      at Test.Runnable.run (lib/runnable.js:290:7)
      at Runner.runTest (lib/runner.js:422:10)
      at lib/runner.js:529:12
      at next (lib/runner.js:342:14)
      at lib/runner.js:352:7
      at next (lib/runner.js:284:14)
      at Immediate._onImmediate (lib/runner.js:320:5)
```

danielstjules:
It technically masks the fact that done() is invoked after an error is thrown, but I think that's fine. At that point, the test has already failed.

boneskull:
:+1: in general this is a good idea, but @frankleonrose can you add a test which uses promises?  because a Promise won't throw, `emitted` won't be set to `true`, so is it possible we could still get the misleading error?

frankleonrose:
@boneskull I don't think there's a way to cause this error with promises. The tests written require both failing on the sync path and calling done() on the async path to trigger the "double call" warning. Promises don't have a second way in to the termination code so they can't fail this way.

boneskull:
makes sense.

boneskull:
@frankleonrose thanks! merged

Issue #2040, opened 2016-01-06
New version of Mocha requested


klamping:
The last release of Mocha was on Nov. 15th and many improvements have been added since then. Specifically for my needs, [the `retry` functionality](https://github.com/mochajs/mocha/pull/1989) was introduced after the latest release. Can we get a new release of Mocha to take advantage of these changes?

Another request for a release was made here: https://groups.google.com/forum/#!topic/mochajs/Olw3berArnA (so i'm not the only one :smile:)

longlho:
:+1:

LordWingZero:
![](https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png)

danielstjules:
In a few days? I'd like to close up some possible bugs and include fixes in the minor release.

LordWingZero:
Sounds good to me.

boneskull:
@danielstjules Please let me know when ready

cc @mochajs/mocha

robraux:
:+1: looking forward to it, thank you

TrejGun:
:+1:

danielstjules:
Seems to be issues with the HTML reporter in master - investigating now

danielstjules:
Sorry for the delay. imo, this should be fixed before the next release: https://github.com/mochajs/mocha/issues/1760

jonny-improbable:
Sorry to bump this issue but it's been a week since the last comment; a 2.3.5 release would be greatly appreciated.

boneskull:
We're stuck on a browser-related issue afaik.  @danielstjules know what's up here?

antinescience:
@danielstjules same thing here - would love to see 2.3.5 hit npm!

klamping:
Any updates on this? Maybe y'all can put out a prerelease so I can move forward (since the bug doesn't affect my needs).

danielstjules:
This is the current blocker. The browser reporter in master (run `make`) is not functional

![screen shot 2016-01-25 at 11 05 07 pm](https://cloud.githubusercontent.com/assets/817212/12574374/3b5c1f0e-c3b8-11e5-93d0-b9b55492b545.png)

danielstjules:
Done :+1:

klamping:
Awesome, thanks! Can't wait to try out the new features!

PR #2066, opened 2016-01-26
Fix test fn's being deleted too early for browser reporter

danielstjules:
Fixes the issue reported here: https://github.com/mochajs/mocha/issues/2040#issuecomment-174864558

screen shot 2016-01-25 at 11 07 05 pm

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2066)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -726,17 +726,8 @@ Runner.prototype.uncaught = function(err) {
 };
 
 /**
- * Cleans up the reference to the test's deferred function.
- * @see cleanSuiteReferences for details.
- * @param {Test} test
- */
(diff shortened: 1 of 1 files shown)


chrmod:
confirming that fix works. 

Though there is something strange about it as I see one failed test but two error reports (on html reporter) - previously I've seen two failed tests and one report.

Have noticed that on master 'fail' hook is called multiple times.

danielstjules:
@chrmod Yup, that's fixed via https://github.com/mochajs/mocha/pull/2068 :)

chrmod:
great job ! :-)

Issue #2067, opened 2016-01-26
HTML/DOC reporter regressions as a result of #2037


danielstjules:
https://github.com/mochajs/mocha/pull/2037

As a result of that PR, the browser & doc reporters cannot handle showing function bodies when exceptions are thrown async, outside of the duration of the originating suite (after the suite has ended). They'll result in: `Uncaught TypeError: Cannot read property 'toString' of undefined`

The line in question:
``` javascript
var pre = fragment('<pre><code>%e</code></pre>', utils.clean(test.fn.toString()));
```

Rather than revert https://github.com/mochajs/mocha/pull/2037, I propose caching the function body strings, which should still be light.

francisco-sanchez-molina:
Hi @danielstjules @boneskull! After _test.fn.toString()_ to _test.body_ change, when "before" throw Error mocha explode in mocha.js:5980 Uncaught TypeError: Cannot read property 'replace' of undefined

2.4.4 version http://jsfiddle.net/38v388kw/

2.3.4 version http://jsfiddle.net/joes4L1q/

asadsahi:
any fix for this? having same issue.

PR #2068, opened 2016-01-26
Fix 2067: HTML/DOC reporter regression with async failures

danielstjules:
Fixes https://github.com/mochajs/mocha/issues/2067

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2068)

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -22,6 +22,7 @@ function Test(title, fn) {
   Runnable.call(this, title, fn);
   this.pending = !fn;
   this.type = 'test';
+  this.body = (fn || '').toString();
 }
 
 /**
--- a/lib/reporters/doc.js
+++ b/lib/reporters/doc.js
@@ -49,13 +49,13 @@ function Doc(runner) {
 
   runner.on('pass', function(test) {
     console.log('%s  <dt>%s</dt>', indent(), utils.escape(test.title));
-    var code = utils.escape(utils.clean(test.fn.toString()));
+    var code = utils.escape(utils.clean(test.body));
     console.log('%s  <dd><pre><code>%s</code></pre></dd>', indent(), code);
   });
 
   runner.on('fail', function(test, err) {
     console.log('%s  <dt class="error">%s</dt>', indent(), utils.escape(test.title));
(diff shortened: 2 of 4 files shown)


boneskull:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mochajs/mocha/2068)*

lib/test.js
```diff
@@ -22,6 +22,7 @@ function Test(title, fn) {
   Runnable.call(this, title, fn);
   this.pending = !fn;
   this.type = 'test';
+  this.body = (fn || '').toString();
```
chrmod: Should `clone` copy this one as well?

danielstjules: No - when the new test is created via `Test#clone`, it'll get the exact same string body https://github.com/danielstjules/mocha/blob/d59cc6c166d0d112a62016909b33e04ed08d2274/lib/test.js#L34

danielstjules: Same reason why `Test#clone` doesn't apply `this.type = 'test';`. It's already handled in the constructor :)

chrmod: right :+1:

lddubeau: Shouldn't the initialization of `this.body` belong to `Runnable`'s constructor rather than `Test`?

Right now, if you get an error in a hook, Mocha will bomb, because the `Hook` object that it tries to process [here](https://github.com/mochajs/mocha/pull/2068/files#diff-998e3fb668a2094eedccab3e56199e4fR203) does not have a `body` field. Moving the creation of `body` to `Runnable` would make both `Hook` and `Test` get a `body` field.

The issue I'm talking about here is fully reproducible with the code provided in [this comment](https://github.com/mochajs/mocha/issues/2067#issuecomment-181430571). I also am able to reproduce it independently in a test suite of mine (which is why I found the comment and am commenting here).

danielstjules: This will be resolved by https://github.com/mochajs/mocha/pull/2115 Thanks! :)

Issue #2063, opened 2016-01-22
--delay option causes incorrect nesting of suites/tests


sa-0001:
When I recursively run multiple files using the --delay option, all suites and tests are mixed together in one hierarchy and run out of order.

Expected result from running recursively three files, each of which contains two suites, each of which contains one or two tests:

```
  file I, suite A
    ✓ test 01
    ✓ test 02
    file I, suite B
      ✓ test 03
  file II, suite C
    ✓ test 04
    ✓ test 05
    file I, suite D
      ✓ test 06
  file III, suite E
    ✓ test 07
    ✓ test 08
    file III, suite F
      ✓ test 09
```

Actual result:

```
  file I, suite A
    file II, suite C
      file III, suite E
        ✓ test 01
        ✓ test 08
        file I, suite B
          ✓ test 04
          ✓ test 05
          ✓ test 09
          file II, suite D
            ✓ test 07
            ✓ test 02
            ✓ test 06
            file III, suite F
              ✓ test 03
```

I tried removing run() from all files but the last, however the result is the same.  In fact if you look carefully even the test order within a given file is not respected due to a later test being placed in a higher suite than an earlier test!

While the asynchronous test definition of a single file works correctly with --delay, suites and test from different files become interwoven I believe due to this code:

``` javascript
Mocha.prototype.loadFiles = function(fn) {
  ...
  this.files.forEach(function(file) {
    ...
    suite.emit('require', require(file), file, self);
    ...
  });
};
```

Does anyone know of a way to also _require_ the test files asynchronously, or to otherwise prevent a describe/it in one file from being attached to a describe/it in another file?

hollomancer:
@sa-0001 - do you have more details / source files on how you were able to get this to happen?

PR #2069, opened 2016-01-26
(Issue #2063) proposal for asynchronous require when using --delay

sa-0001:
This is how I have locally solved the issue I created (https://github.com/mochajs/mocha/issues/2063).

If this change would affect other features or cause other problems I am not aware I would appreciate feedback on how I can improve it; otherwise for my specific case I have tested the changes and calling run() in each file causes the next file in sequence to be required, or the root suite to run if all files have been required.

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2069)


boneskull:
this sort of thing can be managed with the new `--file` option

PR #2072, opened 2016-01-26
Fix #2071: Removed color code to fix Chalk colors

### LINKED ISSUE #2071: Colors not working correctly
In version 2.3.4 the colors were working fine. The commit that  added in Chalk doesn't work.

Mocha 2.3.4: Mocha 2.3.4

Mocha 2.4.1: Mocha 2.4.1

Removing the String ` '\u001b[' + `  from the `return '\u001b[' + exports.colors[type] + 'm' + str + '\u001b[0m';` , the output shows fine as seen here:

Mocha 2.4.1 Fixed

Using cmder with powershell on Windows 10

Currently works fine as well on Linux and removing the code also works fine.

thedark1337:
Removed extraneous color code that broke output on Windows as shown in https://github.com/mochajs/mocha/issues/2071

[Review on Reviewable](https://reviewable.io/reviews/mochajs/mocha/2072)

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -105,7 +105,7 @@ var color = exports.color = function(type, str) {
   if (!exports.useColors) {
     return String(str);
   }
-  return '\u001b[' + exports.colors[type](str) + '\u001b[0m';
+  return exports.colors[type](str);
 };
 
 /**


danielstjules:
I wonder if chalk handles `\u001b[0m` for us, making resetting to the normal color unnecessary. But I guess given our lack of test coverage around this, it's best to keep the changeset small.

thedark1337:
From reading the source code of what Chalk uses , it appears to reset already for the open / close. Testing with my mocha tests for my ESLInt config, it seems that is true. Should I remove it as well and amend the commit?

danielstjules:
Sure, I'll just wanna test with a bunch of reporters prior to giving a thumbs up

jbnicolai:
> From reading the source code of what Chalk uses , it appears to reset already for the open / close

Yeah, having worked on Chalk, I can definitely confirm that is the case. Both the opening and closing escape can be removed.

> Should I remove it as well and amend the commit?

Yes please :)

> Sure, I'll just wanna test with a bunch of reporters prior to giving a thumbs up
 Show all checks

Absolutely! Although I can confidently say that there's no need to do any extra escaping/wrapping around Chalk.

thedark1337:
Amended the commit to remove the reset code. :+1:

boneskull:
@thedark1337 thanks!

PR #2086, opened 2016-02-02
Fix: html reporter crashing on before hook error

chrmod:
HTML reporter expects that Runnables have a `body` 

https://github.com/mochajs/mocha/blob/d59cc6c166d0d112a62016909b33e04ed08d2274/lib/reporters/html.js#L203


danielstjules:
Thanks for the PR! Not sure why there's a merge conflict, so I'm gonna go ahead and merge https://github.com/mochajs/mocha/pull/2086 isntead

Issue #2089, opened 2016-02-03
Diff doesn't be displayed when object has Symbol value


ryym:
When I assert an object that has a `Symbol` value, the test doesn't display diff even if it fails. For example, the following test fails without generating diff.

```javascript
// test.js
const assert = require('assert');

it('should generate diff', () => {
  const obj = { type: Symbol('foo') };
  assert.deepEqual(obj, { type: null });
});
```
result:

![mocha-symbol-before](https://cloud.githubusercontent.com/assets/12684251/12785091/01ee67a0-cace-11e5-83f9-34577f1872cb.png)

Then I found the `stringify` function in [lib/utils](https://github.com/mochajs/mocha/blob/master/lib/utils.js) throws a following error when the argument object has a `Symbol` value.

```
TypeError: Cannot convert a Symbol value to a string
```

environment:

* mocha: v2.4.5
* node: v4.2.2

glenjamin:
I just ran into this, the `json-stream` and `tap` reports seem to hand this ok, so it's definitely a reporter problem - although the lack of an error is slightly more worrying

ryym:
It seems the reporters like `dot` and `spec` which use [this function] cause this problem. The function generates diff by `util.stringify()` to convert Symbol to string and throws the error.
I sent [PR] for this issue :)

[this function]: https://github.com/mochajs/mocha/blob/2a8594424c73ffeca41ef1668446372160528b4a/lib/reporters/base.js#L299
[PR]: https://github.com/mochajs/mocha/pull/2094

glenjamin:
Good stuff, is it possible to fix the black-holed error too?

ryym:
> is it possible to fix the black-holed error too?

Hmm, I don't know how to fix this. My PR just adds handling of `Symbol` values, so it does nothing about the black-holed error. But as you say, it should be fixed too.

glenjamin:
I've created a draft PR which attempts to solve the swallowed-error problem: #2096

ryym:
Great!

PR #2094, opened 2016-02-06
handle Symbol values in util.stringify()

ryym:
This PR fixes issue #2089.
It enables to generate diff of objects that have Symbol values.

Fixed output:

mocha-symbol-after

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -472,6 +472,7 @@ function jsonStringify(object, spaces, depth) {
         break;
       case 'boolean':
       case 'regexp':
+      case 'symbol':
       case 'number':
         val = val === 0 && (1 / val) === -Infinity // `-0`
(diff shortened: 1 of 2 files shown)


ryym:
The cause of this issue is [this line] of `lib/utils.js`. According to [Symbol type conversions], We have to call `toString()` explicitly to convert Symbol to string.

[this line]: https://github.com/mochajs/mocha/blob/2a8594424c73ffeca41ef1668446372160528b4a/lib/utils.js#L603
[Symbol type conversions]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Symbol_type_conversions

danielstjules:
Looks good, thank you!

spalger:
Would love to see this released :)

PR #2096, opened 2016-02-07
Ensure that errors happening after the test suite is done still get thown

glenjamin:
Most notably this happens if there are problems in a reporter's summary output, such as that described in #2089

I'd like to include a test for this - but I'm not sure where best to put it. Any thoughts?

Here's a sample which will reproduce the error even after #2094 is fixed:

``` js

describe('x', function() {
  it('should', function() {
    var d = new Date();
    // Force an error during rendering
    d.toISOString = function() { throw new Error("Can't display"); };
    assert.deepEqual([d], ['a']);
  });
});
```

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -67,6 +67,7 @@ function Runner(suite, delay) {
   this._delay = delay;
   this.suite = suite;
   this.started = false;
+  this.finished = false;
   this.total = suite.total();
   this.failures = 0;
   this.on('test end', function(test) {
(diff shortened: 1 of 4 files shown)


danielstjules:
> I'd like to include a test for this - but I'm not sure where best to put it. Any thoughts?

If it occurs in all reporters, you could add a fixture and integration spec that tested for the presence of the error.

glenjamin:
I've added a test into the regression section, as that seemed to be where it fit best.

I was briefly going to add tests into `runner.js`, as that's where the fix is - but couldn't find a good way to tie it in. I did start adding some useful tests before giving up on that approach - I've left those in this PR as I think they don't hurt, but I'm happy to remove them if desired.

boneskull:
this was fixed in v5.0.2

PR #2097, opened 2016-02-07
fix diff for objects with custom hasOwnProperty values

mantoni:
Instead of calling `hasOwnProperty` directly, it should be `Object.prototype.hasOwnProperty.call(object, prop)`. This is already the case further up in the `exports.stringify` implementation.


danielstjules:
Awesome, thank you!

PR #2101, opened 2016-02-09
Non string error messages

jkimbo:
This PR checks that when handling assertion error messages in a reporter, the message is actually a string since if it isn't the `message.match` function (https://github.com/jkimbo/mocha/blob/2a1dc7a8f32d7eb65cdf645c30aac3cc37714ae7/lib/reporters/base.js#L208) fails and leads to the test suite failing silently.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -170,8 +170,8 @@ exports.list = function(failures) {
     var msg;
     var err = test.err;
     var message;
-    if (err.message) {
-      message = err.message;
+    if (err.message && typeof err.message.toString === 'function') {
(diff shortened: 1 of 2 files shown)


danielstjules:
How about:
``` javascript
if (err.message && typeof err.message.toString === 'function') {
  message = err.message + '';
} ...
```

jkimbo:
@danielstjules great idea! I'll change it.

danielstjules:
Thanks! Could you rebase/squash? :)

jkimbo:
@danielstjules done!

danielstjules:
Thanks!

PR #2104, opened 2016-02-11
Consistent assertion formatting

avdd:
Doing so wraps the error message in the color format string for the
error title, so that if the error message title were set to another
color code, e.g. '31;1' (bright red), that would leak into the error
message.

Also there was an extra newline before the error message when showing a
diff that wouldn't appear when there is no diff.

Before:
before

After:
after


boneskull:
@avdd Thanks!  Can you please sign the CLA?

Issue #2079, opened 2016-01-27
Add browser test suite to CI


danielstjules:
Should hopefully catch some issues going forward. Very simple cross-browser tests with testling.com/saucelabs would be nice too.

boneskull:
this is necessary before we do any release post-2.4.5.

danielstjules:
Any preference on how this should be done?

* Testling could be configured to use mocha's latest master as a dev dependency https://ci.testling.com/guide/mocha However, it would make testing a bit awkward, and wouldn't work on PRs. For example, how could you test failures/reporter-output/etc?
* Saucelabs with http://webdriver.io/ would certainly work, and could be used to test reporter output in different conditions (failure, success, etc) and browsers. However, it can't run on PRs. Possible workaround: https://github.com/twbs/savage
* casperjs or https://github.com/segmentio/nightmare for webkit testing. Faster, less prone to sporadic network failures as with Testling/Saucelabs. Would work on PRs. But wouldn't give us cross-browser coverage

danielstjules:
How much do we value cross-browser integration/acceptance tests over just plain-old automated browser tests in, e.g. phantomjs? Most of our IE8 issues thus far could have been detected with a grep (it's almost always `Array.prototype.forEach` lol) Aside from that, we've hit some small snags with Safari in the past.

We could certainly kill two birds with one stone, but I feel like the initial focus might be just hooking up basic browser tests with CI?

tschaub:
Cross browser testing would be nice.  But even testing in Phantom would help.  You could take the two tasks on separately.  Phantom only tests could be done on Travis.



> Saucelabs with http://webdriver.io/ would certainly work, and could be used to test reporter output in different conditions (failure, success, etc) and browsers. However, it can't run on PRs.

You can use [Sauce Connect on Travis](https://docs.travis-ci.com/user/sauce-connect/) and have your tests run in as many browsers as you want on every commit (PRs included).

You'll find the same thing elsewhere, but here's an example [Karma](https://github.com/planetlabs/planet-client-js/blob/master/test/karma.config.js) and [Travis](https://github.com/planetlabs/planet-client-js/blob/master/.travis.yml) config and [`package.json` scripts](https://github.com/planetlabs/planet-client-js/blob/master/package.json) for running tests in Node + Browsers (locally and via Sauce Connect in Travis).

danielstjules:
@tschaub From your link at https://docs.travis-ci.com/user/sauce-connect/
[...]

PR #2112, opened 2016-02-16
Fix #2083: HTML reporter regression causing duplicate error output

### LINKED ISSUE #2083: Regression: Failures rendering twice
I just happened to start up a simple hello world mocha test project (to test something else :) ), and I noticed that versions >= 2.4 now double-render the failure in my example.  Reverting to 2.3.3 makes the behavior go away.  I will try to get to the bottom of it, but I see several releases in the last few days so I thought maybe you'd want to know.

https://github.com/ahamid/mocha-phantomjs-blanket/tree/mocha-failure-double-render

double-render-failure

danielstjules:
Fixed #2083

This is a minor refactor of the HTML reporter to listen to `pass/fail/pending` events rather than `test end`, just like the spec/dot reporters.

--------
As seen below, the sync error is displayed only a single time:

--------

screen shot 2016-02-16 at 12 16 50 am

--------

And the async errors are each correctly displayed. This is the behavior https://github.com/mochajs/mocha/pull/1981 was intended to fix. Prior to that PR, none of the errors below were displayed in the HTML reporter.

--------

screen shot 2016-02-16 at 12 16 57 am

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -128,16 +128,70 @@ function HTML(runner) {
     stack.shift();
   });
 
-  runner.on('fail', function(test) {
-    // For type = 'test' its possible that the test failed due to multiple
-    // done() calls. So report the issue here.
-    if (test.type === 'hook'
-      || test.type === 'test') {
-      runner.emit('test end', test);
-    }
+  runner.on('pass', function(test) {
+    var url = self.testURL(test);
+    var markup = '<li class="test pass %e"><h2>%e<span class="duration">%ems</span> '
+      + '<a href="%s" class="replay">‣</a></h2></li>';
+    var el = fragment(markup, test.speed, test.title, test.duration, url);
+    self.addCodeToggle(el, test.body);
+    appendToStack(el);
+    updateStats();
   });
 
-  runner.on('test end', function(test) {
+  runner.on('fail', function(test) {
+    var el = fragment('<li class="test fail"><h2>%e <a href="%e" class="replay">‣</a></h2></li>',
+      test.title, self.testURL(test));
+    var stackString; // Note: Includes leading newline
+    var message = test.err.toString();
+
+    // <=IE7 stringifies to [Object Error]. Since it can be overloaded, we
+    // check for the result of the stringifying.
+    if (message === '[object Error]') {
+      message = test.err.message;
+    }
+
+    if (test.err.stack) {
+      var indexOfMessage = test.err.stack.indexOf(test.err.message);
+      if (indexOfMessage === -1) {
+        stackString = test.err.stack;
+      } else {
+        stackString = test.err.stack.substr(test.err.message.length + indexOfMessage);
+      }
+    } else if (test.err.sourceURL && test.err.line !== undefined) {
+      // Safari doesn't give you a stack. Let's at least provide a source line.
+      stackString = '\n(' + test.err.sourceURL + ':' + test.err.line + ')';
+    }
+
+    stackString = stackString || '';
+
+    if (test.err.htmlMessage && stackString) {
+      el.appendChild(fragment('<div class="html-error">%s\n<pre class="error">%e</pre></div>',
+        test.err.htmlMessage, stackString));
+    } else if (test.err.htmlMessage) {
+      el.appendChild(fragment('<div class="html-error">%s</div>', test.err.htmlMessage));
+    } else {
+      el.appendChild(fragment('<pre class="error">%e%e</pre>', message, stackString));
+    }
+
+    self.addCodeToggle(el, test.body);
+    appendToStack(el);
+    updateStats();
+  });
+
+  runner.on('pending', function(test) {
+    var el = fragment('<li class="test pass pending"><h2>%e</h2></li>', test.title);
+    appendToStack(el);
+    updateStats();
+  });
+
+  function appendToStack(el) {
+    // Don't call .appendChild if #mocha-report was already .shift()'ed off the stack.
+    if (stack[0]) {
+      stack[0].appendChild(el);
+    }
+  }
+
+  function updateStats() {
     // TODO: add to stats
     var percent = stats.tests / this.total * 100 | 0;
     if (progress) {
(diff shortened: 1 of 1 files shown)


danielstjules:
cc @boneskull

danielstjules:
This is compatible with https://github.com/mochajs/mocha/pull/2081 since the `pending` event now fires after checking `isPending()`

boneskull:
@danielstjules cool.  do you have any recommendations for how to test this?

boneskull:
#2079 is still a thing.

boneskull:
I *really* don't want to release anything without #2079 being addressed in some fashion

danielstjules:
Aside from manual testing? If you open test/browser/index.html in master, you'll see that normal errors are duplicated, but async errors after a test completed are correctly shown:

screen shot 2016-02-17 at 9 26 24 am

screen shot 2016-02-17 at 9 26 36 am

Comparing that against the build of this PR/branch, async errors post-test completion are still shown, and normal errors are only rendered once.

danielstjules:
In regards to the async errors, this is the same expected behavior seen from the node reporters:

``` javascript
$ cat example.js
it('test', function(done) {
  setTimeout(done, 10);
  setTimeout(done, 20);
});

it('test2', function(done) {
  setTimeout(done, 30);
});

$ ./bin/mocha example.js


  ․․․

  2 passing (51ms)
  1 failing

  1)  test:
      at Test.Runnable (lib/runnable.js:54:17)
      at new Test (lib/test.js:22:12)
      at context.it.context.specify (lib/interfaces/bdd.js:85:18)
      at Object.<anonymous> (example.js:1:63)
      at Array.forEach (native)
      at Object.<anonymous> (bin/_mocha:403:18)
      at node.js:868:3
```

danielstjules:
> I really don't want to release anything without #2079 being addressed in some fashion

I can work on that tonight. But I don't think I'll have an exhaustive enough automated browser test suite ready to reveal issues beyond manually running those HTML files in IE8/Safari/Chrome.

boneskull:
at minimum then I'd want someone to test IE8, phantomjs 1.x and phantomjs 2.x manually.  specifically, see if mochify breaks.

jnbt:
Any updates on this? I think the HTML reports are still broken...

justinfagnani:
Also curious about this PR, the bug is messing with reporting downstream of mocha. Are all PRs on hold until #2079 is resolved?

dasilvacontin:
@justindujardin pretty much, sorry for the hassle.

justinfagnani:
Honestly, this makes me question our dependence on Mocha. This fix is 2 months old.

boneskull:
@justinfagnani Hi!  Instead of complaining, feel free to help with #2079.

ScottFreeCode:
I've been experimenting with a set of tests to check for problems like this. It isn't pretty, I don't know if it will work in any kind of automatic testing such as discussed in #2112, and the CLI won't work unless I can get the same HTML report as the browser uses; but at least the set of test cases should be a good starting point (although you may have to go through and tweak which situations are intended to be reported as how many passes and failures, I'm not 100% sure I got it right). It's also not really unit-testing the matter -- if there's the wrong number of results there's no indication as to whether it's a flaw in the test runner or the reporter, if there's the right number of results that doesn't mean the responsibilities involved in getting them are correctly divided between the test runner and the reporter -- but it's better than nothing, I figure (I guess if we wanted to unit test Mocha's internals we'd have to use the non-bundled scripts... and to make it run in the browser would probably end up basically creating a second bundle containing the internal unit tests?...).

I've put what I've got so far up in a gist here: https://gist.github.com/ScottFreeCode/f8b0864781414f96594f67b77ddee469

maciejjankowski:
I wonder why is this stalled...
shouldn't you add `PR Please` label?

pluma:
I can see why you would want to fix #2079 first but this change is trivial compared to the rabbit hole that is automated browser testing. I believe greenlighting this fix despite the freeze would help a lot of people. As it is, this bug is making browser reports significantly less useful, so the question of browser support doesn't really come into it.

boneskull:
I'm going to rebase this onto `master` and see what happens.

lib/reporters/html.js
```diff
@@ -128,88 +128,82 @@ function HTML(runner) {
...
+  function appendToStack(el) {
     // Don't call .appendChild if #mocha-report was already .shift()'ed off the stack.
     if (stack[0]) {
       stack[0].appendChild(el);
     }
-  });
+  }
+
+  function updateStats() {
+    // TODO: add to stats
```
boneskull: @danielstjules does this TODO mean this PR isn't ready?

boneskull:
In general, I think this looks good. Just want some feedback from @danielstjules to ensure it's ready to merge.

lib/reporters/html.js
```diff
@@ -128,88 +128,82 @@ function HTML(runner) {
...
-      progress.update(percent).draw(ctx);
+  runner.on('fail', function(test) {
+    var el = fragment('<li class="test fail"><h2>%e <a href="%e" class="replay">‣</a></h2></li>',
+      test.title, self.testURL(test));
+    var stackString; // Note: Includes leading newline
+    var message = test.err.toString();
+
+    // <=IE7 stringifies to [Object Error]. Since it can be overloaded, we
+    // check for the result of the stringifying.
+    if (message === '[object Error]') {
```
dasilvacontin: In the comment you mention `[Object Error]`, but in the code it uses a different casing.

pluma: @dasilvacontin the code is correct, the comment is wrong.

dasilvacontin: Thanks for clearing it up @pluma!

dasilvacontin:
LGTM. Sync error is only reported once now. Waiting for the reply about the `TODO`, will release in a day or two if no response is given.

dasilvacontin:
btw: sorry for the wait internet people.

lib/reporters/html.js
danielstjules: If you look further down, I didn't write the comment, just moved around the code :)

lib/reporters/html.js
danielstjules: The comment already existed above https://github.com/mochajs/mocha/pull/2112/files#diff-998e3fb668a2094eedccab3e56199e4fL141

danielstjules:
The diff is unfortunately long for what was a minor refactor. A lot of the code was moved around rather than modified.

lib/reporters/html.js
dasilvacontin: @danielstjules true :)

lib/reporters/html.js
dasilvacontin: @danielstjules true again

dasilvacontin:
Thanks for the fix @danielstjules :). I released it already.

Issue #2111, opened 2016-02-16
Deprecated graceful-fs


thedark1337:
Hi, I just noticed after installing mocha that it relies on an outdated dependency on glob@3.2.3 which uses graceful-fs@2.0.3 .  

`npm WARN deprecated graceful-fs@2.0.3: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible`

See the tracking issue here: https://github.com/nodejs/node/issues/5213

LinusU:
Would be nice to update glob, the latest version is `7.0.0`...

danielstjules:
Please see: https://github.com/mochajs/mocha/issues/2058#issuecomment-173077615

This won't be needed until Node 7.0. Furthermore, we can't break support for older version of npm atm. This is something we could pursue in a major release at some point.

LinusU:
Hmm, that's to bad :(

Although Node 0.8 is at end of life, it isn't supported anymore. I think it's time to drop support for that from mocha...

danielstjules:
That's something we can def look into in the next major release :) We wouldn't want to break compatibility in 2.x.x

LinusU:
I understand :+1:

Issue #2058, opened 2016-01-19
Upgrade glob to latest


gyandeeps:
Mocha is using `glob` module version from 2013. I would be good if we can upgrade that to latest so that we can use a lot of globbing patterns with it.

If its ok, then I can send the PR.

danielstjules:
We've previously experienced issues with upgrading dependencies: https://github.com/mochajs/mocha/issues/1867 https://github.com/mochajs/mocha/issues/1887 https://github.com/mochajs/mocha/issues/1868

Aside from a bump, we'd want to verify compatibility with old versions of node/npm.

gyandeeps:
What is our official nodejs support range?
If its too wide then this change might cause issue (but i have to try it).

boneskull:
this is not possible as any glob newer than the one we currently have uses fuzzy version matching against minor versions (`^`) which precludes nodejs 0.8.  

I imagine 0.8 will be dropped w/ next major, however.

closing for now

boneskull:
(it doesn't *actually* preclude using 0.8; you can generally still upgrade npm, but some people cannot for various reasons)

danez:
The problem on the other hand is that this old glob, depends on an old version of graceful-fs. And this one is using node apis which are getting removed in the next major. So either beeing backward compatible or forward :/

From install:
```
npm WARN deprecated graceful-fs@3.0.8: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible.
```

danielstjules:
Sounds like it won't get removed until v7: https://github.com/nodejs/node/issues/5213#issuecomment-183633565 But I agree that we can look at breaking compatibility with the next major mocha release :) https://github.com/mochajs/mocha/issues/2111#issuecomment-184599809

isiahmeadows:
If you're testing against 0.8, graceful-fs was rewritten in 4.x to [support 0.8 better](https://github.com/isaacs/node-graceful-fs/issues/47), and it [does test against that version in CI](https://github.com/isaacs/node-graceful-fs/blob/master/.travis.yml). And apparently, [not much changed between 3.x and 4.x](https://github.com/isaacs/node-graceful-fs/issues/470) other than its monkey-patching-by-default behavior (which is likely to break something).

Most of the breakage you might encounter would probably be in node-glob, not graceful-fs, unless you unintentionally rely on graceful-fs' monkey-patching.

danielstjules:
[...]

PR #2113, opened 2016-02-16
package: update glob

jonathanong:
older versions of glob use older version of `graceful-fs`, which will break on newer (future, i assume) versions of node

### DIFF
--- a/package.json
+++ b/package.json
@@ -278,7 +278,7 @@
     "debug": "2.2.0",
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
-    "glob": "3.2.3",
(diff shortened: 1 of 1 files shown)


jonathanong:
btw if u guys are gonna pin dependencies, would be nice to use greenkeeper so you guys stay up to date

danielstjules:
Closing since it breaks compatibility with older versions of NPM. See https://github.com/mochajs/mocha/issues/2058#issuecomment-173077615 and https://github.com/mochajs/mocha/issues/2111

I think we can def bump it in the next major release :) Just not sure when that'll be, and targeting the 3.0.0 branch might not be wort it since it's in limbo.

Issue #270, opened 2012-02-17
browser - exceptions thrown in before/beforeEach are ignored


spmason:
I'm finding that exceptions thrown in my before/beforeEach blocks aren't being handled, it's simple enough to reproduce:

        mocha.setup('bdd');
        describe('testing the tests', function(){
            beforeEach(function(){ // also before(), 
                throw new Error('this should fail');
            });

            it('Works', function(){
            });
        });
        mocha.run();

This is on the most recent version, 0.12.1 on Chrome 17 and FireFox 10

danielstjules:
Not sure when the issue was fixed, but it's no longer a problem in 2.2.0 :)

``` javascript
// When thrown in beforeEach

$ cat beforeEachError.js
describe('issue-270', function() {
  before(function() {
    console.log('before hook');
  });

  beforeEach(function() {
    console.log('beforeEach hook');
    throw new Error('beforeEach hook error');
  });

  afterEach(function() {
    console.log('afterEach hook');
  });

  after(function() {
    console.log('after hook');
  });

  it('test', function() {});
});

// Output

$ mocha beforeEachError.js

  before hook
beforeEach hook

  ․afterEach hook
after hook


  0 passing (6ms)
  1 failing

  1) issue-270 "before each" hook:
     Error: beforeEach hook error
      at Context.<anonymous> (beforeEachError.js:8:11)

// An example when thrown in before hook

$ cat beforeError.js
describe('issue-270', function() {
  before(function() {
    console.log('before hook');
    throw new Error('before hook error');
  });

  beforeEach(function() {
    console.log('beforeEach hook');
  });

  afterEach(function() {
    console.log('afterEach hook');
    invoked++;
  });

  after(function() {
    console.log('after hook');
  });

  it('test', function() {});
});

// Output

$ mocha beforeError.js

  before hook

  ․after hook


  0 passing (6ms)
  1 failing

  1) issue-270 "before all" hook:
     Error: before hook error
      at Context.<anonymous> (beforeError.js:4:11)
```

bitmage:
Wonderful, thanks!  I just didn't want it to slip under the radar.

danielstjules:
No problem! Happy it all worked out.

bitwiseman:
Fantastic.

PR #2115, opened 2016-02-17
Fix hook error in browser reporter by moving body prop from test to runnable

danielstjules:
Fixes `Uncaught TypeError: Cannot read property 'replace' of undefined` when an exception is thrown from a hook.

Example test suite that correctly renders after the PR:
``` javascript
before(function() {
  throw new Error('uh oh!');
});
it('test', function() {});
```

screen shot 2016-02-16 at 6 25 46 pm

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -45,6 +45,7 @@ module.exports = Runnable;
 function Runnable(title, fn) {
   this.title = title;
   this.fn = fn;
+  this.body = (fn || '').toString();
   this.async = fn && fn.length;
   this.sync = !this.async;
   this._timeout = 2000;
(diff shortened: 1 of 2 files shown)


andrei-picus-hs:
Why hasn't this been published to npm?!

screendriver:
+1 for a release because I can't use [Test'em](https://github.com/testem/testem)s CI feature. My CI builds are hanging forever because the browser will not be closed when mocha isn't failing correctly 😭

stevebartholomew:
Presume this is a regression as the same issue was fixed a while ago: https://github.com/mochajs/mocha/issues/270 - +1 for release :)

PR #2119, opened 2016-02-18
prevent html reporter updating url hash

jimenglish81:
I have noticed a small issue with the hash updating when the passed/failure filters are clicked on the html reporter. When testing an app that uses hash navigation this can cause the page to navigate.
Let me know thoughts. Thanks

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -80,7 +80,8 @@ function HTML(runner) {
   }
 
   // pass toggle
-  on(passesLink, 'click', function() {
+  on(passesLink, 'click', function(evt) {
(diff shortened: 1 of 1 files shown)


danielstjules:
Thanks! Would you mind undoing your changes to mocha.js (we only run make before releases, keeps PR/commit diffs cleaner) and squashing?

jimenglish81:
sure no problem thanks

jimenglish81:
Are we ok to merge this? It would help with the filtering of tests for ember-cli projects. Thanks

dasilvacontin:
Thanks @jimenglish81!

Issue #2121, opened 2016-02-22
Looks like glob needs to be upgraded to at least 3.2.4


andresn:
Graceful was removed as of glob 3.2.4:
https://github.com/isaacs/node-glob/commit/49122ad716730a7d56a03edb6a5c15225eebe3f1

Installing latest mocha (mocha@2.4.5 ) yields:

```
npm WARN deprecated graceful-fs@2.0.3: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible.

Install log / dep tree:
├─┬ mocha@2.4.5 
│ ├── commander@2.3.0 
│ ├── diff@1.4.0 
│ ├── escape-string-regexp@1.0.2 
│ ├─┬ glob@3.2.3 
│ │ ├── graceful-fs@2.0.3 
```

danielstjules:
Thanks for the issue! This is something we're currently aware of:  #2117 #2113 #2111 #2058

From https://github.com/mochajs/mocha/issues/2058#issuecomment-186923581
> It's less about node compatibility, and more about npm. Releases [of glob] newer than our current use a caret range (^) in version matching, which isn't supported in older versions of npm.

I'll leave this open for tracking and to prevent additional issues being opened up.

boneskull:
Can anyone show where this breaks in the wild?

SamuelMarks:
```
/tmp/a$ npm install --save mocha
npm WARN deprecated graceful-fs@2.0.3: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible.
mocha@2.4.5 node_modules/mocha
├── escape-string-regexp@1.0.2
├── commander@2.3.0
├── diff@1.4.0
├── growl@1.8.1
├── supports-color@1.2.0
├── debug@2.2.0 (ms@0.7.1)
├── mkdirp@0.5.1 (minimist@0.0.8)
├── jade@0.26.3 (commander@0.6.1, mkdirp@0.3.0)
└── glob@3.2.3 (inherits@2.0.1, graceful-fs@2.0.3, minimatch@0.2.14)
```

Note that [the latest](https://github.com/isaacs/node-glob/blob/2fc2278ab857c7df117213a2fb431de090be6353/package.json) version of glob is 7.0.3.

danielstjules:
@aldeed Regarding https://github.com/mochajs/mocha/pull/2123#issuecomment-193811872

[...]

PR #2124, opened 2016-02-25
updating growl dependency

benjamine:
this updates to last version of growl ( from https://github.com/tj/node-growl/issues/57 )
this for example fixes the way notifications look in OSX (they don't show a success/fail icon currently), with this update now it looks like this:

screenshot 2016-02-25 03 41 43

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,7 +279,7 @@
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
     "glob": "3.2.3",
-    "growl": "^1.9.2",
+    "growl": "1.9.2",
     "jade": "0.26.3",
     "mkdirp": "0.5.1",
     "supports-color": "1.2.0"
--- a/package.json
+++ b/package.json
@@ -279,7 +279,7 @@
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
     "glob": "3.2.3",
-    "growl": "1.8.1",
+    "growl": "^1.9.2",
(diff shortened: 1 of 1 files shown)


danielstjules:
Hey! Sorry, using the tilde breaks compatibility with old versions of npm https://github.com/mochajs/mocha/issues/2121#issuecomment-194095797

We've also had a history of things breaking when bumping dependencies. If any of growl's dependencies use a tilde or don't use strict enough versioning, we could indirectly introduce future issues.

For example, with your current branch:

```
$ grep "mocha" package.json
    "mocha": "benjamine/mocha#master"
danielstjules:~/git/testing
$ node --version
v0.10.11
danielstjules:~/git/testing
$ npm --version
1.2.30
danielstjules:~/git/testing
$ npm install
...
npm ERR! Valid install targets:
npm ERR! ["1.0.1","1.0.2","1.1.0","1.2.0","1.2.1","1.3.0","1.4.0","1.4.1","1.5.0","1.5.1","1.6.0","1.6.1","1.7.0","1.8.0","1.8.1","1.9.0","1.9.1","1.9.2"]
npm ERR!     at installTargetsError 
```

benjamine:
@danielstjules that's perfect!, fixed

danielstjules:
Looks good, everything correctly installs with this latest version of the branch. :) And added bonus that we don't need to worry about nested dependencies, since growl doesn't have any.

Thanks!

PR #2145, opened 2016-03-07
Only replace absolute -> relative paths for stack lines

### LINKED ISSUE #2126: Stack trace filtering causing duplicate error messages
```js
describe('foo', function() {
  it('bar', function() {
    throw new Error('this is a test in ' + process.cwd() + '/foo/bar/test.js and it fails');
  });
});
```

will output

```
  1) foo bar:
     this is a test in /Users/tbn/Code/tmp/mocha-stack-bug/foo/bar/test.js and it fails
  Error: this is a test in foo/bar/test.js and it fails
      at Context.<anonymous> (test/test.js:5:11)
```

and as you can see the `this is a test in` message is duplicated in the output. In this example it is not a big deal but once you have larger error messages this becomes really problematic (see https://github.com/BadgeLabs/mocha-eslint/issues/29).

I've tracked this down to `stackTraceFilter()` being applied before the `BaseReporter.list()` method runs which removes the `message` from the `stack` property. But since the `stack` has been beautified the `messages` will no longer be equal due to the `process.cwd()` removal.

Turbo87:
This PR resolves #2126.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -733,7 +733,11 @@ exports.stackTraceFilter = function() {
       }
 
       // Clean up cwd(absolute)
-      list.push(line.replace(cwd, ''));
+      if (/\(?.+:\d+:\d+\)?$/.test(line)) {
+        line = line.replace(cwd, '');
+      }
+
+      list.push(line);
       return list;
     }, []);
 
--- a/test/utils.js
+++ b/test/utils.js
@@ -119,11 +119,11 @@ describe('utils', function() {
       });
 
       it('should replace absolute with relative paths', function() {
-        var stack = ['Error: failed'
+        var stack = ['Error: ' + process.cwd() + '/bla.js has a problem'
           , 'at foo (' + process.cwd() + '/foo/index.js:13:226)'
           , 'at bar (/usr/local/dev/own/tmp/node_modules/bluebird/js/main/promise.js:11:26)'];
 
-        var expected = ['Error: failed'
+        var expected = ['Error: ' + process.cwd() + '/bla.js has a problem'
           , 'at foo (foo/index.js:13:226)'
           , 'at bar (/usr/local/dev/own/tmp/node_modules/bluebird/js/main/promise.js:11:26)'];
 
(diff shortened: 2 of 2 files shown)


Turbo87:
@danielstjules any comments on this?

danielstjules:
It looks good, but can you post the output of the code snippet from https://github.com/mochajs/mocha/issues/2126#issue-136391738 with your branch? Will take a look later tonight.

Turbo87:
@danielstjules ping

danielstjules:
Sorry for the delay! Confirmed that the following works as described after the PR:


```javascript
describe('foo', function() {
  it('bar', function() {
    throw new Error('this is a test in ' + process.cwd() + '/foo/bar/test.js and it fails');
  });
});
```

```
----------------
Before PR
----------------
  1) foo bar:
     Error: this is a test in /Users/danielstjules/git/mocha/foo/bar/test.js and it fails

----------------
After PR
----------------
  1) foo bar:
     this is a test in /Users/danielstjules/git/mocha/foo/bar/test.js and it fails
  Error: this is a test in foo/bar/test.js and it fails
```

lib/utils.js
```diff
@@ -733,7 +733,11 @@ exports.stackTraceFilter = function() {
       }
 
       // Clean up cwd(absolute)
-      list.push(line.replace(cwd, ''));
+      if (/\(?.+:\d+:\d+\)?$/.test(line)) {
```
danielstjules: What's the capture group for? Could this be simplified, e.g.
``` javascript
if (line.match(/:\d+:\d+$/))
```

danielstjules:
Other than my comment on the regexp, looks good :)

lib/utils.js
Turbo87: that's actually not a capture group, those are escaped parenthesis. your suggestion would only work for stack lines without parenthesis around the file+position info, but according to the `test/utils.js` both seem to exist.

danielstjules: My bad, didn't notice the escape characters :)

PR #2158, opened 2016-03-16
fix more robust checking of timeout value in Runnable

OlegTsyba:


danielstjules:
Thanks for the PR! Unfortunately, but this would cause errors for anyone passing a null/undefined ms. Also, with PRs, please don't run `make` - we only do that before releases to keep diffs/commits a bit cleaner.

OlegTsyba:
I looked at the code, and have noticed that `Runnable.prototype.timeout` is called only by `suite`, which passes its own value of timeout, and, AFAIK, `Runnable.prototype.timeout` will never receive `ms` argument in the form of `string`, so, maybe it makes sense to remove unnecessary call to `milliseconds`, and clarify the documentation by specifying that function receive only `number`. Otherwise this function potentially does not work correctly in all possible cases, for example if the argument is passed as `'0'`

PR #2164, opened 2016-03-21
Fix non standard ANSI colors

jtangelder:
Replaces the non standard 90-97 color range to their standard 30-37 range. This should fix colors from not showing up, like on Travis.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -50,23 +50,23 @@ exports.inlineDiffs = false;
  */
 
 exports.colors = {
-  pass: 90,
+  pass: '30;1',
   fail: 31,
-  'bright pass': 92,
-  'bright fail': 91,
-  'bright yellow': 93,
+  'bright pass': '32;1',
+  'bright fail': '31;1',
+  'bright yellow': '33;1',
   pending: 36,
   suite: 0,
   'error title': 0,
   'error message': 31,
-  'error stack': 90,
+  'error stack': '30;1',
   checkmark: 32,
-  fast: 90,
+  fast: '30;1',
   medium: 33,
   slow: 31,
   green: 32,
-  light: 90,
-  'diff gutter': 90,
+  light: '30;1',
+  'diff gutter': '30;1',
   'diff added': 32,
   'diff removed': 31
 };


danielstjules:
We've had issues with changing colors in the past, so this will prob need a bit of time to review. That said, I haven't had any problems using those values with a toy project a couple years ago https://github.com/danielstjules/pho/blob/master/src/Console/ConsoleFormatter.php#L22-L35

danielstjules:
Since we couldn't upgrade to chalk, we can prob copy values from ansi-styles: https://github.com/chalk/ansi-styles/blob/master/index.js#L27-L60 Note that we can't use ansi-styles directly since it would break compatibility with older version of NPM due to its use of tilde in package.json

boneskull:
@jtangelder can you please rebase this onto `master`?  I want to see how the phantom test(s) are affected.

jtangelder:
Rebased!

boneskull:
@jtangelder Thanks!

Someone (maybe me) should pull down this PR and see what mochify's output looks like.  @mantoni are you interested in doing this?

mantoni:
Sure. Looks like my computer is offline for the day due to some technical issues with my provider. Hopefully I'm able to check tomorrow.

mantoni:
Looking good here. I can't see any difference.

bildschirmfoto 2016-06-18 um 12 33 22

boneskull:
basically floored that nyan works w/ mochify

boneskull:
So this solves #1200, apparently, but [I can't reproduce a problem](https://github.com/mochajs/mocha/issues/1200#issuecomment-229875943), so I'm having trouble understanding how this actually changes anything.    Help?

r-tanner-f:
Hmm, this doesn't seem to fix #1200 when using iTerm and the official Solarized theme. I'm still seeing 'pass' text as the background color.

boneskull:
@jtangelder Can you show an example of where Travis fails to show colors without this change?

jtangelder:
Can't reproduce anymore. But I think it still is a good idea to use the standard colors instead of the non-standard alternatives.

boneskull:
In general I think this is a good idea.  I'm not sure if this got merged in the shuffle.  If anyone wants to double-check or look at the changelog, then I'll close it if it was.

boneskull:
Closing for inactivity; unclear if issue reproducible

Issue #2148, opened 2016-03-08
this.skip() is also skipping teardown()/afterEach()


laughinghan:
``` js
    setup(function() {
      mock.innerHTML = 'Temporary stuff';
    });
    teardown(function() {
      mock.innerHTML = ''; // THIS ISN'T BEING RUN
    });
    test('test', function() {
      if ('some precondition is unmet') {
        this.skip();
      }
    });
```

With Mocha v2.4.5: http://jsbin.com/bijuru/edit?html,output
With latest `master` branch: http://jsbin.com/dajugel/edit?html,output

Observed: `Temporary stuff` is leftover from test setup

Expected: `Temporary stuff` should've been cleared by test teardown

danielstjules:
Confirmed as an issue. Likely extends to the bdd and other UIs as well.

``` javascript
suite('suite', function() {
  suiteSetup(function() {
    console.log('SUITE SETUP');
  });

  setup(function() {
    console.log('SETUP');
  });

  test('spec2', function() {
    console.log('SKIPPED SPEC');
    this.skip();
  });

  teardown(function() {
    console.log('TEADOWN');
  });

  suiteTeardown(function() {
    console.log('SUITE TEARDOWN');
  });
});

// =>

SUITE SETUP
SETUP
SKIPPED SPEC
SUITE TEARDOWN
```

danielstjules:
Note that the behavior is correct when using `test.skip` to skip the function. In that instance, the setup/beforeEach is never ran, so the teardown/afterEach is unnecessary.

It's only when dynamically skipping the spec that this occurs. In that case, the test is pending, however the setup/beforeEach has already ran.

boneskull:
@megagon @danielstjules 8a37e016e5629b46709891740c74d5f0b69929e0 was merged.  Is this still an issue?

laughinghan:
@boneskull: yes it is, I'm able to repro using the testcase from the original issue report that uses latest `master`, you can see "Temporary stuff" isn't cleaned up: http://jsbin.com/dajugel/edit?html,output

ScottFreeCode:
I copied this down locally and tried it with a rebuilt browser mocha.js file and confirmed the issue is still present. In the process I also noticed how this is different from #2286/#2287: that was `afterAll` being skipped if `beforeEach` calls skip, this is `afterEach` being skipped if the test or the `beforeEach` calls skip.

laughinghan:
[...]

PR #2168, opened 2016-03-26
Fix for tearDown on this.skip(), see #2148.

megagon:
Tests passed, new added for checking that tearDown executed.
Something to change?

### DIFF
--- a/test/acceptance/interfaces/tdd.js
+++ b/test/acceptance/interfaces/tdd.js
@@ -38,4 +38,24 @@ suite('Array', function(){
       done();
     });
   });
+
+  suite('tear down should be executed', function () {
+    var shouldBeChangeInTearDown;
+
+    setup(function () {
+      shouldBeChangeInTearDown = 5;
+    });
+
+    test('spec2', function () {
+      this.skip();
+    });
+
+    teardown(function () {
+      shouldBeChangeInTearDown = 10;
+    });
+
+    suiteTeardown(function () {
+      shouldBeChangeInTearDown.should.eql(10);
+    });
+  });
 });
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -101,7 +101,6 @@ program
   .option('--use_strict', 'enforce strict mode')
   .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
   .option('--delay', 'wait for async suite definition')
-  .option('--ignore-file <ignore-file-path>', 'specifies ignore file', './default.mochaignore')
 
 program.name = 'mocha';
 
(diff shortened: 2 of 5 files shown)


megagon:
@boneskull 
Do you have any comments?

boneskull:
see #2148 for question(s)

laughinghan:
@megagon: this doesn't seem to do anything for the testcase in #2148, see how "Temporary stuff" still isn't cleaned up even though I modified it to reference `megagon/master`: http://jsbin.com/kubago/edit?html,output

laughinghan:
@megagon: your PR doesn't modify `mocha.js` at all. Should it? Or is that only rebuilt on releases or something?

ScottFreeCode:
The browser mocha.js file has to be built for release, and may be updated before then, but if I understand correct generally shouldn't be in non-blocking PRs to avoid interfering with other PRs (well, or having to rebase depending on the order of merges, anyway).

test/acceptance/interfaces/tdd.js
```diff
@@ -38,4 +38,24 @@ suite('Array', function(){
...
+
+    teardown(function () {
+      shouldBeChangeInTearDown = 10;
+    });
+
+    suiteTeardown(function () {
+      shouldBeChangeInTearDown.should.eql(10);
+    });
+  });
 });
```
ScottFreeCode: I'd suggest just a boolean flag (and without duplicating "should" in its name), e.g.:
```js
var tornDown;

setup(function () {
  tornDown = false;
});

// ...

tearDown(function () {
  tornDown = true;
});

suiteTeardown(function () {
  tornDown.should.be.ok();
});
```

ScottFreeCode: More importantly, if this is an issue in Runner and fixed in Runner, then it shouldn't be in the TDD-specific tests, but rather in the non-interface-specific tests. (In fact, unless we've pretty severely screwed up the way the interfaces interact with the rest of the system, all we should need to test in the interface-specific tests is that all their versions of the methods work, I think...)

ScottFreeCode:
We should determine/decide whether calling `this.skip` inside a `beforeEach`/`setup` should also still run `afterEach`/`teardown` like it will if called in the test itself, or if we should assume that `beforeEach`/`setup` is going to do any necessary cleanup if it calls `this.skip` and continue skipping `afterEach`/`teardown` in that case.

Then, we should also add a test for whichever behaviour is determined/decided to be correct in that regard.

boneskull:
@megagon Can you please add test(s) which @ScottFreeCode would like to see to this PR?

boneskull:
Closing as inactive; please reopen if you'd like to continue.  Thanks!

PR #2170, opened 2016-03-28
Add pending count to HTML reporter

ericlau-solid:
Having pending count was useful to me, hence sharing it with mocha community.


dovrosenberg:
This would be super helpful.  Is there any chance of it being merged soon?

boneskull:
needs CLA

Issue #2036, opened 2016-01-03
add .mochaignore


TrejGun:
Please add .mochaignore file same as .gitignore
to ignore some files

yes i know there is `.skip` for tests but it would be good to ignore dirs or files with utils and configs

boneskull:
absolutely a good idea, but probably won't get implemented before #1969 unless there's a PR.

megagon:
I want to make PR for this, do you have any wishes about how it should looks like?
my suggestion is to make by default something like mocha.ignore with something like this:
`var skiplist = {folders:[], files: [], suites:[], tests:[]};` 
How is it sound?

ScottFreeCode:
A couple thoughts/questions:
1. Should test files ignored this way be loaded but marked entirely as pending?
2. Can/should something equivalent to this work in the in-browser testing mode?

boneskull:
since #1969 is still basically nowhere, a PR for this would be helpful.

gonnza:
Is the idea to allow a configuration as megagon suggested?

Something like 

`{folders:[], files: [], suites:[], tests:[]};`

?

- folders: Ignores a complete folder and its subdirectories from running tests
- files: Ignore specific files (by name ? by path?)
- suites: to ignore `describe` tests by name
- test: to ignore specific `it` tests by name

Should they show as pending, or as they do not exists?

boneskull:
As of v6, this is implemented as the `excludes` config value.  A `.mochajs.json` which includes the following should exhibit the requested behavior:

```json
{
  "excludes": [
    "some-glob/**",
    "some-other-glob/*.derp"
  ]
}
```

I think that we could alias the `excludes` option to `ignore`.

boneskull:
Would prefer we don't support `.mochaignore`, simply because it's functionality that's now duplicated in a config file.  

FWIW, ESLint [is moving in the direction of dropping its "ignore" file](https://github.com/eslint/rfcs/blob/ea369d753d571128ecb9afb3c459b8ce936fba24/designs/2019-config-simplification/README.md).

plroebuck:
Pre-Mocha-6, this might have been attractive, but now it's just an extra file to process with no benefit not already available.

Close as "wontfix" IMO.

gndelia:
Cool, I did not know the exclude option was already implemented. Then I agree, this should be closed

boneskull:
I'm cool with an `ignore` alias.  I can probably knock that out.

boneskull:
[...]

Issue #1577, opened 2015-03-04
Option exclude certain files by pattern when testing recursively


KylePDavis:
I would like the ability to provide an exclusion pattern so that I can only test files that match my test file pattern. This would allow data files to coexist with test files so long as they follow a reasonable pattern.

I think this would be simply providing an option that sets `ignore` option in `glob`.

Thoughts? I can make a PR for it real quick if you'd like.

alinex:
You may better make two parallel directories, one for tests and the other
for data files. That's what I did.
Am 04.03.2015 15:52 schrieb "Kyle P Davis" <notifications@github.com>:

KylePDavis:
I could (and have in the past) but I'd rather not have to do that anymore now that I've realized that glob ignore is right there.

It should be pretty easy to add a command line argument to set the glob ignore option. I just wanted to get thoughts and discuss a little bit first.

Any objections to me creating a PR for this?

toddbluhm:
I was just needing this too and was surprised to see mocha did not already have it. I think it would be a great addition. We have a "testApp" inside our test folder that we use specifically for testing. We don't want mocha trying to run tests inside that folder. It would be nice to be able to just add an exclude to our mocha.opts file to exclude that specific path rather than having to explicitly include all the other test sub-folders we have. 

While we could move the testApp out, it really doesn't fit anywhere else in the repo and its got relative pathing to certain files in the test folder.

This I think would be a nice addition. Certainly not critical, but helpful.

boneskull:
@KylePDavis @tawdle 

If I needed this, I think I'd probably just

```sh
$ mocha $(find test/ ! -path '*testApp*')
```

So I'm not really thrilled about the idea.  I'm kind of on the fence about the existence of the `recursive` flag even, since the same thing can be achieved with a shell command, `Makefile`, `Gruntfile.js`, `gulpfile.js`, etc., etc.

danielstjules:
I think @boneskull makes a good point - this is already achievable in a variety of ways. And that's not including the fact that you can structure you directories to avoid this altogether. For example, given this structure:

```
$ tree .
.
└── spec
    ├── fixtures
    ├── integration
    └── unit

4 directories, 0 files
```

You just update your package.json so that you can simply run `npm test`
```
  "scripts": {
    "test": "./node_modules/.bin/mocha spec/unit spec/integration"
  }
```

Or with a structure like the following:
```
$ tree .
.
└── src
    └── models
        ├── user.js
[...]

PR #2173, opened 2016-03-29
Feature ignore file #2036

megagon:
Please take a look at the implementation.
Let me know if that's ok - i will make tests for this. If need something to change - i will change.
After this i am going to add support to ignore files/suites/folders etc...

Tests passed locally.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -101,6 +101,7 @@ program
   .option('--use_strict', 'enforce strict mode')
   .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
   .option('--delay', 'wait for async suite definition')
+  .option('--ignore-file <ignore-file-path>', 'specifies ignore file', './default.mochaignore')
 
 program.name = 'mocha';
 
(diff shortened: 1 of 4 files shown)


boneskull:
@megagon Hi, thanks for this.  As you can see in #2036, nobody replied to your question.  We are spread pretty thin at the moment.  

I'm not very enthusiastic about adding features at the moment, and we're not going to release until we get some browser coverage.  IMO any changes before the rewrite should probably just be bug fixes.  But others can weigh in...
 
cc @mochajs/mocha

boneskull:
@megagon We do have browser coverage now and have since released, so we can revisit this.

I have a few comments:

1.  The default file should be `.mochaignore`
2.  It should support a global `mochaignore` file installed at whatever `npm get prefix` is
3.  It should support a per-user `~/.mochaignore`
4.  It should be based upon [ignore](https://www.npmjs.com/package/ignore) as to work how users would expect

Thanks for the PR!

ScottFreeCode:
FWIW, I've discovered that you can actually accomplish exclusion, albeit cryptically and not as easily, using Glob with Mocha as-is: https://github.com/mochajs/mocha/issues/1577#issuecomment-225409794

boneskull:
Closing for inactivity

PR #179, opened 2011-12-25
Xunit reporter

jkrall:
I banged out a simple xunit reporter, because I needed reports in Jenkins and the TAP reporter solution wasn't working for me.

Closes #10

PR #2178, opened 2016-03-31
avoid double and tripple xUnit XML escaping

graingert:


graingert:
Otherwise the title gets triple escaped and the error gets double escaped
```xml
<testcase classname="ex-name" name="should do a thing" time="0.12"><failure><![CDATA[(a =&amp;gt; b + &amp;quot;c&amp;quot;); failed
Error: (a =&gt; b + &quot;c&quot;);
    at Object.&lt;anonymous&gt; (./foo.js:80:19)]]></failure></testcase>
</testsuite>
```

graingert:
@danielstjules bump

graingert:
@dasilvacontin bump

lib/reporters/xunit.js
```diff
@@ -164,11 +164,3 @@ function tag(name, attrs, close, content) {
   }
   return tag;
 }
-
-/**
- * Return cdata escaped CDATA `str`.
- */
-
-function cdata(str) {
-  return '<![CDATA[' + escape(str) + ']]>';
```
dasilvacontin: Where did the `CDATA` tag go?

lib/reporters/xunit.js
```diff
@@ -130,7 +130,7 @@ XUnit.prototype.test = function(test) {
 
   if (test.state === 'failed') {
     var err = test.err;
-    this.write(tag('testcase', attrs, false, tag('failure', {}, false, cdata(escape(err.message) + '\n' + err.stack))));
```
graingert: it was used here, to triple escape the error message.

lib/reporters/xunit.js
graingert: https://github.com/mochajs/mocha/pull/2178/files#diff-d80a115516f0cc2ec0b46a54c60077faL133

lib/reporters/xunit.js
dasilvacontin: But then the output no longer has a surrounding `'<![CDATA[..,]]>'`, right? Don't we need that so that special characters can be used in the error message and trace?

dasilvacontin: Oh, okay, if they are escaped, it shouldn't be necessary. I wonder why it was used in the first place.

graingert: It was double escaping from the first cut: https://github.com/mochajs/mocha/pull/179/files#diff-d80a115516f0cc2ec0b46a54c60077faR88

dasilvacontin: Sorry, I'm don't know much about this topic. I don't see:
- Why (before your changes) the data was being escaped even though it was being put inside a `CDATA` tag. (which afaik is used to avoid having to escape chars)
- Why escaping the data is preferred to just putting the data inside a `CDATA` tag. (since using the `CDATA` tag uses less bytes and makes the data readable by a human on a quick glance)

dasilvacontin: > It was double escaping from the first cut: https://github.com/mochajs/mocha/pull/179/files#diff-d80a115516f0cc2ec0b46a54c60077faR88

I don't see how it was double escaping there, since `escape` was only used in the `cdata` fn, which was given `test.err.stack` directly.

Still, not as important now. I'm more interested in the second bullet point.

graingert: * No idea, ask @jkrall 
* CDATA is for humans to type content into XML, entity escaping is how XML libs do it.

graingert: @dasilvacontin you need to to do CDATA escaping properly the function should be:
```js
function cdata(content) {
    return `<!CDATA[${s.replace(']]>',']]]]><![CDATA[>')}]]>`;
}

while escape within CDATA is indeed "hurr, durr", it's safer than just wrapping in CDATA.

dasilvacontin: Wouldn't it be enough to stop using `escape` on `err.message`, since `escape` is being used in `cdata` already?

Removing the `CDATA` tag is not a change that would go into this major version, for the record.

dasilvacontin: I guess you are removing the `CDATA` tag because if you are escaping the inner content then there's apparently no point for the `CDATA` tag (?)

graingert: @dasilvacontin 
I am removing the CDATA tag because if you are escaping the inner content then there's no point for the CDATA tag

dasilvacontin: Gonna read a bit more on the `CDATA` tag...

graingert: @dasilvacontin pinging you in https://gitter.im/graingert

dasilvacontin:
Okay, without the `CDATA` tag, rendering is identical. node content is equivalent but not the same, though apparently no one cares about that. I'm waiting for @jkrall's reply, just in case: https://github.com/mochajs/mocha/pull/179/files#r59530748.

Otherwise LGTM. But hopefully we can get a reply soon.

dasilvacontin:
Will locally test in a few hours and merge if all is good.

dasilvacontin:
Sauce passed when ran from local. Merging. Thanks @graingert! :)

graingert:
Sweet do you know when this will get a release?

dasilvacontin:
@graingert Now available on npm. ;) (v2.5.2)

PR #2181, opened 2016-04-01
fix check-leaks to catch a leak whose name begins with 'd'

JustATrick:


lib/runner.js
```diff
@@ -838,7 +838,7 @@ Runner.prototype.abort = function() {
 function filterLeaks(ok, globals) {
   return filter(globals, function(key) {
     // Firefox and Chrome exposes iframes as index inside the window object
-    if (/^d+/.test(key)) {
+    if (/^\d+/.test(key)) {
```
JustATrick: Or would `/^\d+$/` be a better match for the comment?

boneskull:
@JustATrick Thanks.  At face value this *looks* like a typo, I'm going to dig a little deeper and make sure it's not this way on purpose.

JustATrick:
@boneskull Thanks - I wouldn't be surprised if I missed something.

And, assuming it is just a typo, thinking about it again, I realise that "fixing" it in either of the ways I suggested might not be the best. Since the condition wasn't doing anything to permit global index properties, and apparently nobody missed that behaviour (annotate suggests it's been there since 2013, v1.10.0), then maybe it's better to remove the comment and condition completely?

boneskull:
@JustATrick Unless this fixes a known, reproducible bug, I think we should probably just leave it as is.  Do you have anything like that?

JustATrick:
--check-leaks reproducibly misses the leak of any variable whose name starts with a 'd'. It catches a leaked 'cat', but misses a 'dog' - in my opinion that's a bug. The test I included in this pull request demonstrates it misses a leaked variable named 'derp'.  When I first discovered the problem it was failing to find a leaked 'document'.

ScottFreeCode:
And if the comment is accurate, I'm for making it `/^\d$/`.

boneskull:
lgtm, thanks

DiscoNova:
While I was looking at the commit alone, I was seriously baffled by "why has the regexp been changed to check if key begins with a number, and a test added to check that a key beginning with 'd' triggers failure". This discussion was truly enlightening regarding the issue.

A few thoughts...

- as @ScottFreeCode above mentions, being more explicit on the pattern would possibly be a good idea; currently the regexp will match keys like '123foo' etc. (which it, as far as I understand, shouldn't) ... same issue is also seen on `filterLeaks()`-method a bit later ... and that piece of code is actually dead at the moment, because this numeric index checking short circuits the tests in the beginning already
- furthermore, with a quick look, I was unable to find any tests that attempt to assert proper behavior when numeric indexes **are** used as keys, which I believe would make it explicit what is and what isn't supposed to get filtered ... however, I may have overlooked something somewhere else in the code

PR #930, opened 2013-07-19
Support for multiple reporters

misterdjules:
Hi! 

This pull request adds support for multiple reporters. 

One of my use cases for multiple reporters is described in my recent post on [mocha's google group](https://groups.google.com/forum/#!topic/mochajs/l7QrifO2bdE). To sum it up, I'm using mocha to run tests on a Bamboo instance, and I use the xunit-file reporter to create a file that can be parsed by Bamboo to display tests results. However, I also need to send tests reports to a CouchDB instance that doesn't accept xml files, but JSON documents. Right now, I'm running the tests twice: first with the xunit-file reporter, and then with the mocha-json-file-reporter to generate the xunit file and the json document. This has a number of cons like slowing down the whole process, and even sometimes not generating the exact same results, which can lead to a lot of confusion.

These changes allow me to run both reporters at the same time, and I think being able to run more than one reporter at a time can be useful to others. All the tests seem to run fine.

PR #2184, opened 2016-04-02
Added support for multiple reporters.

santiagoaguiar:
Rework of #2091, rebased to master.

Includes work from @benvinegar & @misterdjules, this PR also:
- Removes support for `--reporter json:<outputfile>` which was added on previous PRs, instead reporters that  support output to files should use `--reporter-options`.
- Added utility methods on `Base` reporter to open an `output` option provided on reporter options as an output file, and write to it. This is a generalization of the mechanism used by the `xunit` reporter to other reporters: `json`, `json-cov`, `html-cov`, `markdown`. All those reporters, and the `xunit` reporter, now inherit from `Base` and use these utility methods to write their output.
- Fixes lint warnings.
- Fixed issues with parsing the reporterOptions when there were multiple reporters.
- Adds a test case to check that the `json` reporter writes to the file specified by the <reporter>:<outputPath> option.
- Adds method documentation.

Submitted as PR independent of #2091 since it's a somewhat different approach. I think this one is better, simpler, and makes more sense, but I leave the other PR in case we want to go the way @benvinegar was proposing (using the `--reporter <name>:<outputfile>` syntax.)

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -66,10 +66,10 @@ program
   .option('-c, --colors', 'force enabling of colors')
   .option('-C, --no-colors', 'force disabling of colors')
   .option('-G, --growl', 'enable growl notification support')
-  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options')
-  .option('-R, --reporter <name>', 'specify the reporter to use', 'spec')
-  .option('-S, --sort', 'sort test files')
-  .option('-b, --bail', 'bail after first test failure')
+  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options, use name1:{k=v,k2=v2,...},name2:{...} for multiple reporters')
+  .option('-R, --reporter <name>', 'specify the reporter to use, or <name1>,<name2>,... for multiple reporters.', list, ['spec'])
+  .option('-S, --sort', "sort test files")
+  .option('-b, --bail', "bail after first test failure")
   .option('-d, --debug', "enable node's debugger, synonym for node --debug")
   .option('-g, --grep <pattern>', 'only run tests matching <pattern>')
   .option('-f, --fgrep <string>', 'only run tests containing <string>')
@@ -196,16 +196,45 @@ Error.stackTraceLimit = Infinity; // TODO: config
 
 var reporterOptions = {};
 if (program.reporterOptions !== undefined) {
-  program.reporterOptions.split(',').forEach(function (opt) {
-    var L = opt.split('=');
-    if (L.length > 2 || L.length === 0) {
-      throw new Error("invalid reporter option '" + opt + "'");
-    } else if (L.length === 2) {
-      reporterOptions[L[0]] = L[1];
-    } else {
-      reporterOptions[L[0]] = true;
+  var reporterOptionsParser;
+  var multipleReporterFormat = /([^,:]+):{([^}]+)}/g;
+
+  var parseOptions = function (optionsStr) {
+    var options = {};
+    optionsStr.split(',').forEach(function (opt) {
+      var split = opt.split('=');
+      if (split.length > 2 || split.length === 0) {
+        throw new Error("invalid reporter option '" + opt + "'");
+      }
+
+      if (split.length === 2) {
+        options[split[0]] = split[1];
+      } else {
+        options[split[0]] = true;
+      }
+    });
+
+    return options;
+  };
+
+  if (program.reporterOptions.match(multipleReporterFormat)) {
+    // multiple reporter config:
+    //   spec:{a=1,b=2},dot:{c=3,d=4}
+    var match;
+    while ((match = multipleReporterFormat.exec(program.reporterOptions))) {
+      if (match.length !== 3) {
+        throw new Error("invalid multiple reporter options format '" + program.reporterOptions + "'");
+      }
+
+      var reporterName = match[1];
+      var reporterOptionStr = match[2];
+      reporterOptions[reporterName] = parseOptions(reporterOptionStr);
     }
-  });
+  } else {
+    // single reporter config:
+    //   a=1,b=2
+    reporterOptions._default = parseOptions(program.reporterOptions);
+  }
 }
 
 // reporter
@@ -214,16 +243,17 @@ mocha.reporter(program.reporter, reporterOptions);
 
 // load reporter
 
-var Reporter = null;
-try {
-  Reporter = require('../lib/reporters/' + program.reporter);
-} catch (err) {
+program.reporter.forEach(function (reporterName) {
   try {
-    Reporter = require(program.reporter);
-  } catch (err2) {
-    throw new Error('reporter "' + program.reporter + '" does not exist');
+    require('../lib/reporters/' + reporterName);
+  } catch (err) {
+    try {
+      require(reporterName);
+    } catch (err2) {
+      throw new Error('reporter "' + reporterName + '" does not exist');
+    }
   }
-}
+});
 
 // --no-colors
 
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -12,7 +12,7 @@
 
 var escapeRe = require('escape-string-regexp');
 var path = require('path');
-var reporters = require('./reporters');
+var builtInReporters = require('./reporters');
 var utils = require('./utils');
 
 /**
@@ -36,7 +36,7 @@ if (!process.browser) {
 
 exports.utils = utils;
 exports.interfaces = require('./interfaces');
-exports.reporters = reporters;
+exports.reporters = builtInReporters;
 exports.Runnable = require('./runnable');
 exports.Context = require('./context');
 exports.Runner = require('./runner');
@@ -129,23 +129,42 @@ Mocha.prototype.addFile = function (file) {
 };
 
 /**
- * Set reporter to `reporter`, defaults to "spec".
+ * Set reporters to `reporters`, defaults to ['spec'].
  *
- * @param {String|Function} reporter name or constructor
- * @param {Object} reporterOptions optional options
+ * @param {String|Function|Array of strings|Array of functions} reporter name as
+ * string, reporter constructor, or array of constructors or reporter names as
+ * strings.
+ * @param {Object} reporterOptions optional options, keyed by reporter name, or
+ * '_default' for options to use when per-name options are not given.
  * @api public
- * @param {string|Function} reporter name or constructor
- * @param {Object} reporterOptions optional options
  */
-Mocha.prototype.reporter = function (reporter, reporterOptions) {
-  if (typeof reporter === 'function') {
-    this._reporter = reporter;
-  } else {
-    reporter = reporter || 'spec';
+Mocha.prototype.reporter = function(reporters, reporterOptions) {
+  // if no reporter is given as input, default to spec reporter
+  reporters = reporters || ['spec'];
+  reporterOptions = reporterOptions || {};
+
+  if (!utils.isArray(reporters)) {
+    if ((typeof reporters === 'string')
+        || (typeof reporters === 'function')) {
+      reporters = [reporters];
+    } else {
+      throw new Error('invalid reporter "' + reporters + '"');
+    }
+  }
+
+  // Load each reporter, only passing the options for that reporter
+  this._reporters = reporters.map(function(reporter) {
+    if (typeof reporter === 'function') {
+      return {
+        fn: reporter,
+        options: reporterOptions
+      };
+    }
+
     var _reporter;
     // Try to load a built-in reporter.
-    if (reporters[reporter]) {
-      _reporter = reporters[reporter];
+    if (builtInReporters[reporter]) {
+      _reporter = builtInReporters[reporter];
     }
     // Try to load reporters from process.cwd() and node_modules
     if (!_reporter) {
@@ -165,9 +184,12 @@ Mocha.prototype.reporter = function (reporter, reporterOptions) {
     if (!_reporter) {
       throw new Error('invalid reporter "' + reporter + '"');
     }
-    this._reporter = _reporter;
-  }
-  this.options.reporterOptions = reporterOptions;
+    return {
+      fn: _reporter,
+      options: reporterOptions[reporter] || reporterOptions._default || {}
+    };
+  }, this);
+
   return this;
 };
 
@@ -490,7 +512,14 @@ Mocha.prototype.run = function (fn) {
   var options = this.options;
   options.files = this.files;
   var runner = new exports.Runner(suite, options.delay);
-  var reporter = new this._reporter(runner, options);
+
+  // For each loaded reporter constructor, create
+  // an instance and initialize it with the runner
+  var reporters = this._reporters.map(function(reporterConfig) {
+    var _reporter = reporterConfig.fn;
+    options.reporterOptions = reporterConfig.options;
+    return new _reporter(runner, options);
+  });
   runner.ignoreLeaks = options.ignoreLeaks !== false;
   runner.fullStackTrace = options.fullStackTrace;
   runner.hasOnly = options.hasOnly;
@@ -502,21 +531,32 @@ Mocha.prototype.run = function (fn) {
   if (options.globals) {
     runner.globals(options.globals);
   }
+  // Use only the first reporter for growl, since we don't want to
+  // send several notifications for the same test suite
   if (options.growl) {
-    this._growl(runner, reporter);
+    this._growl(runner, reporters[0]);
   }
   if (options.useColors !== undefined) {
     exports.reporters.Base.useColors = options.useColors;
   }
   exports.reporters.Base.inlineDiffs = options.useInlineDiffs;
 
-  function done (failures) {
-    if (reporter.done) {
-      reporter.done(failures, fn);
-    } else {
-      fn && fn(failures);
+  function runnerDone(failures) {
+    function reporterDone(reporterFailures) {
+      if (--remain === 0) {
+        fn && fn(reporterFailures);
+      }
     }
+
+    var remain = reporters.length;
+    reporters.forEach(function(reporter) {
+      if (reporter.done) {
+        reporter.done(failures, reporterDone);
+      } else {
+        reporterDone(failures);
+      }
+    });
   }
 
-  return runner.run(done);
+  return runner.run(runnerDone);
 };
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -5,6 +5,10 @@
  */
 
 var tty = require('tty');
+var fs = require('fs');
+var mkdirp = require('mkdirp');
+var path = require('path');
+var EOL = require('os').EOL;
 var diff = require('diff');
 var ms = require('../ms');
 var utils = require('../utils');
@@ -235,12 +239,15 @@ exports.list = function (failures) {
  * of tests passed / failed etc.
  *
  * @param {Runner} runner
+ * @param {Object} options runner optional options
  * @api public
  */
 
-function Base (runner) {
+function Base (runner, options) {
   var stats = this.stats = { suites: 0, tests: 0, passes: 0, pending: 0, failures: 0 };
   var failures = this.failures = [];
+
+  this.reporterOptions = options ? options.reporterOptions : null;
 
   if (!runner) {
     return;
@@ -334,6 +341,71 @@ Base.prototype.epilogue = function () {
   }
 
   console.log();
+};
+
+/**
+ * Opens for writing the file referenced on the `optionName` option
+ * of the reporter options, if any.
+ * Call this method to support writing to this file when calling
+ * write or writeLine.
+ *
+ * @param {string} optionName the name of the option, defaults to 'output'.
+ * @api public
+ */
+Base.prototype.openOutput = function (optionName) {
+  var output = this.reporterOptions && this.reporterOptions[optionName || 'output'];
+  if (output) {
+    if (!fs.createWriteStream) {
+      throw new Error('file output not supported in browser');
+    }
+
+    if (this.fileStream) {
+      throw new Error('file stream already opened');
+    }
+
+    mkdirp.sync(path.dirname(output));
+    this.fileStream = fs.createWriteStream(output);
+  }
+};
+
+/**
+ * Write to reporter output stream.
+ *
+ * @param {string} str
+ * @api public
+ */
+Base.prototype.write = function (str) {
+  if (this.fileStream) {
+    this.fileStream.write(str);
+  } else {
+    process.stdout.write(str);
+  }
+};
+
+/**
+ * Write to reporter output stream, adding an EOL at the end.
+ *
+ * @param {string} line
+ * @api public
+ */
+Base.prototype.writeLine = function (line) {
+  this.write(line + EOL);
+};
+
+/**
+ * Close the output stream and callback with failures.
+ *
+ * @param failures
+ * @param {Function} fn
+ */
+Base.prototype.done = function (failures, fn) {
+  if (this.fileStream) {
+    this.fileStream.end(function () {
+      fn(failures);
+    });
+  } else {
+    fn(failures);
+  }
 };
 
 /**
--- a/test/reporters/multiple.js
+++ b/test/reporters/multiple.js
@@ -1,0 +1,113 @@
+'use strict';
+
+var Mocha = require('../../');
+var Suite = Mocha.Suite;
+var Runner = Mocha.Runner;
+var Test = Mocha.Test;
+
+describe('multiple reporters', function () {
+  it('should have 1 test failure', function (done) {
+    var mocha = new Mocha({
+      reporter: ['spec', 'json']
+    });
+    var suite = new Suite('Multiple reporters suite', 'root');
+    var runner = new Runner(suite);
+
+    /* eslint-disable new-cap */
+    var specReporter = new mocha._reporters[0].fn(runner);
+    var jsonReporter = new mocha._reporters[1].fn(runner);
+    /* eslint-enable new-cap */
+
+    var testTitle = 'json test 1';
+    var error = { message: 'an error' };
+
+    suite.addTest(new Test(testTitle, function (done) {
+      done(new Error(error.message));
+    }));
+
+    runner.run(function () {
+      // Verify that each reporter ran
+      expect(specReporter).to.have.property('failures');
+      expect(specReporter.failures).to.be.an('array');
+      expect(specReporter.failures).to.have.length(1);
+
+      expect(jsonReporter).to.have.property('failures');
+      expect(jsonReporter.failures).to.be.an('array');
+      expect(jsonReporter.failures).to.have.length(1);
+      done();
+    });
+  });
+
+  it('should pass correct reporter options and path to each reporter', function (done) {
+    var mocha = new Mocha({
+      reporter: [
+        'spec',
+        'dot',
+        'json'
+      ],
+      reporterOptions: {
+        spec: { foo: 'bar' },
+        json: { bar: 'baz' }
+      }
+    });
+
+    var doneCnt = 0;
+    function reporterDone () {
+      doneCnt++;
+      if (doneCnt === 3) {
+        done();
+      }
+    }
+
+    // specReporter
+    mocha._reporters[0].fn = function (runner, options) {
+      expect(options.reporterOptions).to.have.property('foo', 'bar');
+      reporterDone();
+    };
+
+    // dot (no options)
+    mocha._reporters[1].fn = function (runner, options) {
+      expect(options.reporterOptions).to.eql({});
+      reporterDone();
+    };
+
+    // json
+    mocha._reporters[2].fn = function (runner, options) {
+      expect(options.reporterOptions).to.have.property('bar', 'baz');
+      reporterDone();
+    };
+
+    mocha.run();
+  });
+
+  it('should pass _default reporter options to each reporter', function (done) {
+    var mocha = new Mocha({
+      reporter: ['spec', 'json'],
+      reporterOptions: {
+        _default: { foo: 'bar' }
+      }
+    });
+
+    var doneCnt = 0;
+    function reporterDone () {
+      doneCnt++;
+      if (doneCnt === 2) {
+        done();
+      }
+    }
+
+    // specReporter
+    mocha._reporters[0].fn = function (runner, options) {
+      expect(options.reporterOptions).to.have.property('foo', 'bar');
+      reporterDone();
+    };
+
+    // json
+    mocha._reporters[1].fn = function (runner, options) {
+      expect(options.reporterOptions).to.have.property('foo', 'bar');
+      reporterDone();
+    };
+
+    mocha.run();
+  });
+});
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -5,6 +5,7 @@
  */
 
 var Base = require('./base');
+var inherits = require('../utils').inherits;
 
 /**
  * Expose `JSON`.
@@ -17,15 +18,18 @@ exports = module.exports = JSONReporter;
  *
  * @api public
  * @param {Runner} runner
+ * @param {Object} options Runner optional options
  */
-function JSONReporter (runner) {
-  Base.call(this, runner);
+function JSONReporter (runner, options) {
+  Base.call(this, runner, options);
 
   var self = this;
   var tests = [];
   var pending = [];
   var failures = [];
   var passes = [];
+
+  this.openOutput();
 
   runner.on('test end', function (test) {
     tests.push(test);
@@ -54,9 +58,14 @@ function JSONReporter (runner) {
 
     runner.testResults = obj;
 
-    process.stdout.write(JSON.stringify(obj, null, 2));
+    self.write(JSON.stringify(obj, null, 2));
   });
 }
+
+/**
+ * Inherit from `Base.prototype`.
+ */
+inherits(JSONReporter, Base);
 
 /**
  * Return a plain-object representation of `test`
--- a/lib/reporters/markdown.js
+++ b/lib/reporters/markdown.js
@@ -6,6 +6,7 @@
 
 var Base = require('./base');
 var utils = require('../utils');
+var inherits = utils.inherits;
 
 /**
  * Constants
@@ -24,12 +25,16 @@ exports = module.exports = Markdown;
  *
  * @api public
  * @param {Runner} runner
+ * @param {Object} options runner optional options
  */
-function Markdown (runner) {
-  Base.call(this, runner);
+function Markdown (runner, options) {
+  Base.call(this, runner, options);
 
+  var self = this;
   var level = 0;
   var buf = '';
+
+  this.openOutput();
 
   function title (str) {
     return Array(level).join('#') + ' ' + str;
@@ -92,8 +97,13 @@ function Markdown (runner) {
   });
 
   runner.on('end', function () {
-    process.stdout.write('# TOC\n');
-    process.stdout.write(generateTOC(runner.suite));
-    process.stdout.write(buf);
+    self.write('# TOC\n');
+    self.write(generateTOC(runner.suite));
+    self.write(buf);
   });
 }
+
+/**
+ * Inherit from `Base.prototype`.
+ */
+inherits(Markdown, Base);
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -7,10 +7,7 @@
 var Base = require('./base');
 var utils = require('../utils');
 var inherits = utils.inherits;
-var fs = require('fs');
 var escape = utils.escape;
-var mkdirp = require('mkdirp');
-var path = require('path');
 
 /**
  * Save timer references to avoid Sinon interfering (see GH-237).
@@ -35,21 +32,16 @@ exports = module.exports = XUnit;
  *
  * @api public
  * @param {Runner} runner
+ * @param {Object} options Runner optional options
  */
 function XUnit (runner, options) {
-  Base.call(this, runner);
+  Base.call(this, runner, options);
 
   var stats = this.stats;
   var tests = [];
   var self = this;
 
-  if (options.reporterOptions && options.reporterOptions.output) {
-    if (!fs.createWriteStream) {
-      throw new Error('file output not supported in browser');
-    }
-    mkdirp.sync(path.dirname(options.reporterOptions.output));
-    self.fileStream = fs.createWriteStream(options.reporterOptions.output);
-  }
+  self.openOutput();
 
   runner.on('pending', function (test) {
     tests.push(test);
@@ -64,7 +56,7 @@ function XUnit (runner, options) {
   });
 
   runner.on('end', function () {
-    self.write(tag('testsuite', {
+    self.writeLine(tag('testsuite', {
       name: 'Mocha Tests',
       tests: stats.tests,
       failures: stats.failures,
@@ -78,7 +70,7 @@ function XUnit (runner, options) {
       self.test(t);
     });
 
-    self.write('</testsuite>');
+    self.writeLine('</testsuite>');
   });
 }
 
@@ -86,37 +78,6 @@ function XUnit (runner, options) {
  * Inherit from `Base.prototype`.
  */
 inherits(XUnit, Base);
-
-/**
- * Override done to close the stream (if it's a file).
- *
- * @param failures
- * @param {Function} fn
- */
-XUnit.prototype.done = function (failures, fn) {
-  if (this.fileStream) {
-    this.fileStream.end(function () {
-      fn(failures);
-    });
-  } else {
-    fn(failures);
-  }
-};
-
-/**
- * Write out the given line.
- *
- * @param {string} line
- */
-XUnit.prototype.write = function (line) {
-  if (this.fileStream) {
-    this.fileStream.write(line + '\n');
-  } else if (typeof process === 'object' && process.stdout) {
-    process.stdout.write(line + '\n');
-  } else {
-    console.log(line);
-  }
-};
 
 /**
  * Output tag for the given `test.`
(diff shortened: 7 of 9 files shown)


santiagoaguiar:
I rebased it to master.

Anything else that can be done to get this one merged to master? 

Thanks!

dasilvacontin:
Hi @santiagoaguiar! As soon as I'm done with the browser CI integration, I'll start reviewing the most demanded PRs, which includes this one. Should be very soon!

santiagoaguiar:
OK! Didn't want to rush anyone :) Just let me know if there's anything I can do to help or improve! Thanks.

mpahuja:
Hope this will also support custom reporters. Nice feature added instead of needing to depend on other modules like mocha-multi to get this job done.

santiagoaguiar:
You can specify as many reporters as you want, custom or not. 

However, if building a custom reporter that must write to a file, it has to accept `reporterOptions`, like the `xunit` reporter does. If it writes to console, you can use it too, but its output might get mixed with output from other reporters that also write to console (so, make sure you only use one of those :)).

Here's a custom reporter that works OK with this PR (we are using it) and supports `reporterOptions` to write to a file: [mocha-trx-reporter](https://github.com/Infragistics/mocha-trx-reporter)

domharrington:
Any news on when this might be merged? I'm hoping to use https://github.com/issacg/mocha-bamboo-reporter as well as the `spec` reporter in CI.

dasilvacontin:
@domharrington Gonna take a look at it this weekend if no one beats me to it.

bin/_mocha
```diff
@@ -190,16 +190,41 @@ Error.stackTraceLimit = Infinity; // TODO: config
...
+
+      var reporterName = match[1];
+      var reporterOptionStr = match[2];
+
+      reporterOptionStr.split(',').forEach(function(reporterOpt) {
+        var split = reporterOpt.indexOf('=');
+        if (split === -1) {
+          throw new Error("invalid reporter option '" + reporterOpt + "'");
         }
+        reporterOptions[reporterName] = reporterOptions[reporterName] || {};
```
dasilvacontin: Cache the `options` object for this reporter outside the loop, and then you don't have to check it it's initialised to `{}` on each loop. Having a variable with a short name like `options` will also improve the legibility of the code.

After the loop you can assign back `options` to `reporterOptions[reporterName]`.

bin/_mocha
```diff
@@ -190,16 +190,41 @@ Error.stackTraceLimit = Infinity; // TODO: config
...
+    while((match = multipleReporterFormat.exec(program.reporterOptions))) {
+      if (match.length !== 3) {
+        throw new Error("invalid multiple reporter options format '" + program.reporterOptions + "'");
+      }
+
+      var reporterName = match[1];
+      var reporterOptionStr = match[2];
+
+      reporterOptionStr.split(',').forEach(function(reporterOpt) {
+        var split = reporterOpt.indexOf('=');
```
dasilvacontin: I feel like the code is easier to understand if you use `.split('=')` instead of `indexOf('=')`, and then assign the elements to variables e.g. `property` and `value`, for legibility. The assignment is also easier because you'd be using `split[0]` vs `reporterOpt.substr(0, split)`, etc.

bin/_mocha
```diff
@@ -190,16 +190,41 @@ Error.stackTraceLimit = Infinity; // TODO: config
...
+    };
+  } else {
+    // single reporter config:
+    //   a=1,b=2
+    program.reporterOptions.split(',').forEach(function(opt) {
+      var L = opt.split('=');
+      if (L.length !== 2) {
+        throw new Error("invalid reporter option '" + opt + "'");
+      }
+      reporterOptions._default = reporterOptions._default || {};
```
dasilvacontin: Same comments for this loop.

lib/mocha.js
```diff
@@ -143,23 +143,40 @@ Mocha.prototype.addFile = function(file) {
...
-  if (typeof reporter === 'function') {
-    this._reporter = reporter;
-  } else {
-    reporter = reporter || 'spec';
+Mocha.prototype.reporter = function(reporters, reporterOptions) {
+  // if no reporter is given as input, default to spec reporter
+  reporters = reporters || ['spec'];
+  reporterOptions = reporterOptions || {};
+
+  if (!Array.isArray(reporters)) {
```
dasilvacontin: We don't have a polyfill, but we have this: https://github.com/mochajs/mocha/blob/c04c1d73b30979d04d94398309ee47643ede0af4/lib/utils.js#L187-L191

lib/mocha.js
```diff
@@ -143,23 +143,40 @@ Mocha.prototype.addFile = function(file) {
...
+    if ((typeof reporters === 'string')
+        || (typeof reporters === 'function')) {
+      reporters = [reporters];
+    }
```
dasilvacontin: Hmm, maybe throw an `Error` otherwise? Since it's not a type we care for.

lib/mocha.js
```diff
@@ -143,23 +143,40 @@ Mocha.prototype.addFile = function(file) {
...
+  }
+
+  // Load each reporter
+  this._reporters = reporters.map(function(reporter) {
+    if (typeof reporter === 'function') {
+      return {
+        fn: reporter,
+        options: reporterOptions
```
dasilvacontin: Maybe it should be pointed out that when given an array of constructors as the first argument to `Mocha.prototype.reporter` it doesn't make sense to give an options object keyed by reporter name since we can't tell the reporter name from the constructor.

dasilvacontin: Otherwise reader might wonder for a while.

lib/mocha.js
```diff
@@ -10,7 +10,7 @@
 
 var escapeRe = require('escape-string-regexp');
 var path = require('path');
-var reporters = require('./reporters');
+var builtInReporters = require('./reporters');
```
dasilvacontin: Appreciated change. :)

lib/mocha.js
```diff
@@ -471,7 +491,14 @@ Mocha.prototype.run = function(fn) {
   var options = this.options;
   options.files = this.files;
   var runner = new exports.Runner(suite, options.delay);
-  var reporter = new this._reporter(runner, options);
+
+  // For each loaded reporter constructor, create
+  // an instance and initialize it with the runner
+  var reporters = this._reporters.map(function(reporterConfig) {
+    var _reporter = reporterConfig.fn;
+    options.reporterOptions = reporterConfig.options;
```
dasilvacontin: Runners may not clone the passed in options, and instead hold a reference, and even modify its content. I think it would be better to clone the `options` object in each loop, and pass that in to the constructor.

dasilvacontin: I wonder if someone relies on a reference to `options` being held by the reporter. Thoughts @mochajs/mocha?

lib/mocha.js
```diff
@@ -482,21 +509,32 @@ Mocha.prototype.run = function(fn) {
...
 
-  function done(failures) {
-    if (reporter.done) {
-      reporter.done(failures, fn);
-    } else {
-      fn && fn(failures);
+  function runnerDone(failures) {
+    function reporterDone(failures) {
+      if (--remain === 0) {
+        fn && fn(failures);
```
dasilvacontin: I believe there's no need for a `failures` argument in `reporterDone` since `failures` will already have a correct value in the scope.

lib/mocha.js
```diff
@@ -482,21 +509,32 @@ Mocha.prototype.run = function(fn) {
...
+      }
     }
+
+    var remain = reporters.length;
+    reporters.forEach(function(reporter) {
+      if (reporter.done) {
+        reporter.done(failures, reporterDone);
+      } else {
+        reporterDone(failures);
```
dasilvacontin: And then no need to pass it in here.

lib/reporters/base.js
```diff
@@ -231,13 +235,16 @@ exports.list = function(failures) {
...
+ * @param {Object} options runner optional options
  * @api public
  */
 
-function Base(runner) {
+function Base(runner, options) {
   var stats = this.stats = { suites: 0, tests: 0, passes: 0, pending: 0, failures: 0 };
   var failures = this.failures = [];
 
+  this.reporterOptions = options ? options.reporterOptions : null;
```
dasilvacontin: Why not already call `openOutput` in the constructor? If you specify an `output` you probably want any logs coming from the reporter to go to that file, by default.

Also, not sure about the naming, `output`. Not very specific, and output can also mean 'resulting data'. It's always a file right? What about `outfile`?

lib/reporters/base.js
```diff
@@ -333,6 +340,67 @@ Base.prototype.epilogue = function() {
...
+ */
+Base.prototype.openOutput = function(optionName) {
+  var output = this.reporterOptions && this.reporterOptions[optionName || 'output'];
+  if (output) {
+    if (!fs.createWriteStream) {
+      throw new Error('file output not supported in browser');
+    }
+
+    mkdirp.sync(path.dirname(output));
+    this.fileStream = fs.createWriteStream(output);
```
dasilvacontin: If a `fileStream` already existed, it should close it.

lib/reporters/base.js
```diff
@@ -333,6 +340,67 @@ Base.prototype.epilogue = function() {
...
+ * @param failures
+ * @param {Function} fn
+ */
+Base.prototype.done = function(failures, fn) {
+  if (this.fileStream) {
+    this.fileStream.end(function() {
+      fn(failures);
+    });
+  } else {
+    fn(failures);
```
dasilvacontin: Per the change I suggested above, it's not necessary to pass `failures` to `fn`.

lib/reporters/html-cov.js
```diff
@@ -17,25 +18,31 @@ exports = module.exports = HTMLCov;
...
       cov: self.cov,
       coverageClass: coverageClass
     }));
   });
 }
 
 /**
+ * Inherit from `JSONCov.prototype`.
+ */
+inherits(HTMLCov, JSONCov);
```
dasilvacontin: What's the point of extending `JSONCov`?

dasilvacontin: Also, I don't think a comment is necessary, `inherits` is quite self explanatory I'd say.

test/integration/reporters.js
```diff
@@ -60,4 +60,45 @@ describe('reporters', function() {
...
+      var tmpJsonFile = tmpDir + path.sep + 'test-multiple-reporters-' + randomStr + '.json';
+      var tmpXmlFile = tmpDir + path.sep + 'test-multiple-reporters-' + randomStr + '.xml';
+
+      var args = ['--reporter=json,xunit', '--reporter-options', 'xunit:{output=' + tmpXmlFile + '},json:{output=' + tmpJsonFile + '}'];
+      run('passing.js', args, function(err, result) {
+        if (err) return done(err);
+
+        var json = fs.readFileSync(tmpJsonFile, 'utf8');
+        fs.unlinkSync(tmpJsonFile);
+        var xml = fs.readFileSync(tmpXmlFile, 'utf8');
```
dasilvacontin: Reading an empty file will still return an empty 'Buffer' object (which evals to `true`). You have to do `!!` (or `Boolean(x)`) on its `length` or `toString()`.

test/integration/reporters.js
```diff
@@ -60,4 +60,45 @@ describe('reporters', function() {
...
+      var tmpFile = tmpDir + path.sep + 'test-json-reporter-options-' + randomStr + '.json';
+
+      var args = ['--reporter=json', '--reporter-options', 'output=' + tmpFile];
+      run('passing.js', args, function(err, result) {
+        if (err) return done(err);
+
+        var json = fs.readFileSync(tmpFile, 'utf8');
+        fs.unlinkSync(tmpFile);
+        var results = JSON.parse(json);
+        assert(results.passes && results.passes.length === 2, 'JSON did not contain the expected results.');
```
dasilvacontin: `results.passes && results.passes.length === 2` -> `results.passes.length === 2`.

This way we'll know if it failed for a deeper reason (if it didn't even have a `passes` array).

test/reporters/json.js
```diff
@@ -12,7 +12,7 @@ describe('json reporter', function(){
     });
     suite = new Suite('JSON suite', 'root');
     runner = new Runner(suite);
-    var mochaReporter = new mocha._reporter(runner);
+    var mochaReporter = new mocha._reporters[0].fn(runner);
```
dasilvacontin: Hopefully plugins/libraries don't break because we no longer expose `_reporter`.

test/reporters/multiple.js
```diff
@@ -0,0 +1,98 @@
...
+    var mocha = new Mocha({
+      reporter: ['spec', 'json']
+    });
+    suite = new Suite('Multiple reporters suite', 'root');
+    runner = new Runner(suite);
+    specReporter = new mocha._reporters[0].fn(runner);
+    jsonReporter = new mocha._reporters[1].fn(runner);
+
+    var testTitle = 'json test 1';
+    var error = { message: 'oh shit' };
```
dasilvacontin: hmm, these variables are only being used once, it actually makes it harder to read to extract them from the code below.

test/reporters/multiple.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    // specReporter
+    mocha._reporters[0].fn = function(runner, options) {
+      options.reporterOptions.should.have.property('foo', 'bar');
+    };
+
+    // dot (no options)
+    mocha._reporters[1].fn = function(runner, options) {
+      options.reporterOptions.should.eql({});
```
dasilvacontin: Not sure if it's the case, but if it compares object references it should always fail.

test/reporters/multiple.js
```diff
@@ -0,0 +1,98 @@
...
+    };
+
+    // json
+    mocha._reporters[2].fn = function(runner, options) {
+      options.reporterOptions.should.have.property('bar', 'baz');
+      done();
```
dasilvacontin: Rather than making one of the three runners mark the `it` as done, it would be nicer to have a `reporterDone` callback that increases a count and when `== 3` call `done`.

test/reporters/multiple.js
```diff
@@ -0,0 +1,98 @@
...
+
+    // specReporter
+    mocha._reporters[0].fn = function(runner, options) {
+      options.reporterOptions.should.have.property('foo', 'bar');
+    };
+
+    // json
+    mocha._reporters[1].fn = function(runner, options) {
+      options.reporterOptions.should.have.property('foo', 'bar');
+      done();
```
dasilvacontin: Same thing here.

test/reporters/multiple.js
```diff
@@ -0,0 +1,98 @@
...
+var Suite = Mocha.Suite;
+var Runner = Mocha.Runner;
+var Test = Mocha.Test;
+var should = require('should');
+
+describe('multiple reporters', function() {
+  var suite;
+  var runner;
+  var specReporter;
+  var jsonReporter;
```
dasilvacontin: Declaring these vars in this scope has apparently no purpose – probably leftover code.

dasilvacontin:
Thanks lots for the great work! ❤️ 

Left quite a few comments/nitpicks that are mainly styling / legibility / naming issues. Let me know if anything is not clear / didn't explain myself properly for any of those. Cheers!

dasilvacontin:
Also, it would be nice to credit the previous committers in your commit description. :)

lib/mocha.js
santiagoaguiar: Any support for cloning the `options` object? Can't relay on `Object.assign`, no polyfill nor dependency that includes a similar utility. Should I go for a `JSON.stringify`/`parse` route?

lib/mocha.js
santiagoaguiar: I think it's safer, since `reporterDone` might be called as a callback by `reporter.done`. I'll use `reporerFailures` as a name to avoid shadowing the variable which might cause some confusion.

lib/mocha.js
santiagoaguiar: Same as before, `reporter.done` might call `reporterDone`. I'm not sure if we can assume it will pass the same `failures` object.

lib/reporters/base.js
santiagoaguiar: `output` was the name used previously by the `xunit` reporter, I preferred not changing it to keep backward compat.

Also, since `reporterOptions` are in principle reporter specific, I don't know if some other reporter might decide to interpret (or it's already interpreting!) `output` in a different way. That's why I went by making each interested reporter call the purely helper method `openOutput` explicitly.

lib/reporters/base.js
santiagoaguiar: I think `done` already existed as such and received a failures and callback, I only added a default implementation to `base` and use the mechanism used on the previous PR to call done on each reporter. 

Since it already received a failures and a callback, I'm not sure we should change that, but I defer to you on that.

lib/reporters/base.js
santiagoaguiar: Would it be better just to throw an error instead? 

You shouldn't be calling `openOutput` twice, if you do (and for example you happen to write something in the meantime) that data will be lost. Feels like a path to silent errors for most of the times. If we want to support (for some odd reason) calling it twice, maybe we should add an explicit `closeOutput`?

lib/reporters/html-cov.js
santiagoaguiar: I found the line:
```javascript
function HTMLCov(runner, options) {
  JSONCov.call(this, runner, options, false);
```

Confusing at first, since `HTMLCov` it didn't inherit `JSONCov`. I understand is completely valid code and that is being used to reuse the initialization code, but I found odd to call the constructor function of another object while not inheriting, and would find it double odd to call it while inheriting something else :). If you want, I'll change it though.

About the comment, just followed what looked as a common pattern in the rest of the codebase (most times `inherits` is used, a comment with that form appears), I can remove it if you prefer.

test/integration/reporters.js
santiagoaguiar: I almost fainted, but I'm passing the encoding, so it returns a string :). If it's too obscure or you prefer to use the non-encoding version, I'll switch it so it returns the buffer, but most of the other calls in the tests use the form with the encoding.

test/integration/reporters.js
santiagoaguiar: I'll change it to two asserts then, so both things are checked independently, otherwise we'll get a non descriptive error I think.

santiagoaguiar:
I made the changes and rebased, should I replace the current branch or do you want me to create a new PR?

test/reporters/multiple.js
santiagoaguiar: It performs a deep equality check, it passes OK in this case.

bin/_mocha
santiagoaguiar: I moved the parsing of options to an independent function called it from both paths, and added the case for bool options (split length === 1), since the original PR were based on a version where that wasn't supported.

lib/mocha.js
dasilvacontin: But even if `reporter.done` might call `reporterDone`, `failures` will still be in the scope due to `runnerDone`, right?

test/integration/reporters.js
dasilvacontin: Oh, cool, I didn't know it returned a string when using the encoding, my bad. Hmm, maybe an assertion like `json.length > 0` would be more natural.

test/integration/reporters.js
dasilvacontin: :+1:

test/reporters/multiple.js
dasilvacontin: Oh, I was expecting it to have `eql` and `deepEql` or something.

bin/_mocha
```diff
@@ -191,16 +191,45 @@ Error.stackTraceLimit = Infinity; // TODO: config
...
+    optionsStr.split(',').forEach(function(opt) {
+      var split = opt.split('=');
+      if (split.length > 2 || split.length === 0) {
+        throw new Error("invalid reporter option '" + opt + "'");
+      }
+
+      if (split.length === 2) {
+        options[split[0]] = split[1];
+      } else {
+        options[split[0]] = true;
```
dasilvacontin: Nice. This wasn't supported in the previous implementation, was it? Or did I miss it?

dasilvacontin:
Will follow up the replies later today.

lib/mocha.js
santiagoaguiar: It's on scope, but the `failures` passed by `reporter.done` might be a different object. Contrived example:

`reporter` is a reporter with the following `done` implementation:
```javascript
done(failures, cb) {
   cb(failures.concat({}))
}
```

If `reporterDone` used the `failures` captured before calling `reporter.done`, the above code would't work. 

I'm not sure it's a real case, but I think it's safer, and the as far as I know the callback signature is `(failures)`, so I think we should reflect & honor that there and use the actual object passed by arg.

test/reporters/multiple.js
santiagoaguiar: Agree! it's actually `equal` for strict equality and `eql` for deep equality. Anyway, I tried it and it worked, so I guess we are safe here :).

bin/_mocha
santiagoaguiar: Nope, it's on master for single reporter options, but the original PR (the one by the guys working on this originally) was written before it was supported and since it changed that area it removed support for single reporter options accidentally, and I didn't notice the change either. Now I did :).

boneskull:
@santiagoaguiar apologies for the delay, thanks for the work here.  @dasilvacontin did you follow up or no..?

dasilvacontin:
Sadly, no.

bguiz:
*bump* I'd really love to see this get into mocha, any status updates on this?

K76154:
When will this be merged into master?

test/integration/reporters.js
santiagoaguiar: I changed it `json.length > 0`.

bin/_mocha
```diff
@@ -63,8 +63,8 @@ program
   .option('-c, --colors', 'force enabling of colors')
   .option('-C, --no-colors', 'force disabling of colors')
   .option('-G, --growl', 'enable growl notification support')
-  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options')
-  .option('-R, --reporter <name>', 'specify the reporter to use', 'spec')
+  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options, use name1:{k=v,k2=v2,...},name2:{...} for multiple reporters')
```
boneskull: I'm not so keen on this API.  would prefer something slightly more familiar, like [subarg](https://www.npmjs.com/package/subarg)

boneskull:
@santiagoaguiar thanks for your work on this!

boneskull:
@K76154 we don't have timetables for things, generally.

bin/_mocha
santiagoaguiar: Sure, want me to add a dependency to subarg and use that instead? IIRC, This syntax was suggested on the previous PR, but I have no issues using a different one. 

However, I do like the fact  that is the same syntax for the reporter options when using a single reporter, so you can just copy and paste them without the need of reformatting. e.g.: 

`mocha --reporter foo --reporter-options x=y,w=z` 
is equivalent to: 
`mocha --reporter foo --reporter-options foo:{x=y,w=z}`

santiagoaguiar:
@boneskull anything left to be done for this one? I addressed @dasilvacontin reviews, is there anything else I should look into? 

As I see it, current implementation has these benefits:
1. Adds an additional parameter to the reporter constructor with the `options` for the reporter, following the path already taken by the `xunit` reporter, thus not exactly inventing anything new.
2. Provides some helper methods on the `Base` reporter to handle an `output` option (which is now the only option needed by most built in reporters), but doesn't require to actually inherit from it to use reporter options argument for something.
3. Uses an extension of the current syntax for reporter options, extended to support options to different reporters, thus not inventing anything new.

And these cons:
1. Mostly an extension of existing mechanisms, maybe we want to do something more comprehensive to cover some cases we are not thinking of or provide some sort of more general solution?
2. Syntax for passing options to multiple reporters is mocha specific, as it was the syntax for single reporters. Also parsing of options is somewhat fragile (for example, no way to escape/delimiter an option, so you can't pass as an option one of the chars used to delimiter options). This also happens with the current reporter options though (e.g. you can't use a filename including `,` like `--reporter-options "output=tests, and stuff.json"` since `,` is used to delimit options).

boneskull:
@santiagoaguiar I haven't had a chance to look at this.  I can see it landing after 3.0.0, which I hope to get published in the next couple days.

bin/_mocha
boneskull: yeah, I think that's what subarg does.  adding the dep should be fine, assuming it will work in windows.  when you do, please use an exact version instead of a range (`~` or `^`)

santiagoaguiar: OK! I guess we'll need to keep support for the older syntax for a single reporter too, or otherwise we'll need to update the major version?

cletusw:
Now that 3.0.0 has landed, it would be nice to see this get picked up again.

azachar:
Hello, is here any progress?

dasilvacontin:
@azachar All progress for an issue/PR is visible in the issue/PR itself.

santiagoaguiar:
@dasilvacontin should I rebase this for master or do you think we won't be able to merge it anyway? 

I haven't been tracking all the changes lately, maybe there's been structural changes that merit a rewrite more than just rebase?

test/reporters/multiple.js
fearphage: `deepEqual` is much more explicit to the reader... especially those unfamiliar with `should.js` (like me).

dasilvacontin:
@santiagoaguiar  Hmm, I don't think that there've been many changes affecting this PR. Maybe the addition of a linter. It will probably be an easy rebase (?).

Also, we now have a CLA that we require to sign, since Mocha is part of the jQuery/JS Foundation now.

Mainly this PR needs follow up by reviewers. eg myself.

test/reporters/multiple.js
santiagoaguiar: No issue with changing it if you everyone prefers it, though the rest of the code uses `should/expect.eql`, and `eql` is also what's used on `chai` for deep equality (which is what I'm used to). To be honest, I wasn't aware `should.js` provided `deepEqual` as a synonym for `eql`, seems it was added after the first PRs ;). Wondering why there are uses of `expect`/`assert`/`should` depending on the test, maybe one style should be chosen to write the tests?.

ScottFreeCode: I believe `expect` is preferable for browser compatibility (see Travis's Sauce configuration for the list of browsers), `should` is what we were using in most cases before we tried to convert some of the test base to run in the browser and we didn't get around to finishing it (and, fortunately, mostly translates to `expect` straightforwardly, so we should really finish that conversion), and I have no idea why `assert` would be used in other cases.

apowers313:
Wow, this trail goes all the way back to #930 from July, 2013. Is there any Guinness record for "longest open and continuously active PR"? :)

ibratoev:
Yeah, it is amazing that @santiagoaguiar is still willing to continue contributing to this PR. And this is the only recent **high-priority** PR.
@dasilvacontin, can I (or someone else) help to move this PR forward?

lib/mocha.js
dasilvacontin: @santiagoaguiar Since we support IE7, I'd use `stringify`/`parse` 'til we figure out what we do with polyfills / stuff. We lose support for using `function` as value for options (when using Mocha programmatically), but I don't think anyone is doing that.

lib/mocha.js
```diff
@@ -483,21 +512,32 @@ Mocha.prototype.run = function(fn) {
   if (options.globals) {
     runner.globals(options.globals);
   }
+  // Use only the first reporter for growl, since we don't want to
+  // send several notifications for the same test suite
   if (options.growl) {
-    this._growl(runner, reporter);
+    this._growl(runner, reporters[0]);
```
dasilvacontin: Which makes me think: "why does growl need to know anything about reporters".

dasilvacontin: Unrelated to the PR but, apparently, we pass a reporter to `this._growl` because we get `failures` and `duration` from the Reporter.

IMO that info belongs to the Runner. Do you agree @boneskull? Just from looking at the Runner you should be able to tell whether it's running or not, for how long it's been running, how many pass/fails so far, etc.

lib/mocha.js
dasilvacontin: Alright, fair enough, preserves behavior. Post-merge we'll/I'll have to look into if it's any useful being able to change the `failures` number via the callback.

lib/reporters/base.js
dasilvacontin: I see. We could change this we decide to bump major though. Ideally, `outfile`/`output` would work for any reporter (that doesn't move the cursor around), and the reporter implementation could ignore it as long as it is using `Base#write`.

We could leave this improvement for later though.

/cc @boneskull

lib/reporters/base.js
dasilvacontin: Good. We won't change it then, but we should look into its usefulness at some point.

lib/reporters/html-cov.js
dasilvacontin: Hmm, I just noticed neither HTMLCov or JSONCov are part of the codebase anymore.

dasilvacontin: I can see in the CHANGELOG that they were removed in v3. Issue about that: #2356.

test/integration/reporters.js
dasilvacontin: 🙌

test/reporters/multiple.js
```diff
@@ -0,0 +1,109 @@
...
+      reporterOptions: {
+        spec: { foo: 'bar' },
+        json: { bar: 'baz' }
+      }
+    });
+
+    var doneCnt = 0;
+    function reporterDone() {
+        doneCnt++;
+        if (doneCnt === 3) {
```
dasilvacontin: Small nitpick: `3` -> `mocha._reporters.length`, or something of the likes (?).

test/reporters/multiple.js
dasilvacontin: I guess nevermind it for now, and we'll open an issue for assert consistency when merging this PR.

dasilvacontin:
> @apowers313: Wow, this trail goes all the way back to #930 from July, 2013. Is there any Guinness record for "longest open and continuously active PR"? :)

Hopefully the Guinness record would give us more visibility so that more people help maintain the project! \o/

> @ibratoev: Yeah, it is amazing that @santiagoaguiar is still willing to continue contributing to this PR. And this is the only recent high-priority PR.
> @dasilvacontin, can I (or someone else) help to move this PR forward?

No need to ask for permission :). Anyone can chime in and help reviewing issues, PRs. Anyone can become a maintainer. As long as you are respectful and you collaborate a bit. :)

santiagoaguiar:
I rebased, fixed linting, moved `should` to `expect` on tests. Will follow up with the rest of the suggested changes (in particular moving to `subarg`, which I think is the largest one) during the weekend. 

Thanks @dasilvacontin for the review! 

By the way, my suggestion of rewriting was also in case you guys wanted to approach this in a completely different way (given the PR was stuck for so long). I don't mind if this PR gets in or not, as long as we get support for multiple reporters :).

dasilvacontin:
@santiagoaguiar Ah, I see. I can't personally come up with a simpler and cleaner way of implementing the feature.

At most, I would nitpick about whether a Reporter should be responsible for creating/opening its output stream, instead of being handed one. But at this point.. ¯\_(ツ)_/¯

bin/_mocha
```diff
@@ -66,10 +66,10 @@ program
   .option('-c, --colors', 'force enabling of colors')
   .option('-C, --no-colors', 'force disabling of colors')
   .option('-G, --growl', 'enable growl notification support')
-  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options')
-  .option('-R, --reporter <name>', 'specify the reporter to use', 'spec')
-  .option('-S, --sort', 'sort test files')
-  .option('-b, --bail', 'bail after first test failure')
+  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options, use name1:{k=v,k2=v2,...},name2:{...} for multiple reporters')
+  .option('-R, --reporter <name>', 'specify the reporter to use, or <name1>,<name2>,... for multiple reporters.', list, ['spec'])
+  .option('-S, --sort', "sort test files")
```
boneskull: please revert these changes to quotes

bin/_mocha
```diff
@@ -66,10 +66,10 @@ program
   .option('-c, --colors', 'force enabling of colors')
   .option('-C, --no-colors', 'force disabling of colors')
   .option('-G, --growl', 'enable growl notification support')
-  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options')
-  .option('-R, --reporter <name>', 'specify the reporter to use', 'spec')
-  .option('-S, --sort', 'sort test files')
-  .option('-b, --bail', 'bail after first test failure')
+  .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options, use name1:{k=v,k2=v2,...},name2:{...} for multiple reporters')
```
boneskull: I think I may have mentioned this before, but this format is not something people are accustomed to.  can we please use the [subarg](https://www.npmjs.com/package/subarg) format (for lack of anything better that I know of)?  you'll probably want to pull in that package.

bin/_mocha
```diff
@@ -66,10 +66,10 @@ program
...
+  .option('-R, --reporter <name>', 'specify the reporter to use, or <name1>,<name2>,... for multiple reporters.', list, ['spec'])
+  .option('-S, --sort', "sort test files")
+  .option('-b, --bail', "bail after first test failure")
```
boneskull: and here

bin/_mocha
boneskull: all of this custom parsing should be removed if using subarg

lib/mocha.js
boneskull: I have no idea about growl.  we should look at gh bigdata and see if anyone is using it.

apowers313:
Is there a branch this can get merged in to so that this PR doesn't linger on to 2019?

boneskull:
@apowers comments like this don't make anything go faster and mostly just piss people off

apowers313:
Sorry, not trying to be inflammatory -- just trying to find a path forward.

santiagoaguiar:
@boneskull  `subarg` is basically an enhanced `minimist`, and I'm not aware of a way to use it with `commander`. Assuming we don't want to switch from `commander`, the other option I can think of to use `subarg` is to pass the reporter options after `--`, and pass those args to `subarg` (by processing the `program.args` with `subarg`).

This means the syntax would look like:
`mocha --reporters xunit,json -- --xunit [ --output result.xml ] --json [ --output result.json ]` 

Or for a single reporter maybe:  
`mocha --reporters xunit -- [ --output result.xml ]`

Usually the convention is that after `--` anything you put would either be passed verbatim or not interpreted as options. In this case, I guess it could be read as pass this options verbatim to the reporter because `mocha` won't interpret them by itself.

I'll keep looking to see if there are other options, but ideas are welcomed.

boneskull:
@santiagoaguiar Ugh, I didn't realize subarg wrapped minimist.  What happens if you give commander subarg-type arguments?  Does it just barf, or ignore them, or..?

I *would* like to replace commander, but it's pretty low priority.  Not something that should happen in this PR.

santiagoaguiar:
It barfs in different ways depending on how you try to use it. 

Examples:
`mocha --reporter-options [ --foo ]`: barfs saying `--foo` is an unknown option, and uses `[` as the value for `reporter-options` option.
`mocha --reporter-xunit [ --foo ]`: `--reporter-xunit` unknown option.
`mocha --reporter-options [ xunit --foo ]`: accepted, but `xunit` is left on `program.args` and `--foo` is silently discarded. `reporter-options` receives `[` as value.  Doesn't seem trivial to pass the appropriate args to `subarg`, and `--reporter-options` needs to be the last arg.

You _could_ use `--reporter-options "[ xunit --foo]"`, and then try to split the string as args, but is too ugly.

So, should we replace `commander` first in a different PR and then try to get this in?

boneskull:
I know a lot of people have waited a long time for this. I don't want to hold this up because of the CLI syntax. 

However, if we add weird CLI syntax, we will want to change it to something better later on. Of course, that's a breaking change, and we'll know it's coming. Mocha is loaded with enough tech debt as it is.

Different syntax idea: `--reporters=foo,bar --foo="--baz=quxx"`. Option names based on reporter name; options for each should be parseable standalone when in single or double quotes. I've seen similar syntax with utils like `mdfind` (though the stuff in quotes is a query language, but same idea).  I truly hope commander can handle that. 

What do you think?

santiagoaguiar:
To allow it, I think it's necessary to pass `allowUnknownOption` to `commander`. This means we lose showing usage when a mistyped arg is passed. 

We would also need code to parse that string I think, or at least split reasonably well (e.g. handling quotes). 

If we plan to `allowUnknownOption` and add custom parsing, I wonder why not keep the current syntax for reporter options, but support additional flags, like `--reporter-xunit <k=v,k2=v2,...>` to avoid the `name1:{k=v,k2=v2,...},name2:{...}` syntax. 

By the way, adding a prefix to the option(`--reporter-`) might help deprecation if the syntax is changed, and also avoid name conflicts. For example, if at some point `subarg` is supported, it can be switched to `--options-xunit` as a way to pass options to reporters and whatever else, and print deprecation if `--reporter-*` is given.

fgarcia:
I came to this issue because I also wanted this feature. However after noticing the complexity caused by the conflict with `--reporter-options` and other issues, I would like to cast my vote against this feature

Let me explain ;-)

People like me who want to chain multiple reporters AND the options of each one, should just write a custom reporter which is basically a multiplexer of others. In code it would far more cleaner to manage the configuration/order of such combination.

I believe this solution would be far cleaner, reduce complexity within the current codebase and lessen the burden of the mantainers. Since I am grateful of their hard work, I would see them working in more important issues.

Well... actually I did not manage to chain reporters from the one I wrote, but I believe documenting how to do that in a clean way, should have priority before dealing with the command line interface

santiagoaguiar:
I agree, that would be a sensible alternative. 

Wouldn't be against having to write a little code to output to multiple reporters, but the reporter API should be such that it ensures that you can actually multiplex easily to any reporter.

fgarcia:
After working further on my custom reporter and exploring the current ones, I've just realized that multiplexing is not that easy.

I've even noticed that there is one issue that does not seem easy to solve from the command line:

What happens with the `runner.on('end', ...) `of each reporter?

I want to combine a JSON to file reporter with the built in Spec reporter, but I would like to modify the 'end' part of Spec and only use mine.

Basically I want to cherry pick parts of other runners, and once again, I believe enabling a clean way to do this within a custom reporter would provide a more valuable / generic solution to the problem.

boneskull:
I'm going to close this as it's problematic with Mocha's current ability to handle options.

A proper configuration file would be a prerequisite to using multiple reporters, given what we've seen here.

graingert:
@boneskull can we revisit this? We've now got mocha-multi from @glenjamin and mocha-multi-reporters from @stanleyhlng 


/cc @benvinegar

corvinrok:
so we just have to keep on using mocha-multi and hack our way into the future

boneskull:
if there's a solution that has a good solution for multiplexing and individual reporter options, I'm open to it.

ShadSterling:
@boneskull, my PR to add support for individual reporter options was merged into mocha-multi today: https://github.com/glenjamin/mocha-multi/pull/53

Doesn't look like it's made it to NPM yet, but if you're able to use a `git` dependency you could try it out.

hayesgm:
I'm a little confused why this issue was closed due to bad CLI syntax, only to be followed up by new npm modules `mocha-multi` and `mocha-multi-reporters` that seemed to have found a solution to that issue. I understand the need to keep the code-base clean, but this is pushing the extra code complexity on to the projects which use mocha. I suppose this is a long-winded way of saying: is there a chance this issue is going to be re-addressed in the future, or was this a permanent closure?

plroebuck:
Working on plumbing. Once that works and accepted, it'll be much easier to add the multi-reporter capability.

Bradeskojest:
Is there anything new regarding this?

plroebuck:
Not yet.
Working on PR #3874 which will provide file output options for XUnit, TAP, and JSON reporters.
Afterwards, need to standardize reporter options handling.

michaelseno:
Can we define several `--reporter` and `--reporterOption` inside the `mocha.opts` file? It seems I'm able to define but the last report is only prevailing. i'm using the `mochawesome` reporter and also another reporter which is a testrail integration `mocha-testrail-reporter`. Please let me know if this is possible. Thanks

ShadSterling:
@michaelseno, AFAIK using multiple reporters still requires declaring a single reporter that is separately configured to use multiple reporters as an additional step.  I was able to use multiple reporters via https://www.npmjs.com/package/mocha-multi

michaelseno:
@ShadSterling Could you provide an example for using multiple custom 3rd party reporter using mocha-multi? since I cannot find any reference for using multiple 3rd party reporters for mocha. the one that I can see are the the native reporters only for mocha. Thanks in advance

ShadSterling:
IIRC there's no difference in specifying 3rd party vs native.  My test case uses [this set of reporters](https://github.com/ShadSterling/Awaitable/blob/master/mocha-multi.json), and generates all of the reports when I run `npx mocha --reporter mocha-multi --reporter-options mocha-multi=mocha-multi.json build/test`.  (The package scripts runs mocha through nyc to generate coverage reports, and that command seems to use `--reporter` for both coverage reporters and test reporters... or something.)

brettz9:
@plroebuck , if this issue can't be reopened, where would be a good place to subscribe to get updates on this?

Sayan751:
Having a custom solution is quite easy, assuming the reporters you want to combine already exist. What I did is to create a small and naive custom reporter, and used the reporter in `.mocharc.js` config.

```javascript
// junit-spec-reporter.js
const mocha = require("mocha");
const JUnit = require("mocha-junit-reporter");
const Spec = mocha.reporters.Spec;
const Base = mocha.reporters.Base;

function JunitSpecReporter(runner, options) {
    Base.call(this, runner, options);
    this._junitReporter = new JUnit(runner, options);
    this._specReporter = new Spec(runner, options);
    return this;
}
JunitSpecReporter.prototype.__proto__ = Base.prototype;

module.exports = JunitSpecReporter;
```

```javascript
// .mocharc.js
module.exports = {
    reporter: './junit-spec-reporter.js',
    reporterOptions: {
        mochaFile: './tests-results/results.xml'
    }
};
```
The example above shows how to use both spec and junit reporter. For my purpose it is working pretty good. Not sure whether other reporters can be combined in the same manner.

ShadSterling:
@Sayan751, the mocha-multi README gives me the impression that combining multiple reporters isn't always so easy: https://github.com/glenjamin/mocha-multi#user-content-how-it-works

Sayan751:
@ShadSterling Well, it is not that difficult as well, is it? 😄

Jokes apart, it might be difficult if you want to combine the reporters generically. But generally when setting up the project you know which reporters are going to be used. For my case it is 'junit' (for ci) and 'spec' (for development). Combining these 2 together, gives me all info on ci log as well. And if you do know the reporter requirements, then combining those is few lines of code (with ES6/TypeScript, it is even fewer). So, I will take that solution over a hacky solution.

PR #2186, opened 2016-04-04
[Fix] Exclude not own properties when looping on options

JakeChampion:
It seems this was a regression as it was added before -- https://github.com/mochajs/mocha/pull/1586

### DIFF
--- a/support/browser-entry.js
+++ b/support/browser-entry.js
@@ -119,7 +119,10 @@ mocha.ui = function(ui){
 
 mocha.setup = function(opts){
   if ('string' == typeof opts) opts = { ui: opts };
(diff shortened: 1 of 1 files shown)


boneskull:
the chances of this breaking somebody's setup are probably greater than it actually addressing any problems.  if there's a bug here, we can revisit.  thanks though!

JakeChampion:
It broke our setup on https://cdn.polyfill.io :-(

boneskull:
@JakeChampion Where was the check added?  How did your setup break?

triblondon:
@boneskull do you really think people are relying on setting mocha config options by extending the Object prototype?

PR #2199, opened 2016-04-10
upgrade glob to the latest version

thefourtheye:
This is necessary because the version of `graceful-fs` package used by `glob` 3.2.3 is
deprecated. This change would avoid the warning

```
npm WARN deprecated graceful-fs@2.0.3: graceful-fs version 3 and before
will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon
as possible.
```

when installing the latest version of mocha.

### DIFF
--- a/package.json
+++ b/package.json
@@ -278,7 +278,7 @@
     "debug": "2.2.0",
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
-    "glob": "3.2.3",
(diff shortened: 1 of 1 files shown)


graingert:
https://github.com/mochajs/mocha/issues/2121#issuecomment-194095797

boneskull:
see #2121

PR #2212, opened 2016-04-15
Correct regex for function source cleaning

### LINKED ISSUE #2208: Imbalance in function source opening/closing shown in markdown/html reports
This is a bit of an edge case since normally test functions are not named (in the sense of the function itself having a name) and one does not usually compress the body of the function due to readability/structure, but for what it's worth, this code...
```js
var assert = require("assert")

it("has the head of the function but not the closing brace", function opening() {
	assert(true)
})
it("does the same thing with a function declared elsewhere", functionDeclaredOutside)
it("has the closing brace of the function but not the head", function () {assert(true)}) // a space before the parens doesn't affect it, must be the name
it("does not have the closing brace because of just one space after the contents", function() {assert(true) })
it("has the closing brace if the function body is entirely empty", function()
	{}
)
it("does not have the closing brace because of just one space in the body",function(){ })

function functionDeclaredOutside() {
	assert(true)
}
```
...yields this output when run with the markdown reporter...

---

has the head of the function but not the closing brace.

```js
function opening() {
	assert(true)
```

does the same thing with a function declared elsewhere.

```js
function functionDeclaredOutside() {
	assert(true)
```

has the closing brace of the function but not the head.

```js
assert(true)}
```

does not have the closing brace because of just one space after the contents.

```js
assert(true)
```

has the closing brace if the function body is entirely empty.

```js
}
```

does not have the closing brace because of just one space in the body.

```js

```

---

I would guess that the inclusion of `function name() {` in named functions is intentional, so that references to the name from within the function are clear; but in any case the closing brace being included or not should match whether the function name/opening is included, rather than whether there's whitespace before it.

ScottFreeCode:
Fixes #2208

- Treat tabs etc. as spaces.
- Match function name for removal.
- Match only non-`)` inside the function parameter list for removal instead of `.`; prevents matching through later in the first line e.g. `function(){ if ([condition]) { ...`.
- Match 0 or more spaces before } on the end for removal rather than 1 or more.
- Make the closing } optional for arrow functions with a single-statement body just like the opening {.
- Add tests for these cases.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -264,8 +264,8 @@ exports.slug = function(str) {
 exports.clean = function(str) {
   str = str
     .replace(/\r\n?|[\n\u2028\u2029]/g, '\n').replace(/^\uFEFF/, '')
-    .replace(/^function *\(.*\)\s*\{|\(.*\) *=> *\{?/, '')
-    .replace(/\s+\}$/, '');
+    // (traditional)->  space/name     parameters    body     (lambda)-> parameters       body   multi-statement/single          keep body content
+    .replace(/^function(?:\s*|\s+[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\s*\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\s*\}|((?:.|\n)*))$/, '$1$2$3');
 
   var spaces = str.match(/^\n?( *)/)[1].length;
   var tabs = str.match(/^\n?(\t*)/)[1].length;
--- a/test/utils.js
+++ b/test/utils.js
@@ -19,6 +19,30 @@ describe('utils', function() {
     it('should remove tab character indentation from the function body', function() {
       clean('\t//line1\n\t\t//line2').should.equal('//line1\n\t//line2');
     });
+
+    it('should handle functions with tabs in their declarations', function() {
+      clean('function\t(\t)\t{\n//code\n}').should.equal('//code');
+    });
+
+    it('should handle named functions without space after name', function() {
+      clean('function withName() {\n//code\n}').should.equal('//code');
+    });
+
+    it('should handle named functions with space after name', function() {
+      clean('function withName () {\n//code\n}').should.equal('//code');
+    });
+
+    it('should handle functions with no space between the end and the closing brace', function() {
+      clean('function() {/*code*/}').should.equal('/*code*/');
+    });
+
+    it('should handle functions with parentheses in the same line', function() {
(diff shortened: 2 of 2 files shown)


ScottFreeCode:
Updated with tests! https://travis-ci.org/mochajs/mocha/jobs/130488582#L928

test/utils.js
```diff
@@ -19,6 +19,26 @@ describe('utils', function() {
...
       clean('\t//line1\n\t\t//line2').should.equal('//line1\n\t//line2');
     });
+
+    it('should handle functions with tabs in their declarations', function() {
+      clean('function\t(\t)\t{\n//code\n}').should.equal('//code');
+    });
+
+    it('should handle named functions', function() {
+      clean('function withName() {\n//code\n}').should.equal('//code');
+    });
```
dasilvacontin: Can you add an `it` for when the function's name is followed by a space?

lib/utils.js
```diff
@@ -264,8 +264,7 @@ exports.slug = function(str) {
 exports.clean = function(str) {
   str = str
     .replace(/\r\n?|[\n\u2028\u2029]/g, '\n').replace(/^\uFEFF/, '')
-    .replace(/^function *\(.*\)\s*\{|\(.*\) *=> *\{?/, '')
-    .replace(/\s+\}$/, '');
+    .replace(/^function(?:\s*|\s+[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\s*\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\s*\}|((?:.|\n)*))$/, '$1$2$3');
```
dasilvacontin: Is there anyway this can be documented a bit, for legibility?
As in, have a comment in the previous line, and when you reach certain section you write what it's for.

```js
//                           fname   args        code      etc
.replace(/^function(?:\s*|\s+[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\s*\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\s*\}|((?:.|\n)*))$/, '$1$2$3');
```

test/utils.js
dasilvacontin: It's a quite common coding style.

lib/utils.js
ScottFreeCode: Good idea. I had half a mind to break the regex up into multiple variables, but wasn't sure if it would clutter the function as a whole... A comment should be a good compromise I think.

ScottFreeCode:
Updated based on review.

boneskull:
looks good to me.

dasilvacontin:
yay

PR #2214, opened 2016-04-17
Update README's header section

dasilvacontin:
- Use new logo
- Centers logo
- Gives it some margin

I couldn't use the `.svg`'s we have in `/assets/` for some reason, maybe because of the way it was exported. You can try it out for yourselves. /cc @boneskull 

If we were to use an image, we need a `@2x` export (double size) for retina screens. Hmm, I can actually generate one of those myself. But I hope we can solve the issues our `svg` files have.

I'll squash (or use "squash & merge") as necessary once the PR is reviewed, just in case we want to easily revert a subset of the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
+<br><br>
+<p align="center">
+  <img src="https://cldup.com/xFVFxOioAU.svg" alt="Mocha test framework"/>
+</p>
+<br><br>
+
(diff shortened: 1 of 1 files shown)


boneskull:
LGTM

boneskull:
@dasilvacontin What does "couldn't use" mean?

dasilvacontin:
@boneskull image would display as broken, and opening the svg's url would display the file's raw content (xml tags and such), instead of a render of the svg (like happens when you open one of our badge's svg url)

boneskull:
@dasilvacontin I'm not seeing an example of hosting an `.svg` on GitHub and referencing it.  Everyone else seems to use a remote image.  Probably a good idea.  Just stick em on `cloudup.com` or something

dasilvacontin:
@boneskull I tried imgur but they don't support `svg`. I'll try `cloudup`.

Btw, the logo with the `simple, flexible, fun` got the last `n` cut off – I didn't notice before merging, my bad.

PR #2218, opened 2016-04-20
Avoiding possible "TypeError: Cannot call method 'push' of undefined"

vangorra:
The [NodeJS api](https://nodejs.org/api/modules.html#modules_the_module_object) makes no mention about module.paths. Adding a line to avoid a possible TypeError exception.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -25,6 +25,7 @@ exports = module.exports = Mocha;
 
 if (!process.browser) {
   var cwd = process.cwd();
+  module.paths = module.paths || [];
   module.paths.push(cwd, path.join(cwd, 'node_modules'));
 }
 


dasilvacontin:
@vangorra In which scenario do you get the `TypeError`? Isn't this solution like silently failing?

vangorra:
My environment:
Running Ubuntu 15.10 with Meteor 1.3, which I think means node 10.40.x. 

Isn't this solution like silently failing?
No, this solution isn't equivalent to silently failing, rather initializing an uninitialized variable.

Reasoning:
- The PR applies to the root js file and fails before any mocha init can happen. So there is no other mocha code than can set this property.
- module.paths is not a documented property of the module object (in modern node docs) provided by node. So we cannot assume it is initialized nor is an array for that matter. So we must initialize it.
- Seems the paths property was deprecated some time ago anyway. Likely this is occurring because newer version of node no longer provide/support it. https://nodejs.org/docs/v0.4.3/api/all.html#loading_from_the_require.paths_Folders

At any rate, it's a better safe than sorry situation. For the reasons above, we cannot assume the varaible is initialized and It can't hurt to initialize it.

boneskull:
Since `module.paths` is undocumented--and therefore we cannot rely on it--perhaps we should use `NODE_PATH` instead.  Thoughts?

ScottFreeCode:
I think @dasilvacontin was on the right track -- if `modules.paths` is missing, then a Mocha-initialized copy will not have the effects the Node one should have, leading to other problems later when whatever we did with `module.paths` should have made a difference.

boneskull:
I'm in agreement with @ScottFreeCode and @dasilvacontin.  `module.paths`, if it doesn't exist, is a problem, and this PR would cause a silent failure in Mocha's functionality.  We should find a different way to do this (which *may* be `NODE_PATH`?).  Thanks for the PR though!

Issue #2207, opened 2016-04-14
BDD-style interface still active when other interface selected in CLI


ScottFreeCode:
When Mocha is run from the CLI and another interface is selected, BDD-style tests are still run. For instance, create the following two test files and then run `mocha -u exports` to see two tests run instead of only one.

test/bdd.js

``` js
it("should run only if the other does not", function(){})
```

test/exports.js

``` js
exports["should only run if the other does not"] = function(){}
```

dasilvacontin:
> instead of only one

Instead of a `ReferenceError: it is not defined`.

Thanks for pointing this out. It's not a priority though.

jmm:
I [brought this up on gitter](https://gitter.im/mochajs/mocha?at=570cfa8d4c2125fc3f01a23e) a couple days ago, but there was no response. This is really bad behavior. Especially since the [documentation](http://mochajs.org/#hooks) does a poor job of explaining the hooks. The main explanation of it portrays it that the hooks will always be the ones from the `bdd` interface, which combined with this bug makes for a bad situation.

ScottFreeCode:
Thanks to @codermarcos at #2994, I've also discovered that if `--watch` is used and this problem occurs the BDD tests on the subsequent watch-triggered runs will niether throw with a reference error *nor actually register their tests* -- you can get results like "1 passed, 0 pending, 0 failed" the first time followed by results like "0 passed, 0 pending, 0 failed" when the watch is triggerred.

juergba:
Can be closed, was fixed with release v6 and the implementation of `yargs`.

PR #2231, opened 2016-04-28
Make CI run browser tests

### LINKED ISSUE #2079: Add browser test suite to CI

dasilvacontin:
Resolves #2079 – still WIP.

"Difficulties" found so far:

- `zuul` comes with Mocha bundled (they actually have our bundle file in their source 👌 ) – this makes it hard to make some tests work, as they rely on the Mocha code being tested to be also the code that is running the tests. Might be worth taking a look at using Karma-PhantomJS framework-less (without having Mocha bundled).

- `zuul` didn't accept arrays for the `included` / `excluded` browserify options, had to fix and fire a PR: https://github.com/defunctzombie/zuul/pull/280.

- a lot of the tests are written in a way that you have to execute Mocha for each file (hence the super large Makefile with lots of different `test-xxx`). Together with the fact that quite a few tests can only work in `node` (using `fs`, `http.createServer` and such) makes that the subset of tests that can be run in the browser is not as big as one would hope. Rewriting some of the tests + adding new ones would be ideal, but not sure it's super worth considering that we are looking forward to the Mocha's rewrite. Still, those tests could likely be ported to the rewrite, not so bad I guess...

- I might need admin permissions over the org/repo to generate and encrypt sauce labs credentials – not sure yet, but it wouldn't hurt having them I guess. /cc @boneskull

Feedback is welcome – looking forward to getting this done and rocking the of the issues/PRs!


danielstjules:
Have you thought of just using phantomjs/nightmare for browser tests as discussed here https://github.com/mochajs/mocha/issues/2079#issuecomment-175909058 ?

I'd probably not recommend using zuul w/ sauce labs for all our browser tests as it's a flakey setup, though it works great locally. There's just something about running it on Travis. :) I previously used zuul for https://github.com/zendesk/cross-storage The builds end up experiencing frequent connection errors (between travis, localtunnel or saucelabs, etc) Here's an example: https://travis-ci.org/zendesk/cross-storage/builds/123155396 Looking at the build history you can see that they eventually pass, it just often takes a restart or two.

dasilvacontin:
> Edit: Though this could have been resolved in newer releases :)

What are you referring to?

I'm gonna give PhantomJS and other solutions a go – thanks for sharing!

danielstjules:
Oh, just meant that maybe stability improved in a more recent release, though I'm not sure.

dasilvacontin:
Going good so far using `mocha-phantomjs'. I'm able to test the source using the source itself as the test framework.

dasilvacontin:
Status update: I've been sick. I'm feeling better now – looking forward to make progress on the PR.

lib/browser/tty.js
```diff
@@ -1,5 +1,5 @@
 exports.isatty = function isatty() {
-  return true;
+  return false;
```
dasilvacontin: self-note: look into why were we reporting tty as true in the browser

dasilvacontin: It's been there since @tj first committed the file. Not sure why yet.

lib/mocha.js
```diff
@@ -99,22 +99,6 @@ function Mocha(options) {
...
-    exports.it = context.it || context.test;
-    exports.setup = context.setup || context.beforeEach;
-    exports.suiteSetup = context.suiteSetup || context.before;
-    exports.suiteTeardown = context.suiteTeardown || context.after;
-    exports.suite = context.suite || context.describe;
-    exports.teardown = context.teardown || context.afterEach;
-    exports.test = context.test || context.it;
-    exports.run = context.run;
-  });
 }
```
ScottFreeCode: Is this by any chance related to #2207?

dasilvacontin: What this code is doing is making these functions available via `require('mocha').describe` so that you can do `import { describe, it } from 'mocha'`. So yes, it's related, but it's not the part of the code that is setting up the globals.

ScottFreeCode:
For what it's worth, if the idea of using Karma in order to use a git checkout of Mocha instead of a published/bundled version is still being kicked around, Karma + Karma's Mocha runner seems to work with a local checkout of Mocha if you add Mocha's folder to the `NODE_PATH` environment variable (so it gets picked up the same way it would if it were in node_modules). And by "work" I mean it will at least try to run and presumably *should* work just as well as using Mocha with Karma for any other project, not that I actually have got to the point where tests are passing instead of having weird failures like Karma restarting the browser three times for no apparent reason. ;^) Karma's also supposed to a Sauce Labs runner that I would hope can run IE 8, since that's probably the browser that most needs testing against beyond picking any modern browser just to make sure some Node library doesn't break browsers altogether; I haven't even started to try that, though.

The other test runner I've heard of is Testling, which I'm pretty sure I've heard of being used with Mocha, but I haven't yet looked into whether that allows you to provide your own copy of the test library without having to write your own adapter with which to use it.

ScottFreeCode:
Addendum: You also have to build an up-to-date mocha.js file to use Karma's Mocha plugin with `NODE_PATH` including Mocha's folder. That poses a certain amount of potential conflict with the "don't update mocha.js till a release" policy... although not much -- CI isn't getting checked back into Git, so it shouldn't matter if it updates the build in order to run tests, whereas local testing... you'd just have to be sure to tell Git to discard the changes to mocha.js I guess. Well, that's that for what it's worth; I'll report back if I discover anything else worthy of note in the process of actually getting to where tests pass.

ScottFreeCode:
So, I can't seem to get bundling the test files (and by extension their dependencies) like the phantomjs bundle to work with Karma, *but* I did figure out that it works fine if I give Karma the list of test files instead, as long as:
1. I also give it the files for assertion libraries like "should" and any similar dependencies.
2. The test doesn't have any requires that can't be satisfied by the browser files for such libraries.

boneskull:
it appears we'll want to run IE and phantomjs (pre 2.0), so I think the most direct route to that might be to just pick up karma and eschew mocha-phantomjs.

boneskull:
where are the builds?

boneskull:
I have a feeling I screwed something up w/ travis but I have no idea what that could be.

scripts/upgrade-npm.js
```diff
@@ -0,0 +1,27 @@
...
+  exec('npm install --production', function (err) {
+    if (err) {
+      throw new Error(err);
+    }
+    exec('npm install --global npm@2', function (err) {
+      if (err) {
+        throw new Error(err);
+      }
+      // I'm unsure what this one is for, but whatever.
+      exec('npm install --global npm', function (err) {
```
dasilvacontin: npm v1.x can't upgrade directly to v3.x, so we first upgrade to v2, and then to v3.

Why was the script rewritten? With what purpose? The only behavioural change I see is the `if` condition, and that could have been easily changed in the bash script afaik.

dasilvacontin:
@boneskull what builds? afaik the package.json and/or Makefile hasn't been changed so that phantom.js is run on CI. I don't see any code related to spinning up karma either.

If you mean Travis builds, I can see them now, but maybe you've experienced some kind of delay. (?)

dasilvacontin:
I just saw from the other PR that you meant Travis. 🙈

scripts/upgrade-npm.js
boneskull: > npm v1.x can't upgrade directly to v3.x, so we first upgrade to v2, and then to v3.

we don't need to actually upgrade to v3?  so, we can skip that.

the shell script won't run on windows.  this *is* premature optimization, because I know that we'll need to run IE, and I don't want to have anything in here that isn't portable.  it also opens us up to running unit tests on diff platforms eventually.

boneskull:
yes, there was a significant delay.  anyway, the phantomjs version is failing to install on 0.8 for some reason.  but the older ones (1.9.7.x) fail to install on newer versions of node, so ugh.

scripts/upgrade-npm.js
dasilvacontin: > we don't need to actually upgrade to v3? so, we can skip that.

But maybe it wouldn't be a bad idea to test if dependencies install correctly while using the current version of npm.

> the shell script won't run on windows. this is premature optimization, because I know that we'll need to run IE, and I don't want to have anything in here that isn't portable. it also opens us up to running unit tests on diff platforms eventually.

Good call.

dasilvacontin: > But maybe it wouldn't be a bad idea to test if dependencies install correctly while using the current version of npm.

Nvm, we are already doing that when testing on newer Node versions.

ScottFreeCode:
Something that crossed my mind: Will the mocha.js need to have its timestamp tweaked so that make will be able to tell whether to update it for the tests, e.g. like https://gist.github.com/davidwindell/fbfef588c6295666c6a1 or maybe https://gist.github.com/jeffery/1115504 ?

ScottFreeCode:
Check it out! The tests are [finding the actual IE8 incompatibilities now](https://travis-ci.org/mochajs/mocha/jobs/131554812#L2061) (and otherwise passing*)!

*Well, I've still got to figure out what to do about PhantomJS; [latest doesn't work on Node 0.8](https://travis-ci.org/mochajs/mocha/jobs/131554808#L407), but we [got a seemingly unrelated failure downloading 1.9.8 recently](https://travis-ci.org/mochajs/mocha/jobs/131535239#L774).

ScottFreeCode:
Some additional comments:
- I've been testing with IE11 with Karma's HTML tweaked to trigger IE8 compatibility mode (not sure if that's how I ought to be doing it, but it works for most things...). I do not get the failure on the test for length that cannot be coerced to number. Is it likely to be an issue that exists in real IE8 but not IE8 compatibility mode?
- Looks like a lot of these are the `toISOString` issue (cf #2259). If I run the interface browser tests a lot of them are `indexOf`. (Note to self: look at which object/prototype is missing `indexOf` and whether it's being used inside Mocha itself or just in the test code.)
- After we fix the ones we can fix quickly, do we want to disable the difficult IE8-failing tests (say, using reverse grep?) in the IE8 test run only, in order to go ahead and start getting other things tested and merged and be able to come back and fix those tests for IE8 in due time?

ScottFreeCode:
- Do we want to remove any of the earlier changes that were aimed at mocha-phantomjs (e.g. to test/browser/...)?

ScottFreeCode:
@dasilvacontin, @boneskull, I think I've got everything working! Take a look and let me know if there are any issues, or if there's anything else we need to do before merging this. (Thanks!)

ScottFreeCode:
Attempted to squash this a bit... David's contributions are a bit scarce between the fact that some of them ended up not being used and the fact that his improvements to the makefile were absorbed into Chris's; and I'm ambivalent about whether that one commit that changed `return '    "dave"';` to get "dave" from the array being mapped should be absorbed into the previous squash (what I've got here), or put in its own commit for obviousness, or just dropped... But it'll do if it looks good to everyone else. I've saved the original line of commits in the branch [add-browser-to-CI-detailed](../commits/add-browser-to-CI-detailed) if anyone wants to compare (although I don't know that GitHub's comparison features are going to give meaningful info on them, since they amount to the same net changes but by entirely unrelated commit history).

boneskull:
restarting some failed jobs due to failed d/l of phantomjs prebuilt binary.  once these pass, I think we're good.

boneskull:
(thereafter I'm going to squash this in at least a couple places)

lib/browser/tty.js
boneskull: ooh, hmm, missed this.  it shouldn't change unless we have a good reason.

package.json
```diff
@@ -266,7 +267,8 @@
     "node": ">= 0.8.x"
   },
   "scripts": {
-    "test": "make test-all"
+    "preinstall": "node scripts/fake-karma-mocha-peer-dep.js",
```
boneskull: `karma-mocha` has a peer dep on `mocha`, which is a problem.  we want it to use, well, _this_ mocha, so we just symlink `node_modules/mocha` to the root.  

this is causing a problem atm because we're not publishing the `scripts` dir.  I'll figure out a solution; it doesn't need to run anyway unless you're running `npm install` from a working copy

.travis.yml
```diff
@@ -8,22 +8,37 @@ dist: trusty
...
-  - '5'
-  - '4'
-  - 'iojs'
-  - '0.12'
-  - '0.11'
-  - '0.10'
-  - '0.8'
+matrix:
+  include:
+    - node_js: '6'
```
boneskull: we could also add an `osx` target here, if anyone cares.

browser-entry.js
```diff
@@ -159,3 +159,5 @@ Mocha.process = process;
 
 global.Mocha = Mocha;
 global.mocha = mocha;
+
+module.exports = global;
```
boneskull: this needs explanation.

.travis.yml
ScottFreeCode: Worth trying, but probably best done after we get the basic browser testing merged.

browser-entry.js
boneskull: done

scripts/upgrade-npm.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+/**
+ * This upgrades npm on Travis when using Node.js 0.8x
+ */
+
+if (process.env.TRAVIS_NODE_VERSION === '0.8') {
+  var exec = require('child_process').exec;
+  // ensure *dependencies* are installed using provided npm
```
boneskull: (because if they don't, then mocha won't install on 0.8)

dasilvacontin: gonna amend this to `can be installed`

boneskull:
@mochajs/mocha unless someone has an objection, I'll likely merge this today or tomorrow

ScottFreeCode:
Sorry about the to-iso-string mishap; I added it where the other new dependencies were without thinking about the fact that it's used in the framework code itself rather than just the tests.

dasilvacontin:
@ScottFreeCode No worries. Sorry for not noticing that during code review ;)

PR #2201, opened 2016-04-11
Bump glob patch version to remove warning

astorije:
When installing `mocha@2.4.5`, `glob@3.2.3` triggers the following warning:

```
npm WARN deprecated graceful-fs@2.0.3: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible.
```

This commit bumps to the latest patch version before a major version.
Diff at https://github.com/isaacs/node-glob/compare/v3.2.3...v3.2.11.

Issue #2200, opened 2016-04-11
Change jade dependency to pug after their renaming


astorije:
It was already mentioned in https://github.com/mochajs/mocha/issues/1811, and the original discussion comes from https://github.com/pugjs/pug/issues/2184.

At the moment, installing the latest `mocha` (2.4.5), we get the following warning:

```sh
npm WARN deprecated jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
```

I wanted to take a shot at it but realized you've been using a 2012-old version of `jade` for reasons, and it looks like they haven't released old versions under the new name, so unless they do, it's not a trivial change.
Maybe a start could be requesting them to publish specific versions considering they have git tags for them...

boneskull:
We cannot upgrade to `pug` atm as this would cause old versions of `npm` to fail.  For now we support Node.js v0.8.x out of the box.  Support will be dropped @ next major.

silkentrance:
@boneskull Please be more specific. How can this break old versions of npm?

danielstjules:
If a dependency or nested dependency uses tildes/carets in their package.json, it breaks old npm support. Check here for an example: https://github.com/mochajs/mocha/issues/2121#issuecomment-194095797

Using npm 1.2.30, I get several failures since they can't resolve versions with `^`

```
npm ERR! Error: No compatible version found: pug-error@'^1.3.0'
npm ERR! Valid install targets:
npm ERR! ["0.0.0","1.3.0"]
```

donaldpipowitch:
> Support will be dropped @ next major.

Is there an estimated date for this?

dasilvacontin:
@donaldpipowitch There's lots to fix before dropping support for v0.8.x. Is this just because of the warning, or?

donaldpipowitch:
No, just a general question :) I was a little surprised that 0.8.x is still so important?

dasilvacontin:
I think @boneskull has a better insight regarding 0.8 usage.

mitselek:
it just sucks getting this warning message every time I like to use mocha.
I understand there are IT-restrictions and good reasons and what not - it just looks plain ugly and renaming happened a while ago.
Do something about it. Figure it out and fix it, please :) You can't postpone it till "next major" because next major is not about resolving jade issue, package renaming is absolutely irrelevant in next major - weak excuse.

And cheers - I love your work!

dasilvacontin:
[...]

PR #2234, opened 2016-04-30
Update glob for node 6 compatibility

pornel:
Older version of glob uses a version of graceful-fs, which is deprecated and incompatible with node v6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -278,7 +278,7 @@
     "debug": "2.2.0",
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
-    "glob": "3.2.11",
+    "glob": "7.0.3",
     "growl": "1.9.2",
     "jade": "0.26.3",
     "mkdirp": "0.5.1",


calebmer:
This also removes a `graceful-fs` deprecation warning for those with OCD that are don't like that type of thing.

dasilvacontin:
Thanks! It'll have to wait – cannot bump in mocha v2 for the same reason we cannot bump jade: https://github.com/mochajs/mocha/issues/2200#issuecomment-216017864.

pornel:
I admire that you strictly respect semver and still support node 0.8.x users, even though it's not even maintained by the node team. 

However, mocha is also very important on the cutting edge part of the node ecosystem.

To help adoption of Node 6, and also reduce potential for breakage for those upgrading to Node 6, could you release just that small change as mocha 3.0.0 (bumping version just to satisfy semver for node 0.8 users), and rename the currently planned v3.0.0 to v4.0.0?

dasilvacontin:
@pornel I was actually thinking of doing something like that, or maybe `v3.0.0-alpha` or the sorts – thanks for your voice! Making Mocha run on Node v6 is important imo, and you can tell other users think so as well.

What do you peps think, @mochajs/mocha?

ScottFreeCode:
Try `npm install mochajs/mocha` to use the latest master from GitHub and see if that still has the issue? It looks like #2201 should have eliminated the graceful-fs dependency by updating glob to 3.2.11, it just hasn't made it into a release yet. Of course, one would want to use an actual release of Mocha (at least in production); if I'm not mistaken the changes from #2201 will be able to go into a 2.4.6 or 2.5.0 release, without breaking compatibility, once #2079/#2231 is done.

That said, as a user I'm all for having an early copy of the next "major" version though (which should be able to resolve the Jade warning among other things; updating to the latest version of all dependencies if possible might be a good idea at that point), if it isn't too much trouble to maintain both that and the current 2.x.y release line until the Mocha team is ready to switch over fully.

boneskull:
@ScottFreeCode Seems to be correct that Node.js v6.x is *not* broken when installing from `master`, but *is* from the currently published `v2.4.5` (though I have nothing to confirm this except the warning when installing).  

Installing from `master` against Node.js v0.8 yields no errors or warnings.

This leads me to believe we could release `v2.4.6` with an upgraded `graceful-fs`.  *I am extremely hesitant to do this* until we get automated browser tests in place.  I *thought* using the "CLI-only" `chalk` module would not affect browser users; I was incorrect, which is the reason for my hesitation.  Nothing can be so easily assumed when bundling for the browser.

*If I can confirm* `graceful-fs` causes a Mocha installation to *break*--not just yield an install warning--I'll investigate doing a patch release later today; at minimum, we'll want to do some manual browser tests, and update our CI matrix.

cc @mochajs/mocha

boneskull:
Going to close this PR if that is the case; if that *is* the case, the problem has already been solved.

boneskull:
Can anyone confirm a case where Mocha breaks under Node.js v6?  Our test suite [passes with Node.js 6](https://travis-ci.org/boneskull/mocha/jobs/127946069).

I'm having trouble understanding the trail of issues, discussion and code surrounding the deprecation of old `graceful-fs` versions.  Does it *actually break*, or will it *actually break* in Node.js v7?

See isaacs/node-graceful-fs#58 prompting the deprecation warning, and the linked issue in nodejs/node#5213.  

Hopefully one of @isaacs @chalker @jasnell or @thealphanerd can help distill the conversation around those issues...?  Thank you, if you can.

ChALkeR:
@boneskull The deprecation warning upon installation in graceful-fs isn't very precise — it still works with Node.js 6 due to a temporary fix that was landed on Node.js side just for that, see https://github.com/nodejs/node/pull/5102.

That temporary work-around is going to be reverted in Node.js 7.0 (refs: https://github.com/nodejs/node/issues/5213, https://github.com/nodejs/node/pull/6413).

Note that the fix also prints a warning _in runtime_, from Node.js side, and that warning is printed only in those situations that will throw in 7.0.

ChALkeR:
I opened https://github.com/isaacs/node-graceful-fs/issues/64 in regards to the deprecation message.
The initial graceful-fs deprecation message was fine, but it was mistakenly changed in https://github.com/isaacs/node-graceful-fs/issues/61.

boneskull:
@ChALkeR Thanks for the explanation.  I can confirm the warning is now fixed.

boneskull:
To summarize for interested parties:

1.  `glob@3.2.3`, which current `mocha@2.4.5` uses, *does not actually break on Node.js v6*, as you will see from the updated deprecation message of its dependency, `graceful-fs`.
2.  `mocha@2.4.5` *should* break under Node.js v7 in certain (?) circumstances.
3.  We updated to `glob@3.2.11`, which *should not* break under Node.js v7, but have not released this change.
4.  Since nothing is actually breaking right now (there is no Node.js v7 yet), I'm not in a hurry to release anything.
5.  So, I'm going to close this issue, and we'll release a new patch version with some recent fixes (including the upgraded `glob@3.2.11`) once necessary changes for #2231 land.

boneskull:
(warnings can be hidden during `npm install` by using `--loglevel=error` or configured permanently in `.npmrc`, if it's bothering you too much)

darkmorpher:
Related:
#2201 & #2262

PR #2248, opened 2016-05-11
Add --icu-data-dir flag for node

adrian-ludwig:
It is missing from available flags list and I would be nice to have it.


boneskull:
@adrian-ludwig thanks!

PR #2252, opened 2016-05-13
prepare to use karma on saucelabs; fix potential windows problems

boneskull:
- rewrite `scripts/ensure-compatible-npm.sh` in JS
  - detect travis node 0.8 env; just upgrade if found
- symlink `mocha` to itself for `karma-mocha`
  - if explicit `--production` flag is present, skip this
  - don't know how to skip this part if `mocha` is a devDep of some other package

**just triggering a build; please ignore**


boneskull:
actually this won't trigger a build.

PR #2258, opened 2016-05-15
(ab)use issue template to ward off duplicates

ScottFreeCode:
I'm not sure if this is actually a good idea (maybe putting it in CONTRIBUTING.md would be better? do we know if people actually read CONTRIBUTING.md?), and I'm pretty sure it's not the intended use case for issue templates, but I wanted to throw this out there in case it might help people avoid submitting duplicate issues while we're waiting for a new release. The glob/graceful-fs deprecation warning issue is an obvious example, but there have been a few other issues raised that were fixed in master months ago, such as the one about exceptions thrown by the reporter trying to clean a string for errors thrown out of hooks. Feel free to ignore, discard, edit to something completely different, or whatever else seems better.

TODO: if this is accepted create an issue/pull to delete it once we get back to a regular release schedule again.


dasilvacontin:
Browser tests are now finally added to CI.

@ScottFreeCode it's a good idea, though not sure if the flux of issues was big enough to make it worth it (which is why I didn't merge it back then). But it's a very good way to warn users in case we need it in the future – thanks for sharing. :)

ScottFreeCode:
Aha, you beat me to closing it. 8^)

PR #2259, opened 2016-05-16
Fix missing methods/globals in IE7, IE8 environments

ndhoule:
This patch fixes two incompatibilities in pre-ES5 environments:

- Missing `JSON` global (IE7)
- Missing `Date#toISOString` method (IE7, IE8)

### DIFF
--- a/package.json
+++ b/package.json
@@ -269,6 +269,7 @@
     "test": "make test-all"
   },
   "dependencies": {
+    "@segment/to-iso-string": "0.0.2",
     "commander": "2.3.0",
     "debug": "2.2.0",
     "diff": "1.4.0",
@@ -276,6 +277,7 @@
     "glob": "3.2.11",
     "growl": "1.9.2",
     "jade": "0.26.3",
+    "json3": "^3.3.2",
     "mkdirp": "0.5.1",
     "supports-color": "1.2.0"
   },
--- a/lib/context.js
+++ b/lib/context.js
@@ -1,4 +1,10 @@
 /**
+ * Module dependencies.
+ */
+
+var JSON = require('json3');
+
+/**
  * Expose `Context`.
  */
 
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -3,6 +3,7 @@
  */
 
 var EventEmitter = require('events').EventEmitter;
+var JSON = require('json3');
 var Pending = require('./pending');
 var debug = require('debug')('mocha:runnable');
 var milliseconds = require('./ms');
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -4,6 +4,7 @@
  * Module dependencies.
  */
 
+var JSON = require('json3');
 var basename = require('path').basename;
 var debug = require('debug')('mocha:watch');
 var exists = require('fs').existsSync || require('path').existsSync;
@@ -11,6 +12,7 @@ var glob = require('glob');
 var join = require('path').join;
 var readdirSync = require('fs').readdirSync;
 var statSync = require('fs').statSync;
+var toISOString = require('@segment/to-iso-string');
 var watchFile = require('fs').watchFile;
 
 /**
@@ -481,7 +483,7 @@ function jsonStringify(object, spaces, depth) {
       case 'date':
         var sDate = isNaN(val.getTime())        // Invalid date
           ? val.toString()
-          : val.toISOString();
+          : toISOString(val);
         val = '[Date: ' + sDate + ']';
         break;
       case 'buffer':


ndhoule:
CI failure seems unrelated to PR, anyone care to re-run it?

ScottFreeCode:
Actually, the failure might be due to the PR; Node 0.8, and only Node 0.8, is complaining about the added "@segment/to-iso-string" dependency, [reacting as though you'd specified a dependency named just "segment"](https://travis-ci.org/mochajs/mocha/jobs/130645887#L200). It's hard to find documentation that far back to confirm it, but I'd hypothesize that scoped dependencies weren't supported in the version of NPM that comes with Node 0.8...

ndhoule:
Huh—I would think that [this script](https://github.com/mochajs/mocha/blob/master/scripts/ensure-compatible-npm.sh) would update npm.

Not really sure why we try to install non-devDeps using the version of npm bundled with node 0.8; updating npm on that platform is a non-issue. At least that's what we decided back when we Browserified ([ref](https://github.com/mochajs/mocha/pull/1727#issuecomment-109660225)). @dasilvacontin probably has an opinion here.

Also happy to swap the dep out with a non-namespaced one if someone can suggest one. I wasn't able to find any other `toISOString` modules that weren't buggy/incomplete (though I didn't search too hard as I already knew this one existed)

ScottFreeCode:
It looks like #2236 was deliberately designed to check this.

For what it's worth, recently I've seen this discussion about supporting 0.8 out of the box:
https://github.com/mochajs/mocha/issues/2200#issuecomment-216017864
https://github.com/mochajs/mocha/issues/2121#issuecomment-210308408

dasilvacontin:
> Huh—I would think that this script would update npm before installing deps. That was the intention behind adding it, anyway.
> 
> Not really sure why we try to install non-devDeps using the version of npm bundled with node 0.8; updating npm on that platform should be a non-issue. At least that's what we decided back when we Browserified (ref). @dasilvacontin probably has an opinion here.

We are currently supporting it out of the box –  not sure if there's really a reason why users can't upgrade `npm`, or if we do so out of convenience (/cc @boneskull). Still, to stop supporting node v0.8 out of the box would be a breaking change.

> Also happy to swap the dep out with a non-namespaced one if someone can suggest one. I wasn't able to find any other toISOString modules that weren't buggy/incomplete (though I didn't search too hard as I already knew this one existed)

That would be ideal. We can switch to the one you proposed when we bump the major version.

Also, @boneskull, pretty sure these changes will be needed for the browser testing ticket. Otherwise you'll likely bump into the exception.

package.json
```diff
@@ -269,13 +269,15 @@
...
   "dependencies": {
+    "@segment/to-iso-string": "0.0.2",
     "commander": "2.3.0",
     "debug": "2.2.0",
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
     "glob": "3.2.11",
     "growl": "1.9.2",
     "jade": "0.26.3",
+    "json3": "^3.3.2",
```
dasilvacontin: Carret makes `npm` cry.

dasilvacontin:
Thanks for the PR, btw!

ScottFreeCode:
We ended up getting `toISOString` shimmed in #2231; @boneskull, do we want to look at IE7 support and see whether it would take anything more than falling back on a library like this for `JSON` in the event it's missing too?

boneskull:
Has Mocha *ever* supported IE7?

boneskull:
I'm going to rebase this and see what happens.

boneskull:
Oh, right, it's failing b/c of the caret.  It looks like there was some implicit support for IE7 somewhere.

boneskull:
Rebased branch is [here](https://github.com/mochajs/mocha/tree/fix/improve-es3-compat).
Build is [here](https://travis-ci.org/mochajs/mocha/builds/136589926).

ScottFreeCode:
Is it possible to test that we don't add `JSON` to the global namespace? The main reason for that being that we don't want people running tests of code that uses `JSON` and having those tests pass because of the shim/polyfill pulled in by Mocha when in fact their code will fail outside of testing due to lack of the shim/polyfill. (I just manually checked and it seems to be ok, but it'd be nice to confirm it never ends up otherwise, e.g. by future changes that need to use JSON in new places internally.)

I suppose the same thing could hypothetically go for other shims/polyfills too, although it seems like most of the others I've seen are just acting as function variables used instead of the native stuff and not standing in place of native global variables.

Also, if this does work, we should try adding IE7 to the Sauce tests. For that matter, maybe we should see if it will go back all the way to 5 or 6? (Heh heh, do those still exist? Weird fact: the compatibility mode setting in IE11's developer console has modes for 10, 9, 8, 7 and 5. That's right, 5, which nobody even talks about anymore, but not 6.)

ScottFreeCode:
One other question: Does "json3" just return the existing JSON global if it's available? If not, we may want to use `var JSON = typeof JSON !== "undefined" ? JSON : require("json3");` instead of just `var JSON = require("json3");`

dasilvacontin:
@ScottFreeCode Apparently, it does: https://github.com/bestiejs/json3/blob/master/lib/json3.js#L43-L47

But it also checks if the native implementation is spec compliant and stuff.

boneskull:
I've rebased my branch onto `master` and added an IE7 test.  [Build](https://travis-ci.org/mochajs/mocha/builds/136967915)

boneskull:
Alright, this is passing.  I'm going to merge the branch. @ndhoule Thanks for getting this started!

PR #2267, opened 2016-05-21
Add mocha.js to .gitignore

dasilvacontin:
Makes contributing more dev-friendly, specially for newcomers to the
project.


dasilvacontin:
Meh, .gitignore only affects untracked files. My bad.

boneskull:
no, this is a good idea, it just needs more pieces:

- add `mocha.js` to `files` prop of `package.json`
- `git rm mocha.js`
- add `mocha.js` to `.gitignore`
- add `prepublish` script `make mocha.js`

this will ensure `mocha.js` is built and packaged when releasing.

boneskull:
actually looks like we're using an `.npmignore`.  I'll convert it to `files`

boneskull:
ehh we are using both, we only need one--I'll remove `.npmignore`

dasilvacontin:
I thought we had the browser bundle in git for people to download it from the repo or something.

boneskull:
> I thought we had the browser bundle in git for people to download it from the repo or something.

I don't know why.  You can download it with zip files or whatever.  But if you need the latest version and happen to be on a system without `make`, or stuck with an old `npm`, you're kind of screwed.  

See #2268.  I almost want to roll that into this; it shouldn't take too long to just convert the targets to npm scripts... 😛

boneskull:
I think we can get away with just making the `mocha.js` target an npm script for now

boneskull:
(squashed)

boneskull:
amended

boneskull:
actually nevermind.  we lose bower support if we remove the artifact from vcs.

ScottFreeCode:
Bower looks at the specific release, right? So we could check it in for the release commit, then delete it in the first commit to master after the release?

boneskull:
@ScottFreeCode bower doesn't look at anything except VCS tags, which means you have to include your build artifacts in VCS.  some projects make a separate repo just for this.  npmcdn.com is another option, but I couldn't get it working trivially (try `bower install https://npmcdn.com/mocha@2.4.5/bower.zip`; I was getting a HTTP 500)

ScottFreeCode:
I mean, could we explicitly add it (overriding .gitignore) only in the tagged commits for each release and then take it back out in the next commit after so it's in VCS for Bower looking at the tags but .gitignore keeps people from committing it in general?

boneskull:
@ScottFreeCode that seems like a pain in the butt 😛

dasilvacontin:
Also, think of cdnjs.

PR #2270, opened 2016-05-22
Support test suite on Windows

### LINKED ISSUE #1813: Tests are broken in Windows

TimothyGu:
Based on work by Tyler Waters <tyler.waters@gmail.com> in #1814.

This commit contains the following changes:

- Add quotation marks to Makefile variables for programs. The variables
  use POSIX-style paths, which cannot be used on Windows to launch a
  program except when quoted. Using double quotation marks instead of
  single since the latter is not available on Windows
- Use os-tmpdir module to get an acceptable directory for temporary
  usage instead of relying on the POSIX /tmp
- Use process.execPath as an authoritative path for Node.js executable
- Detect whether symbol links are supported on the platform before
  testing. On Windows, creating symlinks can fail since it needs
  additional user permissions
- Fix hook tests. The tests parse output of the "dot" reporter to
  separate output of individual tests. The "dot" reporter uses "·"
  symbol (U+2024 ONE DOT LEADER) under POSIX environments and "." symbol
  (U+002E FULL STOP) under Windows, which means that having "." in the
  echoed message makes it ambiguous to be parsed in Windows. To fix this
  issue, two separate changes are necessary:
  - Use a dynamically created regular expression to split the tests
    based on the specific dot character used on the platform
  - Replace "." with "-" in echoed messages in fixtures for hook tests
    to avoid ambiguity with the character output by the reporter

Changes from #1814 include:

- Rebasing
- Use process.execPath as an authoritative path for Node.js executable
- Avoid external dependencies for child_process.spawn()
- Detect whether symbolic links are supported on the platform before
  testing. On Windows, creating symlinks can fail since it needs
  additional user permissions

Fixes #1813.

### DIFF
--- a/test/acceptance/fs.js
+++ b/test/acceptance/fs.js
@@ -1,16 +1,18 @@
 var fs = require('fs');
+var path = require('path');
+var tmpFile = path.join.bind(path, require('os-tmpdir')());
 
 describe('fs.readFile()', function(){
   describe('when the file exists', function(){
     it('should succeed', function(done){
-      fs.writeFile('/tmp/mocha', 'wahoo', done)
+      fs.writeFile(tmpFile('mocha'), 'wahoo', done)
     })
   })
 
   describe('when the file does not exist', function(){
     it('should fail', function(done){
       // uncomment
-      // fs.readFile('/tmp/does-not-exist', done);
+      // fs.readFile(tmpFile('does-not-exist'), done);
       done();
     })
   })
--- a/test/acceptance/lookup-files.js
+++ b/test/acceptance/lookup-files.js
@@ -1,56 +1,83 @@
 var utils = require('../../lib/utils');
 
 describe('lookupFiles', function() {
-  var fs = require('fs'), path = require('path'), existsSync = fs.existsSync ||
-    path.existsSync;
+  var fs = require('fs'),
+    path = require('path'),
+    existsSync = fs.existsSync || path.existsSync,
+    tmpDir = require('os-tmpdir')(),
+    tmpFile = path.join.bind(path, tmpDir),
+    symlinkSupported = false;
+
+  fs.writeFileSync(tmpFile('mocha-utils.js'), 'yippy skippy ying yang yow');
+  try {
+    fs.symlinkSync(tmpFile('mocha-utils.js'), tmpFile('mocha-utils-link.js'));
+    symlinkSupported = true;
+  } catch (ignored) {
+  }
+
+  cleanup();
 
   beforeEach(function() {
-    fs.writeFileSync('/tmp/mocha-utils.js', 'yippy skippy ying yang yow');
-    fs.symlinkSync('/tmp/mocha-utils.js', '/tmp/mocha-utils-link.js');
+    fs.writeFileSync(tmpFile('mocha-utils.js'), 'yippy skippy ying yang yow');
+    if (symlinkSupported) {
+      fs.symlinkSync(tmpFile('mocha-utils.js'), tmpFile('mocha-utils-link.js'));
+    }
   });
 
-  it('should not choke on symlinks', function() {
-    expect(utils.lookupFiles('/tmp', ['js'], false))
+  (symlinkSupported ? it : it.skip)('should not choke on symlinks', function() {
+    expect(utils.lookupFiles(tmpDir, ['js'], false))
       .to
-      .contain('/tmp/mocha-utils-link.js')
+      .contain(tmpFile('mocha-utils-link.js'))
       .and
-      .contain('/tmp/mocha-utils.js')
+      .contain(tmpFile('mocha-utils.js'))
       .and
       .have
       .length(2);
-    expect(existsSync('/tmp/mocha-utils-link.js'))
+    expect(existsSync(tmpFile('mocha-utils-link.js')))
       .to
       .be(true);
-    fs.renameSync('/tmp/mocha-utils.js', '/tmp/bob');
-    expect(existsSync('/tmp/mocha-utils-link.js'))
+    fs.renameSync(tmpFile('mocha-utils.js'), tmpFile('bob'));
+    expect(existsSync(tmpFile('mocha-utils-link.js')))
       .to
       .be(false);
-    expect(utils.lookupFiles('/tmp', ['js'], false))
+    expect(utils.lookupFiles(tmpDir, ['js'], false))
       .to
       .eql([]);
   });
 
   it('should accept a glob "path" value', function() {
-    expect(utils.lookupFiles('/tmp/mocha-utils*', ['js'], false))
+    var res = utils.lookupFiles(tmpFile('mocha-utils*'), ['js'], false)
+      .map(path.normalize.bind(path));
+
+    var expectedLength = 0;
+    var ex = expect(res)
       .to
-      .contain('/tmp/mocha-utils-link.js')
-      .and
-      .contain('/tmp/mocha-utils.js')
-      .and
+      .contain(tmpFile('mocha-utils.js'));
+    expectedLength++;
+
+    if (symlinkSupported) {
+      ex = ex.and
+        .contain(tmpFile('mocha-utils-link.js'));
+      expectedLength++;
+    }
+
+    ex.and
       .have
-      .length(2);
+      .length(expectedLength);
   });
 
-  afterEach(function() {
+  afterEach(cleanup);
+
+  function cleanup() {
     [
-      '/tmp/mocha-utils.js',
-      '/tmp/mocha-utils-link.js',
-      '/tmp/bob'
+      'mocha-utils.js',
+      'mocha-utils-link.js',
+      'bob'
     ].forEach(function(path) {
       try {
-        fs.unlinkSync(path);
+        fs.unlinkSync(tmpFile(path));
       } catch (ignored) {
       }
     });
-  });
+  }
 });
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-BROWSERIFY := node_modules/.bin/browserify
-ESLINT := node_modules/.bin/eslint
-KARMA := node_modules/.bin/karma
+BROWSERIFY := "node_modules/.bin/browserify"
+ESLINT := "node_modules/.bin/eslint"
+KARMA := "node_modules/.bin/karma"
 
 REPORTER ?= spec
 TM_BUNDLE = JavaScript\ mocha.tmbundle
--- a/mocha.js
+++ b/mocha.js
@@ -2474,7 +2474,7 @@ function coverageClass(coveragePctg) {
   return 'terrible';
 }
 
-}).call(this,require('_process'),"/lib/reporters")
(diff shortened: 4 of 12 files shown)


test/acceptance/fs.js
```diff
@@ -1,16 +1,18 @@
...
   })
 
   describe('when the file does not exist', function(){
     it('should fail', function(done){
       // uncomment
-      // fs.readFile('/tmp/does-not-exist', done);
+      // fs.readFile(t('does-not-exist'), done);
       done();
     })
   })
```
ScottFreeCode: Note to self: go through the test suites and find things like this that are failure-tests but either commented out or otherwise unusually handled and if possible change them to assert that the failure throws or passes an error to a callback or whatever the expectable failure mode is.

dasilvacontin: YES ^

test/acceptance/lookup-files.js
```diff
@@ -1,56 +1,79 @@
...
+    if (symlinkSupported) {
+      ex = ex.and
+        .contain(t('mocha-utils-link.js'));
+    }
+
+    ex.and
       .have
-      .length(2);
+      .length(1 + (+symlinkSupported));
   });
```
ScottFreeCode: For the expected length, I'd rather see something like:
```js
// at start of test
var expectedLength = 0;
//...after the .contain(t('mocha-utils.js'));
expectedLength += 1;
//...after the .contain(t('mocha-utils-link.js'));
expectedLength += 1;
//...instead of .length(1 + (+symlinkSupported));
.length(expectedLength);
```
This is, in my opinion, clearer than using the unary `+` to cast the `symlinkSupported` bool to the integer 1, and more flexible in the event this test were ever to be changed such that there would be more than one additional file added if `symlinkSupported`. Although a simpler alternative that requires manual counting (instead of adding a `+= 1` line to everywhere the count should increase) would be to just change `.length(1 + (+symlinkSupported));` to `.length(1 + (symlinkSupported ? 1 : 0));` or `.length(symlinkSupported ? 2 : 1);`.

ScottFreeCode:
- I think I prefer `tmpFile(...)` over `t(...)` for clarity.
- I'd like to see the reason for the changes of `"2.1"`->`"2-1"` documented in the commit logs -- not necessarily thoroughly, but enough that if I see the `.`-to-`-` change in the file diffs, then look at the commit log to find out why, it will both point me in the direction of the code I need to understand and fill in the knowledge of Windows needed to see why that code had to be treated differently, without having to look up the GitHub issue.
- See inline comment about `.length(1 + (+symlinkSupported));`.

Other than that, looks good to me.

test/acceptance/lookup-files.js
boneskull: In agreement.

boneskull:
@TimothyGu Thanks again so much for this!

test/acceptance/fs.js
ScottFreeCode: Moved this into its own issue, since it's nothing to do with this PR: #2309

TimothyGu:
@ScottFreeCode, all comments have been fixed in the latest iteration of the patch.

boneskull:
*fires up virtualbox*

boneskull:
> The "dot" reporter uses "·" symbol (U+2024 ONE DOT LEADER) under POSIX environments and "." symbol (U+002E FULL STOP) under Windows, which means that having "." in the echoed message makes it ambiguous to be parsed in Windows. 

This is crazy.  Awesome catch

boneskull:
@TimothyGu This is really close, but something is killing IE8 tests--unless that's brokenness on SauceLabs' end.  The browser won't "capture".  Trying to find out more.

Related, I'm getting this setup on AppVeyor.

TimothyGu:
@boneskull, is there a publicly available link to look at Sauce Labs output? Either way, that sounds like a Sauce Labs problem...

boneskull:
@TimothyGu I *think* [this is public](https://saucelabs.com/beta/tests/b181b02c723a48e68a61a43ffeb4b008/commands)

Anyway, I reported an issue to SauceLabs, hoping they can point me in the right direction.  I don't know how to debug this myself--do you?

TimothyGu:
> I don't know how to debug this myself--do you?

Nope; in fact, I am totally at a loss here on how to interpret the Sauce Labs reports. Unfortunately I don't think I own a computer with IE8 either.

ScottFreeCode:
I was able to get Karma's tests to trigger IE8 compatibility mode on IE11 by adding a meta tag to Karma's HTML files. Let's see, I think it was... `<meta http-equiv="x-ua-compatible" content="IE=8">` in either `node_modules/karma/static/client.html` or `node_modules/karma/static/context.html` or maybe both. Probably has to be the first thing in `<html><head>`, or maybe the first thing except for a charset meta? There's also manual running of tests (create by hand a page that pulls in the test files you want) with the developer toolbar's compatibility-mode setting, if you want, say, to set allowUncaught and open up the debugger. Either of those could help dig into IE8 issues unrelated to Sauce. The downside is there are a few truly hairy corner cases where things will behave differently in real IE8 than in mere compatibility mode -- I forget what exactly it was, but one of the things I had to fix for the browser testing PR I was only getting the failure on Sauce and not on my local Karma runs (I want to say it was something involving the behavior of replacing `toString`).

ScottFreeCode:
I haven't tried to run it with SauceLabs yet, but preliminary check that it runs on Windows yielded just a couple other bumps:
- In Makefile, `$(MAKE)` may need to be `"$(MAKE)"` (or maybe I'm not setting make right on Cygwin/Git Shell?)
- We're going to have to deal with line endings in the integration diffs tests. Seems like forcing the contents of test/integration/fixtures/diffs to Unix line endings might resolve the issue; alternatively, however, test/integration/diffs.js could try to normalize the line endings before comparison...?

boneskull:
going to close this PR to send a different one.  @TimothyGu 's changes are in it

TimothyGu:
> - In Makefile, `$(MAKE)` may need to be `"$(MAKE)"` (or maybe I'm not setting make right on Cygwin/Git Shell?)

:sigh: Technically, no, since `$(MAKE)` might contain command line options that should not be quoted. If they are, the shell will interpret the options as if they are part of the executable file. I'm not certain enough to say it's a problem with your setup, but then it looks the most possible.

> - We're going to have to deal with line endings in the integration diffs tests. Seems like forcing the contents of test/integration/fixtures/diffs to Unix line endings might resolve the issue; alternatively, however, test/integration/diffs.js could try to normalize the line endings before comparison...?

Hmm, I didn't meet this error. It's entirely possible that the Git configuration has something to do with this. I set Git's `core.autocrlf` to `input` which resulted in UNIX line endings for those files.

We can force Git to store those files with UNIX line endings on all systems using a `.gitattributes` file, or of course we can normalize the test toolchain. Normalization sounds more robust, but the `.editorconfig` file mandates the use of UNIX line endings on all files.

boneskull:
👍  for `.gitattributes`

boneskull:
@TimothyGu at any rate, I think the proof's [in the pudding](https://ci.appveyor.com/project/boneskull/mocha/build/17)

ScottFreeCode:
Ah, I see. I guess I probably set `MAKE` wrong then; I am not that familiar with configuring it, on any system, thanks to packaging on Linux and CMake on Windows.

Looks as though the makefile being run on AppVeyor has some difficulties we don't run into using it in Cygwin-type shells, such as ["find" being the unrelated Windows tool](https://ci.appveyor.com/project/boneskull/mocha/build/17/job/njpmatvtdrug258n#L613). (Also, it's getting a printout of ESLint's help options right after that, so I'm not sure what's going wrong there.) I wonder if we could configure it to run in something like Cygwin until we can replace the makefile with something more cross-platform.

boneskull:
Appveyor has Cygwin and we can use its tools but I don't know if actually
running it in Cygwin is an option. I feel like we should't and it should
just work with a vanilla `cmd`
On Fri, Jul 1, 2016 at 14:47 Scott Santucci <notifications@github.com>
wrote:

> Ah, I see. I guess I probably set MAKE wrong then; I am not that familiar
> with configuring it, on any system, thanks to packaging on Linux and CMake
> on Windows.
>
> Looks as though the makefile being run on AppVeyor has some difficulties
> we don't run into using it in Cygwin-type shells, such as "find" being
> the unrelated Windows tool
> <https://ci.appveyor.com/project/boneskull/mocha/build/17/job/njpmatvtdrug258n#L613>.
> (Also, it's getting a printout of ESLint's help options right after that,
> so I'm not sure what's going wrong there.) I wonder if we could configure
> it to run in something like Cygwin until we can replace the makefile with
> something more cross-platform.
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub
> <https://github.com/mochajs/mocha/pull/2270#issuecomment-230054762>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AA4bMa4u16PfIwFt1A_a5FGF_f2YrFjrks5qRYsEgaJpZM4Ij45S>
> .
>
-- 
Christopher Hiller
https://boneskull.com | https://github.com/boneskull

TimothyGu:
I encountered the find.exe problem too. The only clean way to fix it is to make the MinGW or Cygwin bin path first in the env var, which the AppVeyor PR already does.

Issue #2286, opened 2016-05-29
Regression after doesn't execute if tests skipped in beforeEach


dasilvacontin:
```js

after(function () {
  console.log('after outside suite')
})

describe('suite', function () {
  var count = 0

  beforeEach(function () {
    if (count > 2) this.skip()
  })

  after(function () {
    console.log('after in suite')
  })

  it('test 1', function () { ++count })
  it('test 2', function () { ++count })
  it('test 3', function () { ++count })
  it('test 4', function () { ++count })
  it('test 5', function () { ++count })
  it('test 6', function () { ++count })
})
```
will not output the `after in suite` since v2.5.0 due to PR https://github.com/mochajs/mocha/pull/2081 (found via `git bisect`)

reported by @blueice349, confirmed by myself

dasilvacontin:
It doesn't happen if the `skip` happens inside the `it`.

Simplified case:

```js
after(function () { console.log('after outside suite') })

describe('suite', function () {
  beforeEach(function () { this.skip() })
  after(function () { console.log('after in suite') })
  it('test', function () {})
})
```

boneskull:
See PR #2571

laughinghan:
@dasilvacontin Hey, we're discussing in #2571 what the right behavior is when you call `this.skip()` inside `beforeEach()`, care to weigh in there and explain your use case?

PR #2287, opened 2016-05-29
Add failing test: after hook is not run if test skipped in beforeEach

dasilvacontin:
Related to #2286.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -302,7 +302,8 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
-          suite.pending = true;
+          self.test.pending = true;
+          console.log(self.test);
         } else {
           self.failHook(hook, err);
 
@@ -516,7 +517,7 @@ Runner.prototype.runTests = function(suite, fn) {
     // execute test and hook(s)
     self.emit('test', self.test = test);
     self.hookDown('beforeEach', function(err, errSuite) {
-      if (suite.isPending()) {
+      if (test.isPending()) {
         self.emit('pending', test);
         self.emit('test end', test);
         return next();
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -302,8 +302,13 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
-          self.test.pending = true;
-          console.log(self.test);
+          if (name === 'beforeEach' || name === 'afterEach') {
+            self.test.pending = true;
+          } else {
+            suite.tests.forEach(function(test) {
+              test.pending = true;
+            });
+          }
         } else {
           self.failHook(hook, err);
 
@@ -844,8 +849,8 @@ function filterLeaks(ok, globals) {
     }
 
     // in firefox
-    // if runner runs in an iframe, this iframe's window.getInterface method not init at first
-    // it is assigned in some seconds
+    // if runner runs in an iframe, this iframe's window.getInterface method
+    // not init at first it is assigned in some seconds
     if (global.navigator && (/^getInterface/).test(key)) {
       return false;
     }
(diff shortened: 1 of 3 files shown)


boneskull:
@dasilvacontin I think this fixes it, but I need a sanity check

boneskull:
cc @mochajs/mocha or anyone else who wants to test this out

boneskull:
as per discussion, I've updated my code to be more weird, but also more backwards-compatible.

cc @mochajs/mocha

ScottFreeCode:
I'm wary of depending on the `run` interface when all we want to test is hooks, and of splitting the test out into a separate file so you have to look in those two separate places to figure it out, but the only alternative I can think of will throw from an after or after each hook in case of failure instead (which is arguably an abuse):
```js
describe('issue-2286: after doesn\'t execute if test was skipped in beforeEach', function () {
  var afterWasRun = false
  describe('suite with skipped test for meta test', function () {
    beforeEach(function () { this.skip() })
    after(function () { afterWasRun = true })
    it('should be pending', function () {})
  })
  after('meta test', function () {
    afterWasRun.should.be.ok();
  })
})
```

test/integration/fixtures/regression/issue-2286.js
```diff
@@ -0,0 +1,5 @@
+describe('suite', function () {
+  beforeEach(function () { this.skip() })
+  after(function () { console.log('after in suite') })
+  it('test', function () {})
+})
```
ScottFreeCode: How about some names that make this file's function obvious (both in viewing this file and if it were to end up run among the main tests in addition to the meta test calling it by `run`)? e.g.:
describe: 'testception for issue 2286 meta test'
beforeEach: 'skips tests'
after: 'should run'
it: 'skipped by beforeEach'

lib/runner.js
```diff
@@ -302,7 +302,13 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
-          suite.pending = true;
+          if (name === 'beforeEach' || name === 'afterEach') {
+            self.test.pending = true;
+          } else {
+            suite.tests.forEach(function(test) {
+              test.pending = true;
+            });
```
ScottFreeCode: Couldn't it use `suite.pending` if it's not `beforeEach` or `afterEach`? Or in the older versions does `after` run even if `before` calls this.skip()?

test/integration/fixtures/regression/issue-2286.js
boneskull: @ScottFreeCode I don't see this stuff as absolutely necessary given the filename, but if you wish to add to the branch, go for it

boneskull: actually on second thought, I like your test above.  please use it instead

lib/runner.js
boneskull: setting the suite to be pending, at this point, will cause the `after` hook to not get executed.

boneskull: (so, the answer is "no")

test/integration/fixtures/regression/issue-2286.js
ScottFreeCode: (to the first reply:) Yeah, it's pretty clear from the filename what issue this is testing, but it's not so obvious to me from looking just at this file that this is meant to be run by another test in another file that tests the output of this one (unless that's what all the stuff in this fixtures folder is -- I haven't looked at the files that were already there), so it looked at first as though this weren't actually testing it programmatically but only printing something to look at manually.

ScottFreeCode: (to the second:) You mean the one that uses nested suites with code like this in the inner one and a test in the `after` of the outer one?

ScottFreeCode: (Went ahead and committed both; feel free to rebase away either you don't want.)

test/integration/regression.js
```diff
@@ -1,4 +1,4 @@
-var assert = require('assert');
+﻿var assert = require('assert');
 var fs     = require('fs');
```
ScottFreeCode: Did this... pick up a UTF-8 BOM or something? I will see if I can fix that when I get another free moment.

ScottFreeCode: Or not, if it's already merged into the history and isn't hurting anything...

PR #2299, opened 2016-06-08
Fix % always 0 in HTML reporter

AviVahl:
Hi mocha team.
This is a small fix for the progress bar staying consistently at 0 when using the HTML reporter.
Seems like a regression that happened due to the "double-error" fix/refactor.

Reasoning for the changes:
updateStats() accesses this.total, but this === the global object (window) if not specified explicitly here.
Also added a missing update when all tests end.

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -125,6 +125,7 @@ function HTML(runner) {
 
   runner.on('suite end', function(suite) {
     if (suite.root) {
+      updateStats.call(this);
       return;
     }
     stack.shift();
@@ -137,7 +138,7 @@ function HTML(runner) {
     var el = fragment(markup, test.speed, test.title, test.duration, url);
     self.addCodeToggle(el, test.body);
     appendToStack(el);
-    updateStats();
+    updateStats.call(this);
   });
 
   runner.on('fail', function(test) {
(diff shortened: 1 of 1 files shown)


lib/reporters/html.js
```diff
@@ -125,6 +125,7 @@ function HTML(runner) {
 
   runner.on('suite end', function(suite) {
     if (suite.root) {
+      updateStats.call(this);
```
boneskull: can we just use `Function.prototype.bind()` on `updateStats` instead?

AviVahl:
@boneskull Thanks for reviewing.

You could bind `updateStats` to `runner` and set it as `this.updateStats`.
Both approaches would work. You would still need to add a call on 'suite end'.
I wasn't sure if you want `updateStats` set on the reporter instance here, so I used the a explicit call. Please clarify.

boneskull:
@AviVahl It looks like `updateStats()` is always called with the context of the `Runner` object, so actually `call` and `bind` shouldn't be necessary, yes?

AviVahl:
Calling `updateStats()` in a block inside a context does **not** set that context to the call invocation.
I might have an alternative patch which you would like more.
I'll commit.

AviVahl:
@boneskull new patch allows calling updateStats() directly (without specifying this), as it binds runner to extract the value of `total`.
Please let me know if this is a nicer solution in your opinion.

AviVahl:
@boneskull Can we get this small bug-fix reviewed again?
Two lines change for such an annoying UI issue.
Thanks in advance.

boneskull:
@AviVahl looks good, thanks!

Issue #902, opened 2013-06-18
🚀 Feature: --order option for random test order


sarenji:
An `--order` option would allow people to uncover order dependencies. The three options would be `--order random`, `--order random:seed`, and `--order default`. Each randomized suite outputs the seed it used.

RSpec implements this, but their default order is random. Mocha doesn't have to do that. Some details on their `--order` parameter are here: http://blog.davidchelimsky.net/2012/01/04/rspec-28-is-released/

What do you think?

TimothyGu:
First, it looks like a Fisher-Yates shuffle would do the job here.

Second, I'd rather have `--order random`, `--order random-suites`, and `--order default` as the three arguments, with an optional `:<seed>`.

demisx:
+1. Just found a bug that would've shown up a long time ago if tests were randomized. Similar to how RSpec supports it.

neontapir:
Here's some code that illustrates the usefulness of random test ordering. While there are simpler examples, this is one I just encountered during a TDD demo. If you reverse the order of the tests, the first test always fails.

game.js:

	var express = require('express');
	app = exports.app = express();

	var sum = 0;

	app.post('/bowl/:pins', function(req,res) {
		var score = parseInt(req.params.pins);
		console.log('Bowled ' + score);
		sum += parseInt(req.params.pins);
	});

	app.get('/score', function(req,res) {
		console.log('Sum: ' + sum);
		res.send(sum + '');
	});

	app.listen(process.env.PORT || 3000);

test\gameTest.js:

	var request = require('supertest'),
		should = require('should'),
		game = require('../game.js').app;
		
	describe('a game of bowling', function() {
		describe('a gutter game', function() {
			it('should score 0', function(done){
				request(game).get('/score').expect(200, '0', done);
			});
		});
		
		describe('a single pin game', function() {
			it('should score 20', function(done){
				for(var i = 0; i < 20; i++) {
					request(game).post('/bowl/1').expect(200, done);
				}
				request(game).get('/score').expect(200, '20', done); 
			});
		});
	});

TheSavior:
I would love to have this.

tomconroy:
👍

fishermand46:
Once you get a few globals involved (this is Javascript, remember), start stubbing out server calls, or inserting/removing things from the DOM in tests, it is *very* easy to add some order-dependency to your tests. Randomizing the test order would help discover these early rather than later.
:+1:

TSiege:
:+1:

[...]

PR #2301, opened 2016-06-10
feat: randomization support [WIP REFERENCE - DO NOT MERGE]

boneskull:
Edit (January 2024, @JoshuaKGoldberg): This PR is just here as a reference to support #902. If you want to see this feature land, send a PR yourself. See https://github.com/mochajs/mocha/issues/902#issuecomment-1902527501. Thanks! ✨ 

---

- Adds randomization support via `--random` CLI flag or `random` option in browser
- Random seed is created if not supplied for deterministic test runs

**QUESTIONS**:
### Automatic Seed Creation

How should we handle automatic seed creation?  

Issues:
- Some reporters (XUnit in particular) are extremely fussy about output.
- If we were to simply echo the automatically created seed to the console, problems would happen

Suggestions:
- In the browser, this can simply be displayed.
- Output a `.mocha-seed.txt` file or something.  This file would not be read from; only written to.
- Disable automatic seed creation; _require_ a seed for operation.  Trivially generate a seed via `Mocha.seed()` in the browser, or `--generate-seed` on the CLI.

### "Specification"

- "Automatic" seed generation is not (yet?) available, due to concerns about output and reporter support / conflicts.
- A seed is required for randomization.
- On the command line:
  - Mocha can generate a seed via `mocha --generate-seed`.  A random seed will be printed to the console.
  - `--random <seed>` will enable randomization.
- In the browser:
  - Mocha can generate a seed via `mocha.seed()`.  
  - Enable randomization thru `mocha.options({random: mocha.seed()})` before calling `mocha.run()`.
- For Mocha's purposes, a seed is a *positive integer* which is *less than or equal to* the value of `Number.MAX_SAFE_INTEGER`, which is typically `9007199254740991`.
- Acceptable formats for seeds are numbers (decimal, octal, hexadecimal) or strings that will convert via `parseInt()`.
- Mocha will use a hexadecimal representation for display purposes.
- When randomizing tests, they are only randomized within the context of a suite.  In other words, suites will still run in order, but the tests within them will not.

### TODO

- [x] Add `--generate-seed`
- [x] Add `--random <seed>`
- [x] Seed validation
- [x] Unit tests
- [ ] Display active seed in HTML reporter
- [ ] Docs
- [ ] Support `MOCHA_SEED` environment variable.  This would ease usage in CI envs.
- [ ] Node integration tests
- [ ] Browser integration tests


boneskull:
cc @mochajs/mocha

boneskull:
cc @TimothyGu

lib/mocha.js
```diff
@@ -12,6 +12,8 @@ var escapeRe = require('escape-string-regexp');
 var path = require('path');
 var reporters = require('./reporters');
 var utils = require('./utils');
+var Random = require('random-js');
+var MAX_SAFE_INTEGER = 0x1fffffffffffff;
```
ScottFreeCode: Where is this number coming from?

lib/mocha.js
```diff
@@ -277,15 +283,67 @@ Mocha.prototype.invert = function() {
  * @param {Boolean} ignore
  * @return {Mocha}
  * @api public
- * @param {boolean} ignore
- * @return {Mocha}
  */
```
ScottFreeCode: Some changes like this would be better just applied to master before this PR than applied during this PR; then if we ended up having to revert we wouldn't have to cherry-pick the unrelated changes to keep them, plus it would reduce clutter when reviewing the PR. (I figure something similar goes for `"` to `'` conversion in bin/_mocha, by the way.)

lib/mocha.js
```diff
@@ -277,15 +283,67 @@ Mocha.prototype.invert = function() {
...
+        seed = parseInt(seed, 16);
+      } else {
+        // a decimal number as string, presumably
+        seed = parseInt(seed, 10);
+        if (isNaN(seed)) {
+          throw new Error('Invalid random seed. Expected unsigned 32-bit '
+            + 'integer; got: "' + seed + '"');
+        }
+      }
+    }
```
ScottFreeCode: Is there a particular use case for supporting numbers-in-strings directly in the randomization function? I would think it would be on the caller to do the conversion. Having to specify number conversion behavior in the randomization function itself complicates the API, and for little benefit given that it could be handled separately. And if it were handled separately, it would be easier to update in isolation if need be, or to test, etc.

The one place I can think of it being needed in Mocha is the CLI. Ideally, I'd say the string to seed conversion should be in its own utility function so that in the event it is needed in more than one place it could be reused, so that it can be tested separately from testing the stuff that uses it or the randomization it feeds into, and so we can consider exposing it to users/wrappers of Mocha if they want to get the same conversion method that Mocha uses for the seed.

lib/mocha.js
```diff
@@ -277,15 +283,67 @@ Mocha.prototype.invert = function() {
...
+
+    var engine = Random.engines.mt19937().seed(seed);
```
ScottFreeCode: Is there any option to, say, get a default randomization engine? It's nice to use a library with better randomization quality guaranteed than the browser-variable-quality `Math.random`, and it's not like we need something like cryptographic quality, but it would be cool to be able to pick up improvements to what's considered the average random number generator without having to go into the code and change which one is hardcoded. (For instance, I have heard talk of newer engines that manage to give better quality than the Mersenne Twister *while also being faster*. I don't know if such things are available in common use yet, but it would be neat to pick them up when they are.)

lib/mocha.js
```diff
@@ -277,15 +283,67 @@ Mocha.prototype.invert = function() {
...
+};
+
+/**
+ * Generate a random seed.
+ * @returns {number} Unsigned 32-bit integer
+ */
+Mocha.seed = function seed() {
+  return Random.integer(0, MAX_SAFE_INTEGER)(Random.engines.mt19937()
+    .autoSeed());
+};
```
ScottFreeCode: It'd be nice to have a comment explaining why we use `autoSeed` to generate a seed to start a random generator to use to generate a seed to return rather than just using `autoSeed` to generate a seed to return. (I assume it's something like `autoSeed` spits out floating point or some format that's inadequate for people entering it on the commandline, but I'd like to know for sure... without going and looking up the docs for another library if possible.)

lib/suite.js
```diff
@@ -58,6 +58,7 @@ function Suite(title, parentContext) {
   this._slow = 75;
   this._bail = false;
   this._retries = -1;
+  this._deterministic = false;
```
ScottFreeCode: Bug? I see references to `this._randomized` below instead of `this._deterministic`.

test/acceptance/random.js
```diff
@@ -0,0 +1,63 @@
...
+        .to
+        .be
+        .a('number');
+    });
+
+    it('should return a positive number', function() {
+      expect(Mocha.seed() > 0)
+        .to
+        .be(true);
+    });
```
ScottFreeCode: This will give the nondescript "expected true actual false" thing if it fails. `expect(Mocha.seed()).to.be.above(0)` will be more helpful.

However, the test doesn't match the description; 0 is a positive number. If 0 is a valid output, then [in lieu of a greaterThanOrEqual assertion](https://github.com/Automattic/expect.js/issues/108) we can use `.within(0, Infinity)`. If not, the description should be updated instead.

See: https://github.com/Automattic/expect.js#api

test/acceptance/random.js
```diff
@@ -0,0 +1,63 @@
...
+        expect(mocha.options.randomConfig).not.to.be(undefined);
+      });
+
+      it('should generate a seed', function() {
+        expect(mocha.options.randomConfig.seed).to.be.a('number');
+      });
+
+      it('should generate a hexadecimal representation of a seed', function() {
+        expect(mocha.options.randomConfig.hex).to.be.a('string');
+      });
```
ScottFreeCode: Incomplete test? Description specifies hexadecimal but assertion only tests being a string.

lib/mocha.js
dasilvacontin: I agree. 👍 

Since I tend to make atomic commits, I often just cherry-pick them into `master` and delete them from the branch before creating the PR.

lib/mocha.js
dasilvacontin: I'd rather get the number from the JS API or, if not available, an npm module.

dasilvacontin:
Thanks for putting time into this one peeps!

lib/mocha.js
TimothyGu: I wonder if quality of a PRNG really matters for Mocha. Mersenne Twister famously has a very long period although not being that strong or fast, but period is basically what we need here because we want a maximum number of test order combinations. With that in mind, sure, newer faster and altogether _better_ PRNGs are cool, but not really useful.

lib/mocha.js
TimothyGu: From https://www.npmjs.com/package/random-js#distributions

> - `Random.integer(min, max)(engine)`: Produce an integer within the inclusive range [`min`, `max`]. `min` can be at its minimum -9007199254740992 (-2 ** 53). `max` can be at its maximum 9007199254740992 (2 ** 53).

`Number.MAX_SAFE_INTEGER` is the same value.

boneskull: Yes, sorry.  `0x1fffffffffffff` is the value of `Number.MAX_SAFE_INTEGER` *if* `Number.MAX_SAFE_INTEGER` is implemented.  But we're not guaranteed that, of course, and I didn't want to pull in some polyfill.  I'll add a comment.

boneskull: @dasilvacontin Would prefer not to `npm install --save max-safe-integer` (don't know if that exists; probably does) if I can help it.

lib/mocha.js
boneskull: Good catch, I'll nix this.

lib/mocha.js
boneskull: Right; string to integer conversion is necessary for the CLI only.  At some point, we'll have to do it.  I can move the implementation elsewhere.

lib/mocha.js
boneskull: @TimothyGu I'm unfamiliar with this stuff; what's "period"?

I used Mersenne Twister because the lib purported that it was "more random".  I think "more random than `Math.random()`" is reasonable.

This library does not have a "default" engine, no.  If anyone can recommend a simpler library--having no dependencies--with seed support, and reasonable randomization, I'm happy to use it.

boneskull: (But really, I suppose, as long as we have seeding support and a reasonable "maximum number of test order combinations", as @TimothyGu writes, then I guess we're ok)

lib/mocha.js
boneskull: That comment would be something like "go read the docs for this other library".  😜 

Basically, you can't get a value out of the engine unless it's seeded.  So if we want a seed, then we should randomly generate one.  And if we're randomly generating a seed, why not use the random number generator..?

lib/suite.js
boneskull: yep, missed this, ty

test/acceptance/random.js
boneskull: Ah, good idea to use `within()`.  I'm not familiar with `expect.js`.

test/acceptance/random.js
boneskull: Will match it

boneskull:
@ScottFreeCode @TimothyGu Thanks for looking this over.  I'll implement the suggested changes

boneskull:
I don't think my questions in [the description](https://github.com/mochajs/mocha/pull/2301#issue-159541051) were answered though.  

If I'm deciding, I'll go with `--generate-seed` so I don't have to deal with the file system.

lib/mocha.js
TimothyGu: The length of the period is the number of random numbers generated by the PRNG before it is repeated (I believe). For MT19937, the period is 2^19937 − 1.

Where did it say that it is more random?

Either way, I'm still for leaving the code as it is. MT is more than enough for what we need.

lib/mocha.js
boneskull: Hmm, this may have been just a weird merge.  Anyway, I'll sort it out.

lib/mocha.js
TimothyGu: @boneskull, I don't think you understood @ScottFreeCode's question. The question was if `autoSeed()` generates a number that can be directly returned, i.e. if the following was possible

```js
Mocha.seed = function seed() {
  return Random.engines.mt19937().autoSeed();
};
```

The answer seems to be no, since `autoSeed` **doesn't return a seed;** rather it returns a **seeded random generator** using `Random.generateEntropyArray`. The generator is fed into the function returned by `Random.integer`. `generateEntropyArray` returns an array instead of a seed number.

There's also a minor inaccuracy in the JSDoc (`s/32/53/`), assuming we don't want to allow negative seeds (or random-js doesn't support them).

Altogether the code should IMO look like this:

```js
/**
 * Generate a random seed.
 * @returns {number} Unsigned 53-bit integer
 */
Mocha.seed = function seed() {
  /*
   * Create an auto-seeded random number generator.
   * We can't use `Random.generateEntropyArray` (used by `autoSeed`)
   * for this function since our seed is an integer.
   */
  var generator = Random.engines.mt19937().autoSeed();

  /*
   * Return a random integer that can be used as a seed, using the
   * random number generator we just created.
   */
  return Random.integer(0, MAX_SAFE_INTEGER)(generator);
};
```

TimothyGu:
I'm not familiar with the current situation of browser testing. How do reporters interact with browsers? Can I use, say, the XUnit reporter with the browser? If yes, will the XUnit output be made to the browser console?

boneskull:
so this is how it works on the CLI:

```shell
# generate the seed first
$ mocha --generate-seed 
0x10FFFA5D98AE6F
```

```shell
# run tests with the seed
$ mocha --random 0x10FFFA5D98AE6F test/acceptance/random.js --reporter spec


  random option
    seed
      ✓ should be a valid seed
      ✓ should return a finite number
      ✓ should return a hex string
    when Mocha is instantiated with "random" option
      and that option is "false"
        ✓ should not populate the "randomConfig" option
      and that option is undefined
        ✓ should not populate the "randomConfig" option
      and that option is not an integer or convertable string
        ✓ should throw
      and that option is a valid seed
        randomConfig
          ✓ should define a function to shuffle tests in the "shuffleTests" prop
          ✓ should define a "seed" prop
          ✓ should be defined
          ✓ should define a hexadecimal representation in the "hex" prop


  10 passing (24ms)
```

For comparison, without randomization:

```shell
$ mocha test/acceptance/random.js  --reporter spec


  random option
    seed
      ✓ should return a hex string
      ✓ should be a valid seed
      ✓ should return a finite number
    when Mocha is instantiated with "random" option
      and that option is "false"
        ✓ should not populate the "randomConfig" option
      and that option is undefined
        ✓ should not populate the "randomConfig" option
      and that option is not an integer or convertable string
        ✓ should throw
      and that option is a valid seed
        randomConfig
          ✓ should be defined
          ✓ should define a "seed" prop
          ✓ should define a hexadecimal representation in the "hex" prop
          ✓ should define a function to shuffle tests in the "shuffleTests" prop


  10 passing (24ms)
```

boneskull:
@TimothyGu Generally, if you're in a browser, your only available reporter is the HTML reporter.

If you're in a *headless* reporter, then the HTML reporter is going to just dump HTML into your terminal, and you'll want one of the console-based reporters, like `spec`.  I would also assume something with the complexity of `nyan` would not work in this context.  Furthermore, I doubt a reporter like `json-stream` would work in a PhantomJS context, but who knows.  

I don't use Mocha in a browser context without Karma to hold my hand, so I'm not the best source of information on this topic.  @mantoni is much more qualified to elaborate on which reporters are supported, if he cares to.

boneskull:
What's left to do:  

- [ ] Display active seed in HTML reporter
- [ ] Write integration tests to assert deterministic, random order of tests within a suite context
- [ ] Docs
- [ ] Support MOCHA_SEED environment variable or something.  This would ease usage in CI envs.

boneskull:
note to self:

Karma setup would look something like:

```js
var seed = require('mocha').seed();
console.log('Mocha seed: ' + seed);
config.set({
  frameworks: ['mocha'],
  client: {
    random: seed
  }
});
```

lib/utils.js
```diff
@@ -748,3 +757,52 @@ exports.stackTraceFilter = function() {
...
+ * Given integer `int`, try to convert it to a hex string.
+ * @param {number} int Integer
+ * @returns {string} Hexadecimal representation thereof
+ */
+exports.toHex = function toHex(int) {
+  return '0x' + int.toString(16).toUpperCase();
+};
+
+/**
+ * Returns `true` if `value` is a "safe" 32-bit unsigned int.
```
TimothyGu: Here and in various other places, it's a bit confusing to see uint32_t or integers in the range of [0, 2³² – 1] being called "safe," while what the function actually takes is integers in the range of [1, 2⁵³ – 1]. IMO, what we should allow is [0, 2⁵³ – 1] which can be described as "safe" unsigned 53-bit integer and matches with ECMAScript semantics. (In addition, we can optionally allow negative values as well.)

boneskull:
SauceLabs is taking a big ol' dump trying to run our tests in IE8.  Considering the `test-windows`/`appveyor` branch hasn't been merged, it's not specific to my changes or @TimothyGu's changes there.  Curiously, the `pr` tests pass, but the `push` tests fail.  Must have something to do with trying to run in a branch.  At least that's something to go on for SauceLabs--they're waiting for my reply to a ticket about this.

mantoni:
@TimothyGu @boneskull Pretty much all the reporters work in PhantomJS, too. The trick is to redirect console output and to provide a shim for `process.stdout`. I do this with [brout](https://github.com/mantoni/brout.js) and then set `Mocha.process.stdout = process.stdout` (see [mocaccino browser setup](https://github.com/mantoni/mocaccino.js/blob/master/lib/setup-browser.js#L42)). With this setup, even the nyan reporter output is redirected properly to the terminal.

lib/utils.js
```diff
@@ -282,7 +282,9 @@ exports.slug = function(str) {
 exports.clean = function(str) {
   str = str
     .replace(/\r\n?|[\n\u2028\u2029]/g, '\n').replace(/^\uFEFF/, '')
-    // (traditional)->  space/name     parameters    body     (lambda)-> parameters       body   multi-statement/single          keep body content
+    // (traditional)->  space/name     parameters    body     (lambda)->
+    // parameters       body   multi-statement/single          keep body
+    // content
     .replace(/^function(?:\s*|\s+[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\s*\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\s*\}|((?:.|\n)*))$/, '$1$2$3');
```
ScottFreeCode: The comment wrapping here basically breaks this documentation of the regex. Do we need to stop the linter from paying attention to that comment? Or should we break up the regex into parts (as strings) and compose them using `new RegExp`?

ScottFreeCode: The comment wrapping here basically breaks this documentation of the regex. Do we need to stop the linter from paying attention to that comment? Or should we break up the regex into parts (as strings) and compose them using `new RegExp`?

boneskull: didn't notice this. I'll ensure it's left as-is

macalinao:
woohoo finally

boneskull:
So basically needs more tests. But I don't necessarily need to be the person to write them.

Robdel12:
Is there an updated description of what needs to happen to push this over the line? We would like to randomize our mocha builds to bring the flaky tests to the surface.

boneskull:
not so much.  new features are frozen until we get traction on immediate coverage needs.  though I haven't had any time to pay attention to where we're at there

craigtaub:
@boneskull this sounds good. Worth pursuing now coverage in better place? Or do you think given where codebase is now there's a better alternative approach here?

lib/utils.js
```diff
@@ -767,6 +776,36 @@ exports.stackTraceFilter = function() {
...
+ */
+exports.toSafeUnsignedInteger = function toSafeUnsignedInteger(value) {
+  var intVal;
+  if (/^0[xX]/.test(value)) {
+    intVal = parseInt(value, 16);
+  } else if (/^0[^.]/.test(value)) {
+    intVal = parseInt(value, 8);
+  } else {
+    intVal = parseInt(value, 10);
+  }
```
fearphage, line 802: It feels like this could be simplified a bit.

```suggestion
  var base = 10;
  var intVal;
  if (/^0[xX]/.test(value)) {
    base = 16;
  } else if (/^0[^.]/.test(value)) {
    base = 8;
  }
  
  intVal = parseInt(value, base);
```

lib/utils.js
```diff
@@ -767,6 +776,36 @@ exports.stackTraceFilter = function() {
...
+  return !isNaN(intVal) && parseFloat(intVal) === intVal
+    && exports.isSafeUnsignedInteger(intVal)
+    ? intVal
+    : NaN;
```
fearphage, line 806: `isSafeUnsignedInteger` already checks for an integer so there's no need to double check here. Also if the value is `NaN`, you'll still hit the right checks without the additional `NaN` check.

```suggestion
  return exports.isSafeUnsignedInteger(intVal) ? intVal : NaN;
```

lib/utils.js
```diff
@@ -776,3 +815,22 @@ exports.stackTraceFilter = function() {
...
   return typeof value === 'object' && typeof value.then === 'function';
 };
+
+/**
+ * Given integer `int`, try to convert it to a hex string.
+ * @param {number} int Integer
+ * @returns {string} Hexadecimal representation thereof
+ */
+exports.toHex = function toHex(int) {
+  return '0x' + int.toString(16).toUpperCase();
```
fearphage, line 825: This doesn't have any validation compared to other functions.`toHex('foo')` will happily return `0xFOO`. Is that expected behavior?

fearphage (changes requested):

JoshuaKGoldberg:
@fearphage thanks for the reviews, but this PR is just here as a reference. It's got a whole bunch of merge conflicts. I edited the title & OP to make that more clear.

If you're interested in the work (great! 🙌) the right next step would be to send a separate PR.

Issue #2302, opened 2016-06-10
allowUncaught doesn't work in hooks


dima117:
For example if exception was thrown in `beforeEach` hook then it will be caught even if `mocha.allowUncaught()` was executed. This is because during the hook execution the runner not sets `allowUncaught` parameter for it.

ScottFreeCode:
We should check whether this is related to #1985.

dima117:
Hi,

what is tis issue current state?

boneskull:
I think this is still broken.

givanse:
New PR https://github.com/mochajs/mocha/pull/3669

PR #2303, opened 2016-06-10
Sets the allowUncaught parameter for hooks.

dima117:
https://github.com/mochajs/mocha/issues/2302


boneskull:
@dima117 Thanks!  Can you please add a test in `test/runner.js`?

dima117:
I added a test for allowUncaught parameter in hooks.

dima117:
Do I need to do something else for this PR?

dima117:
@boneskull What is the state of this PR?

ScottFreeCode:
Is it safe to call `suite.beforeEach` inside a test?

dima117:
I think yes. Why it may be unsafe?

boneskull:
@dima117 Apologies for not getting to this earlier.

test/runner.js
```diff
@@ -315,6 +315,17 @@ describe('Runner', function(){
   });
 
   describe('allowUncaught', function() {
+    var immediately;
```
boneskull: Why are we monkeypatching `Runner.immediately()`?

test/runner.js
```diff
@@ -327,6 +338,19 @@ describe('Runner', function(){
       fail.should.throw('allow unhandled errors');
       done();
     });
+
+    it('should allow unhandled errors in hooks to propagate through', function(done) {
```
boneskull: this doesn't appear to be async, so you don't need `done`.

Ideally, I'd like to see both sync and async tests against this behavior.  Can you please add an async version?

test/runner.js
dima117: It was done in order to run the hook synchronously.

test/runner.js
dima117: Thanks, I will correct soon and I will add the asynchronous version

dima117: @boneskull  I'm sorry for delay.

I can not think of a good way, how to check the exception handling asynchronously. Perhaps I not fully understand the problem. Please give me an advice on how to do it correctly.

test/runner.js
boneskull: Why do we need to do that?

dima117: We need it to catch the exception in the test method, instead of catching it on a global error handler.

dima117:
Hello! What is the status of this PR?

boneskull:
I may use some of this, though I don't expect the original author to sign a CLA at this point

boneskull:
see #2302

PR #2325, opened 2016-06-21
use custom glob; closes #2318

### LINKED ISSUE #2318: DoS vulnerability in glob dependency
glob version 3.2.11 includes a dependency on minimatch 0.3.0, which contains a DoS vulnerability as documented here:

https://nodesecurity.io/advisories/118

This should be resolvable by updating to the current version of glob, which uses the up-to-date minimatch version.

boneskull:
This addresses the [security advisory](https://nodesecurity.io/advisories/118) in a non-breaking manner.  It upgrades `minimatch` to `1.0.0`, basically--but from what I can tell, the only meaningful change from `0.3.0` was a feature to support number padding in brace expansion.  

Generally speaking, what we're doing here should be avoided (forking old versions of modules and re-publishing them under new names).  Doing so could potentially expose us to security issues down the road--if an issue is detected in `minimatch` or `glob` and subsequently fixed, we will have to manually backport the fix--assuming we even know about any issue!

I could go either way here; I can't conceive of a manner in which it would affect users of Mocha via usage of `mocha` itself.  This issue does not affect Mocha in the browser.

See the forked/renamed modules [glob-mocha](/mochajs/node-glob) and [minimatch-mocha](/mochajs/minimatch), the latter of which contains a backport for the advisory fix.

cc @ScottFreeCode

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -69,7 +69,7 @@ module.exports = function(config) {
     browserify: {
       debug: true,
       configure: function configure(b) {
-        b.ignore('glob')
+        b.ignore('glob-mocha')
           .ignore('jade')
           .ignore('supports-color')
           .exclude('./lib-cov/mocha');
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -7,7 +7,7 @@
 var basename = require('path').basename;
 var debug = require('debug')('mocha:watch');
 var exists = require('fs').existsSync || require('path').existsSync;
-var glob = require('glob');
+var glob = require('glob-mocha');
 var join = require('path').join;
 var readdirSync = require('fs').readdirSync;
 var statSync = require('fs').statSync;
--- a/package.json
+++ b/package.json
@@ -320,7 +320,7 @@
     "debug": "2.2.0",
     "diff": "1.4.0",
     "escape-string-regexp": "1.0.2",
-    "glob": "3.2.11",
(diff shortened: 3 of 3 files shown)


boneskull:
Without this, I believe `nsp` will complain about *any* package depending upon `mocha`, even if that's a `devDependency`.

boneskull:
Would appreciate another set of eyes on this to ensure I'm not totally out-of-bounds.

cc @mochajs/core

ScottFreeCode:
I haven't reviewed the change itself, but I want to raise a question: Would it be a backwards-incompatible change to Mocha (e.g. due to relevant changes to the behavior of globbing as used by/with Mocha) to fork the latest version of Glob/Minimatch/anything-else-we-have-to and just change the carets? (On the other hand, this might not be scalable -- if for instance the earliest version of library X required by the latest Minimatch has carets in its own dependencies then we also have to fork X to take out the carets... and there may be multiple "X"s in this case?) If it worked, the advantage would be that we can try to keep the forks tracking the current version and getting other fixes relatively easily that way.

In any case, we should start keeping a list of things we want to change in 3.x-not-a-total-rewrite if we hit the point where we can drop Node 0.8 before we hit the point where the rewrite is a viable replacement; updating to a new, non-forked Glob/Minimatch along with following the Pug renaming can both go on it for starters...

boneskull:
I think I want to scrap this idea and just release a major.

PR #2328, opened 2016-06-24
Allow underscores in v8 options

shvaikalesh:
Node allows both `-` (dashes) and `_` (underscores) as separators in v8 flags (e.g. `--allow_natives-syntax`).

This PR allows any v8 flag to be specified with any separator to improve parameters consistency.


bin/mocha
```diff
@@ -16,7 +16,7 @@ var spawn = require('child_process').spawn,
 getOptions();
 
 process.argv.slice(2).forEach(function(arg){
-  var flag = arg.split('=')[0];
+  var flag = arg.split('=', 1)[0].replace(/([a-z])_/g, '$1-');
```
boneskull: would prefer no lines are *changed* in this file.  it appears you can add two new cases for `--es-staging` and `--use-strict` to the hideous `switch` below; this seems like a safer bet.

bin/_mocha
```diff
@@ -81,8 +81,8 @@ program
   .option('--compilers <ext>:<module>,...', 'use the given module(s) to compile files', list, [])
   .option('--debug-brk', "enable node's debugger breaking on the first line")
   .option('--globals <names>', 'allow the given comma-delimited global [names]', list, [])
-  .option('--es_staging', 'enable all staged features')
-  .option('--harmony<_classes,_generators,...>', 'all node --harmony* flags are available')
+  .option('--es-staging', 'enable all staged features')
```
boneskull: I think we can leave this file as-is

boneskull:
@shvaikalesh Hi, thanks for the PR!  I have a couple comments; if you could please address them, we can merge this.

bin/mocha
shvaikalesh: If we do this, we will allow only two flags (`--es-staging` and `--use-strict`) to be written with `_`:

1. There is no real reason to make exceptions for these particular two.
2. That will not allow specify other flags with underscores. If you do `node --v8-options`, you will see that v8 lists options with underscores. If `node` allows two separators, I think mocha should too.

boneskull: I'm not really following, but please let me explain my intent:

1.  `node` only allows interchangeable `-` and `_` in those options listed within `--v8-options`.  For example, you cannot run `node --throw_deprecation`.
2.  This change, as proposed, would actually *add* support for cases such as `--throw_deprecation`.  This is not necessary, since `node` does not support it.
3.  Any modification of the command-line API without integration test coverage is risky.  We don't have a whole lot of coverage against the `mocha` executable in the first place (if any), and I'm almost certain we have *no* coverage for flags shown by `--v8-options`.  Due mainly to lack of integration tests (which would be a pain to write anyway), I'm not convinced that the proposed changes would not break something somebody is doing.
4.  So, adding the two lines as I proposed *should* add support either `-` or `_` for all of the `--v8-options`-type flags the `mocha` executable *currently* supports.  Since no lines are modified (and only added), this is relatively "safer".  Would it not address your use case?

To expand further, the `mocha` allows some--but not all--`node` flags through.  When a flag is added to `node`, someone invariably sends a PR which adds the flag to `mocha` as well.  This doesn't scale very well.  

The only reason (that I'm aware of) for providing *two* executables, `mocha` and `_mocha`, is to allow passing of `node` flags--merely a convenience.  All flags, of course, are supported by running `node [options] /path/to/_mocha`; `mocha` itself only supports a subset, as you can see.

I don't believe I have seen an issue come through where:
- someone needs support for, say, `--use-strict`
- cannot use `--use_strict`
- has no viable workaround

Because a workaround exists, and because the proposed changes are (in my mind, anyway) high-risk and low-reward, I can't really get behind changes to the documented API, nor modification of the option-parsing lines.

hollomancer:
@shvaikalesh, please let us know when you have a chance to address the comments, and we'll take another look!

boneskull:
Closing for inactivity.  Please reopen if you wish!

PR #2329, opened 2016-06-24
fix custom comiler file (absolute path) support in windows system

silentcloud:
`mocha --compilers js:C:\User\fortest\demo\compiler.js`

https://github.com/mochajs/mocha/issues/2310

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -295,9 +295,9 @@ if (program.retries) mocha.suite.retries(program.retries);
 
 var extensions = ['js'];
 program.compilers.forEach(function(c) {
-  var compiler = c.match(/([a-z.]+):(([A-Z]:)?\S*)+/)
-    , ext = compiler[1]
-    , mod = compiler[2];
+  var idx = c.indexOf(':')
+    , ext = c.slice(0, idx)
+    , mod = c.slice(idx + 1);
 
   if (mod[0] == '.') mod = join(process.cwd(), mod);
   require(mod);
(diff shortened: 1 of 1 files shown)


boneskull:
@silentcloud thanks!

bin/_mocha
```diff
@@ -295,9 +295,9 @@ if (program.retries) mocha.suite.retries(program.retries);
 
 var extensions = ['js'];
 program.compilers.forEach(function(c) {
-  var compiler = c.split(':')
-    , ext = compiler[0]
-    , mod = compiler[1];
+  var compiler = c.match(/([a-z.]+):(([A-Z]:)?\S*)+/)
```
boneskull: there are several problems with this regular expression--instead, try `c.split(':', 1)`.  does that work?

silentcloud:
```javascript
var str = 'js:C:\\Users\\fortest\\example\\compiler.js';
console.log(str.split(':', 1))
//['js']
```

boneskull: right, it still splits, just doesn't return it.  how about just:

```js
var idx = c.indexOf(':');
var ext = c.slice(0, idx);
var mod = c.slice(idx + 1);
```

silentcloud: "slice" is better, I will update this PR

ScottFreeCode:
Looks good to me.

silentcloud:
@boneskull   merge & publish a version ？

boneskull:
@silentcloud not sure when the release will be but I can merge this.  thanks again.

PR #2331, opened 2016-06-24
Updated glob library to latest, resolves minimatch security issue

cramhead:
I've updated the package.json file to reference an updated version of glob. All tests pass before and after the library upgrade.


boneskull:
@cramhead thanks, but see #2325

PR #2333, opened 2016-06-26
update dependency of jade

Thomas-P:
Update Jade to the actual version, but should be replaced with pug in
the next time.

- There are some security issues with [Link](https://snyk.io/org/thomas-p/test/github/mochajs/mocha)
- Jade is renamed to Pug [Link](https://github.com/pugjs/pug)

### DIFF
--- a/package.json
+++ b/package.json
@@ -322,7 +322,7 @@
     "escape-string-regexp": "1.0.2",
     "glob": "3.2.11",
     "growl": "1.9.2",
-    "jade": "0.26.3",
+    "jade": "1.11.0",
     "mkdirp": "0.5.1",
     "supports-color": "1.2.0",
     "to-iso-string": "0.0.2"


Thomas-P:
seems the build system hangs up

boneskull:
@Thomas-P Thanks for the PR.  Please search issues before sending, as this has already been covered.  It's not going to be possible to upgrade `jade` to `pug` until we're ready to do a major release--it will break Node.js v0.8 support.

Thomas-P:
Hi, I did not change `jade` into `pug`, just updatet the outdated version of jade to the actual version. I can't make the test, because i am not able to run the make file.

ScottFreeCode:
It's actually the version upgrade (which is required for the rename since the older versions weren't published under the Pug rename) that breaks compatibility with Node 0.8's default NPM version due to the caret in Jade's (or its dependencies') dependency versions.

(Incidentally, the first link in the original post here requires logging in to Snyk; if there's publicly available info we could link to on the security issues in question, it would be more convenient, especially for anyone without a GitHub account who isn't otherwise signed up with that service.)

PR #2336, opened 2016-06-28
ensures slow() called w/o args does not blast value; closes #2311

### LINKED ISSUE #2311: `this.slow()` doesn't work in describe callbacks like `this.timeout`
```js
describe('a', function() {
  this.timeout(5000);
  this.slow();

  it('1', function(cb) {
    setTimeout(cb, 3000);
  });

  it('2', function(cb) {
    this.slow(); // just for contrast with test 1
    setTimeout(cb, 3000);
  });
});
```

If I understand the docs correctly, I expect neither of the tests to timeout or have slow warnings because they're both covered by `this.timeout` and `this.slow` calls in the describe section. Neither of the tests timeout, but the first test (not covered by its own redundant `this.slow`) has a slow warning still:

```
  a
    ✓ 1 (3008ms)
    ✓ 2


  2 passing (6s)
```

mocha version 2.5.3

boneskull:
that's about the size of it

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -96,7 +96,7 @@ Runnable.prototype.timeout = function(ms) {
  * @return {Runnable|number} ms or Runnable instance.
  */
 Runnable.prototype.slow = function(ms) {
-  if (!arguments.length) {
+  if (typeof ms === 'undefined') {
     return this._slow;
   }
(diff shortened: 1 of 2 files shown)


boneskull:
this appears to be low-impact; gonna merge it.

PR #2350, opened 2016-07-03
Release: v3.0.0

boneskull:
OK, so this PR represents v3.0.0, which contains breaking changes.  Notably, Node.js v0.8 support is *dropped*.

This PR will probably undergo some rebasing before it's merged.

TODO:

1. [x] AppVeyor must pass
2. [x] Update `contributors` field of `package.json`
3. [x] Update `CHANGELOG`
4. [x] Update relevant site documentation 

Once 1. is working, I'll publish this *branch* as a pre-release version.

A rough estimation of notable changes:

- Add IE7 support
- Mocha will throw an exception if `done` is passed to a test *and* a `Promise` is returned:
  
  ```js
  it('should now throw an exception', function (done) {
    return new Promise(function (resolve) {
      resolve();
    })
      .then(done);
  });
  ```
- `glob` is updated to latest version; a production `npm install` of `mocha` should not fail an `nsp check` (#2318)
- `jade` is upgraded to latest `pug` (as of last night) (#2200)
- `it.only()` and `describe.only()` no longer leverage "grep" to do their work; these functions should now work like you think they should (#1481)
- Altered `dot` reporter output to be more visually distinctive.
- Mocha will now throw an error if a suite or test is untitled:
  
  ```js
  // this will now fail
  it(function () {
    assert(true);
  });
  ```
- The minimum version of Node.js required is now v0.10.x.  
- The minimum version of `npm` required is now v1.4.x.


boneskull:
Waiting for appveyor sucks.

hairyhenderson:
@boneskull - just curious about Node 0.10 support - it's currently in maintenance mode and support from the foundation ends completely in October of this year, so would it maybe be worth skipping straight to 4.x as minimum, or planning another major bump to explicitly drop support for older Node versions closer to October?

boneskull:
@hairyhenderson The only metrics I've seen around usage recently is the [Node.js 2016 User Survey Report](https://nodejs.org/static/documents/2016-survey-report.pdf).  Page 18 contains:

image

In my experience, when serving web apps, support for browsers would be dropped if their usage falls at or below 1%.  I'm not sure if that is too liberal or too conservative, and I'm not sure how well it translates here, but that's all I got.

Considering this graph, if we're going to require v4.0, we will drop support for nearly *a third* of potential Mocha users.  If we're going to require v0.12 even, that will drop ~13% of potential users.  I'm uncomfortable with both of these numbers.

But also: Mocha's codebase has little to gain from requiring v4.0.  It contains no ES6; IIRC not even polyfills.

We will drop support for 0.10 at some point, but we need *more data* to make these decisions when they aren't obvious.  I don't want to have to wait until the 2017 user survey if we don't have to!  @dasilvacontin has proposed a mechanism (if there's an issue for it, I don't know where it is) wherein users could opt-in to anonymous usage tracking.  [Yeoman](http://yeoman.io) has similar functionality.

boneskull:
See #2356 for potentially dropping [HTMLCov](http://mochajs.org/#htmlcov) and [JSONCov](http://mochajs.org/#jsoncov) reporters.

boneskull:
I've [updated the CHANGELOG](https://github.com/mochajs/mocha/blob/v3.0.0/CHANGELOG.md)

boneskull:
I've [updated the docs](https://github.com/mochajs/mochajs.github.io/pull/48) and will *likely* publish Thursday (7/28), barring any unexpected issues or concerns.

I'm still somewhat concerned about the `.only()` behavior.  If someone feels like testing it manually further, that'd be cool.

mikermcneil:
@boneskull thanks for the update!  I switched over a couple of our modules yesterday and haven't noticed any problems yet, so that's good :)  Although, a grain of salt:  I haven't used only() in quite a while (been using `--grep` and `.skip()` instead)

boneskull:
rebase and merged

hairyhenderson:
Thanks @boneskull, this is awesome! 🎉

dasilvacontin:
w00t

PR #2357, opened 2016-07-06
Exclude the --inspect flag

simov:
PR: https://github.com/nodejs/node/pull/6792
Merged in Node v7.0: https://nodejs.org/download/nightly/v7.0.0-nightly20160619a92089b6bd/
Short article about it: https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27#.cs12vrlsa
Usage:

```bash
mocha --inspect --debug-brk hello.js
```

A quick patch release will be greatly appreciated! Thanks!

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -26,6 +26,7 @@ process.argv.slice(2).forEach(function(arg){
     case 'debug':
     case '--debug':
     case '--debug-brk':
+    case '--inspect':
       args.unshift(arg);
       args.push('--no-timeouts');
       break;


boneskull:
Thanks.  We can add this in v3.1.0.  May need to rebase it onto `master` after the v3.0.0 branch is merged.

As a workaround until then, you can execute `node <flags> /path/to/_mocha -- <mocha-args>`

boneskull:
probably transient CI failures

SimenB:
Any chance for a release of this? 😄

fgarcia:
Thanks a lot!!! This feature was the best surprise I got from the latest release!

However I am curious about why `--debug-brk`is needed. Normally I would just type `debugger` in some test and wait for the debugger to reach that breakpoint. 

If I just use `--inspect` and a `debugger` line, Mocha will start the debug server and run all tests without stopping. Currently I must force it to start with a breakpoint, and manually continue to reach the second breakpoint in my code.

SimenB:
If you don't have `--debug-brk` you have no attached debugger, so `debugger` statements do nothing.

I _think_

fgarcia:
@SimenB is right, I've just tested a small node example and `--inspect` is useless without `--debug-brk` OR with a debugger attached before reaching the breakpoint. 

The current behaviour is the expected one

PR #2362, opened 2016-07-07
update `diff` package to latest version

DZittersteyn:
According to https://github.com/kpdecker/jsdiff/blob/v2.2.3/release-notes.md, the changes mostly include bugfixes, no API changes. `make test` doesn't complain.

(aside: it looks like http://mochajs.org/#testing-mocha is outdated since `test-all` no longer appears in the makefile)

### DIFF
--- a/package.json
+++ b/package.json
@@ -318,7 +318,7 @@
   "dependencies": {
     "commander": "2.3.0",
     "debug": "2.2.0",
-    "diff": "1.4.0",
+    "diff": "2.2.3",
     "escape-string-regexp": "1.0.2",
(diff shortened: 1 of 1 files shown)


boneskull:
Good catch on the developer docs.  But we can't upgrade `diff` atm; it will break compatibility with old IE due to [this bug](https://github.com/kpdecker/jsdiff/issues/124). 

This doesn't show in CI due to restrictions about running tests on SauceLabs in PRs.  Thanks anyway!

DZittersteyn:
Ah, nice catch, thanks for looking into it!

PR #2364, opened 2016-07-08
Add --preserve-symlinks flag for node

rosswarren:
This is a new flag added in Node 6.2.0 https://github.com/nodejs/node/pull/6537

I would like to be able to use it with mocha!


boneskull:
@rosswarren thanks!

PR #2366, opened 2016-07-10
Trying to avoid some warnings.

JJ:
Because we all hate warnings when testing, right?


boneskull:
@JJ thanks but please search the issues before sending a PR

PR #2371, opened 2016-07-12
grep: skip adding connecting space for empty suite titles. Fixes #2360.

### LINKED ISSUE #2360: Q: CLI grep regex behavior?
``` js
it('hey', () => {})
```

``` bash
$ mocha test.js 
# _mocha /(?:)/


  ✓ hey

  1 passing (6ms)

$ mocha test.js -g hey
# _mocha /hey/


  ✓ hey

  1 passing (5ms)

$ mocha test.js -g ^hey$
# _mocha /^hey$/


  0 passing (0ms)
```

Is this the expected behavior?

gurdiga:
Fixes #2360.

@mochajs/core Please take a look.


mocha.js
```diff
@@ -776,7 +776,7 @@ module.exports = function(suite) {
   });
```
gurdiga: @mochajs/core I’m not sure what to do about these changes here. Is it all right to include them in the commit? 😳

gurdiga:
@mochajs/core Hey guys, the AppVeyor seems to fail it can’t download phantomjs. What is the appropriate way to handle this? Is there a way to re-start the build? 🤓

bin/_mocha
```diff
@@ -68,8 +68,8 @@ program
   .option('-S, --sort', "sort test files")
   .option('-b, --bail', "bail after first test failure")
   .option('-d, --debug', "enable node's debugger, synonym for node --debug")
-  .option('-g, --grep <pattern>', 'only run tests matching <pattern>')
-  .option('-f, --fgrep <string>', 'only run tests containing <string>')
+  .option('-g, --grep <pattern>', 'only run tests having full title matching <pattern>')
+  .option('-f, --fgrep <string>', 'only run tests having full title containing <string>')
```
simov: I don't think this is correct. If you want exact matching you have to use `^full title$'. Also this PR doesn't change the way the matching works.

mocha.js
simov: I'm not familiar with Mocha's development, but I think this shouldn't be here.

simov:
@gurdiga can I suggest a slight update to this PR's title:

```diff
- grep: handle empty suite titles. Fixes #2360.
+ grep: skip adding connecting space for empty suite titles. Fixes #2360.
```

mocha.js
gurdiga: Hey @simovm! Thank you for checking this out! 🤓

I agree that this raises a question mark, and this is why I’m bringing this up. I’m not familiar with the process either — this is my first PR — so I guess in this case it’s appropriate to ask those who know. 🤔

simov: If you take a look at the commit log you will see that Mocha is being build before release by the maintainer who releases it.

bin/_mocha
gurdiga: > If you want exact matching you have to use `^full title$'.

Yep! That’s exactly what I’m trying to make a bit clearer with this wording change: the pattern given to `--grep` regards the full test title, and this is especially valuable to understand in cases like the one you brought up in #2360.

> Also this PR doesn't change the way the matching works.

Exactly! It only slightly changes how the full test title is computed, which means it changes the string that in the end is matched against the given pattern; the matching doesn’t need to change as far as I understand. 🤓

gurdiga:
Yeah, I guess that gives a bit more detail regarding on what exactly changes, which I guess this may be a good idea. 🤔 Thanks! 🙂

gurdiga:
(I have force-pushed same changes just to see if it’d make the build pass this time.)

bin/_mocha
simov: Ok, I'm getting your point now, but as you can see I didn't understand your intention at first, so I don't think the wording is the correct one. The way I understand it is that the pattern should match the full title, which obviously is not the case.

mocha.js
gurdiga: Wow — good catch! I’ve checked `git log` and it looks like `mocha.js` is rebuild quite frequently just before the release. That’s good, I can drop those changes than and we’ll get a simpler change! 🤓

bin/_mocha
gurdiga: I agree that it’s tricky — as I mentioned on the issue, I was quite surprised myself to have found how `--grep` actually works. But thinking about it a bit more, I realized that this is actually a good choice. 👷

If you have a clearer way to word this, I’m open to discuss it. 🤓

PR #2373, opened 2016-07-13
Remove stdout: prefix

### LINKED ISSUE #2095: Output in Chrome console includes "stdout:"-prefixed lines
Mocha 2.4.5
Chrome 48.0.2564.103

Pretty much with any reporter I use, Mocha outputs the following poorly-formatted text to the browser console.

Note the weird "stdout:"-prefixed lines.

```
  Foo
     #bar
 stdout: [2K
       1) should work
       ✓ should still work
   1 passing (770ms)
   1 failing
   1) Foo #bar should work:
     Error: oops
      at Context.shouldWork (test/test.js!transpiled:11:35)

 stdout: [0G
 stdout: [2K
 stdout: [0G
```

This seem to be the result of code in most reporters using a combination of calls to console.log() and process.stdout.write().  In the browser, process.stdout is ultimately implemented as calls to the console with "stdout:" prefixed in front on each line (not really sure why, just something BrowserStdout does) which is often just a cursor movement escape code (which, in many cases, like above, really contributes nothing to the final formatting).

So, this just doesn't work well, eh?  Can something be done to make reporter output look better in the browser?  Mocha is advertised as working "in the browser"   ...but it doesn't really.

skeggse:
Fixes #2095, by disabling the `stdout:` prefix with the browser shim.

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -2,7 +2,7 @@
  * Shim process.stdout.
  */
 
-process.stdout = require('browser-stdout')();
+process.stdout = require('browser-stdout')({level: false});
 
 var Mocha = require('./lib/mocha');
 


boneskull:
see dicsussion in #2095.  if there's a better alternative than `browser-stdout`, then I'm open to suggestions

boneskull:
@skeggse I realize this was sent over a year ago, but was wondering if you could sign the CLA?

skeggse:
Happy to sign the CLA. Unfortunately, it appears to be incomplete. It references a document called `[JS Foundation IP Policy]({{ site.url }}{{ site.baseurl }}/pdf/ip-policy.pdf`. Please provide a copy of this document, or fix the Licenses section in the CLA itself.

boneskull:
@skeggse you want [this URL](https://cla.js.foundation/mochajs/mocha?pullRequest=2373) which is linked in the build section above.  Where did you find that..?

skeggse:
image

boneskull:
cc @kborchers

boneskull:
@skeggse FWIW I [found the doc](https://js.foundation/wp-content/uploads/sites/33/2017/03/JS-Foundation-IP-Policy.pdf) if you want to get it moving.  :wink:

skeggse:
@boneskull thank you. I've signed the CLA.

kborchers:
@skeggse @boneskull Thanks! We are aware of the broken link and will fix ASAP. The main blocker is that currently, CLA Assistant treats any change to the CLA as a change that invalidates every previous signature. In general, that would be correct behavior to make sure anyone that has signed is made aware of a change and has to sign again to agree to the change. In this case, it's the equivalent of a typo but the application doesn't know the difference so we are trying to formulate a procedure to make the change and migrate all of the signatures over to the "new" version with minimal (preferably no) impact to the projects.

boneskull:
@skeggse thanks!

PR #2387, opened 2016-07-21
let child suites run if parent is exclusive; closes #2378

### LINKED ISSUE #2378: Nested describes don't run if parent describe has .only (mocha 3)
Previously, if you had a nested describe and tests inside a parent describe with .only, the child-describe blocks would also run.

It seems https://github.com/mochajs/mocha/issues/1481 may have broken this behavior. While I think the actual fix in 1481 makes sense, this side-effect of it seems counter-intuitive, and at least goes against my expectations of how this should work. It's also different from how this works in some other testing tools I've used in the past with similar features.

## Steps to reproduce

1. Create a test file with a root describe using .only.
2. Put another describe inside it, with an it inside
3. Run test

## Expected result

The tests inside the nested describe run

## Actual result

Only tests immediately inside the parent describe block run, at least with mocha@3.0.0-1

## Sample code

```javascript
describe.only('this should run', function() {
  it('an it should also run', function() { });

  describe('this nested describe should run', function() {
    it('and this nested it should run as well', function() { });
  });
});
```

Output:
```
  this should run
    ✓ an it should also run


  1 passing (14ms)
```

Expected output:
```

  this should run
    ✓ an it should also run
    this nested describe should run
      ✓ and this nested it should run as well


  2 passing (17ms)
```

boneskull:
- includes some more refactoring of interface code into `lib/interfaces/common.js` for DRY (still more work to be done here)
- unfortunately said refactoring contains hellish logic which addresses this issue (someone should make a flowchart of this for lolz)
- original PR did not address `exports` interface; this doesn't either
- may need more coverage against `qunit` interface

*This is a PR into the `v3.0.0` branch*

cc @mochajs/core

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -1,13 +1,16 @@
 'use strict';
 
+var Suite = require('../suite');
+
 /**
  * Functions common to more than one interface.
  *
  * @param {Suite[]} suites
  * @param {Context} context
+ * @param {Mocha} mocha
  * @return {Object} An object containing common functions.
  */
-module.exports = function(suites, context) {
+module.exports = function(suites, context, mocha) {
   return {
     /**
      * This is only present if flag --delay is passed into Mocha. It triggers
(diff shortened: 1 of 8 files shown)


lib/interfaces/common.js
```diff
@@ -64,15 +67,60 @@ module.exports = function(suites, context) {
...
+       * @param {string} [opts.file] Filepath where this Suite resides
+       * @param {boolean} [opts.isOnly] Is Suite exclusive?
+       * @returns {Suite}
+       */
+      create: function create(opts) {
+        var suite = Suite.create(suites[0], opts.title);
+        suite.pending = Boolean(opts.pending);
+        suite.file = opts.file;
+        suites.unshift(suite);
+        // I should be pilloried for the following.
```
boneskull: The general idea is that we don't flip a bit on a `Suite` if something like `describe.only()` is used; instead we just reference its `tests` prop to `onlyTests`, so any tests in it (and within child `Suite`s!) will be in the bucket.  

If a `Test` would override this behavior, we [destroy the reference](https://github.com/mochajs/mocha/pull/2387/files#diff-b756b5d38e193a4ff9b810229dbc9b60R139).

boneskull:
I'm *extremely bothered* by the fact that this PR #2387 is a fix for PR #2378.

boneskull:
retrying travis; seems to be transient sauce failure

boneskull:
Guess I'll just merge this and publish another beta.

PR #2400, opened 2016-07-28
Remove carriage return before each test line in spec reporter

Munter:
The spec reporter would output a carriage return char at the beginning of each line in non-TTY's and delete and overwrite the current line in TTY's.

Since this cursor call was only ever used at the beginning of a line it serves no purpose.

See discussion starting at https://gitter.im/mochajs/mocha?at=579a7573ac80b5ea3f149512

### DIFF
--- a/lib/reporters/spec.js
+++ b/lib/reporters/spec.js
@@ -5,7 +5,6 @@
 var Base = require('./base');
 var inherits = require('../utils').inherits;
 var color = Base.color;
-var cursor = Base.cursor;
 
 /**
  * Expose `Spec`.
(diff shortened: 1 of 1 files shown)


boneskull:
@Munter ty.  can you take a screenshot of before/after output in a console?  also would be helpful to see the before/after output when redirecting to a file

boneskull:
also, whatever [mocha-phantomjs](https://www.npmjs.com/package/mocha-phantomjs) outputs should be checked

Munter:
https://github.com/mochajs/mocha/pull/2401 replaces this PR

boneskull:
(This PR is closed because we get better CI coverage if a committer can just use a branch in `mochajs/mocha`... )

PR #2402, opened 2016-07-29
don't use location.pathname in HTML reporter; closes #2285

### LINKED ISSUE #2285: window.location.pathname in links in browser HTML reporter
A link starting with `window.location.pathname` (which is used in the grep links in the HTML reporter) is treated oddly in some browsers, at least when the page is run without a server (straight from the filesystem). For example, Internet Explorer on Windows gets "/C:/path/to/page.html" for `window.location.pathname`, which causes it to try to go to "C:/C:/path/to/page.html?grep=..." when the grep links are clicked. (I am not sure if this issue affects any other browsers. I'm also not sure whether this is strictly a problem for browsing the page straight off the filesystem, although I wouldn't be surprised if it is.) We should find out if simply removing pathname and using just "?grep=..." for the links works in all browsers, or if there's any other fix.

Looks like test/browser/ui.html tests this, but I don't know whether we can get all the same behavior in automated testing for it or not (or even whether the approach used by test/browser/ui.html works for automated testing).

gurdiga:
Related to https://github.com/mochajs/mocha/issues/2285.

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -223,7 +223,7 @@ function makeUrl(s) {
     search = search.replace(/[?&]grep=[^&\s]*/g, '').replace(/^&/, '?');
   }
 
-  return window.location.pathname + (search ? search + '&' : '?') + 'grep=' + encodeURIComponent(escapeRe(s));
+  return (search ? search + '&' : '?') + 'grep=' + encodeURIComponent(escapeRe(s));
 }
 
 /**


gurdiga:
@mochajs/core Please take a look; but be tender, I’m trying to get a hang of what’s the right way to do things here. 😊

boneskull:
@gurdiga Thanks.  I'm not sure if we have automated tests around this.  can you run some manual ones (thinking chrome and ie7/ie8) and post screenshots?  If you don't have these browsers handy, I can get you set up with SauceLabs .

gurdiga:
SouceLabs sounds good, thank you! — I couldn’t find a way to get IE7 now. 🤓

boneskull:
@gurdiga Were you able to get the screenshots?

PR #2411, opened 2016-08-03
fix nested describe.only suites; closes #2406

### LINKED ISSUE #2406: .only() no longer works with nested describe()s in Mocha 3.0.0
I was excited to see improvements to `.only()` in the Mocha 3.0.0 changelog. It's mostly working well, but I found a regression. Consider the following test file with nested `describe()`s:

```

  describe('.push()', function() {

    it('appends a value to the end of the array', function() {
      var array = [1];
      array.push(2);
      assert.deepEqual(array, [1, 2]);
    });

    it('can append more than one value to the end of the array', function() {
      var array = [1];
      array.push(2, 3);
      assert.deepEqual(array, [1, 2, 3]);
    });
  });

  describe('.reverse()', function() {

    it('reverses a non-empty array', function() {
      assert.deepEqual([1, 2, 3].reverse(), [3, 2, 1]);
    });

    it('does nothing to an empty array', function() {
      assert.deepEqual([].reverse(), []);
    });
  });
});
```

In previous versions of mocha, you could add a `.only()` to one of the inner `describe()`s and it would do what you expect. For example, if I changed that file to look like:
```
  describe.only('.push()', function() {
```
It should run 2 tests. That's the way previous versions of Mocha behaved. It's a very useful behavior that my team uses every day.

In Mocha 3.0.0 it runs 0 tests. 

A few other notes:
- `.only()` on the top-most `describe()` in a file works fine
- changing `describe()` to `context()` doesn't affect the behavior (same issue with `context()`)

not-an-aardvark:
Previously, tests in nested `describe.only` suites would not be registered as `only` tests in their parent suite (due to the implementation described in https://github.com/mochajs/mocha/pull/2387#discussion_r71646306). This caused mocha to think that the parent suite had no `only` tests, so the suite was skipped entirely. (See #2406 for more information.)

This PR moves some of the logic for `it.only` and `describe.only` to the test runner, rather than the place where suites are created. This makes the logic slightly simpler, because all of the suites already exist at the time that `only` suites are handled.

---

The correct logic for `only` blocks (from what I can tell, looking at the existing tests and the docs) is as follows:

### To run a suite:

* If a suite contains no `only` suites/tests among all of its descendants, run all of its child suites and tests.
* Otherwise, if a suite contains any `only` tests among its children, run the `only` tests for that suite, and skip all the child suites.
* Otherwise, run all the child `only` suites. Also run any child suites that have `only` suites/tests as descendants.

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -106,15 +106,9 @@ module.exports = function(suites, context, mocha) {
         suite.pending = Boolean(opts.pending);
         suite.file = opts.file;
         suites.unshift(suite);
-        // I should be pilloried for the following.
         if (opts.isOnly) {
-          if (suite.parent && suite.parent.onlyTests) {
-            suite.onlyTests = suite.parent.onlyTests === suite.parent.tests ? suite.tests : [];
-          } else {
-            suite.onlyTests = suite.tests;
-          }
-        } else {
-          suite.onlyTests = suite.parent && suite.parent.onlyTests === suite.parent.tests ? suite.tests : [];
+          suite.parent._onlySuites = suite.parent._onlySuites.concat(suite);
+          mocha.options.hasOnly = true;
         }
         if (typeof opts.fn === 'function') {
           opts.fn.call(suite);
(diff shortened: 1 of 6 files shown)


boneskull:
Thanks for this.  I'm re-running the appveyor build; it may have been timing flake.

boneskull:
At first glance, this looks good.  If it passes CI, I'll see if I can't release it as v3.0.1 tonight or tomorrow.

boneskull:
@not-an-aardvark Is there anything I should know about with the signed commit?  I haven't worked  with these before, and don't know the implications (if any).

not-an-aardvark:
I don't think there are any implications to worry about, it basically just means you can be sure that the person with the not-an-aardvark GitHub account is the same person who made the commit. (This information would be more useful if another user had a branch with some of my commits on them, for example.)

boneskull:
@not-an-aardvark tyvm!

PR #2419, opened 2016-08-05
fix .only suites nested within each other

### LINKED ISSUE #2417: infinite loop with nested .only()
Run the following test through mocha 3.0.1:
```

  describe.only('.push()', function() {

    it.only('appends a value to the end of the array', function() {
      var array = [1];
      array.push(2);
      assert.deepEqual(array, [1, 2]);
    });
  });
});

```

Mocha exits with an exception that seems to indicate an infinite loop occurred:
```
RangeError: Maximum call stack size exceeded
    at filterOnly (/Users/adam/workspace/cirrus/node_modules/mocha/lib/runner.js:857:23)
    at /Users/adam/workspace/cirrus/node_modules/mocha/lib/runner.js:861:9
    at Array.forEach (native)
    at filterOnly (/Users/adam/workspace/cirrus/node_modules/mocha/lib/runner.js:857:23)
    at /Users/adam/workspace/cirrus/node_modules/mocha/lib/runner.js:861:9
    at Array.forEach (native)
    at filterOnly (/Users/adam/workspace/cirrus/node_modules/mocha/lib/runner.js:857:23)
    ...
```

This is a simplified example. In practice, I sometimes focus on getting a single `it.only()` test passing, and then go "up a level" and test a group of related tests via a `describe.only()` block.

Not a blocker. Workaround: don't use nested `.only()`

not-an-aardvark:
fixes #2417

The issue was that instead calling itself with child suites, the function for filtering a `.only` suite was calling itself with the same suite. This resulted in infinite recursion.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -855,10 +855,10 @@ function filterOnly(suite) {
     // Otherwise, do not run any of the tests in this suite.
     suite.tests = [];
     suite._onlySuites.forEach(function(onlySuite) {
-      // If there are other `only` tests/suites nested in the current `only` suite, then filter the current suite.
+      // If there are other `only` tests/suites nested in the current `only` suite, then filter that `only` suite.
       // Otherwise, all of the tests on this `only` suite should be run, so don't filter it.
       if (hasOnly(onlySuite)) {
-        filterOnly(suite);
+        filterOnly(onlySuite);
       }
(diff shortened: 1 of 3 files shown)


boneskull:
awesome, thanks again @not-an-aardvark

PR #2435, opened 2016-08-13
Fix required test files caching, closes #736

### LINKED ISSUE #736: Running the test suite twice skips tests that were already required
I wanted to run the test suite twice against different browsers on Saucelabs. I'm using the technique to run mocha programatically but on the second run mocha would always report that it did not run any tests.

After a lot of digging around it seems that since mocha uses require() to load the test files, they get cached and if they're cached they're not added to the suite.

I can manually clear all the tests from the require cache, but I'm wondering if maybe mocha shouldn't do this by default?

This is a simplified example of what I end up doing for it to work:

``` javascript
var Mocha = require('mocha');
var path  = require('path');

var testFile = path.resolve(__dirname, './test_file.js');

var set1 = new Mocha({ui: 'tdd'});
set1.addFile(testFile);

var set2 = new Mocha({ui: 'tdd'});
set2.addFile(testFile);

set1.run(function(failures1){
  console.log(require.cache['/Users/patrick/Downloads/mocha_argh/test_file.js']);
  // Need to delete the test from the cache or else
  // the test won't run.
  delete require.cache['/Users/patrick/Downloads/mocha_argh/test_file.js'];
  set2.run(function(failures2){
    process.exit(failures1 && failures2);
  });
});
```

1999:


1999:
@boneskull it seems like appveyor test fail is not because of this PR. Could you take a look please?

1999:
@travisjeffery @boneskull @danielstjules is there anybody going to listen to my sto-o-o-ry all about the pull request to merge :)

boneskull:
@1999 thanks for the PR

lib/mocha.js
```diff
@@ -247,6 +253,28 @@ Mocha.prototype._growl = function(runner, reporter) {
 };
 
 /**
+ * Get require.cache files (node only)
```
boneskull: a problem is that this code runs regardless of whether or not we're in node.

lib/mocha.js
```diff
@@ -247,6 +253,28 @@ Mocha.prototype._growl = function(runner, reporter) {
...
+ *
+ * @api private
+ */
+Mocha.prototype._getRequireCache = function() {
+  return Object.keys(require.cache || {});
```
boneskull: ...and because of the above, this will break ES3.  the "jsapi" tests don't get run in the browser.  should they?  not sure.

lib/mocha.js
boneskull: see line 228 for suggestion

lib/mocha.js
```diff
@@ -220,6 +224,8 @@ Mocha.prototype.loadFiles = function(fn) {
     suite.emit('require', require(file), file, self);
     suite.emit('post-require', global, file, self);
   });
+
+  this._flushRequireCache();
```
boneskull: only call this if `process.browser` is falsy

boneskull:
This should be hidden behind an option.  See [here](https://github.com/mochajs/mocha/blob/master/lib/mocha.js#L321) for what this should look like.

Otherwise it's a breaking change.  I can produce an example if you are unconvinced.

In the spirit of DRY, the "purge" function should be called by `bin/_mocha` [here](https://github.com/mochajs/mocha/blob/master/bin/_mocha#L372) as well.  

Looking further at the "jsapi" tests, they are inappropriate for the browser as written; don't worry about trying to get the browser to run them.

lib/mocha.js
1999: Hm.. but I thought that `_getRequireCache` will return either cached files (node) or an empty array and that would be okay, won't it? Should I instead check for `process.browser` and don't run this function at all?

1999:
Could you help me a bit?

> This should be hidden behind an option

This issue doesn't seem to be a breaking change, should it still be hidden behind a flag? What would be the meaning of this flag?

> In the spirit of DRY, the "purge" function should be called by bin/_mocha here as well.

Ah, you mean that I should move `_getRequireCache` and `_flushRequireCache` to separate file and call it inside bin/_mocha instead of `purge` function?

1999:
cc @boneskull

lib/mocha.js
```diff
@@ -247,6 +258,28 @@ Mocha.prototype._growl = function(runner, reporter) {
...
+};
+
+/**
+ * Flush test files from require.cache
+ *
+ * @api private
+ */
+Mocha.prototype._flushRequireCache = function() {
+  this._getRequireCache().forEach(function(filePath) {
+    if (this._requireCache.indexOf(filePath) === -1) {
```
boneskull: Since `require()` is imperative, we don't have a guarantee here that the things we're removing from the cache were actually loaded by Mocha--just that they appeared in the cache between *a* and *b*. 

We also don't have a guarantee that everything being removed from the cache is everything that *should* be removed, since your test file could execute a `require()` in async code (use of `--delay` is a more likely scenario).

We could inspect the module's `parent` to see that it was required by `lib/mocha.js`, but we're still going to be unsure about whether module `x` loaded by test `y` should be removed.

Mocha tests are written with the understanding (assumption, more likely) that Mocha doesn't molest `require.cache`.  This change would break assumptions.  I'm also not convinced this should be the default behavior.

Since this is intended to be used programmatically, an option should suffice (a command-line flag probably isn't necessary at this point).

boneskull:
Will add back to v3.1.0 if it's ready in time.

lib/mocha.js
1999: Thank you for such a full explanation, I finally got your point.
Just added dropRequireCacheBeforeRun() method for this and updated tests.

1999:
cc @boneskull have you seen the changes?

PR #2440, opened 2016-08-14
Update TAP reporter. fix #2410

### LINKED ISSUE #2410: Improve tap reporter
```
1..1
ok 1 describe it
# tests 1
# pass 1
# fail 0
```

for

```js
describe('describe', () => {
  it('it', () => {
    expect(true).to.eql(true)
  })
})
```

```
    ✔ Description


  ✖ No tests found
```

```
1..1
# describe
ok 1 it
# tests 1
# pass 1
# fail 0
```
prints
```

  describe

    ✔ it


  total:     1
  passing:   1
  duration:  8ms
```

apkiernan:
Updates the TAP reporter to print the `describe` blocks as diagnostic messages to allow for proper TAP output.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -27,7 +27,11 @@ function TAP (runner) {
 
   runner.on('start', function () {
     var total = runner.grepTotal(runner.suite);
-    console.log('%d..%d', 1, total);
+    console.log('1..' + total);
+  });
+
+  runner.on('test', function (test) {
+    console.log('# ' + title(test.parent.fullTitle()));
   });
 
   runner.on('test end', function () {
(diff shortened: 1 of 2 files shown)


apkiernan:
Each failure appears to be an issue with PhantomJS not installing correctly. Not sure how to remedy this

Munter:
Did the behavior of the reporter also change from a streaming output to a buffer that is only flushed at the end?

boneskull:
Yeah, it's not clear to me why we're buffering things now... @apkiernan ?

apkiernan:
Apologies for the delay in responding, started a new job recently.

Looking back at this I think I had originally made the buffering change for testing purposes. I noticed the reassignment of console.log in another test and made the change to these tests as well so the buffering isn't needed. This has been changed back to streaming the output as it was originally.

Munter:
Looks good to me.

I'm a bit confused about not seeing any tap reporter test output in the CI log. I'll see if I can have a look into that. Wouldn't want the tests to somehow not be run, now that we have them :)

PR #2441, opened 2016-08-14
Only check if stdout is a tty.

paparomeo:
Mocha only writes to `stdout`, and users may want to redirect `stderr`. Mocha ends up with unnecessarily constrained output in this case, since while `stderr` will be a pipe `stdout` will still be a tty. This PR changes the code to only check if `stdout` is a tty instead of checking both `stdout` and `stderr`.

In  the `master` branch try in bash:

```
bin/mocha 2> >(cat)
```

Mocha assumes `stdout` is a pipe and only uses the default 75 character screen width. 

If running the same command in this branch, Mocha will correctly detect the screen width.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -28,10 +28,10 @@ var clearInterval = global.clearInterval;
 /* eslint-enable no-unused-vars, no-native-reassign */
 
 /**
- * Check if both stdio streams are associated with a tty.
+ * Check if stdout streams is associated with a tty.
  */
 
-var isatty = tty.isatty(1) && tty.isatty(2);
+var isatty = tty.isatty(1);
 
 /**
  * Enable coloring by default, except in the browser interface.


boneskull:
Thanks!  @paparomeo I'm not sure what bug(s) this fixes.  Can you show how this fix can be used, and what happens before the fix?

paparomeo:
@boneskull: in this [screencast](https://asciinema.org/a/e24bmahty1pmvaozd08d1mh38) you can see how in master, when redirecting `stderr` mocha *does not* make use of the whole width of the terminal, only using the default 80 columns for non-ttys. Then it switches to my branch and you see it using the whole width of the terminal when redirecting `stderr`. Hopefully this will demonstrate the issue I describe above. Please let me know if not and if you need further clarification.

paparomeo:
@boneskull can this be merged? Are you happy with my explanation above?

ScottFreeCode:
I can't speak for @ boneskull re. the explanation, but for my two cents' worth:

It may be helpful to have a test for this, which would
1. demonstrate what the change fixes in terms of (preferably realistic or at least non-contrived) example usage,
2. in the codebase for posterity (no need to look up the discussion/recording to see what difference the new logic makes), and
3. ensure that we don't accidentally change it back without realizing it (regression)

PR #2442, opened 2016-08-15
Change in some text color for better readability.

vajahath:
The previously used dark grey reduces readability of some text. So changed it to a better color.


boneskull:
Thanks!  We're not going to change the colors for now (see #802, #1200, etc, etc).

Issue #2383, opened 2016-07-20
consistently name (and rename) test files


boneskull:
We should rename test files in `test/**/*.js` to `test/**/*.spec.js`.  This would help with quickly opening files in an IDE or text editor.  I often open the wrong `runnable.js`:

![image](https://cloud.githubusercontent.com/assets/924465/16971513/85350eb4-4df2-11e6-83cd-53e5ee404116.png)

Renaming the test files will eliminate potential confusion here and improve DX.  

However, not everything in the glob `test/**/*.js` is a test.

Fixtures (e.g. `test/integration/fixtures/**/*.js` and others) should *not* be renamed to `*.spec.js`.  Ideally we could settle on a convention (`*.fixture.js`?) and eliminate the weirdo filenames with umpteen extensions; `multiple.done.before.js` could be `multiple-done-before.fixture.js`, for example.  

Mostly I'm looking for consistency in recognizing that I'm about to open a source file, a test file, a fixture, or whatever else.

PR #2443, opened 2016-08-15
fix test files not using .spec suffix

makepanic:
PR for #2383 
- changes all test files from `test/**/*.js` to `test/**/*.spec.js`
- changes filenames with dot to use `-` instead (`http.meta.2.js` -> `http-meta-2.spec.js`)

Open questions:
- Should i also go ahead and change the fixtures filename to use `.fixture.js`?
- Should https://github.com/mochajs/mocha/blob/master/test/compiler/foo.js also get a `.spec` suffix?
- Is it ok that the commit includes the generated `mocha.js` file?


makepanic:
I rebased onto current master and renamed the fixture files. The generated mocha file was also removed from the commit.

boneskull:
@makepanic Missed a few fixtures, but I've added those and merged this.  Thanks again; killer work!

PR #2445, opened 2016-08-15
Exit process with correct error codes

### LINKED ISSUE #2438: Ambiguous cli exit status
Mocha exits with an exit status equal to the amount of errors in the test run. This is a potentially huge problem causing ambiguous exit codes.

Exit codes have a range from 0-255. An exit value greater than 255 returns an exit code modulo 256.

This means a test with 256 errors will exit with code 0. Run this example as a proof:

```js
for (var i = 0; i < 256; i += 1) {
    it('test ' + i, function () {
        throw new Error('waat');
    });
}
```

then `echo $?`

Further more, certain exit codes have special meaning which can cause tools further down a unix pipe to misbehave. See http://tldp.org/LDP/abs/html/exitcodes.html

Mocha cli should always exit with code 1 if there are tests failing and code 0 if there are no tests failing

Munter:
- Exit with code 130 on SIGINT
- ~~Exit with code 1 if any error occurred~~
- Exit with an exit code of max 255 to avoid 8-bit overflow mistakingly reporting no errors

Closes #2438

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -442,7 +442,7 @@ if (program.watch) {
 
 function exitLater(code) {
   process.on('exit', function() {
-    process.exit(code);
+    process.exit(Math.min(code, 255));
   });
 }
 
@@ -452,7 +452,7 @@ function exit(code) {
   // https://github.com/visionmedia/mocha/issues/333 has a good discussion
   function done() {
     if (!(draining--)) {
-      process.exit(code);
+      process.exit(Math.min(code, 255));
     }
   }
 
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -384,7 +384,7 @@ if (program.watch) {
   process.on('SIGINT', function() {
     showCursor();
     console.log('\n');
-    process.exit();
+    process.exit(130);
   });
 
   var watchFiles = utils.files(cwd, [ 'js' ].concat(program.watchExtensions));
@@ -470,6 +470,11 @@ function exit(code) {
 
 process.on('SIGINT', function() {
   runner.abort();
+
+  // This is a hack:
+  // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+  // The amount of failures will be emitted as error code later
+  runner.failures = 130;
 });
 
 /**


1999:
Many developers are waiting for failed number of tests in exit code of mocha, not just for 0, 1 and 130. Is this a really necessary change?

RobertWHurst:
@1999 That may be true, but it's a pretty odd way to get the error count. Exit codes aren't really meant to be used that way.

Munter:
Whether the full scope of the change is necessary is up for debate.  Certainly the false pass has to be fixed.

> Many developers are waiting for failed number of tests in exit code of mocha

Do you have any examples in the wild or any known numbers or is this just a case of ad-hoc statistics?

The current state of Mocha is at odds with proper unix exit code semantics. You could also argue that these many people have been doing it wrong all the time and misinterpreting signals like "exit caused by SIGINT", "exit caused by invalid command", "file not found" etc as indicating the number of test failures, while the actual error condition was a completely different one

1999:
Well in my personal opinion this is a necessary change. I noticed that our QA guys are looking at the process exit code as the `failuresNum` but I usually ask them not to do that.

boneskull:
I don't see this as a necessary breaking change unless someone can document that their workflow is broken because of it.  I agree it's "incorrect" and should probably be fixed at some point, but we need to be actually helping people when we break the API--not just in theory.

However, the exit code should be capped at 255 or whatever so as to not exit with a false positive 0 if there happen to be 256 failing tests.

@Munter Can you modify the PR?

Munter:
@boneskull Updated

Munter:
All green after restarting all the tests that failed because of broken phantomjs install

bin/_mocha
```diff
@@ -428,7 +428,10 @@ function exit(code) {
   done();
 }
 
-process.on('SIGINT', function() { runner.abort(); })
+process.on('SIGINT', function() {
+  runner.abort();
+  process.exit(130);
```
boneskull: This is going to cause problems, because `runner.abort()` is attempting to exit cleanly.

Instead, you should be able to override `runner.failures`; [see here](https://github.com/boneskull/mocha/blob/5e1cd44c1f4e7301b6368209abf986bd6becca92/lib/runner.js#L817-L817).

boneskull:
Please ignore the "LGTM" comment; I don't see how to remove it.  But line 433 will need fixing..after that this is OK (after conflicts are resolved).  Thanks!

Munter:
@boneskull Requested changes are implemented, branch is rebased and all test are green

73rhodes:
For anyone following this issue: mocha 11.3.0 now supports a --posix-exit-codes command line option that will use a non-zero exit code for fatal signals on unix systems (https://github.com/mochajs/mocha/pull/4989).

PR #2450, opened 2016-08-21
Report non-match to STDERR and exit if no files are matched

### LINKED ISSUE #2194: Feature request: better error reporting when mocha can't find tests (getting started)
The getting started tutorial on the home page looks pretty simple, but if you put the test in a different directory than "test" it's suddenly not very friendly:

/usr/local/lib/node_modules/mocha/lib/utils.js:628
        throw new Error("cannot resolve path (or pattern) '" + path + "'");
        ^

Error: cannot resolve path (or pattern) 'test'
    at Object.lookupFiles (/usr/local/lib/node_modules/mocha/lib/utils.js:628:15)
    at /usr/local/lib/node_modules/mocha/bin/_mocha:326:30
    at Array.forEach (native)
    at Object.<anonymous> (/usr/local/lib/node_modules/mocha/bin/_mocha:325:6)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Function.Module.runMain (module.js:447:10)
    at startup (node.js:142:18)
    at node.js:939:3

Munter:
This displays a warning on STDERR for each glob pattern that is passed in as a binary execution argument if the glob pattern results in no files matched.

If no files are matched across all the passed glob patterns it exits with error code 1

This allows for running tests for any glob pattern that correctly matches files, while still making the user aware of their other invalid glob patterns. Previously we would throw if any pattern did not match any files and the error message was not the best one to figure out the problem.

Fixes #2194

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -358,8 +358,24 @@ if (!args.length) {
 }
 
 args.forEach(function(arg) {
-  files = files.concat(utils.lookupFiles(arg, extensions, program.recursive));
+  var newFiles;
+  try {
+    newFiles = utils.lookupFiles(arg, extensions, program.recursive);
+  } catch (err) {
+    if (err.message.indexOf('cannot resolve path') === 0) {
+      console.error('Warning: Could not find any test files matching pattern: ' + arg);
+      return;
+    } else {
+      throw err;
+    }
+  }
+
+  files = files.concat(newFiles);
 });
+
+if (files.length === 0) {
+  process.exit(1);
+}
 
 // resolve
 
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -365,15 +365,16 @@ args.forEach(function(arg) {
     if (err.message.indexOf('cannot resolve path') === 0) {
       console.error('Warning: Could not find any test files matching pattern: ' + arg);
       return;
-    } else {
-      throw err;
     }
+
+    throw err;
   }
 
   files = files.concat(newFiles);
 });
 
-if (files.length === 0) {
+if (!files.length) {
+  console.error('No test files found');
   process.exit(1);
 }
 
--- a/test/acceptance/glob/glob.sh
+++ b/test/acceptance/glob/glob.sh
@@ -22,8 +22,19 @@ cat /tmp/mocha-glob.txt | grep -q -F '["end",{"suites":1,"tests":1,"passes":1,"p
     exit 1
 }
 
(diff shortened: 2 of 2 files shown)


bin/_mocha
```diff
@@ -323,9 +323,25 @@ var args = program.args;
...
 args.forEach(function(arg){
-  files = files.concat(utils.lookupFiles(arg, extensions, program.recursive));
+  var newFiles;
+  try {
+    newFiles = utils.lookupFiles(arg, extensions, program.recursive);
+  } catch(err) {
+    if (err.message.indexOf('cannot resolve path') === 0) {
+      console.error('Warning: Could not find any test files matching pattern: ' + arg);
+      return;
+    } else {
```
boneskull: This `else` is unnecessary.

bin/_mocha
```diff
@@ -323,9 +323,25 @@ var args = program.args;
...
+      throw err;
+    }
+  }
+
+  files = files.concat(newFiles);
 });
 
+if (files.length === 0) {
```
boneskull: prefer `if (!files.length)`

bin/_mocha
```diff
@@ -323,9 +323,25 @@ var args = program.args;
...
+  process.exit(1);
```
boneskull: Should probably add an error message here about no files being found, given you can simply execute `mocha` in a directory with empty subdir `test`, and this is the current output:

```shell
$ mocha


  0 passing (3ms)

```

bin/_mocha
boneskull: (not required)

bin/_mocha
Munter: Isn't it good practice to re-throw any other error that might not be this specific one? Otherwise it will just be swallowed

bin/_mocha
Munter: Fixed

bin/_mocha
Munter: Fixed

bin/_mocha
Munter: .. And now I get it. Was a bit distracted. Fixed

Munter:
@boneskull Changes implemented, branch rebased and all tests are passing

boneskull:
Thanks @Munter

PR #2468, opened 2016-09-02
Make unittests pass under node 6.5.0; closes #2467

### LINKED ISSUE #2467: Unittests fail in node 6.5.0
With Node 4 unittests fail:

 1) Suite .beforeAll() wraps the passed in function in a Hook adds it to _beforeAll:
 2) Suite .afterAll() wraps the passed in function in a Hook adds it to _afterAll:
 3) Suite .beforeEach() wraps the passed in function in a Hook adds it to _beforeEach:
 4) Suite .afterEach() wraps the passed in function in a Hook adds it to _afterEach:

All have a similar Error:
`AssertionError: expected '"before all" hook: fn' to be '"before all" hook'`

Reason:
In node 6.5.0 anomynous functions have a name when assigned to a variable:

```
var myFunc = function(){};
console.log(myFunc.name); // ->  'myFunc'
```

Prior to node 6.5.0 the function name would be empty

palortoff:
make function to be passed to hook truly anomynous

### DIFF
--- a/test/suite.js
+++ b/test/suite.js
@@ -131,7 +131,7 @@ describe('Suite', function(){
 
     describe('wraps the passed in function in a Hook', function(){
       it('adds it to _beforeAll', function(){
-        var fn = function(){};
+        var fn = (function(){return function(){};})();
         this.suite.beforeAll(fn);
 
         this.suite._beforeAll.should.have.length(1);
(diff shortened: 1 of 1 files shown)


boneskull:
@palortoff Hey, thanks for this fix.  Instead of this approach, can we modify the assertions to be fuzzy?  Something like:

```js
beforeAllItem.title.should.match(/^"before all" hook: namedFn/);
```

palortoff:
I'm not sure I understand what you mean.
The cause for this fix was that anonymous functions have a name under certain circumstances in Node 6.5

How would changing the assertions into regular expressions fix the case where we want to test that there is no title for the hook?

boneskull:
not sure how to explain further; please see #2477 instead, since this problem is failing CI.  thanks!

PR #2470, opened 2016-09-04
avoid test flake in "delay" test; closes #2469

### LINKED ISSUE #2469: appveyor test flake in "options" test for "--delay"
I think it's mostly because `setTimeout()` isn't guaranteed to be precise.  I'll probably just remove one of the assertions.

boneskull:


boneskull:
failure is sauce flake.  I think the sauce connect tunnel is not getting shut down quickly enough.  see #2478

dasilvacontin:
Wait, what? How come that test didn't work? Do you have any insight @boneskull? It would be nice to have a link to an example of this test causing the run to fail.

boneskull:
We have a limited number of concurrent sessions (5) with SauceLabs.

If we limit the concurrent sessions within Karma itself, our tests should not flake, *if and only if* we have **only one (1) build running at any given time**.

If we have multiple builds running, we cannot ensure we won't exceed our concurrent session limit.

I researched this, and there seems to be *no way to limit the number of concurrent builds* on Travis CI.  You can limit the number of concurrent *jobs* (which is only *part of* a build), but this is not necessary, and slows CI *way* down, since it has to run all jobs in serial.

Potential solutions:

1.  Don't run Karma tests on Travis CI; run them on a service or CI server which supports this behavior.
2.  Run Karma tests in Travis CI, but keep them in a separate repository.  That config should have a single job in its matrix, which would use Travis CI's concurrency limit of 1, and Karma's concurrency limit of 5.
3.  Build with Jenkins or something else we have more control over.

From my conversations with @drazisil, it *seems like* CircleCI wouldn't be helpful to us, given the use case.  But I'd love to know his opinion after having looked at our Travis config, and the problems we're having.

boneskull:
@dasilvacontin ^

dasilvacontin:
@boneskull But I thought we were only using Karma (for the browser tests) on one of the TravisCI jobs (?) eg Node 7

paulmelnikow:
Thought I'd pipe in and offer to help, should you decide to move part of the work to CircleCI. I'm pretty familiar with that service, and like it a lot. I've done some work with Sauce Labs in the past, on a [python project](https://github.com/bodylabs/drf-to-s3).

@boneskull BTW, I just recently sent you a note on gitter. I'm looking for a good place to plug in!

PR #2229, opened 2016-04-28
Use the correct option to handle the global error

amsul:

PR #2474, opened 2016-09-11
Use the correct option to handle the global error

amsul:
Continuing from https://github.com/mochajs/mocha/pull/2229

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -60,7 +60,7 @@ process.on = function (e, fn) {
   if (e === 'uncaughtException') {
     global.onerror = function (err, url, line) {
       fn(new Error(err + ' (' + url + ':' + line + ')'));
-      return !mocha.allowUncaught;
+      return !mocha.options.allowUncaught;
     };
     uncaughtExceptionHandlers.push(fn);
   }


boneskull:
@amsul Thanks.  A rebase onto `master` should fix the CI failures.

I have some questions.

So, if we return `false` (when `allowUncaught` is `true`), this means the default event handler will not be executed.  Until this change, it executes.  Does this mean:

- it overrides the browser's default behavior (whatever that is)?  
- it overrides any other global event handler that has been set?  
- can a user set multiple global handlers for a particular event?
- are there any browsers (I'm looking at PhantomJS and IE) that have peculiar behavior here?

If you're unsure about some of these, would it be possible to write an integration test for this?

boneskull:
Moreover, what bug does this fix?

amsul:
The bug this fixes is mentioned here: https://github.com/mochajs/mocha/pull/2229#issuecomment-226589541

Essentially, there was no way to allow errors from propagating to the window and being handled in the dev tools (since mocha catches those errors to allow tests to continue running). Adding this option meant that there is an opportunity to debug these errors by pausing for uncaught errors. However, the implementation had a bug, which is what this corrects.

In regards to your questions:

> it overrides the browser's default behavior (whatever that is)?

It doesn't override the browser's default behavior - just prevents mocha from "catching" the error and allows it to propagate up.

> it overrides any other global event handler that has been set?

Nope

> can a user set multiple global handlers for a particular event?

Not sure I understand what you mean by this..

> are there any browsers (I'm looking at PhantomJS and IE) that have peculiar behavior here?

Nope..they all seem to handle this behavior pretty consistently

boneskull:
@amsul I realize this is really old, but if you feel like signing the CLA & writing a test for this, we can go ahead.  Please let me know, otherwise I'll try to work it out myself.

PR #2475, opened 2016-09-12
Dont count xunit failures as errors

mlucool:
Per http://reflex.gforge.inria.fr/xunit.html#xunitReport a failure and an error are different concepts and must be counted as such. Some parsers will fail if the number of expected tests do not add up.

### DIFF
--- a/test/reporters/xunit.spec.js
+++ b/test/reporters/xunit.spec.js
@@ -5,6 +5,7 @@ var mkdirp = require('mkdirp');
 var path = require('path');
 var assert = require('assert');
 var createStatsCollector = require('../../lib/stats-collector');
+var EventEmitter = require('events').EventEmitter;
 var reporters = require('../../').reporters;
 var XUnit = reporters.XUnit;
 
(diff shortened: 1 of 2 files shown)


boneskull:
@mlucool Thanks. What's an "error"?  Shouldn't we report the number of them instead of `0`?

boneskull:
This is a bug if the output is incorrect, but I don't know enough about XUnit to say whether or not this is true.

mlucool:
@boneskull the documentation says: 
> @errors indicates the number of assertions that weren't true.
> @failures indicates the number of test cases that fail to run. A single report should report 1 or 0, but if several reports are merged, a greater number could be reported.

Then in the xunit.xml file on the ```<test-suite>``` level:
> @errors : the number of test cases that report at least one error
> @failures : the number of test cases that report at least one failure

Reporting failures as errors it is clearly the wrong thing to do, but from a reread, we might be that we want to flip errors and failures in this commit (such that failures is always 0 while errors is the number of tests with issues). As far as I can tell, this does not gracefully handle the error case and when something can't run, it exits so no xunit.xml would be created.

boneskull:
OK.  To summarize, it seems that we need to report *errors* and not *failures* as a first pass.  Otherwise, if a "before" or "beforeEach" hook fails, and thus causes tests to *not run*, we should count those as *failures*.  But that needn't be part of this PR.  Can you please make the necessary change?  Thanks!

mlucool:
Updated - please squash and merge unless you have another change

mlucool:
@boneskull can we merge this?

mlucool:
Reminder: @boneskull

mlucool:
@craigtaub @Munter Any thoughts?

mlucool:
Reminder: @boneskull @ScottFreeCode @Munter

boneskull:
sign CLA if still interested.

craigtaub:
@mlucool poke.

mlucool:
@craigtaub thanks for the poke. Test cases added

@boneskull 
> Thanks. That's weird; why wouldn't a test case run?

I tried to quickly google this, but didn't get far. I *think* errors is the case failing and failures is when you can detect ahead of time that the environment is not correct (e.g. an external dependency is down).

boneskull:
This will need conflicts resolved

mlucool:
@boneskull fixed the conflicts

test/reporters/xunit.spec.js
```diff
@@ -320,6 +321,57 @@ describe('XUnit reporter', function() {
         expect(expectedWrite, 'to be', expectedTag);
       });
     });
+    describe('should write expected summary statistics', function() {
```
craigtaub: LGTM. Although this needs to be `it(`

mlucool: Fixed the `describe` vs. `it` - thanks!

Although these changes test for that issue, the unit test should be there to ensure that all xunit outlook looks correct.

boneskull:
LGTM.  Probably better to use Sinon to mock stuff, but we can change that later.

boneskull:
thanks!

PR #2489, opened 2016-09-19
cleanup after browser tests; closes #2488

### LINKED ISSUE #2488: browser tests should cleanup after themselves
Currently we have to build Mocha (I think) to use it with `karma-mocha`.  But after the tests are complete, we tend to have a modified `mocha.js`.

Solution:  modify the `clean` target to `git checkout -f -- mocha.js` instead of `rm -f mocha.js`.

I _hope_ that works in Windows.

boneskull:
let's see if AppVeyor likes this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ mocha.js: $(SRC) browser-entry.js
 
 clean:
 	@printf "==> [Clean]\n"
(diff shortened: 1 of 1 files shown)


boneskull:
the answer is no

PR #2499, opened 2016-09-22
fix diff output for String objects; closes #2496

### LINKED ISSUE #2496: Objects prematurely coerced into string primitives
This is regarding an issue I reported on the WebStorm tracker: https://youtrack.jetbrains.com/issue/WEB-23383

Differences between symbols that are not strictly equal, appear as being equal and are thus hard to read and confusing.

Here is a simplified version of a test where I noticed this behavior:
```js
"use strict";

const chai           = require( "chai" );
const expect         = chai.expect;

chai.config.truncateThreshold = 0;

describe( "Strings", () => {
    it( "should correctly display differences", () => {
        expect( new String( "foo" ) ).to.equal( "foo" );
    } );
} );
```

Because `new String( "foo" )` is not equal to the primitive `"foo"`, the test will correctly fail. However, in the resulting output, the difference is not immediately apparent.

In this simple test, the `AssertionError` *will* pretty clearly identify the issue. However, in larger objects, this can be much harder to identify, *especially* if the truncation threshold is not raised.

boneskull:
- rename some exported functions/vars as to use locally
- update JSDoc comments

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -89,7 +89,7 @@ exports.map = function(arr, fn, scope) {
  * @param {number} start
  * @return {number}
  */
-exports.indexOf = function(arr, obj, start) {
+var indexOf = exports.indexOf = function(arr, obj, start) {
   for (var i = start || 0, l = arr.length; i < l; i++) {
     if (arr[i] === obj) {
       return i;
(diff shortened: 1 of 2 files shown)


boneskull:
Failing in IE7/IE8; needs more investigation.  We're using json3 to avoid the broken `JSON` implementations, so it shouldn't be that...

boneskull:
`String` seems broken in IE7/IE8 as verified through a SauceLabs session.  The return value of `new String('foo')` is `[undefined, undefined, undefined]`.

PR #2513, opened 2016-09-30
use semistandard

### LINKED ISSUE #2490: use semistandard
We have some code standards in place, but not enough.  We have disabled many warnings which we probably shouldn't have, to avoid changing too much code when we picked up ESLint originally.  I'd also prefer to use a widely-used style instead of a custom style.

- Install eslint-semistandard-config and all necessary dependencies
- Update `.eslintrc` to be just `{"extends": "semistandard"}`
- `eslint --fix` everything
- Manually fix everything else

If anyone disagrees, please let me know.  Unless that disagreement involves "semicolons"; if so, please keep it to yourself.  :smile:

makepanic:
should fix #2490


makepanic:
I think there's a problem because [config-semistandard](https://github.com/Flet/eslint-config-semistandard/blob/master/package.json#L10) and [config-standard](https://github.com/feross/eslint-config-standard/blob/master/package.json#L14) depend on eslint > 3. 
Eslint > 3 requires node 4 or greater.

boneskull:
I'm fine with requiring Node.js 4.x for linting (and newest ESLint).  We only really need to *run tests* on multiple versions of Node.js--and even then, just in CI.  

Best way to accomplish that?  Not sure.  Maybe `prepublish` script which installs ESLint if the Node.js is new enough.  `Makefile` should then check the version before trying to run it.

makepanic:
I rebased with some pinned versions which works on all our AppVeyor node versions, so it shouldn't be a problem anymore.

boneskull:
fine as is for now.  thanks a ton!

PR #2536, opened 2016-10-13
Add --no-diff option (fixes #2514)

### LINKED ISSUE #2514: Option to disable diff
Hi,

We currently have a use case where we're testing a large array for deep equality against something returned from a function. The problem is that when the test fails the output in the reporter is very long and clogs all the tests.

Is there any option to disable the diff? In this case it would be very useful, since we don't really need to know where the output is different from the expected result, but just if it's deeply equal to it or not.

Thanks!

CapacitorSet:
Adds a `--no-diff` flag, which disables diffs completely (takes priority over `--inline-diffs`). Made to fix #2514.


mocha.js
```diff
@@ -1641,6 +1641,20 @@ Mocha.prototype.useInlineDiffs = function (inlineDiffs) {
 };
 
 /**
+ * Do not show diffs at all.
```
boneskull: please don't commit changes to this file.  it's lame we keep it under VCS (b/c bower)

CapacitorSet:
Actually, nevermind, it appears that `err`s already have a `showDiff` property. I'll use that instead.

CapacitorSet:
Done! I squashed the changes as recommended in CONTRIBUTING.md.

CapacitorSet:
Ping @boneskull. This PR still has the *pr-needs-work* label, but I think it's fixed now.

rhali786:
Hey did this get pulled out? I don't see it in the code. I was looking forward to this.

boneskull:
someone needs to review this and probably resolve the conflicts.  as to how soon that'll happen, I can't tell you

boneskull:
closed by 7ceaaf22c934afcd3b39aa57d15c4324f28c1e27 and 575d83d76b97dfa426ee9f1aa436e27f9e96b283

Issue #2535, opened 2016-10-12
--watch crashes mocha on startup


villesau:
I'm getting error below when running mocha in watch mode. As you can see, the file it tries to find is not even js file. Apparently the file link which breaks the mocha is somehow broken. It shouldn't break the mocha though. Without --watch, mocha runs just fine. i'm running mocha version 3.1.2 with following command from package.json:
"test-rc": "mocha 'frontend/*_/_.spec.js' --watch"

The error is following:

Error: ENOENT: no such file or directory, stat '/Users/me/code/my-project/public/index-apitest.php'
at Error (native)
at fs.statSync (fs.js:987:18)
at /Users/me/code/my-project/node_modules/mocha/lib/utils.js:255:11
at Array.forEach (native)
at Object.exports.files (/Users/me/code/my-project/node_modules/mocha/lib/utils.js:253:6)
at /Users/me/code/my-project/node_modules/mocha/lib/utils.js:256:17
at Array.forEach (native)
at Object.exports.files (/Users/me/code/my-project/node_modules/mocha/lib/utils.js:253:6)
at Object.<anonymous> (/Users/me/code/my-project/node_modules/mocha/bin/_mocha:408:26)
at Module._compile (module.js:556:32)
at Object.Module._extensions..js (module.js:565:10)
at Module.load (module.js:473:32)
at tryModuleLoad (module.js:432:12)
at Function.Module._load (module.js:424:3)
at Module.runMain (module.js:590:10)
at run (bootstrap_node.js:394:7)
at startup (bootstrap_node.js:149:9)
at bootstrap_node.js:509:3

villesau:
if following would be wrapped to try-catch, things does not break anymore:

https://github.com/mochajs/mocha/blob/master/lib/utils.js#L255

ScottFreeCode:
Simply ignoring errors at that point isn't likely to be correct behavior -- in this case, as far as I can tell that function shouldn't even be getting paths for files that don't exist, so there's almost certainly something else going wrong that should be addressed; more importantly, in general, suppressing errors could hide and/or worsen other problems.

Any idea where Mocha is getting the path `public/index-apitest.php` from? Do you have a `mocha.opts` file?

villesau:
[...]

PR #2538, opened 2016-10-14
ignore files files which are not found by files()

villesau:
fixes https://github.com/mochajs/mocha/issues/2535#issuecomment-253363169

broken symlinks caused file not found -errors and the system should not fail on those.


lib/utils.js
```diff
@@ -252,10 +252,16 @@ exports.files = function (dir, ext, ret) {
     .filter(ignored)
     .forEach(function (path) {
       path = join(dir, path);
-      if (statSync(path).isDirectory()) {
-        exports.files(path, ext, ret);
-      } else if (path.match(re)) {
-        ret.push(path);
+      try {
```
boneskull: Hi!  Thanks for the PR.

I'm not sure I agree with the strategy here.  If we want to provide "expected" behavior--and we should--then a broken symlink should still be watched. If/when the symlink is fixed while watching, and the symlink is to a directory, we'd want to watch files in *that* directory as well (if requested).  What happens instead is that broken symlinks silently fail, which causes the same problem for users that this PR was meant to address: "it's not working, and I don't know why".

But that implementation sounds like a pain, and I'm not really interested in significant efforts against the `--watch` functionality as part of Mocha core short of low-severity bugfixes.

Instead of trying to further enhance `--watch`, I'd be happy with a solution which leverages `fs.lstatSync()` and re-throws with a helpful error if a symlink is broken.  That'd let the user know something is wrong, and that they'll need to address it if they want Mocha to do what they're telling it to do.

villesau: Thanks for the comments @boneskull !

watching broken symlinks and their content would be nice, but right now it wouldn't follow the current convention: Watch seems to watch only files which mocha can resolve on startup. Thus i think it would be too much of effort to fix the behaviour only for symlinks. 

I wasn't aware of lstatSync function so i ended up to solution I currently propose in this PR. But indeed lstatSync would be more feasible solution here. I played around with lstatSync a bit and it wouldn't throw at all if the symlink is broken - which is fine in my use case at least. Instead, mocha would fail later on to 'module not found' if it tries to require module from broken symlink, which again sounds like a reasonable behaviour to me. That exception also points out the failing module so it's easy to sort out why it is failing.

So what I propose is that I remove the try-catch and replace statSync with lstatSync here.

Watch option built into mocha is pretty important to us since starting up and setting things up takes pretty long time in larger projects.

villesau:
@boneskull what do you think of current approach?

lib/utils.js
```diff
@@ -252,9 +254,9 @@ exports.files = function (dir, ext, ret) {
     .filter(ignored)
     .forEach(function (path) {
       path = join(dir, path);
-      if (statSync(path).isDirectory()) {
+      if (lstatSync(path).isDirectory()) {
         exports.files(path, ext, ret);
-      } else if (path.match(re)) {
+      } else if (path.match(re) && existsSync(path)) {
```
ScottFreeCode: Using `existsSync` here still silently ignores the problem rather than making the error message useful as discussed in the previous review, doesn't it?

test/acceptance/file-utils.spec.js
```diff
@@ -7,8 +7,8 @@ var os = require('os');
 var mkdirp = require('mkdirp');
 var rimraf = require('rimraf');
 
-describe('lookupFiles', function () {
-  var tmpDir = path.join(os.tmpDir(), 'mocha-lookup-files');
+describe('files utils', function () {
+  var tmpDir = path.join(os.tmpDir(), 'mocha-files');
```
ScottFreeCode: `'mocha-files'` isn't any less function-specific since the other function's name is just `.files`, it's just more vague (outside of the context of the function names anyway) and less resistant to name collision; I'd recommend leaving this directory name as `'mocha-lookup-files'` or, if you really want to avoid implying it related only to one of the two functions, changing to `'mocha-file-lookup'`.

test/acceptance/file-utils.spec.js
```diff
@@ -36,46 +36,68 @@ describe('lookupFiles', function () {
...
-      .be(true);
-    fs.renameSync(tmpFile('mocha-utils.js'), tmpFile('bob'));
-    expect(existsSync(tmpFile('mocha-utils-link.js')))
-      .to
-      .be(false);
-    expect(utils.lookupFiles(tmpDir, ['js'], false))
-      .to
-      .eql([]);
+  describe('.lookupFiles', function () {
+    (symlinkSupported ? it : it.skip)('lookupFiles keke should not choke on symlinks', function () {
```
ScottFreeCode: Test description should omit "lookupFiles" (which is in the suite name/description, which is generally printed before the test name) and "keke".

test/acceptance/file-utils.spec.js
```diff
@@ -36,46 +36,68 @@ describe('lookupFiles', function () {
...
+        .have
+        .length(expectedLength);
+    });
   });
 
-  it('should accept a glob "path" value', function () {
-    var res = utils.lookupFiles(tmpFile('mocha-utils*'), ['js'], false)
-      .map(path.normalize.bind(path));
+  describe('.files', function () {
+    (symlinkSupported ? it : it.skip)('should not choke on broken symlinks', function () {
```
ScottFreeCode: One of the test descriptions specifies "broken" and the other does not; these should be consistent, since it's substantially the same test for the two different functions.

lib/utils.js
villesau: Yeah true. I removed  existsSync now so it still fails if .js symlinks files are broken, but also gives better error message when trying to require it later on. To me personally both are fine as our major problem was with other files than js.

villesau:
@ScottFreeCode @boneskull changed the logic to not to hide the broken symlink.

ScottFreeCode:
Something interesting that may be relevant: I tried this out, and it turns out that broken symlinks to files with this change actually *are* watched -- as in, if you later create the file to which the symlink should have been pointing, it triggers the test suite to be re-run. So, except for the case of directories (where, I think, villesau has a good point that adding more files after watch start would be inconsistent with current watch behavior in general), this actually seems to fix the behavior altogether, if I'm understanding correctly. (Unless the concern is that the symlink might be broken because it's got the wrong name for a file that does exist but isn't being watched because the symlink to it is broken?) @boneskull What do you think? (GitHub still lists your review as outstanding, by the way.

boneskull:
@ScottFreeCode thanks for the heads-up.  

If a symlink was intended to be watched, is broken, and *then* is fixed while watching, this is arguably "inconsistent" with the watch behavior elsewhere.  

But fixing a broken symlink is not conceptually the same as adding a file.

I don't want to overthink this too much.  LGTM

@villesau thanks!

PR #2544, opened 2016-10-16
Fix .only() behaviour when running with watch option, fixes #2429

### LINKED ISSUE #2429: --watch option works strangely with `.only()`
### steps to reproduce

``` js
// test.js 
// Sample code to reproduce 
  it('should be ok', () => {
    //
  });
    //
  });
});
```

```
// output
  test with --watch
    ✓ should be ok
    ✓ should be ok too

  2 passing (8ms)
```

```
// output
  test with --watch
    ✓ should be ok too

  1 passing (2ms)
```

```
// output
  0 passing (1ms)
```
### Expected Behavior

### Environmet
- node : 6.3.0
- mocha : 3.0.2

davidspinat:
Since version `3.0.0` the `watch` option does not work as expected in combination with `.only()`.

@fediev already described in detail https://github.com/mochajs/mocha/issues/2429 the misbehaviour.


boneskull:
@davidspinat thanks.  can you please [sign the CLA](https://cla.js.foundation/mochajs/mocha?pullRequest=2544)

davidspinat:
@boneskull of course 👍

davidspinat:
@boneskull may I help you with something else? :)

boneskull:
@davidspinat Sorry, bandwidth problems.  

*This may not be feasible*, but can you provide an integration test?  If you're having trouble, I can dig deeper and suggest something.

Failing that, maybe you see if you can't re-create the problem via other means...

davidspinat:
@boneskull Sorry haven't managed to provide an integration test. However I would suggest the following:

In https://github.com/mochajs/mocha/blob/master/test/integration/helpers.js#L64 we might extend the signature of `runMocha` with the an argument called times e.g.:

`runMocha: function (fixturePath, args, fn, times) ...`

Further we might sequential call `invokeMocha` according the passed times value and push all the results/summaries into an Array. The callback `fn` should be resolved with this Array.

What's your opinion about my suggestion?

jzaefferer:
For the record, I had the same issue as reported in #2429 and just tested this branch locally, it works and fixes issue. Not the same as automated tests, but considering the two changes relatively trivial, maybe its good enough.

davidspinat:
@boneskull how should we proceed here? :)

threepears:
@boneskull : just curious if this is being, or has been, integrated yet. My team is waiting to update our Mocha version until this issue is fixed, as we found the problem with ".only()" to break behavior we had relied on.  This ticket is still listed as open, but if you've integrated this into a recent version, please let us know.  Thank you!

juank11memphis:
Is this bug getting fixed any time soon? 
Thanks!

davidspinat:
Is there any chance this PR will be merged? How might we add a test which covers this fix? I guess everyone would love to see this merged. Please leave some feedback.

Greetings

boneskull:
@davidspinat Can you please revert the change to the root `mocha.js`?  This file is generated (manually, currently) when we release.

boneskull:
A proper harness for testing watch-related behavior would be necessary if we were to make enhancements in this subsystem (or re-implement it).

The risk here is that the change involves code that's run regardless of whether `--watch` is invoked.  If someone (like @jzaefferer?) has been running against this branch for awhile, it'd help to know that it's not causing problems elsewhere.

lib/runner.js
```diff
@@ -795,7 +795,7 @@ Runner.prototype.run = function (fn) {
   var rootSuite = this.suite;
 
   // If there is an `only` filter
-  if (this.hasOnly) {
+  if (hasOnly(rootSuite)) {
```
boneskull: Okay, so it looks like with the addition of `--forbid-only`, this may need some changes.

`Runner#hasOnly` is set if e.g. a `describe.only()` is encountered (see `lib/interfaces/common.js`).

The `--forbid-only` functionality will check this value.  At the point that it's checked, the `Runner` may or may not have a `suite` property.  This means there's likely some weird interactions between `--forbid-only` and `--watch`.

A solution may be to set `Runner#hasOnly` to `true` if the result of `hasOnly()` is `true`.  `hasOnly()` may also want to *check* the value of `Runner#hasOnly`.  When the `Runner`'s `end` event is handled, it may also need to reset `Runner#hasOnly` to `false`.

I haven't fiddled with this too much, but this is the matrix we're trying to assert is OK:

1.  `describe.only()`
2.  `it.only()`
3.  `--watch`
4.  `--forbid-only`

For example, run a `--watch` with `--forbid-only`, then add a `describe.only()`, save, then remove it again & save.  What happens?

boneskull: (careful with nested suites here; `hasOnly()` looks like it only inspects the innards of any given suite, and not if that suite was itself marked as an "only" suite; that information is stored in the aforemention suite's parent! :smile:)

ScottFreeCode: Worth noting that #2874 also touches `hasOnly`-using code in addition to the `--forbid-only` code already merged doing so.

(I plan on a more detailed review later this week.)

jzaefferer:
> The risk here is that the change involves code that's run regardless of whether --watch is invoked. If someone (like @jzaefferer?) has been running against this branch for awhile, it'd help to know that it's not causing problems elsewhere.

I can't say how long I was running this branch locally, might've been a few weeks though. I never noticed any regressions.

Looks like this will requires further changes now - let me know if I can help testing the result.

ScottFreeCode:
Waaait a second, I just realized something else: @mochajs/core *is `runner.hasOnly`, the (potentially stale) property, meant to be public?* Because if so:

- We'd have to wait for a "major" release to remove it. (As previously stated, I believe it is ultimately correct to replace it with a function that checks the current testsuite so stateleness and any other source-of-truth problems cannot occur; so I would advocate for doing so eventually.)
- Until then, we'd want to fix it being incorrect on subsequent runs for consumers of the runner and not just for this built-in behavior.

In other words, ***if*** `runner.hasOnly` is public, I think the right thing to do may be to table this PR in favor of #2934 or something like it, and instead use my recommendations (remove setting the `hasOnly` property to ensure it can no longer be used, fix failing tests by updating uses to use the `hasOnly` function) when the time comes to make a "major" release.

(That's not my preferred course of action in principle simply because I'd rather get the root cause corrected now, but backwards compatibility of public interfaces would necessitate it ***if*** `runner.hasOnly` is meant to be public...)

boneskull:
if it's documented or marked `@api` or `@public` then yes. otherwise search github and see if anyone is consuming it (unlikely but you never know). 

I have no problems with a major release. lets fix root causes.

ScottFreeCode:
Quick assessment on that:
- `Runner`'s `this.hasOnly` is only used in `Runner.prototype.run`, which is public but does not document `hasOnly` in the comment; no other reference to `hasOnly` the property is in `lib/runner.js` (the only other `hasOnly` is the function).
- `runner.hasOnly` is set by the `Mocha.prototype.run` function, which is public but merely copies the property from its `this.options`, with no mention of `this.options` in its comment doc.
- `this.options` is set by the `Mocha` constructor, which is public but has a list of `options` that `hasOnly` is not one of.
- There are no other mentions of `hasOnly` in `lib/mocha.js` either.
- `lib/interfaces/common.js` sets `mocha.options.hasOnly` in its `only` functions and (redundantly?) in one of the `create` functions. None of these are documented in comments as being public -- the whole file seems to be intended for implementing the actual interfaces. (Speaking of which, the proposed change should not have any additional backwards-compatibility issues with properly written custom interfaces, as the interface would currently need to be setting both `hasOnly` and `_onlyTests`/`_onlySuites`, and we'd merely be making the former ignored in favor of the latter.)
- No other files besides these three reference `hasOnly`.

Seems like a bit of a side-channel on top of everything else...

davidspinat:
Wow, just failed rebasing that one. Will fix that.

davidspinat:
@ScottFreeCode @boneskull 

Just updated the PR to include the requested changes. Please let me know if i didn't manage to meet your feedback. I'm happy to help.

- [x] Rebase with master
- [x] Remove all additional occurrences of the hasOnly property

Cheers.

ScottFreeCode:
Went through all the GitHub code results (yes, all forty pages) for `hasOnly extension:js extension:coffee extension:ts extension:typescript extension:flow extension:es extension:mjs extension:jsx` (also looked at `extension:cs` but it gave me seven pages of C# and no CoffeeScript) and here's what I found:
- irrelevant:
  - `hasOnly` as a function, or as a variable rather than as a property (so obviously not it)
  - other test runners I'd never heard of
  - `hasOnly` differently capitalized and/or as the start of some other name instead of as the whole word/identifier (so also obviously not it)
  - forks of Mocha such as [mocha-nightwatch](/nightwatchjs/mocha-nightwatch)
  - copies of the Mocha code in other projects such as https://github.com/Dudinia/Arithmetic/tree/3c90c6e4b62b5a3cdf064b1775a884993a5ead47/TestProject/lib/mocha
- relevant:
  - this: https://github.com/pflannery/mocha-ui-esm/blob/8a6440f42685bfd6c1f2b5f0576c1c28abeb5c3c/index.js#L27 which seems to be setting the flag and therefore as previously discussed shouldn't be affected?
  - this: https://github.com/malexandert/metamocha/blob/c36b077e5c215595ec2ffd8fae57f6032846a071/lib/metadata_ui.js#L88 which I'm somewhat more confident is setting the flag such that it won't be affected as previous discussed
  - this: https://github.com/twolfson/multi-image-mergetool/blob/43c06fb4d5c8d751ccbd06cb7910817e7197a5b8/test/browser/utils/application.js#L127-L151 (see next)
  - ...and this: https://github.com/twolfson/react-playground/blob/5e559f37d3106a340781de5d5d6cfc3d82715f66/test/common/utils/react.js#L35-L61 which seem to be a couple instances of reading (rather than writing) `options.hasOnly` that would therfore be affected -- BUT, I believe we could remove the `runner.hasOnly = options.hasOnly` line *without* removing the setting of `options.hasOnly = true` and avoid affecting them until a semver "major" release if we want to be cautious.

I'm marking my prior request for changes as having been satisfied, in any case.

lib/runner.js
ScottFreeCode: This should be merged right now, I think.

ScottFreeCode:
Still trying to figure out if I'm merging master into this correctly (which was necessary because conflicts); there's one failing test, looking into that.

lib/runner.js
```diff
@@ -429,7 +429,7 @@ Runner.prototype.runTest = function (fn) {
   if (!test) {
     return;
   }
-  if (this.forbidOnly && this.hasOnly) {
+  if (this.forbidOnly && hasOnly(this.parents().reverse()[0])) {
```
ScottFreeCode: ...Just realized this *still* isn't right, because it won't work for tests outside of any `describe`. Remind me to add a test for that, but first, gonna fix this and merge...

PR #2559, opened 2016-10-28
fix circular objects in json reporter. Closes #2433

### LINKED ISSUE #2433: JSON reporter doesn't work (no output)
The issue occur to me on mocha v3.0.2 and node v5.8.0 on Mac OS X. Here's my test file

``` javascript
'use strict';
var app = require('./app');

describe('Converting strings.', function() {

  it('Can convert a string with one space', function() {
    var result = app.convertString('hello world');
    result.should.eql('hello-world');
  });

  it('Replaces two spaces with one dash', function() {
    app.convertString('hello  world').should.eql('hello-world');
  });
});
```

and the code I'm testing

``` javascript
function convertString(str) {
  return str.split(' ').join('-');
}

module.exports = {
  convertString: convertString
};
```

With `json-stream` reporter everything works fine:

```
$ mocha --require should --reporter json-stream test.js
["start",{"total":2}]
["pass",{"title":"Can convert a string with one space","fullTitle":"Converting strings. Can convert a string with one space","duration":2,"currentRetry":0}]
["fail",{"title":"Replaces two spaces with one dash","fullTitle":"Converting strings. Replaces two spaces with one dash","duration":0,"currentRetry":0,"err":"expected 'hello--world' to equal 'hello-world'","stack":"AssertionError: expected 'hello--world' to equal 'hello-world'\n    at Assertion.fail (node_modules/should/cjs/should.js:205:17)\n    at Assertion.Object.defineProperty.value (node_modules/should/cjs/should.js:277:19)\n    at Context.<anonymous> (test.js:12:46)"}]
["end",{"suites":1,"tests":2,"passes":1,"pending":0,"failures":1,"start":"2016-08-12T15:34:45.144Z","end":"2016-08-12T15:34:45.155Z","duration":11}]
```

but with the json reporter nothing comes out:

```
$ mocha --require should --reporter json test.js  # No output!
```

Am I doing something wrong?

jeversmann:
In #2433, @boneskull recommended using `jsonStringify()` from `./lib/utils`, but that function isn't exported by the utils module, and more importantly it isn't actually circular-reference safe.

Instead, I took the custom replacer from [this StackOverflow post](http://stackoverflow.com/questions/11616630/json-stringify-avoid-typeerror-converting-circular-structure-to-json) and added it to the reporter's invocation of `JSON.stringify`.

The test for this issue is mostly a copy of the 1st json reporter test. It has fewer asserts because there's no way to test for the change to the object that is made by the replacer, but the test _does_ fail without the additional code.

### DIFF
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -54,18 +54,7 @@ function JSONReporter (runner) {
 
     runner.testResults = obj;
 
-    var cache = [];
-    process.stdout.write(JSON.stringify(obj, function (key, value) {
-      if (typeof value === 'object' && value !== null) {
-        if (cache.indexOf(value) !== -1) {
-          // Instead of going in a circle, we'll print [object Object]
-          return '' + value;
-        }
-        cache.push(value);
-      }
-
-      return value;
-    }, 2));
+    process.stdout.write(JSON.stringify(obj, null, 2));
   });
 }
 
@@ -78,17 +67,43 @@ function JSONReporter (runner) {
  * @return {Object}
  */
 function clean (test) {
+  var err = test.err || {};
+  if (err instanceof Error) {
+    err = errorJSON(err);
+  }
+
   return {
     title: test.title,
     fullTitle: test.fullTitle(),
     duration: test.duration,
     currentRetry: test.currentRetry(),
-    err: errorJSON(test.err || {})
+    err: cleanCycles(err)
   };
+}
+/**
+ * Replaces any circular references inside `obj` with '[object Object]'
+ *
+ * @api private
+ * @param {Object} obj
+ * @return {Object}
+ */
+function cleanCycles (obj) {
+  var cache = [];
+  return JSON.parse(JSON.stringify(obj, function (key, value) {
+    if (typeof value === 'object' && value !== null) {
+      if (cache.indexOf(value) !== -1) {
+        // Instead of going in a circle, we'll print [object Object]
+        return '' + value;
+      }
+      cache.push(value);
+    }
+
+    return value;
+  }));
 }
 
 /**
- * Transform `error` into a JSON object.
+ * Transform an Error object into a JSON object.
  *
  * @api private
  * @param {Error} err
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -54,7 +54,18 @@ function JSONReporter (runner) {
 
     runner.testResults = obj;
 
-    process.stdout.write(JSON.stringify(obj, null, 2));
+    var cache = [];
+    process.stdout.write(JSON.stringify(obj, function (key, value) {
+      if (typeof value === 'object' && value !== null) {
+        if (cache.indexOf(value) !== -1) {
+          // Instead of going in a circle, we'll print [object Object]
+          return '' + value;
+        }
+        cache.push(value);
+      }
+
+      return value;
+    }, 2));
   });
 }
 
--- a/mocha.js
+++ b/mocha.js
@@ -3061,7 +3061,18 @@ function JSONReporter (runner) {
 
     runner.testResults = obj;
 
-    process.stdout.write(JSON.stringify(obj, null, 2));
+    var cache = [];
+    process.stdout.write(JSON.stringify(obj, function (key, value) {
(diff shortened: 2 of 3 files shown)


Munter:
Nice one. I'm not a big fan of the duplication though. Should this be put in its own reusable util function?

jeversmann:
There's no duplication here? This one reporter is the only spot I'm aware of with this problem.

ScottFreeCode:
IMHO if the version in our `util` component isn't actually circular-safe we should fix it (or make it a reference to a dependency that's safe), and if it isn't exported it should be, so other components such as reporters can just call on the util function.

boneskull:
hmm, yes, it seems [canonicalize()](https://github.com/mochajs/mocha/blob/36184aa78593772032988efa8845f3dc996b5c6f/lib/utils.js#L582-L582) is circular-safe, which is what I thought I was pointing you to.  oops.  but that function is for displaying diffs.

boneskull:
It would *seem* a public API to output JSON--while avoiding circular references--would be helpful to reporter implementations.  That's tangential to #2433.

Looking more closely at the JSON reporter, it already [purports](https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/lib/reporters/json.js#L61-L61) to kill circular references.  The bug is that *it doesn't actually succeed in doing this*.

The JSON stream reporter works because it doesn't attempt to "stringify" an `Error`--it just yanks the `message` and `stack` properties out.

I think what this PR should do is actually very fine-tuned to this bug.  @jeversmann Can you instead modify [errorJSON()](https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/lib/reporters/json.js#L86-L86) or [clean()](https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/lib/reporters/json.js#L69) to skip objects it's already seen?  

I don't know what the "best" implementation would look like here, but we should retain as much of the `Error` as is feasible.  We shouldn't take the easy way out and do what the JSON stream reporter does. That would would be a breaking change, and not a bug fix.

jeversmann:
I moved the circular object logic into a helper function that's called in `clean` so it behaves according to the comment above it.

The 'gotcha' here is that `Error` objects don't `JSON.stringify` correctly, so formatting `test.err` is a two-step process which might behave strangely if a test throws an object that isn't an `instanceof Error` but contains one or more `Error`s as children. [The suggested fix seems to be defining `Error.prototype.toJSON`](http://stackoverflow.com/a/18391400) but that seems like overkill unless the current method causes problems for anyone.

ScottFreeCode:
> ...might behave strangely if a test throws an object that isn't an  instanceof Error  but contains one or more  Error s as children.

I thought Mocha doesn't accept non-`Error` failures in the first place?

jeversmann:
The problem in #2433 deals specifically with an object whose prototype chain contains `Error`, but the test I wrote to provoke the circular parsing problem just `throw`s a regular object with a circular reference in it.

I can modify the test to build the thrown object from an `Error` but the code to handle printing it shouldn't change.

ScottFreeCode:
Yeah, I'm not too concerned with having a test that focuses on the circular reference thing without worrying about rigging up a "real" `Error`; I just wasn't sure how the comment about handling the case of `Error` as a member of non-`Error` objects was relevant. Sorry for any confusion!

jeversmann:
@boneskull This is still labeled as pr-needs-work, but I think I've addressed the feedback.
The bot is right that this PR has become stale, but #2433 is still around, so I'd rather merge this in.

boneskull:
@jeversmann thanks for the heads-up.  someone will take a look at this, maybe even me!

xyleen:
Hey! I've stumbled upon this situation myself and the proposed commits seem to resolve the issue for me. 

Do you plan on merging this pull request sometime? Is this PR missing anything that prevents it from being mergable?

boneskull:
continued in #3318.  thanks

boneskull:
landed in https://github.com/mochajs/mocha/commit/741b0bd4eff23b2388b605c8f4cda99d2d7ea38a

PR #2560, opened 2016-10-28
Add a more clear timeout message. Closes #2294

### LINKED ISSUE #2294: Clarify timeout message
We should consider updating the timeout error message from "timeout of [time] exceeded. Ensure the done() callback is being called in this test." to something like "timeout of [time] exceeded. Ensure the test completes within [time] or increase the timeout set, and ensure the done() callback is being called when this test completes." The current message makes it sound like the problem is forgetting to call `done` even when the problem is simply that the test took too long or that some asynchronous action never got called back.

jeversmann:
Since there isn't currently a way to tell whether the test tried to return a promise or not where we throw the timeout error, I simply went with the suggestion by @dasilvacontin.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -229,7 +229,8 @@ Runnable.prototype.resetTimeout = function () {
     if (!self._enableTimeouts) {
       return;
     }
-    self.callback(new Error('timeout of ' + ms + 'ms exceeded. Ensure the done() callback is being called in this test.'));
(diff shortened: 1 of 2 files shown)


jeversmann:
The Travis failure happened in `npm install`. Is there a way to re-👊  it?

Munter:
Rerun and passed. Thanks!

PR #2587, opened 2016-11-22
FEATURE Allow comments in `mocha.opt` file.

igwejk:
This should enable switching an option off/on with ease by
commenting/uncommenting the option. A sample use case is the need
to enable the `--debug-brk` option  during debugging but disabling
it otherwise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ else
 endif
 
 TM_BUNDLE = JavaScript\ mocha.tmbundle
-SRC = $(shell find lib -name "*.js" -type f | sort)
+SRC = $(shell find lib -name "*.js" -type f | LC_ALL=C sort)
 TESTS = $(shell find test -name "*.js" -type f | sort)
 
 all: mocha.js


boneskull:
@igwejk I don't understand why this is necessary.  If you just need to add `--debug-brk`, execute:

```shell
$ mocha --debug-brk
```

and it will be merged w/ the rest of your flags defined in `mocha.opts`.  

Can you please explain?

igwejk:
Hi Christopher, that example was just a single use case which is also the simplest.

More generally I could have multiple flags, some of which are useful at one instance but not at a different instance. When I don't find a flag useful (but I expect it to fill my need at some other point in time) I would like to comment it out instead of deleting it completely.

Also, consider that the suggested feature can serve as enabling a means for easily documenting used options.
________________________________

jwalton:
Here's another use case which I'd unfortunately like to use right now:

```
--compilers js:babel-register,jsx:babel-register
--check-leaks
# Add `addresses` here for now because https://github.com/webpack/webpack/issues/3173
--globals window,navigator,localStorage,document,addresses
```

ELLIOTTCABLE:
@igwejk any particular reason you closed this? I'd personally still love to see this behaviour, for the same reason @jwalton mentioned.

igwejk:
Nice to know there are others who want this. I'll reopen it by the end of the day. Thanks.

/c @ELLIOTTCABLE , @jwalton

igwejk:
This PR has been reopened via #2870 2870

PR #2590, opened 2016-11-23
Add Browser Support Matrix to README.md

dasilvacontin:
Will look like this: 
screen shot 2016-11-23 at 6 47 44 pm

It's kinda the best I managed after quite a few tries.


boneskull:
don't mind if I squash, do you? 😉

dasilvacontin:
@boneskull I want all the contribs I can get!

PR #2593, opened 2016-11-24
Revert test pass symbol in List reporter

Aldaviva:
Hello,

The List reporter's output regressed during a [past refactoring](https://github.com/mochajs/mocha/commit/bec7a6e1686cf1f4c26ef6c2ba25eda36ccdd6c5#diff-855bef20eedfd15010e9e05ba5ddf02f). This pull request reverts the output to the [old correct output](https://mochajs.org/#list) but it still retains the benefit of the refactoring.

All tests passing.

### DIFF
--- a/lib/reporters/list.js
+++ b/lib/reporters/list.js
@@ -42,7 +42,7 @@ function List (runner) {
   });
 
   runner.on('pass', function (test) {
-    var fmt = color('checkmark', '  ' + Base.symbols.dot) +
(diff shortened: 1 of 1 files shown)


Aldaviva:
Thank you! I noticed that my Putty window was showing green question mark icons for passing tests, even though my font (Consolas) has the `CHECK MARK (\u2713)` glyph. Turns out Consolas doesn't have `MIDDLE DOT (\u00B7)`.

boneskull:
@Aldaviva nice one.

PR #2601, opened 2016-11-26
refactors .only() behaviour, closes #2429 and #2606

### LINKED ISSUE #2429: --watch option works strangely with `.only()`

villesau:
This pr moves hasOnly logic into suites and tests instead of global state and filter tests and suites based on suites' and tests' hasOnly and isOnly flags. Also removes need to duplicate suite arrays which means less state to maintain.

fixes #2429 and #2606

Unfortunately i was not able easily to write tests for #2429 even though this pr solves it. eg. regression tests does not have option to run same instance of mocha continuously which makes validating the fix more difficult. If you have any hints for writing tests for this, feel free to suggest :)

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -3,8 +3,8 @@
 var Suite = require('../suite');
 
 function setOnlies (suite) {
-  if ((suite.hasOnly || suite.isOnly) && suite.parent) {
-    suite.parent.hasOnly = true;
+  if ((suite._hasOnly || suite._isOnly) && suite.parent) {
+    suite.parent._hasOnly = true;
     setOnlies(suite.parent);
   }
 }
@@ -113,7 +113,7 @@ module.exports = function (suites, context, mocha) {
         suite.file = opts.file;
         suites.unshift(suite);
         if (opts.isOnly) {
-          suite.isOnly = true;
+          suite._isOnly = true;
         }
         setOnlies(suite);
         if (typeof opts.fn === 'function') {
@@ -137,7 +137,7 @@ module.exports = function (suites, context, mocha) {
        * @returns {*}
        */
       only: function (mocha, test) {
-        test.isOnly = true;
+        test._isOnly = true;
         setOnlies(test);
         return test;
       },
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -2,13 +2,6 @@
 
 var Suite = require('../suite');
 
-function setOnlies (suite) {
-  if ((suite._hasOnly || suite._isOnly) && suite.parent) {
-    suite.parent._hasOnly = true;
-    setOnlies(suite.parent);
-  }
-}
-
 /**
  * Functions common to more than one interface.
  *
@@ -115,7 +108,6 @@ module.exports = function (suites, context, mocha) {
         if (opts.isOnly) {
           suite._isOnly = true;
         }
-        setOnlies(suite);
         if (typeof opts.fn === 'function') {
           opts.fn.call(suite);
           suites.shift();
@@ -138,7 +130,6 @@ module.exports = function (suites, context, mocha) {
        */
       only: function (mocha, test) {
         test._isOnly = true;
-        setOnlies(test);
         return test;
       },
 
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -2,6 +2,13 @@
 
 var Suite = require('../suite');
 
+function setOnlies (suite) {
+  if ((suite.hasOnly || suite.isOnly) && suite.parent) {
+    suite.parent.hasOnly = true;
+    setOnlies(suite.parent);
+  }
+}
+
 /**
  * Functions common to more than one interface.
  *
@@ -74,7 +81,6 @@ module.exports = function (suites, context, mocha) {
        * @returns {Suite}
        */
       only: function only (opts) {
-        mocha.options.hasOnly = true;
         opts.isOnly = true;
         return this.create(opts);
       },
@@ -107,9 +113,9 @@ module.exports = function (suites, context, mocha) {
         suite.file = opts.file;
         suites.unshift(suite);
         if (opts.isOnly) {
-          suite.parent._onlySuites = suite.parent._onlySuites.concat(suite);
-          mocha.options.hasOnly = true;
+          suite.isOnly = true;
         }
+        setOnlies(suite);
         if (typeof opts.fn === 'function') {
           opts.fn.call(suite);
           suites.shift();
@@ -131,8 +137,8 @@ module.exports = function (suites, context, mocha) {
        * @returns {*}
        */
       only: function (mocha, test) {
-        test.parent._onlyTests = test.parent._onlyTests.concat(test);
-        mocha.options.hasOnly = true;
+        test.isOnly = true;
+        setOnlies(test);
         return test;
       },
 
--- a/test/integration/fixtures/options/only/bdd.fixture.js
+++ b/test/integration/fixtures/options/only/bdd.fixture.js
@@ -47,26 +47,43 @@ describe.only('should run this suite, even', function () {
 describe('should run this suite with an exclusive test', function () {
   it.only('should run this test', function () {});
 
-  describe('should not run this nested suite', function () {
-    describe.only('should not run this doubly-nested suite', function () {
-      it('should not run this test', function () {});
+  it('should not run this test', function () {
+    (true).should.equal(false);
+  });
 
-      it('should not run this test', function () {});
+  describe('should run this nested suite', function () {
+    it('should not run this test', function () {
+      (true).should.equal(false);
+    });
 
-      it('should not run this test', function () {});
+    describe.only('should run this doubly-nested suite', function () {
+      it('should run this test', function () {});
+
+      it('should run this test', function () {});
+
+      it('should run this test', function () {});
     });
   });
 });
 
 describe('should run this suite with an exclusive test (reverse order)', function () {
-  describe('should not run this nested suite', function () {
-    describe.only('should not run this doubly-nested suite', function () {
-      it('should not run this test', function () {});
+  describe('should run this nested suite with an exclusive suite', function () {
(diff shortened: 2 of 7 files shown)


lib/interfaces/common.js
```diff
@@ -2,6 +2,13 @@
 
 var Suite = require('../suite');
 
+function setOnlies (suite) {
+  if ((suite.hasOnly || suite.isOnly) && suite.parent) {
+    suite.parent.hasOnly = true;
+    setOnlies(suite.parent);
+  }
+}
+
```
ScottFreeCode: Seeing as this is called with both tests and suites, it is somewhat confusing that the parameter name suggests it is only meant to be called with suites. I also believe it would be preferable to document this function even though it is private.

test/integration/fixtures/options/only/bdd.fixture.js
```diff
@@ -47,26 +47,43 @@ describe.only('should run this suite, even', function () {
...
+      (true).should.equal(false);
     });
   });
+
   it.only('should run this test', function () {});
+
+  it('should not run this test', function () {
+    (true).should.equal(false);
+  });
 });
```
ScottFreeCode: Adding failing assertions to the should-not-run tests here is misleading, since what matters is the integration test verifying that they did not run (and that the should-run tests did). It could, however, be appropriate to add a comment explaining that setup instead, if you'd like to help make it more obvious for future contributors.

ScottFreeCode:
Re. regression testing for the issue, I imagine it would have to be somewhat like the integration test that this currently changes, where the real test runs a separate instance of Mocha that uses some tests just so the real test can check its behavior when run with them; except that instead of running once with one set of tests, it would need to be run twice with two different sets that differ only in their use of `only`, or perhaps run with `--watch` and different sets of tests that differ only in their use of `only` could be copied over the directory that it's set to watch.

villesau:
@ScottFreeCode thanks for the comments!

- I think it's fine to merge the other pr first, so that it fixes the --watch issue and then this so that this is scoped only to nested onlies. Refactoring easily tends to have effect to multiple issues.

- The benefits of getting rid of _onlySuites and _onlyTests are that suites and tests are not duplicated into separate arrays, but there is more like single source of truth. Having suites in multiple places is potential risk for inconsistent data.

- _onlySuites and _onlyTests were treated as public even though by convention they were private, see the line where private members are accessed in public manner: https://github.com/mochajs/mocha/pull/2601/files#diff-b3b53682a18f203ac8d29b0e277cad26L872 Thus i think it's more convenient to remove underscore to state that they really are public. If it's necessary to have isOnly and hasOnly private, one way of achieving this would be to move filtering logic into suites which might also be feasible solution.

- Having both hasOnly and isOnly makes filtering easier: When parent knows it has only-suites or tests, we do not need to map over _all_ the tests and suites on every run. The idea is to propagate the info "hey, i have some childs with only flag on, go ahead and filter others out". hasOnly is more like a signpost to right direction where every point of nesting is crossroad with multiple potential paths. I'm not 100% happy to have both flags either, but so far it felt most feasible solution to the problem.
The actual difference of isOnly and hasOnly comes from this line: https://github.com/mochajs/mocha/pull/2601/files#diff-b3b53682a18f203ac8d29b0e277cad26R863
This could be maybe documented better into functions dealing with these.

test/integration/fixtures/options/only/bdd.fixture.js
villesau: yeah you are correct. I just tried to follow convention in lines above. I could fix the rest of the test cases too wit appropriate comment(s)

villesau:
And what comes to testing, i could try to come up with some. The case with --watch is that it's currently non-trivial to test so it would require probably more a lot more work to test than what it took to make the fix. Testing this using programmatic api is currently pretty difficult too as it does not expose stats when tests are run, just failures.

ScottFreeCode:
>Testing this using programmatic api is currently pretty difficult too as it does not expose stats when tests are run, just failures.

That's odd -- there isn't pending and passing too? If not, we should probably fix that; I can't imagine what good failure count does without the other two (and if we ever want to drop or replace them, we'd have to anyway if failures is already out there).

ScottFreeCode:
>•I think it's fine to merge the other pr first, so that it fixes the --watch issue and then this so that this is scoped only to nested onlies.

:+1:

>•The benefits of getting rid of _onlySuites and _onlyTests are that suites and tests are not duplicated into separate arrays, but there is more like single source of truth. Having suites in multiple places is potential risk for inconsistent data. I think it's better to have single arrays for both and then filter them based on flags.

Very good point -- surprised I missed that. `_only(Suites|Tests)` combines the info on which are `only` with duplicate info on what are the children of the suite.

Re. `hasOnly`:

My concern with this is that it's also no longer a single source of truth. As far as I can tell, this is basically acting as a cache -- you could still search all the suites for `isOnly` (and if I read the code right the original is searching all the suites for `_only...`, so I don't think changing to `isOnly` makes performance any worse than it already is), you can just do it faster with the same info in multiple places (setting `isOnly` also setting `hasOnly` on the line of parentage to indicate the presence of `isOnly` further up the chain). But a sound cache, to my knowledge (and setting aside ones that are time-bound because staleness isn't a correctness issue), can detect staleness without the thing that changes the underlying data (`isOnly`) having to separately call on the cache to update -- in other words, it's automatic so it can't get out of sync. Furthermore, a sound cache doesn't let the cache data be updated independently of the underlying data (see the privacy issue below) -- which would be setting it out of sync almost by definition. (I imagine this could be solved by making the only property not a raw piece of public data but some kind of getter and setter where the setter would be hooked into the tree cache mechanism behind the scenes?) I'd be interested in knowing if there's evidence of performance issues before we get too sidetracked on figuring out a way to solve them without too much consistency risk.

>•_onlySuites and _onlyTests were treated as public even though by convention they were private

Not really; some of Mocha's modules are being treated like what C++ programmers call "friend classes": components so tightly coupled (whether rightly or wrongly) that they get access to each others' implementation details that are private to the rest of the program. It's a separate issue whether to imply to downstream users of the library that they should use this data exposed in this way and that if we need to make further changes to it they will wait for a "major" release.

For the record, I'm inclined to say that `only`ness *should* be publicly exposed in *some* way, but I'd have to think more about the right way. For instance: do we want to expose directly which tests and suites had `only` set, or only expose which tests and suites Mocha determines should run based on the `only` settings, or both? *And this compounds with the source-of-truth issue for `hasOnly`: if any part of the program with access to the suites can set their `hasOnly`, it can mislead the search algorithm.*

> If it's necessary to have isOnly and hasOnly private, one way of achieving this would be to move filtering logic into suites which might also be feasible solution.

It might improve the organization and reusability of the functionality/components in general if we can come up with a way to do this, depending on how it interacts with other issues (such as optimizing the search vs. avoiding redundant info on where the `only`s are), although I don't think it would be strictly necessary for this PR (in other words, feel free if it makes the other points easier to resolve, or feel free not to at this point if it does not).

> Refactoring easily tends to have effect to multiple issues.

Indeed, although it's always nice when it can be accomplished incrementally, to make sure each of the issues is addressed cleanly and correctly. I've found it's easier to push through big sets of changes -- and to be sure they're all right/good -- when there's a clear path of little changes from here to there.

Setting aside for the moment how we want the nested `only` logic to work, I would definitely be interested in getting in a change from `_only(Suites|Tests)` to `_isOnly` simply for sake of reducing the risk of error (assuming as I expect that no new performance issues are introduced by that), then from there move on to further optimize if need be, expose more publicly if appropriate, and make behavioral changes if they're desirable.

Thanks again for working with us on this! Definitely some good thoughts bubbling up here that will help improve things, whatever we decide on.

villesau:
> That's odd -- there isn't pending and passing too? If not, we should probably fix that; I can't imagine what good failure count does without the other two (and if we ever want to drop or replace them, we'd have to anyway if failures is already out there).


Actually runner seems to contain the info, but it's not received as an object if you look at ie. json.spec.js. It might also be possible to listen to events, but most convenient way in my opinion would be to pass stats to callback of .run method. I think it would be relatively easy to implement too.

villesau:
I did some testing with ~700 test cases and there is no anywhere near as big hit when iterating all the tests trough as i believed it would have. filtering 700 test cases takes less than 0.140 without .only tests and 0.8ms with some .only tests. with _hasOnly times was 0.088ms without .only tests and 0.220ms with some .only tests. So it takes ~4 times more to filter the tests, but it's still practically nothing.

villesau:
> Not really; some of Mocha's modules are being treated like what C++ programmers call "friend classes": components so tightly coupled (whether rightly or wrongly) that they get access to each others' implementation details that are private to the rest of the program. It's a separate issue whether to imply to downstream users of the library that they should use this data exposed in this way and that if we need to make further changes to it they will wait for a "major" release.


I'm not really familiar with C++ conventions so i find this pretty alien to me. I think this might be something like protected in java world.

> For the record, I'm inclined to say that onlyness should be publicly exposed in some way, but I'd have to think more about the right way. For instance: do we want to expose directly which tests and suites had only set, or only expose which tests and suites Mocha determines should run based on the only settings, or both? And this compounds with the source-of-truth issue for hasOnly: if any part of the program with access to the suites can set their hasOnly, it can mislead the search algorithm.


Good points! I don't have any strong opinions on this as i haven't used Mocha programmatically too much. Is there some roadmap, requirements or needs regarding mochas programmatic api? It seems to be a bit difficult to approach at the moment.

boneskull:
Please see my comments in #2606.  @villesau Thanks anyway--glad you're enthusiastic about the project.  Even though the maintainers may move pretty slowly, it's always a good idea to discuss these types of PRs before coding them to yourself some time.

Issue #2602, opened 2016-11-27
investigate jwt for saucelabs integration in PRs


boneskull:
Apparently, using json web tokens, we can run our browser tests (on SauceLabs) on pull requests.  Currently, we're not doing this, because security.

Not sure when Travis-CI introduced its [JWT addon](https://docs.travis-ci.com/user/jwt/), but we should be able to leverage it.

boneskull:
this will be closed by #2861

ScottFreeCode:
Did this end up happening in the Sauce launcher consolidation, or do we still need to give it a try?

boneskull:
@ScottFreeCode I haven't looked at JWT, no.

craigtaub:
This has surfaced again from tests passing on PR but causing master to fail. I can have a go.

juergba:
> The JWT addon is deprecated and will be discontinued on April 17, 2018.

PR #2603, opened 2016-11-27
Move Sauce key to JWT

ScottFreeCode:
I figure we'll test whether this works for a PR, then look at the branch and test whether it works for non-PR commits before merging into master, and if both those work then we can consider #2602 handled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ language: node_js
 
 env:
   global:
+    - SAUCE_USERNAME=mochajs
     # phantomjs hosts binaries @ bitbucket, which has fairly restrictive
     # rate-limiting.  pull it from this sketchy site in China instead.
     - PHANTOMJS_CDNURL='https://cnpmjs.org/downloads'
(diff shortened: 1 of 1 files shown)


ScottFreeCode:
Clearly I missed something, as this isn't getting all the Sauce credential stuff it's looking for: https://travis-ci.org/mochajs/mocha/jobs/179248317#L876

ScottFreeCode:
Hmm. Even with `SAUCE_USERNAME` separately set, it's still getting this: https://travis-ci.org/mochajs/mocha/jobs/179251815#L876 Maybe it needs the `SAUCE_ACCESS_KEY` alone re-encrypted (because the old encrypted setting contained both)?

dasilvacontin:
> Maybe it needs the SAUCE_ACCESS_KEY alone re-encrypted (because the old encrypted setting contained both)

Sounds likely!

boneskull:
@ScottFreeCode `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` is *also* in the [settings](https://travis-ci.org/mochajs/mocha/settings) which may be conflicting?

boneskull:
This is likely superceded by #2861.

PR #2611, opened 2016-11-30
Please, include me to the list of contributors

Rulikkk:
This update includes me to the list of contributors in package.json
Couple years ago I did this one: https://github.com/mochajs/mocha/pull/1043/files and I still see some of its' code in master.
Please let me be in contributors list.


dasilvacontin:
Weird, I thought we autogenerated this list, but in that case you should be there already. Any idea @boneskull?

boneskull:
@danielstjules It's auto-generated, but it's also a manual process which doesn't happen often.  I can't actually remember what I used to do it last time.

PR #2623, opened 2016-12-03
fix weird "pending" behavior; closes #2286

### LINKED ISSUE #2286: Regression after doesn't execute if tests skipped in beforeEach

boneskull:
continuation of #2571 with behavior suggested by @laughinghan.

this is targeted to `beta` branch for pre-release

### DIFF
--- a/test/integration/regression.spec.js
+++ b/test/integration/regression.spec.js
@@ -5,6 +5,7 @@ var fs = require('fs');
 var path = require('path');
 var run = require('./helpers').runMocha;
 var runJSON = require('./helpers').runMochaJSON;
+var map = require('../../lib/utils').map;
 
 describe('regressions', function () {
   it('issue-1327: should run all 3 specs exactly once', function (done) {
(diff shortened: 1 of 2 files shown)


boneskull:
@ScottFreeCode any further comments?

test/integration/regression.spec.js
```diff
@@ -54,17 +55,175 @@ describe('regressions', function () {
...
+       *   "afterAll" x 2.  There should be 10 items in this array.
+       * @param {string} [mode=always] - One of 'always' or 'once'
+       */
+      function testPendingRunnables (name, expected, mode) {
+        mode = mode || 'always';
+        var spies = [];
+
+        function spy (skip) {
+          function wrapped () {
+            if ((!wrapped.runCount++ || mode === 'always') && skip) {
```
fearphage: `wrapped.runCount++ === 0` would add a bit more clarity and make it very explicit what you're looking for.

test/integration/regression.spec.js
```diff
@@ -54,17 +55,175 @@ describe('regressions', function () {
...
+              return Boolean(spy.runCount);
+            })
+              .should
+              .deepEqual(expected);
+          });
+        });
+      }
+
+      testPendingRunnables('beforeAll', [
+        true, // beforeAll
```
fearphage: Booleans that map to calls are not intuitive. What if you used strings instead?

Possibly use `beforeAll` and `beforeAll-called-skip` (or something similar) to denote the state?

Then you wouldn't need comments to explain what's everything means.

test/integration/regression.spec.js
boneskull: Thanks for your thoughts!  

I usually just prefer to use `!` when a falsy check is adequate.  It's a common style, so I'm not too concerned about it.

Though, some prefer explicit addition (`+`) over increment operator(s).  This is more obvious for readers of JavaScript who are, say, unfamiliar with the difference between `foo++` and `++foo` (or even `+foo`).

But, if it's opaque, I'll find another way to skin this particular cat.

test/integration/regression.spec.js
boneskull: That's a good suggestion.  

This entire test is a pretty rotten hack, which I am unhappy with.  If anyone can think of a better solution--and/or implement the requisite harness(es)--I'd be grateful!

craigtaub:
@boneskull struggling to confirm if this is still useful. Could you confirm pls?

PR #2629, opened 2016-12-06
allow xunit reporter options to specify a title for the test groups

elangelo:


boneskull:
thanks

lib/reporters/xunit.js
```diff
@@ -65,7 +69,7 @@ function XUnit (runner, options) {
 
   runner.on('end', function () {
     self.write(tag('testsuite', {
-      name: 'Mocha Tests',
+      name: self.title ? self.title : 'Mocha Tests',
```
ScottFreeCode: If I'm not mistaken, this is equivalent to `self.title || 'Mocha Tests'`.

lib/reporters/xunit.js
```diff
@@ -51,6 +51,10 @@ function XUnit (runner, options) {
     self.fileStream = fs.createWriteStream(options.reporterOptions.output);
   }
 
+  if (options && options.reporterOptions && options.reporterOptions.title) {
+    self.title = options.reporterOptions.output;
```
dasilvacontin: I think you meant `options.reporterOptions.title` here?

dasilvacontin: Otherwise I'm missing context.

boneskull:
old

PR #2642, opened 2016-12-19
allow + in "grep" querystring; closes #2626

### LINKED ISSUE #2626: 3.x: url encoded grep not properly unescaped
HI,

Running `3.0.2`, in browser tests `grep` argument 
won t work properly with an url like `http://localhost:3000/test/?grep=CustomAjaxTable+should+update+header+class+on+sortable+header+click`
It will work if the spaces are not replaced by `+` sign,
`http://localhost:3000/test/?grep=CustomAjaxTable should update header class on sortable header click`

It s a bit of problem, my test need to manage url parameters, it uses the URL / URLSearchParams API (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), and it seems like to encode it in such way mocha won t follow up.

I gone through the changelog, look likes it was not fixed since the reported version.

boneskull:


boneskull:
eh, I'll merge it anyway.

dasilvacontin:
LGTM.

PR #2649, opened 2016-12-28
options.reporterOptions are used for progress reporter

canoztokmak:
Progress reporter had multiple options for configuring the reporter,
but it needed the optional fields under `options` object, while
those fields are provided under `options.reporterOptions`.
This commit enables usage of optional fields from `reporterOptions`.

### DIFF
--- a/lib/reporters/progress.js
+++ b/lib/reporters/progress.js
@@ -39,11 +39,13 @@ function Progress (runner, options) {
 
   // default chars
   options = options || {};
-  options.open = options.open || '[';
-  options.complete = options.complete || '▬';
-  options.incomplete = options.incomplete || Base.symbols.dot;
-  options.close = options.close || ']';
-  options.verbose = false;
+  var reporterOptions = options.reporterOptions || {};
+
+  options.open = reporterOptions.open || '[';
+  options.complete = reporterOptions.complete || '▬';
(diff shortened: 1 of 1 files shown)


canoztokmak:
solution for issue https://github.com/mochajs/mocha/issues/2661

xxczaki:
@boneskull i think this should be merged & the issue #2661 closed + this is a pretty old pr :/

boneskull:
old PR's happen

boneskull:
LGTM, thanks

boneskull:
@canoztokmak Can you please try this again?  It seems this introduced breaking changes

canoztokmak:
Hey @boneskull,

The reporter options are passed as a field inside `options`. See example below:
https://github.com/mochajs/mocha/blob/master/lib/reporters/xunit.js#L52

But in test, the options are passed directly. So the tests are validating the current behaviour, which is wrong. See below:
https://github.com/mochajs/mocha/blob/master/test/reporters/progress.spec.js#L99

There was no tests when I introduced the fix, that's why build did not failed back then. I think I can fix the tests and send another PR including with this change in a few days.

canoztokmak:
@boneskull I've just sent another PR to solve the issue.
https://github.com/mochajs/mocha/pull/3182

PR #2653, opened 2017-01-02
Update LICENSE

Scottkao85:
update copyright year


Munter:
I thought copyright notices were supposed to contain the year of the original publication of the work, plus comma separated amended years of new publications with alterations. Updating to the current year seems to me to void copyright of previous years, opening the project up to copyright claims from forks from previous years.

IANAL though, so I'd love some legal advice on this one

PR #2660, opened 2017-01-04
Double timeouts on integration tests

Munter:
The current timeout of 1000ms gives us a lot of false negatives on different node tests on Travis. The problem might be related to running lots of tests in parallel. Since this is hard to confirm on a local machine I have no scientific grounds for setting 2000ms specifically

### DIFF
--- a/test/integration/suite.spec.js
+++ b/test/integration/suite.spec.js
@@ -5,7 +5,7 @@ var run = require('./helpers').runMocha;
 var args = [];
 
 describe('suite w/no callback', function () {
-  this.timeout(1000);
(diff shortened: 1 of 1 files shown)


test/integration/suite.spec.js
```diff
@@ -5,7 +5,7 @@ var run = require('./helpers').runMocha;
 var args = [];
 
 describe('suite w/no callback', function () {
-  this.timeout(1000);
+  this.timeout(2000);
```
sul4bh: Probably pull this value out to a variable?

Munter:
Irony:

image

PR #2662, opened 2017-01-08
Replaced embedded unicode characters in html reporter with their hex encoded equivalents

### LINKED ISSUE #2283: Incorrect character in browser report
The HTML reporter is supposed to use ‣ for its grep buttons (&amp;#x2023; is the hex code I believe), but I'm getting this symbol instead: â€

rotemdan:
Fixes #2283.

The issue causes the HTML reporter to display garbled characters (`â€`) instead of the [triangular bullet character](http://graphemica.com/%E2%80%A3) (`‣`) unless the containing HTML page has explicitly set:
```html
<meta charset="utf-8" />
```

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -136,7 +136,7 @@ function HTML (runner) {
   runner.on('pass', function (test) {
     var url = self.testURL(test);
     var markup = '<li class="test pass %e"><h2>%e<span class="duration">%ems</span> ' +
-      '<a href="%s" class="replay">‣</a></h2></li>';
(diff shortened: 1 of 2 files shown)


Munter:
Thank you for contributing :D

dasilvacontin:
This bugfix has been released in [v3.3.0](https://github.com/mochajs/mocha/releases/tag/v3.3.0)! 🎉

TriMoon:
Wouldn't it have been better and more to the point to add that meta tag in the generated page's head?
Besides it already seems set in [Line5 of template.html](https://github.com/mochajs/mocha/blob/master/lib/template.html#L5) :confused:

PR #2674, opened 2017-01-18
chore: pin karma to 1.3.0

boneskull:


boneskull:
wow, this is pretty awesome if it works.  thanks @greenkeeperio!

PR #2675, opened 2017-01-18
limit browser concurrency on sauce; closes #2478 and #2667

### LINKED ISSUE #2478: address saucelabs flake
We're hitting some flake when running tests against SauceLabs; from time-to-time the browsers fail to launch, for whatever reason.  Investigate, fix, etc.

boneskull:
We need to pin Karma to v1.3.0 for whatever reason, and also pin the # of concurrent Karma sessions to 5.  This seems to imrpove the situation, though I can't guarantee it's "fixed".

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -174,6 +174,8 @@ function addSauceTests (cfg) {
     startConnect: true
   };
 
+  cfg.concurrency = 5;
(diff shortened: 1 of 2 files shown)


Munter:
I'll take it :)

Issue #2628, opened 2016-12-05
Support configurable name for XUnit reporter


ngeor:
The XUnit reporter produces an XML report that has a hard-coded name 'Mocha Tests'. In the XML this appears as: `<testsuite name="Mocha Tests"`. When using mocha for various types of tests in the same CI build, it would be great to be able to specify this name to something more descriptive, e.g. "Unit Tests", "Integration Tests", etc.

PR #2676, opened 2017-01-19
Fixing issue #2628 allowing to override the default test suite name of…

ngeor:
Fixing issue #2628 allowing to override the default test suite name of the xunit reporter.

### DIFF
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -43,13 +43,28 @@ function XUnit (runner, options) {
   var tests = [];
   var self = this;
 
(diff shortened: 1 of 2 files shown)


boneskull:
this looks good.  it should go into v3.6.0 or v4.0.0, whichever comes first.

boneskull:
@ngeor thanks!

ScottFreeCode:
Should we update the assertion library in this to be the same one used by the existing tests for this reporter, or should we merge and update to use the same assertion library afterward?

PR #2677, opened 2017-01-19
in favor of 'istanbul report' over wrapper

a0viedo:
Is there any specific reason for keeping `istanbul-combine`? For what I can see the results are the same.


a0viedo:
Hey @probot-stale[bot], I appreciate you care about this. I don't know if this has changed in master but I really think it's an improvement. Could you advice how can I get more attention?

a0viedo:
thanks @stale, this is no longer relevant

PR #2680, opened 2017-01-22
Increase tests coverage for base reporter

epallerols:
I am adding 4 new test scenarios for the base reporter.

Coverage diff:
old coverage

new coverage

There is one test that covers a specific chaijs scenario related to this [lines](https://github.com/mochajs/mocha/blob/master/lib/reporters/base.js#L212-L213), it was introduced in this [PR](https://github.com/mochajs/mocha/commit/34116d5d67b75fd189637cc773ce550ba956bc36), since I didn't want to add an extra dev dependency to cover this single case I added a comment trying to clarify it.

There is also some clean up that can be done ([see this PR](https://github.com/mochajs/mocha/pull/2681) ) on that file that could make the code cleaner and the coverage increase, but I didn't want to add to many things in one PR.

Hope it helps.

### DIFF
--- a/test/reporters/base.spec.js
+++ b/test/reporters/base.spec.js
@@ -113,6 +113,53 @@ describe('Base reporter', function () {
     });
   });
 
+  describe('Inline strings diff', function () {
+    it('should show single line diff if property set to `true`', function () {
+      var err = new Error('test');
+      var errOut;
+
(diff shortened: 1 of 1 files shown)


Munter:
Thank you for contributing!

PR #2682, opened 2017-01-23
Creating "Step" reporter

SpaceMacGyver:
This PR attempts to create a reporter that is largely similar in nature to the existing spec reporter, but with output that includes hooks.  The primary motivation is to add value to tests whose hooks have significant meaning and whose output are intended to serve as documentation.

## Rationale
The main use case for this is [Gherkin](https://github.com/cucumber/cucumber/wiki/Gherkin)-like structure for tests.  The various cucumber-esque interfaces and reporters that have been written to address this space make heavy use of the `it()` function to result in meaningful rendered output, which:
* seems to be a misuse of assertion clauses, and
* is an unfortunate side effect of no existing mechanisms for displaying hooks being built into Mocha.

To illustrate the difference, consider a [sample project](https://gitlab.com/bellwether-open-source/pattern-matcher/blob/master/test/matcher.js) whose tests are written with Gherkin syntax in mind when providing descriptions:

**using spec reporter**
```
  Feature: Pattern matcher
    Scenario: Matcher invoked as function with no args
      √ then should evaluate to undefined
    Scenario: Matcher provided default resolution
      √ then should evaluate to expected default value
    Scenario: Matcher provided default resolution against value
      √ then should evaluate to expected default value
    Scenario: Matcher provided matcher rules
      √ then should evaluate to expected default value
    Scenario: Matcher evaluated against multiple args
      √ then should evaluate to expected default value
```

**using step reporter**
```
    Feature: Pattern matcher
      Scenario: Matcher invoked as function with no args
        "before all" hook: when matcher invoked
        "before all" hook: when matcher evaluated
        √ then should evaluate to undefined
      Scenario: Matcher provided default resolution
        "before all" hook: given default evaluation handler
        "before all" hook: when matcher invoked with default handler
        "before all" hook: when matcher evaluated
        √ then should evaluate to expected default value
      Scenario: Matcher provided default resolution against value
        "before all" hook: given default evaluation handler
        "before all" hook: when matcher invoked with default handler
        "before all" hook: when value evaluated by matcher
        √ then should evaluate to expected default value
      Scenario: Matcher provided matcher rules
        "before all" hook: given matching rule
        "before all" hook: given non-matching rule
        "before all" hook: when matcher invoked with rules
        "before all" hook: when value evaluated by matcher
        √ then should evaluate to expected default value
      Scenario: Matcher evaluated against multiple args
        "before all" hook: given matching rule
        "before all" hook: when value evaluated by matcher
        √ then should evaluate to expected default value
```

### DIFF
--- a/lib/reporters/index.js
+++ b/lib/reporters/index.js
@@ -16,4 +16,5 @@ exports.XUnit = exports.xunit = require('./xunit');
 exports.Markdown = exports.markdown = require('./markdown');
 exports.Progress = exports.progress = require('./progress');
 exports.Landing = exports.landing = require('./landing');
+exports.Step = exports.step = require('./step');
 exports.JSONStream = exports['json-stream'] = require('./json-stream');
--- a/lib/reporters/step.js
+++ b/lib/reporters/step.js
@@ -1,0 +1,55 @@
+'use strict';
+
+var Base = require('./base');
+var inherits = require('../utils').inherits;
+
+function Step (runner) {
+  Base.call(this, runner);
+
+  var self = this;
+  var suiteStack = [];
+  var failingTestCount = 0;
+
+  function indent (offset) {
+    var indentCount = (offset || 0) + suiteStack.length;
+
+    return Array(indentCount).fill('  ').join('');
+  }
+
+  runner.on('suite', function (suite) {
+    suiteStack.push(suite);
+    console.log(indent() + suite.title);
+  });
+
+  runner.on('suite end', function () {
+    suiteStack.pop();
+  });
+  runner.on('hook', function (hook) {
+    console.log(indent(1) + hook.title);
+  });
+  runner.on('pass', function (test) {
+    var fmt;
+
+    if (test.speed === 'fast') {
+      fmt = indent() +
+        Base.color('checkmark', '  ' + Base.symbols.ok) +
+        Base.color('pass', ' %s');
+      console.log(fmt, test.title);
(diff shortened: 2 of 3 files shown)


boneskull:
Hi, thanks for the PR.  We're not currently merging new feature PRs--but you can help us get back on the rails.  See our [README](https://github.com/mochajs/mocha/blob/master/README.md) for more info

SpaceMacGyver:
Thanks @boneskull -- I can utilize my fork for personal development if I get desperate enough, and will leave this in the queue in the meantime until it receives the opportunity for review.

Munter:
@SpaceMacGyver This looks quite useful. I recommend you release this reporter as a separate project, which would allow other users to pull it in on their own while still staying on the official mocha releases

Munter:
Would the same functionality also be covered by https://www.npmjs.com/package/mocha-steps ?

SpaceMacGyver:
@Munter Looking at that suggestion, it seems like rather than making hooks useful it removes them entirely from the equation.  If you consider its output, every step is a "test" (even if no verification is taking place), and while this does seem to be a popular solution (e.g. [mocha-gherkin](https://github.com/mklabs/mocha-gherkin/blob/master/lib/parser.js#L88)), the checkmarks add noise and obscure the system under test.  It is a solution, though, that many people seem to think acceptable.

boneskull:
again this is more appropriate as an external reporter.

PR #2698, opened 2017-02-02
Increase tests coverage for xunit reporter

craigtaub:
I am adding 11 new test scenarios for the XUnit reporter. 
The only branch which is missing is the falsy of the `attr` iterations property check here https://github.com/mochajs/mocha/blob/master/lib/reporters/xunit.js#L158 (which does not seem possible to catch).

Coverage diff:
screen shot 2017-01-31 at 17 55 27
screen shot 2017-02-02 at 18 14 58

Hope it helps

### DIFF
--- a/test/reporters/xunit.spec.js
+++ b/test/reporters/xunit.spec.js
@@ -107,7 +107,7 @@ describe('XUnit reporter', function () {
   });
 
   describe('done', function () {
-    describe('if fileStream is truthly', () => {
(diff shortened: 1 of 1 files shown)


Munter:
Tahnk you!

plroebuck:
https://github.com/mochajs/mocha/blame/master/test/reporters/xunit.spec.js#L196

Curious how this even works with current codebase.
I see error removing 'uncaught' handler

PR #2700, opened 2017-02-03
Allow prepending hooks beforeAll/afterAll/beforeEach/afterEach

carsonreinke:
A `prepend` argument allows for a hook to placed in front of the list.

Thoughts?  Let me know.

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -31,8 +31,8 @@ module.exports = function (suites, context, mocha) {
      * @param {string} name
      * @param {Function} fn
      */
-    before: function (name, fn) {
-      suites[0].beforeAll(name, fn);
+    before: function (name, fn, prepend) {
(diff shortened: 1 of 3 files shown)


Munter:
Since you've obviously gone through a lot of trouble to implement this, maybe you could also spend a bit of time trying to clarify the use case that fuels the need for this change? I'm having a hard time understanding what this is about

carsonreinke:
@Munter this is in case you need to push a hook on to beginning of the list.  Came across this issue using a set of shared tests, but a particular test case required a `beforeEach` hook ahead of the shared hooks.

Maybe it is better to just allow access the array of hooks instead.

boneskull:
> @Munter this is in case you need to push a hook on to beginning of the list

Why would this be necessary?

boneskull:
inactive

PR #2705, opened 2017-02-08
cover to-iso-string with tests

seppevs:


ScottFreeCode:
Since Mocha 4 dropped support for older environments, we are no longer using this shim (:tada:).

PR #2708, opened 2017-02-10
Added feature to xit that allows someone to chain 'only' onto it

wi-ski:
I really wanted this stupid change so I made quick stab at it to get the idea across.

### DIFF
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -106,6 +106,12 @@ module.exports = function (suite) {
     };
 
     /**
+     * Because willdembinski really, really wanted it.
+     */
+
+    context.xit.only = context.xit;
+
+    /**
      * Number of attempts to retry.
      */
     context.it.retries = function (n) {


Munter:
What problem is this PR intending to solve?

willdembinski:
Entirely trivial.  But I would really like to be able to do:

```
it.only('description text',function(){})
xit.only('description text',function(){})  <- Have this be quiet-ed.
xit('description text',function(){})
```

It comes up frequently enough (In my life at least) to have wanted to suggest it as a feature.

Munter:
The semantics of using multiple `.only` calls are a bit broken. In general I think of putting any of these modifiers in your test cases to single our or exclude test is an anti pattern, since people might accidentally check them in. Have you tried the `--grep` option to isolate tests?

If we weren't interested in merging this specific PR, would a switch around of priority between `skip` and `only` solve the use case you run into? If `skip` always took precedence I assume you wouldn't run into these problems either

willdembinski:
In my mind, this new behavior wouldn't be related to the semantics of `.only()` or `.skip()`. It would *only* attach attributes to `xit()` that are essentially noOps (They just defer they behavior to the original behavior of `xit()`)

Ultimately, `xit()` takes precedence over any behavior that could result from `.only()` or `.skip()`. This change wouldn't effect the actual behavior of `.only()` or `.skip()` even if their semantics change in the future in the library at large.

The `--grep` simply takes longer (Agreed, this is pure laziness talking) than pressing the "Home" key and "x"-ing out the test case.

TL;DR: I wouldn't want the priority of `skip` or `only` to be considered with this PR because `xit.only()` or `xit.skip()` should only be concerned with the behavior of `xit()`.

wi-ski:
@Munter - How might I improve this PR? ^^

Munter:
I don't think this belongs in core. The crazy edge cases are already far to many when people start using `xit`, `iit`, `it.skip` and `it.only`. I don't want to support even more pitfalls with combinations.

I already see the use of most of these as anti patterns because they might slip into the code base. Combinations are even worse.

If you feel strongly for this, extend `it.xit` in your local code base or release a separate interface that you can use instead of bdd.

Issue #2713, opened 2017-02-15
Mocha exits with 0 exit code with failing tests


chrisleck:
Node: 7.2.1
Mocha: 3.2.0
MacOSX: Sierra (10.12.3)
Cmdline: `mocha my_test.js`

Seeing mocha exit with a 0 exit code after failing a test in an `after each` hook.  I traced it down to this code: https://github.com/mochajs/mocha/blob/9ae6a85b6ba50ea415affc88456d469d6915a157/bin/_mocha#L425

The function is called with the correct exit code, but then `stream.write('', done);` doesn't call the callback before the process terminates.

chrisleck:
I added some tracing statement to the code as follows and see the following output:
```
function exit (code) {
  // flush output for Node.js Windows pipe bug
  // https://github.com/joyent/node/issues/6247 is just one bug example
  // https://github.com/visionmedia/mocha/issues/333 has a good discussion
  function done () {
    draining--;
    console.log(`Draining down to ${draining}`);
    if (draining <= 0) {
      process.exit(Math.min(code, 255));
    }
  }

  process.on('exit', function(realExitCode) {
    console.log(`Process is exiting with ${realExitCode} but should exit with ${code}.`);
  });

  var draining = 0;
  var streams = [process.stdout, process.stderr];

  streams.forEach(function (stream) {
    // submit empty write request and wait for completion
    draining += 1;
    console.log(`Draining up to ${draining}`);
    stream.write('', done);
  });

  console.log('Starting extra call to done().');
  done();
  console.log('Extra call to done() finished.');
}
```
Results in:
```
Draining up to 1
Draining up to 2
Starting extra call to done().
Draining down to 1
Extra call to done() finished.
Process is exiting with 0 but should exit with 1.
```

chrisleck:
Proposal: eagerly set the `process.exitCode` property and eliminate the extraneous call to `done`.  Will send a PR.

chrisleck:
FWIW, I'm also able to reproduce this in a Linux container running on CoreOS with the same Node and Mocha versions and same test.

percyhanna:
Running into this issue on our CI running on Travis. Tests fail, but Mocha exits with `0`, so the build does not fail.

dasilvacontin:
The `Eagerly set exitCode` fix has been released in [v3.4.2](https://github.com/mochajs/mocha/releases/tag/v3.4.2)! 🎉 

Sorry it took long!

olastor:
[...]

PR #2714, opened 2017-02-15
Fix for issue #2713

chrisleck:
* Eagerly set the process exitcode in case the write callback isn't executed before the process terminates.
* Remove the extraneous call to `done()` at the bottom of the function.
* Cleanup the `draining` reduction and value check in `done` into separate lines for easier reading.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -466,12 +466,19 @@ function exitLater (code) {
 }
 
 function exit (code) {
+  var clampedCode = Math.min(code, 255);
+
+  // Eagerly set the process's exit code in case stream.write doesn't
+  // execute its callback before the process terminates.
+  process.exitCode = clampedCode;
+
   // flush output for Node.js Windows pipe bug
   // https://github.com/joyent/node/issues/6247 is just one bug example
   // https://github.com/visionmedia/mocha/issues/333 has a good discussion
   function done () {
-    if (!(draining--)) {
-      process.exit(Math.min(code, 255));
+    draining--;
+    if (draining <= 0) {
+      process.exit(clampedCode);
     }
   }
 
@@ -483,8 +490,6 @@ function exit (code) {
     draining += 1;
     stream.write('', done);
   });
-
-  done();
 }
 
 process.on('SIGINT', function () {


Munter:
Is is possible to consistently recreate the issue from #2713 ? If it is, a test case to guard against regresssions would be helpful

chrisleck:
I wasn't able to create a standalone test case. Seems timing dependent in node.

chrisleck:
Ping for action?

chrisleck:
Can we get this PR approved or follow up with review comments?

chrisleck:
Ping-a-ling-a-ping-ping

chrisleck:
@Munter Sorry for the direct ping.  Mind taking another look?

Munter:
I understand the value of eagerly setting the exitCode with `process.exitCode = ...`. That part seems good to me.

Without tests I'm not sure what the bigger consequences are of removing the `done()` call at the end

chrisleck:
@Munter What's your concern about the `done()` call at the end?  The existing code is pretty clearly unbalanced - each stream instance adds one to `draining` and each callback removes one.  The `done()` call at the end causes `draining` to reach 0 before the second stream has called `done()`, thereby defeating the purpose of this ref-counting mechanism.

An alternative to my approach for making the system balanced is to initialize `draining` to 1 such that the call to `done()` at the end of the method is the third callback needed to get `draining` down to 0. That may be even safer than my approach and I'm happy to make that change if you'd like.  Please advise.

chrisleck:
@Munter waiting for your response.

chrisleck:
@Munter ping for response.

ScottFreeCode:
Why not just address the setting of `process.exitCode` and the correctness or incorrectness of the draining logic (involving the `done()` call) in separate issues/PRs?

(Actually, we should double-check whether the draining logic's `done` calls have come up before -- we might already have an issue about investigating it. But in any case, there's benefit to setting `process.exitCode` once we know how we're going to exit *even if we don't touch the draining logic*, so...)

chrisleck:
@ScottFreeCode That sounds reasonable.  I'll send out a new PR shortly.

dasilvacontin:
> whether the draining logic's done calls have come up before 

@ScottFreeCode can you elaborate?

The logic rewrite looks good to me.

Regarding that extraneous `done` call, take a look at the old condition:

```js
if (!(draining--)) {
```

The condition will be true when the value of `draining` is `0` before hitting the `if`. `draining--` returns then current value of the variable and _then_ it decrements the variable.

So the logic was correct. But laid out in a very confusing way.

ScottFreeCode:
Thanks, that's what I remember discussing (or at least looking at) before... The parentheses make it particularly confusing because normally they're used to force things to happen in a certain order, but in this case we've got a rare instance of C/C++-style variable-value separation that leaked into JS and even the parentheses don't affect that. (I guess they ensure that the decrement is applied to the variable and not to the result of the negation?)

So it's equivalent to:
```js
if (!drain) {
  ...exit...
}
drain -= 1
```
...and not to:
```js
drain -= 1
if (!drain) {
  ...exit...
}
```

And so we don't need to change it for correctness... But we really ought to change it for clarity.

In fact, I would love to see an ESLint rule against postincrement/postdecrement operators. (Wouldn't even say no to such a rule against preincrement/predecrement as well, considering command-query separation ought to be respected if you're going to be using mutable variables in the first place...)

boneskull:
This was fixed via #2820

PR #2715, opened 2017-02-16
add gdb support to mocha

ObiWahn:
Hi,

I am developing a c++ and nodejs driver (using Nan) for a database. There was a segmentation fault in my bindings and nodejs crashed without a proper trace or other useful information. Our js guru was not able to isolate the error (I personally guess he was too lazy to switch the tests off one by one). So it seemed useful to me to have the option to start node with an attached debugger (even over core files). The js guy had less work and I was able to find the bug in the native code in about 5 seconds.

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -11,6 +11,7 @@ var spawn = require('child_process').spawn;
 var path = require('path');
 var getOptions = require('./options');
 var args = [path.join(__dirname, '_mocha')];
+var gdb = false;
 
 // Load mocha.opts into process.argv
 // Must be loaded here to handle node-specific options
@@ -47,6 +48,9 @@ process.argv.slice(2).forEach(function (arg) {
     case '--perf-basic-prof':
       args.unshift(arg);
       break;
+    case '--gdb':
+        gdb = true;
+        break;
     default:
       if (arg.indexOf('--harmony') === 0) {
         args.unshift(arg);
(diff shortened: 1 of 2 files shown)


bin/mocha
```diff
@@ -65,7 +69,14 @@ process.argv.slice(2).forEach(function (arg) {
   }
 });
 
-var proc = spawn(process.execPath, args, { stdio: 'inherit' });
+var binary = process.execPath;
+if (gdb) {
+  args.unshift(binary);
+  args.unshift('--args');
+  binary = '/usr/bin/gdb';
```
ScottFreeCode: Is GDB guaranteed to be installed at `/usr/bin`? In every distro? When running on Mac via MacPorts or Homebrew? When running on Windows via MinGW or Cygwin? As a rule, it would probably be safer to assume that it's on the path and just use the executable name, I think. (This is a moot point if we don't add code specifically for GDB, but I want to throw it out there for reference.)

Munter:
Debugging is already supported with the `--debug` and `--inspect`. I don't feel that we have the man power to add support for more specific debugging tools that don't run on the remote debugging concept and possibly require platform specific hacks and caveats.

If you want gdb to work with mocha I'd definitely recommend the approach Scott already outlined with wrapping around the mocha binary, like istanbul for example already does

PR #2722, opened 2017-02-25
Add observable support with `symbol-observable`

### LINKED ISSUE #1898: Observables
So there's this cool thing called `Observable`s that is (likely) coming to the language soon ([ref 1](https://github.com/zenparsing/es-observable), [ref 2](https://github.com/tc39/ecma262)).

They are similar to promises in a number of ways.
- They both attempt to deal with asynchronous actions in a reasonable manner.
- They both define a standard interface that can be implemented and extended by third parties.

Since Mocha supports returning promises in order to avoid using the `done` callback, it'd be nice if it did the same thing with observables:

``` js
it('should do something with an Observable', () => {
  return Observable.of(1, 2, 3).subscribe(...);
});
```

People are already using these with http://reactivex.io/.

cc @blesh

mAAdhaTTah:
Handles returned observables the same way we handle returned promises.
Use the shared `symbol-observable` library for interop and consume the
stream until it ends. If an error is emitted, call callback with error object
from stream.

Fixes #1898.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -6,6 +6,7 @@
 
 var EventEmitter = require('events').EventEmitter;
 var JSON = require('json3');
+var $$observable = require('symbol-observable').default;
 var Pending = require('./pending');
 var debug = require('debug')('mocha:runnable');
 var milliseconds = require('./ms');
@@ -355,6 +356,19 @@ Runnable.prototype.run = function (fn) {
         function (reason) {
           done(reason || new Error('Promise rejected with no or falsy reason'));
         });
+    } else if (result && typeof result[$$observable] === 'function') {
+      self.resetTimeout();
+      var observable = result[$$observable]();
+
+      observable.subscribe({
+        error: function (reason) {
+          done(reason || new Error('Observable errored with no or falsy reason'));
+        },
+
+        complete: function () {
+          done();
+        }
+      });
     } else {
       if (self.asyncOnly) {
         return done(new Error('--async-only option in use without declaring `done()` or returning a promise'));
--- a/package.json
+++ b/package.json
@@ -316,7 +316,8 @@
     "json3": "3.3.2",
     "lodash.create": "3.1.1",
     "mkdirp": "0.5.1",
-    "supports-color": "3.1.2"
+    "supports-color": "3.1.2",
+    "symbol-observable": "^1.0.4"
   },
   "devDependencies": {
     "assert": "^1.4.1",
--- a/test/runnable.spec.js
+++ b/test/runnable.spec.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var $$observable = require('symbol-observable').default;
 var mocha = require('../');
 var utils = mocha.utils;
 var Runnable = mocha.Runnable;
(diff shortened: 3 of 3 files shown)


mAAdhaTTah:
Bumping this. Would love to get Observable support in Mocha, similar to how Promises are handled. Anything I can do to help?

boneskull:
@mAAdhaTTah Thanks for the PR.  

Even though #1898 was created in 2015, observables are still in [stage 1](https://github.com/tc39/proposals).

Until the proposal reaches stage 3 (or 4 if we're cautious; see [process document](https://tc39.github.io/process-document/)), adding explicit support for them is, IMO, jumping the gun.

It's likely this will also break compatibility with older browsers that we still support.  It'd be a good reason (among many) *to* break compatibility with those browsers--when the feature is ready.

That's my take on this; if anyone else has another opinion, please speak up!  

cc @mochajs/core

* * *

As an aside, from an interface perspective, I'm interested in how best to handle observables which never complete.  It wouldn't make much sense to `return` one, since your test will timeout, though it's still valuable to make assertions against them.

Given that *some* observables could be returned and *others* could not, does it make sense to add support for this at all?  A `Promise` either fulfills or it doesn't.  Not so with observables.

mAAdhaTTah:
> Until the proposal reaches stage 3 (or 4 if we're cautious; see process document), adding explicit support for them is, IMO, jumping the gun.

Fair point. I thought I had read somewhere it was moving to stage 2 soon, but I don't recall. Even still, we're still a ways off from Observables getting specced.

> It's likely this will also break compatibility with older browsers that we still support.

It shouldn't; `symbol-observable` is a polyfill, so it checks if `Symbol` exists in the environment and falls back to a plain string should it not be found.

> A Promise either fulfills or it doesn't. Not so with observables.

Strictly speaking, you can write a Promise that never resolves:

```js
const myPromise = new Promise((resolve, reject) => {
    // do nothing
});

myPromise.then(() => {
    console.log('Will never get called');
});
```

But yes, you're much more likely to have an Observable that never ends than a Promise that never resolves. That said, one of the things I like about Observables is you can use something like `take` to only get the next X number of values out of it, which provides a kinda, sorta `plan` for your tests (although it won't help if the Observables accidentally produce extra values).

Other than that, I don't think you _can_ make assertions against an Observable that doesn't complete, or at least, I'm don't think Mocha can provide any special handling for it, similar to just timing out when a Promise fails to resolve.

Hope that addresses some of your concerns. If you'd prefer to wait until Observables reach stage 3+, I'll keep this PR open and look at providing this functionality through a library similar to how `mocha-as-promised` handled it.

mAAdhaTTah:
I'm going to close this. My approach to testing with observable is to use spies on `next` & `error` and verify the values in `complete`, where I call `done` manually. This provides better test planning than `take`, as extra values are allowed to flow through without ending the stream and ensures you're testing for the exact values you expect.

Also, most Observable implementations have a `toPromise` method, if you prefer to return the Observable, so there isn't really a good reason to add support to Mocha directly.

ORESoftware:
maybe someone could create `Observable.to(Promise)`, then return that in mocha hooks? Not sure if that's possible, but maybe.

PR #2727, opened 2017-03-04
Please make the build reproducible.

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that node-mocha could not be built reproducibly.

 [0] https://reproducible-builds.org/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ else
 endif
 
 TM_BUNDLE = JavaScript\ mocha.tmbundle
-SRC = $(shell find lib -name "*.js" -type f | sort)
+SRC = $(shell find lib -name "*.js" -type f | LC_ALL=C sort)
 TESTS = $(shell find test -name "*.js" -type f | sort)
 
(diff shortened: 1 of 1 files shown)


ScottFreeCode:
Is it by any chance possible to test this? Seeing as we'd like to get rid of the makefile, it'd be handy to know that we're not undoing the fix here in the process of replacing it.

That being said, the change itself looks good to me; I think we can merge this as soon as we get the CLA confirmation.

lamby:
> Is it by any chance possible to test this?

Not easily without, for example, our special filesystem that returns shell globs deliberately wonky to surface issues.


Not easily without, for example, our special filesystem that returns shell globs deliberately wonky to surface issues.

> Seeing as we'd like to get rid of the makefile, it'd be handy to know that we're not undoing the fix here in the process of replacing it.

As long as you sort the inputs, you should be fine.

> CLA confirmation.

On it...

ScottFreeCode:
Thanks!

Issue #2298, opened 2016-06-08
Suite `only` not anchored to title start


ScottFreeCode:
In the process of looking into https://github.com/mochajs/mocha/issues/2157#issuecomment-224475198, I noticed that the BDD interface's `it.only` escapes the regular expression and requires it to match the whole title with `^` and `$`, but `describe.only` merely passes the string to `grep`, which escapes it but does not require it to begin with `^`. This same issue also affects the QUnit and TDD interfaces.

Contrived example:
```js
describe.only("one", function() {
  it("two", function() { /* succeed */ })
})
describe("you know", function() {
  it("one of these should not run", function() {
    throw new Error("Specifically, this one should not run!")
  })
})
```

ScottFreeCode:
Something similar goes for the links in the HTML reporter; the test ones should use `^` and `$` to require matching the whole title, and the suite ones should use `^` to require matching the start of the title.

boneskull:
@ScottFreeCode I merged bbeeff09c2dff0790b7358d4cf37b57661c04ce9 into v3.0.0 branch.  does this solve the issue?

ScottFreeCode:
Sorry I missed this!

Making `only` programmatic resolved the primary issue here.

It looks like the HTML links still have the same sort of problem, however. Do we have another issue for it? If not, I will open one instead of keeping this issue open for it.

PR #2728, opened 2017-03-05
Fix/html reporter

### LINKED ISSUE #2300: HTML reporter grep links do not respect fgrep, handle existing ? wrong
The grep links in the HTML reporter replace an existing `grep=...` parameter in the URL, but not `fgrep=...`. Furthermore, they tend to append an extra `?` if `?fgrep=...` is already in the URL instead of appending with `&` in that case.

Note that this code is in the same area as #2285.

ScottFreeCode:
Fixes potential issues with unescaped text, including in the querystring used as part of the grep links.

Also fixes #2300 and fixes the HTML half of #2298

Fixing #2285 is tricky. Per #1464, we can't just take the pathname out. There are a lot of ways to manipulate this, but all of them have some weird problem involved... with one possible exception: using the href and just dropping the search/hash from it.

```js
window.location.href.replace(new RegExp('^(.*?' + escapeRe(window.location.pathname) + ').*$'), '$1')
// or
window.location.href.replace(new RegExp('^(.*)' + escapeRe(window.location.search + window.location.hash) + '$'), '$1')
```


ScottFreeCode:
What? No, seriously -- what? I change `"` to `'` and coverage drops over one whole percent?? Is there some massive element of nondeterminism in play here or something?

boneskull:
@ScottFreeCode looks like you just volunteered to write tests against the html reporter :wink:

ScottFreeCode:
@boneskull

> @ScottFreeCode looks like you just volunteered to write tests against the html reporter :wink:

I just had a brainwave that I could do that fairly easily:
- in a test file that is only run in Karma
- have the tests create the `<div id="mocha">` element (if one already exists, temporarily save it and change its ID)
- then create a new instance of the HTML reporter with a dummy runner and emit events on the dummy runner
- restore anything that was changed re. divs and IDs (this could be additionally put in an `afterEach` hook in case anything goes wrong during the preceding parts of the test)
- assert on the temporary, test-created id="mocha" div having been correctly populated by the reporter

Preliminary experimentation indicates not only that this should be doable, but that it is way, way easier than any idea I've had for it in the past (e.g. using JSDOM or running multiple instances of Mocha somehow). I would actually like to finish this at some point...

The downside? I have no idea how to get a coverage report out of our browserifying Karma setup, and we are not currently set up to send Coveralls coverage reports from more than one CI job (although if we got that figured out, we could check coverage in a wider variety of environments and get rid of a lot of missed branches in the code that are handling different environments; e.g. the `to-iso-string` module is the other least covered module besides the HTML reporter, and it's uncovered because it's only used if date objects lack a built-in `toISOString` function; additional "although": it seems like this should be able to be set up either using the `istanbul-combine` command we already have or else using [Coveralls' "parallel" support](https://coveralls.zendesk.com/hc/en-us/articles/203484329)).

On the other hand, the fact that the HTML reporter is non-covered on two counts (lack of tests for it and lack of running in Node even if we had tests for it) strongly suggests that any affect this PR has on coverage is coincidental and/or non-meaningful.

(In fact, I'm not sure why its coverage counts aren't just a big flat bunch of 0s. Do we have somewhere that is loading, but not using, the file in Node somewhere in the tests?? It might be nice to get proper browser coverage and figure out for sure that there isn't already something trying to test this, before I get too thorough with my new tests...)

PR #2746, opened 2017-03-17
Make sure `mocha -h` consistently shows help

### LINKED ISSUE #2745: `mocha -h` doesn't always display help
Pretty straightforward testcase:

```
zarel@Phenylalanine ~> mocha -h

  Usage: mocha [debug] [options] [files]

  [...]

zarel@Phenylalanine ~> mkdir test
zarel@Phenylalanine ~> echo "test" > test/mocha.opts
zarel@Phenylalanine ~> mocha -h
No test files found
```

In other words, if `test/mocha.opts` exists, `mocha -h` won't display help.

Zarel:
Previously, having options in `test/mocha.opts` would add those to `process.argv` to be passed to `program.parse`, which would make it stop showing help.

Fixes #2745


Zarel:
The problem is when `program.parse` is called at [line 188 of `bin/_mocha`](https://github.com/mochajs/mocha/blob/b4ebabd1b76c02cbadd544544fb4b171a84ea698/bin/_mocha#L189):

```
program.parse(process.argv);
```

`process.argv` has been filled with the contents of `test/mocha.opts`

```
['/usr/local/bin/node', '/Users/zarel/Workspace/mocha/bin/_mocha', 'test', '-h']
```

`commander` seems to only show help if `process.argv` is exactly

```
['/usr/local/bin/node', '/Users/zarel/Workspace/mocha/bin/_mocha', '-h']
```

ScottFreeCode:
This looks good to me; @mochajs/core do we want to add a test for this behavior?

dasilvacontin:
There should be, imo.

boneskull:
@Zarel Thank you for the PR.  Would you be willing to add a test (probably somewhere [in here](https://github.com/mochajs/mocha/blob/master/test/integration/options.spec.js))?

Zarel:
@boneskull, I tried. I'm not sure if I did it the way you'd prefer, but this is definitely a test that fails without this commit and passes with it.

Zarel:
I'm not sure why the test failed in AppVeyor. Does anyone have advice?

ScottFreeCode:
I can't see the actual failure in the AppVeyor test either...

Travis is only failing the new test on the oldest version of Node. Unfortunately, its not too clear *why* it's failing (except that it's the assertion that failed) since the assertion library used in that test file is boolean-based. It may be necessary to switch it to one of the other assertion libraries used in Mocha's tests that supports a `contains` assertion, in order to find out what is wrong there.

Zarel:
@ScottFreeCode can you do that? I don't really know how to.

test/integration/options.spec.js
```diff
@@ -180,4 +183,28 @@ describe('options', function () {
       });
     });
   });
+
+  describe('--help', function () {
+    it('works despite the presence of mocha.opts', function (done) {
+      var output = '';
```
kunagpal: This initialization is redundant since an implicit casting is being done on line 195.

test/integration/options.spec.js
```diff
@@ -180,4 +183,28 @@ describe('options', function () {
...
+      try {
+        fs.mkdirSync(path.resolve(__dirname, 'test-env'));
+        fs.mkdirSync(path.resolve(__dirname, 'test-env/test'));
+        fs.writeFileSync(path.resolve(__dirname, 'test-env/test/mocha.opts'), 'foo');
+        var mochaLoc = path.resolve(__dirname, '../../bin/mocha');
+        output = '' + childProcess.execSync(mochaLoc + ' -h', {cwd: path.resolve(__dirname, 'test-env')});
+      } catch (e) {}
+      try {
```
kunagpal: The sequence of `try...catch` blocks here are for setup and teardown, and should be shifted to `before` and `after` blocks, outside this `it` block.

test/integration/options.spec.js
```diff
@@ -180,4 +183,28 @@ describe('options', function () {
...
+        fs.unlinkSync(path.resolve(__dirname, 'test-env/test/mocha.opts'));
+      } catch (e) {}
+      try {
+        fs.rmdirSync(path.resolve(__dirname, 'test-env/test'));
+      } catch (e) {}
+      try {
+        fs.rmdirSync(path.resolve(__dirname, 'test-env'));
+      } catch (e) {}
+      assert(output.indexOf('Usage:') >= 0);
+      done();
```
kunagpal: The entire test is synchronous, and as such, the `done` callback is not needed here.

kunagpal:
@Zarel What @ScottFreeCode was referring to was tweaking:
```js
assert(output.indexOf('Usage:') >= 0)
```
to something like:
```js
expect(output).to.contain('Usage:');
```

Here, `expect` here is an available global. Additionally, you could also log the actual output so that we can see what exactly is coming up in the CI builds.

ScottFreeCode:
Sorry about the late followup...

> Additionally, you could also log the actual output so that we can see what exactly is coming up in the CI builds.

This should happen automatically if a more detailed assertion is used such as that `contains` suggestion, and that avoids interleaving console logs with the test output.

Aside from that, @Zarel, @kunagpal's suggestions all look good to me at a glance; feel free to try applying any/all of them.

This might also end up having to be... merged a bit with other updates that have been made to `master`; but I'd be happy to take care of that.

ScottFreeCode:
Actually, looked at the conflict and -- seriously -- GitHub was confused because two entirely separate sets of tests were added to the same file? It couldn't figure out to just keep both additions? Eh, whatever; I went ahead and resolved that.

Zarel:
It looks like the problem for AppVeyor is:

```
'C:\projects\mocha\bin\mocha' is not recognized as an internal or external command,
operable program or batch file.
```

Is Mocha built/run a different way in Windows?

ScottFreeCode:
The Travis failure on Node 0.10 turns out to be due to [child_process.execSync](https://nodejs.org/dist/latest-v8.x/docs/api/child_process.html#child_process_child_process_execsync_command_options) not existing till Node 0.11...

The Windows error... sounds to me as though it is trying to run the Mocha script as though it were an executable?

I'm sure we have some helpers in Mocha's testsuite somewhere that will run a separate CLI instance of Mocha properly on both OSes and all versions of Node; let me see if I can find them and get back to you.

ScottFreeCode:
So, I found an integration test helper and used it to do this (added a commit to the branch), although... it runs Mocha from the current working directory, so if we wanted to be absolutely sure that the opts file contains "foo" (rather than presuming that a mocha.opts file at all would trigger the issue we're fixing), then we'd have to tweak the helper to make it possible to override the current working directory. Hmm.

Zarel:
UOnly certain things in `mocha.opts` trigger this issue. `foo` is one of those things.

Zarel:
Still, though, this level of struggling seems excessive to leave a bug unfixed. (I don't have mocha options memorized and it's mildly annoying to have to `cd` out of my project directory to look up help for it.)

Zarel:
It seems weird to leave a bug unfixed because your test framework doesn't support testing the fix?

boneskull:
@ScottFreeCode does this need anything or can it be merged?

Zarel:
Rebased to fix the merge conflict.

Zarel:
For the record, I have trouble testing this; I get a different error every time I run `npm test`; most recently:

```
17 10 2017 18:39:16.173:INFO [launcher]: Starting browser PhantomJS
17 10 2017 18:39:16.218:ERROR [karma]: { Error: spawn ENOTDIR
    at _errnoException (util.js:1021:11)
    at ChildProcess.spawn (internal/child_process.js:325:11)
    at exports.spawn (child_process.js:494:9)
    at spawnWithoutOutput (/Users/zarel/Workspace/mocha/node_modules/karma/lib/launchers/process.js:168:24)
    at Object.ProcessLauncher._execCommand (/Users/zarel/Workspace/mocha/node_modules/karma/lib/launchers/process.js:76:21)
    at Object.PhantomJSBrowser._start (/Users/zarel/Workspace/mocha/node_modules/karma-phantomjs-launcher/index.js:80:10)
    at Object.<anonymous> (/Users/zarel/Workspace/mocha/node_modules/karma/lib/launchers/process.js:19:10)
    at emitOne (events.js:120:20)
    at Object.emit (events.js:210:7)
    at Object.BaseLauncher.start (/Users/zarel/Workspace/mocha/node_modules/karma/lib/launchers/base.js:42:10)
    at Object.j (/Users/zarel/Workspace/mocha/node_modules/karma/lib/launcher.js:116:17)
    at Object.setTimeout.bind.j (/Users/zarel/Workspace/mocha/node_modules/qjobs/qjobs.js:143:18)
    at ontimeout (timers.js:471:11)
    at tryOnTimeout (timers.js:306:5)
    at Timer.listOnTimeout (timers.js:266:5)
 code: 'ENOTDIR', errno: 'ENOTDIR', syscall: 'spawn' }
make: *** [test-browser-unit] Error 1
```

ScottFreeCode:
I think the initial test was correct but didn't work on Windows, then my attempt to use the existing cross-platform helpers (because Node really makes portable spawning/forking difficult) worked on Windows but didn't actually prove that the fix works (unlike the original test).

I verified this by pulling down the code and trying the test with and without the fix in the source. So... had to think about it a while, but I *should* have something that can test this correctly cross-platform and will be pushing it up shortly. If that passes CI we should go ahead and merge this already -- it's waited longer than it should have.

ScottFreeCode:
All CI passed except linting (oops, should have done that before pushing); pushed a fix for the linting error, going to merge.

Issue #2010, opened 2015-12-16
White colour scheme hurts my eyes.


bcowgill:
I've had a personal mocha.css/js which uses a dark colour scheme because I can't look at a white scheme all day long.  I've got code changes which I could commit but don't know the contribution process.

What I've done is add a body class of mocha-dark or mocha-light and mocha.css changes to apply a dark scheme if mocha-dark is present.

In code, it stores a cookie mocha-scheme=mocha-light or mocha-dark with an expiry of  1 year.
The mocha object gets two new methods:
mocha.initColorScheme(options.colorScheme) which reads the cookie and sets the body class
mocha.setColorScheme('mocha-dark') which sets the body class and the cookie value.

Default is to use mocha-light scheme so no one's the wiser.

Comments ? Criticisms ? What to do to submit it?

bcowgill:
I've ran all the tests against my change and they pass. Don't see any tests for browser_entry.js or progress.js so didn't write any new ones. I have a patch as I couldn't create a branch.

From 99f3be1466792e5cd205a4151efe3b8990a74bda Mon Sep 17 00:00:00 2001
From: "Brent S.A. Cowgill" <brent.cowgill@workshare.com>
Date: Wed, 16 Dec 2015 15:15:49 +0000
Subject: [PATCH] issue 2010 mocha-dark colour scheme by Brent S.A. Cowgill

---
 lib/browser/progress.js  |  9 +++++++++
 mocha.css                | 49 ++++++++++++++++++++++++++++++++++++++++++++++++
 support/browser-entry.js | 32 +++++++++++++++++++++++++++++++
 3 files changed, 90 insertions(+)

diff --git a/lib/browser/progress.js b/lib/browser/progress.js
index 3186b6e..8e0235f 100644
--- a/lib/browser/progress.js
+++ b/lib/browser/progress.js
@@ -109,6 +109,15 @@ Progress.prototype.draw = function(ctx) {
     var text = this._text || (percent | 0) + '%';
     var w = ctx.measureText(text).width;
 
+    try {
+      /* global document */
+      if (document.body.className.match(/\bmocha-dark\b/)) {
+        ctx.fillStyle = 'yellow'; // BSAC DARK SCHEME
+      }
+    } finally {
+      // default colour scheme if error
+    }
+
     ctx.fillText(text, x - w / 2 + 1, y + fontSize / 2 - 1);
   } catch (err) {
     // don't fail if we can't render progress
[...]

PR #2747, opened 2017-03-19
Issue#2010 Switch to white or black background

bcowgill:
https://github.com/mochajs/mocha/issues/2010
body class=mocha-light or mocha-dark added based on
cookie mocha-scheme setting if found.
A little button added to bottom of test page which will allow the user to change the setting easily.

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -168,26 +168,6 @@ mocha.run = function (fn) {
   });
 };
 
-mocha.setColorScheme = function (scheme) {
-  var cookieValue;
-  var maxAgeInSeconds = (60 * 60 * 24 * 365);
-
-  try {
-    document.body.className = document.body.className
-      .replace(/\bmocha-[a-z]+\b/g, '');
-    if (scheme) {
-      scheme = scheme.trim();
-      document.body.className += ' ' + scheme;
-      cookieValue = scheme;
-    } else {
-      cookieValue = 'mocha-light';
-    }
-    document.body.className = document.body.className.trim();
-    document.cookie = 'mocha-scheme=' + cookieValue +
-      ';max-age=' + maxAgeInSeconds;
-  } finally {}
-};
-
 mocha.initColorScheme = function (scheme) {
   try {
     var mochaScheme = document.cookie
@@ -201,6 +181,28 @@ mocha.initColorScheme = function (scheme) {
     }
   } finally {}
   return mochaScheme;
+};
+
+mocha.setColorScheme = function (scheme, cookieOnly) {
+  var cookieValue;
+  var maxAgeInSeconds = (60 * 60 * 24 * 365);
+
+  try {
+    if (scheme) {
+      scheme = scheme.trim();
+      cookieValue = scheme;
+    } else {
+      cookieValue = 'mocha-light';
+    }
+    if (!cookieOnly) {
+      document.body.className = document.body.className
+        .replace(/\bmocha-[a-z]+\b/g, '');
+      document.body.className += ' ' + cookieValue;
+      document.body.className = document.body.className.trim();
+    }
+    document.cookie = 'mocha-scheme=' + cookieValue +
+      ';max-age=' + maxAgeInSeconds;
+  } finally {}
 };
 
 mocha.getColorScheme = function (scheme) {
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -142,7 +142,6 @@ mocha.setup = function (opts) {
 
 mocha.run = function (fn) {
   var options = mocha.options;
-  mocha.initColorScheme('mocha-light');
   mocha.globals('location');
 
   var query = Mocha.utils.parseQuery(global.location.search || '');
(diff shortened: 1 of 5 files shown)


bcowgill:
The AppVeyor failure on node 0.12.18 doesn't happen for me locally.
mocha (master % u=)$ bin/mocha --inspect test/acceptance/http.spec.js 
/home/me/.nvm/versions/node/v0.12.18/bin/node: bad option: --inspect
mocha (master % u=)$ bin/mocha test/acceptance/http.spec.js 


  http
    ✓ should provide an example


  1 passing (19ms)

ScottFreeCode:
(Does anybody know a way to get AppVeyor to actually show the test failure result details? I am only seeing that the test failed, but not what the error/failure was.)

I'm not sure why we would need something like this built in when it's straightforward just to add custom CSS to the HTML (which is part of the project using Mocha rather than part of Mocha) or even to re-use that custom CSS by including it by a `<link>` to a file that's shared in a repo or package. That being said, if we want to consider it then the following issues would need to be addressed:
- I don't get anything unless I include JQuery on the page. Doesn't seem like it's being used for anything that isn't easy to do with the DOM APIs (however much many developers may dislike them). I think we should not require projects using Mocha to add JQuery for this, and should just use the DOM APIs considering Mocha doesn't already have JQuery bundled.
- The symbol is small and almost invisible by default, and frankly doesn't do a very good job of conveying its purpose before I click it even when I do see it and hover over it.
- The scheme does not correctly change the background color back when it is dark and I click the button to change back to light (some other things do change back, leading to unreadable results such as black text on black background).
- The root `mocha.js` file will be overwritten when the bundle is updated; code changes would need to go into the source files.

bcowgill:
ScottFreeCode, thanks for your comments. 
What browser did you try it on?
Did you use file:// or an http:// protocol when loading?
I don't see the usage issues you are reporting on Chrome.

Most of it can be done with CSS except the top right progress circle because it uses a canvas you cannot change the fillStyle with CSS.

I based the opacity to match the other |> symbols but can certainly make it more visible.

bcowgill:
ok, I've removed the UI toggler for now and just use the color scheme setting from the cookie.
I've been waiting quite a while for this so I hope we can get this in.

To test it I just added class="mocha-dark" to the body tag of my HTML runner and ran the page.
After that deleting the class still keeps the dark scheme by reading the cookie.

boneskull:
@bcowgill 

Hi, thanks for the PR.  I'm not sure why the JavaScript is necessary to implement a dark scheme.  Simply using your own custom styles should work; there's no requirement to use `mocha.css` (or any CSS, really) to execute tests in the browser.

I'm disinclined to merge this, because it increases complexity and maintenance overhead to benefit a relatively small portion of our userbase (those executing browser-based unit tests in a non-automated fashion) .  Mocha can't afford that.

I'd encourage you to create a [gist](https://gist.github.com) or find another way to share your dark color scheme, as I'm sure others may find it useful.  Also, there are plugins for browsers such as [Dark Reader](https://chrome.google.com/webstore/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh) which you might find useful.

Anyway, even if this is not merged, please do not be discouraged from contributing to Mocha.  :smile: You may have even created this PR or issue before we added [contribution guidelines](https://github.com/mochajs/mocha/blob/master/CONTRIBUTING.md).  Feature PRs are more likely to be merged if they arrive after discussion.  There are plenty of other areas in which Mocha needs help, as you can see in [the README](https://github.com/mochajs/mocha/blob/master/README.md)!

steve-taylor:
This is worth revisiting in another PR given widespread support for [`prefers-color-scheme`](https://caniuse.com/#search=prefers-color-scheme). There's no need for any JavaScript or cookies.

boneskull:
@steve-taylor agree.

bcowgill:
@steve-taylor, @boneskull

It's been a long time, but I'm updating one of my projects that uses mocha and so am updating this again for mocha 8.3.0.  And have a bit of time available.

The use of prefers-color-scheme CSS would be a nice (minimal) thing to add but people with vision difficulties may want to turn on light/dark schemes for things independently of changing their entire OS color scheme.
For example, there are still a lot of apps that do not render well when the OS dark scheme is set as they hard code some colors to black still.
Sometimes vision problems are temporary (before / after cataract surgery, speaking from experience.)
Also, for example, I use Linux with i3 window manager which is a minimalist environment and don't think it even supports specifying the prefers-color-scheme.
Use of things like chrome extensions to invert colors is helpful but usually you get a wonky color scheme when you simply invert them.

For the best accessibility experience the user needs to be able to switch between light and dark as needed.
Adding a mocha-dark class to the body in the test page index.html is a good compromise.
Normal team members can use the usual index.html page and another team member can simply use a git stash to apply the class to the body when they are running the tests or keep an index-dark.html alongside to choose their scheme.
But ideally the user would want to be able to click something on the page to just toggle the color scheme and have it be remembered when they refresh.
In my local changes, I've used the circular progress indicator as the place to click to toggle the scheme and use localStorage to save the setting only once the user has done so.
So if the CSS prefers is present it will work and if the user overrides the setting in the body or by clickin it will work.

A Javascript change is also needed to properly effect the color scheme change because the circular mocha progress indicator is rendered in a canvas and cannot be seen when you go to dark theme. (see screenshot)

mocha-light-dark-canvas

So at the very least, even if prefers-color-scheme is used the canvas code must be made aware of that to change the text color.
If the click to toggle is present it needs to re-render the progress indicator to correct the rendered colors of text against the background.

We could get by with a fixed text color of orange or blue for the progress indicator which is pretty visible against both white and black. (obviously we wouldn't want to use green or red as those have semantic meaning about passing and failing.)

I can create a fresh pull request which switches the scheme based on prefers-color-scheme and a mocha-dark class name if you're willing.

Alternatively I'll create a mocha-dark package on NPM and include the css and javascript file there with full functionality.

PR #2754, opened 2017-03-29
fix the erorr that it can not unhide all suites

ZeroBinbin:
by convert from HTMLCollection to Array

### DIFF
--- a/mocha.js
+++ b/mocha.js
@@ -2883,7 +2883,7 @@ function hideSuitesWithout (classname) {
  * Unhide .hidden suites.
  */
 function unhide () {
-  var els = document.getElementsByClassName('suite hidden');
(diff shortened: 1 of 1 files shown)


ScottFreeCode:
- Is there a bug report for this issue? I am not sure why or in which environment(s) the current code doesn't work.
- Don't update the bundle, update the relevant source file (most of them are in `lib`, I think some browser-specific stuff is in `browser-entry.js`).

ZeroBinbin:
It is the problem with the reporter page,while my tests are all pass ,so when I click the 'failures' button to filter the result ,they all hidden, it`s right. But when I click the 'passes' button ,only some of them are shown.

see this [https://zerobinbin.github.io/mocha-test-report/runner.html](url)

And I`m sorry to update the bundle  ，because it`s my first time to pull a request ,I did it in hurry

PR #2759, opened 2017-04-06
test fallback to ChromeCanary/Chromium/PhantomJS in macOS

teabyii:


boneskull:
the config lacks context about why Chrome is the default. it's the default because at the time of the change, PhantomJS v1.9.8 [was causing seg faults](https://github.com/ariya/phantomjs/issues/14558).

But...I can no longer reproduce the seg fault.  It's likely the executable or macOS has been fixed in the meantime.  So PhantomJS should be the default across all OS's.

So, while we can't accept this PR--because the conditional should be removed altogether (which I'm about to do)--we can accept a PR in the future that adds support for headless Chrome, or make it easier for users to specify which browser (ChromeCanary, etc) they want to launch via the command line.

That could start with ditching `make` as a build tool.  :wink:

Issue #2772, opened 2017-04-20
Add tests for loading reporters w/ relative/absolute paths


dasilvacontin:
I'd like to have tests for the new feature added in #2659.

Gonna do this today after I review the rest of the un-released commits. Let me know if you want to tackle it for me.

sul4bh:
I can help with writing a test to cover #2659 if you give me some pointers on what to cover in the test.

dasilvacontin:
@sul4bh Thanks!

It should cover:
- using a relative path as the argument for the `reporter` option
- using an absolute path as the argument for the `reporter` option

So maybe write a very short and useless reporter, and use this helper to execute the mocha binary, and check its output and that there haven't been any error: https://github.com/mochajs/mocha/blob/7a05a6c8c196bb9a8dcfb156499507238f6f7e5a/test/integration/helpers.js#L44-L83.

Ping me on Gitter if anything! :)

sul4bh:
Awesome. I am on it!

dasilvacontin:
@sul4bh ❤️

dasilvacontin:
Closed via #2773.

PR #2773, opened 2017-04-20
Adds tests for loading reporters w/ relative/absolute paths

sul4bh:
@dasilvacontin How does this look for https://github.com/mochajs/mocha/issues/2772?

### DIFF
--- a/test/integration/reporters.spec.js
+++ b/test/integration/reporters.spec.js
@@ -76,7 +76,7 @@ describe('reporters', function () {
 
     it('loads a reporter from an absolute path', function (done) {
       // Generates an absolute path string
-      var reporterAtAnAbsolutePath = process.cwd() + '/test/integration/fixtures/simple-reporter.js';
+      var reporterAtAnAbsolutePath = path.join(process.cwd(), 'test/integration/fixtures/simple-reporter.js');
 
       var args = ['--reporter=' + reporterAtAnAbsolutePath];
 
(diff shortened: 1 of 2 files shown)


test/integration/reporters.spec.js
```diff
@@ -60,4 +60,29 @@ describe('reporters', function () {
...
+
+      run('passing.fixture.js', args, function (err, result) {
+        assert(!err);
+        done();
+      });
+    });
+
+    it('loads a reporter from an absolute path', function (done) {
+      // Generates an absolute path string
+      var reporterAtAnAbsolutePath = process.cwd() + 'test/integration/fixtures/simple-reporter.js';
```
dasilvacontin: Hmm, weird that this works, `.cwd()` doesn't finish with a `/` on my computer.

dasilvacontin: Could you assert the value of `result` on both tests? I'm a bit skeptical this works.

sul4bh:
I totally botched the test in previously. It looks better now.

test/integration/reporters.spec.js
ScottFreeCode: For the first thing, `path.join` is real nice in terms of not having to care whether there's already a `/`.

dasilvacontin:
True, I thought of that when I saw the '+' concatenation, but I ended up forgetting to comment it. Good catch!

sul4bh:
travis failure is due to a test infrastructure issue.

ScottFreeCode:
Reran, looks like that passed now.

dasilvacontin:
LGTM, thanks @sul4bh! ❤️

PR #2648, opened 2016-12-28
use `semistandard` directly

kt3k:
This PR simplifies the lint tool dependencies by using `semistandard` directly.
We don't need to keep track of each peer dep of `eslint-config-semistandard` anymore by this change.

---
closes #2633

PR #2795, opened 2017-05-09
reverts use of semistandard directly (#2648)

boneskull:
Abstractions around ESLint are not widely supported in text editors or IDEs.  For example, I could tell WebStorm to use `node_modules/eslint-config-semistandard/eslintrc.json`, but there's no way for WebStorm to know about the overrides in the [`semistandard` prop of `package.json`](https://github.com/mochajs/mocha/blob/master/package.json#L374), nor the overrides specified for [test files in `Makefile`](https://github.com/mochajs/mocha/blob/master/Makefile#L38).

`semistandard` is arguably easier to use on the CLI, and requires less configuration files, but the cost to users primarily interfacing with ESLint via contextual information in an IDE (like myself!) outweighs the benefit.

cc @kt3k

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,0 +1,11 @@
+env:
+  node: true
+  browser: true
+parserOptions:
+  ecmaVersion: 5
+  sourceType: script
+extends: semistandard
+rules:
+  strict:
+    - error
+    - safe
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BROWSERIFY := "node_modules/.bin/browserify"
-SEMISTANDARD:= "node_modules/.bin/semistandard"
+ESLINT := "node_modules/.bin/eslint"
 KARMA := "node_modules/.bin/karma"
 MOCHA := "bin/mocha"
 NYC := "node_modules/.bin/nyc"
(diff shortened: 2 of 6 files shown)


boneskull:
Worth mentioning: WebStorm supports `standard`, but not `semistandard`, nor `xo`.

kt3k:
@Flet
Can you see this issue? How do you think?

kt3k:
Personally I don't use editor integration for linting. I think this issue depends on people and the team who use it. If you evaluate the negative impact is bigger than the benefits here, I think it makes sense to switch back to eslint cli.

Flet:
Yeah, @boneskull is right in that IDE integration can be tricky if its not supported by an existing plugin. For WebStorm, it should be possible to configure it manually similar to how it can be done with `standard`: https://standardjs.com/webstorm.html
- be sure to turn on them semicolons in the xml :)
- for .eslintrc.json use `{"extends": ["semistandard"]}`

`semistandard` and `standard` have some ignore paths baked-in that eslint does not (including ignoring everything in `.gitignore`), so to remedy that, you may also need an `.eslintignore` file: http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories.

There are `semistandard` plugins for Sublime Text and VS Code which work great, but yeah it can be a pain in other IDEs :/. The JetBrains folks might be receptive to adding `semistandard` too, but I don't know if anyone has asked :).

For this particular PR, I'd say adding a `.eslintignore` file would be needed.

boneskull:
> For this particular PR, I'd say adding a .eslintignore file would be needed.

We have an `.eslintignore` file already, which was not removed.  This should be GTG.

PR #2806, opened 2017-05-15
Linting: enforce unix line endings. Closes #2800

### LINKED ISSUE #2800: Unix/Linux line endings in CLI need to be enforced
`bin/mocha` and `bin/_mocha` need to have Unix line endings. Currently Git's autocrlf setting can screw this up. (See also #2799 for an example of when this goes wrong.) Our tests don't check for it either.

We will need to either check for or fix the line endings in these files as part of publishing. Or, if doing it on publish isn't feasible (prepublish vs. prepublishOnly seems to be a bit of an issue?), on version (most of our other publishing prep is in versioning already anyway...).

Ideally, I'd like to just run a Node command and convert the line endings, so it doesn't matter what Git did (or the user, etc) -- this isn't something that should have to be fixed manually when it is detected.

Munter:


Munter:
Per our .editorconfig we want unix newlines on everything. This is just an enforcement of this policy, which happens to also completely cover your error case

ScottFreeCode:
Interesting; well, if we want to go that way, we should really eliminate the Git settings issue in the first place by adding the same effect to `.gitattributes` too. Then nothing has to happen manually.

My vague impression is that Windows development is smoothest with per-OS newlines rather than always-Unix/Linux newlines, but I don't use a lot of fancy software most of the time either, so I am not sure just how much of a difference it makes to the typical Windows developer. I know Notepad is one thing that still can't handle other OSes' line endings, but there's Gitpad (basically a blocking Notepad with 'nix newline support) for people hacking on a Git repo...

dasilvacontin:
Sorry @Munter, don't these config files inherit from those found higher in the file hierarchy? I'd just like to avoid redundancy, if possible. Not sure if it's the case, just asking.

dasilvacontin:
Apparently they do inherit: http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy.

Having those rules in our main `.estlinrc` file should be enough. Do you agree?

boneskull:
Yes.

lib/browser/.eslintrc.yaml
```diff
@@ -2,3 +2,8 @@ env:
   node: false
```
boneskull: revert this file

test/.eslintrc.yaml
```diff
@@ -5,3 +5,8 @@ globals:
   assert: false
```
boneskull: revert this file

dasilvacontin:
Thanks! 😊

Issue #2805, opened 2017-05-14
Eliminate shell scripts from Mocha's test suite


ScottFreeCode:
Besides getting ourselves off the Makefile (#2268?), the other thing that doesn't work on a default Windows development environment is Unix/Linux shell scripts. (Even with a Posixy environment, they can have problems with line endings like #2801; this would not be an issue if they weren't Posixy shell scripts in the first place.) As far as I am aware, Node's `child_process` actions can run anything in a shell the same as if it were in a `.sh` script, so we should be able to convert any such test files to proper JS test files.

The one of these of which I am aware is:
- test/acceptance/glob/glob.sh

(We can make an exception for scripts used only by CI and not local test runs, of course.)

Munter:
Relevant to migrating away from shellscripts for test:

https://gitter.im/unexpectedjs/unexpected?at=5919695c2b926f8a6757de02

kunagpal:
At present, the following three shell scripts remain:

[test/acceptance/glob/glob.sh](https://github.com/mochajs/mocha/blob/master/test/acceptance/glob/glob.sh)
[scripts/travis-before-install.sh](https://github.com/mochajs/mocha/blob/master/scripts/travis-before-install.sh)
[scripts/travis-before-script.sh](https://github.com/mochajs/mocha/blob/master/scripts/travis-before-script.sh)

If glob.sh is indeed a valid exception, then this issue can be closed, as #2807 has landed in master, and has been released in [v3.4.2](https://github.com/mochajs/mocha/releases/tag/v3.4.2)

ScottFreeCode:
To clarify, the parenthetical about CI scripts was referring to those Travis scripts, and glob.sh was not proposed as an exception but as an example of the sort of scripts the issue is about -- required for Mocha's complete test suite to be run yet themselves requiring a Posix shell and/or environment (both in this case).

Munter:
The travis scripts are fine. They are never required for local execution. We need to convert the glob script and also get rid of the makefile

PR #2807, opened 2017-05-15
Move linting into an npm script. Relates to #2805

Munter:


dasilvacontin:
@Munter why the `to-merge` label?

Munter:
@dasilvacontin I was assuming it means this is ready to merge in. But it's probably easier for the next person finishing up a realease to do it, since they won't have to hunt down changes in reverse from the git log

dasilvacontin:
@Munter Hmm, I think it's simpler to merge directly. Simpler, less rules to follow, less assumptions to make.

I'd prefer to enforce all changes to go through PRs, so when doing the reverse scan on the git log I can safely focus only on the merge commits.

Also, if you wait to merge a bunch of PRs in one go, there could be conflicts. That would be horrible. 🤔

Munter:
Yeah, didn't think that all the way through

dasilvacontin:
All good.  😊

Issue #2811, opened 2017-05-18
Allowing functions to be passed in as titles as long as they have names


armsteadj1:
This would allow functions or even React classes to be passed in when testing against them and allow for easier refactoring than praying IntelliJ finds all of your stings with that function in them.

For example: 

```
function addition (a, b) { return a+b; };
describe(addition, function () {
    it('will add the two numbers', function () {
       ....
    });
});
```

Now if `addition` were ever renamed to `addNumbers` our test would nicely be updated along the way.

ScottFreeCode:
...I might be missing something, but wouldn't it do to just write it like this?
```js
function addition (a, b) { return a+b; };
describe(addition.name, function () {
    it('will add the two numbers', function () {
       ....
    });
});
```

armsteadj1:
Considering the backwards compatibility issues with what I tried to create. :). That actually may be better off. I figured it would just be nice if that was wrapped up in mocha to clean up the code. I'm totally fine with your suggested route though.

boneskull:
@armsteadj1 Relying on `Function.prototype.name` can be folly due to transpilation, minification, lack of support, etc.  It also makes for terse reporter output, as the function names cannot have a host of characters in them.

Maybe [reflect-metadata](https://npm.im/reflect-metadata) would be useful, or even TypeScript decorators?  Tagged template literals? This would be cool for some future Mocha.

Either way, if what you have works for you, you can create a custom ui and load it.

I'm going to close the associated PR, but thanks for bringing it up!

PR #2812, opened 2017-05-18
#2811 - allowing functions to be passed in as titles as long as they have names 

armsteadj1:
#2811

This would allow functions or even React classes to be passed in when testing against them and allow for easier refactoring than praying IntelliJ finds all of your stings with that function in them.

For example: 

```
function addition (a, b) { return a+b; };
describe(addition, function () {
    it('will add the two numbers', function () {
       ....
    });
});
```

Now if `addition` were ever renamed to `addNumbers` our test would nicely be updated along the way.

### DIFF
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -43,6 +43,7 @@ exports.create = function (parent, title) {
  * @param {Context} parentContext
  */
 function Suite (title, parentContext) {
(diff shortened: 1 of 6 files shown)


boneskull:
See #2811 for discussion

PR #2813, opened 2017-05-20
Pass flags for NAPI modules

jupp0r:
This change enables testing of NAPI modules[1]. This is achieved
by passing the appropriate flags to node inside the mocha executable.

[1]: https://nodejs.org/api/n-api.html


boneskull:
@jupp0r thanks!

PR #2818, opened 2017-05-22
Remove call to deprecated os.tmpDir (#2802)

### LINKED ISSUE #2802: Deprecation warning in Mocha's test
https://github.com/mochajs/mocha/blob/0a93024b475e58e75c5def6b2f5228bd363874de/test/integration/reporters.spec.js#L39 is using `os.tmpDir` and that's printing a deprecation warning.

Depending on supported versions of Node and how far back the replacement `os.tmpdir` goes, we may have to dynamically detect the availability of `os.tmpdir`.

Otherwise, we should just change that one letter and be done with it.

makepanic:
fixes #2802 

`os.tmpdir` was added in 0.9.9 (see https://github.com/hapijs/hapi/issues/3369).
With mocha having a minimum node version of 0.10 it can be dropped without issues.

### DIFF
--- a/test/acceptance/file-utils.spec.js
+++ b/test/acceptance/file-utils.spec.js
@@ -8,7 +8,7 @@ var mkdirp = require('mkdirp');
 var rimraf = require('rimraf');
 
 describe('file utils', function () {
-  var tmpDir = path.join(os.tmpDir(), 'mocha-file-lookup');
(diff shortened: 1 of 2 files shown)


ScottFreeCode:
Thanks!

dasilvacontin:
This fix has been released in [v3.4.2](https://github.com/mochajs/mocha/releases/tag/v3.4.2)! 🎉

PR #2820, opened 2017-05-23
Simplified fix for issue #2713

chrisleck:
Simpler version of PR #2714 which only eagerly sets the exitcode without playing with the draining semantics


chrisleck:
CI failure looks like a flake.

dasilvacontin:
LGTM! Thanks @chrisleck!

PR #2829, opened 2017-05-24
add config for probot-stale integration

### LINKED ISSUE #2828: configure probot-stale
it's [this thing](https://github.com/probot/stale)

boneskull:
closes #2828.

Issues and PRs will be marked `stale` after 4 mos. of inactivity and closed if there's no activity after 2 more weeks.

*Most* issues with labels are exempt from the check.  I'm not sure how well this is going to work, but we can always change it later.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,39 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 120
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pr-please
+  - confirmed
+  - future
+  - bug
+  - chore
+  - feature
+  - unconfirmed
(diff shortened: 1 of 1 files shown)


dasilvacontin:
I'm finding it hard to understand which issues are worthy of being marked as stale and being closed.

I can come up with:
- help issues that ask for help with a concrete code example, we reply to them, and we never hear back from OP.

Any others?

boneskull:
@dasilvacontin IMO, we should mark the following as stale:

- PRs that need work and have become rotten
- Stuff marked `needs-feedback` that never got feedback
- Unlabeled stuff without activity

4 mos is a long time, and there's a warning, so a user can reply and say "yes, this is still a question/problem/etc".

In the config, we're blacklisting almost every label except a few (`duplicate`, `needs-feedback`, `pr-needs-work`).  Really, we need a whitelist instead (see probot/stale#22).

Issue #2796, opened 2017-05-10
🚀 Feature: Test duration not shown for failed tests, but does for passed tests


aDu:
Concerning runners, listening to 'test end' events.

For passed tests, test.duration is given in milliseconds.

However, when there is a failed test, test.duration is not defined.

Why is that? Can test.duration be implemented for failing tests?

Thanks in advance! =)

Elergy:
Hey @AlanDu50,

Could you please provide some information about how you run tests (a sample of code would be helpful) because I tried to reproduce the situation and couldn't do that [here](https://github.com/mochajs/mocha/issues/2834)

Thanks!

BrandonZacharie:
I think I'm seeing this as well...

![Screen Shot 2019-09-19 at 16 55 54](https://user-images.githubusercontent.com/27306/65289197-9ce03b00-dafe-11e9-855a-b60282d7a8c2.png)

The test that was run after the test that failed somehow had `duration` undefined. I placed that log statement on the first line in `runner` callback...

![image](https://user-images.githubusercontent.com/27306/65289361-27289f00-daff-11e9-9e9f-01d0a2f86f10.png)

🤷🏾‍♂️

JoshuaKGoldberg:
I don't know what to make of this issue 😞 nobody ever posted a full reproduction. So I'm going to close it as stale.

But if someone is facing this, please do file a new issue with one of our new issue templates. They prompt for the info needed to diagnose problems.

Cheers, all! ☕

PR #2834, opened 2017-05-26
#2796 - a test case for "test end" event

Elergy:
I couldn't reproduce a bug from #2796 
But I implemented a couple of tests to make this piece of code covered

### DIFF
--- a/test/runner.spec.js
+++ b/test/runner.spec.js
@@ -433,4 +433,92 @@ describe('Runner', function () {
       });
     });
   });
+
+  describe('.run', function () {
+    describe('"test end" event', function () {
+      it('should emit "test end" event for each test', function (done) {
+        var finishedTestTitles = [];
+
+        suite.addTest(new Test('first test', noop));
+        suite.addTest(new Test('second (failed) test', function () {
+          var falseValue = false;
+          falseValue.should.be.equal(true);
+        }));
+        suite.addTest(new Test('third (failed because of an exception) test', function () {
+          throw new Error('FAIL TEST');
+        }));
+        suite.addTest(new Test('fourth test', noop));
+
+        runner.on('test end', function (test) {
+          finishedTestTitles.push(test.title);
+        });
+
+        runner.run(function () {
+          finishedTestTitles.should.deepEqual([
+            'first test',
+            'second (failed) test',
+            'third (failed because of an exception) test',
+            'fourth test'
+          ]);
+          done();
+        });
+      });
+
+      it('should have a proper duration field', function (done) {
+        var finishedTestDuration = [];
+
+        function waitForSomeTime (timeInMs) {
+          var endTestTime = new Date().getTime() + timeInMs;
+
+          while (new Date().getTime() < endTestTime) {
+            //
+          }
+        }
+
+        function doSomethingForSeveralMilliseconds () {
+          waitForSomeTime(10);
+        }
+
+        function doneAfterTimeout (done) {
+          setTimeout(done, 10);
+        }
+
+        function failAfterSeveralMilliseconds () {
+          waitForSomeTime(10);
+
+          var falseValue = false;
+          falseValue.should.be.equal(true);
+        }
+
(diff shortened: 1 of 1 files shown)


boneskull:
@Elergy Thanks!

CI failure is due to integration tests running too slowly on AppVeyor infrastructure.   I'll address this; please rebase onto `master` once I do

test/runner.spec.js
```diff
@@ -433,4 +433,87 @@ describe('Runner', function () {
...
   });
+
+  describe('.run', function () {
+    describe('"test end" event', function () {
+      it('should emit "test end" event for each test', function (done) {
+        var finishedTestTitles = [];
+
+        suite.addTest(new Test('first test', noop));
+        suite.addTest(new Test('second (failed) test', function () {
+          expect(1 + 1).to.equal(3);
```
ScottFreeCode: I'd prefer `expect(false).to.equal(true)` or better yet `throw new Error("FAIL TEST")` in place of all the `expect`s that are intended to fail in these new tests, for safety (no chance of changes to the variables and values involved ending up in a passing `expect` in the future), simplicity and clarity.

test/runner.spec.js
```diff
@@ -433,4 +433,87 @@ describe('Runner', function () {
...
+      });
+
+      it('should have a proper duration field', function (done) {
+        var finishedTestDuration = [];
+
+        function doSomethingForSeveralMilliseconds () {
+          var result = 0;
+          for (var i = 0; i < 10000000; i++) {
+            result = i;
+          }
```
ScottFreeCode: Let's not count on the JS engine never figuring out that it could just set result to 100000000 in a single operation, skipping the loops and thus shaving off all this time; instead, use something like `var then = Date.now() + <desired milliseconds later>; while (then < Date.now()) {}` (assuming some such technique is available in all our supported environments...) to explicitly control the delay in synchronous tests and `setTimeout` to explicitly control the delay in asynchronous tests.

test/runner.spec.js
```diff
@@ -433,4 +433,87 @@ describe('Runner', function () {
...
+
+        function doneAfterTimeout (done) {
+          setTimeout(function () {
+            var result = 0;
+            for (var i = 0; i < 10000000; i++) {
+              result = i;
+            }
+            if (result > 0) {
+              done();
+            }
```
ScottFreeCode: If the test were supposed to be able to fail, a specific error would be better than a timeout unless a timeout is the specific failure being tested; and in this case it appears the test is not supposed to be able to fail. Either way, don't call `done` conditionally with the alternative being to niether call `done` nor throw.

test/runner.spec.js
```diff
@@ -433,4 +433,87 @@ describe('Runner', function () {
...
+        suite.addTest(failedAsynchronousTest);
+        failedAsynchronousTest.timeout(20);
+
+        runner.on('test end', function (test) {
+          finishedTestDuration.push(test.duration);
+        });
+
+        runner.run(function () {
+          finishedTestDuration.forEach(function (duration) {
+            duration.should.be.above(0);
```
ScottFreeCode: If possible, assertion library (and its style in the case of libraries like Chai with multiple styles) should be consistent within the file. (Caveats: I haven't looked at whether the existing tests in this file are using `expect`, `should` or both, and most or all of the `expect`s could be replaced with a `throw` as mentioned above.)

test/runner.spec.js
Elergy: Thanks, it think it could be even better to add one more case for test which is failed because of an exception

test/runner.spec.js
Elergy: Ok, I'll change samples of tests to `should` instead of `expect`

boneskull: Ideally we'd only use _one_ assertion library.  alas...

Elergy: Yes, thanks, it's a reasonable requirement.

Firstly I tried to separate _our_ tests from _users_, using another assertion style, but now this file contains only `should`

Issue #2819, opened 2017-05-23
this.skip() in before fails to skip tests with nested describe calls


fearphage:
Based on the documentation:

> To skip multiple tests in this manner, use `this.skip()` in a “before” hook

I expected the test below to never run.

```js
describe('skip', function() {
  before(function() {
    this.skip();
  });

  describe('bug', function() {
    it('should never be called', function() {
      throw new Error('Why is this reached?');
    });
  });
});
```

However, here is the output.

```shell
✗ mocha bug.js


  skip
    bug
      1) should never be called


  0 passing (7ms)
  1 failing

  1) skip bug should never be called:
     Error: Why is this reached?
      at Context.<anonymous> (bug.js:8:19)
```

This seems like a bug to me. If it is expected behavior, then the documentation should be updated to reflect this.

kunagpal:
@fearphage I'm able to reproduce this with Mocha v3.4.2. It appears that calling `this.skip()` doesn't work correctly for nested suites. 😕

RyanZim:
I've come across this issue as well. mocha@3.4.2

plasticrake:
Just ran into this issue as well in v4.0.1

paranoidjk:
ran into this issue too, any process?

paranoidjk:
cc @Munter @boneskull

mrosett:
I see this behavior too. I'm pretty sure it's a bug that was introduced at some point between late 2016 and the date that this thread was opened. I've been editing some tests that skipped correctly when I looked at them a year ago and now aren't being skipped.

aikrasnov:
5.0.4, ran into this issue too

jdmarshall:
I suspect there is something being mis-handled with "current test" and before resolution.  This feels like it may be a different symptom of the problem underlying #2134

tex0l:
I confirm I have this bug as well

ViKomprenas:
Yet another confirmation post. Is anyone going to get on this?

djmitche:
Looking at the number of open issues, it's doubtful :)

I have worked around this by setting a scoped `skipping` variable true if skipping, and then checking that in a before function in each subsuite.  It's not pretty, but it's a nice chance to link to this issue :)

```js
describe('outer', function() {
  let skipping = false;
  before(function() {
    if (...) {
      this.pending = 1;
      skipping = 1;
    }
  });

  describe('inner', function() {
    // work around https://github.com/mochajs/mocha/issues/2819
[...]

PR #2836, opened 2017-05-27
#2819 - this.skip() in before fails to skip tests with nested describe calls

Elergy:
this.skip() in before fails to skip tests with nested describe calls

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -148,6 +148,14 @@ Runnable.prototype.isPending = function () {
 };
 
 /**
+ * Mark the runnable as pending or not pending
+ * @param pending
+ */
+Runnable.prototype.setPending = function (pending) {
+  this.pending = pending;
+};
+
+/**
  * Set number of retries.
  *
  * @api private
--- a/test/acceptance/interfaces/bdd.spec.js
+++ b/test/acceptance/interfaces/bdd.spec.js
@@ -32,3 +32,33 @@ context('test suite', function () {
     expect(this.number).to.equal(5);
   });
 });
+
+describe('a suite skipped by "before" hook', function () {
+  before(function () {
+    this.skip();
+  });
+
+  it('should skip this test', function () {
+    expect(2).to.equal(1);
+  });
+
+  it('should skip this test too', function () {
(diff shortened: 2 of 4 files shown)


boneskull:
@Elergy See conversation in #2571.  I'm not sure if we ever came to an agreement about how this should work.  There are differing opinions.

Elergy:
Thank you, I didn't see #2571 before. 
I'm going to understand the scenarios and return :-)

test/acceptance/interfaces/bdd.spec.js
```diff
@@ -32,3 +32,33 @@ context('test suite', function () {
...
   });
 });
+
+describe('a suite skipped by "before" hook', function () {
+  before(function () {
+    this.skip();
+  });
+
+  it('should skip this test', function () {
+    expect(2).to.equal(1);
```
fearphage: In all the failure cases, you should call fail directly: `expect().to.fail('message here')`

lib/runnable.js
```diff
@@ -148,6 +148,14 @@ Runnable.prototype.isPending = function () {
 };
 
 /**
+ * Mark the runnable as pending or not pending
+ * @param pending
+ */
+Runnable.prototype.setPending = function (pending) {
```
fearphage: What is the value of switching from property to a method?

Elergy: In most cases, it's better to use setters instead of using properties directly.

It's a common problem with using properties (and I encountered this problem while fixing this bug) that it's not easy to understand where someone changes the property.

You know that the property was changed, but you don't know when - all you can do is to debug the whole application. 
You can search for a string 'pending = ', but it can be a problem when several objects contain fields with the same name. 

if you have a setter, you can easily set a breakpoint inside the setter and find who called this method very quickly.

test/acceptance/interfaces/bdd.spec.js
Elergy: Thanks!

craigtaub:
@Elergy i think the fix for this has now gone into master (via https://github.com/mochajs/mocha/pull/3225). Closing this PR.

PR #2841, opened 2017-05-27
revert 93392dd; downgrade to phantomjs v1.9.8; closes #2840

### LINKED ISSUE #2840: fix broken phantomjs test execution
At some point we wound up running PhantomJS v2.x, which is why 93392ddb061e6870eb3fc524a7eff600c881cef3 happened.

But this was erroneous--we should still be running PhantomJS v1.9.8 to assert compatibility.  And no, I still can't run that version on macOS Sierra.

93392ddb061e6870eb3fc524a7eff600c881cef3 should be reversed.  [karma-phantomjs-launcher](https://npm.im/karma-phantomjs-launcher) installs PhantomJS v2.x [since v1.0.1](https://github.com/karma-runner/karma-phantomjs-launcher/commit/9ae8ada4e59c38a2c4d47041bd13e4be56816091), and I think this doesn't play nicely with npm v4+.  It should be downgraded.

boneskull:


boneskull:
This downgrades karma-phantomjs-launcher to v0.2.3, which doesn't require `phantomjs-prebuilt`, nor does it have such a `peerDependency`.  This got messed up with bb743fffd83d7c99748769dd8b73b35bb7f8dc96.

PR #2842, opened 2017-05-27
reorganize test hierarchy; closes #2838

### LINKED ISSUE #2838: better test directory organization
The directory hierarchy of the `test` dir is confusing and doesn't have much relationship to how the tests are actually run.

In `Makefile`, our tests are put into many different buckets.  This is typically because each bucket requires unique command-line options.  Some options (such as setting the interface) cannot be set from tests themselves, so we have to split them up in this manner.

Let's rearrange these files into directories which correspond to the divisions as in `Makefile`.

boneskull:
Alright, so this moves some crap around and **BONUS** we get a few more tests running in the browser.

- Anything called "acceptance" (which basically means anything) is now "unit"
- Node-specific unit tests live in `test/node-unit`
- Unit tests for any environment are in `test/unit`
- Moved some tests for the `utils` module, since there were two of these; Node-specific tests were put into `test/node-unit/stack-trace-filter.spec.js`
- Be more specific about file extension (`.js` ➡️  `.spec.js`)
- Took all subdirs out of `test/unit` and moved them into `test`
- Ensure `process.stdout` is provided in Browserified test code (see `test/browser-fixtures/*`), now that some tests rely on it.
- Probably some reformatting.  Sorry!


boneskull:
Yeah, I'm gonna need to run this in Linux or something to reproduce the problems.

dasilvacontin:
Lovely, a better structure was indeed needed.

boneskull:
One thing was needing to avoid `process.nextTick()`, which I did.

test/browser/README.md
```diff
@@ -0,0 +1,3 @@
+These files need to be run manually by loading the `.html` file(s) in a browser.
+
+It would be awesome if we could automate that!
```
ScottFreeCode: Created a dedicated issue for this: #2862

dasilvacontin:
LGTM, much needed – thanks! ❤️

Issue #2295, opened 2016-06-03
Base reporter, add unified diff separator


olsonpm:
Right now the unified diff output has no separators.  Resulting in output similar to this:

![no-diff-seperator](https://cloud.githubusercontent.com/assets/5957709/15793934/c98aae62-29ac-11e6-8d8a-d0a5954c5b87.png)

Having not used mocha in some time, I had completely forgotten it splits before/after 4 lines from the affected text and spent a while trying to figure out how "PD" could possibly be having values of "15/30" or "50/100".

I don't know what kind of separator to add, but a blank line or two should suffice for majority of cases.  I'd have to check to see if other libraries have tackled this problem how they went about it.

The code to modify would be this:
https://github.com/mochajs/mocha/blob/7493bca76662318183e55294e906a4107433e20e/lib/reporters/base.js#L161-L207

ScottFreeCode:
Git has a format something like this: `@@ -<line number of start of this segment in previous file version>,<number of lines removed> +<line number of start of this segment in next file version>,<number of lines added> @@`
It's blue, where the + are green and the - are red. And I'm not sure I'm interpreting those numbers exactly right -- maybe the numbers after the commas include the unchanged lines shown?

olsonpm:
hmm - the problem with generating line numbers is that these diffs are for formatted objects, thus line numbers don't seem appropriate.  

However I did notice `git grep` separates sections with `\n--\n` if multiple matches are found.  That seems intuitive enough.

boneskull:
@olsonpm Can you show what *should* be output?

olsonpm:
As I said I'm open for discussion on what makes the most sense, but my vote leans toward the output of `git grep -C<number>`, where sections are delimited by a double dash.  E.g.

![asdf](https://cloud.githubusercontent.com/assets/5957709/16383688/caa21d24-3c4b-11e6-85f8-a4d285ec165b.png)

olsonpm:
@boneskull - trying to prevent the issue from being closed.  I provided the requested feedback, so at least remove that label please.

ScottFreeCode:
I'd be fine with just `--` as a separator if no one has better suggestions or other concerns. Do we want to put this behind a flag for compatibility? (Another advantage to a flag would be we could provide the ability to override `--` with some other non-dynamic string...)

boneskull:
[unexpected](https://npm.im/unexpected) may have a good solution we can pilfer; let's take a look at what their messaging looks like.

[...]

PR #2846, opened 2017-05-31
add unified diff separator

olsonpm:
- fix issue #2295
 - the purpose of this feature is to make the unified diff
   format more readable.  Without it, different portions
   of the same actual/expected diff look contiguous even
   though they are actually separated by the default
   context of four lines.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -404,7 +404,7 @@ function unifiedDiff (err, escape) {
       return indent + colorLines('diff removed', line);
     }
     if (line.match(/@@/)) {
-      return null;
+      return '--';
     }
     if (line.match(/\\ No newline/)) {
       return null;
@@ -415,7 +415,7 @@ function unifiedDiff (err, escape) {
     return typeof line !== 'undefined' && line !== null;
   }
   var msg = diff.createPatch('string', err.actual, err.expected);
-  var lines = msg.split('\n').splice(4);
+  var lines = msg.split('\n').splice(5);
   return '\n      ' +
     colorLines('diff added', '+ expected') + ' ' +
     colorLines('diff removed', '- actual') +
--- a/test/reporters/base.spec.js
+++ b/test/reporters/base.spec.js
@@ -14,6 +14,14 @@ function makeTest (err) {
   };
 }
 
+function createElements (argObj) {
+  var res = [];
+  for (var i = argObj.from; i <= argObj.to; i += 1) {
+    res.push('element ' + i);
+  }
+  return res;
+}
+
 describe('Base reporter', function () {
   var stdout;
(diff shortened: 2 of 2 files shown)


olsonpm:
I know the label said 'semver major' - but I didn't see a major branch to point the PR to.  I figured master was the correct one, but please let me know if that's incorrect.

test/reporters/base.spec.js
```diff
@@ -160,6 +168,46 @@ describe('Base reporter', function () {
...
+        /- {2}"element 11"/,
+        /]/,
+        /test/,
+        /expected/,
+        /actual/
+      ];
+
+      regexesToMatch.forEach(function (aRegex) {
+        errOut.should.match(aRegex);
+      });
```
ScottFreeCode: Instead of matching each line's regex against the whole output string, which does not guarantee order as far as I am aware, let's either match a single regex covering multiple lines, or else use `.forEach(function (aRegex, index) { <next line> errOut.split('\n')[index].should...` instead of `.forEach(function (aRegex) { <next line> errOut.should...`

lib/reporters/base.js
```diff
@@ -393,7 +393,7 @@ function inlineDiff (err, escape) {
  */
 function unifiedDiff (err, escape) {
   var indent = '      ';
-  function cleanUp (line) {
+  function cleanUp (line, idx) {
```
ScottFreeCode: The `idx` added here does not appear to be used?

test/reporters/base.spec.js
olsonpm: When I PR - I try to avoid optimizations like these and just stick with the patterns in the existing codebase.  I too thought this pattern was weird, but it's [what's in the code](https://github.com/mochajs/mocha/blob/master/test/reporters/base.spec.js#L149-L159).

If you really want me to make it different for this particular test I will, but I suggest refactoring them all in a separate PR for more accurate test coverage.

lib/reporters/base.js
olsonpm: Whoops - sorry about this.  An artifact of my debugging.

test/reporters/base.spec.js
ScottFreeCode: Ouch! My bad; I hadn't noticed those tests before.

In that case I'd be fine merging this as-is for consistency and then doing a separate thorough sweep to correct all instances of that pattern.

Thanks for identifying that!

olsonpm:
@ScottFreeCode - Just pushed the removal of the unused `idx` you pointed out.  All we need is a branch to push to now.

olsonpm:
i appreciate the time you spent with this.  i know it wasn't a high priority nor highly visible issue.

PR #2850, opened 2017-05-31
fixed array test name in mocha/test/browser/index.html

PopradiArpad:
The bug seems to be only a typo.


boneskull:
Yeah, we need to do something with these, but thanks!

PopradiArpad:
@ScottFreeCode,@boneskull  Thanks for your review and merge!

PR #2858, opened 2017-06-03
ignore lockfiles for now

boneskull:
This ensures we don't have to deal with `package-lock.json` and `yarn.lock` for now.


boneskull:
@ScottFreeCode It makes sense for "apps".

PR #2868, opened 2017-06-07
fix CI; closes #2867

### LINKED ISSUE #2867: fix false positives in CI
First bad revision is e00d27b.  Probably something to do with either Node.js 8 or npm v5.

boneskull:
I think the `package-lock.json` creation created by npm@5 was causing us to be unable to do `npm install` after an `npm install --production`.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -14,6 +14,15 @@ module.exports = function (config) {
       'expect',
       'mocha'
     ],
+    plugins: [
+      'karma-browserify',
+      'karma-chrome-launcher',
+      'karma-phantomjs-launcher',
+      'karma-expect',
+      'karma-mocha',
+      'karma-spec-reporter',
+      require('@coderbyheart/karma-sauce-launcher')
+    ],
     files: [
       // we use the BDD interface for all of the tests that
       // aren't interface-specific.
--- a/package.json
+++ b/package.json
@@ -320,6 +320,7 @@
     "supports-color": "3.1.2"
   },
   "devDependencies": {
+    "@coderbyheart/karma-sauce-launcher": "coderbyheart/karma-sauce-launcher#5259942cd6d40090eaa13ceeef5b0b8738c7710f",
     "assert": "^1.4.1",
     "browserify": "^13.0.0",
     "coffee-script": "^1.10.0",
@@ -337,7 +338,6 @@
     "karma-expect": "^1.1.2",
     "karma-mocha": "^1.3.0",
     "karma-phantomjs-launcher": "0.2.3",
-    "karma-sauce-launcher": "coderbyheart/karma-sauce-launcher",
     "karma-spec-reporter": "0.0.26",
     "nyc": "^10.0.0",
     "os-name": "^2.0.1",
--- a/test/jsapi/index.js
+++ b/test/jsapi/index.js
@@ -9,8 +9,7 @@ var mocha = new Mocha({
   growl: true
 });
 
-// mocha.reporter('spec');
-require('should');
+global.expect = require('expect.js');
 
 mocha.addFile('test/unit/suite.spec.js');
 mocha.addFile('test/unit/runner.spec.js');
--- a/test/unit/runnable.spec.js
+++ b/test/unit/runnable.spec.js
@@ -38,7 +38,7 @@ describe('Runnable(title, fn)', function () {
     it('should set the timeout', function () {
       var run = new Runnable();
       run.timeout(1000);
-      run.timeout().should.equal(1000);
+      expect(run.timeout()).to.equal(1000);
     });
   });
 
@@ -46,7 +46,7 @@ describe('Runnable(title, fn)', function () {
     it('should set disabled', function () {
       var run = new Runnable();
       run.timeout(1e10);
-      run.enableTimeouts().should.be.false;
+      expect(run.enableTimeouts()).to.be(false);
     });
   });
 
@@ -54,7 +54,7 @@ describe('Runnable(title, fn)', function () {
     it('should set enabled', function () {
       var run = new Runnable();
       run.enableTimeouts(false);
-      run.enableTimeouts().should.equal(false);
+      expect(run.enableTimeouts()).to.equal(false);
     });
   });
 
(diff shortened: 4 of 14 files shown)


boneskull:
See npm/npm#16967

boneskull:
Nope, that wasn't it.  I have no idea.  just gonna downgrade npm.

boneskull:
This may have been an incompatible "Git-style" dependency.

boneskull:
Trying to just downgrade Node.js itself back to v7 for browser tests...

boneskull:
@dasilvacontin @ScottFreeCode if this goes green please merge it unless I get to it first.  we should then be good to rebase the debug upgrade in branch `debug` onto `master`--at that point, it's probably worth cutting the release.

dasilvacontin:
Looks pretty green – but it's be nice to review it nevertheless. 😬

karma.conf.js
```diff
@@ -14,6 +14,15 @@ module.exports = function (config) {
...
       'mocha'
     ],
+    plugins: [
+      'karma-browserify',
+      'karma-chrome-launcher',
+      'karma-phantomjs-launcher',
+      'karma-expect',
+      'karma-mocha',
+      'karma-spec-reporter',
+      require('@coderbyheart/karma-sauce-launcher')
```
boneskull: The plugin is named 'karma-sauce-launcher', but the package is named something else, so we have to do this.  Unfortunately, if you start specifying plugins, you have to specify *all* of them.

package.json
```diff
@@ -320,6 +320,7 @@
     "supports-color": "3.1.2"
   },
   "devDependencies": {
+    "@coderbyheart/karma-sauce-launcher": "coderbyheart/karma-sauce-launcher#5259942cd6d40090eaa13ceeef5b0b8738c7710f",
```
boneskull: `karma-sauce-launcher` was a misnomer.  npm@3 and npm@4 install this anyway, but npm@5 does not; see npm/npm#17193.

I'm using the changeset SHA here to avoid pulling down any subsequent commits to `master`, since this is someone else's repo.

I have another branch wherein the Mocha org owns a fork, so we have complete control over how it behaves -- and what upgrades it receives.

(This was done in the first place because the original package is not-very-well maintained)

test/jsapi/index.js
```diff
@@ -9,8 +9,7 @@ var mocha = new Mocha({
   growl: true
 });
 
-// mocha.reporter('spec');
-require('should');
+global.expect = require('expect.js');
```
boneskull: As noted in #2882, `should` is problematic for non-ES5-compliant environments, so we can't run tests using `should` in those environments.

jsapi runs a subset of the unit tests, which now all use `expect` rather than `should`.

test/unit/runnable.spec.js
```diff
@@ -142,9 +142,24 @@ describe('Runnable(title, fn)', function () {
...
           test.run(function (err) {
-            calls.should.equal(1);
-            test.duration.should.be.type('number');
-            done(err);
+            if (err) {
+              done(err);
+              return;
+            }
+
+            try {
```
boneskull: this test was written in a way that assertions would still be made even if the callback received an error.

test/unit/utils.spec.js
```diff
@@ -498,25 +498,30 @@ describe('lib/utils', function () {
...
+      expect(isBuffer({}))
+        .to
         .equal(false);
     });
   });
 
   describe('map()', function () {
     var map = utils.map;
     it('should behave same as Array.prototype.map', function () {
+      if (!Array.prototype.map) {
```
boneskull: this comparison isn't possible in non-ES5-compliant environments

ScottFreeCode:
There are a couple points where this skips tests if `Object.create` is absent. In one case this seems to be using `Object.create` as a proxy for `defineProperty`, which is being used to get a non-writable stack on an error to test that Mocha is able to handle those; I'd prefer to have that one check whether `defineProperty` is available *and can create a non-writable property*, but I suppose what we've got is OK since the important thing is we test this in a modern environment where that works and don't attempt it in older environments where it would blow up for lack of `defineProperty`. In the other case, I am not actually sure what `Object.create` is being checked for -- is it needed for Mocha's behavior? If so, should we be attempting to polyfill that or something instead?

Other than that, this looks fine. I mean, I'm still not sure I understand what we're fixing, but the code after the changes looks OK. And, let's face it, switching from `should` to `expect` was long overdue. 🎉

test/unit/runner.spec.js
```diff
@@ -285,21 +294,26 @@ describe('Runner', function () {
...
       runner.on('fail', function (test, err) {
-        err.message.should.equal('the array [\n  1\n  2\n] was thrown, throw an Error :)');
+        expect(err.message).to.equal('the array [\n  1\n  2\n] was thrown, throw an Error :)');
         done();
       });
       runner.fail(test, err);
     });
 
     it('should recover if the error stack is not writable', function (done) {
+      if (!Object.create) {
```
boneskull: checking for `Object.defineProperty` seems reasonable, but it isn't really, because IE8 implements the function incorrectly.

test/unit/runner.spec.js
```diff
@@ -119,9 +125,12 @@ describe('Runner', function () {
     });
 
     it('should not fail when a new common global is introduced', function () {
+      if (process.browser) {
```
boneskull: This test makes some assumptions about global properties that we can't guarantee across browsers.

package.json
ScottFreeCode: Looking forward to switching to our fork, although I have some questions I need to get on there and ask about that too...

test/unit/runner.spec.js
boneskull: though I think the lines [here](https://github.com/mochajs/mocha/pull/2868/files#diff-84453848e82796cadaea28db8fca1ab2R84) might mean this never happens.

boneskull:
@ScottFreeCode @dasilvacontin I went thru and made a bunch of explanations.  please let me know if you're still unsure about something

boneskull:
the `Object.defineProperty` failure in IE8 would manifest itself as an exception, I believe.  it only works on DOM nodes.  so it exists, but...

To be clear, I'm not just handling quirks in IE7 and IE8; PhantomJS as well.

test/unit/runner.spec.js
boneskull: ..no, actually, we don't want this to run in *any* browser, regardless if it's ES5-spec-compliant.

ScottFreeCode:
Thanks!

Yeah, if it's going to be... complicated... to check that `Object.defineProperty` is not only available but correctly functioning, then we may as well just check for `Object.create` instead. Ideally we may want a comment in the code alluding to the issues checking for `Object.defineProperty`, but I'm not going to be a stickler about that.

I'm still unsure about the purpose of the other `Object.create` check -- the one in the `checkGlobals` suite I think it was.

boneskull:
@dasilvacontin Speak now or forever hold your peas

dasilvacontin:
I like my veggies.

PR #2870, opened 2017-06-08
FEATURE Allow comments in `mocha.opts` file.

igwejk:
This should enable switching an option off/on with ease by commenting/uncommenting the option.

It can be useful to document certain options to facilitate better collaborations.

/cc @jwalton, @ELLIOTTCABLE

### DIFF
--- a/bin/options.js
+++ b/bin/options.js
@@ -23,7 +23,7 @@ function getOptions () {
 
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
-      .replace(/^\s*#.*/gm, '')
+      .replace(/^\s*([^#]*)#.*/gm, '')
       .replace(/\\\s/g, '%20')
       .split(/\s/)
       .filter(Boolean)
--- a/bin/options.js
+++ b/bin/options.js
@@ -23,7 +23,7 @@ function getOptions () {
 
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
-      .replace(/^\s*([^#]*)#.*/gm, '')
+      .replace(/^\s*([^#]*)#.*$/gm, '')
       .replace(/\\\s/g, '%20')
       .split(/\s/)
       .filter(Boolean)
--- a/bin/options.js
+++ b/bin/options.js
@@ -23,6 +23,7 @@ function getOptions () {
 
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
+      .replace(/^\s*#.*/gm, '')
       .replace(/\\\s/g, '%20')
       .split(/\s/)
       .filter(Boolean)


igwejk:
There have been requests to reopen the closed #2587 PR.

bin/options.js
```diff
@@ -23,6 +23,7 @@ function getOptions () {
 
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
+      .replace(/^\s*#.*/gm, '')
```
kunagpal: This could be changed to:
```js
    .replace((/^(\S*)\s*#.*$/m, '$1')`
```
to better accommodate comments at the end of a line, like so:
```
--recursive # To dive to deeper depths
```

igwejk: That's an excellent idea, I'll apply it

igwejk:
The updated regex looks like this `/^\s*([^#]*)#.*/gm` and you can see a test [here](http://www.regexpal.com/?fam=97840)

I'll make another PR that tests the `getOptions( ... )` function.

ELLIOTTCABLE:
Just :-1: on supporting end-of-line comments, due to the complexity and validity of comment-styled filenames.

As an alternative possibility, and one I actually prefer — why not just declare *any* line that doesn't begin with `\s*-` a “comment?” This actually makes `mocha.opts` basically ‘literary programming’, in the Knuth style:

```
Allow for blah:
    --blah

… but not for widget!
   --no-widget

Use a pretty reporter, by default:
    --reporter spec
```

It's also trivial to implement, user-friendly (pretty hard to fuck it up — `#` comments *and* `//` comments would all work as expected; only Lua-style comments would fail, for obvious reasons), and easy to explain.

Just a thought.

ScottFreeCode:
> Just 👎 on supporting end-of-line comments, due to the complexity and validity of comment-styled filenames.

Fair enough argument there.

Anyone else want to weigh in for/against supporting end-of-line comments? I'd love to have a better idea of whether anyone even wants them.

> As an alternative possibility, and one I actually prefer — why not just declare any line that doesn't begin with `\s*-` a “comment?”

It really looks nice, but it's a non-starter unless we can come up with a way to differentiate comments from file globs that has basically no risk of the user writing something they intend as a path that Mocha interprets as a comment and ignores (thus not only behaving incorrectly but doing *nothing* to indicate what's wrong or why, which kills usability).

jwalton:
+1 for not supporting end-of-line comments.

danday74:
Knowing the difficulties with end of line comments (from #2963) I would vote for 'something is better than nothing' and focus on implementing # comments and or // comments.

I think // is a good idea because mocha is a JavaScript library.

However @ScottFreeCode raises a good point in #2963 when he says:

> My gut instinct is that the comment indicator should be # (as in #2870) as in shell scripting because mocha.opts is just a list of commandline arguments

I would thus suggest that everyone agrees to not support end-of-line comments and the last matter of debate is whether we use # or // or both

My vote is for both!

ELLIOTTCABLE:
@ScottFreeCode woap, shows what I get for speaking out of my ass! I didn't realize `mocha.opts` could include anything other than flags; ignore my comment entirely!

igwejk:
Hey I'm late to the party this time, but I'm here anyway.

> we could add comment support at the start of lines first and then later add comment support on the same line after options

I agree, we could indeed approach it like so.

@ScottFreeCode, I stated 

> I'll make another PR that tests the getOptions( ... ) function.

And duly followed up by writing a test for the `bin/options` module: #2881

boneskull:
closing b/c of inactivity

plroebuck:
How odd. I was just about to create new issue to allow for start-of-line-only '#' comment lines.
Though not completely following the whole what-else-can-go-in-"mocha.opts"-besides-cmdline-args discussion ScottFree and ELLIOT had above.

PR #2872, opened 2017-06-11
eslint fixes

EsrefDurna:


ScottFreeCode:
What exactly is this fixing? The current code is not having any problems with ESLint that I can tell; the changes, at a glance, appear to be all semicolon removals, but our ESLint configuration *requires* semicolons rather than the other way around.

PR #2874, opened 2017-06-11
Print failures for pending/only forbidden

ScottFreeCode:
Followup to #2696; this is going to look crazy when somebody puts `.only` or `.skip` on a suite, but I can live with that (whereas I can't take not having any printout for these failures at all) and if we figure out how to emit one failure per relevant suite we can always tweak that later (whether it has to be semver-major or not).

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -537,11 +537,10 @@ Runner.prototype.runTests = function (suite, fn) {
         test.isPending = alwaysFalse;
         self.fail(test, new Error("Pending test forbidden"));
         delete test.isPending;
-        self.emit('test end', test);
       } else {
         self.emit('pending', test);
-        self.emit('test end', test);
       }
+      self.emit('test end', test);
       return next();
     }
 
@@ -553,11 +552,10 @@ Runner.prototype.runTests = function (suite, fn) {
           test.isPending = alwaysFalse;
           self.fail(test, new Error("Pending test forbidden"));
           delete test.isPending;
-          self.emit('test end', test);
         } else {
           self.emit('pending', test);
-          self.emit('test end', test);
         }
+        self.emit('test end', test);
         return next();
       }
       if (err) {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -429,6 +429,10 @@ Runner.prototype.runTest = function (fn) {
   if (!test) {
     return;
   }
+  if (this.forbidOnly && this.hasOnly) {
+    fn(new Error("`.only` forbidden"))
+    return
+  }
   if (this.asyncOnly) {
     test.asyncOnly = true;
   }
@@ -529,8 +533,15 @@ Runner.prototype.runTests = function (suite, fn) {
     }
 
     if (test.isPending()) {
-      self.emit('pending', test);
-      self.emit('test end', test);
+      if (self.forbidPending) {
+        test.isPending = alwaysFalse;
+        self.fail(test, new Error("Pending test forbidden"));
+        delete test.isPending;
+        self.emit('test end', test);
+      } else {
+        self.emit('pending', test);
+        self.emit('test end', test);
+      }
       return next();
     }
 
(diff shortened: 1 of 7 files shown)


lib/runner.js
```diff
@@ -529,7 +533,13 @@ Runner.prototype.runTests = function (suite, fn) {
     }
 
     if (test.isPending()) {
-      self.emit('pending', test);
+      if (self.forbidPending) {
+        test.isPending = alwaysFalse;
```
ScottFreeCode: I'm not entirely happy with this hack to circumvent `Runner.prototype.fail`'s check that ignores pending tests based on `test.isPending()`, but I'm not sure if there's a better option. Any opinions would be appreciated, provided we can avoid stalling over the matter.

lib/runner.js
```diff
@@ -538,7 +548,13 @@ Runner.prototype.runTests = function (suite, fn) {
...
+        if (self.forbidPending) {
+          test.isPending = alwaysFalse;
+          self.fail(test, new Error("Pending test forbidden"));
+          delete test.isPending;
+        } else {
+          self.emit('pending', test);
+        }
         self.emit('test end', test);
         return next();
       }
```
ScottFreeCode: These two `if (test.isPending()) { ...` sections are identical or nearly so; we should consider breaking it out into a function or something.

lib/runner.js
ScottFreeCode: (Actually, I honestly wonder why `Runner.prototype.fail` needs to ignore pending tests. Wouldn't it be more correct not to call it on pending tests? Do we know where it's being called on pending tests so we could evaluate whether that could be adjusted instead?)

boneskull: A pending test is a failure, because a pending test (a synchronous one, anyway) actually throws an exception.  We catch that in the `Runnable`, then set the pending flag, then continue on to the "fail" handler in `Runner`.  At least, if memory serves...

boneskull: And `Runner.prototype.fail` is what reporters depend on for events...

boneskull:
@ScottFreeCode Is this needed to release alongside #2696?  In other words, can we release #2696 without it?

ScottFreeCode:
In my own opinion: yes.

In more detailed and less opinion: #2696 adds the options and a return code, but no output to actually indicate the reason for the failing return code. This adds the output. #2696, or #2696 plus this if the two are released together, is a new feature. We can release #2696 without this if we are willing to either:
- Consider this a bugfix as there should have been output to begin with, or...
- Make this semver-major as it significantly changes output.

(I don't have any strong preference which, so long as it happens; my stronger preference is that the decision not be necessary by releasing this and #2696 together as one feature.)

If it makes any difference:
- The options in question are likely to be used in CI, leading to CI failures with no message about why.
- The output added here not only indicates that these particular options were violated, but indicates which test(s) are at fault (albeit clumsily in the case of suites, as it will list every one of their tests rather than listing the suite or suites with the `only` or `skip` set).

lib/runner.js
ScottFreeCode: Hmm. The Pending exception is being handled in the `Runner` around [line 533](https://github.com/mochajs/mocha/pull/2874/files#diff-b3b53682a18f203ac8d29b0e277cad26L553), in which a `pending` event is emitted instead of going to the `fail` function at all. (The `fail` function ignores pending tests altogether, rather than emitting anything pending for them; hence the need to circumvent it in this line.)

The `Runnable` is only throwing (and in one case passing to `done`) `Pending` exceptions, not catching any (it also only emits an "error" event, in the case that `done` is called multiple times, so it's not handling the pending emission before sending an ignored error of some sort to the `Runner` either).

ScottFreeCode: Additionally, Coveralls indicates that the `fail` function's pending test ignore branch is never hit: https://coveralls.io/builds/11947829/source?filename=lib%2Frunner.js#L224 However, to confirm that we aren't just missing a test for this case, I'd have to go look at whether we have tests for all... five? six?... cases where tests can be pending: `describe.skip`, `it.skip`, `it("title and not callback")`, `this.skip` in `it`, `this.skip` in `beforeEach`, `this.skip` in `before`...

ScottFreeCode:
I'm toying with adding a test for `describe.only` as well (similar to my test for `describe.skip`). It's redundant with this implementation, where any `it` run when `only` is in effect is what will error, but may be nice to have explicitly tested so the test is already there if we ever change the implementation -- or if anything accidentally affects the implementation, for that matter.

It'd be nice if we could get the CI issues fixed first so pushing a commit with a test we know passes won't make everything go from green like it is now to red for no reason.

ScottFreeCode:
Added that last test previously mentioned, and it looks like we're passing CI here.

PR #2881, opened 2017-06-12
Improved options module for testability

igwejk:
- Wrote unit tests for the options feature
- Refactored options module for testability


igwejk:
Interestingly the test **locally** executes with success on Node 6.

igwejk:
@ScottFreeCode thought I should bring your attention to this now that we're on the topic.

bin/options.js
```diff
@@ -1,41 +1,67 @@
 'use strict';
```
boneskull: imo nothing should live in `bin` except the two executables.  maybe we should move this.

bin/options.js
```diff
@@ -1,41 +1,67 @@
...
-
 var fs = require('fs');
 
-/**
- * Export `getOptions`.
- */
+var defaultOptionsPath = 'test/mocha.opts';
 
-module.exports = getOptions;
+var options = Object.create(null, {
```
boneskull: what's the necessity of `Object.create()` and the properties here?

package.json
```diff
@@ -344,7 +345,8 @@
     "os-name": "^2.0.1",
     "phantomjs": "1.9.8",
     "rimraf": "^2.5.2",
-    "should": "^9.0.2",
+    "should": "^11.1.1",
```
boneskull: we should not be upgrading modules not associated with this pr

package.json
```diff
@@ -326,6 +326,7 @@
     "coffee-script": "^1.10.0",
     "coveralls": "^2.11.15",
     "cross-spawn": "^5.1.0",
+    "del": "^3.0.0",
```
boneskull: what's this for?

package.json
boneskull: actually I recall we dropped `should` entirely!

test/unit/options.spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+      argv = [];
+      sinon.spy(argv, 'indexOf');
+      path = options.path(argv, options.defaultOptionsPath);
+      expect(argv.indexOf.calledOnce).to.be.true;
+      expect(path).to.equal(options.defaultOptionsPath);
+
+      argv = ['--opts', '/user/passed/mocha.opts'];
+      sinon.spy(argv, 'indexOf');
+      path = options.path(argv, options.defaultOptionsPath);
```
boneskull: I don't really see the value in testing this

test/unit/options.spec.js
```diff
@@ -0,0 +1,90 @@
...
+    it.skip('reads the given path when it exists', function () {
+      var fs = require('fs');
+      var del = require('del');
+
+      var now = Date.now();
+      var text = '--opt-1 value-1\n--opt-2 value-2';
+
+      fs.mkdirSync('test-data');
+      fs.writeFileSync('test-data/mocha' + now + '.opts', text);
+      expect(function () {
```
boneskull: I don't understand the need for the nested assertion?

bin/options.js
igwejk: I agree, will review and change.

bin/options.js
igwejk: The aim is to collect and expose in a secure manner, the set of related _options_ units in a map. Using `Object.create` is a discipline. I believe there would be less JS bashing if the bashers applied this very discipline more.

package.json
igwejk: Must be a mistake, will fix.

test/unit/options.spec.js
igwejk: It doesn't hurt though? :-)

test/unit/options.spec.js
igwejk: Sure we can do without it, but it does help with organising the report. Doesn't the following output make good sense?

image

bin/options.js
igwejk: @boneskull I recommend we move the _options_ module from the `bin` directory in a different change. Does that sound good to you as well?

package.json
igwejk: In the module `options.spec.js` (see lines `:58` and `:66`), a test data was created and `del` is used to remove it before the running test concludes.

test/unit/options.spec.js
igwejk: I have dropped `sinon` and assertions that depend on it.

igwejk:
@boneskull I've made the changes due to your review.

test/unit/options.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      expect(path).to.equal('/user/passed/mocha.opts');
+    });
+  });
+
+  describe('#read', function () {
+    var options = require('../../lib/options');
+
+    it('throws when given a path that does not exist', function () {
+      // 1. no path is passed
+      expect(function () {
```
boneskull: unless there's a function context involved, this can be simplified to
```js
expect(options.read).to.throwError();
```

test/unit/options.spec.js
```diff
@@ -0,0 +1,87 @@
...
+
+    it('property should have value \'test/mocha.opts\'', function () {
+      expect(options.defaultOptionsPath).to.equal('test/mocha.opts');
+    });
+
+    it('property should NOT be writable/configurable', function () {
+      expect(function () {
+        options.defaultOptionsPath = '/new/path/mocha.opts';
+      }).to.throwError(function (error) {
+        expect(error).to.be.a(TypeError);
```
boneskull: eww... this is why I never use `expect.js` haha

lib/options.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+var fs = require('fs');
+
+var defaultOptionsPath = 'test/mocha.opts';
+
+var options = Object.create(null, {
```
boneskull: just so it doesn't get lost, we don't need to set properties and use `Object.create()`; let's just make a plain old object.

it's a private API, and if someone is going to monkey with it, then that's their problem.  we don't need to lock down the values via `writable: false`.  IIRC there isn't a precedent for this in the codebase.

lib/options.js
```diff
@@ -0,0 +1,71 @@
...
+    .filter(Boolean)
+    .map(function (value) {
+      return value.replace(/%20/g, ' ');
+    });
+}
+
+function load () {
+  var options, optionsPath, optionsFileText;
+
+  if (process.argv.length === 3 && (process.argv[2] === '-h' || process.argv[2] === '--help')) {
```
boneskull: wait, what, we only display help if `--help` is the last argument?  ergh. (not your fault)

test/unit/options.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      });
+    });
+  });
+
+  describe('#path', function () {
+    var options = require('../../lib/options');
+
+    it('should return user passed options path if it exists, or the default path', function () {
```
boneskull: please split into two tests; we try to do 1 assertion per test

test/unit/options.spec.js
boneskull: yes, but there's no need to wrap an assert in *another* assert:

```js
// this:
expect(() => expect(someAssertion)).not.to.throwError();
// should just be:
expect(someAssertion);
```
I *think* what you may be getting at here is something like `try` / `finally` to cleanup the `test-data` dir, which can be accomplished via `afterEach()` -- use another suite if necessary.

test/unit/options.spec.js
```diff
@@ -0,0 +1,87 @@
...
+    var options = require('../../lib/options');
+
+    it('parse the given option', function () {
+      var opts = options.parse('--opts-1 value-1\n--opts-2 value-2');
+
+      expect(opts).eql(['--opts-1', 'value-1', '--opts-2', 'value-2']);
+    });
+
+    it('throws when passed an argument that is not a string', function () {
+      expect(function () {
```
boneskull: like above can be simplified to:

```js
expect(options.parse).to.throwError()
```

boneskull: sorry this review is kind of haphazard!

boneskull:
cc @mochajs/core I think we can ditch the `pr-needs-work` label and just use github reviews instead.

test/unit/options.spec.js
igwejk: Good to know, that's way more succinct.

test/unit/options.spec.js
igwejk: I see what you mean now. Didn't notice when I wrote it originally and didn't notice yesterday neither.

lib/options.js
igwejk: Didn't take a critical look at that during merge resolution. Will fix in a different PR.

craigtaub:
@igwejk have you had any time to look at this again?
Having a look not sure the lockfile update is required and as options now lives in `lib` believe it will need correct JSDoc access tags (in this case `@private`).

igwejk:
Just back from my slumber... Sorry I couldn't follow up on this since it kinda went beyond its scope. Thanks everyone for your inputs and resolving the issue (albeit through another PR).

PR #2886, opened 2017-06-15
Cover Browser Tests (and all Node versions)

### LINKED ISSUE #2883: in-browser coverage numbers
We have coverage numbers, but only in Node.js.  I'd like to see the coverage numbers against tests run with Karma.  

Gathering coverage from Karma seems fraught with peril.

I would entertain other suggestions to improve how we run tests in a browser environment.  My only requirement is that the tests run in a strict browser-like environment--`jsdom` shenanigans or Electron won't do.

ScottFreeCode:
Fixes #2883.

Depends on #2868.


ScottFreeCode:
Most of the commits in this should disappear after #2868 is merged.

ScottFreeCode:
We keep getting Sauce 20,000ms disconnects on Edge, IE7 and Safari... Not sure if that's actually related to any of these changes or not.

ScottFreeCode:
Relative coverage has gone down because I made the browser coverage include browser-only files that the Node coverage was ignoring -- so there's more source being counted in the first place now. The current counts are "2097 of 2439 relevant lines covered (85.98%)" and "824 of 1038 branches covered (79.38%)" whereas, if you click on "PR base - master" you see "1988 of 2249 relevant lines covered (88.39%)" and "774 of 955 branches covered (81.05%)" -- so that's actually an increase of 109 covered lines and 50 covered branches (it just happens there are even more newly counted lines/branches that were never covered). Furthermore, if you look through the list of changes on the PR coverage page, none of the files had decreases in lines or branches covered.

The browser files just haven't had coverage to begin with; but we can measure and start correcting that now!

ScottFreeCode:
Thinking of rebasing onto `master` to:
- get rid of extraneous commits from the CI-fixes branch in the history (instead just add the coverage-related changes against `master` now that `master` has the CI fixes merged in), thus ensuring that when this is merged it doesn't do anything weird to the overall commit history
- correct the bug in the couple of bugged commits rather than having a separate commit for bug-correction
- move the timeout adjustment to before the browser coverage

The latter two would allow both the browser coverage commits to each be tested on Travis and show valid results, demonstrating any changes in behavior from the first of the two browser coverage commits to the second, which *may* tell us something about the Sauce 20,000ms disconnect problem.

I will probably create a copy of this branch beforehand so we can bring it back if I goof up the rebase in any way.

ScottFreeCode:
Ok, so, this is exactly what I wanted to be able to see:
- The commit *before* adding coverage to Karma has those Sauce 20,000ms disconnect errors on Edge, IE7 and Safari.
- It only has them on the Travis PR build and not the Travis Push build.

That means, among other things:
- It doesn't seem to be the addition of coverage that's at fault. Yay!
- Something's different between this code in and of itself and this code pulled onto master? Despite the fact that this was just rebased onto master?? How is that possible?!

ScottFreeCode:
Why is Snyk flagging this PR? I click on "details" and I get a page showing the vulnerable dependencies we had before this PR. (And I double-checked, they're not also dependencies of the dependencies added in this PR.) Furthermore, the only dependency changes in this PR are to dev dependencies...

ScottFreeCode:
Rebased onto master to get various fixes, but it seems the flakiness is still prevalent with coverage instrumentation (even though past commits demonstrated that it happened on this branch *before* adding instrumentation), and apparently Phantom is now even slower?

ScottFreeCode:
See #2890 concerning the test failures. Adding that to my follow-up list.

boneskull:
I'd like to rebase this and test it again.

PR #2860, opened 2017-06-03
upgrade to debug@2.6.8; closes #2859

boneskull:
- ensure we can run the test suite with `DEBUG=mocha*` by removing it
  when spawning `mocha`
- remove some cruft in the test helpers

PR #2889, opened 2017-06-17
Stop timing out

ScottFreeCode:
Corrects a bug in #2860


ScottFreeCode:
Cancelling the Travis builds from before this and the Travis builds from this (waited to see that at least one passed to confirm the fix worked), so that the Travis build for the merge into master can go head sooner.

ScottFreeCode:
(That's this build, in case anyone's wondering: https://travis-ci.org/mochajs/mocha/builds/243891881)

PR #2897, opened 2017-06-21
Fix Node 0.10 AppVeyor failure

ScottFreeCode:
During the `debug` package upgrade, a tweak to the integration test helper's spawn usage was applied that cleared the environment variables. For whatever reason, that worked on every OS and version of Node except for 0.10 on Windows. This PR reverts that particular tweak, and nothing else, allowing CI to pass again.

If not clearing the environment variables is still an issue for some usage, we should attempt a new, separate fix that we can wait on merging till we can get it to avoid breaking anything.

(It is not immediately clear to me whether the breakage is due to lack of environment variables and could be fixed just by changing `{ env: {} }` to something old-Node-compatible equivalent of `{ env: Object.assign({}, process.env, { DEBUG: ""}) }`, or whether the breakage is instead due to some other effect of passing the options object argument to `spawn` -- perhaps in Node 0.10 for Windows it unset some other default setting?)

We may wish to switch back to `child_process` instead of dev-depending on `cross-spawn` if the reason `cross-spawn` was added was an attempt to fix these Node 0.10 AppVeyor failures (it did not).

### DIFF
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -115,10 +115,10 @@ module.exports = {
 
 function invokeMocha (args, fn) {
   var output, mocha, listener;
-  // ensure DEBUG doesn't kill tests
+
   output = '';
   args = [path.join('bin', 'mocha')].concat(args);
-  mocha = spawn(process.execPath, args, {env: {}});
+  mocha = spawn(process.execPath, args);
 
   listener = function (data) {
     output += data;


ScottFreeCode:
There appear to be Travis failures only in the browser tests, which should not be affected by this PR. Ironically, they're not even the same browser failures we were having a few days ago... This time it looks like something changed in browserify or Karma that breaks Phantom 1.x and old IE. If that's the case, it may really be time to make another semver-major and drop old browsers. (Although I don't know whether we consider Phantom 1.x "old" or not; I can't remember off the top of my head why we don't use Phantom 2.)

ScottFreeCode:
Well `npm view karma time` and `npm view browserify time` doesn't show any update in the past few days; perhaps one of their dependencies... but I don't really know how to check that for all of them quickly...

ScottFreeCode:
#2898 fixes the Travis browser failures seen here.

boneskull:
merged via 2408d90

PR #2898, opened 2017-06-21
Fix Travis browser tests dependency

ScottFreeCode:
Does NOT fix the Sauce 20,000ms disconnect issue.

Does fix the newer issues that appeared just this past day.

This fixes Travis browser tests (except for the Sauce 20,000ms disconnect issue), #2897 fixes AppVeyor Node 0.10 tests; put them both in master and then merge into anything else that's waiting for CI checks to pass, then hope the Sauce 20,000ms issue stays away for a while, ha.

(For the record: the issue is that readable-stream 2.3.0 breaks Phantom 1.x due to use of `Function.prototype.bind` and breaks IE7 and 8 due to pulling in `safe-buffer`, which uses `Object.keys`.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -343,6 +343,7 @@
     "nyc": "^10.0.0",
     "os-name": "^2.0.1",
     "phantomjs": "1.9.8",
+    "readable-stream": "2.2.11",
     "rimraf": "^2.5.2",
     "should": "^9.0.2",
     "through2": "^2.0.1",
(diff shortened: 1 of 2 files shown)


ScottFreeCode:
This may end up superseded by https://github.com/nodejs/readable-stream/issues/295 and https://github.com/feross/safe-buffer/issues/15

ScottFreeCode:
If we do end up specifying version 2.2.11 or 2.2.x of readable-stream ourselves as this PR does, I'd really prefer a lockfile over this quick hack. I threw this together without trying to figure out how to create a lockfile that uses an older-than-latest version of a dependency for the time it's created, and without waiting to follow up on the recent concerns as to whether lockfiles are workable for Mocha.

Issue #2893, opened 2017-06-20
browser reporter suite unhide is broken


ScottFreeCode:
Found out exactly what's going on in #2754: the HTMLCollection returned by `document.getElementsByClassName`, at least in some browsers, actually drops elements that get their `className` changed such that they no longer match the criteria on which the collection was based; thus, Mocha will only unhide the first half of the hidden suites.

#2754 attempted to fix this by converting the collection to an array. Currently Mocha runs in some pretty old environments. If we have a shim for the array `slice` method we should use that rather than assuming the native `[].slice`. If we don't have such a shim, we should simply reverse the `for` loop: start at `.length - 1` and decrement while `>= 0` instead of starting at `0` and incrementing while `< .length`.

(We should also get a test for the unhiding of more than one suite. I will incorporate this into my work on testing the HTML reporter and mark it as `.skip` until a fix can also be PR'd.)

I'd like to thank @zerobinbin for bringing this to our attention, and apologize for taking so long to get around to digging into it.

ScottFreeCode:
This could probably be fixed with array methods now that Mocha 4 supports only newer environments? In any case, it's mostly waiting on #2900, which I need to be able to devote some undivided time to per https://github.com/mochajs/mocha/issues/2947#issuecomment-338714711

PR #2900, opened 2017-06-22
Unit-test: the Last Reporter

ScottFreeCode:
Everything changed when the browser got coverage.


ScottFreeCode:
Depends on (built on top of) #2886

Also fixes #2893

test/browser-reporters/html.spec.js
```diff
@@ -0,0 +1,1918 @@
...
+        runner.emit('test end', rootSuite.suites[0].tests[0]);
+        runner.emit('suite end', rootSuite.suites[0]);
+        runner.emit('suite', rootSuite.suites[1]);
+        // no 'test' event is emitted for pending tests
+        runner.emit('pending', rootSuite.suites[1].tests[0]);
+        runner.emit('test end', rootSuite.suites[1].tests[0]);
+        runner.emit('test', rootSuite.suites[1].tests[1]);
+        rootSuite.suites[1].tests[1].pending = true;
+        rootSuite.suites[1].tests[1].duration = 0;
+        runner.emit('pending', rootSuite.suites[1].tests[1]);
```
ScottFreeCode: Could have pulled in the real `Runner`, but then this would have been more like an integration test than a unit test... (I made a tradeoff to reuse the real `Suite` and `Test` objects mainly because I felt that the code would be too bloated if I tried to recreate them completely but too brittle if I tried to recreate just the bits and pieces that the HTML reporter currently happens to be relying upon.)

test/browser-reporters/html.spec.js
```diff
@@ -0,0 +1,1918 @@
...
+    return tmp !== realOutput ? tmp : undefined;
+  }
+
+  before(saveRealOutput); // so that the following afterEach is guaranteed to work
+
+  afterEach(function () { restoreOutput(); }); // WARNING: this is solely a fallback so *further* tests can be guaranteed to report correctly; these tests will not be reported correctly unless `restoreOutput` is called at the end of the test, before throwing any errors
+
+  it('should show an error if the tag is missing', function () {
+    try {
+      // TODO: Figure out why it's ok to add an element to display this error but not ok to just add the missing element in the first place.
```
ScottFreeCode: We can feel free to remove this comment and replace it with a new GitHub issue. I just needed to jot it down while it was on my mind but didn't want to get sidetracked from the work at hand.

test/browser-reporters/html.spec.js
```diff
@@ -0,0 +1,1918 @@
...
+      realOutput.id = 'mocha';
+      realOutputStats.id = 'mocha-stats';
+      realOutputReport.id = 'mocha-report';
+    }
+    return tmp !== realOutput ? tmp : undefined;
+  }
+
+  before(saveRealOutput); // so that the following afterEach is guaranteed to work
+
+  afterEach(function () { restoreOutput(); }); // WARNING: this is solely a fallback so *further* tests can be guaranteed to report correctly; these tests will not be reported correctly unless `restoreOutput` is called at the end of the test, before throwing any errors
```
ScottFreeCode: The comment here might not be true when running in Karma and not opening the debug page -- I don't know enough about how it works. I am certain that it is crucial for using the HTML reporter to report the HTML reporter tests, however (e.g. in a manual page such as the one added in this commit as well, or in Karma's debug usage of Mocha's HTML reporter). I'm considering updating the comment to clarify this. It's the explanation for the try-catching in all the tests, so I hope it's clear enough.

karma.conf.js
```diff
@@ -34,7 +34,9 @@ module.exports = function (config) {
       // we use the BDD interface for all of the tests that
       // aren't interface-specific.
       'test/browser-fixtures/bdd.fixture.js',
-      'test/unit/*.spec.js'
+      'test/unit/*.spec.js',
+      'test/browser-unit/*.spec.js',
```
ScottFreeCode: Planning on filling `test/browser-unit` with tests for `lib/browser/*.js`, especially any that don't just end up covered by extension of being used in other tested stuff.

karma.conf.js
```diff
@@ -34,7 +34,9 @@ module.exports = function (config) {
...
+      'test/browser-reporters/*.spec.js'
     ]),
```
ScottFreeCode: We could put the HTML reporter test in the browser unit folder (it's not like we're planning on adding multiple browser reporters in core), but this is more obvious by parallel of `test/reporters`.

ScottFreeCode:
I think the Snyk error may be because `master` has gotten an update to resolve the `debug` "vulnerability" (not that it's a real vulnerability, it's using the `ms` function that is not reportedly affected...), even though this PR doesn't introduce that "vulnerability" either.

ScottFreeCode:
As this builds on #2886, it will need to be rebased from the old #2886 commits (last commit was 58e4b5d93feeb6f8ff3e24dd27d7a9d7a4889fd6), at some point at least after we find ways to resolve the CI issues for #2886 (rebase onto the new #2886 in that case) and preferably after merging #2886 (rebase onto `master` in that case).

Issue #2529, opened 2016-10-10
Can't see how retried test failed before passing.


JustinRainwater:
It would be great when using the retries feature if I could get the exception and stack trace info on the failed attempts before it eventually passes. 

Currently it looks like `this.currentTest.err` remains undefined unless the last retry is a failure. Instead when going through `afterEach()` before retrying the test I'd like to be able to log the contents of `this.currentTest.err`.

JustinRainwater:
Playing around locally it looks like this could be resolved just by adding `test.err = err` after this line: https://github.com/mochajs/mocha/blob/master/mocha.js#L5039

boneskull:
It might be difficult to do this in a way that works with all of the reporters.  Which reporter are you using?

JustinRainwater:
I wouldnt expect this to work with any reporters unless we created a new state like "Passed After Retry" to go with Passed and Failed. I'm just console logging this info so I can parse it in Jenkins console output. I originally wanted to differentiate retried tests in the reporter but that doesnt seem possible.

To answer your question we're using https://github.com/juhovh/mocha-jenkins-reporter when running on jenkins and when running locally we just write to console. I'm not sure if that is default spec reporter or what.

If we wanted this to work in reporters then I could imagine test.err being an array where each index would store the error output for a specific retry so the reporter could dump all of the errors related to the retries. I'm not sure that is all worth it though. It makes sense that the reporter would report the ultimate result of the test "Passed/Failed" and then within the test run we can log details related to tests passing after retry.

LToronto:
I also need this feature for a custom reporter.
see https://github.com/mochajs/mocha/pull/2901

example:
runner.on('test', function (test) {
    if(test._currentRetry > 0) {
      reportErr(test, test.retryErr[test._currentRetry-1]);
    } 
});

catdad:
I am also trying to take a crack at this, as it has recently come up with a large legacy test suite I am working on.

From what I have gathered, it's really up to the reporters to decide what to do with passing or failing tests. In the case of mocha and retries, the runner simply ignores the failure of a test when it is possible to retry it,  clones that test, and adds it back to the tests array to be executed again. This is a fine approach, but it completely discards the failure information, so a reporter cannot take action on it.

I think rather than having mocha do any magic to include retried test information, it should simply make that information available to reporters, allowing reporters to choose what to do with that information.

[...]

PR #2901, opened 2017-06-22
Add err to test.retryErr list if retrying the test #2529

LToronto:
Adds the ability to see the errors from earlier attempts.  Occasional
the test state is different during the retry and the last test error is
not helpful.  Is this case the first error (test.retryErr[0]) is
useful.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -556,6 +556,8 @@ Runner.prototype.runTests = function (suite, fn) {
           } else if (retry < test.retries()) {
             var clonedTest = test.clone();
             clonedTest.currentRetry(retry + 1);
+            clonedTest.retryErr = test.retryErr || [];
+            clonedTest.retryErr.push(err);
             tests.unshift(clonedTest);
 
             // Early return + hook trigger so that it doesn't


LToronto:
It looks like node version 0.10 is broken in master.  There's nothing I should do here to fix it.

ScottFreeCode:
Hi @LToronto, thanks for the PR! And sorry about the delay responding to it.

One question I have about this is whether it would suffice to emit an event when a test is retried, that would contain the failure info such that you could save it if you want to or print it at the time it happens (rather than having to wait until the last test attempt ends to get the list of retried failures). While that's a little more work on the part of the consumer of that info, I believe it serves a superset of the use cases of this approach; I've also discussed that aspect in https://github.com/mochajs/mocha/issues/2529#issuecomment-315510743 and that other approach should be implemented by #2926 if you want to check it out and try it.

(By the way, the Node 0.10 failure should be fixed in master now if you want to rebase -- or add tests.)

LToronto:
Closed in favor of https://github.com/mochajs/mocha/pull/2926

PR #2904, opened 2017-06-28
Test: add bdd interface test case, coverage increased

Gerhut:


boneskull:
@Gerhut thanks!

failure was due to unsupported version of node.js

PR #2908, opened 2017-07-01
Tap reporter: report thrown error messages

chrmod:
This PR extendis tap reporter to print error messages. It is useful in faster understanding on what exactly when wrong.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -46,6 +46,9 @@ function TAP (runner) {
   runner.on('fail', function (test, err) {
     failures++;
     console.log('not ok %d %s', n, title(test));
+    if (err.message) {
+      console.log(err.message.replace(/^/gm, '  '));
+    }
     if (err.stack) {
       console.log(err.stack.replace(/^/gm, '  '));
     }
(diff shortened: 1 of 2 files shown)


chrmod:
@boneskull what can be done to get this one in?

boneskull:
@chrmod Does this output adhere to the TAP specification?

boneskull:
Since our tap reporter is meant to be machine-consumable, any change to its output is a breaking one.

chrmod:
@boneskull I would say error messages fall unto same specification clause as the stack trackes - that is anything else https://testanything.org/tap-specification.html#anything-else 
So it is up to test runner to decide what to do with them. 
I've been using such modified test reporter with testem for quite some time now. Without error messages, some test failures may be hard to understand. 

Do you see a way we could get this upstream?

boneskull:
@chrmod Given that I don't use the `tap` reporter, and am unsure how popular it is, and *you* find this change useful, I don't have any good reason *not* to merge this (though it must be released in a major).  :smile:

This needs to be rebased onto current `master` so the build can be run again.  Can you please do that?

boneskull:
Continued in #3317.

PR #2912, opened 2017-07-05
Fix spec paths in test HTML files

eugenet8k:
* Finish the work started at 9bd9389c58e81b10bef9f4b2e5a54b77a026d2d0
* Fix Array test case. It’s not clear why `1` expected to be `-1`.

### DIFF
--- a/test/browser/array.spec.js
+++ b/test/browser/array.spec.js
@@ -23,7 +23,7 @@ describe('Array', function () {
 
 describe('Array', function () {
   describe('#pop()', function () {
-    it('should remove and return the last value', function () {
+    it('should remove and return the last value with expected error', function () {
       var arr = [1, 2, 3];
       assert(arr.pop() === 3);
       assert(arr.pop() === 2);
--- a/test/browser/index.html
+++ b/test/browser/index.html
@@ -12,8 +12,8 @@
       }
     </script>
     <script src="array.spec.js"></script>
-    <script src="../acceptance/duration.spec.js"></script>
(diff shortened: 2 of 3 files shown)


ScottFreeCode:
The failing assertion was actually an intentional demonstration of failed tests: https://github.com/mochajs/mocha/commit/e12ff8f7b634fc2ba4715fd0b68758030773e4dc

Other than that, the path changes look good to me.

AppVeyor failure is [unrelated](/mochajs/mocha/pull/2897).

Travis browser failures are [also unrelated](/mochajs/mocha/issues/2913).

eugenet8k:
@ScottFreeCode thanks for the information. It was confusing to see the error there because test description doesn't mention that error is expected.

Do you think it's reasonable to mention the error expectation? Something like `'should remove and return the last value with expected error'`. Other tests where an error is expected have similar descriptions, like `'should report an error if done was called more than once'` in [multiple-done.spec.js](https://github.com/mochajs/mocha/blob/master/test/browser/multiple-done.spec.js).

ScottFreeCode:
Updating the title/description instead of the behavior definitely sounds reasonable for now; thanks!

Also, if I haven't already said so, thanks for the path updates!

@mochajs/core Does anyone have any serious concerns with the proposed wording or with putting in the path fixes and the test clarification together rather than separately? If not then I believe this should be good to merge.

ScottFreeCode:
Seeing as it's been long enough for the bot to call this stale and no one has raised any concerns, I'm going to go ahead and put this in!

Thanks again!

PR #2918, opened 2017-07-09
Use JS for globbing integration test instead of bash shell

### LINKED ISSUE #2805: Eliminate shell scripts from Mocha's test suite

ScottFreeCode:
Closes #2805


ScottFreeCode:
I've been able to confirm on my Windows box that Node 0.10.x inconsistently fails the error-only glob tests due to getting an empty string for `stderr`. I am going to see if I can find a reason why this happens in order to cook up a proper fix; but if I can't find anything I may just make those tests have a retry count when run under Node 0.10 on Windows.

ScottFreeCode:
Noted that when Mocha doesn't find any files it calls `process.exit(1)`.

Tried calling `child_process.exec` on an exhaustive combination of:
- `node bin/mocha -R json-stream "./**/*-none.js"` (aka same as the integration test)
- same but `_mocha` instead of `mocha` (to confirm it's not just the `mocha` wrapper around `_mocha`)
- with `--no-exit` added (but this probably did nothing, see above)
- a script that just prints the same warning message to stderr and calls `process.exit(1)`
- same but with the `process.on("exit", function() { process.exit(1) })` logic from `--no-exit` (the `exitLater` function to be precise)
- same but with the draining logic from Mocha's usual exit path
- same but with the draining logic except calling the `exitLater` function instead of `process.exit` at the end
- same but with `process.on("exit"` calling the draining function instead of `process.exit`
- same but without calling anything to exit (just letting the script end after printing the message)

In all cases, if I ran this 15 times, eventually it would get an empty string for stderr on Node 0.10 on Windows.

That tells me for sure it is a Node issue and not anything we're doing wrong, since no strategy for exiting can prevent the problem from happening and it's reproducible in a trivial script too; so I am going to brute-force ignore it -- but with a slight twist... rather than just setting the tests to retry, I am going to make the exec-helper function that's already in there robust against getting an empty string for stderr on Node 0.10 on Windows (have it recursively call itself again when that happens). I've even confirmed this workaround is correct by running the test locally a few times and seeing that occasionally the tests that used to sometimes fail instead sometimes take twice as long (implying the integration test call was run twice).

ScottFreeCode:
For the record, here are the passing tests:
- https://ci.appveyor.com/project/boneskull/mocha/build/1019
- https://ci.appveyor.com/project/boneskull/mocha/build/1020
- https://travis-ci.org/mochajs/mocha/builds/260034468
- https://travis-ci.org/mochajs/mocha/builds/260034489

Going to merge this such that the original commits are retained, in case anyone wants to examine the different steps (conversion from makefile+shell to integration JS, addition of more complete globbing combinations to the test, and finally addition of Node 0.10 Windows workaround).

PR #2920, opened 2017-07-12
Ignore difference of newline output in diffing test

### LINKED ISSUE #2801: Line endings and test files
Similar to #2800, while the test suite can be run on Windows (e.g. using MinGW make with either Cygwin or Git Shell), certain tests fail if certain files do not have the right line endings:
- ~`test/acceptance/glob/glob.sh` -- this just needs to be Linux line endings, should probably be converted to them by the test run if possible regardless of how they may have been changed~
- something in `test/integration/fixtures/diff` -- perhaps just the `output` file? and not sure if the right line endings here are always Linux, or depend on the OS, or depend on whether the tests are being run from a Posixy environment (as is necessary to use the Makefile)

More meta-like, if this ends up requiring per-file fiddling, then it would be really, really helpful to have some way of guaranteeing that our CI runs won't succeed just because they happen to have the right global settings that developers may or may not have. It doesn't make sense to have something that we need to set up correctly for certain tests that won't fail CI if we get it wrong.

ScottFreeCode:
Fixes #2801 the brute-force but simple and easy way without refactoring the test design.

### DIFF
--- a/test/integration/diffs.spec.js
+++ b/test/integration/diffs.spec.js
@@ -7,7 +7,7 @@ var fs = require('fs');
 var getDiffs = helpers.getDiffs;
 
 function getExpectedOutput () {
-  var output = fs.readFileSync('test/integration/fixtures/diffs/output', 'UTF8');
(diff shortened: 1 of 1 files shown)


ScottFreeCode:
CI for this is green after a rebase.

PR #2921, opened 2017-07-12
Simplify publishing process

ScottFreeCode:
Eliminates need for `--git-tag-version` in the `npm version` step/command.

Should eliminate need for an explicit push-to-github step and ensure that if it goes to NPM, it goes to GitHub too.


ScottFreeCode:
All checks should be passing; the one that isn't shouldn't even be applied because of `[ci skip]`...

AppVeyor's PR mode seems to be acting up -- this isn't the first perfectly mergeable PR for which AppVeyor has claimed it can't build a "non-mergeable" PR... Although if I can find which other PR was affected by that issue, it might be worth looking at whether any of the commits include `[ci skip]`, in case it's a bug in AppVeyor's `[ci skip]` handling.

boneskull:
@ScottFreeCode I'm confused about what this `git-tag-version` thing is.  When you run `npm version`, it will tag *by default*, won't it?

Pushing on post-publish seems reasonable, though.

PR #2922, opened 2017-07-12
Test compilers lookup

ScottFreeCode:
Ensure that `--compilers` file lookup behavior works and not just that transpilation works.


ScottFreeCode:
All tests passed with the rebase; merging.

PR #2798, opened 2017-05-14
chore(package): bump growl to 1.9.3

faust64:
As suggested by mochajs/mocha#2791, try bumping growl to 1.9.3

Issue #2791, opened 2017-05-08
Vulnerability in dependency growl 1.9.2


balsamiqFlorian:
Hi,

I got a report that mocha includes growl 1.9.2 which is vulnerable to an arbitrary code injection.
https://snyk.io/vuln/npm%3Agrowl%3A20160721?utm_campaign=vuln_alert&utm_medium=email&utm_source=Vuln

This issue is closed in growl 1.9.3.

Thanks.

ScottFreeCode:
Is there a way we can tell Snyk that Growl 1.9.3 does not in fact exist, and that the commit they reference is only in a fork (so it is possible that in the future Growl 1.9.3 could be published before fixing this issue, thus rendering their instructions incorrect once 1.9.3 does exist)?

If anybody wants to look at whether Mocha is actually at risk from this, I believe these would be the lines to examine: https://github.com/mochajs/mocha/blob/8a3cab063cfe68fa2e9aea8feada23aaf1a4db6f/lib/mocha.js#L241-L257

andris310:
Got same warning in Node Security Platform https://nodesecurity.io/advisories/146

Fzzr:
There is no growl 1.9.3, the issue is not actually fixed. See issue https://github.com/tj/node-growl/issues/60 and PR https://github.com/tj/node-growl/pull/61

papakai:
Got same warning using retire.js

davorivas:
I am getting the same error and my tests in circleCI are not passing because of this "vulnerability"


(+) 1 vulnerabilities found
┌───────────────┬───────────────────────────────────────────────────────┐
│               │ Command Injection                                     │
├───────────────┼───────────────────────────────────────────────────────┤
│ Name          │ growl                                                 │
├───────────────┼───────────────────────────────────────────────────────┤
│ CVSS          │ 9.8 (Critical)                                        │
├───────────────┼───────────────────────────────────────────────────────┤
│ Installed     │ 1.9.2                                                 │
├───────────────┼───────────────────────────────────────────────────────┤
│ Vulnerable    │ All                                                   │
├───────────────┼───────────────────────────────────────────────────────┤
│ Patched       │ None                                                  │
├───────────────┼───────────────────────────────────────────────────────┤
│ Path          │ my-project > mocha@3.4.2 > growl@1.9.2 │
├───────────────┼───────────────────────────────────────────────────────┤
│ More Info     │ https://nodesecurity.io/advisories/146                │
[...]

PR #2924, opened 2017-07-14
chore(package): bump growl to 1.10.0

faust64:
Following-up on mochajs/mocha#2798 & mochajs/mocha#2791, update growl that is known to be vulnerable


faust64:
allright.... let's forget about it for now.
Will check with new maintainer, why did he refuse my PR - that didn't break mocha tests, at least against node6.

PR #2926, opened 2017-07-14
adding a "retryable fail" event for a test that fails but can be retried

catdad:
Addresses issue #2529 by allowing reporters to take action on tests that have failed but can be retried.


ScottFreeCode:
Thanks for the submission!

As I mentioned on #2529, I don't have strong opinions for or against the feature; but I do want to put in here that I like that the implementation is this simple, and especially that you've provided a test (and at a glance it looks like that function might not have been tested before to begin with, too, so if there are objections to the feature we might still want a similar test...). :+1: Nicely done!

catdad:
@ScottFreeCode, thanks for the feedback. If there is anything I can do to make this go faster or smoother, please let me know.

LToronto:
👍  well done @catdad this fill my need very nicely.

mamanakisdm:
This is a GREAT idea!
We had this need so bad that one of our QA spent some time and developed a similar solution, and we thought about PUSHING the idea into this Repo. But then, he found this solution. 
Please, for the love of all that is holy, MERGE this Pull Request!

catdad:
I updated this PR with the latest code from master, since there were conflicts. Though, it's been 10 months now, so I am starting to lose hope.

test/unit/runner.spec.js
```diff
@@ -385,6 +385,40 @@ describe('Runner', function() {
...
+      var ERR = new Error('bear error');
+
+      var TEST = new Test('im a test about bears', function() {
+        if (retryableFails < retries) {
+          throw ERR;
+        }
+      });
+
+      TEST.retries(retries);
+      suite.retries(retries);
```
outsideris: `suite.retries(retries);` is enought for tests. 
IMO `TEST.retries(retries);` not unneccesary here.

catdad: Thanks for the feedback. I agree and have updated the PR.

outsideris:
Markdown linting is weired. Current master branch has been fixed.
It it rebased on master, CI will pass.

boneskull:
I'll see if I can merge this manually.

boneskull:
see #3408

PR #2931, opened 2017-07-18
chore(package): bump growl

faust64:
mocha currently includes a version of growl that is known to be vulnerable (https://snyk.io/test/npm/mocha/3.4.2)

vulnerability got fixed, eventually. although support for nodejs pre-4 broke.
node-growl maintainer was warned, he recommends adding an option (`--harmony`) running node, ...
forking growl to provide with aforementioned fix, without dropping older nodejs versions support.

Now, in  all fairness, having mocha tests pass with the original growl would be better.
Otherwise, I don't mind forking, growl has no dependency, not much code, relatively clear, ...

FYI:
Exhaustive diff between current node-growl and fork: https://github.com/tj/node-growl/compare/master...faust64:master
Exhaustive diff between vulnerability fix and our fork: https://github.com/tj/node-growl/compare/d71177d5331c9de4658aca62e0ac921f178b0669...faust64:master

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -239,7 +239,7 @@ Mocha.prototype.loadFiles = function (fn) {
  * @api private
  */
 Mocha.prototype._growl = function (runner, reporter) {
-  var notify = require('growl');
+  var notify = require('mygrowl');
 
   runner.on('end', function () {
     var stats = reporter.stats;
--- a/mocha.js
+++ b/mocha.js
@@ -1499,7 +1499,7 @@ Mocha.prototype.loadFiles = function (fn) {
  * @api private
  */
 Mocha.prototype._growl = function (runner, reporter) {
-  var notify = require('growl');
+  var notify = require('mygrowl');
 
   runner.on('end', function () {
     var stats = reporter.stats;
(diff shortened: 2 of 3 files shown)


ScottFreeCode:
Thanks for the PR!

For what it's worth, using a fork is one of our options I laid out here: https://github.com/mochajs/mocha/pull/2930#issuecomment-315798341 So now we've got PRs for a couple options out of the list... I might be overthinking the possible solutions there, but feel free to let me know if you have any thoughts (especially about the background info such as the browser stuff).

faust64:
AFAIR:
- using growl-1.10.0 didn't catch errors
- 1.10.1 was failing with some `Block-scoped declarations not yet supported outside strict mode` prior to node 5
- 1.10.2 was failing with `Use of const in strict mode` running with pre-node4 versions

I don't think I've run the browser support tests yet (most of these I ran on my laptop, without travis).
Although you make a good point in that comment you've linked: do we need notifications to begin with? I know I've never seen anything in XFCE notifier. And frankly, I'm not looking for these to start working ...

boneskull:
thanks.  we've worked this in to another set of PRs

PR #2934, opened 2017-07-23
Set hasOnly to false in _mocha rerun; closes #2429

### LINKED ISSUE #2429: --watch option works strangely with `.only()`

diegobg:
This is an alternative fix for the --watch with only bug. It's a single line of code and only changes the _mocha script, which doesn't appear to have any unit tests.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -459,6 +459,7 @@ if (program.watch) {
     if (!program.grep) {
       mocha.grep(null);
     }
+    mocha.options.hasOnly = false;
     mocha.suite = mocha.suite.clone();
     mocha.suite.ctx = new Mocha.Context();
     mocha.ui(program.ui);


ScottFreeCode:
Thanks for taking a shot at this!

In Mocha 4.0 we were able to replace the `hasOnly` property with a function to ensure the result is always up to date, which should make `--watch` and `only` work together correctly. Let us know on #2429 if there are still any issues!

PR #2937, opened 2017-07-28
Add coverage of Base reporter cursor

jsocol:
This gets coverage of `lib/reporters/base.js` to over 92% on branches, statements, functions and lines.

I had to add `isatty` to exports to be able to control it during tests.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -33,7 +33,7 @@ var clearInterval = global.clearInterval;
  * Check if both stdio streams are associated with a tty.
  */
 
-var isatty = tty.isatty(1) && tty.isatty(2);
+exports.isatty = tty.isatty(1) && tty.isatty(2);
 
 /**
  * Enable coloring by default, except in the browser interface.
@@ -118,7 +118,7 @@ exports.window = {
   width: 75
 };
 
-if (isatty) {
+if (exports.isatty) {
   exports.window.width = process.stdout.getWindowSize
       ? process.stdout.getWindowSize(1)[0]
       : tty.getWindowSize()[1];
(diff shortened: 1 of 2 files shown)


jsocol:
The test failures seem like flakes: an ECONNRESET during `npm install`; a timeout in a run where a _lot_ of tests were slower than normal; a bad tarball header.

I should also probably add a second set of tests, for `isatty = false` cases.

ScottFreeCode:
> I had to add  isatty  to exports to be able to control it during tests.

We shouldn't do this unless there's a non-test use case for allowing users to override it.

Otherwise, if we really want to test those branches by forcing `isatty` to one value or another, we should use something like Proxyquire to do it.

However, I'm inclined to think it would be more correct to cover those branches by testing Mocha in both TTY and non-TTY environments, since this is essentially not part of the interface (except for the `--colors`/`--no-colors` override) but just internal compatibility handling.

That may happen automatically between [adding coverage to all environments](https://github.com/mochajs/mocha/pull/2886) and later switching from the Makefile to NPM scripts (there's a test in the Makefile that seems like it would cover this but is suppressing whether the tests actually pass, but if we changed it into an NPM script with `&&` chains that would eliminate the result suppression and allow it to be added to the overarching `test` command).

ScottFreeCode:
(P.S. Thanks for the PR -- we will review the coverage improvements here in general, in any case.)

jsocol:
I was kind of surprised none of this code got executed in any of the reporter tests—even if the statement coverage wasn't 100% I thought the line coverage would be there. That makes me wonder if this just isn't used at all by the built-in reporters? Especially if it's "not part of the interface", maybe it's dead code?

Sorry to leave this hanging, I'm going on vacation for a couple weeks. If you want to go another direction feel free to close this out. It's probably not the right PR to add a TTY/non-TTY harness. Otherwise I'll happily try to help move it along when I get back!

jsocol:
Seems like this is the wrong direction so going to close it to avoid lingering PRs, but let me know if it's worth picking up again!

PR #2938, opened 2017-07-28
don't attempt to import growl in browser build

ahamid:
Does growl work in the browser?  I had to add this in order to perform es6 imports in a webpack build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -367,6 +367,7 @@
     "./index.js": "./browser-entry.js",
     "fs": false,
     "glob": false,
+    "growl": false,
     "path": false,
     "supports-color": false
   },


ScottFreeCode:
That's a good question -- if Growl doesn't work in browsers to begin with, [we'll want to know for sake of determining how to handle the vulnerability fix too](https://github.com/mochajs/mocha/pull/2930#issuecomment-315798341).

ScottFreeCode:
(P.S. Thanks for the PR! If we can confirm that Growl shouldn't even be used in the browser, this update looks good to me.)

BuonOmo:
I had a project were I needed notifications in browser. I ended up setting `growl` to true and overriding `mocha._growl`. I can push here this override of `_growl` (that uses [web notifications](//mdn.io/notifications)). What do you think about it?

Also by reading the code I could see that the best place to put this overriding is `browser-entry.js`, right?

ahamid:
@BuonOmo were you using webpack to package the js? My issue was that webpack was including the npm -sourced growl package automatically, it was not just a runtime configuration issue.

BuonOmo:
@ahamid Yes I packaged the whole js with webpack, using the mocha-loader. I've even created [a question on SO](https://stackoverflow.com/q/45509216/6320039) about it to see if someone had encountered the same issue.

ahamid:
@BuonOmo interesting, i wasn't aware of `mocha-loader`, I will give that a shot

BuonOmo:
It is not very well documented, I'd suggest you to read the code in order to see what it does. Anyway, for my previous question: may I implement notifications in browser for mocha ? (@ScottFreeCode)

ScottFreeCode:
> Also by reading the code I could see that the best place to put this overriding is  browser-entry.js , right?

> Anyway, for my previous question: may I implement notifications in browser for mocha ?

You're certainly welcome to give it a try! I can't promise it will be evaluated promptly (we've already got a bit of a backlog of stuff to go through at the moment), but we'd be happy to at least look at it at some point and see what it's like, then decide what we want to do with it.

As for browser-entry.js, I don't happen to know of anywhere better to handle it in Mocha's codebase off the top of my head, though someone else on the team might.

On the other hand, it may be worth asking whether Growl would want browser integration like that (in which case we could just ensure Growl is usable in the browser and we'd get the functionality), provided we resolve the previously mentioned issue of supporting newer versions of Growl somehow.

BuonOmo:
Ok thank I'll make a PR and reference it here. I don't think growl is supporting browser, but maybe we should name the option after `notify` instead of `growl`, this would go straight to the point and users can use it whether they are in browser or not without having to care about how the notification is done.

Anyway, let's discuss this in the according PR :)

boneskull:
I am pretty sure that #2962 preempts this.

PR #2948, opened 2017-08-08
testing

prianchiru:


ScottFreeCode:
Thanks, but there shouldn't generally be a need for updates that have been merged into `master` to be merged into other branches that [are themselves being prepared for merging into master](/mochajs/mocha/pull/2900). Once the other enhancements on which this branch depends are merged we plan on rebasing it to pick up those enhancements with a clean history, and that will also pick up any fixes in `master` that are needed for checks to pass.

PR #2954, opened 2017-08-11
Correct issue with timeout 0 closing prematurely

### LINKED ISSUE #2537: Inconsistent timeout policy

```
  this.timeout(0);

  it("is here now", function(){
    console.log("Allgood");
  });
});
```

Current result 
```
  should never end
R:\>
```
Expected result
```
  should never end
R:\>
```

ScottFreeCode:
...without causing any new issues allowing --no-exit to close

Fixes #2537

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -260,6 +260,7 @@ Runnable.prototype.run = function (fn) {
   var ctx = this.ctx;
   var finished;
   var emitted;
+  var keepalive = setInterval(function () {}, 30 * 1000);
 
   // Sometimes the ctx exists, but it is not runnable
   if (ctx && ctx.runnable) {
@@ -277,6 +278,7 @@ Runnable.prototype.run = function (fn) {
 
   // finished
   function done (err) {
+    clearInterval(keepalive);
     var ms = self.timeout();
     if (self.timedOut) {
       return;
--- a/test/integration/fixtures/timeout-0-done.fixture.js
+++ b/test/integration/fixtures/timeout-0-done.fixture.js
@@ -1,0 +1,6 @@
+'use strict';
+
+it('never finishes', function (done) {
+  process.send('process started');
+  this.timeout(0);
+});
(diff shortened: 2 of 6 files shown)


ScottFreeCode:
Can anyone help me understand the CI results I'm seeing?

- In older Node: `clearInterval` is saved the same way `setInterval` is saved; if `setInterval` isn't affected by global monkey-patching, then niether should `clearInterval` be... yet `clearInterval` is failing the test for robustness against global monkey-patching, while the earlier `setInterval` is not!
- In newer Node: it just hangs after the timeout unit tests??

If you run the three new tests locally you can see they all pass, so I'm having a hard time believing that I actually screwed anything up this badly...

ScottFreeCode:
What am I thinking? This solution is insane. Sure, intuitively if you never complete the test you'd expect Mocha to just wait for it indefinitely. But *there is no other work scheduled* -- in all but the oldest version of Node (which would simply continue to have the broken behavior, thus what I'm about to suggest is kinda-sorta backwards-compatible), if we *don't* schedule work specifically to keep the test alive, [we can detect when Node would be ready to close for lack of work scheduled and intercept that](https://nodejs.org/api/process.html#process_event_beforeexit). So instead of deliberately keeping the process alive when the test might never complete, we should just have `Runnable` listen for the event that Node has nothing else scheduled and turn that into a test failure (which resumes processing things and thus gets more scheduled, preventing the close), with text something like:
```js
new Error("Test has no more asynchronous actions scheduled and will never complete; please call `done` or resolve the promise to complete the test.")
```

I'll code up that solution later and redesign the integration tests to match the expectation that if it works correctly they'll error rather than running indefinitely (and treat actually running indefinitely as an error -- hmm, I might be able to make these regular integration tests with a normal timeout). (The `--no-exit` one is still valid as-is since we need to be sure that *any* solution to this doesn't accidentally stop the process from closing normally when tests are done. Although I suppose it ought to use a short-but-still-async test, or two one with `done` and one with a promise, to be absolutely certain it's triggerring the new solution. Hey, it could use them in addition to the sync test even, then we know we're trying everything...)

ScottFreeCode:
Updated `--no-exit` test and pushed to `timeout-0-indefinite` branch so we can compare this solution with the `'beforeExit'` one, which I will push to the `timeout-0` branch once it's ready.

ScottFreeCode:
New PR is #2959

Issue #2955, opened 2017-08-12
ANSI escape codes in messages break xunit file


jkrems:
When the error message or stack contains any non-safe characters, it breaks the xunit file and the resulting report is no longer valid XML.

Repro:

```bash
# Explicit:
npm init --yes
npm i -D mocha
echo "it('throws', () => { throw new Error('\x1B[32mfoo\x1B[0m'); })">test.js
./node_modules/.bin/mocha -R xunit -O output=report.xml test.js

# One-liner:
npm init --yes >/dev/null && npm i -D mocha >/dev/null && echo "it('throws', () => { throw new Error('\x1B[32mfoo\x1B[0m'); })">test.js && ./node_modules/.bin/mocha -R xunit -O output=report.xml test.js
```

Running `report.xml` through an [online validator](http://www.xmlvalidation.com/) will trigger messages like the following:

> An invalid XML character (Unicode: 0x1b) was found in the element content of the document.

PR #2957, opened 2017-08-13
Add `utils.escape` tests and fix unicode escaping

jkrems:
I started walking down the path of replacing those myself but once I hit surrogate pairs it seemed to become silly. That's why I went with just adding `he`.

Fixes https://github.com/mochajs/mocha/issues/2955

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,6 +18,7 @@ var statSync = require('fs').statSync;
 var watchFile = require('fs').watchFile;
 var lstatSync = require('fs').lstatSync;
 var toISOString = require('./to-iso-string');
+var he = require('he');
 
 /**
  * Ignored directories.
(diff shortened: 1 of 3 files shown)


boneskull:
@jkrems Thanks!!

JamesMGreene:
The `{ useNamedReferences: true }` part of this PR breaks the XUnit reporter, which now outputs things like `&grave;` (which is only valid in HTML, not XML).

jkrems:
Oops, could've anticipated that. @JamesMGreene can you confirm that setting that option to false fixes everything? I'll assume that we could live with slightly uglier `&quot;` escaping as long as the files are all valid.

jkrems:
Assuming that we're fine with "ugly" escapes for now, there's a potential fix in https://github.com/mochajs/mocha/pull/3003 now. @JamesMGreene Can you give that branch a try to verify that it fixes your issue?

PR #2959, opened 2017-08-13
Fail test with timeout 0 that completes without calling `done` or resolving a returned promise

### LINKED ISSUE #2537: Inconsistent timeout policy

ScottFreeCode:
Fixes #2537 

Also improves on the original attempt (in the `timeout-0-indefinite` branch, where the process is kept alive doing nothing forever in the same situation this fixes to an error) in handling `allowUncaught`, which could otherwise lead to situations wherein the handler was not removed at the end of a test.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -249,6 +249,17 @@ Runnable.prototype.globals = function (globals) {
 };
 
 /**
+ * Allow uncaught-exception failures to trigger cleanup that would otherwise be altogether missed
+ *
+ * @api private
+ */
+Runnable.prototype.cleanUpHandlers = function () {
+  if (this._closeIntercept) {
+    process.removeListener('beforeExit', this._closeIntercept);
+  }
+};
+
+/**
  * Run the test and invoke `fn(err)`.
  *
  * @param {Function} fn
@@ -260,6 +271,15 @@ Runnable.prototype.run = function (fn) {
   var ctx = this.ctx;
   var finished;
   var emitted;
+  process.on('beforeExit', detectAsyncUnscheduled);
+  if (this._closeIntercept) {
+    process.removeListener('beforeExit', this._closeIntercept);
+  }
+  this._closeIntercept = detectAsyncUnscheduled;
+
+  function detectAsyncUnscheduled () {
+    done(new Error('After ' + (new Date() - start) + ' milliseconds there were no more asynchronous actions scheduled. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.'));
+  }
 
   // Sometimes the ctx exists, but it is not runnable
   if (ctx && ctx.runnable) {
@@ -277,6 +297,7 @@ Runnable.prototype.run = function (fn) {
 
   // finished
   function done (err) {
+    process.removeListener('beforeExit', detectAsyncUnscheduled);
     var ms = self.timeout();
     if (self.timedOut) {
       return;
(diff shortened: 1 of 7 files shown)


ScottFreeCode:
I've manually checked what happens in the same situation on Karma and in a manually opened browser with and without each fix. It seems that:
- manually opened browser will just hang no matter what we do (there's no event I know to listen for lack of scheduled async work, but also nothing closes if we don't keep things moving either), and it doesn't hurt anything because that's not sending a programmatic result code back anyway
- Karma with its usual timeouts just gets a timeout error (better than nothing and closer to the new error than to infinite hanging) whether fixed or not and no matter which fix
- Karma with `browserNoActivityTimeout` set to `0` just hangs indefinitely much like the manual browser. Probably nothing we can do about that for similar reasons?

I'd love to get Karma tested for behavior in this case, but it's another thing I'll need to find some way to do browser integration tests if we're going to do it.

boneskull:
I want to see the interaction with the v4 behavior

lib/runnable.js
```diff
@@ -260,6 +271,15 @@ Runnable.prototype.run = function (fn) {
...
   var finished;
   var emitted;
+  process.on('beforeExit', detectAsyncUnscheduled);
+  if (this._closeIntercept) {
+    process.removeListener('beforeExit', this._closeIntercept);
+  }
+  this._closeIntercept = detectAsyncUnscheduled;
+
+  function detectAsyncUnscheduled () {
+    done(new Error('After ' + (new Date() - start) + ' milliseconds there were no more asynchronous actions scheduled. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.'));
```
boneskull: are there any other cases than `--timeout 0` or `this.timeout(0)` where this error would happen?  if not, we can be more specific.

test/integration/fixtures/timeout-0-noexit.fixture.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+describe('all test possibilities', function () {
+  it('finishes', function () {
+    process.send('process started');
```
boneskull: I don't quite understand why we need the IPC.

test/integration/fixtures/timeout-0-noexit.fixture.js
```diff
@@ -0,0 +1,39 @@
...
+  it('errors asynchronously', function (done) {
+    setTimeout(function () { throw new Error('ignore me'); }, 10);
+  });
+
+  it('fails (with done) asynchronously', function (done) {
+    setTimeout(done.bind(null, new Error('ignore me')), 10);
+  });
+
+  it('returns a promise that eventually rejects', function () {
+    return new Promise(function (resolve, reject) { setTimeout(reject.bind(null, new Error('ignore me')), 10); });
```
boneskull: `Promise`s are never not asynchronous, so you shouldn't need to call `setTimeout()` here.  Also, I believe this code is functionally equivalent to the test on L24, and can be removed.

We can also probably just use `process.nextTick()` instead of `setTimeout()`.

test/integration/options.spec.js
```diff
@@ -280,4 +282,110 @@ describe('options', function () {
       });
     });
   });
+
+  describe('--no-exit', function () {
```
boneskull: I'm not sure we need this, as it seems to pass with or without the changes in `runnable.js`.  Either way I'm trying to combine it with what we have for `--exit` tests, as I wrote for v4.

test/integration/timeout.spec.js
```diff
@@ -18,4 +18,73 @@ describe('this.timeout()', function () {
       done();
     });
   });
+
+  describe('0 (no timeout)', function () {
+    var itIfNodeWorthy = (function () {
```
boneskull: happily, we can ditch this.

(I *am* merging and fiddling with this code, so I don't expect you to do it)

PR #2961, opened 2017-08-15
Allow comments in mocha.opts

danday74:
Any line starting with // in mocha.opts will be ignored, viz treated as a comment


danday74:
gonna add some unit tests

PR #2963, opened 2017-08-16
Introduced comments to mocha.opts

danday74:
A line starting with // in mocha.opts will be treated as a comment.

If there is anything wrong with this PR then please advise. 

Many thanks.

Dan Lewis

### DIFF
--- a/bin/options.js
+++ b/bin/options.js
@@ -5,6 +5,7 @@
  */
 
 var fs = require('fs');
+var getOptionsPath = require('./optionsPath');
 
 /**
  * Export `getOptions`.
@@ -18,11 +19,12 @@ module.exports = getOptions;
 
 function getOptions () {
   var optsPath = process.argv.indexOf('--opts') === -1
-    ? 'test/mocha.opts'
+    ? getOptionsPath.getPath()
     : process.argv[process.argv.indexOf('--opts') + 1];
 
   try {
     var opts = fs.readFileSync(optsPath, 'utf8')
+      .replace(/^\/\/.*$/mg, '')
       .replace(/\\\s/g, '%20')
       .split(/\s/)
       .filter(Boolean)
--- a/bin/optionsPath.js
+++ b/bin/optionsPath.js
@@ -1,0 +1,10 @@
+'use strict';
+
+// Putting this method in its own module facilitates ease of stubbing for unit testing purposes
+function getPath () {
+  return 'test/mocha.opts';
+}
+
+module.exports = {
+  getPath: getPath
+};
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "Cory Thomas (https://github.com/dump247)",
     "cybertk (https://github.com/cybertk)",
     "Daniel Ericsson (https://github.com/monowerker)",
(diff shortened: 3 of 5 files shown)


ScottFreeCode:
Hi @danday74, thanks for the PR! I especially appreciate the tests added here; we may adapt those even if we go with #2870 instead.

My gut instinct is that the comment indicator should be `#` (as in #2870) as in shell scripting because `mocha.opts` is just a list of commandline arguments. However, we'll need to look into potential compatibility issues with filenames (`mocha.opts` can specify the test file pattern/path), especially if may want to add we add end-of-line comment support (as discussed below). To my knowledge:
- double-slashes are valid in filepaths and can occur:
  - at the start of the path on some OSes where this indicates the path is on another machine (admittedly an unusual use case that would only show up in, say, a private and peculiar enterprise environment rather than something public and portable)
  - when a tool, program or user uses naive path concatenation to generate the path (e.g. joining `directory/` and `/filename.js` to `directory//filename.js`)
- `#` is valid in filenames, at the start or otherwise, but:
  - it is by convention reserved for special purposes such as Emacs' backup files

I could be wrong about some of those points and they may not be exhaustive, however; we may want to look for an expert in file name/path problems (or ask on Stack Exchange) for further info.

I'd also be curious whether we would want to support comments on the same line as options, e.g. `--compilers myspecialthing:path/to/mything # myspecialthing is really just a fork of CoffeeScript that fixes my pet peeves and adds my favorite feature from Elm or whatever`; in that case, ensuring that the regex does not begin with `^` would probably do -- although I also presume we could add comment support at the start of lines first and then later add comment support on the same line after options.

danday74:
Many thanks for the feedback on this.

Happy if you decide in favour of [#2870](https://github.com/mochajs/mocha/pull/2870).

Also happy if you want me to make changes so that we can get this in. I'm very active so can make changes quickly and would be proud to be a contributor. Very experienced with unit testing too.

Re: // syntax

Interesting discussion on "naive path concatenation". Good thinking. We could ensure that the line starts with // OR that // is preceded by a space before considering it a comment?

Re: # syntax

Since filenames can also contain these, how about ensuring that the line starts with # OR that # is preceded by a space?

See also .. https://stackoverflow.com/questions/45742287/javascript-regex-to-distinguish-between-a-comment-and-a-file-path

Happy to change comment syntax from // to # (or support both?)

Happy also to add support for same line comments.

If you want me to take any action then please let me know, I am very keen, I love mocha :)

danday74:
Can we take some action here?

ScottFreeCode:
Hey @danday74, just a heads-up, there's some discussion going on at #2870 currently that may get us to where we can move forward with these decisions and merge these. Feel free to drop in any thoughts you have there!

danday74:
Thanks for the heads up, have chipped in on #2870 as suggested :)

igwejk:
@danday74 , I commend your effort very much.

Though, as you know already #2870 introduced support for comments in `mocha.opts`.

We should also take note that `bin/options.js` was refactored and tested fully in #2881.

Let's throw our weight behind the PRs that have prospectively improved mocha options instead of doing them all over in another PR.

danday74:
agreed i will close this off

PR #2978, opened 2017-09-02
Update lodash.create

bdsomer:
As you can see through [this visualization](http://npm.anvaka.com/#/view/2d/mocha), updating [`lodash.create`](https://npmjs.com/package/lodash.create) would drastically reduce the number of dependencies that [`mocha`](https://npmjs.com/package/mocha) has, causing **faster install times** and occupying **less disk space**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -315,7 +315,7 @@
     "glob": "7.1.1",
     "growl": "1.9.2",
     "json3": "3.3.2",
-    "lodash.create": "3.1.1",
+    "lodash.create": "4.2.0",
     "mkdirp": "0.5.1",
     "supports-color": "3.1.2"
   },


ScottFreeCode:
Thanks for the PR! This sounds really helpful; I was looking for ways to cut down on install time not too long ago.

To be sure the new version works in all the environments Mocha supports, I think we'll want to rebase this after we [get the JWT stuff for SauceLabs set up](/mochajs/mocha/issues/2602) so that the Travis jobs are actually testing all the browsers (whereas right now there's a separate job for each browser but for PRs from forks they redundantly test PhantomJS instead).

bdsomer:
@ScottFreeCode Okay, sounds good!

boneskull:
See #2979; `lodash.create` *may* be able to be dropped when that happens.

boneskull:
linking this to #3018.

boneskull:
ok, we don't need this.  thanks anyhow!

bdsomer:
@boneskull Sounds good!

PR #2981, opened 2017-09-04
Spelling

jsoref:
### Description of the Change

Spelling fixes

### Alternate Designs

This PR can be squashed, and individual changes can be dropped.

These are fixes to existing content.

### Applicable issues

* Is this a breaking change (major release)?
I don't think so, unless as noted in drawbacks above
* Is it an enhancement (minor release)?
I believe so.

### DIFF
--- a/test/browser/ui.spec.js
+++ b/test/browser/ui.spec.js
@@ -19,7 +19,7 @@ describe('$.jQuery', function () {
 });
 
 // another generic describe block to verify it is absent
-// when greeping on $.jQuery
+// when grepping on $.jQuery
 describe('@Array', function () {
   it('.pop()', function () {
     assert(true);
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,7 +17,7 @@ We must be able to understand the design of your change from this description. K
 
 ### Why should this be in core?
 
-<!-- Explain why this functionality should be in mocha as opposed to it's own package -->
+<!-- Explain why this functionality should be in mocha as opposed to its own package -->
 
 ### Benefits
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1091,7 +1091,7 @@ Thanks @entertainyou, @SimenB, @just-paja for the heads-up.
 ==================
 
   * add: sort test files with --sort (#813)
-  * update: diff depedency to 1.0.7
+  * update: diff dependency to 1.0.7
   * update: glob dependency to 3.2.3 (#927)
   * fix: diffs show whitespace differences (#976)
   * fix: improve global leaks (#783)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -863,7 +863,7 @@ Thanks @entertainyou, @SimenB, @just-paja for the heads-up.
   * utils.stringify should handle objects without an Object prototype
   * in runnable test, comparing to undefined error's message rather than a literal
   * Fix test running output truncation on async STDIO
-  * ammended for deprecated customFds option in child_process
+  * amended for deprecated customFds option in child_process
 
 2.1.0 / 2014-12-23
 ==================
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -156,7 +156,7 @@ exports.cursor = {
 };
 
 /**
- * Outut the given `failures` as a list.
+ * Output the given `failures` as a list.
  *
  * @param {Array} failures
  * @api public
--- a/test/unit/ms.spec.js
+++ b/test/unit/ms.spec.js
@@ -29,7 +29,7 @@ describe('.ms()', function () {
     });
   });
 
-  describe('minutess representation', function () {
+  describe('minutes representation', function () {
     it('should return short format', function () {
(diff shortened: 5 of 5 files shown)


mocha.js
```diff
@@ -13074,7 +13074,7 @@ var process = module.exports = {};
 var cachedSetTimeout;
 var cachedClearTimeout;
 
-function defaultSetTimout() {
+function defaultSetTimeout() {
```
ScottFreeCode: To answer a concern from the description: given that this function seems to be local, I don't think it's an API, let alone a public one, so as long as all the references to it were updated this should be fine. (Now, if the name of something it is assigned to were updated, we'd need to check whether *that* is an API, but I'm pretty sure none of the left-sides for this were changed...)

test/unit/throw.spec.js
```diff
@@ -91,7 +91,7 @@ describe('a test that throws', function () {
 
     it('should not pass if throwing async and test is async', function (done) {
       var test = new Test('im async and throw null async', function (done2) {
-        process.nexTick(function () {
+        process.nextTick(function () {
           throw null;
         });
       });
```
ScottFreeCode: Judging from the looks of these corrected tests, I *think* they were *supposed* to `throw undefined` and `throw null` in the `process.nextTick` callback, *but* were actually throwing due to `process.nexTick` being undefined, and the tests only checked that the attempt failed somewhere along the line and not that it got to the callback and failed by throwing out of the callback. I don't know off the top of my head whether those tests should be checking something to be sure that `undefined` and `null` are what's thrown respectively, but they should probably be doing something to prevent treating throws outside the callback as passing. Perhaps wrapping the synchronous portion  with `try {` and `} catch (ignore) { /* ensure that early failure due to any problems reaching the callback is not mistaken for the failure we're expecting to get out of the callback when it is asynchronously called */ }`?

In any case, some extra eyes on them would be helpful just to make sure that we really are making them better... I'm not absolutely certain I've ascertained their purpose correctly here, and, frankly, I can't quite figure out why there are tests failing in CI now...

mocha.js
```diff
@@ -7430,7 +7430,7 @@ function slowToString (encoding, start, end) {
     return ''
   }
 
-  // Force coersion to uint32. This will also coerce falsey/NaN values to 0.
+  // Force coersion to uint32. This will also coerce falsy/NaN values to 0.
```
ScottFreeCode: I'm not so sure that "falsey" is incorrect -- niether spelling should show up in a normal spellchecker/dictionary as it's programming jargon, but I've definitely seen the spelling with an E used.

test/browser/ui.spec.js
```diff
@@ -19,7 +19,7 @@ describe('$.jQuery', function () {
 });
 
 // another generic describe block to verify it is absent
-// when greeping on $.jQuery
+// when greping on $.jQuery
```
ScottFreeCode: I'm also fairly sure that "grepping" is even better than "greping" per typical English language consonant-in-middle-of-word rules, unless using some kind of quote or codeblock-backtick to separate the `grep` from the "ing", but, eh, anything's better than "greeping" (:laughing:).

mocha.js
jsoref: Both are used in this file. This spelling is only used once in the entire repository.

test/browser/ui.spec.js
jsoref: Err, yeah. I have no idea why I picked that spelling. Will fix.

mocha.js
jsoref: Hrm, by ignoring the mocha.js, the thing goes away, which means that it's coming from some included module. If you happen to know which module, I can fix it there. Otherwise, this will disappear as I rebase back/forth...

mocha.js
jsoref: Ok, turns out this is coming from: https://github.com/defunctzombie/node-process/blob/master/browser.js (or some variant of that). If you can help find the right project for it, that'd help me.

test/unit/throw.spec.js
jsoref: I'm going to split this out and leave it for someone else to solve.
I'm on a spelling binge. While I do understand JS and node, I don't really want to spend the time figuring out the right magic to make the test happy.

mocha.js
ScottFreeCode: > Both are used in this file. This spelling is only used once in the entire repository.

Huh, interesting catch -- consistency is definitely a plus then. Well, maybe less so for the package it comes from if it's the only use of the term there, but...

Speaking of which, I may or may not be able to track down which project it's from... It's generally a matter of scrolling up (or is it down? I forget) in the bundled file till you find the name in the bundle glue (the beginning of which will be the package, but the full module path should be in there), then `npm view <package name>` and looking for GitHub URLs. But if there aren't any GitHub URLs for the package, it gets harder...

mocha.js
ScottFreeCode: I'd imagine that that is in fact the right project. As I understand it, this defunctzombie guy maintains a bunch of packages that are basically clones of the Node builtins that can be pulled in by Browserify and Webpack.

ScottFreeCode:
New set of changes looks good! And test failure was unrelated timeout flake; reran and it passed.

PR #2997, opened 2017-09-09
import/require xit, fixes #2972

### LINKED ISSUE #2972: import / require xit
Is there a way to `import` / `require` to satisfy my tooling?

I can do

```javascript
import { it } from 'mocha';
```

but not

```javascript
import { xit } from 'mocha';
```

or

```javascript
import { xit } from 'mocha/lib/interfaces/bdd';
```

solodynamo:
### Description of the Change

Adds `xit` into the exports in the files `lib/mocha.js` and `mocha.js`.

To individually require/import` xit` from the library.

### Benefits

It will allow devs to skip a particular test. Another alternative is `it.skip`


lib/mocha.js
```diff
@@ -204,6 +204,7 @@ Mocha.prototype.ui = function (name) {
     exports.before = context.before || context.suiteSetup;
     exports.describe = context.describe || context.suite;
     exports.it = context.it || context.test;
+    exports.xit = context.xit || context.test;
```
ScottFreeCode: (**This is the only change that should be needed to merge this.** Any others suggested are optional.)

So, the `context.test` on the right side of the `||` isn't right, that's the same as for the non-skipped `it`. (With that said, you can skip to the very bottom for recommended fix; the rest here is explanation.) I'm not 100% certain whether this involves the QUnit interface at all, but it seems these are basically enabling `require`-based used of either the BDD or the TDD interface when either the BDD or TDD interface is used; so, the `||` here would be to replicate the BDD interface in the event the TDD interface is chosen. So, I double-checked what the TDD interface does for skipping.

In the BDD interface, `xit`, `xspecify` and `it.skip` are all set to a function that just calls `it` while only using the title. In the TDD interface, there are no `x` functions, and `test.skip` delegates to the "common interface" `test.skip`, which in turn relies on the TDD interface's `test` function and -- like the BDD function -- just passes it only the title. (Weird coupling between the "common interface" and the TDD interface... but nothing worth worrying about here.)

*So it looks like it should be perfectly safe to change this to `|| context.test.skip`.*

mocha.js
```diff
@@ -1464,6 +1464,7 @@ Mocha.prototype.ui = function (name) {
     exports.before = context.before || context.suiteSetup;
     exports.describe = context.describe || context.suite;
     exports.it = context.it || context.test;
+    exports.xit = context.xit || context.test;
```
ScottFreeCode: Small tip: you don't actually have to update the `mocha.js` file at the base of the project, as it is generated from the other files as part of the publication process. (On the other hand, for the same reason, you don't have to not update it either -- it will simply be overwritten later, as long as you update the other files in addition. I just mention this because it's useful to know you don't have search through the gigantic bundle for the code to update and do have to update other files for changes to actually be picked up.)

lib/mocha.js
solodynamo: @ScottFreeCode  Updated the PR with suggested changes. This time not updated mocha.js at root level.

mocha.js
solodynamo: @ScottFreeCode Updated the PR with suggested changes. This time not updated mocha.js at root level.

ScottFreeCode:
Squashed commits and merged in 1cc0fc0e6153bbd746b0c2da565363570432cdf7 (sorry that didn't automatically update this PR); thanks again!

PR #3003, opened 2017-09-11
Use safer XML escaping

jkrems:
### Description of the Change

Fixes a regression introduced in https://github.com/mochajs/mocha/pull/2957. It was too aggressive in its use of named escapes which broke XML files that might not support all of the named escapes available in HTML.

### Alternate Designs

We could try to get support for safe named escapes into `he`. But given that this won't affect the visible output, I focussed on fixing the functionality first.

### Possible Drawbacks

This makes the HTML output a bit uglier. E.g. it won't use the nice `&quot;` anymore.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -36,7 +36,7 @@ exports.inherits = require('util').inherits;
  * @return {string}
  */
 exports.escape = function (html) {
-  return he.encode(String(html), { useNamedReferences: true });
+  return he.encode(String(html), { useNamedReferences: false });
 };
 
 /**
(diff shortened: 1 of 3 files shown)


boneskull:
thanks

JamesMGreene:
Thanks for the quick turnaround here, @jkrems and @boneskull.  :+1:

This was making hundreds of our CI builds fail today.  😵

jkrems:
@JamesMGreene Sorry for that & thanks for the report! :)

JamesMGreene:
@boneskull Could you kindly drop a comment here when `v3.5.3` is published to NPM?  Thanks!

boneskull:
published as v3.5.3

arwilczek90:
Thank you so much! That threw me for a loop all morning!

Issue #3016, opened 2017-09-27
drop Node.js v0.10 and v0.12 support


boneskull:
We can no longer run our development environment on node.js v0.10/v0.12 (see request/request#2772) without adding a `npm-shrinkwrap.json` or doing other high-effort/low-reward things to work around this.

furthermore, it's making it more difficult to apply security fixes.  at least one of these Node.js  versions (v0.10) has known vulnerabilities which have gone unpatched (in an official capacity).

we should not continue to supply updates for these environments.

- [x] remove `0.10`, `0.12`, `iojs` & `5` from `.travis-ci.yml` and `appveyor.yml`
- [x] update `engines` field in `package.json`
- [x] necessary security upgrades (this includes `debug` and `growl`)
- [x] upgrade anything else we can get away with (stick to dev deps)

To be clear, Mocha *still runs* in Node.js v0.10 and v0.12; you just can't clone a working copy and run the test suite.

boneskull:
also dump `iojs` because

boneskull:
also Node.js v5 [is no longer maintained](https://github.com/nodejs/Release#release-schedule1), so drop that too.

ScottFreeCode:
This week is the week we will merge as many [semver-major](https://github.com/mochajs/mocha/labels/semver-major) updates as we can?

boneskull:
sure.  I'll try to release EOD Fri to give people the weekend to find whatever we broke.

boneskull:
WIP PR: #3017

boneskull:
@ScottFreeCode please throw stuff into kanban if you want to see it land

ScottFreeCode:
Will do; gotta turn in for the night but can focus on it tomorrow so we have the middle of the week to evaluate stuff.

boneskull:
closed via e39a867

ScottFreeCode:
@boneskull Sorry this is later than I meant to get it, but I put a bunch of issues in a new kanban column "Evaluate for V4" -- the idea here being that we don't have to sort through the Todo or In Progress columns looking for which are semver-major *and* **these don't necessarily need to go in, I'd just like them *considered*.** (Again, apologies that there isn't much time left to consider them -- although some are either trivial or already have PRs, so there's that -- but that's also why I'm not saying any of them necessarily need to go in. Some of them you may have already looked at since they were labelled semver-major, as well.)

ScottFreeCode:
[...]

PR #3017, opened 2017-09-27
drop support for unmaintained versions of Node.js; upgrade stuff

boneskull:
Addresses #3016

- Drops support for Node.js v0.10, v0.12, v5.x and iojs
- Address potential [arbitrary code injection vulnerability](https://snyk.io/vuln/npm:growl:20160721) in `growl`
- Address potential [regex DoS vulnerability](https://snyk.io/vuln/npm:debug:20170905) in `debug`
- Upgrade `eslint` and plugins and sweep up code
- Upgrade `should`, `growl`, and `debug`
- Update `engines` field in `package.json`

I'm going to see what else I can squeeze out of this.


boneskull:
This will be (squashed) and merged when the build's green.

lib/runner.js
```diff
@@ -229,7 +229,7 @@ Runner.prototype.fail = function (test, err) {
   ++this.failures;
   test.state = 'failed';
 
-  if (!(err instanceof Error || err && typeof err.message === 'string')) {
+  if (!((err instanceof Error || err) && typeof err.message === 'string')) {
```
ScottFreeCode: Are we sure this shouldn't be `if (!(err instanceof Error || (err && typeof err.message === 'string'))) {`? Reading the original, and not remembering exactly how JS evaluates differing chained boolean operators without parentheses, it *looks* like the original included `err` as a condition to guard against accessing `.message` of `undefined`.

lib/utils.js
```diff
@@ -540,8 +540,8 @@ function jsonStringify (object, spaces, depth) {
     --length;
     str += '\n ' + repeat(' ', space) +
       (isArray(object) ? '' : '"' + i + '": ') + // key
-      _stringify(object[i]) +                    // value
-      (length ? ',' : '');                       // comma
+      _stringify(object[i]) + // value
+      (length ? ',' : ''); // comma
```
ScottFreeCode: [unimportant] I long for the day that linters can recognise when whitespace has been properly used to make things, any things, line up across multiple lines. [/unimportant]

lib/runner.js
boneskull: I think you're right.

PR #3019, opened 2017-09-27
consolidate saucelabs tests to address flake

### LINKED ISSUE #2890: Mocha's hanging CI browser tests
Of late, there has been a failure in CI that looks like a browser flake... except that it's far too consistent to be a flake. Doesn't always happen, but usually does, and always the same browsers (and usually the same point in the test). It seems like the browser hangs or something -- Sauce disconnects with a timeout of 20,000ms partway through the testsuite.

It's blocking #2886, notably, but some testing determined that it affected the commits there even before coverage was actually added, so it seems to be something that coverage makes more likely to happen but isn't the sole cause.

I have had a... cryptic breakthrough, inasmuch as while I was experimenting with something else I accidentally wound up reproducing a hang in PhantomJS with Node 8.0.0 at the same points where Sauce is disconnecting. I do not know what about my local experiment triggerred this, and am going to move further experimentation to a new checkout of Mocha to ensure this one stays where it can reproduce the hangy issue.

Here's what I have been able to discover:
- In Travis, it occurs for Edge, IE7 and Safari.
- It seems like it may happen more consistently with my PR of adding coverage to Karma runs, but it has happened to builds that lack those changes as well (e.g. master, other PRs).
- Both in travis and locally, it happens after printing the success result for `should format functions saved in windows style - spaces` and for the last unit test.
- Adding console.logs to later tests indicates that, when it hangs after the "- spaces" test, not only is the next test started, but **many more tests are actually run even though no more get their results printed**; however, at some point there is a test that is never console.logged, thus indicating that it also stops running tests somewhere along the line.
- The point at which it actually stops running tests does not seem to be consistent.
- Adding console.logs may affect the point where it stops running tests and/or whether the issue occurs at all. It's hard to tell for sure. Sometimes it changes after I add a console.log and then changes back when I remove that console.log; other times it changes after I add a console.log and does not change back when I remove it.
- On the theory that the continued running of tests past the point where the reporter ceases output suggest that output is somehow gummed up and perhaps the system hangs for real when buffers waiting for the output to ungum finally fill up, I tried upgrading the two things that I presume affect output: karma and karma-spec-reporter. Results were much like adding console.logs: *it seems like these upgrades may affect whether and when the issue occurs, but niether one (nor even both together) consistently eliminates the issue.*
- By saving the Browserify bundle of the test files, I see that the tests when this issue occurs (where I've been adding console.logs) is not minified or obfuscated, so it is unlikely to be due to a bug in bundling that the exact code affects nondeterministically. At least, I think so anyway.

I feel just as stuck as before, but maybe this information will ring a bell to somebody.

boneskull:
may or may not fix #2890.


boneskull:
I don't know how to actually test for the existence of flake except to merge this and see if it works consistently.

boneskull:
tentatively stuffing into v4, but reserve the right to pull out if it becomes a morass.

I'll keep rebasing this.

ScottFreeCode:
> I don't know how to actually test for the existence of flake except to merge this and see if it works consistently.

That's pretty much where I'm at too. No idea what causes it, just that it always fits a certain pattern (specific browsers and hanging after one of two particular tests) when it does happen. Some of my investigation suggested a similar-looking issue *could* be caused by something going wrong in Mocha leading to a hanging test run after a suppressed internal error having to do with stringification and circular references, but that doesn't necessarily mean that that issue I discovered *is* the cause of the CI flake -- they could be unrelated problems that would just happen to look similar in their results.

Keeping the Sauce tunnel open seems like a good idea anyway, so it's definitely worth a shot.

Probably the only caveat I have about fiddling with how we run browser tests is that right now it's easy to see in the Travis jobs list which browser failed, but if we merged it into one Travis job we'd have to dig through pages of Karma output to find any errors. (This goes not just for flakes like this but for any test failure in a browser run.) If for some reason we need to do that, it isn't unmanageable; it's just the only potential con I can think of to any of the options we have.

On the other hand, besides potentially fixing the flake, this looks like it makes the CI run much, much faster. Tradeoffs...

test/node-unit/http.spec.js
```diff
@@ -1,16 +1,24 @@
...
+  before(function () {
+    return getPort().then(function (p) {
+      port = p;
+      server.listen(port);
+    });
+  });
+
   it('should provide an example', function (done) {
-    http.get({ path: '/', port: 8888 }, function (res) {
+    http.get({ path: '/', port: port }, function (res) {
```
ScottFreeCode: Just my two cents' worth, but I don't think we should be adding dependencies to tweak this test. It doesn't seem to be a real unit test of Mocha to begin with -- probably more like an example use case. If it's causing trouble we should stop running it and (if we want to keep it as an example) move it into documentation somewhere instead.

boneskull:
@ScottFreeCode yeah, it does give better visibility into *which* browser failed.  

that being said, I'm fully behind trying to get Jenkins running, since it'll give us more control (and information).  I have a droplet; just need to configure the thing.

boneskull:
so far the only failures I've seen have *not* been of the "couldn't start a browser" variety, and I've ran quite a few builds today.

boneskull:
@ScottFreeCode anyhow, this LGTM.  I'll merge it tomorrow w/ your blessing

PR #3024, opened 2017-09-28
upgrade deps; closes #3018 (WIP--DO NOT MERGE)

### LINKED ISSUE #3018: upgrade (mostly) production deps
after #3016 but before v4.0.0, these production deps should be upgraded to the latest version, if not too painful:

- `commander`
- `debug`
- `diff`
- `glob`
- `lodash.create`
- `readable-stream`
- `supports-color`

the following dev deps should be upgraded as well, with caveats:

- `karma-phantomjs-launcher`: this should be part of #2979 
- `karma`: depends on previous item
- `browserify`: unknown effect on browser bundle

boneskull:
### upgrades: production

- [x] `commander`
- [x] `debug`
- [x] `diff`
- [x] `glob`
- [x] `supports-color`
- [x] `browserify` (affects production)

### upgrades: development

- [x] `karma-phantomjs-launcher`
- [x] `karma`
- [x] `karma-sauce-launcher`

### modules added

- [x] `buffer`

### modules removed

- [x] `lodash.create` 
- [x] `phantomjs` (installed via `karma-phantomjs-launcher` instead)
- [x] `readable-stream`
- [x] `json3` (this was for IE7/8's crap implementation)


boneskull:
The story on `buffer` is that `browserify@14.4.0` ships with `buffer@5` which is incompatible with IE10 and older.  So we have to explicitly depend on `buffer@4` and tell `browserify` to use it.

boneskull:
okay, I'm going to merge this.  maybe a bit of rebasing but I'll try to keep things separated.

PR #3033, opened 2017-09-29
Remove generated file from VCS

### LINKED ISSUE #2615: do not commit generated files to VCS
When we added Bower support long ago (#584), it was implemented by committing the bundle (the root `mocha.js`) to VCS.

This, of course, is bad practice.  

I have no numbers anywhere on how many people install Bower via Mocha, but it's probably enough to cause pain if we stop supporting it.

Instead, we can create a script that dumps the dist file(s) into [another repository](/mochajs/bower-mocha), *a la* AngularJS.  The dist file(s) would still be *published* to npm.

Practically speaking, whenever anybody runs the browser-based tests via Karma, we generate a new bundle so that `karma-mocha` uses Mocha to test itself.  That's inconvenient for contributors and causes confusion.

* * *

An alternative or supporting solution is to consume [unpkg](https://unpkg.com/#/) in some way.  I'm a little hesitant to depend on this service.  It's open-source, so we could run our own server, but $.

ScottFreeCode:
Resolves #2615


ScottFreeCode:
Do we have any mentions of Bower support we would need to edit away for this too?

boneskull:
Yes, so, `bower.json` needs to be removed, and we will need to update any relevant documentation.

I'm not sure if we need to do anything with Bower itself, like deprecate the package or whatever.  Please open an issue [here](https://github.com/bower/bower) or hit someone up in Discord chat...

boneskull:
This will also likely break `mocha init`.

ScottFreeCode:
> This will also likely break  mocha init .

Only for usage from git, right? From NPM `mocha init` should just keep using the published file?

> •we may need to change how we're using the  browser  field of  package.json  and just do that stuff on the command-line, then point  browser  to  mocha.js . this may impact users who are bundling Mocha themselves, but maybe not.

I'm not clear why our workflow in this regard would be affected; this may be a good change to make but if we're still publishing the same stuff it shouldn't be necessitated by the removal of the built file from version control.

> •contact Bower team if necessary

There's an unregister command, but I don't know if we want to *unregister* it given that versions < 4 should continue to work... Asking around and will update here when I have anything to show for it.

> •remove  bower.json
> •ensure  mocha.js  is in  files  field of  package.json

Will update these now...

ScottFreeCode:
`mocha.js` is already in `package.json`'s `files`; `bower.json` used to be in it, removed that. Also removed `bower.json`

ScottFreeCode:
According to https://github.com/bower/bower/issues/2482, I believe the idea is to make a copy of this repo containing only the old versions, then point Bower to that instead of this repo, then leave that repo as-is unless and until we have a need to change anything on Bower.

@boneskull Do you have the Bower permissions needed to do that?

boneskull:
no, I didn't register it initially.  you'll just have to ask them to change it--please let me know if you need assistance

boneskull:
> Only for usage from git, right? From NPM mocha init should just keep using the published file?

Yeah, you're right, n/m

From what I understand from that issue, basically all we need to do is remove `bower.json` when we tag v4.0.0.  This will break users who have depended on `mocha: "*"`, but really, that's not our problem.

boneskull:
so, I think I'm gonna close this.  I don't think we need to muck with `browser` for now, since it won't affect our builds.

PR #3034, opened 2017-09-29
Switch default from forced exit to no-exit

### LINKED ISSUE #2879: Do not force exit of a process (by default)
I'm talking about behaviour that can be switched off with `--no-exit` option.

Having this as default looks dangerous, as it may hide serious issues in tests (any orphaned async flow is just silenced, leaving no clue to developer)

e.g. on my side I had a promise with crashing test not returned properly. Still test process as handled by mocha was exiting smiling with a success. It took me a while to figure out why it happens.

ScottFreeCode:
Allows `--exit` to override and force exit as before. Resolves #2879

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -94,7 +94,7 @@ program
   .option('--inspect-brk', 'activate devtools in chrome and break on the first line')
   .option('--interfaces', 'display available interfaces')
   .option('--no-deprecation', 'silence deprecation warnings')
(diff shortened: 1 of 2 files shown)


boneskull:
Also, wait, does this actually work, just by changing it like that?

ScottFreeCode:
This doesn't affect the browser one way or the other as it's purely the Node CLI `process.exit` call.

ScottFreeCode:
And yeah, I think so -- just alter the Commander line and it switches the default value.

PR #3038, opened 2017-09-29
Deprecate --compilers

ScottFreeCode:
This attempt is a bit rough; we might have to tweak tests to tell Node to suppress the deprecation warning, for instance.


boneskull:
this won't make it into v4.0.0

boneskull:
well, hmm... eh.  it'll dump to STDERR.  *should* be okay...I can live with the fallout.

boneskull:
I'm gonna tweak the text a bit.

Issue #3035, opened 2017-09-29
Nuke `should`


ScottFreeCode:
I am tired of running into this: https://travis-ci.org/mochajs/mocha/builds/281135003?utm_source=github_status&utm_medium=notification [To save others digging: these tests are fine, they pass on Node, they would pass in browser but we're using only expect.js in browser because there used to be no `should` compatibility there.]

We need to go through the remaining uses of `should` in our tests and change them to `expect`, then remove `should` from our mocha.opts and any fixtures, then remove `should` from package.json!

ngeor:
I can pick this one up.

ngeor:
I run into a problem with `diffs.spec.js`, which might be a breaking change. In short, using `expect` gives less verbose diffs compared with `should`.

With expect:

```
  1) diffs
       should display a diff for small strings:
     Error: expected 'foo rar baz' to equal 'foo bar baz'
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
```

With should:

```
  1) diffs
       should display a diff for small strings:

      AssertionError: expected 'foo rar baz' to be 'foo bar baz'
      + expected - actual

      -foo rar baz
      +foo bar baz

      at Assertion.fail (node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
```



This is how `should` outputs errors (notice the diff):

```
9) diffs
       should work with objects:

      AssertionError: expected Object { name: 'tobi', species: 'ferret', color: 'white', age: 2 } to equal Object { name: 'loki', species: 'ferret', color: 'brown', age: 2 } (at name, A has 'tobi' and B has 'loki')
      + expected - actual

       {
         "age": 2
      -  "color": "white"
      -  "name": "tobi"
      +  "color": "brown"
      +  "name": "loki"
         "species": "ferret"
       }
      
      at Assertion.fail (node_modules/should/cjs/should.js:275:17)
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
```

On the other hand, these tests are effectively testing the behavior of `should` so I don't know if mocha should have them here in the first place.

I'd like some feedback on how to tackle this one :-)

ngeor:
PR created, waiting for feedback :)

ScottFreeCode:
Fixed in #3042

PR #3042, opened 2017-10-02
Nuke `should` #3035

ngeor:
### Description of the Change

The `should` library has been removed. All tests that were using it are now using expectJS.

Note that I also removed completely the diff integration test, because it was testing the output format provided by `should`.

### Benefits

One less package to worry about.

### Applicable issues

* This PR solves [issue 3035](https://github.com/mochajs/mocha/issues/3035)
* It shouldn't impact production code as this removes a dev dependency and only tests have been modified

### DIFF
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -31,7 +31,7 @@ describe('json reporter', function () {
       expect(runner).to.have.property('testResults');
       expect(runner.testResults).to.have.property('failures');
       expect(runner.testResults.failures).to.be.an('array');
-      expect(runner.testResults.failures.length).to.equal(1);
+      expect(runner.testResults.failures).to.have.length(1);
 
       var failure = runner.testResults.failures[0];
       expect(failure).to.have.property('title', testTitle);
@@ -52,7 +52,7 @@ describe('json reporter', function () {
       expect(runner).to.have.property('testResults');
       expect(runner.testResults).to.have.property('pending');
       expect(runner.testResults.pending).to.be.an('array');
-      expect(runner.testResults.pending.length).to.equal(1);
+      expect(runner.testResults.pending).to.have.length(1);
 
       var pending = runner.testResults.pending[0];
       expect(pending).to.have.property('title', testTitle);
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -27,16 +27,16 @@ describe('json reporter', function () {
     }));
 
     runner.run(function (failureCount) {
-      failureCount.should.be.exactly(1);
-      runner.should.have.property('testResults');
-      runner.testResults.should.have.property('failures');
-      runner.testResults.failures.should.be.an.instanceOf(Array);
-      runner.testResults.failures.should.have.a.lengthOf(1);
+      expect(failureCount).to.be(1);
+      expect(runner).to.have.property('testResults');
+      expect(runner.testResults).to.have.property('failures');
+      expect(runner.testResults.failures).to.be.an('array');
+      expect(runner.testResults.failures.length).to.equal(1);
 
       var failure = runner.testResults.failures[0];
-      failure.should.have.property('title', testTitle);
-      failure.err.message.should.equal(error.message);
-      failure.should.have.properties('err');
+      expect(failure).to.have.property('title', testTitle);
+      expect(failure.err.message).to.equal(error.message);
+      expect(failure).to.have.property('err');
 
       done();
     });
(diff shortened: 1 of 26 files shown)


boneskull:
awsome

ScottFreeCode:
Haven't reviewed all the changes in depth, but want to say *thank you* for taking this on!!

I just discovered that `expect.js` doesn't provide the diffs for its strict-equality comparison (`equal`/`be`) but does provide them for its deep-equality comparison (`eql`, no idea why not `deepEqual`). Do the diffing tests pass if `equal`/`be` in the assertion(s) that generate the diffs is/are changed to `eql`?

ngeor:
No problem at all, I was looking for something to help with and this one looked clear enough. By the way I'm more used to `chai` and it's also `eql` there, which one could argue it's not straightforward. I'll give it a try tomorrow with changing the assertions to `eql`, maybe that helps.

ngeor:
Good morning @ScottFreeCode , you were correct, switching to `eql` just worked for the diff tests. I'd be interested in your opinion on why we should keep these tests, given that they depend on an external tool to pass or fail (i.e. what would happen if `expect` had its completely own weird output format).

I squashed my commits into one. Looking forward to your feedback :-)

ScottFreeCode:
> Good morning @ScottFreeCode , you were correct, switching to  eql  just worked for the diff tests. 

🎉 

> I'd be interested in your opinion on why we should keep these tests, given that they depend on an external tool to pass or fail (i.e. what would happen if  expect  had its completely own weird output format).

The expected vs. actual format isn't specific to `expect.js` or `should`, it's common to many assertion libraries (I'm not sure if it's a standard, but it's at least a convention). So they don't really depend on one particular assertion library, just *an* assertion library with that common behavior.

If we don't already have unit tests that do so, we might be better off creating errors and assigning the expected and actual values to them ourselves and then throwing them or passing them into the `diff`ing stuff, instead of using an assertion library (it would be more... "unit-ey", for lack of a better term, and less "integration-ey"), but that doesn't need to happen at this point.

> I squashed my commits into one. Looking forward to your feedback :-)

Will give it a look soon! And thanks again!

test/reporters/json.spec.js
```diff
@@ -27,16 +27,16 @@ describe('json reporter', function () {
...
-      failureCount.should.be.exactly(1);
-      runner.should.have.property('testResults');
-      runner.testResults.should.have.property('failures');
-      runner.testResults.failures.should.be.an.instanceOf(Array);
-      runner.testResults.failures.should.have.a.lengthOf(1);
+      expect(failureCount).to.be(1);
+      expect(runner).to.have.property('testResults');
+      expect(runner.testResults).to.have.property('failures');
+      expect(runner.testResults.failures).to.be.an('array');
+      expect(runner.testResults.failures.length).to.equal(1);
```
ScottFreeCode: Nit: whereas `should` has `.have.a.lengthOf`, the equivalent in `expect` is `.have.length`, which yields better error messages than `equal`... Would merge this either way, but figure it's worth pointing out.

test/reporters/json.spec.js
```diff
@@ -27,16 +27,16 @@ describe('json reporter', function () {
...
 
       var failure = runner.testResults.failures[0];
-      failure.should.have.property('title', testTitle);
-      failure.err.message.should.equal(error.message);
-      failure.should.have.properties('err');
+      expect(failure).to.have.property('title', testTitle);
+      expect(failure.err.message).to.equal(error.message);
+      expect(failure).to.have.property('err');
```
ScottFreeCode: Wow, weird that we've got this test that's asserting the existence of a property only after using said property such that it would get a null reference exception if it didn't exist.

Let's not fix it in this PR, just keep this to should->expect updates. But we may want to drop in a follow-up commit to fix it.

test/reporters/json.spec.js
```diff
@@ -48,14 +48,14 @@ describe('json reporter', function () {
...
-      failureCount.should.be.exactly(0);
-      runner.should.have.property('testResults');
-      runner.testResults.should.have.property('pending');
-      runner.testResults.pending.should.be.an.instanceOf(Array);
-      runner.testResults.pending.should.have.a.lengthOf(1);
+      expect(failureCount).to.be(0);
+      expect(runner).to.have.property('testResults');
+      expect(runner.testResults).to.have.property('pending');
+      expect(runner.testResults.pending).to.be.an('array');
+      expect(runner.testResults.pending.length).to.equal(1);
```
ScottFreeCode: Another instance of the `length` nit discussed above.

PR #3048, opened 2017-10-04
Refine issue template

ScottFreeCode:
- Reorder to put transpiler troubleshooting directives after general troubleshooting directives.
- Flesh out troubleshooting directives.
- Add a few "if applicable" types of qualifiers.
- Make versions list visible by default.
- Add description of the meaning and relevance of a(n?) MCVE.


.github/ISSUE_TEMPLATE.md
```diff
@@ -23,6 +25,10 @@ Place an `x` between the square brackets on the lines below for every satisified
 <!--
 Please add a series of steps to reproduce the problem. See https://stackoverflow.com/help/mcve for in depth information
 on how to create a mimimal, complete, and verifiable example.
+
+TL;DR version: If we get a report like: "I use `library/tool X` and here's the one line of code that the error points to," we can provide troubleshooting suggestions but we generally can't know that if we use `library/tool X` we'll be doing the same thing that causes your issue. *Ideally* we would like to be able to copy-paste the code contents of your report (the smaller the better) into an empty directory and run `npm i mocha <and any other packages you specify>` followed by `mocha <arguments you specified>` (and/or any other commands necessary) and see the problem occur.
```
boneskull: This is too long; I didn't read it. :wink:

.github/ISSUE_TEMPLATE.md
```diff
@@ -23,6 +25,10 @@ Place an `x` between the square brackets on the lines below for every satisified
...
+
+*That doesn't mean we won't field issues that can't be so easily reproduced*, and we don't expect you to dig into Mocha specifically (we know not everyone is as familiar with it, especially some of the nuances under the hood). However, if it's possible to create such a case it will often narrow down the problem from "could be almost anything" to one of a few (or even just one) specific thing(s)! Isolating the problem like that is something we can't always do without lots of info from you (info that a reproducible test case can comprehensively provide -- see the link above for more on how and why), but *is a crucial step in getting you a solution*. Plus, creating such test cases is a great exercise in troubleshooting even outside of issue reports, so it helps you personally too!
```
boneskull: I think we can probably put this type of thing in the wiki or whatever for now instead of the issue template.  I'd like to ideally _remove stuff_ from the template since keeping it short is better DX, IMO

.github/ISSUE_TEMPLATE.md
```diff
@@ -32,16 +38,14 @@ on how to create a mimimal, complete, and verifiable example.
...
- If applicable, please specify:
- * The output of `mocha --version` and `node node_modules/.bin/mocha --version`:
- * The output of `node --version`:
- * The version and architecture of your operating system:
- * Your shell (bash, zsh, PowerShell, cmd, etc.):
- * Your browser and version (if running browser tests):
- * Any other third party Mocha related modules (with versions):
- * The code transpiler being used:
--->
+<!-- Please specify all of the following that are applicable: -->
```
boneskull: good

.github/ISSUE_TEMPLATE.md
ScottFreeCode: As it is I'm not sure everyone is following and reading the link to the detailed explanation/rationale (from StackExchange) for MCVEs. I was hoping to give an *alternative* that's available right here, right now and briefer -- without leaving out any of the necessary steps and without just demanding "Do X and Y and Z" without a reason (or some sympathy, since it is more involved for the user). I'd rather try to tweak it to streamline that than risk just replicating the "people appear not to be reading the StackExchange link" problem with our own wiki, if possible.

As long as the only feature GitHub gives us is to put in a template full of *information*, we are kind of between a rock and a hard place getting people to pay attention to all the relevant information. If we could actually place requirements on fields to fill out that would be validated before the issue is filed, we could solve this, but as-is the best we can hope for is a balance between "too much information in the template" and "too much information outside the template". Unfortunately I suspect the users who really need the information spelled out for them are also the users most sensitive to that balance...

ScottFreeCode: How about the following simplification?

> Ideally, include minimal code that we can copy-paste into an empty project and run to see the issue. Not every issue has to meet that ideal, but for those that can, it ensures we have all the info needed to look for a solution to your problem -- and you may even solve it yourself while trying to narrow the code down!

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,18 @@
...
+    * Transpilers often do nothing to indicate that they ran OK but didn't change the code you expected them to change; if the problem appears when using a transpiler and Mocha, it usually is still there if you use just the transpiler, but you have to actually run the transpiled code to tell (since the transpiler doesn't know that what it transpiled didn't include what you wanted it to include).
+    * Using multiple sets of tools together, such as a transpiler, a bundler and a browser, can make it inobvious which tool isn't doing its part (e.g. a bundler or a browser might be handling syntax you thought you transpiled, when in fact the transpiler is not working), especially if the tests do not use the entire same toolset and environment.
 * [ ] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
 `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.
```
ScottFreeCode: We could also add to the prerequisites list a link to a new wiki page "Common Issues". This could be in addition to the entries already here, or in place of some of the details about things like transpilers. I wouldn't want it to replace the recommendation to search the issue tracker in various ways; but it might be best to link to the "Common Issues" page before the issue tracker since it would provide more straightforward fixes *and* more detailed explanations with less noise. That page could have the following format:

# Common Issues

## How to fix

### Name of Issue

- examples: "error message here"
- fix: do x instead of y
- more info: [link to corresponding issue in details section]

### Name of another issue

same type of list here

## Details (or "Why?")

### Name of Issue

This is an issue that occurs when you foo the bar. Bar expects to be fizzed or buzzed rather than foo'd, so it throws an error. You can manually fizz and buzz the bar instead, but it's easier and more reliable to tell Mocha you're using a bar and let Mocha handle the fizzing and buzzing.

### Name of Other Issue

This is an issue with bundling and transpiling at the same time and then running against a browser vs. the current version of Node only partly supporting New Idea Voodoo. As a language feature there's nothing Mocha can do to implement it (and it should not anyway, since your tests should indicate whether the code will work in environments that don't have the test runner "helping" them), and nothing Mocha can do to interfere with it when it is implemented either. However, the implementations available at this point can be very complex due to the interaction between transpilers and bundlers and the differences between how browsers have done it and how Node has done it. We recommend consulting the documentation and support for these projects, where you can get more detailed and up-to-date info and better assistance.

ScottFreeCode: (Tangentially, this would be the sort of wiki page that might be better off as part of our new docs collection once we split up the site into multiple pages of documentation.)

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,18 @@
...
+* [ ] Checked that your issue isn't already filed by searching for relevant keywords in the [issue tracker](https://github.com/mochajs/mocha/issues) and cross-referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20) <!-- NOTE: in the event that you find a relevant existing issue but its solution does not work for you, it is possible that you have a different although similar issue; in such cases opening a new issue is appropriate. ***However***, in the event you do wish to add to an existing issue (e.g. because it has no solution yet), **we highly recommend following the same steps and instructions as in reporting a new issue**, to minimize comments troubleshooting your particular case on someone else's issue. If you don't have new information to provide, just hit the "thumbs-up" reaction on the issue; we can look at those numbers to get an idea of how many people are affected without spam in the comments. -->
```
boneskull: my main worry about this template is that people won't read it because it's too verbose, honestly.  the more detailed information can be moved into `CONTRIBUTING.md` or some new document / wiki entry about how issues should be handled

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,18 @@
...
+* [ ] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself. If using any additional tooling make sure to use the exact same toolset outside Mocha (see below).
+* [ ] Checked next-gen ES issues and syntax problems (if any) by using ***the same*** *environment and/or transpiler configuration* ***without Mocha*** to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code. For example:
+    * Transpilers often do nothing to indicate that they ran OK but didn't change the code you expected them to change; if the problem appears when using a transpiler and Mocha, it usually is still there if you use just the transpiler, but you have to actually run the transpiled code to tell (since the transpiler doesn't know that what it transpiled didn't include what you wanted it to include).
```
boneskull: lines 14 and 15 should be moved away from the issue template b/c TL;DR.  people barely read these anyway.

boneskull: a concise checklist is preferable, imo

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,18 @@
...
+* [ ] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself. If using any additional tooling make sure to use the exact same toolset outside Mocha (see below).
```
boneskull: unless we can provide a simple, straightforward way to determine if the issue is *Mocha* or *something else* (which we can't, sadly), we could trim this down to something like:

* * *

* [ ] If a bug, I am reasonably confident Mocha is to blame

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,18 @@
 <!--
 Have you read Mocha's Code of Conduct? By filing an Issue, you are expected to comply with it, including treating everyone with respect: https://github.com/mochajs/mocha/blob/master/.github/CODE_OF_CONDUCT.md
```
boneskull: this can be more concise

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,18 @@
...
+* [ ] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself. If using any additional tooling make sure to use the exact same toolset outside Mocha (see below).
+* [ ] Checked next-gen ES issues and syntax problems (if any) by using ***the same*** *environment and/or transpiler configuration* ***without Mocha*** to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code. For example:
```
boneskull: like previously:

* * *

* [ ] Ensure your environment supports all JavaScript features you are attempting to use

.github/ISSUE_TEMPLATE.md
boneskull: this is pretty much the definition of a FAQ.

.github/ISSUE_TEMPLATE.md
```diff
@@ -21,8 +24,7 @@ Place an `x` between the square brackets on the lines below for every satisified
 ### Steps to Reproduce
 
 <!--
-Please add a series of steps to reproduce the problem. See https://stackoverflow.com/help/mcve for in depth information
-on how to create a mimimal, complete, and verifiable example.
+Please add a series of steps to reproduce the problem. See https://stackoverflow.com/help/mcve for in depth information on how to create a mimimal, complete and verifiable example. TL;DR version: *Ideally*, include *all* (but *only*) the necessary *code, configuration and commands* that we can copy-paste into an empty project and run to see the issue. Not every issue has to meet that ideal, but for those that can, it ensures we have all the info needed to look for a solution to the problem -- and you may even solve it yourself while narrowing the code down!
```
boneskull: can we just copy what SO has written and provide the URL? ..

* * *

Please provide a [Minimal, Complete, Verifiable Example](https://stackoverflow.com/help/mcve).

- **Minimal** – Use as little code as possible that still produces the same problem
- **Complete** – Provide all parts needed to reproduce the problem
- **Verifiable** – Test the code you're about to provide to make sure it reproduces the problem

.github/ISSUE_TEMPLATE.md
boneskull: actually, grammar:

* * *

```markdown
<!-- Please specify any of the following which are applicable: -->
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -32,16 +34,14 @@ on how to create a mimimal, complete, and verifiable example.
...
- * The version and architecture of your operating system:
- * Your shell (bash, zsh, PowerShell, cmd, etc.):
- * Your browser and version (if running browser tests):
- * Any other third party Mocha related modules (with versions):
- * The code transpiler being used:
--->
+<!-- Please specify all of the following that are applicable: -->
+* The output of `mocha --version` and `node node_modules/.bin/mocha --version`:
+* The output of `node --version`:
+* The version and architecture of your operating system:
```
boneskull: some may not know what this means.  maybe this instead:

* * *
```markdown
* Your operating system name and version:
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -32,16 +34,14 @@ on how to create a mimimal, complete, and verifiable example.
...
+* Your shell (bash, zsh, PowerShell, cmd, etc.):
+* Your browser and version (if running browser tests):
```
boneskull: the "ifs" here would be redundant because of the "where applicable" comment

.github/ISSUE_TEMPLATE.md
```diff
@@ -32,16 +34,14 @@ on how to create a mimimal, complete, and verifiable example.
...
+* Any other third party Mocha related modules (with versions):
+* The code transpiler being used (if any):
```
boneskull: maybe just "Relevant transpiler configuration"?

boneskull: Also, please add "The command used to execute Mocha"

PR #3051, opened 2017-10-05
Update growl to 1.10.3

dpogue:
### Description of the Change

Update growl to 1.10.3

This avoid a warning about an eslint peerDependency on install of mocha:

```
eslint-plugin-node@5.2.0 requires a peer of eslint@>=3.1.0 but none is installed.
```


boneskull:
@dpogue thanks!

PR #3055, opened 2017-10-06
Add test for Mocha.bail()

38elements:
### Description of the Change

This adds a test case when `mocha.bail()` has  no arguments.


boneskull:
@38elements nice, thanks.

PR #3061, opened 2017-10-08
Remove unused variables

38elements:
### Requirements

The `escape` argument for `inlineDiff()` and `unifiedDiff()` in `Base.list()` is always false.
[inlineDiff()](https://github.com/mochajs/mocha/blob/9f204ba2b27e254e70e88ab353e8c4cedc7b67f6/lib/reporters/base.js#L372) and [unifiedDiff()](https://github.com/mochajs/mocha/blob/9f204ba2b27e254e70e88ab353e8c4cedc7b67f6/lib/reporters/base.js#L406) are only used in `Base.list()` .
I think the `escape` argument is unnecessary.

https://github.com/mochajs/mocha/blob/9f204ba2b27e254e70e88ab353e8c4cedc7b67f6/lib/reporters/base.js#L188-L220

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -367,7 +367,7 @@ function pad (str, len) {
  * @return {string} Diff
  */
 function inlineDiff (err) {
-  var msg = errorDiff(err, 'WordsWithSpace', false);
+  var msg = errorDiff(err);
 
   // linenos
   var lines = msg.split('\n');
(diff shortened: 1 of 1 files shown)


boneskull:
good catch btw--it looks like this variable became useless somewhere around when inline diffs were added

38elements:
Thank you for reviewing.
Since `errorDiff()` is only used in `inlineDiff()`, I removed `type` and `escape`  arguments.

boneskull:
@38elements thanks!

PR #3064, opened 2017-10-09
normalize getters and setters (fixes #3058)

### LINKED ISSUE #3058: normalize setters and getters
The main classes (`Runnable`, `Test`, `Context`, `Suite`, `Mocha`, etc.) have these setter/getter functions, e.g. `Runnable#timeout()`.  Not all of these, however, implement the "getter" part, where it's supposed to just return the value if it receives no arguments.

Furthermore, there are some methods that might check if the first argument is `undefined`; they need to check length of `arguments` instead, because `undefined` or `void 0` is a valid argument.

makepanic:
### Description of the Change

This PR adds missing argument checks and adds a couple getters to some of the main classes.
Should fix #3058.

If there are additional getters that I've missed, I can add them.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -29,7 +29,7 @@ Context.prototype.runnable = function (runnable) {
 };
 
 /**
- * Set test timeout `ms`.
+ * Set or get test timeout `ms`.
  *
  * @api private
  * @param {number} ms
@@ -51,18 +51,24 @@ Context.prototype.timeout = function (ms) {
  * @return {Context} self
  */
 Context.prototype.enableTimeouts = function (enabled) {
+  if (!arguments.length) {
+    return this.runnable().enableTimeouts();
+  }
   this.runnable().enableTimeouts(enabled);
   return this;
 };
 
 /**
- * Set test slowness threshold `ms`.
+ * Set or get test slowness threshold `ms`.
  *
  * @api private
  * @param {number} ms
  * @return {Context} self
  */
 Context.prototype.slow = function (ms) {
+  if (!arguments.length) {
+    return this.runnable().slow();
+  }
   this.runnable().slow(ms);
   return this;
 };
@@ -78,7 +84,7 @@ Context.prototype.skip = function () {
 };
 
 /**
- * Allow a number of retries on failed tests
+ * Set or get a number of allowed retries on failed tests
  *
  * @api private
  * @param {number} n
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -91,14 +91,14 @@ Runnable.prototype.timeout = function (ms) {
 };
 
 /**
- * Set & get slow `ms`.
+ * Set or get slow `ms`.
  *
  * @api private
  * @param {number|string} ms
  * @return {Runnable|number} ms or Runnable instance.
  */
 Runnable.prototype.slow = function (ms) {
-  if (typeof ms === 'undefined') {
+  if (!arguments.length || typeof ms === 'undefined') {
     return this._slow;
   }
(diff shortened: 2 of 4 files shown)


boneskull:
@makepanic thanks

lib/runnable.js
```diff
@@ -91,14 +91,14 @@ Runnable.prototype.timeout = function (ms) {
...
- * Set & get slow `ms`.
+ * Set or get slow `ms`.
  *
  * @api private
  * @param {number|string} ms
  * @return {Runnable|number} ms or Runnable instance.
  */
 Runnable.prototype.slow = function (ms) {
-  if (typeof ms === 'undefined') {
+  if (!arguments.length || typeof ms === 'undefined') {
```
boneskull: I think we can drop the `ms` check

lib/suite.js
```diff
@@ -165,7 +165,7 @@ Suite.prototype.slow = function (ms) {
 };
 
 /**
- * Sets whether to bail after first error.
+ * Sets or gets whether to bail after first error.
```
boneskull: maybe just `Set or get` like the others?

lib/context.js
```diff
@@ -44,25 +44,31 @@ Context.prototype.timeout = function (ms) {
...
 /**
- * Set test timeout `enabled`.
+ * Set or get test timeout `enabled`.
  *
  * @api private
  * @param {boolean} enabled
  * @return {Context} self
  */
 Context.prototype.enableTimeouts = function (enabled) {
+  if (!arguments.length) {
```
boneskull: (fwiw, this is a breaking change for those who are consuming it directly.  as it's undocumented, they shouldn't be.)

lib/runnable.js
makepanic: I've initially removed the `ms` check but tests were failing [here](https://github.com/mochajs/mocha/blob/master/test/unit/runnable.spec.js#L80).
I'm unsure if removing the ms check will introduce a breaking change.

lib/context.js
makepanic: So should i remove the arguments check from enableTimeouts?

makepanic:
I've updated the PR to fix the jsdoc issue and remove the arguments check from `enableTimeouts`.

lib/runnable.js
boneskull: ugh, hum, let me think about this.

boneskull: that test is invalid and should be removed, *but* it's also a breaking change, so let's leave it for now.

lib/context.js
boneskull: sorry to confuse; no, please leave it.  I don't have any evidence this is used directly.

boneskull: (looks like you'll need to revert)

makepanic:
I've reverted removing the `enableTimeouts` getter and left the `Runnable.slow` ms check in.

lib/runnable.js
fesebuv: The reason why you test fails, is because when you invoke `run.slow(undefined);` `argument.length` is equal to 1. Instead of checking the number of arguments passed, you should just check if `ms` is undefined. You can do something like this:
```
if(!ms) {
  ....
}
```

boneskull: @fesebuv that's correct.  but we just want to leave as-is for now.  there's no change necessary here, since it'll be breaking.

boneskull:
@makepanic thanks!

lib/runnable.js
fesebuv: gotcha!

Issue #3068, opened 2017-10-13
Karma + Mocha + reporter options crash


Itee:
### Prerequisites

* [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
* [x] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
* [x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
* [x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
`node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description
Mocha crash when setup option that are not prototyped (a.k.a function)
Error is: Uncaught TypeError: this[opt] is not a function at node_modules/mocha/mocha.js:135

### Steps to Reproduce
Create a karma boilerplate and setup in karma.conf.js:

```javascript
        // Pass configuration options directly to mocha
        client: {
            mocha: {
                reporter:        'myReporter',
                reporterOptions: { /*Ka-boom*/ }
            }
        }
```

**Expected behavior:** This to be fixed ?

**Actual behavior:** This crash the test run when set reporter options in karma.conf.js

**Reproduces how often:** 100%

### Versions
Mocha version: 4.0.1
Node version: 8.7.0
System: Win10 x64
Shell: any
Browser: any
Third party: Karma v1.7.1
No transpiling

### Additional Information

https://github.com/Itee/karma-mocha-reporterOption-crash
PR: #3069

ScottFreeCode:
Hi @ltee, could you elaborate on what you're trying to accomplish with a custom reporter in Karma?

If you want to customize the console output when running with Karma I believe you'd have to write a reporter for Karma rather than Mocha, and in that case Mocha's reporter options would not apply, unless I'm mistaken.

[...]

PR #3069, opened 2017-10-13
Allow to set reporterOptions from setup

Itee:
This is a hot fix for #3068 

When karma start running Mocha it set the Mocha options with mocha.setup with apply options like this:

```javascript
mocha.setup = function (opts) {

    opts = { ui: opts };
  }

  for (var opt in opts) {
    if (opts.hasOwnProperty(opt)) {
      this[opt](opts[opt]);                 // BOOM !!!
    }
  }
  return this;
};
```

But... reporterOptions is not a function so karma crash at line 144.

This PR fix it

### Requirements

### Description of the Change

Add a Mocha prototype to handle the reporterOptions as function when call inside mocha.setup

### Alternate Designs
Check for options type inside setup instead of make a savage function call ?

Because this fix a bug that crash test runner ?

### Benefits
This fix the crash

### Possible Drawbacks
Allow user to set the reporterOptions ?

### Applicable issues
This is a patch !

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -185,7 +185,7 @@ Mocha.prototype.reporter = function (reporter, reporterOptions) {
  */
 Mocha.prototype.reporterOptions = function (reporterOptions) {
 
-  this.options.reporterOptions = this.options.reporterOptions || reporterOptions;
+  Object.assign(this.options.reporterOptions, reporterOptions);
   return this;
 
 }
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -180,6 +180,17 @@ Mocha.prototype.reporter = function (reporter, reporterOptions) {
 };
 
 /**
+ * Set reporter options from mocha.setup call
+ * @param {Object} reporterOptions optional options
+ */
+Mocha.prototype.reporterOptions = function (reporterOptions) {
+
+  this.options.reporterOptions = this.options.reporterOptions || reporterOptions;
+  return this;
+
+}
+
+/**
  * Set test UI `name`, defaults to "bdd".
  *
  * @api public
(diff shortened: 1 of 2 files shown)


lib/mocha.js
```diff
@@ -180,6 +180,17 @@ Mocha.prototype.reporter = function (reporter, reporterOptions) {
 };
 
 /**
+ * Set reporter options from mocha.setup call
+ * @param {Object} reporterOptions optional options
+ */
+Mocha.prototype.reporterOptions = function (reporterOptions) {
+
+  this.options.reporterOptions = this.options.reporterOptions || reporterOptions;
```
ScottFreeCode: This code as written will ignore the specified options if `this.options.reporterOptions` already exists; that behavior is likely to be confusing at best and may even be erroneous.

There's some question, in my mind, as to whether existing options should be *merged* with new options, e.g. using `Object.assign(this.options.reporterOptions, reporterOptions)`, or just overwritten altogether. Assuming this function is desirable, whether to merge or overwrite would merit further discussion.

Itee: I'm totally agreed ! I don't known what will be the options management inside the mocha core, and what are the desired usage for it.

All i known for the moment is that the current Mocha prototypes doesn't allow to use the setup methods for all available options in Mocha.

Itee: I will add Object.assign in next commit. The only case where this could happen is, i create a new Mocha instance with reporterOptions as arguments, and next call setup methods on it. Is this a real use case ? Anyway this need to be properly handle... so...

Itee:
Due to #3068 
Should i close this PR or the context to support options from setup call is still relevant ?

ScottFreeCode:
Yeah, I think we'll pass on this unless and until there's a need. Thanks for giving it a try anyway, though!

Itee:
No problem !

PR #3082, opened 2017-10-24
Move re variable definition in lookupFiles Utility

Bamieh:
### Description of the Change

I am using mocha programmatically, along with its utility `lookupFiles`.

Upon calling `lookupFiles` A regular expression must be evaluated to lookup files although it is not needed in the majority of cases. Shifting the `re` variable definition down to where it is needed will avoid this issue.

The main reason for this change is to eliminate the requirement of always sending an array to the extensions parameter, moving the `re` variable definition to just right above where it is needed will achieve both goals.

### Alternate Designs

An if statement with a default value for the extensions input param. The proposed version is the neatest and most verbose way to do it.

I am using mocha programmatically along with its utility, this is just a tiny enhancement. also it dictates less knowledge about the function internals to use it since it marks the extension parameter as optional, rather than a required array.

### Benefits

1. Extensions input parameter to the function is now optional, no need to always send an array.
2. No need to evaluate a regex that might not be used in the majority of use cases.

### Possible Drawbacks

Its a super minor change so i dont see any drawbacks.

### Applicable issues
Patch release, a tiny enhancement.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -499,7 +499,6 @@ exports.lookupFiles = function lookupFiles (path, extensions, recursive) {
     return;
   }
 
-  var re = new RegExp('\\.(' + extensions.join('|') + ')$');
   readdirSync(path).forEach(function (file) {
     file = join(path, file);
     try {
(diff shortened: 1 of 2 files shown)


boneskull:
@Bamieh Thanks.  This looks good.  Was wondering if you could add a test to assert that `extensions` is no longer required?

Bamieh:
@boneskull i added a couple of unit tests

Issue #3084, opened 2017-10-25
mocha no-ops when calling require() on file that is in node cache


finnigantime:
### Prerequisites
* [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
* [x] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
* [x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
* [x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
`node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description
Mocha no-ops when calling require() on file that is in node cache. This occurs when using mocha programmatically. We have a test script that calls mocha twice - the first time, we don't actually run tests, we just build up the mocha suites and tests, and we run some filtering on them. We return the subset of files that include tests matching these filters. On the second time we call mocha, we actually run the tests on the set of matching files. However, in some node environments I am seeing `require(file)` pick up the file from the node cache instead of actually loading it and parsing it, so the `require(file)` call ends up being a no-op for mocha (mocha depends on the file being executed during `require(file)` to invoke `describe()` and `it()` statements and emit test code). Seems like this problem can be easily avoided by ensuring the file is not in the require cache before calling `require(file)`.

### Steps to Reproduce
Invoke mocha in two passes in a test script.

In the first pass, load files and look over `mocha.suite.eachTest` and do some filtering. Return a subset of files that we want to run tests on for this script run. 
```
  mocha.files = files;
  mocha.loadFiles();
```

In the second pass, with a separate mocha instance, for each matched file we add it to mocha and then call `mocha.run()`.
```
  mocha.addFile(file);
  mocha.run();
```

**Expected behavior:** In the second pass, mocha parses tests in the file and runs them.

**Actual behavior:** In the second pass, mocha no-ops.

**Reproduces how often:** 100% (in my node env - node v6.11.0)

### Versions
mocha --version: 4.0.1
node --version: 6.11.0
OSX Sierra 10.12.6
zsh

### Additional Information
N/A

[...]

PR #3085, opened 2017-10-25
ensure file is not in cache before calling require

finnigantime:
### Description of the Change

Fixes https://github.com/mochajs/mocha/issues/3084.

### Alternate Designs

Could have modified `require.cache` directly, but seems easier to add the `decache` dependency and it handle modifying the `require.cache` for us.

Mocha makes an assumption that when loading files, `require(file)` will parse a file and encounter mocha directives. For this to be valid, mocha should ensure the file is not in the require cache.

### Benefits

Allows mocha to be used programmatically in scripts/environments where modules mocha is going to run tests on have already been loaded (perhaps by a previous invocation of mocha itself).

### Possible Drawbacks

None that come to mind immediately, other than risk by having the `decache` dependency, though I think having that dependency makes sense here.

### Applicable issues

bug fix

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -10,6 +10,7 @@
  * Module dependencies.
  */
 
+var decache = require('decache');
 var escapeRe = require('escape-string-regexp');
 var path = require('path');
 var reporters = require('./reporters');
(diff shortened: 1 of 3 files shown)


finnigantime:
How do I fix `Error: Cannot find module 'decache'` in travis CI?

boneskull:
thanks.  as @ScottFreeCode wrote in #3084, we can't just blast the cache wholesale.  if we're going to have this capability baked in to core, the solution is going to be much more nuanced.

PR #3101, opened 2017-11-10
Remove unused argument of lib/ms

38elements:
### Description of the Change

Since `options` argument is unused in Mocha, this removes it.

https://github.com/mochajs/mocha/blob/2303c669ef1b47fef7cf86f5ef486d537033d443/lib/suite.js#L108-L110

https://github.com/mochajs/mocha/blob/2303c669ef1b47fef7cf86f5ef486d537033d443/lib/runnable.js#L82-L84

https://github.com/mochajs/mocha/blob/1bb6b3977a482c144a61d68ea2e499b22a7db01b/lib/reporters/base.js#L324-L326

### DIFF
--- a/lib/ms.js
+++ b/lib/ms.js
@@ -19,16 +19,13 @@ var y = d * 365.25;
  *
  * @api public
  * @param {string|number} val
- * @param {Object} options
  * @return {string|number}
  */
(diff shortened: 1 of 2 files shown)


boneskull:
@38elements lgtm, thanks!

Issue #3071, opened 2017-10-18
html reporter needs tests


boneskull:
We have decent coverage in *most* places now.  The glaring exception is the HTML reporter, which is completely untested.

Add tests for the HTML reporter.

(Note: this can probably be accomplished with [jsdom](https://npm.im/jsdom))

ScottFreeCode:
I have branches that do this with Karma that I need to rebase (and see if the issues on Sauce went away or not). Bonus: the test coverage for the browser, while I implemented it for the reporter, theoretically opens the door to browser-specific units tests if we still have anything that would need it (I dunno if all of those were shims we've now gotten rid of) or browser integration tests (e.g. for https://github.com/mochajs/mocha/pull/1800#issuecomment-319479473) with alternate Karma settings (something I still need to explore to ensure they're set up right).

rugk:
@ScottFreeCode You are still assigned to this even though a PR is there in https://github.com/mochajs/mocha/pull/3102.

craigtaub:
@rugk seems like the owner has lost interest, perhaps checkout branch, rebase and re-open PR if you are still interested. It sounded close if passes IE11 testing.

juergba:
@giltayar do you have any suggestions how to start with this one?
We also have some funds available we could invest.

juergba:
There is _playwright_ of Microsoft.
They also offer their own test runner for Node. _Folio_, "a highly customizable test framework", available in preview.
This is going to be a very competitive player in test frameworks.

giltayar:
There's no lack of browser automation tools, although not sure this is needed. I would go with a simpler snapshot test (i.e. save the HTML baseline file, and always compare against that). Usually, I'm not a fan of these, but given that we're looking at one or two tests, I think it's the easiest solution.

If you _do_ want a browser automation test. Given that there aren't a lot of tests here, I would go with the tried and true: Selenium WebDriver.

SmokeDawg92967:
I don't believe this

mark-wiemer:
Marking this as one I'm interested in, I use Playwright as a Microsoft employee and it's excellent (I'm biased ;) ) but I agree it may not be necessary. Happy to work with whoever wants to get started on this, as I'm currently focusing on some other issues.

PR #3102, opened 2017-11-11
Add unit test for html reporter

38elements:
### Description of the Change

This adds unit test for html reporter.
This is related to #3071 .


38elements:
I am sorry.
I did not notice the assignee.

ScottFreeCode:
Actually, *I'm* sorry I didn't reply to this sooner -- no need to worry that the issue's assigned (that just helps us track who's looking at it) or even that there's already work toward it, we're definitely open to alternative attempts and new contributions! While my preference is to ensure things are tested and coverage collected in all the "real" browsers Mocha supports, I still need to figure out whether that's even going to be doable with the recent upgrades relating to Mocha's test harness for browser usage. So, while I can't guarantee which we'd pick at this point, if you've already coded it up we may as well see how it compares. Would you mind keeping this open for further assessment (or, at least, make sure the fork/branch stays available)? Thanks! (And thanks for taking a shot at adding test for this, in any case!)

38elements:
Hi @ScottFreeCode.
Thank you for your reply.
I reopen this pull request.

boneskull:
a jsdom-based test is better than nothing imo.

38elements:
Thank you for reply.
I will change this in within 3 days.

38elements:
I changed this.:smile:

test/browser-specific/html-reporter.spec.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+
+  function addMochaElem () {
+    var mochaElem = document.createElement('div');
+    mochaElem.id = 'mocha';
+    document.body.appendChild(mochaElem);
+  }
+
+  function sleep (fn, ms) {
```
boneskull: this in particular is going to be brittle / susceptible to race conditions.  but this test suite is unlike anything else we have, so we don't have the tooling.

all of assertions involving markup in `innerHTML` are also going to be brittle, but again, need better tooling.

I'm going to grab this PR and test it against saucelabs' browsers.  if it is flaky, we'll have to change it.  but if not, we'll just merge and deal w/ it later--unless, of course, you feel like pulling in [Nightwatch](http://nightwatchjs.org).

test/browser-specific/html-reporter.spec.js
```diff
@@ -0,0 +1,157 @@
...
+    return new Promise((resolve) => {
```
boneskull: IE11 breaks with these

38elements:
Thank you for reviewing.
I will change this within 5 days.

38elements:
I replaced the arrow functions and the comparing HTML attributes.
I added `es6-promise`.
But, I did not test by IE11.

test/browser-specific/html-reporter.spec.js
38elements: Thank you for reviewing.
Since I think I will not be able to achieve it,
Would you please grab this pull request ?

Nokel81:
Have you managed to test on IE11 yet?

rugk:
Any news here?

rugk:
@38elements So what?

PR #3143, opened 2017-12-11
return the created test in `xit` for bdd interface

### LINKED ISSUE #3142: Add timeout option to xits using arrow functions
### Prerequisites

### Description

When using the arrow function syntax to override the default timeout. You specifiy it on the return value of the it function.

    it("should override the timeout", () => { }).timeout(100);

But if we want to ignore the test by adding an x we get an exception, which is annoying.

### Steps to Reproduce

    // Throws
    xit("should override the timeout", () => { }).timeout(100);

**Expected behavior:** The test should be ignored

**Actual behavior:** We get an exception
>> Mocha` exploded!
>> TypeError: Cannot read property 'timeout' of undefined

### Versions
mocha 4.0.1

Bamieh:
### Description of the Change

Fixes #3142 

The bigger reason for this fix is to ensure consistency across across both active and pending tests (ie `it` and `xit`).

I've also added a test suite for `setting timeout by appending it to test`

### Alternate Designs

We can also remove the return statement to disable this feature in the bdd interface, but i personally favor having it around, maybe even document it after adding some more test cases!

### Benefits

Since @boneskull does not encourage the use of the `this` keyword in writing tests, documenting this change might create a shift to increased usage of arrow functions and less usage of `this`.

### Applicable issues
**patch**

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,7 @@ install:
   - set CI=true
   - set PATH=%APPDATA%\npm;c:\MinGW\bin;%PATH%
   - set PHANTOMJS_CDNURL=https://cnpmjs.org/downloads
+  - npm install -g npm
   - npm install
   - copy c:\MinGW\bin\mingw32-make.exe c:\MinGW\bin\make.exe
 matrix:
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -102,7 +102,7 @@ module.exports = function (suite) {
      */
 
     context.xit = context.xspecify = context.it.skip = function (title) {
-      context.it(title);
+      return context.it(title);
     };
 
     /**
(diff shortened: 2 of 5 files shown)


boneskull:
test failure on AppVeyor is curious

boneskull:
@Bamieh it's helpful to write "closes #xxxx" in the commit message somewhere as it will automatically close the associated issue upon merge, as well as link the commit to the issue (#3143)

boneskull:
we're seeing a *seemingly* random npm failure under Node.js v4, so I'm upgrading to npm v5.6.0.  If this fails I'll create `npm-debug.log` as an artifact and try to get more info...

boneskull:
saucelabs flake? really?  at least it's flake I haven't seen before.  retrying

Issue #3139, opened 2017-12-09
Can not import Mocha as a native ES6 module in a browser


vitalets:
### Prerequisites
* [ x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
* [x ] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
* [ x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
* [ x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
`node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description
Can not import Mocha as a native ES6 module in a browser (Chrome 62). I tried several variant with `import` statement but with no success:
```js
import 'https://unpkg.com/mocha@4.0.1/mocha.js';
// or 
import mocha from 'https://unpkg.com/mocha@4.0.1/mocha.js';
// or 
import mocha from 'https://unpkg.com/mocha@4.0.1/index.js';
```

### Steps to Reproduce
See jsFiddle: https://jsfiddle.net/Lph8mrbe/2/

**Expected behavior:**
Mocha will be loaded into global context without errors.

**Actual behavior:**
Error: 
```
mocha.js:1 Uncaught ReferenceError: require is not defined at mocha.js:1
```

**Reproduces how often:**
Always

### Versions
4.0.1

ScottFreeCode:
Sounds like it's caused by the same thing as #3091?

boneskull:
duplicate of #3091

boneskull:
@vitalets please pull down PR #3145 to test a fix

vitalets:
> @vitalets please pull down PR #3145 to test a fix

@boneskull I've checked -  importing Mocha a ES module works for `mocha.js` in this PR.

PR #3145, opened 2017-12-12
fix bundling; closes #3091

### LINKED ISSUE #3091: Cannot find module issue with mocha.js in browser
Hi,

I'm using mocha with requirejs for browser testing and got a weird issue when updated to 4.0.1. 

By some reason I got the following issue when run tests:
screenshot 2017-11-03 20 26 37

You can find the code that reproduce the issue here: https://github.com/killmenot/mocha-cannot-find-module-issue (just do `npm install` and then open testrunner.html in browser or run `npm run test`)

I took a look at mocha.js (I run `make mocha.js`) I found the following:
screenshot 2017-11-03 20 28 40

When I remove `require=` from mocha.js file the test works fine.

Any advices?

boneskull:
This became too difficult to manage on the command-line.

What we're trying to do here is ship a version of the `Buffer` shim
which plays well with IE9/IE10.  Browserify ships with a version which
does NOT play well, meaning we have to force it to use the version
we choose (`buffer@4.9.x`).

The fix is in two parts:

1.  `insertGlobalVars` option replaces usages of global `Buffer` with
`require('/path/to/mocha/node_modules/buffer').Buffer`
2.  Any *other* module which explicitly requires `buffer` or, yes,
`buffer/`, must *also* use `/path/to/mocha/node_modules/buffer`

If *both* of these are not in place, Browserify will use its *own*
version of the `buffer` shim.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-BROWSERIFY := "node_modules/.bin/browserify"
+BROWSERIFY := "scripts/build.js"
 KARMA := "node_modules/.bin/karma"
 MOCHA := "bin/mocha"
 NYC := "node_modules/.bin/nyc"
@@ -20,13 +20,7 @@ all: mocha.js
 mocha.js: $(SRC) browser-entry.js
 	@printf "==> [Browser :: build]\n"
 	mkdir -p ${@D}
-	$(BROWSERIFY) ./browser-entry \
-		--require buffer/:buffer \
-		--plugin ./scripts/dedefine \
-		--ignore 'fs' \
-		--ignore 'glob' \
-		--ignore 'path' \
-		--ignore 'supports-color' > $@
+	$(BROWSERIFY) > $@
 
 clean:
 	@printf "==> [Clean]\n"
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ NYC := "node_modules/.bin/nyc"
 
 ifdef COVERAGE
 define test_node
-	$(NYC) --no-clean --report-dir coverage/reports/$(1) $(MOCHA)
+$(NYC) --no-clean --report-dir coverage/reports/$(1) $(MOCHA)
 endef
 else
-	test_node := $(MOCHA)
+test_node := $(MOCHA)
 endif
 
 TM_BUNDLE = JavaScript\ mocha.tmbundle
@@ -32,7 +32,7 @@ lint:
 
 test-node: test-bdd test-tdd test-qunit test-exports test-unit test-integration test-jsapi test-compilers test-requires test-reporters test-only test-global-only
 
-test-browser: clean mocha.js test-browser-unit test-browser-bdd test-browser-qunit test-browser-tdd test-browser-exports
+test-browser: clean mocha.js test-browser-unit test-browser-bdd test-browser-qunit test-browser-tdd test-browser-esm
 
 test: lint test-node test-browser
 
@@ -42,15 +42,19 @@ test-browser-unit:
 
 test-browser-bdd:
 	@printf "==> [Test :: Browser :: BDD]\n"
-	MOCHA_UI=bdd $(MAKE) test-browser-unit
+	MOCHA_TEST=bdd $(MAKE) test-browser-unit
 
 test-browser-qunit:
 	@printf "==> [Test :: Browser :: QUnit]\n"
-	MOCHA_UI=qunit $(MAKE) test-browser-unit
+	MOCHA_TEST=qunit $(MAKE) test-browser-unit
 
 test-browser-tdd:
 	@printf "==> [Test :: Browser :: TDD]\n"
-	MOCHA_UI=tdd $(MAKE) test-browser-unit
+	MOCHA_TEST=tdd $(MAKE) test-browser-unit
+
+test-browser-esm:
+	@printf "==> [Test :: Browser :: ESM]\n"
+	MOCHA_TEST=esm $(MAKE) test-browser-unit
 
 test-jsapi:
 	@printf "==> [Test :: JS API]\n"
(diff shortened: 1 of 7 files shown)


boneskull:
WIP pending some way of testing this so it doesn't happen again.

boneskull:
oh also this is necessary because `—require` in browserify doesn’t just “alias” stuff like the “browser” field of `package.json` which doesn’t work for any modules outside of your own sources anyway...

boneskull:
I started to wonder if this would be easier with Webpack then realized the build script I just wrote is likely smaller than even the most basic Webpack config 😬

Bamieh:
@boneskull I highly discourage using webpack in mocha, mainly because it overrides the native module implementation of node, which might lead to really odd bugs for some end users. It also limits some possible features of mocha, such as using node VMs for isolating the runner from the user code.

boneskull:
@Bamieh We don't bundle for Node.js, only for the browser.  So we don't need to worry about native modules getting blasted

Bamieh:
@boneskull okay, i will test this at my work tomorrow because we have the appropriate testing environment for this issue.

boneskull:
@Bamieh I've automated it.. please review

Makefile
```diff
@@ -1,14 +1,14 @@
-BROWSERIFY := "node_modules/.bin/browserify"
+BROWSERIFY := "scripts/build.js"
 KARMA := "node_modules/.bin/karma"
 MOCHA := "bin/mocha"
 NYC := "node_modules/.bin/nyc"
 
 ifdef COVERAGE
 define test_node
-	$(NYC) --no-clean --report-dir coverage/reports/$(1) $(MOCHA)
+$(NYC) --no-clean --report-dir coverage/reports/$(1) $(MOCHA)
```
boneskull: from what I can tell the leading tab here is unneccessary

Makefile
```diff
@@ -1,14 +1,14 @@
...
 endef
 else
-	test_node := $(MOCHA)
+test_node := $(MOCHA)
```
boneskull: and here

karma.conf.js
```diff
@@ -4,6 +4,9 @@ var fs = require('fs');
 var path = require('path');
 var mkdirp = require('mkdirp');
 var baseBundleDirpath = path.join(__dirname, '.karma');
+var builder = require('./scripts/build');
```
boneskull: this allows us to reuse the config in karma

karma.conf.js
```diff
@@ -32,23 +35,24 @@ module.exports = function (config) {
...
-        b.ignore('glob')
-          .ignore('fs')
-          .ignore('path')
-          .ignore('supports-color')
-          .require(path.join(__dirname, 'node_modules', 'buffer'), {expose: 'buffer'})
+        build(b)
           .on('bundled', function (err, content) {
-            if (!err && bundleDirpath) {
+            if (err) {
+              throw err;
```
boneskull: apparently we were eating errors before... oops.

karma.conf.js
```diff
@@ -32,23 +35,24 @@ module.exports = function (config) {
...
+            }
+            if (bundleDirpath) {
               // write bundle to directory for debugging
-              fs.writeFileSync(path.join(bundleDirpath,
-                'bundle.' + Date.now() + '.js'), content);
+              fs.writeFileSync(path.join(bundleDirpath, 'mocha.' + Date.now() +
```
boneskull: I changed this filename

karma.conf.js
```diff
@@ -124,21 +117,50 @@ module.exports = function (config) {
...
+        'test/interfaces/' + MOCHA_TEST + '.spec.js'
+      ];
+      break;
+
+    case 'esm':
+      // for now we will only run against Chrome to test this.
+      if (cfg.sauceLabs) {
+        cfg.sauceLabs.testName = 'ESM Integration Tests';
+        cfg.browsers = ['chrome@latest'];
+        var launcher = cfg.customLaunchers['chrome@latest'];
```
boneskull: sorry this is awkward; we need to update the dev scripts to ES2015+.

karma.conf.js
```diff
@@ -124,21 +117,50 @@ module.exports = function (config) {
...
+        cfg.customLaunchers = {
+          'chrome@latest': launcher
+        };
+      } else if (!env.TRAVIS) {
+        cfg.browsers = ['Chrome'];
```
boneskull: what other browsers currently support native modules? @mochajs/core

karma.conf.js
```diff
@@ -124,21 +117,50 @@ module.exports = function (config) {
...
+      } else {
+        console.error(
+          'skipping ESM tests & exiting; no SauceLabs nor local run detected');
+        process.exit(0);
+      }
+      cfg.files = [
+        'test/browser-fixtures/esm.fixture.html',
```
boneskull: can we force Karma to load a `.js` file as a module instead of script?  I have no idea.

test/browser-fixtures/esm.fixture.html
```diff
@@ -0,0 +1,7 @@
+<script>
+  delete window.require;
```
boneskull: this ensures that breakage will occur if the suffix on the distfile is bad.

boneskull: see #3139 for what I mean

boneskull:
You know, it might be a better idea to just land #3148 instead.

karma.conf.js
Bamieh: in accordance to MDN

> This feature is only just beginning to be implemented in browsers natively at this time. It is implemented in many transpilers, such as TypeScript and Babel, and bundlers such as Rollup and Webpack.

image

Bamieh:
@boneskull i am totally in favor of #3148 instead of these hacks. But we should keep the tests you wrote if we agree on dropping support for IE9, IE10.

boneskull:
need to pull my tests out of here

Issue #3128, opened 2017-12-08
stop using phantomjs to run browser tests


boneskull:
headless Chromium probably makes sense.  I'm not sure how far along headless FF is yet, but would prefer this over Chromium.

- [ ] remove karma-phantomjs-launcher
- [ ] add [karma-headless-chrome-launcher](https://npm.im/karma-headless-chrome-launcher)
- [ ] reconfigure `karma.conf.js` as appropriate

boneskull:
(the reason is phantomjs is no longer maintained afaik and probably fewer problems than a major browser in headless mode)

akrawchyk:
@boneskull @ScottFreeCode PR submitted let me know what you think

Issue #3149, opened 2017-12-13
Drop Node.js 4.x support


boneskull:
(Did I get your attention?)

Node.js v4.x [will no longer be maintained](https://github.com/nodejs/Release) as of April 2018.  When this happens, Mocha should follow suit and drop support.

Please note that old versions of Mocha will still work on unsupported platforms.

Bamieh:
in highlight of AWS lambda functions, AWS is already supporting node v4.3.2 and 6.10.3 (and probably more by the time we drop v4.x)

outsideris:
Now AWS lambda supports [v8.10, 4.3.2 and 6.10.3](https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html).

PR #3152, opened 2017-12-13
#3128 replace phantomjs with puppeteer for browser tests

akrawchyk:
### Description of the Change

Replaces PhantomJS with Puppeteer and updates Karma to use `ChromeHeadless` browser.

### Alternate Designs

The issue https://github.com/mochajs/mocha/issues/3128 describes using `karma-headless-chrome-launcher` for this.

However, the project already has `karma-chrome-launcher` as a dependency, see https://github.com/mochajs/mocha/pull/3152#issuecomment-353445511. ~~which has documentation for setting up `ChromeHeadless`, see https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer. I thought this would be preferred, so I configured Karma to use Puppeteer.~~

~~Also, during my testing the `karma-headless-chrome-launcher` was unable to locate Chromium. I'm not sure it's supported [by the library used to locate Chrome](https://github.com/CliftonH/karma-headless-chrome-launcher/blob/master/index.js#L6) for that plugin, see https://github.com/hughsk/chrome-location/issues/7.~~

`karma-chrome-launcher` does not locate Chromium by default either. Both Travis and Appveyor use Google Chrome, but you can use Chromium to run tests locally by setting the `CHROME_BIN` env var:

```
CHROME_BIN=$(which chromium-browser) make test-browser
```

https://github.com/mochajs/mocha/issues/3128#issuecomment-350177398

### Benefits

https://github.com/mochajs/mocha/issues/3128#issuecomment-350177398

### Possible Drawbacks

~~Puppeteer downloads a usable Chrome binary on install. This download is currently >95MB in size compared to PhantomJS <20MB, almost 5x bigger download. This is undesirable for obvious reasons.~~


akrawchyk:
Looks like [puppeteer requires Node >=v6](https://github.com/GoogleChrome/puppeteer#installation). If this is the desired path, it would have to wait on https://github.com/mochajs/mocha/issues/3149.

akrawchyk:
@boneskull thanks for the feedback, I fixed the commit message.

> Unfortunately, there's no such thing as an "optional dev dependency". This would be nice since we don't actually need it, since we only run the browser tests from Node.js v8.x.

Agreed, but it is nice to be able to depend on puppeteer's binary instead of adding dependencies to resolve the paths to Chrome.

> Would downgrading to a version that supports Node.js v4.x work? Happy to hear other suggestions as well.

I looked through the commit history for the project and I don't think this is viable since the project makes heavy use of ES6 features (like destructuring and default function params) that are not supported on older versions of Node, unfortunately.

I'm going to follow [this guide](https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai) and tinker some more with the Karma config to try and get this working without puppeteer as a dependency.

akrawchyk:
@boneskull I removed puppeteer since `karma-chrome-launcher` appears to work out of the box (see thread at https://github.com/karma-runner/karma/issues/2489 as well as the guide I linked above).

I updated the Travis config based on https://docs.travis-ci.com/user/chrome but I don't think we need to do the additional setup for Chrome headless here since Karma takes care of launching Chrome.

The Appveyor config doesn't need to be updated since Chrome is included in the build worker images (see https://www.appveyor.com/docs/build-environment/#test-runners).

akrawchyk:
@boneskull let me know your feedback once you get a chance to review the changes I pushed, thanks!

boneskull:
thanks

boneskull:
I'm going to pull this down and try it out; I'll address the conflict in `package-lock.json` as well.

boneskull:
awesome.  merged via e54370eaa4b16db96f14157fdce06272e2b4ec68

thanks @akrawchyk

PR #3156, opened 2017-12-13
Added a link to the walkthrough video

xxczaki:
Link: [HERE](https://youtu.be/zLayCLcIno0)

This will help future contributors understanding mocha's codebase :)


boneskull:
@xxczaki The build error was seemingly transient, but the code that is supposed to avoid it wasn't working, for whatever reason.  We can figure it out later.

For documentation-only changes, please add `[ci skip]` to your commit message so that the build isn't invoked--at least until we have some sort of CI for the documentation.

Issue #3127, opened 2017-12-08
drop node.js v7 from build matrix; add node.js v9


boneskull:
I think it's safe to do this now that npm v5.6.0 has been released.

xxczaki:
Yeah, if npm is working on Node.js 9.0 we should definitely add it, but i'm not sure about dropping Node.js 7.0 from build matrix...

boneskull:
node 7 is not maintained afaik. safe to drop

Bamieh:
closing this since the implementation is merged on master (fixed by  https://github.com/mochajs/mocha/pull/3185)

PR #3160, opened 2017-12-14
Added Node.js 9 support, removed Node.js 7

xxczaki:
Issue: #3127

I don't know if we should also add node.js 9 for the browser test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,10 @@ env:
 matrix:
   fast_finish: true
   include:
+    - node_js: '9'
+      env: TARGET=test-node COVERAGE=true
     - node_js: '8'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,10 +15,10 @@ env:
 matrix:
   fast_finish: true
   include:
+    - node_js: '9'
+      env: TARGET=test-node COVERAGE=true
     - node_js: '8'
       env: TARGET=test-node COVERAGE=true
```
boneskull: we only need to have `COVERAGE=true` on _one_ of these, so please remove it from line 21.

xxczaki:
I wasn't able to edit this PR, so i make a new one: https://github.com/mochajs/mocha/pull/3185

Issue #3111, opened 2017-11-20
Warn on incorrect Growl usage


ScottFreeCode:
### Description

In #3088 and #3067, Growl throws a not particularly helpful error if used without the OS-specific non-Node library.

We could consider having Mocha detect this issue and transform the error into something more helpful. For instance, it could be a warning at the start of the test run, before anything might otherwise be printed: "Growl mode activated in the options sent to Mocha, but Growl could not find the corresponding library; please see <link to installation instructions on the Growl repo> for assistance." Or instead of a warning it could be a hard error at whatever point Mocha first attempts to use Growl, without preemptively checking before the test run -- basically just catch the error from Growl, check if it's a file not found from the spawn command and if so throw the more helpful error, otherwise rethrow. In any case, besides CLI option (and/or `mocha.opts`), this ought to be useful when used through the programmatic API or with Grunt.

### Steps to Reproduce

Run with `--growl` on a machine that does not have Growl installed.

**Expected behavior:** A warning or error that tells me I need to install Growl or stop using the option.

**Actual behavior:** An error along the lines of "ENOENT libnotify".

**Reproduces how often:** 100%

### Versions
 * The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 4.x -- the version of Growl used by earlier versions of Mocha seems to have been bugged so as to suppress this error even though the readme specified the needed installation even back then

### Additional Information

Or since this is probably just biting people due to switching to Mocha 4, we could just field it till most people have transitioned and consider it handled. :man_shrugging: (And [if Grunt has defaults exasperating this then those should get fixed too](https://github.com/mochajs/mocha/issues/3088#issuecomment-345576861).)

xxczaki:
I will just transform this error into a warning message like "You need to have Growl installed in order to use this command" and we should also add an information about this in the [documentation](https://mochajs.org/) or maybe in the help message, for example:

`-G, --growl                             enable growl notification support. You need to have Growl installed!`

ematipico:
Can I take over this issue? It seems the old PR is closed, I could try to implement the snipped suggested inside the closed ticked

outsideris:
@ematipico Yes, you can.

PR #3161, opened 2017-12-14
Added more information about Growl, according to issue #3111

xxczaki:
Issue: #3111

I added the information about Growl in `bin/ _mocha`  and in the documentation to let users know, that they need to have growl installer in order to use this command correctly. We should also catch this error and add some form of message, but i don't really know where to start 😢 

@boneskull


boneskull:
it sounds like this won't get done; please reopen if I'm wrong

PR #3173, opened 2017-12-21
lint the markdown, closes #3172

### LINKED ISSUE #3172: lint the markdown
We have enough markdown in here that it should be linted via [markdownlint](https://npm.im/markdownlint).

boneskull:
add `@mocha/markdownlint-cli`; closes #3172

- forked [markdownlint-cli](https://npm.im/markdownlint-cli) to fix a glob-related bug; will abandon our version if/when it's merged
- add `.markdownlint.json` config
- remove "allow trailing spaces" for Markdown in `.editorconfig`
- add to `lint` target & create `markdownlint` script in `package.json`, which is dumb, but we're stuck with it for now
- linted the markdown

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,3 @@ indent_size = 2
 
 [Makefile]
 indent_style = tab
-
-[*.md]
-trim_trailing_whitespace = false
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,6 +1,7 @@
 # Mocha Code of Conduct
 
 ## Be friendly and patient
+
 We understand that everyone has different levels of experience or knowledge in many diverse fields, be it technical or
 non-technical in nature. We also have areas of knowledge we are eager to expand; we want to be a community where people
 can not only contribute, but feel comfortable to ask questions as well and learn along the way. If someone says something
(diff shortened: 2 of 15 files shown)


boneskull:
the worst was probably the changelog...

boneskull:
(netlify builds will fail until #3170 is merged)

boneskull:
@Bamieh Both `remark-cli` and `markdownlint-cli` seem to have about the same amount of adoption, so I picked the one which seemed more straightforward.  All we need is linting, after all.  Conversion of markdown to HTML is done with kramdown by way of [Jekyll](https://jekyllrb.com). 

Given that, why would Mocha need `remark` and its plugins?

Generally I'm in agreement on the line length, but I think there was a reason I didn't enforce this.  I just don't remember what it is...  I'll take another look and see if there are any speed bumps here.

boneskull:
I've rebased this onto `master` so it's synced with @Munter's changes.

@Munter This mainly looked OK, but I ended up doing two things:

1.  Replaced some HTML with Markdown in the `nav` tag.
2.  Reverted a change to a link you made (`https` ➡️ `http`) due to broken SSL on a site

boneskull:
...looks like one of my changes broke.

boneskull:
That Netlify preview is pretty awesome.

Bamieh:
@boneskull I suggested remark for mainly because you had to fork `markdownlint` to get it to work, so i suggested an alternative that i personally favor due to its awesome extendability. Given the current state of the PR, I believe there is no need for such change.

We can enforce the 80 line length rule later on as the project sees fit.

Awesome work! LGTM

boneskull:
`markdownlint-cli` accepted my PR and published, so we're using that instead of our fork.

PR #3177, opened 2017-12-24
Update README.md for better readability [ci skip]

xxczaki:
I think it is looking a bit better now 😄


Bamieh:
@xxczaki haha wow, i've just made the same update on a separate PR.

boneskull:
yeah, we also need to update the "plea for help", it's been up awhile.

Issue #2661, opened 2017-01-06
cannot pass reporterOptions to progress reporter


canoztokmak:
progress reporter reads values from `options`, but reporterOptions retrieved from args are passed in `options.reporterOptions` field.

submitted this pr (https://github.com/mochajs/mocha/pull/2649) for the solution.

boneskull:
closed via #2649

PR #3182, opened 2017-12-27
`options.reporterOptions` are used for progress reporter

canoztokmak:
Progress reporter had multiple options for configuring the reporter,
but it needed the optional fields under `options` object, while
those fields are provided under `options.reporterOptions`.
This commit enables usage of optional fields from `reporterOptions`.
Updated test cases as well.

Fix for #2661


boneskull:
@canoztokmak LGTM, thanks for fixing this.

PR #3185, opened 2017-12-31
Added Node.js 9 support, removed Node.js 7 

xxczaki:
I wasn't able to edit the old PR (https://github.com/mochajs/mocha/pull/3160), so i make this one. As @boneskull requested, i removed the coverage from Node.js 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ env:
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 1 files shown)


boneskull:
this isn't breaking; Mocha should still run on Node.js v7, but I don't want to be in the business of running CI against unmaintained versions of whatever; we've done that for too long

PR #3187, opened 2018-01-02
Ensure consistent calculation of duration

FND:
### Description of the Change

minor adjustment; probably doesn't make much of a difference, but seems more correct this way

### Alternate Designs

N/A

N/A

### Benefits

* minor performance enhancement
* slightly more consistent reporting

### Possible Drawbacks

none, I hope

### Applicable issues

N/A


boneskull:
yes, that seems more correct to me. thanks

PR #3189, opened 2018-01-03
Update license year.

sdbedi:
Update license year to -present format. Now no need to update it every year.

### Requirements

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 (The MIT License)
 
-Copyright (c) 2011-2017 JS Foundation and contributors, https://js.foundation
+Copyright (c) 2011-Present JS Foundation and contributors, https://js.foundation
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the


boneskull:
There's no legal requirement to update it every year anyway...

sdbedi:
It's my understanding that there is no formal legal requirement to have a Copyright Notice _at all_.

But if you are going to have a Copyright Notice-which this license does-best practice is to either (a) have the Notice contain just the initial year of publication (in this case 2011), or (b) a range of years, from the first date of publication to the latest changes to the work, ie. 2011-2018 or 2011-present.

We may be only a few days in to 2018, but there are already changes to the source code. I believe having a 2011-Present satisfies best practice for a copyright notice, while eliminating the need to update it in the future.

boneskull:
the task of updating the year in copyright notices is to show that "yes, this _has_ been updated recently".  so this PR would effectively _eliminate_ that benefit, since once it's merged, we no longer need to update it!  :wink:

I think we'll pass on this, but thank you!

Issue #3134, opened 2017-12-09
Document the `done()` contract


boneskull:
_From @FagnerMartinsBrack on February 22, 2016 6:57_

When I call `done()` passing a non `Error` instance (let's say an error Object Literal, for instance) I get the following message:

> done() invoked with non-Error: [Object object]

If the contract is that you should pass an `Error` instance to the `async` `done` handler, then that should be documented. I couldn't find that when looking at the site, only looking at the internals (which, as we all know, should not be relied upon).

If you pass a String (`done("it should not do this")`), according to the source, it is supposed to be casted accordingly, but then the result will be something like:

> done() invoked with non-Error: it should not do this

That is interpreted as a mistake, but then the docs should reflect what should be done to prevent that mistake.

Below is the relevant piece of code from source (2.4.5):

``` javascript
function callFnAsync(fn) {
  fn.call(ctx, function(err) {
    if (err instanceof Error || toString.call(err) === '[object Error]') {
      return done(err);
    }
    if (err) {
      if (Object.prototype.toString.call(err) === '[object Object]') {
        return done(new Error('done() invoked with non-Error: '
          + JSON.stringify(err)));
      }
     return done(new Error('done() invoked with non-Error: ' + err));
    }
    done();
  });
}
```

**TL DR;**

Document that `done` accepts either `undefined` or `an Error object` to reflect the current undocumented behavior.

_Copied from original issue: mochajs/old-mochajs-site#24_

k4113h:
Commenting here even though the issues is closed already: I think the docs are still unclear about the treating of Error instances:

> This callback accepts both an Error instance (or subclass thereof) or a falsy value; anything else will cause a failed test.

To me this sounds like passing an Error instance to done() will cause the test to _pass_, but this is obviously not the case.

PR #3192, opened 2018-01-03
[Document]: Error/undefined params to the 'done' callback

maraisr:
Resolves: #3134

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -155,7 +155,7 @@ Mocha allows you to use any assertion library you wish.  In the above example, w
 
 ## Asynchronous Code
 
-Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test. This callback accepts both an `Error` (or subclassed) instance, or `undefined`, any non-falsy will cause a failed test.
+Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test. This callback accepts both an `Error` instance (or subclass thereof) *or* a falsy value; anything else will cause a failed test.
 
 ```js
 describe('User', function() {
@@ -171,7 +171,7 @@ describe('User', function() {
 });
 ```
 
-To make things even easier, the `done()` callback also accepts an `Error` instance (ie, `new Error()`), so we may use this directly:
+To make things even easier, the `done()` callback also accepts an `Error` instance (i.e. `new Error()`), so we may use this directly:
 
 ```js
 describe('User', function() {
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -155,7 +155,7 @@ Mocha allows you to use any assertion library you wish.  In the above example, w
 
 ## Asynchronous Code
 
-Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test.
+Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test. This callback accepts both an `Error` instance, or `undefined`.
```
boneskull: #3134 isn't quite accurate about what needs to be documented.

We need to document that *any non-falsy, non-`Error` instance* passed into `done()` will cause Mocha to fail.

Subclasses of `Error` are allowed.

docs/index.md
```diff
@@ -184,6 +184,7 @@ describe('User', function() {
 });
 ```
 
+
```
boneskull: please remove

docs/index.md
maraisr: Can you check if that is accurate?

maraisr:
@boneskull do you think we can get this merged? Kinda excited to have committed something that was merged into Mocha.

docs/index.md
```diff
@@ -171,7 +171,7 @@ describe('User', function() {
 });
 ```
 
-To make things even easier, the `done()` callback accepts an error, so we may use this directly:
+To make things even easier, the `done()` callback also accepts an `Error` instance (ie, `new Error()`), so we may use this directly:
```
boneskull: Should be `i.e.` not `ie`

docs/index.md
```diff
@@ -155,7 +155,7 @@ Mocha allows you to use any assertion library you wish.  In the above example, w
 
 ## Asynchronous Code
 
-Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test.
+Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test. This callback accepts both an `Error` (or subclassed) instance, or `undefined`, any non-falsy will cause a failed test.
```
boneskull: Sorry, let's just eliminate `undefined` altogether here.  it's also a bit of a run-on sentence.  how about:

```markdown
This callback accepts both an `Error` instance (or subclass thereof) *or* a falsy value; anything else will cause a failed test.
```

boneskull:
thanks!

PR #3194, opened 2018-01-07
Rewrite Makefile using NPS Scripts; Closes #2352

### LINKED ISSUE #2352: replace GNU make with something else
[scripty](https://npmjs.com/package/scripty) is a thing and so is [p-s](https://npmjs.com/package/p-s).

I don't want the overhead of a full-blown task runner like Grunt or Gulp, as I don't want to rely on plugins.

Thoughts?

@kentcdodds Have you seen scripty?  p-s looks similar.  Maybe you guys should pool efforts? 

cc @searls
cc @mochajs/core

TedYav:
Addresses #2352

### Description of the Change
Where possible, I tried to exactly duplicate the design of the Makefile using `nps`. I've tested it by running `build` and `test` commands, as well as granular testing of individual commands against a vanilla repo to ensure it's running the same commands.

Important notes:
- removed the `==> [<ACTION> :: <SUBACTION>]` output from makefile as `nps` does this automatically
- did not recreate `make .PHONY` as this appeared to be a hack for GNU make. I'm not sure if some kind of file watching capability is desired. Perhaps this can be added back if needed.
- removed the dependency check (`SRC_FILES` variable) from Makefile as I didn't see a big reason to include this in `nps build` command
- code coverage can now be enabled by:
  - setting environment variable `COVERAGE` to any value
  - passing "--coverage" in command line to nps (though `nps` makes this inconvenient)
  - adding the `withCoverage` directive to the `nps` task list, i.e. `nps test.node.bdd withCoverage`
- `make` is now `nps build`
- any scripts which depend on `make` but were not in `package.json` will be broken 
- `nps` will need to be installed globally to work well, or at least be included in user's `$PATH`
- `npm run <xxx>` commands are now `npm start <xxx>` or `nps <xxx>`

### Alternate Designs

Possible modifications:
- Can use `npsUtils.concurrently.nps` to run tests in parallel. This has pros and cons.
- Can add documentation on how to use `nps` if needed
- I'm tempted to eliminate all the constants from the top for `NYC`, `MOCHA`, etc, though I imagine these may be useful in case a developer wants to temporarily direct these scripts to use a different installation

Discussed this briefly with @boneskull on Gitter back in early 2016. Didn't get around to doing it till now (my bad).

### Benefits

Makefile gone!

### Possible Drawbacks

Any code that depends on `make` commands will need to be updated.

### Applicable issues

This should not be a breaking change, unless there is some code I'm not aware of that depends on make commands (possibly some CI type operations)


boneskull:
@TedYav Looks like you'll need to sign or re-sign the CLA; maybe it couldn't link your account?

TedYav:
@boneskull ahh I pushed this from my work laptop and I think I had different email in my git settings, will resubmit with proper configuration. Also will update CI configurations

PR #3195, opened 2018-01-08
Rewrite Makefile using NPS Scripts. Closes #2352

### LINKED ISSUE #2352: replace GNU make with something else

TedYav:
Addresses #2352

Updated version of https://github.com/mochajs/mocha/pull/3194
Commit has correct Git credentials this time and Travis / Appveyor configs are now properly updated.

### Description of the Change
Where possible, I tried to exactly duplicate the design of the Makefile using `nps`. I've tested it by running `build` and `test` commands, as well as granular testing of individual commands against a vanilla repo to ensure it's running the same commands.

Important notes:
- removed the `==> [<ACTION> :: <SUBACTION>]` output from makefile as `nps` does this automatically
- did not recreate `make .PHONY` as this appeared to be a hack for GNU make. I'm not sure if some kind of file watching capability is desired. Perhaps this can be added back if needed.
- removed the dependency check (`SRC_FILES` variable) from Makefile as I didn't see a big reason to include this in `nps build` command
- code coverage is still enabled by setting `COVERAGE` environment variable to true. *I removed the ability to set this via a command line flag as it created issues when `nps` spawns child processes*
- `make` is now `nps build`
- any scripts which depend on `make` but were not in `package.json` will be broken 
- `npm run <xxx>` commands are now `npm start <xxx>` or `nps <xxx>`
- scripts can only be run as `nps <xxx>` if `nps` is in user's path. However, `npm start <xxx>` will work regardless.

### Alternate Designs

Possible modifications:
- Can use `npsUtils.concurrently.nps` to run tests in parallel. This has pros and cons.
- Can add documentation on how to use `nps` if needed
- I'm tempted to eliminate all the constants from the top for `NYC`, `MOCHA`, etc, though I imagine these may be useful in case a developer wants to temporarily direct these scripts to use a different installation

Discussed this briefly with @boneskull on Gitter back in early 2016. Didn't get around to doing it till now (my bad).

### Benefits

Makefile gone!

### Possible Drawbacks

Any code that depends on `make` commands will need to be updated.

### Applicable issues

This should not be a breaking change, unless there is some code I'm not aware of that depends on make commands (possibly some CI type operations)


TedYav:
Not sure where to configure `netlify` --> it is failing because `npm run buildDocs` is now `npm start buildDocs`.

TedYav:
Not sure why CI is failing.

Travis failed because a test timed out:

```
  1) file utils
       "before each" hook for "should accept a glob "path" value":
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

And appveyor appears to be having an issue executing Mocha from the CLI, possibly due to a difference in the way `make` versus `nps` runs commands on Windows. Perhaps I need to invert the `/` delimiter between `bin/mocha` or add `./` ?:

```
nps is executing `test.node.bdd` : bin/mocha --ui bdd test/interfaces/bdd.spec
'bin' is not recognized as an internal or external command,
operable program or batch file.
The script called "test.node.bdd" which runs "bin/mocha --ui bdd test/interfaces/bdd.spec" failed with exit code 1 https://github.com/kentcdodds/nps/blob/v5.7.1/other/ERRORS_AND_WARNINGS.md#failed-with-exit-code
```

Don't have Windows machine near to test.

boneskull:
you could try `node bin/mocha`

boneskull:
travis test may be flake.  you could try to increase the timeout to like 3000ms for that test

boneskull:
I'm not sure where the netlify configuration lives

boneskull:
OK, we're waiting on @Munter for information about the netlify failure, but since everything else is passing I can review this.  😝

Makefile
```diff
@@ -1,176 +0,0 @@
-BROWSERIFY := "node_modules/.bin/browserify"
```
boneskull: all of this red is the best part of the PR

package-scripts.js
```diff
@@ -0,0 +1,106 @@
...
+
+const MOCHA = `node ${path.join('bin', 'mocha')}`;
+const TM_BUNDLE = 'JavaScript\\ mocha.tmbundle';
+
+function test (testName, params) {
+  const prefix = process.env.COVERAGE ? `nyc --no-clean --report-dir coverage/reports/${testName}` : '';
+  return `${prefix} ${MOCHA} ${params}`;
+}
+
+module.exports = {
```
boneskull: I would absolutely love it if we added description fields to everything here

package-scripts.js
```diff
@@ -0,0 +1,106 @@
...
+  scripts: {
+    build: `browserify ./browser-entry --plugin ./scripts/dedefine --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' > mocha.js`,
+    lint: {
+      default: 'nps lint.code lint.markdown',
+      code: 'eslint . bin/*',
```
boneskull: even though we're not running this on AppVeyor, it should still work on windows.  please use double quotes around globs: `'eslint . "bin/*"'`

package-scripts.js
```diff
@@ -0,0 +1,106 @@
...
+        spec: test('non-tty-dot', '--reporter spec test/interfaces/bdd.spec 2>&1 > /tmp/spec.out && echo "spec:" && cat /tmp/spec.out')
+      }
+    },
+    tm: `open editors/${TM_BUNDLE}`,
+    prepublishOnly: 'nps test clean build',
+    coveralls: 'nyc report --reporter=text-lcov | coveralls',
+    prebuildDocs: 'rm -rf docs/_dist && node scripts/docs-update-toc.js',
+    buildDocs: 'nps prebuildDocs && bundle exec jekyll build --source ./docs --destination ./docs/_site --config ./docs/_config.yml --safe --drafts && nps postbuildDocs',
+    postbuildDocs: 'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
+    prewatchDocs: 'node scripts/docs-update-toc.js',
```
boneskull: I can't remember if `nps` supports `pre-` and `post-` scripts.  does it?  when we were running this with `npm`, `prewatchDocs` would get run before `watchDocs` when doing `npm run watchDocs`

package-scripts.js
```diff
@@ -0,0 +1,106 @@
...
+        },
+        globalOnly: {
+          default: 'nps test.node.globalOnly.bdd test.node.globalOnly.tdd test.node.globalOnly.qunit',
+          bdd: test('global-only-bdd', '--ui bdd test/only/global/bdd.spec'),
+          tdd: test('global-only-tdd', '--ui tdd test/only/global/tdd.spec'),
+          qunit: test('global-only-qunit', '--ui qunit test/only/global/qunit.spec')
+        }
+      },
+      browser: {
+        default: 'nps ' +
```
boneskull: would prefer one of a multiline template string or array w/ `join`

boneskull: likewise for other multiline strings in this file

package-scripts.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+const path = require('path');
+
+const MOCHA = `node ${path.join('bin', 'mocha')}`;
+const TM_BUNDLE = 'JavaScript\\ mocha.tmbundle';
+
+function test (testName, params) {
```
boneskull: please add a little comment explaining what this does

package-scripts.js
TedYav: Gotcha. If anyone has access to a Windows machine too it'd be awesome to do a dry run and be sure these commands work as expected.

package-scripts.js
TedYav: Kk — check over them in case anything is off. 👍

Munter:
@TedYav I've just merged a netlify.toml config file on `master`. If you rebase you should be able to adjust configuration of the build script on this branch

package-scripts.js
TedYav: Don't think it does, but the `pre` and `post` commands are explicitly inserted into the nps commands for that. I did move `prepublishOnly` back to `package.json` as I suspect it wouldn't trigger on `npm publish` if it was in `package-scripts.js`

TedYav:
I'm going to rebase and squash my commits. Hopefully Travis failure was a one off and we should be good to go. Let me know if anything else is desired (i.e. updating docs to say `npm start <xxx>` instead of `make <xxx>`

TedYav:
I noticed another commit dropped TextMate integration. I can eliminate the TM command to even this up. Also can merge master into this branch to fix conflicts if desired.

boneskull:
Thanks.  super awesome!

PR #3197, opened 2018-01-09
Added Netlify config file

Munter:
The Netlify build configuration currently only resides in their admin panel. This configuration file should hopefully override those settings to allow outside contributors to make refactoring changes

Refs https://github.com/mochajs/mocha/pull/3195

I'm letting this first build run with a non-valid configuration to see it fail. Then I'll switch it over to be in sync with the admin panel. After merge, current PR's can rebase and make changes to it

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,13 @@
+# Netlify deploy configuration
+# See https://www.netlify.com/docs/continuous-deployment/#deploy-contexts
+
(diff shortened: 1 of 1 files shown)


Munter:
Travis is having trouble connecting to headless chrome at the moment. However, the only change on this branch is for the docs deployment on netlify, so I'm going ahead and merging this so #3195 can become unblocked

PR #3201, opened 2018-01-13
Seperate diff generation code from Base reporter; closes #3011

### LINKED ISSUE #3011: 🚀 Feature: Expose public API for accessing diffs
### Prerequisites

### Description

Mocha prints nice diffs for failed assertions. It'd be great to show these diffs when running mocha tests in IDE.
Currently, IDE accesses `err` object directly to format error info: https://github.com/JetBrains/mocha-intellij/blob/3c40655fcb8e4b216586d19ab83cac8c3f98e4a9/lib/mochaIntellijReporter.js#L157-L176 and doesn't print diffs in console (only clicking "Click to see difference" opens a separate diff dialog).

What do you think about allowing IDE to show unified diffs for failed assertions? // cc @boneskull 

A possible API might look this:
```javascript
// lib/utils.js

exports.prettifyError = function (err) {
  return {
    msg: '<colored text with unified diff>',
    stack: '<colored tail of stack>'
  };
};
```
In this case, both `list` function from `lib/reporters/base.js` and IDE may use `prettifyError`.

Original issue in WebStorm's tracker: https://youtrack.jetbrains.com/issue/WEB-28905

HarrySarson:
### Description of the Change

Moves the code used to generate "nice diffs" out of `lib/reporters/base.js` and into a new file which I have called `lib/formatDiffs.js`. This allows access to "nice diffs" by IDE's.

To make this change I also had to move move code which deals with the coloring of mocha output out of `lib/reporters/base.js` as it is needed to generate the diffs. I put this into `lib/colorUtils.js`.

### Alternate Designs

This is a first step towards exposing a public API, I have left function definitions as unchanged as possible but they should be changed if this API is to be exposed.

### Benefits

Allows IDE to produce prettier output (#3011)

### Possible Drawbacks

Doing `Base.colors = { /* ... */ }` will no longer change the colors used by mocha as `Base.colors` is now a reference to the object defined in `lib/colorUtils.js`.

If this is a problem `Base.colors` could be defined using getters/setters or made readonly.

### Applicable issues
#3011 

This should be semver patch.

### DIFF
--- a/lib/colorUtils.js
+++ b/lib/colorUtils.js
@@ -1,0 +1,41 @@
+'use strict';
+
+exports.colors = {
+  pass: 90,
+  fail: 31,
+  'bright pass': 92,
+  'bright fail': 91,
+  'bright yellow': 93,
+  pending: 36,
+  suite: 0,
+  'error title': 0,
+  'error message': 31,
+  'error stack': 90,
+  checkmark: 32,
+  fast: 90,
+  medium: 33,
+  slow: 31,
+  green: 32,
+  light: 90,
+  'diff gutter': 90,
+  'diff added': 32,
+  'diff removed': 31
+};
+
+/**
+ * Color `str` with the given `type`.
+ *
+ * **Note:** `str` will only be coloured if `options.useColors === true`.
+ *
+ * @param {string} type
+ * @param {string} str
+ * @param {object} options
+ * @return {string}
+ * @api private
+ */
+exports.color = function color (type, str, options) {
+  if (!options.useColors) {
+    return String(str);
+  }
+  return '\u001b[' + exports.colors[type] + 'm' + str + '\u001b[0m';
+};
--- a/lib/formatDiffs.js
+++ b/lib/formatDiffs.js
@@ -1,0 +1,126 @@
+'use strict';
+
+var colorUtils = require('./colorUtils');
+var diff = require('diff');
+
+/**
+ * Pad the given `str` to `len`.
+ *
+ * @api private
+ * @param {string} str
+ * @param {string} len
+ * @return {string}
+ */
+function pad (str, len) {
+  str = String(str);
+  return Array(len - str.length + 1).join(' ') + str;
+}
+
+/**
+ * Returns an inline diff between 2 strings with coloured ANSI output
+ *
+ * @api private
+ * @param {Error} err with actual/expected
+ * @return {string} Diff
+ */
+exports.inlineDiff = function inlineDiff (err, options) {
+  var msg = errorDiff(err, options);
+
+  // linenos
+  var lines = msg.split('\n');
+  if (lines.length > 4) {
+    var width = String(lines.length).length;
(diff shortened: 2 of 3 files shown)


boneskull:
Thanks!  @HarrySarson Does this actually *close* #3011?  It looks like this is just part of the solution, right?

boneskull:
cc @segrey

HarrySarson:
@boneskull yeah sorry this is not a complete solution to #3011. It was my intention to ask for feedback and about the desired shape of the public API.

I think the exported functions `inlineDiff` and `unifiedDiff` (in fornatError.js) are what were requested by @segrey. These two functions generate coloured diffs based on the error. Would love feedback about whether this is alone the right lines to solve this use case.

boneskull:
@segrey I'm not sure this is really going to solve your issue, because Mocha is not the only place diffs come from.  Some assertion libraries provide their own diffs, appended to the `message` prop of whatever `Error` they give us.

We may overhaul all of this.  See chaijs/chai#1120 for more discussion.

I think we should probably sit on #3011 this unless @segrey is fine w/ supporting a temporary change.

segrey:
@HarrySarson Thanks! AFAIU, a reporter can use this API in the following way:
```javascript
var BaseReporter = requireMochaModule('./lib/reporters/base');
var formatDiffs = requireMochaModule('./lib/formatDiffs');

function extractErrorDetails(err) {
  var message = err.message, stack = err.stack;
  var index = stack.indexOf(message);
  if (index >= 0) {
    message = stack.slice(0, index + message.length);
    stack = stack.slice(message.length);
  }
  var diff;
  if (BaseReporter.inlineDiffs) {
    diff = formatDiffs.inlineDiff(err, {useColors: BaseReporter.useColors});
  } else {
    diff = formatDiffs.unifiedDiff(err, {useColors: BaseReporter.useColors});
  }
  return {
    message : message,
    diff: diff,
    stack : stack
  }
}
```

Since mocha already does something similar in Base reporter (it extracts message, diff and stack from `Error` object), it seems possible to expose its result as `err.details` with the following properties: `{message, diff, stack}`?
BTW, this `err.details` object can be expanded with `{actualStr, expectedStr}` properties (so IntelliJ reporter may no longer call `stringify` from `./lib/utils`).
Seems this API allows to leverage diffs provided by assertion libraries: mocha can simply set `err.details.diff` to a needed diff. @boneskull what do you think?

segrey:
Anyway, I'm fine with a quick temporal solution too. Seems it can be smaller, like this one:
```javascript
// lib/reporters/base.js
exports.unifiedDiff = unifiedDiff;
exports.inlineDiff = inlineDiff;
```

Then, IntelliJ reporter could use it in this way:
```javascript
  var diff;
  if (BaseReporter.inlineDiffs) {
    diff = BaseReporter.inlineDiff(err);
  } else {
    diff = BaseReporter.unifiedDiff(err);
  }
```

HarrySarson:
@segrey how does #3213 look?

Closing this.

PR #3202, opened 2018-01-14
fix id and class definition

DanielRuf:
### Requirements

### Description of the Change

This should add the right class and id definitions as they are currently missing.

### Benefits

Right image dimensions.

### Applicable issues

Docs

### DIFF
--- a/docs/_includes/backers.md
+++ b/docs/_includes/backers.md
@@ -3,5 +3,5 @@
 Find Mocha helpful?  Become a [backer](https://opencollective.com/mochajs#support) and support Mocha with a monthly donation.
 
 <!-- markdownlint-disable MD034 -->
-{: id="_backers" class="image-list" }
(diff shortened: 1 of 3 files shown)


DanielRuf:
> Thanks. I can't see any difference between mochajs.org and the preview. The image-list class still seems to be missing...

Mh, what is the needed markdown / kramdown markup here? I do not have `make` and would need a setup for Windows for further testing.

DanielRuf:
Information on the needed code is quite rare =/

DanielRuf:
I guess I fixed it.
Kramdown should be after the actual block / element / code.

DanielRuf:
Currently testing the ID which is still missing.

DanielRuf:
Seems to be fixed now. https://deploy-preview-3202--mocha.netlify.com/

Munter:
Nice catch. Thank you for the contribution.

Next time I need to do anything like that I'll just go pure HTML. At least I know what I'm doing there :D

PR #3210, opened 2018-01-18
#1577 Add "--exclude" Option

### LINKED ISSUE #1577: Option exclude certain files by pattern when testing recursively

metalex9:
### Description of the Change
Adds option `--exclude` to exclude matched files.
For example, this will match any files ending with "*.spec.js" **unless** they're inside the "node_modules" directory:
```bash
mocha "**/*.spec.js" --exclude "node_modules/**"
```
This required a small change to `bin/mocha.js`. As the file arguments are evaluated by mocha, they are checked against the pattern(s) provided by `--exclude`. Any files matching any of the patterns in `--exclude` are ignored.

I used [minimatch](https://github.com/isaacs/minimatch) to check the filenames against the patterns. This is the same library used by [glob](https://github.com/isaacs/node-glob).

As noted in #1577, this isn't strictly necessary. `mocha "./{,!(node_modules)/**/}*.spec.js"` would have the same effect as above. However, the `--exclude` option is easier for users who don't speak glob.

Situations like the above example are common. Other tools (like [Karma](https://github.com/karma-runner/karma)) include this behavior as core functionality.

### Benefits

This is easier for users to understand how to exclude specific files or patterns, giving them more granular control over which files are used for testing without having to use complex glob patterns.

### Possible Drawbacks

This is probably less performant than using glob patterns, and it's yet another option to support.
(If there is a more performant approach for implementing this behavior I'd be happy to change my code).

### Applicable issues
Resolves #1577.
Enhancement.


bin/_mocha
```diff
@@ -494,6 +496,13 @@ args.forEach(arg => {
     throw err;
   }
 
+  if (typeof newFiles !== 'undefined') {
+    if (typeof newFiles === 'string') {
+      newFiles = [newFiles];
+    }
+    newFiles = newFiles.filter(fileName => program.exclude.every(pattern => !minimatch(fileName, pattern)));
```
boneskull: you may be able to just chain this (`.filter` and everything after) to L489.

test/integration/fixtures/options/exclude/fail.spec.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
boneskull: fixture files should all have the extension `.fixture.js`

test/integration/options.spec.js
```diff
@@ -415,4 +415,41 @@ describe('options', function () {
...
     });
   });
+
+  describe('--exclude', function () {
+    /*
+     * Runs mocha in ./fixtures/options/exclude with the given args.
+     * Calls handleResult with the result
+     */
+    function runExcludeTest (args, handleResult, done) {
+      directInvoke(args, function (error, result) {
```
boneskull: unless there's a Good Reason to use `directInvoke()`, please use `runMocha()` instead

bin/_mocha
```diff
@@ -494,6 +496,13 @@ args.forEach(arg => {
     throw err;
   }
 
+  if (typeof newFiles !== 'undefined') {
```
boneskull: I don't think `newFiles` will be anything other than an `Array` of `string`s, will it?

metalex9: `newFiles` is the result of `utils.lookupFiles` which can return a string of arrays, a string, or `undefined`.

metalex9:
Changes made:
* [x] --exclude should allow itself to be specified multiple times instead of expecting a comma-delimited value
* [x] rename fixtures to have extension .fixture.js
* [x] use runMocha() instead of directInvoke() in options.spec.js

Regarding these two:
* [ ] ensure no unnecessary work is being done to check the type of newFiles on L500
* [ ] move the filter on L503 to chain on the call to utils.lookupFiles() on L489

`newFiles` is returned from `utils.lookupFiles`. If you look at the definition for `lookupFiles` in `lib/utils.js` L477, it looks to me like it could technically return an array of strings (L488, L523), a string (L495), or `undefined` (L499). Perhaps the documentation for that function needs to be updated?

Thanks for looking at this!

boneskull:
This looks good.  Not worried about conflicting with `--file`, as that doesn't accept a glob.

joaovieira:
@boneskull any idea on when `v5.1.0` is gonna be released? This option comes in handy :)

PR #3213, opened 2018-01-19
Exposes generation of diffs from base reporter

harrysarson:
#### Second attempt at solution for #3011, improves on PR #3201.

PR based on suggestion by @segrey in https://github.com/mochajs/mocha/pull/3201#issuecomment-358933824. Exposed function `generateDiff(expected, actual)` as a static member of `Base`.

The value of `Base.inlineDiffs` is used to determine the form of the diff.

### Alternate Designs

#3201

Allows IDE's to produce nice output when integrated with mocha.

### Benefits

See #3011.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -172,17 +172,14 @@ function stringifyDiffObjs (err) {
  * The diff will be either inline or unified dependant on the value
  * of `Base.inlineDiff`.
  *
- * @api private
- * @param {String} actual
- * @param {String} expected
+ * @param {string} actual
+ * @param {string} expected
  * @return {string} Diff
  */
 var generateDiff = exports.generateDiff = function (actual, expected) {
-  if (exports.inlineDiffs) {
-    return inlineDiff(actual, expected);
-  } else {
-    return unifiedDiff(actual, expected);
-  }
+  return exports.inlineDiffs
+    ? inlineDiff(actual, expected)
+    : unifiedDiff(actual, expected);
 };
 
 /**
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -167,6 +167,25 @@ function stringifyDiffObjs (err) {
 }
 
 /**
+ * Returns a diff between 2 strings with coloured ANSI output.
+ *
+ * The diff will be either inline or unified dependant on the value
+ * of `Base.inlineDiff`.
+ *
+ * @api private
+ * @param {String} actual
+ * @param {String} expected
+ * @return {string} Diff
+ */
+var generateDiff = exports.generateDiff = function (actual, expected) {
+  if (exports.inlineDiffs) {
+    return inlineDiff(actual, expected);
+  } else {
+    return unifiedDiff(actual, expected);
+  }
+};
+
+/**
  * Output the given `failures` as a list.
  *
  * @param {Array} failures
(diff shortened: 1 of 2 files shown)


segrey:
Looks good to me. @HarrySarson thanks!

segrey:
I'll update IntelliJ reporter once the PR is merged.

lib/reporters/base.js
```diff
@@ -166,6 +166,25 @@ function stringifyDiffObjs (err) {
...
+ *
+ * @api private
+ * @param {String} actual
+ * @param {String} expected
+ * @return {string} Diff
+ */
+var generateDiff = exports.generateDiff = function (actual, expected) {
+  if (exports.inlineDiffs) {
+    return inlineDiff(actual, expected);
+  } else {
```
boneskull: this `else` is not necessary

lib/reporters/base.js
```diff
@@ -166,6 +166,25 @@ function stringifyDiffObjs (err) {
   }
 }
 
+/**
+ * Returns a diff between 2 strings with coloured ANSI output.
+ *
+ * The diff will be either inline or unified dependant on the value
+ * of `Base.inlineDiff`.
+ *
+ * @api private
```
boneskull: I think the intent is that this is a public function, so you can remove this, right?

lib/reporters/base.js
```diff
@@ -166,6 +166,25 @@ function stringifyDiffObjs (err) {
...
+ * @param {String} actual
```
boneskull: lowercase `string`

test/reporters/base.spec.js
```diff
@@ -136,6 +136,36 @@ describe('Base reporter', function () {
...
 
+  describe('Diff generation', function () {
+    it('should generate unified diffs if `inlineDiff === false`', function () {
+      var actual = 'a foo unified diff';
+      var expected = 'a bar unified diff';
+
+      var oldInlineDiffs = Base.inlineDiffs;
+      Base.inlineDiffs = false;
+
+      var output = Base.generateDiff(actual, expected);
```
boneskull: please do the patching in a `beforeEach` of the suite and restore in `afterEach`

boneskull:
@HarrySarson Thanks for this!! in the future, please just reuse your PRs and force-push changes--it's kind of difficult to track the conversation otherwise

segrey:
@boneskull Regarding `err.details`, I meant to enhance standard `err` object that is passed to reporters in `runner.on('fail', function (test, err) {})`. The idea is to have some common properties that are calculated by mocha already and might be interested for reporters:
* `message` and `stack`: calculated using logic similar to https://github.com/mochajs/mocha/blob/v5.0.1/lib/reporters/base.js#L185-L205
* `diff`: the result of `generateDiff(err.actual, err.expected)`.

As a benefit, mocha will have a mechanism to easily change how these values are calculated, e.g.  supply diffs provided by assertion libraries without changes in reporters.

test/reporters/base.spec.js
HarrySarson: When doing the patching, should I be setting the value of `Base.inlineDiffs` in the `beforeEach`?

HarrySarson:
Test failure is due to linting of markdown. Would be resolved by #3236

harrysarson:
@boneskull can this be merged?

boneskull:
@segrey Got it.  So it looks like this PR is the first step to get there.

boneskull:
blah, meant to squash.

PR #3214, opened 2018-01-19
extract checking AMD bundle as own test command

### LINKED ISSUE #3208: travis "before script" should only be run once
This script basically asserts `define` doesn't appear in the browser distfile (`mocha.js`).

- This check probably needs to have its own line in the build matrix
- It should be invoked via a script in `package-scripts.js` so developers can run it locally

outsideris:
### Description of the Change
As #3208 , I moved checking AMD bundle into `package-script.js`.
I'm not sure the name which is`test.browser.amd` is properly.

### Alternate Designs
I considered between `nps lint.amd` and `npm test.browser.amd`.
I thought it is a kind of testing rather than linting.

See #3208

### Benefits
Developers can run it locally.

### Possible Drawbacks
N/A

### Applicable issues
Fix #3208

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -40,7 +40,7 @@ module.exports = {
     test: {
       default: 'nps test.all',
       all: {
-        script: 'nps lint.code test.node test.browser',
+        script: 'nps lint.code test.node test.browser test.bundle',
         description: 'Lint code and runs node / browser environment tests'
       },
       node: {
@@ -204,6 +204,17 @@ module.exports = {
         spec: {
           script: test('non-tty-dot', '--reporter spec test/interfaces/bdd.spec 2>&1 > /tmp/spec.out && echo "spec:" && cat /tmp/spec.out'),
           description: 'Test non-TTY spec reporter'
+        }
+      },
+      bundle: {
+        default: 'nps test.bundle.all',
+        all: {
+          script: 'nps clean build.mochajs test.bundle.amd',
+          description: 'Compile Mocha and run all tests for bundle files'
+        },
+        amd: {
+          script: test('amd', 'test/bundle/amd.spec'),
+          description: 'Test bundle files for AMD'
         }
       }
     },
--- a/scripts/travis-before-script.sh
+++ b/scripts/travis-before-script.sh
@@ -2,6 +2,3 @@
 
 # bundle artifacts to AWS go here
 mkdir -p .karma
-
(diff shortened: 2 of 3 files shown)


package-scripts.js
```diff
@@ -185,6 +185,10 @@ module.exports = {
         esm: {
           script: 'MOCHA_TEST=esm nps test.browser.unit',
           description: 'Test mocha ESM support'
+        },
+        amd: {
+          script: 'if grep -q "define.amd" mocha.js; then exit 1; fi',
```
boneskull: this will likely fail in windows.  so maybe better to write an actual test (put in`test/` somewhere) which:

1. triggers `npm start build` to ensure `mocha.js` is present
1. reads `mocha.js` into a string via `fs.readFile()`
1. asserts `"define.amd"` is not present in the string

what do you think?

outsideris: I got what you mean. It makes sense and looks like more maintainable.
I will try to fix it.

outsideris:
@boneskull Could you check this?

package-scripts.js
```diff
@@ -185,6 +185,10 @@ module.exports = {
         esm: {
           script: 'MOCHA_TEST=esm nps test.browser.unit',
           description: 'Test mocha ESM support'
+        },
+        bundle: {
```
boneskull: this isn't really a _browser_ test because we're not invoking Karma.  maybe should just move it to be its own thing, e.g. `test.bundle`. also, there's a `test()` function to help.

test/browser-specific/bundle.spec.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+var path = require('path');
+var fs = require('fs');
+
+it('should build a non-broken bundle for AMD', function (done) {
+  var bundle = path.join(process.cwd(), 'mocha.js');
+  fs.readFile(bundle, function (err, data) {
```
boneskull: specify utf8 encoding to get a string

test/browser-specific/bundle.spec.js
```diff
@@ -0,0 +1,16 @@
...
+    if (err) { return done(err); }
+
+    var content = data.toString();
```
boneskull: it might be more straightforward to use `expect(content).not.to.match(/define.amd/)` or something like this

outsideris:
I moved the test under `test.bundle.amd` and fixed it as feedbacks.

PR #3217, opened 2018-01-23
update docs for the glob

### LINKED ISSUE #3098: File globbing not documented on site
How to configure it? This for example
> By default, mocha looks for the glob ./test/*.js and ./test/*.coffee, so you may want to put your tests in test/ folder.

No any info in whole [docs](http://mochajs.org/) and wiki

### LINKED ISSUE #3132: Update comment about default glob
_From @ScottFreeCode on October 4, 2017 3:52_

Here: https://mochajs.org/#the-test-directory

Currently says:

> By default, `mocha` looks for the glob `./test/*.js` and `./test/*.coffee`, so you may want to put your tests in `test/` folder.

This is inaccurate in one big way and one small way.

Big way: `.coffee` no longer used by default.

Small way: technically, the default glob is `test`; when a directory is specified Mocha looks in it for `*.js` files (or `*.{js,<other extensions from --compilers}`) *or for `**/*.js` files when `--recursive` is specified.* This is really only important if you need to understand adding `*.{<extensions>}` when replacing `--compilers` or if you need to understand `--recursive`, but we do occasionally get people confused about the interaction between globbing and `--recursive`...

_Copied from original issue: mochajs/old-mochajs-site#82_

outsideris:
### Requirements

### Description of the Change
`.coffee` doesn't supported now. And I enhance the explanation of glob.

Please check my English sentences.

### Alternate Designs
N/A

`.coffee` doesn't supported now.

### Benefits
Users don't confuse anymore.

### Possible Drawbacks
N/A

### Applicable issues
Close #3132
Close #3098

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1272,7 +1272,7 @@ $ mocha --reporter list --growl
 
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js` and `./test/*.coffee`, so you may want to put your tests in `test/` folder.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, because of `./test/*.js` is only matched with files which in the first depth of `test` directory and `./test/**/*.js` is only matched with files which in the second depth of `test` directory.
 
 ## Editor Plugins
 


Munter:
Love it! Thank you for contributing!

docs/index.md
```diff
@@ -1272,7 +1272,7 @@ $ mocha --reporter list --growl
 
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js` and `./test/*.coffee`, so you may want to put your tests in `test/` folder.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, because of `./test/*.js` is only matched with files which in the first depth of `test` directory and `./test/**/*.js` is only matched with files which in the second depth of `test` directory.
```
HarrySarson: > ./test/**/*.js` is only matched with files which in the second depth of `test` directory

Is this true? I though `'/**/'` matches one or more directory level.

HarrySarson: From https://www.npmjs.com/package/glob

> ** If a "globstar" is alone in a path portion, then it matches zero or more directories and subdirectories searching for matches. It does not crawl symlinked directories.

Munter: This depends on if you pass a glob pattern as a string or use shell expansion. Shell expansion will only expand to the second level, and actually **exactly** the second level, not the first one, so that is also a bit confusing to many

PR #3225, opened 2018-01-29
Issue/2819 - Fix / Clarify this.skip behavior in before hooks

bannmoore:
This PR contains a possible fix for [Issue 2819](https://github.com/mochajs/mocha/issues/2819).

### Description of the Change

Let me preface this by saying I'm new to open source contributing, so if I messed something up please let me know. I assure you in advance that it was an unintentional mistake rather than deliberate violation of the rules. :)

I adapted this change from [Elergy's PR](https://github.com/mochajs/mocha/pull/2836/files), since it seemed to be stuck in a limbo. I also referred to [this discussion](https://github.com/mochajs/mocha/pull/2571) regarding skip and hooks. It seems like all the discussions have been inactive for a while, so I'm interested to know if there's been a resolution.

Changes:
- Added integration tests for `this.skip` in `before` that verify hooks don't execute and nested describes are skipped.
  - Modified existing `this.skip` / `before` tests to show that subsequent hooks aren't executed.
- Modified `runner.js` to set this.suite.pending instead of just the tests.
- Added a proposed doc change for the interaction of this.skip and nested describes (I'm fine dropping this commit if needed, but it'd be nice if the docs spoke to this situation).

This PR's primary goal is to get a consensus on what the expected behavior is regarding `this.skip` in non-trivial cases involving hooks and describe. At the very least, it'd be helpful for the documentation / tests to address these cases one way or the other. If this fix works, so much the better.

Here are the behaviors I'm concerned about (my first commit contains the failing tests). I've lumped the two issues together since they seem to stem from the same root cause; namely, that the suite itself is not marked as pending when skipped in a before.
- When `this.skip` is called in a `before` block, subsequent before blocks are still executed.
  - Expected: Subsequent hooks are not executed if the suite is skipped.
- When `this.skip` is called in a `before` block, tests in nested describes are not skipped (issue 2819).
  - Expected: All tests in a block are skipped, even if nested.

### Alternate Designs

If these behaviors are NOT bugs, then perhaps we can make that clear in the docs.

This issue speaks to the behavior of this.skip, which is in core.

### Benefits

At a minimum, I'd like to reach a point where the documentation addresses these cases. This will help address potential user confusion in the future. I'm happy to collaborate on the documentation itself. :)

### Possible Drawbacks

I'm not sure, but I'm hoping someone else has feedback.

### Applicable issues

[Issue 2819](https://github.com/mochajs/mocha/issues/2819)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -317,6 +317,9 @@ Runner.prototype.hook = function(name, fn) {
             suite.tests.forEach(function(test) {
               test.pending = true;
             });
+            suite.suites.forEach(function(suite) {
+              suite.pending = true;
+            });
             // a pending hook won't be executed twice.
             hook.pending = true;
           }
--- a/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
@@ -20,15 +20,15 @@ describe('outer suite', function () {
     });
 
     beforeEach(function () {
-      throw new Error('never thrown');
+      throw new Error('beforeEach should not run');
     });
 
     afterEach(function () {
-      throw new Error('never thrown');
+      throw new Error('afterEach should not run');
     });
 
     it('should not run this test', function () {
-      throw new Error('never thrown');
+      throw new Error('inner suite test should not run');
     });
 
     after(function () {
@@ -41,7 +41,7 @@ describe('outer suite', function () {
       });
 
       it('should not run this test', function () {
-        throw new Error('never thrown');
+        throw new Error('skipped suite test should not run');
       });
 
       after(function () {
--- a/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
@@ -1,0 +1,60 @@
+'use string';
+
+describe('outer suite', function () {
+
+  before(function () {
+    console.log('outer before');
+  });
+
+  it('should run this test', function () {});
+
+  describe('inner suite', function () {
+
+    before(function (done) {
+      console.log('inner before');
+      var self = this;
+      setTimeout(function () {
+        self.skip();
+        done();
+      });
+    });
+
+    beforeEach(function () {
+      throw new Error('never thrown');
+    });
+
+    afterEach(function () {
+      throw new Error('never thrown');
+    });
+
+    it('should not run this test', function () {
+      throw new Error('never thrown');
+    });
+
+    after(function () {
+      console.log('inner after');
+    });
+
+    describe('skipped suite', function () {
+      before(function () {
+        console.log('skipped before');
+      });
+
+      it('should not run this test', function () {
+        throw new Error('never thrown');
+      });
+
+      after(function () {
+        console.log('skipped after');
+      });
+    });
+
+  });
+
+  it('should run this test', function () {});
+
+  after(function () {
+    console.log('outer after');
+  });
+
+});
--- a/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
@@ -16,7 +16,7 @@ describe('outer suite', function () {
       setTimeout(function () {
         self.skip();
         done();
-      });
+      }, 0);
     });
 
     beforeEach(function () {
--- a/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
@@ -1,4 +1,4 @@
-'use string';
+'use strict';
 
 describe('outer suite', function () {
 
@@ -6,7 +6,7 @@ describe('outer suite', function () {
     console.log('outer before');
   });
 
-  it('should run this test', function () {});
+  it('should run this test', function () { });
 
   describe('inner suite', function () {
 
@@ -51,7 +51,7 @@ describe('outer suite', function () {
 
   });
 
-  it('should run this test', function () {});
+  it('should run this test', function () { });
 
   after(function () {
     console.log('outer after');
--- a/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
@@ -1,0 +1,35 @@
+describe('skip in before with nested describes', function () {
+  before(function (done) {
+    var self = this;
+    setTimeout(function () {
+      self.skip();
+      done();
+    }, 50);
+  });
+
+  it('should never run this test', function () {});
+
+  describe('nested describe', function () {
+    before(function () {
+      throw new Error('should not run');
+    });
+
+    it('should never run this test', function () {});
+
+    after(function () {
+      throw new Error('should not run');
+    });
+
+    describe('nested again', function () {
+      before(function () {
+        throw new Error('should not run');
+      });
+
+      it('should never run this test', function () {});
+
+      after(function () {
+        throw new Error('should not run');
+      });
+    });
+  });
+});
--- a/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
@@ -4,7 +4,7 @@ describe('skip in before with nested describes', function () {
     setTimeout(function () {
       self.skip();
       done();
-    }, 50);
+    }, 0);
   });
 
   it('should never run this test', function () { });
--- a/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
@@ -7,28 +7,28 @@ describe('skip in before with nested describes', function () {
     }, 50);
   });
 
-  it('should never run this test', function () {});
+  it('should never run this test', function () { });
 
   describe('nested describe', function () {
     before(function () {
-      throw new Error('should not run');
+      throw new Error('first level before should not run');
     });
 
-    it('should never run this test', function () {});
+    it('should never run this test', function () { });
 
     after(function () {
-      throw new Error('should not run');
+      throw new Error('first level after should not run');
     });
 
     describe('nested again', function () {
       before(function () {
-        throw new Error('should not run');
+        throw new Error('second level before should not run');
       });
 
-      it('should never run this test', function () {});
+      it('should never run this test', function () { });
 
       after(function () {
-        throw new Error('should not run');
+        throw new Error('second level after should not run');
       });
     });
   });
--- a/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
@@ -7,14 +7,18 @@ describe('skip in before with nested describes', function () {
     }, 0);
   });
 
-  it('should never run this test', function () { });
+  it('should never run this test', function () {
+    throw new Error('never run this test');
+  });
 
   describe('nested describe', function () {
     before(function () {
       throw new Error('first level before should not run');
     });
 
-    it('should never run this test', function () { });
+    it('should never run this test', function () {
+      throw new Error('never run this test');
+    });
 
     after(function () {
       throw new Error('first level after should not run');
@@ -25,7 +29,9 @@ describe('skip in before with nested describes', function () {
         throw new Error('second level before should not run');
       });
 
-      it('should never run this test', function () { });
+      it('should never run this test', function () {
+        throw new Error('never run this test');
+      });
 
       after(function () {
         throw new Error('second level after should not run');
--- a/test/integration/fixtures/pending/skip-async-before.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before.fixture.js
@@ -1,18 +1,21 @@
 'use strict';
 
-describe('skip in before', function () {
-  before(function (done) {
-    var self = this;
-    setTimeout(function () {
-      self.skip();
-    }, 50);
+describe('outer describe', function () {
+
+  it('should run this test', function () {});
+
+  describe('skip in before', function () {
+    before(function (done) {
+      var self = this;
+      setTimeout(function () {
+        self.skip();
+      }, 50);
+    });
+
+    it('should never run this test', function () {});
+    it('should never run this test', function () {});
   });
 
-  it('should never run this test', function () {
-    throw new Error('never thrown');
-  });
+  it('should run this test', function () {});
 
-  it('should never run this test', function () {
-    throw new Error('never thrown');
-  });
 });
--- a/test/integration/fixtures/pending/skip-async-before.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before.fixture.js
@@ -1,7 +1,6 @@
 'use strict';
 
 describe('outer describe', function () {
-
   it('should run this test', function () {});
 
   describe('skip in before', function () {
@@ -12,10 +11,13 @@ describe('outer describe', function () {
       }, 0);
     });
 
-    it('should never run this test', function () {});
-    it('should never run this test', function () {});
+    it('should never run this test', function () {
+      throw new Error('never run this test');
+    });
+    it('should never run this test', function () {
+      throw new Error('never run this test');
+    });
   });
 
   it('should run this test', function () {});
-
 });
--- a/test/integration/fixtures/pending/skip-async-before.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before.fixture.js
@@ -9,7 +9,7 @@ describe('outer describe', function () {
       var self = this;
       setTimeout(function () {
         self.skip();
-      }, 50);
+      }, 0);
     });
 
     it('should never run this test', function () {});
--- a/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
@@ -5,14 +5,11 @@ describe('skip in beforeEach', function () {
     var self = this;
     setTimeout(function () {
       self.skip();
+      done();
     }, 50);
   });
 
-  it('should never run this test', function () {
-    throw new Error('never thrown');
-  });
-
-  it('should never run this test', function () {
-    throw new Error('never thrown');
-  });
+  it('should skip this test', function () {});
+  it('should error before reaching this test', function () {});
+  it('should error before reaching this test', function () {});
 });
--- a/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
@@ -6,7 +6,7 @@ describe('skip in beforeEach', function () {
     setTimeout(function () {
       self.skip();
       done();
-    }, 50);
+    }, 0);
   });
 
   it('should skip this test', function () {});
--- a/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
@@ -9,7 +9,13 @@ describe('skip in beforeEach', function () {
     }, 0);
   });
 
-  it('should skip this test', function () {});
-  it('should error before reaching this test', function () {});
-  it('should error before reaching this test', function () {});
+  it('should skip this test', function () {
+    throw new Error('never run this test');
+  });
+  it('should not reach this test', function () {
+    throw new Error('never run this test');
+  });
+  it('should not reach this test', function () {
+    throw new Error('never run this test');
+  });
 });
--- a/test/integration/fixtures/pending/skip-async-spec.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-spec.fixture.js
@@ -5,7 +5,7 @@ describe('skip in test', function () {
     var self = this;
     setTimeout(function () {
       self.skip();
-    }, 50);
+    }, 0);
   });
 
   it('should run other tests in the suite', function () {
--- a/test/integration/fixtures/pending/skip-async-spec.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-spec.fixture.js
@@ -8,7 +8,5 @@ describe('skip in test', function () {
     }, 0);
   });
 
-  it('should run other tests in the suite', function () {
-    // Do nothing
-  });
+  it('should run other tests in the suite', function () {});
 });
--- a/test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
@@ -39,7 +39,7 @@ describe('outer suite', function () {
       });
 
       it('should never run this test', function () {
-        throw new Error('skipped suie test should not run');
+        throw new Error('skipped suite test should not run');
       });
 
       after(function () {
--- a/test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
@@ -1,0 +1,57 @@
+'use strict';
+
+describe('outer suite', function () {
+
+  before(function () {
+    console.log('outer before');
+  });
+
+  it('should run this test', function () {});
+
+  describe('inner suite', function () {
+    before(function () {
+      this.skip();
+    });
+
+    before(function () {
+      console.log('inner before');
+    });
+
+    beforeEach(function () {
+      throw new Error('never thrown');
+    });
+
+    afterEach(function () {
+      throw new Error('never thrown');
+    });
+
+    after(function () {
+      console.log('inner after');
+    });
+
+    it('should never run this test', function () {
+      throw new Error('never thrown');
+    });
+
+    describe('skipped suite', function () {
+      before(function () {
+        console.log('skipped before');
+      });
+
+      it('should never run this test', function () {
+        throw new Error('never thrown');
+      });
+
+      after(function () {
+        console.log('skipped after');
+      });
+    });
+  });
+
+  it('should run this test', function () {});
+
+  after(function () {
+    console.log('outer after');
+  })
+
+});
--- a/test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
@@ -6,7 +6,7 @@ describe('outer suite', function () {
     console.log('outer before');
   });
 
-  it('should run this test', function () {});
+  it('should run this test', function () { });
 
   describe('inner suite', function () {
     before(function () {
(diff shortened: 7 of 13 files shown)


boneskull:
@brittanymoore Thanks for this, and welcome.  PRs and issues--usually of the compex variety--tend to linger around here, due to lack of time to address them (we could always use more help!).

To your points:

> When this.skip is called in a before block, subsequent before blocks are still executed.
> - Expected: Subsequent hooks are not executed if the suite is skipped.

I don't think this is right (though I may be disagreeing with my past self).  A `this.skip()` in a `before`  isn't really "skipping a suite".  This is how I think it should work:

Hooks are for setting up and tearing down.  If you set something up, you must tear it down.  The problem is that we don't know which "tear-down" hooks (`after`, `afterEach`) correspond to which "set-up" hooks (`before`, `beforeEach`), *or if they correspond at all.*  We can't even be sure that whatever happened in a `before` will be torn down by an `after` in the *same* suite.

This means that if a `before` has `this.skip()` in it, we have to run any `after` hooks to ensure the tear-down occurs.  But if we are running **any** `after` hooks, we have to run **all** `before` hooks, because we don't know which `before` any given `after` is tearing down.

It could be argued that we should continue into nested suites, and execute *those* `before`/`after` hooks, because a nested `after` can tear down what a parent `before` did.  The following is contrived--*and an antipattern*--but maybe it should work:

```js
  let stubA;

  before(function () {
    stubA = stubThingA();
    if (reasons()) {
      this.skip();
    }
  });

  describe('something else', function () {
    let stubB;

    before(function () {
      stubB = stubThingB();
    });

    after(function () {
      // this nested `after` does *all* of the teardown.
      stubA.restore();
      stubB.restore();
    });
  });

  // what the author should do instead is create an `after`
  // hook here, which calls `stubA.restore()`.
});
```

I don't really want to encourage stuff like the above, so probably best to leave it out until we see a reasonable use-case.

However, the `this.skip()` in a `before` *does* mean that no tests will be executed within the suite.  If no tests are executed, then no `beforeEach` / `afterEach` hooks should be executed either.

Feel free to disagree with the above!

> When this.skip is called in a before block, tests in nested describes are not skipped (issue 2819).
> - Expected: All tests in a block are skipped, even if nested.

I agree with this.  All tests should be skipped, even in nested suites.

You may also find it useful to know that `this.skip()` inside a *suite itself* is valid--that skips the entire suite's contents.

I'll wait on reviewing your PR until I hear back--not sure if what I'm saying will impact it.  Thanks!

brittanymoore:
@boneskull Thank you for the detailed response! I hope you didn't interpret the "limbo" comment above as criticism - I completely understand that much of open source is done by volunteers, and I appreciate all the work you do.

I submitted a new commit based on your feedback.

Your reasoning for running same-level `before` / `after` hooks make sense. If `this.skip` is called from a `before`, we have no way of knowing that teardown isn't necessary, so we run both. Personally, I can live with that. 👍 

However, I do feel like we could safely skip _nested_ hooks. Partly to dissuade folks from the anti pattern you posted, and partly because they aren't subject to the ambiguous setup / teardown situation above.

So, here's where I'm at, when `this.skip()` is called in `before`:
- run any additional `before` / `after` hooks at the same level
- skip any `beforeEach` / `afterEach` hooks at the same level
- skip all tests at the same level or in nested suites
- skip all hooks in nested suites

lib/runner.js
```diff
@@ -314,6 +314,7 @@ Runner.prototype.hook = function (name, fn) {
             suite.tests.forEach(function (test) {
               test.pending = true;
             });
+            suite.parent.pending = true;
```
leedm777: I believe this will still skip the current suite's before/after hooks. It also might skip further tests and after hooks of the parent suite, but I'm not totally sure of that. 

You could try `suite.suites.forEach(function (suite) { suite.pending = true; })` instead.

bannmoore: @leedm777 Thanks for the suggestion! I reworked my tests and realized that you are correct, any `after` hooks placed under the skipped `describe` were not executing. The only problem I'm having with your solution is when `this.skip` is called asynchronously. I updated my commit with the latest work.

From what I can tell, the asynchronous skip is aborting the test run, giving me outputs like this (there should be another skipped test in `inner inner`:

```
  outer
    ✓ should run
    inner
      - should not run
      inner inner


  1 passing (2s)
  1 pending
``` 

I feel like I'm missing something, so if someone with more experience in the code base has any insight I'd be glad to hear it.

test/integration/pending.spec.js
```diff
@@ -119,7 +185,7 @@ describe('pending', function () {
 
     describe('in beforeEach', function () {
       it('should skip all suite specs', function (done) {
-        run('pending/skip-sync-beforeEach.fixture.js', args, function (err, res) {
```
bannmoore: This test wasn't calling the `async` fixture, and it started failing when I fixed it. This is unrelated to my changes, as far as I can tell (see [this branch](https://github.com/bannmoore/mocha/tree/pending-async-beforeEach).

bannmoore:
@boneskull I did some clean up and all the tests are passing, with one execption.

It looks like there was a typo in pending.spec.js that was running the sync version of the beforeEach fixture instead of async. The test is failing (with or without my change); what behavior do we expect there?

HarrySarson:
The linting error has now been fixed in master so next time Travis runs it should pass the linting job.

michaelbeauchemin:
Just curious as to what's holding this up from being merged? I could use this functionality in a test suite I'm currently working on.

boneskull:
@michaelbeauchemin It hasn't been reviewed yet.

boneskull:
@bannmoore That sounds OK to me, but I would like feedback from @Munter, @outsideris and/or @Bamieh.  This is a breaking change, so we need to be sure we want this.

lib/runner.js
```diff
@@ -314,6 +314,9 @@ Runner.prototype.hook = function (name, fn) {
             suite.tests.forEach(function (test) {
               test.pending = true;
             });
+            suite.suites.forEach(function (suite) {
```
boneskull: This is for `beforeEach`/`afterEach` -- **not** what we wanted to do, right?

test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
```diff
@@ -0,0 +1,60 @@
+'use string';
```
boneskull: typo

boneskull: otherwise, good work on these fixtures.

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
```diff
@@ -0,0 +1,35 @@
+describe('skip in before with nested describes', function () {
+  before(function (done) {
+    var self = this;
+    setTimeout(function () {
```
boneskull: what's the necessity of doing this asynchronously?

test/integration/pending.spec.js
```diff
@@ -59,6 +61,42 @@ describe('pending', function () {
...
           }
           assert.equal(res.stats.pending, 2);
+          assert.equal(res.stats.passes, 2);
+          assert.equal(res.stats.failures, 0);
+          assert.equal(res.code, 0);
+          done();
+        });
+      });
+      it('should run before and after hooks', function (done) {
+        runMocha('pending/skip-sync-before-hooks.fixture.js', args, function (err, res) {
```
boneskull: You *should* be able to get the titles out of `run` / `runJSON`; it's easier to use than inspecting the standard output.

test/integration/pending.spec.js
boneskull: But it's fixed now..?

bannmoore: @boneskull I updated the test to pass _based on the current behavior_. If the expectations in that test are wrong, please let me know. :)

lib/runner.js
bannmoore: Currently, my changes are based off these criteria (also listed above).

If a `before` is skipped:
- run any additional before / after hooks at the same level
- skip any beforeEach / afterEach hooks at the same level
- skip all tests at the same level or in nested suites
- skip all hooks in nested suites

I've been operating under the assumption that `before` hooks are executed prior to `beforeEach` hooks, which would mean that there's no context in which we'd still need to execute `beforeEach` in a skipped suite. Is that the case?

test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
bannmoore: Oops - fixing that now.

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
bannmoore: @boneskull I'm not sure what the "real" use case would look like, but this test validates that an asynchronous skip works with a synchronous nested `describe`. It seemed like something worth checking, even as an edge case. 🤷‍♀️

test/integration/pending.spec.js
bannmoore: @boneskull The issue here is that I'm verifying hooks, not `describe` or `it` blocks. The log statements are used to verify that each hook executes / doesn't execute.

This pattern is based off [another test in the suite](https://github.com/mochajs/mocha/blob/master/test/integration/hooks.spec.js).

bannmoore:
@boneskull I addressed a couple of your comments in a fixup commit.

@Bamieh I think part of the problem is that many of us have different ideas of what `skip` in `before` is _supposed_ to do. Getting to consensus is the goal of this PR. 😄

Could you clarify what you mean by `skip` in `before` skipping _only_ that hook? It currently skips `it` blocks in the same `describe`, and the [documentation](https://mochajs.org/#inclusive-tests) supports that: "To skip multiple tests in this manner, use this.skip() in a “before” hook:". It _seems_ like the intent is to use it to skip suites rather than hooks.

Bamieh:
@bannmoore I meant that this is what I believe is "expected" from my personal point of view:

calling `this.skip()` in the `describe` function itself will skip the describe hooks and test cases.

calling `this.skip()` in a `hook` will only skip that hook.

lib/runner.js
boneskull: Sorry, my bad--was looking at the wrong code.  Yes, we are in agreement.

boneskull:
@Bamieh 

> This is a highly arguable topic. I'd vote for simplicity: Skipping a before is like skipping a test case; the function of that hook only is not executed. Users can skip the corresponding teardown after and other hooks as needed.

I disagree with this, as the same functionality could be expressed by a `return` statement or `done()` call.

Furthermore, this introduces inefficiencies.  Users can skip a suite if `this.skip()` is placed in a `describe` block, but that's currently problematic because determining whether to call `this.skip()` must be asynchronous unless `--delay` is used, further complicating matters.  

If we can't asynchronously determine (with a conditional) whether to skip an entire suite in a `before` hook, then it must be determined in a `beforeEach` hook, which is just wasted cycles if the conditional is invariant.  `this.skip()` in a `beforeEach` is better suited to conditionally skipping on a test-by-test basis.

If anything `this.skip()` in a `describe` is effectively *useless*, because users can simply wrap tests, hooks, and nested suites in a conditional:

```js
describe('skipping this', function () {
  if (reasons()) {
    this.skip();
  }

  it('should be skipped', function () {
    // stuff
  });
});
```

is functionally equivalent to:

```js
describe('skipping this', function () {
  if (!reasons()) {
    it('should be skipped', function () {
      // stuff
    });
  }
});
```

outsideris:
I understand the context of this PR now. 
But I need some time to consider the functionality and review this PR because I haven't ever used `this.skip()`.
It is a great work, so I want to that we make a consensus for `this.skip()` clearly via this PR.

bannmoore:
@boneskull I added some documentation updates. I tried to keep it brief - let me know if you think more detail is warranted, or if it's not what you had in mind.

bannmoore:
@djmitche These are the changes currently covered by this PR:
- fix `before` `skip` behavior in nested `describes`
- update tests to document the new behavior
- update docs to reflect the same

The fix itself is minimal - I'm not making any explicit changes to hooks. My comments above simply describe how hooks _will_ behave after the nested bug is fixed. Since that behavior is not currently documented, the discussion and tests are (in my opinion) an important part of this change. 🤷‍♀️ 

That being said, if the group thinks it's best to split this up I'm not totally opposed... I'd just hate to see this discussion get sidetracked into yet another PR. 😂

djmitche:
Sounds good - maybe I misunderstood the conversation.

outsideris:
@bannmoore Could rebase this onto master branch to resolve conflicting files. IMO the changes in `package-lock.json` doesn't need here.
And it's better if you migrate your tests with [unexpected](https://github.com/mochajs/mocha/pull/3343).

outsideris:
BTW, I agreed with the skip rules. It makes sense for me.

* run any additional before / after hooks at the same level
* skip any beforeEach / afterEach hooks at the same level
* skip all tests at the same level or in nested suites
* skip all hooks in nested suites

bannmoore:
@outsideris @boneskull 

Rebased to fix conflicts. Since `pending.spec.js` doesn't use `chai.expect` and hasn't been migrated to `unexpected` in master, I'm going to hold off rewriting the tests for now. That's not really in the scope of this change.

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
plroebuck: Many of these `setTimeout()`-based tests are simply race conditions that happen to have worked in @bannmoore's favor. Shouldn't we just disallow use of async `this.skip()`? Otherwise, the results are not deterministic...

plroebuck:
Usage as I understand intent:
* Calling `this.skip()` as _first_ thing in `describe()` applies to _everything_ in same scope (including nested `describe()`s).
  Same as `describe('whatever', func(){}).skip()`
* Calling `this.skip()` inside `before()` would _still_ run **all** `before()`s and `after()`s, but nothing else.
  Rationale: Once begun, both `setup` and `teardown` **must** run to completion.
* Calling `this.skip()` inside `beforeEach()` would (wastefully) skip each test.
  Same as `it('test', func(){}).skip()` for every test in same scope.

Suggest we deprecate calling `this.skip()` in `beforeEach()`, `after()`, and `afterEach()`.

plroebuck:
@outsideris, you disagree with my review then?

outsideris:
@plroebuck I just want to say LGTM. And I didn't mean that I disagreed with you.
What points did I disagree do you think? I thought that there are no blocking issues, so I approved. Discussing could continue after I approved it.

Honestly, I really want to discuss many things in mocha with maintainers and contributors. However, I couldn't write my thought exactly in English now. So, I am trying to focus on what I can do currently and improve my English skills.
Make sense?

plroebuck:
I'd like to get something merged as well... just not _this_ in its _current_ state.

As an native English speaker, I take _"approved"_ (in context of a PR) to mean:
 **"IMO, no further discussion or changes need happen. Merge immediately as-is!"**

Just clarifying your intent.

BTW, could you add me as a reviewer so I can comment inline? Thx.

plroebuck:
BTW, could you add me as a reviewer so I can comment inline? Thx.

plroebuck:
@outsideris , assumed maintainer role could add reviewers, as was done about a month ago here.
> boneskull requested review from Munter, outsideris, boneskull and Bamieh 27 days ago

plroebuck:
From README.md (just below the additions):
 > Before Mocha v3.0.0, `this.skip()` was not supported in asynchronous tests and hooks.

I've looked around but saw nothing in 3.0 release notes specifically about this.
What PR added async support?

bannmoore:
@plroebuck I'm responding to your comments in rough top-to-bottom.

- `package-lock.json` is in there because it was marked as modified when I ran `npm install`. I didn't add any dependencies in this PR. 🤷‍♀️ I don't think it makes a huge difference, but I'll defer to the maintainers wishes on whether it should go.
- I'm operating under the assumption that async `skip` is allowed, since I haven't seen or heard anything to the contrary. Your note that the tests include race conditions is accurate. The goal of some of the new tests is to document the behavior _as is_, so that we can be certain of what's expected going forward. 
-  I personally don't care for the name "skip-before-skips-inner-only.fixture.js" for that test, since the hooks are the point of that test more than the "inner/outer" structure. Once again, I'll defer to the group consensus since I think there are several viable options.
- I can update the error messages in `it` blocks and hooks to be less generic (see latest commit). 👍 

I could use some clarification re. this:
> Add beforeEach()/afterEach() to (at least some) tests where it() should be skipped.

The fixtures `skip-async-before-hooks.fixture.js` and `skip-sync-before-hooks.fixture.js` validate the behavior of `beforeEach` and `afterEach`. If you can provide me an example of usage that isn't covered by the existing tests, that would help.

outsideris:
> assumed maintainer role could add reviewers, as was done about a month ago here.

@plroebuck  I guess we can add only maintainers. I can't search you.

outsideris:
@bannmoore If you remove `package-lock.json`, it is better. `package-lock.json` is hard to manage.

You can revert it like:
```sh
$ git checkout b4b85b5^ -- package-lock.json
$ git commit
```

plroebuck:
"skip-async-before-hooks.fixture.js":
L9:
```js
  it('should run this test', function () {
    console.log('outer suite test 1');
  });
```
L16-19:
```js
      setTimeout(function () {
        console.log('skip via setTimeout');
        self.skip();
        done();
      });
```
L23:
```js
      throw new Error('inner beforeEach should not run');
```
L27:
```js
      throw new Error('inner afterEach should not run');
```

Move L33-36 to L51 (move after() after skipped suite)

L54:
```js
  it('should run this test', function () {
    console.log('outer suite test 2');
  });
```

plroebuck:
@bannmoore 
* `package-lock.json` should go. Has nothing to do with this PR.
* Nvm about my previous not doing async skip; running `.skip()` within `done()` is fine. No race   
  conditions exist using callback. For whatever reason this morning, I didn't notice your use of 
  `done()` within `setTimeout()` function.
* Think more descriptive filename would be better - don't care what it is. Maybe add a couple 
  comments in the file header that gives an executive summary of what the test does.

> Add beforeEach()/afterEach() to (at least some) tests where it() should be skipped.
Those two files are fine. Difficult reviewing multiple files (not inline) and unsure of remaining time; I wrote a note JIC.

plroebuck:
"skip-async-before-nested.fixture.js":
L4-7:
```js
    setTimeout(function () {
      console.log('skip via setTimeout');
      self.skip();
      done();
    });
```
L10:
```js
  it('should never run this test', function () {
    throw new Error('outer test should not run');
  });
```
L17:
```js
    it('should never run this test', function () {
      throw new Error('nested test should not run');
    });
```
L28:
```js
    it('should never run this test', function () {
      throw new Error('nested second level test should not run');
    });
```

plroebuck:
"skip-sync-before-hooks.fixture.js":
L9:
```js
  it('should run this test', function () {
    console.log('outer suite test 1');
  });
```
L12-18:
```js
    before(function () {
      console.log('inner before 1');
      console.log('skip');
      this.skip();
    });

    before(function () {
      console.log('inner before 2');
    });

    beforeEach(function () {
      throw new Error('inner beforeEach should not run');
    });

    afterEach(function () {
      throw new Error('inner afterEach should not run');
    });
```
L42:
```js
        throw new Error('skipped suite test should not run');
```
L51:
```js
  it('should run this test', function () {
    console.log('outer suite test 2');
  });
```

plroebuck:
You know, maybe we just refactor the `begin()` function argument that sets the `skip()` out into a function[] and run the test twice, once for sync, then async. Eliminates two test files.

bannmoore:
@outsideris Okay - `package-lock.json` is out. Thanks!

boneskull:
I'm satisfied with this.  Unless there's objection, I'd like to merge it.  I'll merge it Monday, August 13 otherwise.  It will go into the next major release, which may be the next release.

boneskull:
@bannmoore *Thank you* for your patience and sticking with this by the way.  I know we don't make it easy.

bannmoore:
@boneskull No problem at all! I'm happy to hear it's on track for merge. Just let me know if there's anything else you need from me.

boneskull:
@plroebuck *ping*.  I want to merge this.  Are your concerns addressed (enough) that we can do that, and make any other changes afterwards?

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+      self.skip();
+      done();
+    }, 50);
```
plroebuck: Why was delay not removed?

test/integration/fixtures/pending/skip-async-before.fixture.js
```diff
@@ -1,18 +1,21 @@
...
-    throw new Error('never thrown');
-  });
+  it('should run this test', function () {});
+
+  describe('skip in before', function () {
+    before(function (done) {
+      var self = this;
+      setTimeout(function () {
+        self.skip();
+      }, 50);
```
plroebuck: and here

test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
```diff
@@ -5,14 +5,11 @@ describe('skip in beforeEach', function () {
     var self = this;
     setTimeout(function () {
       self.skip();
+      done();
     }, 50);
```
plroebuck: and here

test/integration/pending.spec.js
```diff
@@ -127,6 +187,64 @@ describe('pending', function() {
             return;
           }
           assert.equal(res.stats.pending, 2);
+          assert.equal(res.stats.passes, 2);
+          assert.equal(res.stats.failures, 0);
+          assert.equal(res.code, 0, 'code');
```
plroebuck: Should `'code'` have been on the others asserting `res.code`?

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  it('should never run this test', function () { });
```
plroebuck: Should this throw?

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+
+  describe('nested describe', function () {
+    before(function () {
+      throw new Error('first level before should not run');
+    });
+
+    it('should never run this test', function () { });
```
plroebuck: Should this throw then?

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+    after(function () {
+      throw new Error('first level after should not run');
+    });
+
+    describe('nested again', function () {
+      before(function () {
+        throw new Error('second level before should not run');
+      });
+
+      it('should never run this test', function () { });
```
plroebuck: Should this throw then?

test/integration/fixtures/pending/skip-async-before.fixture.js
```diff
@@ -1,18 +1,21 @@
...
+    });
 
-  it('should never run this test', function () {
-    throw new Error('never thrown');
+    it('should never run this test', function () {});
+    it('should never run this test', function () {});
```
plroebuck: Should these throw then?

test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
```diff
@@ -5,14 +5,11 @@ describe('skip in beforeEach', function () {
...
-  it('should never run this test', function () {
-    throw new Error('never thrown');
-  });
-
-  it('should never run this test', function () {
-    throw new Error('never thrown');
-  });
+  it('should skip this test', function () {});
+  it('should error before reaching this test', function () {});
+  it('should error before reaching this test', function () {});
```
plroebuck: Should these throw then?

test/integration/fixtures/pending/skip-sync-before-nested.fixture.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+describe('skip in before with nested describes', function () {
+  before(function () {
+    this.skip();
+  });
+
+  it('should never run this test', function () { });
```
plroebuck: Should this throw then?

test/integration/fixtures/pending/skip-sync-before-nested.fixture.js
```diff
@@ -0,0 +1,33 @@
...
+
+  describe('nested describe', function () {
+    before(function () {
+      throw new Error('first level before should not run');
+    });
+
+    it('should never run this test', function () { });
```
plroebuck: Should this throw then?

test/integration/fixtures/pending/skip-sync-before-nested.fixture.js
```diff
@@ -0,0 +1,33 @@
...
+    after(function () {
+      throw new Error('first level after should not run');
+    });
+
+    describe('nested again', function () {
+      before(function () {
+        throw new Error('second level before should not run');
+      });
+
+      it('should never run this test', function () { });
```
plroebuck: Should this throw then?

test/integration/fixtures/pending/skip-sync-before.fixture.js
```diff
@@ -1,15 +1,18 @@
...
+
+  describe('skip in before', function () {
+    before(function () {
+      this.skip();
+    });
 
-  it('should never run this test', function () {
-    throw new Error('never thrown');
+    it('should never run this test', function () {});
+    it('should never run this test', function () {});
```
plroebuck: Should these throw then?

plroebuck:
@boneskull, added my final comments. Most are minor, but really think the fixture tests titled "should not throw" _should_ throw as their implementation. Would _prefer_ to remove the delays since they add nothing but extra time to run the tests (especially in CI).

test/integration/fixtures/pending/skip-async-spec.fixture.js
```diff
@@ -5,7 +5,7 @@ describe('skip in test', function () {
     var self = this;
     setTimeout(function () {
       self.skip();
-    }, 50);
+    }, 0);
```
plroebuck: Just remove the value. If ommitted, `0` is used.

test/integration/fixtures/pending/skip-sync-before-hooks.fixture.js
```diff
@@ -0,0 +1,57 @@
...
+      throw new Error('inner suite test should not run');
+    });
+
+    describe('skipped suite', function () {
+      before(function () {
+        console.log('skipped before');
+      });
+
+      it('should never run this test', function () {
+        throw new Error('skipped suie test should not run');
```
plroebuck: spelling

test/integration/fixtures/pending/skip-async-before-nested.fixture.js
```diff
@@ -7,14 +7,18 @@ describe('skip in before with nested describes', function () {
     }, 0);
   });
 
-  it('should never run this test', function () { });
+  it('should never run this test', function () {
+    throw new Error('never thrown');
```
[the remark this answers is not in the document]
bannmoore: Fixed.

plroebuck:
Looks good. Sorry if you felt the requested changes were too pedantic...

bannmoore:
Rebased onto latest Master.

bannmoore:
> Looks good. Sorry if you felt the requested changes were too pedantic...

No worries @plroebuck . You caught me with some spare time today. 👍

boneskull:
happy to finally get this merged.  sorry it took forever; thanks for sticking with it @bannmoore!

PR #3234, opened 2018-02-06
add option to show error details immediately in dot, list, min, and spec reporters

jedwards1211:
### Requirements

### Description of the Change

Now if you run `mocha` with `--reporter-options showErrorsImmediately=1` (or whatever string) the dot, list, min, and spec reporters will output error details immediately after test failure instead of after all tests have finished.

To support this I extracted a `Base.logTestFailure` function from `Base.list` so that reporters could call it with a single failed test when the `showErrorsImmediately` option is given.

I also made `Base.prototype.epilogue` not output error details if called with `{listFailures: false}`; the dot, list, min, and spec reporters use this to prevent error details from being output again at the end if they were already output inline.

### Alternate Designs

I had considered making a third-party reporter package that logs errors immediately, but since Mocha doesn't provide a clean way to use multiple reporters at once, such a package wouldn't be very convenient.

I had also considered outputting errors immediately in other console reporters like landing, progress, and nyan, but since it would garble the appearance of their output I decided against it.

It keeps this option consistent among reporters for which it is sensible to log errors immediately and ensures that immediate error details will have the same formatting as epilogue error details.

### Benefits

Developers who have long-running tests will be able to use this option to see error details immediately upon test failure instead of waiting a long time for all tests to finish to see the error details.

### Possible Drawbacks

Some users may be confused why other reporters don't support `showErrorsImmediately`.

### Applicable issues

This is an enhancement (minor release)


docs/index.md
```diff
@@ -1175,6 +1175,15 @@ Mocha allows you to define custom reporters. For more information see the [wiki]
 
 **The HTML reporter is not intended for use on the command-line.**
 
+### Printing error messages immediately
+
+If you have long-running tests you may want to print errors immediately to the console so that you don't have to wait for the tests to finish to see why a given test failed.
```
boneskull: this is kind of a long sentence.  maybe:

> By default, error details will be output at the end of the test run.  If you have long-running tests, you may want to print errors as they occur.

docs/index.md
```diff
@@ -1175,6 +1175,15 @@ Mocha allows you to define custom reporters. For more information see the [wiki]
...
+The Spec, Dot Matrix, List, and Min reporters support a `showErrorsImmediately` option that will output error details immediately when they occur instead of after all tests have finished:
```
boneskull: please refer to the reporters in preformatted text (and lowercase), so: `spec`, `dot`, `list`, and `min`.  also this can be simplified to something like:

> The `spec`, `dot`, `list`, and `min` reporters support a `showErrorsImmediately` option which will print errors as they occur:

docs/index.md
```diff
@@ -1175,6 +1175,15 @@ Mocha allows you to define custom reporters. For more information see the [wiki]
 
 **The HTML reporter is not intended for use on the command-line.**
 
+### Printing error messages immediately
```
boneskull: Thanks for updating the docs!

lib/reporters/base.js
```diff
@@ -317,10 +322,14 @@ function Base (runner) {
...
  * @api public
+ *
+ * @param {Object} options
+ * @param {[boolean]} options.listFailures - if false, will not output failure information
  */
-Base.prototype.epilogue = function () {
+Base.prototype.epilogue = function (options) {
   var stats = this.stats;
   var fmt;
+  var listFailures = !options || options.listFailures !== false;
```
boneskull: if we're going with an object parameter, please do this instead:

```js
options = options || {};
var listFailures = options.listFailures !== false;
```

this will simplify things if we need to add options later.

lib/reporters/base.js
```diff
@@ -317,10 +322,14 @@ function Base (runner) {
  * the bundled reporters.
  *
  * @api public
+ *
+ * @param {Object} options
+ * @param {[boolean]} options.listFailures - if false, will not output failure information
```
boneskull: I haven't seen this syntax before, but I believe the following works (and it also shows what the default is; also `options` is optional):

```js
/*
 * @param {Object} [options] - Options
 * @param {boolean} [options.listFailures=true] - If false, will not output failure information
 */
```

lib/reporters/dot.js
```diff
@@ -19,9 +19,13 @@ exports = module.exports = Dot;
  *
  * @api public
  * @param {Runner} runner
+ * @param {Object} options
```
boneskull:
```js
/*
 * @param {Object} [options] - options
 * @param {Object} [options.reporterOptions] - reporter options
 * @param {boolean} [options.reporterOptions.showErrorsImmediately=false] - if truthy, errors will be 
 *   shown immediately after tests fail instead of after all tests have finished.
 */
```

`boolean` type makes it more clear that this is a *flag*.

lib/reporters/dot.js
```diff
@@ -19,9 +19,13 @@ exports = module.exports = Dot;
...
+ * @param {any} options.reporterOptions.showErrorsImmediately - if truthy, errors will be shown
+ * immediately after tests fail instead of after all tests have finished.
  */
-function Dot (runner) {
+function Dot (runner, options) {
```
boneskull: You know what?  We don't actually use anything in `options` passed to these reporters *except* `reporterOptions`, so I think we could just refactor across the board so it's not nested this way within another object.  You're welcome to refactor it, or not.  :wink:

lib/reporters/base.js
boneskull: I'm thinking we might want to rename `listFailures` to `showFailures` to keep our verbs in sync with the reporter option...

test/reporters/min.spec.js
```diff
@@ -15,6 +17,12 @@ describe('Min reporter', function () {
     process.stdout.write = function (string) {
       stdout.push(string);
     };
+    useColors = Base.useColors;
```
boneskull: I'm going to assume this is necessary because it's easier to parse the output?

jedwards1211: Right, the other reporter tests do this

jedwards1211:
Oh, I didn't know about the `--bail` option.  That might help, but I still think it might be convenient to keep running other tests while investigating why one failed.  I miss being able to see stack traces immediately  with tools like JUnit in the Java world...I don't really see why not have the option to get them as the tests run.  Also, does `--bail --watch` work?

docs/index.md
jedwards1211: Yeah that's better

jedwards1211:
I'll work more on this tomorrow, your comments all make sense

lib/reporters/dot.js
jedwards1211: @boneskull well, even if you guys don't use it, some 3rd-party reporters do for better or worse, so that would have to be `semver-major`.

lib/reporters/dot.js
jedwards1211: @boneskull it's not possible for it to be boolean if it's coming from CLI though is it?  I'm assuming any values from `--reporter-options` will be `string`s.

jedwards1211:
@boneskull the CI build is complaining of lint errors in `CHANGELOG.md` that aren't due to this PR, other than that it passed

jedwards1211:
@boneskull my current workaround for this is
```js
const {reporters: {Base}} = require('mocha')

afterEach(async function () {
  if (this.currentTest.state === 'failed') Base.list([this.currentTest])
}
```

jedwards1211:
@boneskull are you waiting on anything for this?  The CI build failure was not this PR's fault.

plroebuck:
Think use of `--bail` would eliminate much of need for this change.

jedwards1211:
Speak for yourself.  I'd like to be able to keep running the rest of my (numerous, slow) selenium integration tests while hunting down the cause of failures that have already occured.  Believe me, it would save me time.

plroebuck:
> Speak for yourself. [SNIP]

My apologies at having offended. My assumption you were using a CI environment. If so, you could run your suite both with and without `--bail`. From a testing perspective, the first error often causes a cascade. My point being that you could start working on any problem using the result of `--bail` while your long-running (`--no-bail`) version continues to test for everything else.

lib/reporters/dot.js
boneskull: hmm, you're probably right.  not a big deal.  this is javascript after all

lib/reporters/dot.js
boneskull: speaking of 3rd-party reporters--what's stopping you from taking this change and making it into a 3rd-party reporter?

boneskull:
conflicts will need to be resolved ...

jedwards1211:
@plroebuck it's more convenient for me to just use the following semi-hack for now:
```js
const {reporters: {Base}} = require('mocha')

afterEach(function () {
  const {state, title} = this.currentTest
  if (state === 'failed') {
    Base.list([this.currentTest])
  }
}
```
I just wanted to get this merged so I don't have to rely on internal API.

jedwards1211:
@boneskull okay, I resolved all the conflicts

docs/index.md
plroebuck: Think this sentence reads better without the word "immediately".

| The `spec`, `dot`, `list`, and `min` reporters support a `showErrorsImmediately` option that will output error details when they occur instead of after all tests have finished:

boneskull:
this is going to need more attention to get past CI, but haven't looked into it

craigtaub:
@boneskull have you had any time to look into this? Sounds useful.

JoshuaKGoldberg:
🤖 Closing out old PRs to keep the review queue small. Please yell at me if you need it re-opened and can continue working on it. Cheers!

PR #3239, opened 2018-02-11
generate API documentation

dfberry:
Changes in jsdocs comments only in order to product api docs.

### Description of the Change

changes to get markdown for api docs using documentation.js

### Alternate Designs

The current code exports, inherits, and layers make jsdocs a bit of a mash up. There are some errors here but I'm not sure if the easier fix is jsdocs annotations or code changes. 

Shows APIs

### Benefits

Better communication for API usage.

### Possible Drawbacks

Some assumptions are made about inheritance and relationships. I may have left things off or added things that should be there. 

### Applicable issues

I need to understand how jsdocs is built by documentationjs to resolve some of the existing issues.


boneskull:
Thanks.  @dfberry `package-scripts.js` needs to be linted

PR #3241, opened 2018-02-12
Use execArgv over process.argv to support passing `--experimental-modules` and `--loader` arguments

guybedford:
### Description of the Change

I'm running NodeJS with `--experimental-modules` and `--loader` flags, and during the bootstrap of mocha (running `bin/mocha`) I'd like for it to use the same loader and modules settings as my main Node process that is running `bin/mocha`.

By switching to using `process.execArgv` over `process.argv` we can ensure these flags are passed.

### Alternate Designs

The alternative is to take flags to mocha itself, but in this context that does not scale.

This is a key functionality of the boostrap process.

### Benefits

Mocha will work naturally with `node bin/mocha --experimental-modules --loader my-custom loader`.

### Possible Drawbacks

The only issues I could see is if users currently run mocha bin with certain flags they don't want pased to the subprocess. But I can't envisage why this would ever be a problem as most users wouldn't even necessarily know that mocha is running a subprocess.

### Applicable issues

I couldn't find any previous issues on `execArgv` discussion.

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -16,7 +16,7 @@ const args = [path.join(__dirname, '_mocha')];
 // Must be loaded here to handle node-specific options
 getOptions();
 
-process.argv.slice(2).forEach(arg => {
+process.execArgv.forEach(arg => {
   const flag = arg.split('=')[0];
 
   switch (flag) {
@@ -52,7 +52,9 @@ process.argv.slice(2).forEach(arg => {
       args.unshift(arg);
       break;
     default:
-      if (arg.indexOf('--harmony') === 0) {
+      if (arg.indexOf('--experimental-modules') === 0) {
+        args.unshift(arg);
+      } else if (arg.indexOf('--harmony') === 0) {
         args.unshift(arg);
       } else if (arg.indexOf('--trace') === 0) {
         args.unshift(arg);


bin/mocha
```diff
@@ -52,7 +52,9 @@ process.argv.slice(2).forEach(arg => {
       args.unshift(arg);
       break;
     default:
-      if (arg.indexOf('--harmony') === 0) {
+      if (arg.indexOf('--experimental-modules') === 0) {
+        args.unshift(arg);
```
guybedford: The only reason this line is necessary is to ensure that `--harmony` flags aren't moved to come before it, as this needs to be an early flag.

guybedford:
Any thoughts on this?

boneskull:
Taking a look now.... @guybedford can you please sign the CLA?

bin/mocha
```diff
@@ -16,7 +16,7 @@ const args = [path.join(__dirname, '_mocha')];
 // Must be loaded here to handle node-specific options
 getOptions();
 
-process.argv.slice(2).forEach(arg => {
+process.execArgv.forEach(arg => {
```
boneskull: This breaks a lot of stuff... I'm not sure this is what you meant to do.  Basically any argument that isn't a `node` argument won't get passed to `_mocha`.

bin/mocha
boneskull: lgtm, though this code sucks right here

bin/mocha
boneskull: (see [the build](https://travis-ci.org/mochajs/mocha/builds/340422371?utm_source=github_status&utm_medium=notification) for what's going on)

guybedford:
Thanks so much for the review.

I actually managed to work around this in the end by specifying the loader through the `NODE_OPTIONS` environment variable instead of using command line flags, so this isn't necessary anymore.

Issue #3184, opened 2017-12-28
Multiple `describe.skip` marks other describes as pending


MCSH:
### Prerequisites

* [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
* [x] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
* [x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
* [x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
`node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description
I ran into something weird while using `describe.skip`.  If I have nested describes, with at least 2 skipped describes inside one, all other describes will be tagged pending.

### Steps to Reproduce

First file names 10.test.js:
```
var assert = require('assert');

describe("Some tests", function(){
  describe("First describe", function(){
    it("should pass", function(){
      assert.equal(1,1)
    })
  })

  describe.skip("Some disabled test");
  describe.skip("Another disabled test");
  describe.skip("One more disabled test");
})
```

Second file named 20.test.js
```
var assert = require('assert');

describe("Some other tests", function(){
  describe("a new describe", function(){
    it("should still pass", function(){
      assert.equal(1,1)
    })
  })
})
```

Results in:
```


  Some tests
    First describe
      ✓ should pass
    Some disabled test
      Another disabled test
        Some other tests
          a new describe
            - should still pass


  1 passing (5ms)
  1 pending
```

Note that it is possible to do this with only one file as well:
test.js:
```
var assert = require('assert');

describe("Some tests", function(){
  describe.skip("Some disabled test");
  describe.skip("Another disabled test");
  describe("First describe", function(){
    it("should pass", function(){
      assert.equal(1,1)
    })
  })
})

```

Results in:
```  Some tests
    Some disabled test
      Another disabled test
        First describe
          - should pass


  0 passing (4ms)
  1 pending


```

**Expected behavior:** Only the skipped describes should be marked pending

**Actual behavior:** All describes after the first two skipped ones will be marked pending.

[...]

Issue #2014, opened 2015-12-17
🐛 Bug: Contexts not properly reset between tests within the same level


moll:
Hey,

I wouldn't be surprised if related to my years old bug report and gripe: https://github.com/mochajs/mocha/issues/1195

``` js
var assert = require("assert")

describe("Parent", function() {
  beforeEach(function() { this.value = 42 })

  it("must pass 1", function() { assert.equal(this.value, 42) })
  it("must pass 2", function() { assert.equal(this.value, 42) })

  describe("Child", function() {
    beforeEach(function() {
      assert.equal(this.value, 42)
      this.value = 13
    })

    it("must pass 1", function() { assert.equal(this.value, 13) })
    it("must pass 2", function() { assert.equal(this.value, 13) })
  })
})
```

Fails with:

```
  1) Parent Child "before each" hook for "must pass 2":

      AssertionError: 13 == 42
      + expected - actual

      -13
      +42

      at Context.<anonymous> (test/index_test.js:11:14)
```

danielstjules:
What are you expecting in terms of behavior? With a simpler example, would you expect this to pass:

``` javascript
var assert = require('assert');

describe('Suite', function() {
  it('test1', function() {
    this.value = 'foo';
    assert.equal(this.value, 'foo');
  });

  it('test2', function() {
    assert.notEqual(this.value, 'foo');
  });
});
```

```
  Suite
    ✓ test1
    1) test2


  1 passing (12ms)
  1 failing

  1) Suite test2:

      AssertionError: 'foo' != 'foo'
      + expected - actual
```

moll:
I would expect to _not_ have `this.value` set in "test2". Sharing properties between tests such a way is despicable. Even if they were left there, assignments in `beforeEach` must take precedence. ;)

danielstjules:
I agree that sharing state between tests is a bad idea :) But it's up to you to clean up state in an afterEach/beforeEach hook. I see these two suites as equal:

``` javascript
var assert = require('assert');

describe('Suite1', function() {
  it('test1', function() {
    this.value = 'foo';
    assert.equal(this.value, 'foo');
  });

  it('test2', function() {
    assert.notEqual(this.value, 'foo');
  });
});

describe('Suite2', function() {
  var value;

  it('test1', function() {
    value = 'foo';
    assert.equal(value, 'foo');
  });

  it('test2', function() {
    assert.notEqual(value, 'foo');
  });
});
```

```
  Suite1
    ✓ test1
[...]

PR #3242, opened 2018-02-14
keep hierarchy for skipped suites w/o a callback

outsideris:
### Requirements

### Description of the Change
As my local test for #3184, `describe.skip` without a callback couldn't keep a hierarchy of suites.

For Instance.

```js

describe("Some tests", function(){
  describe.skip("Some disabled test", function() {});
});

describe("Some other tests", function(){
  it("should still pass", function(){
    assert.equal(1,1);
  });
});
```

is show below output.

```sh
  Some other tests
    ✓ should still pass


  1 passing (5ms)
```

It is correct. But when omit a callback in `describe.skip`.

```js

describe("Some tests", function(){
  describe.skip("Some disabled test");
});

describe("Some other tests", function(){
  it("should still pass", function(){
    assert.equal(1,1);
  });
});
```

A hierarchy of the suites is wrong like below.

```sh
  Some tests
    Some other tests
      ✓ should still pass


  1 passing (5ms)
```

So, this PR add `suites.shift();` for a skipped suite w/o a callback.

### Alternate Designs
I'm not sure.

Some test can be skipped not intentionally.

### Benefits
Skipped suites don't make side effect for other tests.

### Possible Drawbacks
I'm not sure.

### Applicable issues
#3184 

I don't agree #2014 is related to this issue.

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -113,6 +113,8 @@ module.exports = function (suites, context, mocha) {
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
           throw new Error('Suite "' + suite.fullTitle() + '" was defined but no callback was supplied. Supply a callback or explicitly skip the suite.');
+        } else if (!opts.fn && suite.pending) {
+          suites.shift();
         }
 
         return suite;
--- a/test/integration/fixtures/pending/skip-hierarchy.fixture.js
+++ b/test/integration/fixtures/pending/skip-hierarchy.fixture.js
@@ -1,0 +1,9 @@
+'use strict';
+
+describe('a suite', function(){
+  describe.skip('skipped suite 1');
+  describe.skip('skipped suite 2');
+  describe('another suite', function(){
+    it('a test', function(){})
+  })
+});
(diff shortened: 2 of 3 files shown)


lib/interfaces/common.js
```diff
@@ -113,6 +113,8 @@ module.exports = function (suites, context, mocha) {
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
           throw new Error('Suite "' + suite.fullTitle() + '" was defined but no callback was supplied. Supply a callback or explicitly skip the suite.');
+        } else if (typeof opts.fn === 'undefined' && suite.pending) {
```
boneskull: please simplify

```js
} else if (!opts.fn && suite.pending) {
```

test/integration/fixtures/pending/skip-hierarchy.fixture.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+describe('a suite', function(){
+  describe.skip('skipped suite 1');
+  describe.skip('skipped suite 2');
+  describe('anothor suite', function(){
```
boneskull: spelling: "another"

boneskull:
@outsideris Thanks!

PR #3243, opened 2018-02-14
Give a deprecation warning if a suite callback returns a value.

### LINKED ISSUE #2975: Describe block with async function behaving weirdly
Mocha **3.5**
Node **8.4**

```
describe('Test Suite 1', async function(){
  let email = await faker.internet.email();
  it('should print email', function(){
     email.should.not.be.a('string')
   )}
})
```
Running this test should give
```
Test Suite 1
   ✓should print email
```
But its giving
`0 passing`

Also modifying the above code to
```
describe('Test Suite 1', async function(){
  let email;
  it('should print email', async function(){
     email = await faker.internet.email()
     email.should.not.be.a('string')
   )}
})
```
Runs the test but doesn't print the name of describe i.e Test Suite 1
 ` ✓should print email`
Removing the async from **describe** and putting it in **it** works fine
```
 describe('Test Suite 1', function(){
  it('should print email',async  function(){
     let email = await faker.internet.email();
     email.should.not.be.a('string')
   )}
})
```
```
Test Suite 1
   ✓should print email
```

Also .**only** and **.skip** do not work when **async** function is included in the **describe** block

jleedev:
### Description of the Change

As long as this is not supported, the behavior is confusing. It can produce a suite that is not properly initialized. Turning this into a hard error will make it clear that this is not supported.

In short, the test suite below currently does nothing; this change will make it into an error.

```javascript
describe("my suite", async function() {
  await Promise.resolve();
  it("does a thing", function() {
    assert(false);
  });
});
```

Other usage could possibly produce a partially-initialized test suite that fails for unknown reasons; this will make it fail obviously.

### Alternate Designs

Original proposal was to throw, but settled for a deprecation warning in the accepted version.
This would need be revisited in future to throw an Error instead.

N/A

### Benefits

Reduction in user confusion/expectations attempting to use `describe` function as anything but a namespace.

### Possible Drawbacks

Users will get deprecation warning for unsupported behavior.
No _real_ downside to knowledge.

### Applicable issues

Fixes #2975
semver-patch

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var Suite = require('../suite');
+var utils = require('../utils');
 
 /**
  * Functions common to more than one interface.
@@ -110,8 +111,8 @@ module.exports = function (suites, context, mocha) {
         }
         if (typeof opts.fn === 'function') {
           var result = opts.fn.call(suite);
-          if (result && typeof result.then === 'function') {
-            throw new Error('Suite "' + suite.fullTitle() + '" was given an async callback. This will not behave as you expect.');
+          if (typeof result !== 'undefined') {
+            utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');
           }
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -111,7 +111,7 @@ module.exports = function (suites, context, mocha) {
         }
         if (typeof opts.fn === 'function') {
           var result = opts.fn.call(suite);
-          if (result) {
+          if (typeof result !== 'undefined') {
             utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');
           }
           suites.shift();
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -112,7 +112,7 @@ module.exports = function (suites, context, mocha) {
         if (typeof opts.fn === 'function') {
           var result = opts.fn.call(suite);
           if (typeof result !== 'undefined') {
-            utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');
+            utils.deprecate('Deprecation Warning: Suites do not support a return value;' + opts.title + ' returned :' + result);
           }
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -109,7 +109,10 @@ module.exports = function (suites, context, mocha) {
           suite.parent._onlySuites = suite.parent._onlySuites.concat(suite);
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          var result = opts.fn.call(suite);
+          if (result && typeof result.then === 'function') {
+            throw new Error('Suite "' + suite.fullTitle() + '" was given an async callback. This will not behave as you expect.');
+          }
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
           throw new Error('Suite "' + suite.fullTitle() + '" was defined but no callback was supplied. Supply a callback or explicitly skip the suite.');
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -111,7 +111,7 @@ module.exports = function (suites, context, mocha) {
         }
         if (typeof opts.fn === 'function') {
           var result = opts.fn.call(suite);
-          if (typeof result !== 'undefined') {
+          if (result) {
             utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');
           }
           suites.shift();
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -544,15 +544,30 @@ exports.getError = function (err) {
 };
 
 /**
- * Show a deprecation warning
+ * Show a deprecation warning. Each distinct message is only displayed once.
  *
  * @param {string} msg
  */
-exports.deprecate = process.emitWarning
+exports.deprecate = function (msg) {
+  msg = String(msg);
+  if (seenDeprecationMsg.hasOwnProperty(msg)) {
+    return;
+  }
+  seenDeprecationMsg[msg] = true;
+  doDeprecationWarning(msg);
+};
+
+var seenDeprecationMsg = {};
+
+var doDeprecationWarning = process.emitWarning
   // Node.js v6+
   ? function (msg) { process.emitWarning(msg, 'DeprecationWarning'); }
+  // Node.js v4+
+  // process.emitWarning emits on the next tick; emulate that.
+  : process.nextTick ?
+  function (msg) { process.nextTick(() => console.warn(msg)); }
   // Everything else
-  : function (msg) { console.warn(msg); };
+  : function (msg) { Promise.resolve().then(() => console.warn(msg)); };
 
 /**
  * @summary
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -564,10 +564,10 @@ var doDeprecationWarning = process.emitWarning
   ? function (msg) { process.emitWarning(msg, 'DeprecationWarning'); }
   // Node.js v4+
   // process.emitWarning emits on the next tick; emulate that.
-  : process.nextTick ?
-  function (msg) { process.nextTick(() => console.warn(msg)); }
-  // Everything else
-  : function (msg) { Promise.resolve().then(() => console.warn(msg)); };
+  : process.nextTick
+    ? function (msg) { process.nextTick(() => console.warn(msg)); }
+    // Everything else
+    : function (msg) { Promise.resolve().then(() => console.warn(msg)); };
 
 /**
  * @summary
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -563,7 +563,7 @@ var doDeprecationWarning = process.emitWarning
   // Node.js v6+
   ? function (msg) { process.emitWarning(msg, 'DeprecationWarning'); }
   // Everything else
-  :  function (msg) { process.nextTick(() => console.warn(msg)); }
+  : function (msg) { process.nextTick(() => console.warn(msg)); };
 
 /**
  * @summary
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -544,6 +544,17 @@ exports.getError = function (err) {
 };
 
 /**
+ * Show a deprecation warning
+ *
+ * @param {string} msg
+ */
+exports.deprecate = process.emitWarning
+  // Node.js v6+
+  ? function (msg) { process.emitWarning(msg, 'DeprecationWarning'); }
+  // Everything else
+  : function (msg) { console.warn(msg); };
+
+/**
  * @summary
  * This Filter based on `mocha-clean` module.(see: `github.com/rstacruz/mocha-clean`)
  * @description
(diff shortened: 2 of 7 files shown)


Bamieh:
@jleedev Awesome!

Indeed the `describe` function should not be async, the approach is to use `before` for such cases.

boneskull:
Thanks.  I'm not sure adding more checks around ways in which Mocha can be misused is something we need to be in the business of.  Nowhere in the documentation does it state that the return value of a suite callback is consumed.  Similarly, you could pass a `done` callback to the suite, which is also unsupported--should we add a special exception for that?  It's really a slippery slope...

boneskull:
@Bamieh Given my comments above, do you still think this should be merged?

jleedev:
My argument here is that "just slap async on the function" is _so tempting_ to integrate with whatever resources are needed, and it's easy to break your entire test suite in a subtle way. I don't think it's at all the same as attemping to pass `done` to the suite (and not just because calling `done()` would raise a runtime error). It's easy to forget that `function` and `async function` have a big difference.

That being said, it could be beneficial to call out a suggested fix — "You probably want to use `before()` instead."

ORESoftware:
if the describe block returns *anything* truthy, throw an error

boneskull:
I get what you're saying about `before()`, but it happens at a totally different phase of execution.  You don't want to mix that stuff up.

I like @ORESoftware's suggestion, actually, but that is a breaking change, because it doesn't necessarily cause anybody's tests to fail.  For example:

```js
  it('is my test', function () {
    // ...
  });

  return true;
});
```

It's pointless, but it currently works, unlike using `async function`.  

What would we think of adding a soft deprecation warning if a value is returned from a `describe` block, instead of an error?  Next major would then toss an exception.

ORESoftware:
@boneskull add a warning if anything truthy is return for the next few patches, then maybe next minor release throw an error?

boneskull:
@ORESoftware well, next *major* since it's breaking, but yeah, that's the idea.

ORESoftware:
yeah I guess that makes sense

boneskull:
@jleedev are you on-board changing this to a deprecation message?  basically:

- If *anything* is returned from a suite, let's warn.  Use something like:
  ```js
  // lib/utils.js (untested)
  exports.deprecate = process.emitWarning 
    // Node.js v6+
    ? function (msg) { process.emitWarning(msg, 'DeprecationWarning'); }
    // Everything else
    : function (msg) { console.warn(msg); };

  // elsewhere
  if (suiteReturnsSomething) {
    utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');
  }
  ```
- Ensure this is done in a way which will work in a browser

jleedev:
Yes, I like that. Branch updated with this strawman. Robots are running the browser tests.

test/integration/suite.spec.js
```diff
@@ -59,7 +59,7 @@ describe('suite returning a value', function () {
         done(err);
         return;
       }
-      var pattern = new RegExp('DeprecationWarning', 'g');
+      var pattern = new RegExp('deprecated', 'g');
```
Mingun: Why not use regexp literal? AFAIK this will be compiled every time as this code executed

jleedev: And this code is executed exactly once, so it doesn't matter, and I was copying the style used in the rest of this file.

lib/interfaces/common.js
```diff
@@ -109,7 +110,10 @@ module.exports = function (suites, context, mocha) {
           suite.parent._onlySuites = suite.parent._onlySuites.concat(suite);
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          var result = opts.fn.call(suite);
+          if (typeof result !== 'undefined') {
```
outsideris: Could you simplify this?
```js
if (result) {
```

test/integration/suite.spec.js
Bamieh: @Mingun as this code is only ran once, it is better to compile it at run-time when it is called, rather that having it loaded on script load time.

lib/interfaces/common.js
jleedev: Done.

lib/interfaces/common.js
```diff
@@ -109,7 +110,10 @@ module.exports = function (suites, context, mocha) {
           suite.parent._onlySuites = suite.parent._onlySuites.concat(suite);
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          var result = opts.fn.call(suite);
+          if (result) {
+            utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');
```
Bamieh: As we know `async` functions will return a `<Promise>` however this might not be clear to the average user and thus the message will be vague (they are not returning anything yet this error is stating that they are).

> utils.deprecate('Returning a value from a Suite is deprecated; this will throw an exception in the next major version of Mocha');

Maybe we can reword to something in this context:

```
utils.deprecate('Deprecation Warning: Suites do not support a return value;' + suite.title + 'returned :' + result);
```

lib/interfaces/common.js
```diff
@@ -109,7 +110,10 @@ module.exports = function (suites, context, mocha) {
           suite.parent._onlySuites = suite.parent._onlySuites.concat(suite);
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          var result = opts.fn.call(suite);
+          if (result) {
```
Bamieh: I believe that keeping this `typeof result === 'undefined'` is more consistent since we are deprecating all return values.

outsideris: Original code was `if (typeof result !== 'undefined') {` and I asked to simplify it.
I believe that `if (result)` is enough even if we are deprecating all return values.
What do you think?

Bamieh: @outsideris I'd vote for consistency. Also i find `typeof ... 'undefined'` simple for all JS developers IMO

jleedev: My vote's for `typeof ... 'undefined'` as well; this codebase is not 100% consistent in this matter but I would agree that it's considered idiomatic (if quirky) javascript.

outsideris: @jleedev Sorry for making to fix it twice.

lib/interfaces/common.js
jleedev: Makes sense. Updated.

lib/utils.js
```diff
@@ -544,6 +544,17 @@ exports.getError = function (err) {
   return err || exports.undefinedError();
 };
 
+/**
+ * Show a deprecation warning
+ *
+ * @param {string} msg
+ */
+exports.deprecate = process.emitWarning
```
boneskull: Sorry to complicate this, but it's a bit more complicated.  😝 

1.  [ ] The user should only see this notice *once*.  For each unique `msg`, we need some to keep some flag (probably in an object, as its key, since we have no access to `Set`) and do nothing if the flag is already set.  For safety, coerce `msg` into a `string` via `msg = String(msg)`
2.  [ ] Internally, `process.emitWarning` emits on the next tick.  So, `console.warn(msg)` must happen within `process.nextTick` as well.  This is to avoid bloating or interfering w/ the call stack

boneskull: Also, please rebase onto `master` to make the Snyk warning go away

jleedev: If `process.nextTick` is unavailable (outside of Node), is it preferred to use `Promise.resolve().then(…)` or `setTimeout(…, 0)`?

jleedev: Does this also need to hook into the `--no-deprecation` flags?

lib/utils.js
```diff
@@ -537,6 +537,32 @@ exports.getError = function (err) {
...
+ * Show a deprecation warning. Each distinct message is only displayed once.
+ *
+ * @param {string} msg
+ */
+exports.deprecate = function (msg) {
+  msg = String(msg);
+  if (seenDeprecationMsg.hasOwnProperty(msg)) {
+    return;
+  }
+  seenDeprecationMsg[msg] = true;
```
boneskull: good

lib/utils.js
```diff
@@ -537,6 +537,32 @@ exports.getError = function (err) {
...
+
+var doDeprecationWarning = process.emitWarning
+  // Node.js v6+
+  ? function (msg) { process.emitWarning(msg, 'DeprecationWarning'); }
+  // Node.js v4+
+  // process.emitWarning emits on the next tick; emulate that.
+  : process.nextTick
+    ? function (msg) { process.nextTick(() => console.warn(msg)); }
+    // Everything else
+    : function (msg) { Promise.resolve().then(() => console.warn(msg)); };
```
boneskull: This isn't necessary; we can assume `process.nextTick` will always be defined.  The browser gets a polyfill.  We can safely remove this last branch of the ternary.

jleedev: Done, and rebased again.

jleedev:
> unit test suite for the utils module

I actually added to the integration tests. Is it possible to verify a call to `console.warn` otherwise?

boneskull:
@plroebuck 

> The message itself implies that this was previously supported

I disagree; it was *not disallowed*, not *supported*.  If you pass an extra parameter to any given function in Mocha's API, if it doesn't break, that doesn't imply it's supported.

@Bamieh 

> @boneskull IMHO we should not be warning users for mocha misuse, so in most cases we should not be doing this. However i feel that this warning is reasonable to prevent very subtle errors, but again it is restricting our users for no real reason on the library level, a small warning text in the documentation might suffice here.

I agree that we shouldn't generally be in the business of warning users about misusing the API.  The only reason I'm entertaining it is because of the potential for causing difficult-to-solve problems.

IMO the *real* problem here is that suites don't support async behavior.  Maybe that should change.  If we *do* want that to change (at some point), we shouldn't merge this, and instead add a note to the docs.

@mochajs/core would love further thoughts

Bamieh:
@boneskull I agree with you about not merging this if there is a possibility of having async suites.


I do not see the value of supporting async suites, as I usually encourage people to write all the logic in the `before` of the suite, the describe body can only hold empty `var`/`let` variables to be assigned in the test cases of the suite to solve the function scoping (some people use `this` directly to avoid this as well, but we do not encourage it). Following that logic, having async describe might be contradictory and unnecessary.

jleedev:
I don't understand why the possibility of async suites in the future should prevent merging this right now. Either way, adding async suites will break compatibility to the same degree, and the warning can be removed. No one should be returning a Promise from a suite before they're actually supported.

plroebuck:
@boneskull were your concerns addressed?
if so, could you resolve any remaining conversations and _explicitly_ approve?

boneskull:
> I don't understand why the possibility of async suites in the future should prevent merging this right now. Either way, adding async suites will break compatibility to the same degree, and the warning can be removed. No one should be returning a Promise from a suite before they're actually supported.

churn, mainly.  but async suites is probably a ways off.

test/integration/deprecate.spec.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var assert = require('assert');
+var run = require('./helpers').runMocha;
+var args = [];
+
+describe.only('utils.deprecate test', function () {
```
boneskull: `.only` needs removed

test/integration/deprecate.spec.js
```diff
@@ -0,0 +1,20 @@
...
+  it('should print unique deprecation only once', function (done) {
+    run('deprecate.fixture.js', args, function (err, res) {
+      if (err) {
+        done(err);
+        return;
+      }
+      console.trace(res);
```
boneskull: remove debug

boneskull:
@plroebuck 	I've noted two problems.  fine to merge and then fix them after; I don't want to ask @jleedev to do more work.

boneskull:
I've "approved", but the stuff I noted must be addressed one way or another

plroebuck:
I'll ask. Since this was intended for patch release, plan going forward is to?
1. correct to throw error instead of deprecation warn (for 6.0)
1. merge as is and throw error (for 7.0)

test/integration/deprecate.spec.js
plroebuck: done

test/integration/deprecate.spec.js
plroebuck: done

Issue #3093, opened 2017-11-03
Replacing current working directory in stack trace message not always working as intended.


zrbecker:
I had the following test fail when running `npm test` in a docker container where I had mounting the mocha source at `/local`.
https://github.com/mochajs/mocha/blob/master/test/unit/runner.spec.js#L406

The issue is here.
https://github.com/mochajs/mocha/blob/1bb6b3977a482c144a61d68ea2e499b22a7db01b/lib/utils.js#L600
The replace call, causes '/usr/local/dev' to be changed to '/usrdev/, and the string comparison fails in the assert.

I imagine you would have to do a fancier regex replacement to get this to work correctly. In practice this might not be a huge issue since users tend to be in `/home/username/path` which should be reasonably unique.

However, I can imagine a situation where some files being stored at `/tmp/<random_string>/home/username/path/filename` or a similar location would occur in a stack trace message and be replaced to `/tmp/<random_string>/filename`.

ScottFreeCode:
Hmm... I can't think of a reason other than temporary files to be running JS outside the nice little project-local stack managed by NPM, and I am having a hard time imagining a temporary file reusing the project's path (wherever the project is) inside the temporary files directory, but...

If there's a standard way for a JS regex built from strings to escape one particular part of the regex, we'd just need to escape `cwd` that way and then prepend "^" to it, I believe, so that seems like it would be easy enough provided we aren't writing custom regex voodoo.

ScottFreeCode:
Followup -- I don't know about a standard way, but elsewhere in Mocha (e.g. the `fgrep` function) we use the `var escapeRe = require('escape-string-regexp');` package, like so: `new RegExp(escapeRe(string))` -- presumably `line = line.replace(escapeRe(cwd), '');` would do? (It'd be nice to get an integration test for the filepath formatting in this type of case, and thereby confirm that the fix works.)

boneskull:
closed via #3244

PR #3244, opened 2018-02-15
Persist paths in stack trace which have cwd as infix

outsideris:
### Requirements

### Description of the Change
As #3093 ,  when paths in stack trace messages have a current working directory as infix, it should not be replaced.  So, I added `(` in front of `cwd` to replace only in a case of the paths started with `cwd`. I assumed `(` is always be there.

### Alternate Designs
@ScottFreeCode suggested `'escape-string-regexp'` package. However, I don't understand how it fixes this issue.

It is related with display stack trace in mocah.

### Benefits
This is an edge case, so the most of users don't experience the issue. Some users like in docker can see correct path in a stack trace.

### Possible Drawbacks
If a case of `(` is in front of a path can exist, the paths in a stack trace are not replaced with a current working directory.

### Applicable issues
#3093

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -597,7 +597,7 @@ exports.stackTraceFilter = function () {
 
       // Clean up cwd(absolute)
       if (/\(?.+:\d+:\d+\)?$/.test(line)) {
-        line = line.replace(cwd, '');
+        line = line.replace('(' + cwd, '(');
       }
 
       list.push(line);
--- a/test/node-unit/stack-trace-filter.spec.js
+++ b/test/node-unit/stack-trace-filter.spec.js
@@ -88,6 +88,23 @@ describe('stackTraceFilter()', function () {
         expect(filter(stack.join('\n')))
           .to.equal(expected.join('\n'));
       });
+
+      it('should not replace absolute path which has cwd as infix', function () {
+        var stack = [
+          'Error: /www' + process.cwd() + '/bla.js has a problem',
+          'at foo (/www' + process.cwd() + '/foo/index.js:13:226)',
+          'at bar (/usr/local/dev/own/tmp/node_modules/bluebird/js/main/promise.js:11:26)'
+        ];
+
+        var expected = [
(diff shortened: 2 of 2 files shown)


HarrySarson:
How about replacing 

```js
if (/\(?.+:\d+:\d+\)?$/.test(line)) {
  line = line.replace(cwd, '');
}
```

with 

```js
if (/\(?.+:\d+:\d+\)?$/.test(line) && line.indexOf(cwd) === 0) {
  line = line.replace(cwd, '');
}
```

outsideris:
@HarrySarson 
The line is like `at EventEmitter.<anonymous> (/usr/local/dev/test.js:16:12)`. So, `cwd` couldn't be 0 of index.

HarrySarson:
Yeah, good point @outsideris

lib/utils.js
```diff
@@ -597,7 +597,7 @@ exports.stackTraceFilter = function () {
 
       // Clean up cwd(absolute)
       if (/\(?.+:\d+:\d+\)?$/.test(line)) {
-        line = line.replace(cwd, '');
+        line = line.replace('(' + cwd, '(');
```
boneskull: by using `escape-string-regexp`, @ScottFreeCode meant do this:

```js
line = line.replace(require('escape-string-regexp')(cwd), '');
```

Does that not work?

outsideris: No, that doesn't.

In this case, `escape-string-regexp` change nothing on the path. (I didn't check it with `escape-string-regexp` on Windows) 
```
> var cwd = '/Users/outsideris/github/mocha/';
undefined
> require('escape-string-regexp')(cwd)
'/Users/outsideris/github/mocha/'
```

So, `line = line.replace(require('escape-string-regexp')(cwd), '');` will change `at foo (/www/Users/outsideris/github/mocha/foo/index.js:13:226)` to `at foo (/wwwfoo/index.js:13:226)`. 
It resolves nothing here and results same as before I change it.

Issue #3193, opened 2018-01-04
Ability to show only changes and deletions in diffs, not additions


Mingun:
### Prerequisites

- [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
- [x] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
- [x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
- [x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description

I use `chai` assertion library and [`chai-like`](https://github.com/zation/chai-like) library for the partial comparing of objects: I am interested in existence and compliance only of those properties that are listed in an `expected` object. Existence of additional properties in an `actual` object permissibly. Unfortunately, when diff is created it is not considered in any way and those properties, existence which for the test it is indifferent get to diff. It would be desirable to have a way to suppress their generation in a diff. Example:

Test:
```js
'use strict';

let chai = require('chai');
let expect = chai.expect;
chai.use(require('chai-like'));

describe("diff test", function() {
  let AST = {
    type: 'grammar',
    rules: [
      { type: 'rule', name: 'rule1' },
      { type: 'rule', name: 'rule2' },
    ]
  };
  it("chai-like", function() {
    expect(AST).like({
      rules: [
        { name: 'rule' },
        { name: 'rule2' }
      ]
    });
  });
});
```

This is current actual output which is too large in some cases and contains unnecessary details
```console
  1) diff test
       chai-like:

      AssertionError: expected { Object (type, rules) } to be like { Object (rules) }
      + expected - actual

       {
         "rules": [
           {
      -      "name": "rule1"
      -      "type": "rule"
      +      "name": "rule"
           }
           {
             "name": "rule2"
      -      "type": "rule"
           }
         ]
      -  "type": "grammar"
       }

      at Context.<anonymous> (test\test.js:24:17)
```
But I want that diff looked here so:
```diff
 {
   "rules": [
     {
-      "name": "rule1"
[...]

PR #3251, opened 2018-02-24
[WIP] Implementation of support for diff property in AssertionError

Mingun:
### Description of the Change

The purpose of this PR -- check of hypotheses and testing of changes (because local execution of mocha tests -- pain, fear and horror) and continuation of the discussion begun in https://github.com/chaijs/chai/issues/1120.

- The first commit contains preparatory work for a possibility of getting the structured diff from AssertionError
- The second commit changes the order `actual` and `expected` in diff output, for more logical, with `expected` as a base point. Meaningful of the diff -- changes which need to be applied to `expected` to get `actual`
- The third commit adds a feature for get diff from an error. Its specific implementation -- a subject for discussion. For demonstration of a new ability I made [demo](https://github.com/Mingun/chai-like/tree/demo) branch in the https://github.com/Mingun/chai-like repository. Just clone this repo, switch to the demo branch. Last commit from it uses diff from an error, penultimate -- mocha built-in.

One of current problems with this solution -- the ["List reporter on fail should immediately construct fail strings"](https://github.com/mochajs/mocha/blob/master/test/reporters/list.spec.js#L214-L215) test. It expects that `expected` and `actual` will be strings, but because of the lazy nature of diff generation they cannot be turned into strings, otherwise it is impossible to guarantee correct operation ([see implementation](https://github.com/Mingun/chai-like/blob/demo/lib/diff.js#L41-L43) in chai-like). As such conversion was carried out only if diff was required that I do not think that it important and it is worth saving such behavior.

### Alternate Designs

Instead of `diff` function in an error it is possible to use the property containing both types of diffs (unified and inline), however such solution can decelerate diff generation (there were already issues with speed of diff generation)

### Benefits

Correct diff can be constructed only by assertion library, test runner (mocha) has no enough information for this that the situation with chai-like shows. On the other hand, only test runner knows how diff will be displayed (coloring etc.) assertion library cannot do this.

Transmission of the structural information necessary for creation of text representation, is necessary also enough. After I did this implementation, I agree with this opinion of @keithamus, stated to them in discussion in https://github.com/chaijs/chai/issues/1120.

### Possible Drawbacks

No. mocha still uses built-in diff if it they not exist in an error.

### Applicable issues

#3193

Is it an enhancement (minor release)


lib/reporters/base.js
```diff
@@ -156,14 +156,8 @@ exports.cursor = {
...
-  return err && err.showDiff !== false && sameType(err.actual, err.expected) && err.expected !== undefined;
-}
-
-function stringifyDiffObjs (err) {
-  if (!utils.isString(err.actual) || !utils.isString(err.expected)) {
-    err.actual = utils.stringify(err.actual);
-    err.expected = utils.stringify(err.expected);
-  }
+  return (utils.type(err.diff) === 'function') ||
+         (err.showDiff !== false && sameType(err.actual, err.expected) && err.expected !== undefined);
```
boneskull: please use `typeof err.expected === 'undefined'` for safety

test/integration/fixtures/diffs/output
```diff
@@ -1,91 +1,91 @@
 // DIFF
-      -foo rar baz
-      +foo bar baz
+      -foo bar baz
```
boneskull: why did this have to happen?

Mingun: It and similar changes because of the second commit which changes the order of `actual` and `expected` in diff generation

lib/reporters/base.js
Mingun: Ok. Maybe use `utils.type(err.expected) === 'undefined'` for uniformity with a code in the line above?

test/integration/fixtures/diffs/output
boneskull: why did we change the order?

lib/reporters/base.js
```diff
@@ -368,105 +354,124 @@ function pad (str, len) {
...
   if (lines.length > 4) {
     var width = String(lines.length).length;
-    msg = lines.map(function (str, i) {
-      return pad(++i, width) + ' |' + ' ' + str;
-    }).join('\n');
+    lines = lines.map(function (line, i) { return indent + pad(i + 1, width) + ' | ' + line; });
+  } else {
+    lines = lines.map(function (line) { return indent + line; });
   }
+  lines.push('');
```
boneskull: I'm having trouble following this diff.  Why is this here?

lib/reporters/base.js
```diff
@@ -368,105 +354,124 @@ function pad (str, len) {
...
+      }
+      if (change.removed) {
+        return color('diff removed', line);
+      }
+      return line;
+    }
+
+    change.value.split('\n').forEach(function (line, i) {
+      line = colorize(line);
+      if (i === 0) {
```
boneskull: what's this?

lib/reporters/base.js
```diff
@@ -368,105 +354,124 @@ function pad (str, len) {
...
-  return '\n      ' +
-    colorLines('diff added', '+ expected') + ' ' +
-    colorLines('diff removed', '- actual') +
+
+  var lines = [];
+  diff.hunks.forEach(function (hunk, i) {
+    if (i !== 0) {
+      lines.push('--');
+    }
+    hunk.lines.forEach(function (line) {
```
boneskull: maybe use `reduce` here..?

package.json
```diff
@@ -309,7 +309,7 @@
     "browser-stdout": "1.3.0",
     "commander": "2.11.0",
     "debug": "3.1.0",
-    "diff": "3.3.1",
+    "diff": "3.4.0",
```
boneskull: why?

lib/reporters/base.js
```diff
@@ -156,14 +156,8 @@ exports.cursor = {
...
 function showDiff (err) {
-  return err && err.showDiff !== false && sameType(err.actual, err.expected) && err.expected !== undefined;
-}
-
-function stringifyDiffObjs (err) {
-  if (!utils.isString(err.actual) || !utils.isString(err.expected)) {
-    err.actual = utils.stringify(err.actual);
-    err.expected = utils.stringify(err.expected);
-  }
+  return (utils.type(err.diff) === 'function') ||
```
boneskull: so it was agreed that `err.diff` is a *function*?  what's its signature?  can you point me to that discussion?  again, I'm having trouble following / understanding what exactly people want here.

test/integration/fixtures/diffs/output
Mingun: It is necessary for more natural message in diff. diff shows the reason for which `actual` does not correspond to `expected`. If the reason that at `actual` there is an excess element, then the reason shall show that this excess element is added, thus, I know that I need to find the place where the excess element is added and not to add it.  Thus, for 
* `expected = [1, 2]`
* `actual = [1]`

diff shall look so:
```diff
 1
-2
```
and for 
* `expected = [2]`
* `actual = [1, 2]`

diff shall look so:
```diff
+1
 2
```

Without this change everything was on the contrary that at least confuses me a little

lib/reporters/base.js
Mingun: This is add empty line after diff, just as in [old code], but use more high-level abstraction, which allow write similar code in the `inlineDiff` and `unifiedDiff` functions

[old code]: https://github.com/mochajs/mocha/blob/master/lib/reporters/base.js#L396

lib/reporters/base.js
Mingun: This prevent adding delimiters between hunks before first hunk

package.json
Mingun: Just to keep dependencies actual. The [changes] made between versions do not influence the functions used by us, but the fresh version of library will simplify the analysis of changes if such appear and they will influence our code

[changes]: https://github.com/kpdecker/jsdiff/compare/v3.3.1...v3.4.0

lib/reporters/base.js
Mingun: I don't see how it can be useful here?

lib/reporters/base.js
Mingun: Function is used because we need one of two formats for different types of diff as one of possible solutions for this. Perhaps, it will be possible to integrate them in one format, in that case function is not required and this field will be able to be a simple object in this format. In this case this it will be necessary to make utility functions for conversion of results of `diff` library to this universal format. Since in https://github.com/chaijs/chai/issues/1120 already expressed idea of object diff library, it can be done in there.

At the moment I only used existing formats which are provided by `diff` library that simplifies implementation of the user-defined diffs. Signature:
```js
/**
@param {boolean} inline Generate inline diffs if true, generate unified diff otherwise
@return {Object} The same as return value of `diff.structuredPatch` from the `diff` library if `inline` is falsely
                 The same as return value of `diff.diffWordsWithSpace` from the `diff` library if `inline` is truely
*/
function diff(inline) { ... }
```

However, if we are going to make a utility library for generation of diff between objects in the format suitable us

Mingun: Though use of universal format will give nothing -- for generation it is necessary to use different functions. As the choice of a format is carried out by mocha, and generation of diff by assertion library, it is necessary either to use function, or to generate directly 2 kinds of diff, which may be time expensive

Mingun:
In the commit f6d63ca525aa7c7f395608a7c31a22f2fb8c6779 I add support for per-character diffs in unified diffs. In principle, existence of such information allows to manage only one type of diff therefore I have a question and suggestion: can be to delete support of `inlineDiffs`? With a new opportunity it does not look the especially useful, and existence only of one format simplifies support and will allow the `err.diff` field to be not function, but a diff object.

In image below you can see possible result. Before tildas you can see new unified format, below -- existent inline format:
new-diff

There files on which diffs is generated:
* [expected.txt](https://github.com/mochajs/mocha/files/1796543/expected.txt)
* [actual.txt](https://github.com/mochajs/mocha/files/1796544/actual.txt)

Mingun:
I find a way to use one data structure for both kinds of diff and push changes soon. Thus, a data structure, I think, is defined. Other changes:
* The quantity of context lines is defined by the reporter. In the future this can be configurable if you wish
* Also I implement concealment of big pieces of not changed lines in the inline diff mode
* For inline diff line numbers now displayed correclty. Now it is not arbitrary values of the simple counter, but real line numbers, one for old and one for new version
* Also I enable line numbers for unified diff too. In both diff modes line numbers shows if line count `>= 4`
* In unified mode intra line diff is shown
* The word diff algorithm is tuned (pay attention to text near parentheses)

Also I completely rework a method by which diff is displayed on the screen and now it looks as:
new-diff-2
For comparing, the same files in the current implementation
old-diff

----
The next thing -- tests. There are any ideas about what needs to be tested?

package.json
```diff
@@ -309,12 +309,12 @@
     "browser-stdout": "1.3.1",
     "commander": "2.11.0",
     "debug": "3.1.0",
-    "diff": "3.5.0",
     "escape-string-regexp": "1.0.5",
     "glob": "7.1.2",
     "growl": "1.10.3",
     "he": "1.1.1",
     "mkdirp": "0.5.1",
+    "structured-diff": "Mingun/structured-diff",
```
Mingun: After the publication of a package I will replace the reference to a npm repository here

boneskull:
> The next thing -- tests. There are any ideas about what needs to be tested?

Everything that's changed, essentially.  Tests should run in both browser and Node.js.  You'll want unit tests for the basic functionality, then a handful of integration/functional tests to ensure that it enables the desired behavior.

boneskull:
Also, because we have so many open PRs, I'm going to close this because it's a work-in-progress.  **This does not mean I'm rejecting the PR**, but rather I prefer you hold it until it is "ready".  Feel free to add any further questions to the associated open issue(s).

Issue #3006, opened 2017-09-18
Support ES6 style tests without transpiler usage


SGD1953:
### Prerequisites

* [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
* [x] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
* [x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
* [x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
`node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description
Before I start, there are already some closed issues regarding this topic but as the prerequisites have changed I would like to start a new attempt.

Now that node supports running [EMCAScript modules](https://nodejs.org/dist/latest-v8.x/docs/api/esm.html) (yes, I know it is experimental) it would be great to see mocha to work in conjunction with `mjs` test definitions.

### Steps to Reproduce

I have a very simple test
```
describe('Test', function () {
});
```
Which i have saved as `test.js` and `test.mjs`

**Expected behavior:** I would like both tests to show
```
- test/test.js 
  0 passing (1ms)
(node:70422) ExperimentalWarning: The ESM module loader is experimental.
```

**Actual behavior:** While the `js` test works, the `mjs` test gives me
```
- test/test.mjs 
module.js:658
    throw new errors.Error('ERR_REQUIRE_ESM', filename);
    ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/dgehl/Repositories/LOreal/code/ecom-lora/test/frontend/components/global/test.mjs
```

**Reproduces how often:** 100%

### Versions
` node --version` - v8.5.0
`mocha --version` - 3.5.3

### Additional Information
I _think_ that this might be that mocha's runner is using commonjs and nodejs' current implementation disallows to use ECMAScript modules from a commonjs context.

Please don't reply with "use a transpiler", I want to explicitly not use one.

Edit: in an earlier version I accidentally used jsx instead of mjs.

ScottFreeCode:
[...]

PR #3253, opened 2018-02-26
WIP: use dynamic import to load es module tests.

harrysarson:
### Description of the Change

Uses the new (and as yet highly experimental) support in node for dynamically loading modules using `const obj = await import('./file-name.mjs');`.

I have added a new command line flag `--es-modules` which causes mocha to use `import()` instead of `require()` to load test files.

This allows tests to `import obj from './somefile.mjs'` without having to transpile tests before running them.

Obviously this only works in the very latest node version. This commit does not break mocha in `node <= 8` unless the user passes the `es-modules` flag.

Currently running a script in node with `--experimental-modules` fails if the file does not have an extension which means that `node --experimental-modules bin/_mocha` files. From my understanding (see https://github.com/nodejs/node/pull/18728) node v9.7 will be able to run scripts without the .js extension.

### Alternate Designs

Current design works fine as long as users are happy to use a transpiler.

Cannot be implemented outside of core.

### Benefits

Tests can be written as es modules.

### Possible Drawbacks

Current support for modules in node is experimental. It is likely that mocha will want to wait untill node starts properly supporting modules before merging this pr.

### Applicable issues

#3006 

Probably semver minor but could be major.


boneskull:
This seems to break AppVeyor's build, but why?

.travis.yml
```diff
@@ -14,17 +14,7 @@ matrix:
   fast_finish: true
   include:
     - node_js: '9'
-      env: TARGET=test.node COVERAGE=true
-    - node_js: '8'
```
boneskull: we can't just remove all of these from the build matrix...

bin/mocha
```diff
@@ -10,7 +10,7 @@
 const spawn = require('child_process').spawn;
 const path = require('path');
 const getOptions = require('./options');
-const args = [path.join(__dirname, '_mocha')];
+const args = [path.join(__dirname, '_mocha.js')];
```
boneskull: why is this necessary?

lib/mocha.js
```diff
@@ -234,6 +234,34 @@ Mocha.prototype.loadFiles = function (fn) {
   fn && fn();
 };
 
+/**
+ * Load registered files via es6 dynamic imports.
+ *
+ * Note: `eval` is used in this function or else mocha will not
+ * compile on older node versions.
```
boneskull: I don't understand where this is supposed to work.

lib/mocha.js
```diff
@@ -234,6 +234,34 @@ Mocha.prototype.loadFiles = function (fn) {
...
+ */
+Mocha.prototype.dynamicallyImportFiles = function (fn) {
+  var self = this;
+  var suite = this.suite;
+  Promise
+    .all(this.files.map(function (file) {
+      file = path.resolve(file);
+      suite.emit('pre-require', global, file, self);
+      return eval('import(file)')
+        .then(function (module) {
```
boneskull: probably shouldn't use `module`

lib/mocha.js
harrysarson: The new `import()` operator allows ECMAScript Modules to be loaded dynamically. The hope was that tests written as modules (i.e. tests including `import obj from './script.js'` statements) can be loaded into Mocha using dynamic imports without having to first use a transpiler to convert ECMAScript Modules into common js modules.

boneskull: right, but I'm unclear on why we're removing old versions of Node.js from the matrix.

> Current support for modules in node is experimental. It is likely that mocha will want to wait untill node starts properly supporting modules before merging this pr.

if we're going to break backwards compatibility, then it'd have to wait until the oldest version of Node.js that supports modules is no longer maintained.  that'll be a few years.  so, that's why I'm trying to understand what I'm supposed to do with this...

harrysarson: Sorry, I didn't mean to add that change to the pull request, I will revert it.

The idea was that mocha could add a new CLI flag which would cause mocha to use the new `import()` instead of `require()` to load test files.

This new CLI flag could only be used on the most recent versions of node but mocha would work normally if this flag was not set.

The tests fail because I added a test for the new flag and travis/appveyor runs this test on old versions of node, I didn't know how to work around that.

demurgos:
Why was this PR closed?

I am currently using Mocha together with `@std/esm` to test my .mjs files but this is suboptimal (edge cases around interop with cjs or code coverage). Now that dynamic imports are available it would be nice to be able to use ESM directly with Mocha.

boneskull:
Until modules are unflagged and/or no longer considered "experimental", I'm concerned about adopting something too soon.  Specifically, when there's agreement in core around interop between `require()` and `import`, filename extensions, etc.--then it's probably safe to move.  Too early, and it'll be expensive to fix any mistakes we made by jumping on it too soon.

demurgos:
Thanks for the reply: it's a fair reason.

I was worried that there was an issue with the PR preventing it from working. Still, it's good to know that it's a Node stability issue instead and that you are able to add support for it when possible.

harrysarson:
I was a bit optimistic with this PR, definately need to wait until modules are finalised in node.

The other problem, which is why my work was so hacky, is that the dynamic `import()` loads modules asynchronously whereas mocha currently loads tests synchronously (using `require()`) so it is not as simple as just switching one for the other.

demurgos:
Hi,
The modules discussion is still ongoing, but it seems that there's agreement that the imports will be async.

A possible way to prepare the support for ESM would be to force the current imports to be async by default: replace the resolution code from `require("test.spec.js")` to `Promise.resolve(require("test.spec.js"))`.
Replacing the inner `require` by `import` would then be trivial.

Looking at the PR, it seems that most of the code is already async. I am not familiar enough with mocha's codebase so maybe my suggestion is not needed. I just hope that Mocha will be able to support ESM when available.

harrysarson:
@demurgos from memory it should be fairly straightforward to make that change

Issue #3223, opened 2018-01-29
can wonky reporting and statistics be avoided?


boneskull:
If an task created by a test would throw an exception (or rejection) *and* this test has already passed or failed, then Mocha's output gets wonky (using `spec` reporter, for example):

- the test name may repeat multiple times, with different results for each
- we can't just limit it, because suppression thereof means errors can't easily be cross-referenced by number
- the total pass/fail/pending count will be larger than the total number of tests executed, because a test will have been reported to be in multiple states
- multiple errors may be output for a single test

* * *

Here are two ideas to mitigate this:

1.  Decouple any exception a test causes *after* it has completed from the test itself, for reporting purposes.  Such errors would simply *not* count towards statistics, but would display nevertheless.  The *test*, then, would not necessarily fail (unless it did originally), but the *test run* would fail.
    This could be significantly disruptive to the reporting system.  Simply swapping a reporter out for another should not change the results!

2.  Determine, by way of async hooks or otherwise, when a test is truly "complete".  Don't record the test result until there are no further tasks in the queue.
    Difficult.  May slow down some tests significantly.  Would need to cancel the timeout when `done()` is called (or whatever); might want to record the delta between when the test finishes and when the test's tasks complete.  Consuming async hooks / zones / realms etc. could either be great (eliminate many edge and corner cases) or terrible (increased complexity and black magic).

There's no end to problems from these types of errors.  They shatter Mocha's assumptions about the current state of things; Mocha is trying to run tests serially, after all.  Test B should never seem to fail because of something Test A did; and even so, we should know that it was Test A's fault.  

* * *

Here's a philosophical question:

Can Mocha rely *on the user* to tell Mocha when a test completes, and how it completed?

Nokel81:
This also seems quite interesting and does seem to be related to #3256. Re the philosophical question would it be possible to tell what async hooks were sent off within a test? If so could it be used that if the test has "completed" by normal means, ie done() has been called or if the test wasn't told to be async then the function completed, but there are still async hooks present then fail the test with a suitable error message saying that an async function did not complete before the test completed?

ctaylo21:
I know this is a tricky issue, but there might be some potential insight someone can offer. Is there a way to debug these sort of errors? We have a tricky race condition that will sometimes cause a test to fail after it. In the spec reporter, it will mark the test as both passed and failed.

[...]

PR #3257, opened 2018-03-01
do not eat exceptions thrown asynchronously from passed tests; closes #3226

### LINKED ISSUE #3226: Errors after test completion are swallowed
### Prerequisites

- [X] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description

Sometimes tests are written incorrectly and throw errors _after_ they're complete.  This applies to all three of Mocha's test forms (sync, promise, callback).  If you write this sort of code in a normal Node program then the errors usually bubble up and kill the Node event loop, an error message is printed, and the process exits with a non-zero code.  However when run with Mocha these errors are completely swallowed and never appear anywhere, and the process exits with code 0.

Obviously these tests are written badly, but I would expect Mocha to at least exit with an error code if a developer makes this sort of mistake!  I can see how it may be difficult or impossible for Mocha to figure out which test caused the error, but it shouldn't just swallow the error completely.

### Steps to Reproduce

```Javascript
describe('error after test complete', function() {

    it('sync', function() {
        setImmediate(() => {
            throw new Error('oops');
        });
    });

    it('promise', function() {
        setTimeout(() => {
            throw new Error('oops');
        }, 1000);
        return Promise.resolve();
    });

    it('callback', function(done) {
        setImmediate(() => {
            done();
            throw new Error('oops');
        });
    });
});
```

**Expected behavior:**

At minimum Mocha executable should fail with a non-zero error code and print out some sort of error message.  Ideally Mocha would also be able to associate the error with the test that caused it, but I realize that may not be possible.

**Actual behavior:**

Tests pass, Mocha exits with code zero.
```
  error after test complete
    ✓ sync
    ✓ promise
    ✓ callback


  3 passing (16ms)
```

### Versions

Tested with Mocha 5.0

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 5.0.0
- The output of `node --version`: v8.9.3
- The version and architecture of your operating system: Ubuntu 16.04, x86 64
- Your browser and version (if running browser tests): n/a
- Any other third party Mocha related modules (with versions): n/a
- The code transpiler being used: none

boneskull:
Currently, Mocha eats the error thrown below, because the test has already passed:

```js
it('should bail if a successful test asynchronously fails', function(done) {
  done();
    throw new Error('global error');
  });
});
```

If the above happens, Mocha can't recover gracefully b/c of #3223.  This PR causes Mocha to bail if the above situation is encountered.  Exactly *how* this is reported is "best effort", based on queued tasks and timing--the reporter output may not be pretty, but you *should* at least get a traceback.

One thing we *can* do to mitigate #3223-related weirdness is force the reporters to only respond to the `end` event *once*, which is what I've done.  Without this, machine-readable reporters in particular--such as `json` or `xunit`--get their wires crossed.  No reporter should be dumping the summary twice anyway.

Also added a couple helper functions to `Runnable.prototype`.


Bamieh:
Looks good, I love the usage of `once` instead of `on`.

boneskull:
> Looks good, I love the usage of once instead of on.

yeah, but it's actually a hack around the real problem, which is that `end` shouldn't be emitted more than once.

Bamieh:
@boneskull regardless if it is triggered once or multiple times, we should be using `once` to automatically detach the event listener.

rsheldiii:
On upgrading from mocha 3.5.3 to 5.2.0 (latest), we noticed that our test coverage started silently skipping a bunch of tests while still passing. I bisected and found that this commit is the commit that causes it. I'm aware now that this is because one or more of our tests are throwing an error in asynchronous code, possibly after they pass, but we've got quite a few tests, so it's going to take a while to pinpoint. For now we're downgrading to 4.1.0, which still runs the entire suite. Just want to leave in case it helps someone

plroebuck:
@rsheldiii, two choices:
* Run Mocha-5.2+ with `--exit` cmdline option which provides pre-4.0+ exit behavior

```bash
$ mocha --exit my-shitty-code.spec.js 
```

* Find and fix problems in your test suite. The [wtfnode](https://www.npmjs.com/package/wtfnode) package can help with this. 

```bash
$ wtfnode ./node_modules/.bin/_mocha my-shitty-code.spec.js 
```

plroebuck:
Also note that the change to `.once` for the reporter event handler also prevents `mocha` instance from being reusable for those doing this programmatically. We need another method to recreate the internal pieces/parts (e.g., reporters, growl) that have event listeners using `.once`.

Issue #3260, opened 2018-03-01
better test harness for reporters


boneskull:
There's a lot of duplicated code in the reporter tests.  Let's extract that into helper modules/functions/etc.

jMuzsik:
Hi! I'd be willing to do this

Bamieh:
@jMuzsik Hello!, thanks for your interest to contribute 🎉. Just say you're going to work on it and start. Here is how it generally works in all OSS; You fork the project and create a new branch from master with a link to the issue number you are working as a commit message and push it before writing any code. This will link your branch to this issue, letting people know that you are working on it and follow on the progress. If you need comments on the code or your branch is ready, submit a PR and we'll have a look at it to get it merged!

Cheers

PR #3272, opened 2018-03-08
Refactor Reporter tests

jMuzsik:
### Description of the Change

In reference to issue: #3260 

I did my best at refactoring the code. Though the part of the test files that was quite obviously capable to become more modular was the runner events. I could make all of those calls from one function, ie:

```javascript
runner.on = runner.once = function...
// into
runner.on = runner.once = runnerEvent(whatevery necessary parameters);
```

Otherwise I tried to make variables that I noticed were repeatedly called throughout singular files to be called only one time rather then the continual `var something = 'something'` over and over.

### Alternate Designs

I believe that it can still be worked on. It is not perfect. The `runnerEvent` function could be more beautiful/intuitive. Some tests can make more sense. BaseReporter tests run but it is not obvious, I have no idea how i'd implement that.

Quite a bit less lines of code for the same functionality. 

### Benefits

Less intensive use of memory and variable calls, etc. More modular code.

### Possible Drawbacks

I had to alter quite a bit of code that a lot of people put lots of time into. There are no errors atm, and I expect there not to be.

### Applicable issues

Not that I know of

Open to doing more if there is a recommendation.

### DIFF
--- a/test/reporters/doc.spec.js
+++ b/test/reporters/doc.spec.js
@@ -3,7 +3,7 @@
 var reporters = require('../../').reporters;
 var Doc = reporters.Doc;
 
-var runnerEvent = require('./helpers.js').runnerEvent;
+var createMockRunner = require('./helpers.js').createMockRunner;
 
 describe('Doc reporter', function () {
   var stdout;
@@ -11,7 +11,6 @@ describe('Doc reporter', function () {
   var runner;
   beforeEach(function () {
     stdout = [];
-    runner = {};
     stdoutWrite = process.stdout.write;
     process.stdout.write = function (string) {
       stdout.push(string);
@@ -27,7 +26,7 @@ describe('Doc reporter', function () {
         title: expectedTitle
       };
       it('should log html with indents and expected title', function () {
-        runner.on = runner.once = runnerEvent('suite', 'suite', null, null, suite);
+        runner = createMockRunner('suite', 'suite', null, null, suite);
         Doc.call(this, runner);
         process.stdout.write = stdoutWrite;
         var expectedArray = [
@@ -43,7 +42,7 @@ describe('Doc reporter', function () {
           title: unescapedTitle
         };
         expectedTitle = '&#x3C;div&#x3E;' + expectedTitle + '&#x3C;/div&#x3E;';
-        runner.on = runner.once = runnerEvent('suite', 'suite', null, null, suite);
+        runner = createMockRunner('suite', 'suite', null, null, suite);
         Doc.call(this, runner);
         process.stdout.write = stdoutWrite;
         var expectedArray = [
@@ -59,7 +58,7 @@ describe('Doc reporter', function () {
         root: true
       };
       it('should not log any html', function () {
-        runner.on = runner.once = runnerEvent('suite', 'suite', null, null, suite);
+        runner = createMockRunner('suite', 'suite', null, null, suite);
         Doc.call(this, runner);
         process.stdout.write = stdoutWrite;
         expect(stdout).to.be.empty();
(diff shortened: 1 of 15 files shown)


boneskull:
great effort, thanks.  I'll review this when I get a minute

test/reporters/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+        callback(test, {});
+      }
+      if (event === ifStr2) {
+        callback(test);
+      }
+      if (event === ifStr3) {
+        callback(test);
+      }
+    };
+  }
```
outsideris: How about throwing an exception here to prevent typo or new runStr?

test/reporters/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+*/
+function runnerEvent (runStr, ifStr1, ifStr2, ifStr3, arg1, arg2) {
+  var test = null;
+  if (runStr === 'start' || runStr === 'pending' || runStr === 'end') {
+    return function (event, callback) {
+      if (event === ifStr1) {
+        callback();
+      }
+    };
+  } else if (runStr === 'pending test' || runStr === 'pass' || runStr === 'fail' || runStr === 'end' || runStr === 'suite' || runStr === 'suite end' || runStr === 'test end') {
```
outsideris: It's hard to read and makes a verbose diff when we add a new case.
What do you think about format it like:

```
} else if (
  runStr === "pending test" ||
  runStr === "pass" ||
  runStr === "fail" ||
  runStr === "end" ||
  runStr === "suite" ||
  runStr === "suite end" ||
  runStr === "test end"
) {
```

test/reporters/doc.spec.js
```diff
@@ -3,6 +3,8 @@
 var reporters = require('../../').reporters;
 var Doc = reporters.Doc;
 
+var runnerEvent = require('./helpers.js').runnerEvent;
```
boneskull: I'm not sure what to call this, but `runnerEvent` isn't descriptive enough.

Also, I'm not sure `runner` needs to be created as an empty object in `beforeEach`...

Instead, maybe this could just be a factory function for a `runner` object?  It would then create `.on` and `.once` as appropriate.

Then call it something like `createMockRunner`

test/reporters/helpers.js
```diff
@@ -0,0 +1,139 @@
...
+
+/*
+  This function prevents the constant use of creating a runnerEvent.
+  runStr is the argument that solely defines the runnerEvent.
+  ifStr1 is one possible reporter argument, as is ifStr2, and ifStr3
+  arg1 and arg2 are the possible variables that need to be put into the scope of this function for the tests to run properly.
+*/
+function runnerEvent (runStr, ifStr1, ifStr2, ifStr3, arg1, arg2) {
+  var test = null;
+  if (runStr === 'start' || runStr === 'pending' || runStr === 'end') {
```
boneskull: don't be afraid to use a `switch` in here, or an object lookup.

test/reporters/helpers.js
```diff
@@ -0,0 +1,139 @@
...
+    err: err,
+    titlePath: function () {
+      return ['test title'];
+    }
+  };
+}
+
+function createElements (argObj) {
+  var res = [];
+  for (var i = argObj.from; i <= argObj.to; i += 1) {
```
boneskull: please use `i++`

test/reporters/json-stream.spec.js
```diff
@@ -3,27 +3,46 @@
...
+
   beforeEach(function () {
     stdout = [];
     runner = {};
     stdoutWrite = process.stdout.write;
     process.stdout.write = function (string) {
       stdout.push(string);
     };
+
+    expectedTitle = 'some title';
```
boneskull: It doesn't make a lot of sense to declare all this above then define it here, does it?  Can we not just define it all above?

test/reporters/tap.spec.js
```diff
@@ -46,21 +53,12 @@ describe('TAP reporter', function () {
...
-      };
-      runner.on = runner.once = function (event, callback) {
-        if (event === 'test end') {
-          callback();
-        }
-        if (event === 'pending') {
-          callback(test);
-        }
-      };
+      // var test = {
```
boneskull: what's this?

boneskull:
@jMuzsik LGTM!  Thanks a ton.

PR #3273, opened 2018-03-08
append filepath to "timeout exceeded" exception; closes #627

### LINKED ISSUE #627: Add test file location on timeout exceeded test failure
This is a problem when running a directory of multiple tests.

Test failures on assertion or exception give the correct file from the stack trace.

A timeout however leaves us grepping for the correct file:

```
  2) lophilo watch watching...:
     Error: timeout of 2000ms exceeded
      at Object.<anonymous> (/home/rngadam/local/node/lib/node_modules/mocha/lib/runnable.js:158:14)
      at Timer.list.ontimeout (timers.js:101:19
```

boneskull:
- all `Runnable`s *should* now have a `file` property (in Node.js)
- filepath is appended to the `Error` message in parens
- DRY-style refactors

* * *

### Alternate Designs

this could be formatted differently.

Because I wanted to send a PR for the oldest issue I could find

### Benefits

Somebody wanted this, once upon a time

### Possible Drawbacks

Any machine trying to parse these errors may run into difficulties

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -252,8 +252,7 @@ Runnable.prototype.resetTimeout = function () {
     if (!self._enableTimeouts) {
       return;
     }
-    self.callback(new Error('Timeout of ' + ms +
-      'ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.'));
+    self.callback(self._timeoutError(ms));
     self.timedOut = true;
   }, ms);
 };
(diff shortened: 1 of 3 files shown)


boneskull:
Looks like I'll need to address the coverage issues.  Wonder what I did wrong here...

boneskull:
If we don't count comments, I actually reduced LoC, in essence the coverage remains the same.  The percentage drops because there are simply fewer LoC to test.

boneskull:
The current coverage drop:

**Coverage decreased (-0.04%) to 89.991%**

PR #3276, opened 2018-03-09
fix(changelog): fix links to some PRs

tagoro9:
### Description of the Change

I was checking the changelog of the latest versions and realized that some PR links had the wrong URL. They were `https://github.com/mochajs/mocha/pulls/:id` and they should be `.../pull/:id`.

This PR fixes those broken links.


outsideris:
Good catch.

boneskull:
Thanks!

PR #3278, opened 2018-03-11
fix to bail option works properly with hooks

### LINKED ISSUE #3096: Bail with failing after() runs suite twice
### Prerequisites

### Description
With a failing test suite using the `bail` option in conjuncture with an `after` clause that throws an error.
Mocha bails properly, executes the after as expected.
Then it does it all again. [But Why?](https://media.giphy.com/media/s239QJIh56sRW/giphy.gif)

### Steps to Reproduce
Very simple repro steps.
File new with latest mocha.
```  
"scripts": {
    "test": "./node_modules/.bin/mocha --reporter spec --bail"
  }
```
and
```
describe("some suite", function() {
    it("does the first test", function() {
        throw new Error('first error')
    });
    after(() => {
        throw new Error('after error')
    })
})
```
yarn test, observe double suite run
remove error in after, observe one suite run

test suite is only ran once
test suite is ran twice
100%

### Versions

node node_modules/.bin/mocha --version => `4.0.1`
macOS Sierra 10.12.6
zsh

outsideris:
### Requirements

### Description of the Change
With bail option, all tests must be stopped after first test failure. However, bail option doesn't stop hooks such as `after` or `afterEach`. 

### Alternate Designs
I have no idea.

It's a bug. It makes bail works correctly.

### Benefits
A user can use bail intentionally.

### Possible Drawbacks
No.

### Applicable issues
Fix  #3096 .
It is patch release.

### DIFF
--- a/test/integration/fixtures/options/bail-with-after.fixture.js
+++ b/test/integration/fixtures/options/bail-with-after.fixture.js
@@ -1,0 +1,11 @@
+'use strict';
+
+describe('suite1', function () {
+  it('should only display this error', function () {
+    throw new Error('this should be displayed');
+  });
+
+  after(function () {
+    throw new Error('this hook should not be displayed');
+  });
+});
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -265,10 +265,10 @@ Runner.prototype.failHook = function (hook, err) {
     hook.title = hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
   }
 
-  this.fail(hook, err);
   if (this.suite.bail()) {
     this.emit('end');
   }
+  this.fail(hook, err);
 };
 
 /**
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -68,6 +68,22 @@ describe('options', function () {
         done();
       });
     });
+
+    it('should stop all hooks after the first error', function (done) {
+      run('options/bail-with-after.fixture.js', args, function (err, res) {
+        if (err) {
+          done(err);
+          return;
+        }
+        assert.equal(res.stats.pending, 0);
+        assert.equal(res.stats.passes, 0);
+        assert.equal(res.stats.failures, 1);
+
+        assert.equal(res.failures[0].title, 'should only display this error');
+        assert.equal(res.code, 1);
+        done();
+      });
+    });
   });
 
   describe('--sort', function () {


test/integration/fixtures/options/bail-with-after.fixture.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+describe('suite1', function () {
+  it('should only display this error', function (done) {
```
boneskull: please remove the `done` parameter in this function and below; they aren't necessary for what these tests are proving.

outsideris:
I removed `done` parameter. I will merge this after confirm that CI passed.

plroebuck:
IMO, this should be backed out. Believe it's completely wrong.

boneskull:
Why?

boneskull:
bail should call those events you mentioned. if it’s not, I agree it’s wrong. past that, whether more hooks should run is a matter of opinion. some would prefer an immediate process exit with exception, others would like cleanup. but fwiw I’m feeling wishy-washy on it

plroebuck:
Think:
* if either `before`/`beforeEach` fail, don't run test and die.
* if test fails, `afterEach`/`after` should be run, then die.

If we manage to do the setup, we should be responsible for the teardown.

boneskull:
Was this fix to #3096 overkill, then?  @outsideris?

plroebuck:
IMO, it definitely doesn't seem to be `semver-patch` either.
The "fix" has **major** ramifications to event logic.

plroebuck:
> However, I still didn't understand why people want to clean up them when they run Mocha with bail.

If you launch a background server (in setup), it will be using a port. If you terminate without shutting the server down, that port will still be in use. Now you can't trivially restart your test without going back to shell to determine the server pid to kill it manually. Would make `--watch` (or your nodemon equivalent) rather cumbersome to use, no?

outsideris:
Why the case which you provide is handled by `--bail`? Without `--bail`, there is no problem. 
I think a purpose of `--bail` is prechecking for all tests are passed before deployment or publishing like there is no reason to proceed further if there is a failure already.
Please tell me If I missed something. I don't use `--bail` heavily.

plroebuck:
I don't use `--bail` much either. But I can definitely see it potentially being used with `--watch` during refactoring.

outsideris:
Users of two cases, only `--bail` and `--bail` with `--watch` have different needs. I guess former case want to stop running at first exception and latter case want to run clean up.
I'm not sure how we handle both cases gracefully in `--bail.

PR #3281, opened 2018-03-12
Ensure all after hooks are executed, even if one of them fail

johanblumenberg:
### Description of the Change

If you have tests that require some cleanup, for example to close open files, network connections, processes, or other things, you usually have an `after()` or `afterEach()` hook to do that.

```js
describe('suite A', function () {
  afterEach(function () {
    cleanupItemX();
    cleanupItemY();
  });

  .
  .
  .
});
```

Now, if there is an exception from the first cleanup method, the second one will not be run.

This PR enables you to write cleanup code like this:

```js
describe('suite A', function () {
  afterEach(function () {
    cleanupItemX();
  });

  afterEach(function () {
    cleanupItemY();
  });

  .
  .
  .
});
```

This will make sure that both hooks are always executed, and errors are reported properly if any of them fail.
Before this PR, the second after each hook was not executed if the first one throws an exception. But the fact that hooks in the parent suite are executed leads me to believe that this was simply overlooked, maybe because several after hooks in the same suite is not very common.

### Alternate Designs

It would be possible to fix this by adding a bunch of try/finally blocks. However, there are some problems with that.
 - The code is cluttered with a bunch of try/finally blocks
 - Only one error is reported from the test suite. You want to log all errors that happen.

```js
describe('suite A', function () {
  afterEach(function () {
    try {
      cleanupItemX();
    } finally {
      try {
        cleanupItemY();
      } finally {
        cleanupItemZ();
      }
    }
  });

  .
  .
  .
});
```

Because there is an inconsistency in how after hooks work. Hooks in the same suite are not run if the first one fails, but hooks in parent suites are run. To be consistent, it should work the same for all hooks.

### Benefits

Consistency in how hooks work, and possibility to write less try/finally clauses in the after hooks.

### Possible Drawbacks

Cannot see any drawbacks

### Applicable issues

I say this is a patch release, because it only affects what is executed when something fails. Also you should not be surprised that your after hook is executed when something fails, because that is the purpose of having it.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -318,6 +318,7 @@ Runnable.prototype.run = function (fn) {
     self.clearTimeout();
     self.duration = new Date() - start;
     finished = true;
+    delete self.callback;
     if (!err && self.duration > ms && self._enableTimeouts) {
       err = new Error('Timeout of ' + ms +
       'ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.');
@@ -325,7 +326,7 @@ Runnable.prototype.run = function (fn) {
     fn(err);
   }
 
-  // for .resetTimeout()
+  // for .resetTimeout() and uncaught errors
   this.callback = done;
 
   // explicit async with `done` argument
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -284,10 +284,10 @@ Runner.prototype.hook = function (name, fn) {
   var hooks = suite['_' + name];
   var self = this;
 
-  function next (i) {
+  function next (i, prevErr) {
     var hook = hooks[i];
     if (!hook) {
-      return fn();
+      return fn(prevErr);
     }
     self.currentRunnable = hook;
 
@@ -320,13 +320,18 @@ Runner.prototype.hook = function (name, fn) {
         } else {
           self.failHook(hook, err);
 
-          // stop executing hooks, notify callee of hook err
-          return fn(err);
+          if (name === 'beforeAll' || name === 'beforeEach') {
+            // stop executing hooks, notify callee of hook err
+            return fn(err);
+          } else {
+            // continue calling all after and afterEach hooks
+            prevErr = err;
+          }
         }
       }
       self.emit('hook end', hook);
       delete hook.ctx.currentTest;
-      next(++i);
+      next(++i, prevErr);
     });
   }
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -724,8 +724,6 @@ Runner.prototype.uncaught = function (err) {
     return;
   }
 
-  runnable.clearTimeout();
-
   // Ignore errors if already failed or pending
   // See #3226
   if (runnable.isFailed() || runnable.isPending()) {
@@ -733,33 +731,15 @@ Runner.prototype.uncaught = function (err) {
   }
   // we cannot recover gracefully if a Runnable has already passed
   // then fails asynchronously
-  var alreadyPassed = runnable.isPassed();
-  // this will change the state to "failed" regardless of the current value
-  this.fail(runnable, err);
-  if (!alreadyPassed) {
-    // recover from test
-    if (runnable.type === 'test') {
-      this.emit('test end', runnable);
-      this.hookUp('afterEach', this.next);
-      return;
-    }
+  if (runnable.isPassed() || !runnable.callback) {
+    // this will change the state to "failed" regardless of the current value
+    this.fail(runnable, err);
 
-    // recover from hooks
-    var errSuite = this.suite;
-    // if hook failure is in afterEach block
-    if (runnable.fullTitle().indexOf('after each') > -1) {
-      return this.hookErr(err, errSuite, true);
-    }
-    // if hook failure is in beforeEach block
-    if (runnable.fullTitle().indexOf('before each') > -1) {
-      return this.hookErr(err, errSuite, false);
-    }
-    // if hook failure is in after or before blocks
-    return this.nextSuite(errSuite);
+    // bail
+    this.emit('end');
+  } else {
+    runnable.callback(err);
   }
-
-  // bail
-  this.emit('end');
 };
 
 /**
--- a/test/integration/uncaught.spec.js
+++ b/test/integration/uncaught.spec.js
@@ -16,7 +16,7 @@ describe('uncaught exceptions', function () {
       assert.equal(res.stats.failures, 1);
 
       assert.equal(res.failures[0].fullTitle,
-        'uncaught "before each" hook');
+        'uncaught "before each" hook for "test"');
       assert.equal(res.code, 1);
       done();
     });
--- a/test/integration/fixtures/hooks/after-hook-async-error.fixture.js
+++ b/test/integration/fixtures/hooks/after-hook-async-error.fixture.js
@@ -19,3 +19,18 @@ describe('spec 2', function () {
     console.log('test 3');
   });
 });
+describe('spec 3', function () {
+  after(function (done) {
+    console.log('after 1');
+    process.nextTick(function () {
+      throw new Error('after hook error');
+    });
+  });
+  after(function (done) {
+    console.log('after 2');
+    done();
+  });
+  it('should call all after hooks also if one fails', function () {
+    console.log('test 4');
+  });
+});
--- a/test/integration/fixtures/hooks/after-hook-error.fixture.js
+++ b/test/integration/fixtures/hooks/after-hook-error.fixture.js
@@ -17,3 +17,15 @@ describe('spec 2', function () {
     console.log('test 3');
   });
 });
+describe('spec 3', function () {
+  after(function () {
+    console.log('after 1');
+    throw new Error('after hook error');
+  });
+  after(function () {
+    console.log('after 2');
+  });
+  it('should call all after hooks also if one fails', function () {
+    console.log('test 4');
+  });
+});
--- a/test/integration/fixtures/hooks/afterEach-hook-async-error.fixture.js
+++ b/test/integration/fixtures/hooks/afterEach-hook-async-error.fixture.js
@@ -19,3 +19,18 @@ describe('spec 2', function () {
     console.log('test 3');
   });
 });
+describe('spec 3', function () {
+  afterEach(function (done) {
+    console.log('after each 1');
+    process.nextTick(function () {
+      throw new Error('after each hook error');
+    });
+  });
+  afterEach(function (done) {
+    console.log('after each 2');
+    done();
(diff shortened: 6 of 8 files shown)


boneskull:
@johanblumenberg Thanks.  Can you please sign the CLA?

johanblumenberg:
> @johanblumenberg Thanks. Can you please sign the CLA?

Done

lib/runnable.js
```diff
@@ -318,14 +318,15 @@ Runnable.prototype.run = function (fn) {
     self.clearTimeout();
     self.duration = new Date() - start;
     finished = true;
+    delete self.callback;
```
boneskull: what's this for?

lib/runner.js
```diff
@@ -724,42 +729,22 @@ Runner.prototype.uncaught = function (err) {
...
-  // this will change the state to "failed" regardless of the current value
-  this.fail(runnable, err);
-  if (!alreadyPassed) {
-    // recover from test
-    if (runnable.type === 'test') {
-      this.emit('test end', runnable);
-      this.hookUp('afterEach', this.next);
-      return;
-    }
+  if (runnable.isPassed() || !runnable.callback) {
```
boneskull: I'm unclear what we're doing with the `callback`

lib/runner.js
```diff
@@ -320,13 +320,18 @@ Runner.prototype.hook = function (name, fn) {
         } else {
           self.failHook(hook, err);
 
-          // stop executing hooks, notify callee of hook err
-          return fn(err);
+          if (name === 'beforeAll' || name === 'beforeEach') {
+            // stop executing hooks, notify callee of hook err
+            return fn(err);
+          } else {
```
boneskull: this `else` is not necessary

test/integration/uncaught.spec.js
```diff
@@ -16,7 +16,7 @@ describe('uncaught exceptions', function () {
       assert.equal(res.stats.failures, 1);
 
       assert.equal(res.failures[0].fullTitle,
-        'uncaught "before each" hook');
+        'uncaught "before each" hook for "test"');
```
boneskull: where is this coming from?

lib/runner.js
```diff
@@ -320,13 +320,18 @@ Runner.prototype.hook = function (name, fn) {
...
+            // continue calling all after and afterEach hooks
+            prevErr = err;
+          }
         }
       }
       self.emit('hook end', hook);
       delete hook.ctx.currentTest;
-      next(++i);
+      next(++i, prevErr);
```
boneskull: so I understand: we're trapping the first error we find, then passing it along to the rest of the `after` and `afterEach` hooks?  then, if there's no hooks remaining, we finally bail?

lib/runner.js
```diff
@@ -724,42 +729,22 @@ Runner.prototype.uncaught = function (err) {
...
   // See #3226
   if (runnable.isFailed() || runnable.isPending()) {
     return;
   }
   // we cannot recover gracefully if a Runnable has already passed
   // then fails asynchronously
-  var alreadyPassed = runnable.isPassed();
-  // this will change the state to "failed" regardless of the current value
-  this.fail(runnable, err);
-  if (!alreadyPassed) {
```
boneskull: why was all of this removed?

boneskull:
FWIW:  I'm unsure if we want to merge this. 

To some of your points:

> Only one error is reported from the test suite. You want to log all errors that happen.

The user can log via `console.error`, etc.  Currently, Mocha does not support "multiple" exceptions in any context.  For example, if a test fails *then fails again*, Mocha swallows the second exception because there's no provision to handle such a state in the reporters:

```js
it('fails twice', function (done) {
  process.nextTick(() => {
    done(new Error('once'));
    process.nextTick(() => {
      throw new Error('twice');
    });
  });
});
```

See #3223 for further information.

> But the fact that hooks in the parent suite are executed leads me to believe that this was simply overlooked, maybe because several after hooks in the same suite is not very common.

So, to confirm, you're saying:

```js
describe('parent', function () {
  afterEach(function () {
    // this gets run even after 'accck' is thrown?
  });
 
  describe('child', function () {
    afterEach(function () {
      throw new Error('accck!');
    });
   
    it('does stuff', function () {
    });
  }); 
});
```

plroebuck:
This would seem to change the symmetry expectations of Mocha hooks. You only want to do this for after()/afterEach(), but not before()/beforeEach(), right?

Seems like you could just wrap the portions of your own code that might throw in a try/catch, and this problem goes away without changes to Mocha.

```js
const _ = require('lodash');
const errors = [];

function justKeepSwimming(fn) {
  let maybeError = _.attempt(fn);
  if (_.isError(maybeError)) {
    errors.push(maybeError);
  }
}

function logErrorsAndReset() {
  if (errors.length > 0) {
    errors.forEach((err) => console.error(err));
    errors.length = 0;
  }
}

describe('suite A', function () {
  afterEach(function () {
    justKeepSwimming(cleanupItemX);
    justKeepSwimming(cleanupItemY);
  });

  afterEach(function () {
    logErrorsAndReset();
  });

  // tests...
});
```

boneskull:
I'm calling this inactive.  please rebase/reopen if interested

lib/runnable.js
johanblumenberg: This is to ensure the `done()` function is only invoked once for a test

lib/runner.js
johanblumenberg: The callback is what is to be executed if the test fails. It's the `done()` function of the current test.

If there is a current test, then that test should be cancelled and all after hooks should be run on any uncaught exceptions.
If there is no current test, then uncaught exceptions simply mark the current suite as failed.

lib/runner.js
johanblumenberg: ok, it can be removed

test/integration/uncaught.spec.js
johanblumenberg: It's just a bonus improvement, that came for free when improving the handling of uncaught exceptions, since it now invokes the current test callback.

It's better to have the error message contain which test that failed.

lib/runner.js
johanblumenberg: yes

The prevErr is just to keep track of the last error that happened. The first after hook might throw an error, and the next ones succeed. Then the error has to be remembered so it can be used when calling the callback fn.

lib/runner.js
johanblumenberg: All this just another implementation of what is already in the tests `done()` function. So it's duplicate code.

It's trying to figure out which hook is being executed, and take certain actions depending on that. All that is already done, in a better way, in the tests `done()` function.

johanblumenberg:
@boneskull 

> So, to confirm, you're saying:
> // this gets run even after 'accck' is thrown?

yes


>> Currently, Mocha does not support "multiple" exceptions in any context. For example, if a test fails then fails again, Mocha swallows the second exception because there's no provision to handle such a state in the reporters:

In the example you provided, the test is completed when the `done()` function is called. All after hooks are run, and the next test is executed. The second exception, in the next tick, will be an unhandled exception, caught outside of any test. It will fail the current test suite. No exceptions are swallowed, they are all reported. I'm not changing any of that. And I think that is correct. If any code has a side effect that an exception is thrown asynchronously, that should still fail the test (if possible) or the test suite, so I notice the bug.

johanblumenberg:
@plroebuck 

> This would seem to change the symmetry expectations of Mocha hooks. You only want to do this for after()/afterEach(), but not before()/beforeEach(), right?

Yes

It cannot be done for the before hooks, for the same reason that tests cannot be run if a before hook fails. There is no point in continuing execution after an exception in the before hooks. But you need to make sure that you do cleanup of whatever you were setting up.

> Seems like you could just wrap the portions of your own code that might throw in a try/catch, and this problem goes away without changes to Mocha.

Problems with this code:
 - It's a lot more noisy
 - It doesn't handle asynchronous cleanup functions
 - It's logging the failures, but the test suite will still succeed, and no one will notice the errors
 - Only if you are looking at stdout will you see the errors, no reporters will be aware of the errors

johanblumenberg:
I would like this ticket to be reopened, however I cannot find the button to do so

johanblumenberg:
I'm not sure why this has to be a major semver update.

The only change is that all after hooks are run on test failures.
If tests are passing, there is no change.
If tests are failing, then possibly another after hook will be run. But since the tests are already failing, this won't affect the outcome of the test run.

outsideris:
Reopen button is also disabled for me. You should open it as new pull request if you want.

boneskull:
> If tests are failing, then possibly another after hook will be run. But since the tests are already failing, this won't affect the outcome of the test run.

in that case, it doesn't  necessarily need to be a semver-major, but I prefer to err on the side of caution with anything that fundamentally changes test execution.

plroebuck:
@johanblumenberg

> @plroebuck
> 
> > Seems like you could just wrap the portions of your own code that might throw in a try/catch, and this problem goes away without changes to Mocha.
> 
> Problems with this code:
>     * It's a lot more noisy
>     * It doesn't handle asynchronous cleanup functions
>     * It's logging the failures, but the test suite will still succeed, and no one will notice the errors
>     * Only if you are looking at stdout will you see the errors, no reporters will be aware of the errors

1. No argument there.
2. I only spent about 10 mins on it. But I could add a `done` argument to the `afterEach()`...
3/4. See below...

```js
function logErrorsAndFail() {
  if (errors.length > 0) {
    var err;
    if (errors.length === 1) {
      err = errors.shift();
    } else {
      errors.forEach((err) => console.error(err));
      err = new Error('multiple errors occurred... see console');
    }
    this.test.error(err);
  }
}
```

johanblumenberg:
@plroebuck I don't see how asynchronous errors would be handled in the code you provided.

You would need to provide a `done` argument. You would also need to check each cleanup function to see if it is a synchronous or asynchronous function, based on if it returns a promise or not. If it does return a promise, you need to wait with executing the next cleanup function, until the first one is done. Then, when all cleanup functions are done, you need to invoke the `done` callback.

This would be possible, of course, but it is a lot of code to write. It also puts the responsibility of making this work properly on the test writer, not on the test framework.

Also, all this code already exists in mocha. Why not leverage the existing code? Why require the user of the library to write the same code? Write it once, and it will work for everyone.

johanblumenberg:
> 4\. Same

@plroebuck I don't see how test reporters would be aware of the exceptions thrown. All they would receive would be the exception `new Error('multiple errors occurred during teardown... see console')`, no?

johanblumenberg:
There is also the inconsistency that if I put after hooks in different describe contexts, they are always executed, as shown in the second example here: https://github.com/mochajs/mocha/pull/3281#issuecomment-374300622

PR #3285, opened 2018-03-20
feature: add classbdd interface style

alfonso-presa:
This change adds a new interface called 'classbdd' which uses bdd interface under-the-hood but accepts class syntax suites as the following:

```js
module.exports = class IntegerPrimitivesArithmetic{

  before() {
    this.number1 = 1;
  }

  beforeEach() {
    this.number2 = 2;
  }

  itShouldFail() {
    expect(1).to.equal(2);
  }

  itShouldAdd() {
    expect(this.number1 + this.number2).to.equal(3);
  }

};
```

### Alternate Designs

I have not taken into account any other design strategies.

I think there's not a way to include a new interface from a plugin.

### Benefits

Less verbose and more friendly structure

### Possible Drawbacks

- Can't access to test context (right now)
- Test and suite names are methods (possible to clean by changing from camel case to regular text)
- Can't put suites inside others

### Applicable issues

N/A

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -761,7 +761,7 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
     -r, --require <name>                    require the given module
     -s, --slow <ms>                         "slow" test threshold in milliseconds [75]
     -t, --timeout <ms>                      set test-case timeout in milliseconds [2000]
-    -u, --ui <name>                         specify user-interface (bdd|tdd|qunit|exports)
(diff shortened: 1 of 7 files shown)


boneskull:
Thanks for the suggestion!  There is, in fact, a way to use an external interface: [see this wiki entry](https://github.com/mochajs/mocha/wiki/Third-party-UIs).

I'm going to close this, as Mocha isn't interested in adding new interfaces right now.

PR #3288, opened 2018-03-21
remove default js in "--watch-extensions" option; closes #3275

### LINKED ISSUE #3275: watch always watches .js
### Prerequisites

### Description

This is not an issue but feature request.

Watch mode always watches for ".js" files even when I specify --watch-extensions to be "ts,tsx". Is it possible to change the behaviour to strictly watch only the specified extensions?

TheDancingCode:
### Description of the Change

Changes the behaviour of  the `--watch-extensions` option. No longer includes `js` by default, meaning that the specified list becomes an exclusive and not an additional set of extensions.

### Benefits

Closes #3275.

### Possible Drawbacks

Mocha will not behave as expected for people who rely on `js` files always being watched.

### Applicable issues

#3275

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -200,7 +200,7 @@ program
   .option('--trace-deprecation', 'show stack traces on deprecations')
   .option('--trace-warnings', 'show stack traces on node process warnings')
   .option('--use_strict', 'enforce strict mode')
-  .option('--watch-extensions <ext>,...', 'additional extensions to monitor with --watch', list, [])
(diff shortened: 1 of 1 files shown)


boneskull:
Originally I called this a "bug", but it's not.  You can see from the [original help text](https://github.com/mochajs/mocha/pull/3288/files#diff-789b16a197fe1df6ad7686554bb32265L203) that it's acting in a way consistent with the documentation.

That makes this a purely breaking change, so it necessitates a major release.  This PR LGTM, but I want to discuss a branching/release plan with @mochajs/core before this is merged.

PR #3290, opened 2018-03-22
add update-contributors script; closes #3289

### LINKED ISSUE #3289: automate contributors list
Ideally this would be done upon release, but otherwise triggered manually. 

previously I had used [git-authors](https://github.com/tj/git-extras) to get the initial list--that's what `.mailmap` is for--but it required too much post-processing.

boneskull:
This updates the `contributors` prop of `package.json` and adds tooling to help.

This grabs information from Git only.  I am working on a better way to ensure more people are recognized, but this is what I have for now.  On the plus side, it's very fast.

We lose the GitHub user URLs, but not a big deal IMO.


scripts/update-contributors.js
```diff
@@ -0,0 +1,59 @@
...
+  (err, res) => {
+    if (err) {
+      throw err;
+    }
+    const contributors = Array.from(new Set(res.trim().split(/\r?\n/)))
+      .filter(contributor => BLACKLIST.indexOf(contributor) < 0)
+      .sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
+    const newCount = contributors.length;
+    const pkgFilepath = path.join(ROOT, 'package.json');
+    const pkg = JSON.parse(fs.readFileSync(pkgFilepath, 'utf8'));
```
outsideris: How about you just do `require(pkgFilepath);`?

boneskull: In this case, it’s mainly semantic. If I don’t need nor want the caching which `require()` provides, then I shouldn’t use it. In this script it doesn’t really matter since we don’t use the result elsewhere, but I feel it better illustrates the intent.

outsideris: I understood what you mean. :)

PR #3299, opened 2018-04-01
doc precision for exclusive tests

nicgirault:
This is a proposition of doc precision. My English is not so good so don't hesitate to propose a rephrase if you find the idea good but the sentence is not OK for you


boneskull:
thanks

Issue #3292, opened 2018-03-23
investigate use of Travis-Ci jobs and stages, etc.


boneskull:
This is a [beta feature](https://docs.travis-ci.com/user/build-stages/) on Travis-CI, but we could (potentially) leverage it to make it more obvious what's happening during each build in the matrix.

It may not be worth doing, either.  We don't have a bunch of packaging, deployment or artifact generation happening (though we have some).

Bamieh:
@boneskull Nice! I have been missing this feature on travis, while circleCI has a similar one. I will definitely try it out soon and give feedback/PR.

outsideris:
I closed this because #3302 has merged.

Issue #3155, opened 2017-12-13
investigate seemingly non-deterministic nyc output


boneskull:
I don't believe the coverage data out of [nyc](https://npm.im/nyc).  I've also seen [this sort of thing](https://github.com/mochajs/mocha/pull/2728#issuecomment-284276545) happen with my own two eyes.

It is likely a misconfiguration on our part, but I don't know where.  It may involve the way we are aggregating the results.

I would be indebted to @bcoe if he could take a look at [our command-line usage](https://github.com/mochajs/mocha/blob/master/Makefile#L8) and [config file](https://github.com/mochajs/mocha/blob/master/.nycrc).  Is there something obviously amiss?  

Hey, at least we aren't in sourcemap hell...

boneskull:
Why would a PR without code changes [like this](https://github.com/mochajs/mocha/pull/3150#issuecomment-351366080) exhibit the same behavior?

boneskull:
...need to make sure it's not bad math on Coveralls' part...

boneskull:
this is either no longer happening or I just haven't noticed.

PR #3302, opened 2018-04-03
use build stages for Travis CI

outsideris:
### Description of the Change
Introduce TravisCI build stages; it's still beta to make our builds more obviously and easier to understand. It groups our builds as `lint` or `test.`
I run linting first and then run tests because linting is much cheaper than tests.

### Alternate Designs
We can add more stages.
For example, we can use a `build` stage instead `before_script` or a `coverage` stage instead `after_success`.

IMO, it's more obviously, but it's tiny.
I'm not sure that we should use this yet. It's still beta.

### Benefits
Easy to understand our build status.

### Possible Drawbacks
If there are some bugs in this feature or Travis change something, our build might be broken, or we should update our `.travis.yml`.

### Applicable issues
https://github.com/mochajs/mocha/issues/3292

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,25 +6,37 @@ language: node_js
 
 matrix:
   fast_finish: true
-  include:
-    - node_js: '9'
-      env: TARGET=test.node COVERAGE=true
-    - node_js: '8'
-      env: TARGET=test.node
-    - node_js: '6'
-      env: TARGET=test.node
-    - node_js: '4'
-      env: TARGET=test.node
-    - node_js: '8'
-      env: TARGET=lint
-    - node_js: '8'
-      env: TARGET=test.browser
 
 before_install: scripts/travis-before-install.sh
 
 before_script: scripts/travis-before-script.sh
 
-script: npm start $TARGET
+jobs:
+  include:
+    - stage: lint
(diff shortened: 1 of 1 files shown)


outsideris:
Coverage change is annoying. I will dig [the problem](https://github.com/mochajs/mocha/issues/3155) when I have enough time.

.travis.yml
```diff
@@ -6,25 +6,37 @@ language: node_js
...
+
+before_script: scripts/travis-before-script.sh
+
+jobs:
   include:
-    - node_js: '9'
+    - stage: lint
+      node_js: '8'
+      env: TARGET=lint
+      script: npm start $TARGET
```
boneskull: I am guessing this can be simplified into something like

```yml
script: npm start $TARGET
node_js: '9'
jobs:
   include:
     - stage: lint
        env: TARGET=lint
     - stage: test # use default node version
        env: TARGET=test.node COVERAGE=true
     - stage: test
        node_js: '8'
        env: TARGET=test.node 
     - stage: test
        node_js : '6'
        env: TARGET=test.node
```

see https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Build-Matrix-Expansion

PR #3308, opened 2018-04-04
update stages and build matrices

boneskull:
- fiddle with caching 
  - don't cache for smoke tests
  - npm@5 comes with node@8 or newer, so cache `~/.npm` as well
  - otherwise, don't.
- smoke test for each node.js version (stage 1)
- lint (stage 2)
- tests for each node.js version and browsers (stage 3)
- addons (artifacts, sauce labs, chrome) are only present for browser tests!

YAML aliases are basically voodoo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,18 +21,15 @@ jobs:
       script: npm start test.node
 
     - <<: *node
+      cache: false
       node_js: '8'
 
     - <<: *node
-      cache:
-        directories:
-          - node_modules
+      cache: false
       node_js: '6'
 
     - <<: *node
-      cache:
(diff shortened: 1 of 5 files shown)


outsideris:
It looks good!

There is one thing.



It's hard to distinguish `test.node` and `test.browser` with node 9. Setting environment varibles can address this.
What do you think?

boneskull:
I raised this issue with Travis CI. If I set an environment variable then the cache can not be reused between the node and browser tests. There’s unfortunately no other way to distinguish them except to add a new stage at this point. Or just don’t worry about the caching

PR #3311, opened 2018-04-05
#3111 - Added a console warn to let users know they don’t have growl …

chadrickman:
…installed.

Updated the growl dependency to the latest version which returns the appropriate error when growl is not installed on system.

Also updating the logic when Mocha attempts to use a growl notification to log a warning if growl errors.

### Benefits

This will allow users who are attempting to use the growl notifications the ability to know they do not have it installed and provide them a link to find out how to resolve that.

### Possible Drawbacks

Additional messages in the console

### Applicable issues

#3111

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -244,14 +244,22 @@ Mocha.prototype._growl = function (runner, reporter) {
 
   runner.on('end', function () {
     var stats = reporter.stats;
+    var errMsg = 'The prerequisites for Growl support are missing. Please see https://github.com/mochajs/mocha/wiki/Growl-Notifications for details';
+    var msg = '';
+
     if (stats.failures) {
-      var msg = stats.failures + ' of ' + runner.total + ' tests failed';
-      notify(msg, { name: 'mocha', title: 'Failed', image: image('error') });
+      msg = stats.failures + ' of ' + runner.total + ' tests failed';
+      notify(msg, { name: 'mocha', title: 'Failed', image: image('error') }, function (err) {
+        if (err && err.code === 'ENOENT') {
+          console.warn(errMsg);
+        }
+      });
     } else {
-      notify(stats.passes + ' tests passed in ' + stats.duration + 'ms', {
-        name: 'mocha',
-        title: 'Passed',
-        image: image('ok')
+      msg = stats.passes + ' tests passed in ' + stats.duration + 'ms';
+      notify(msg, { name: 'mocha', title: 'Passed', image: image('ok') }, function (err) {
+        if (err && err.code === 'ENOENT') {
+          console.warn(errMsg);
+        }
       });
     }
   });
--- a/package-lock.json
+++ b/package-lock.json
@@ -5154,9 +5154,9 @@
       }
     },
     "growl": {
-      "version": "1.10.3",
-      "resolved": "https://registry.npmjs.org/growl/-/growl-1.10.3.tgz",
-      "integrity": "sha512-hKlsbA5Vu3xsh1Cg3J7jSmX/WaW6A5oBeqzM88oNbCRQFz+zUaXm6yxS4RVytp1scBoJzSYl4YAEOQIt6O8V1Q=="
+      "version": "1.10.5",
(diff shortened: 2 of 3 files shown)


lib/mocha.js
```diff
@@ -244,14 +244,22 @@ Mocha.prototype._growl = function (runner, reporter) {
...
   runner.on('end', function () {
     var stats = reporter.stats;
+    var errMsg = 'The prerequisites for Growl support are missing. Please see https://github.com/mochajs/mocha/wiki/Growl-Notifications for details';
+    var msg = '';
+
     if (stats.failures) {
-      var msg = stats.failures + ' of ' + runner.total + ' tests failed';
-      notify(msg, { name: 'mocha', title: 'Failed', image: image('error') });
+      msg = stats.failures + ' of ' + runner.total + ' tests failed';
+      notify(msg, { name: 'mocha', title: 'Failed', image: image('error') }, function (err) {
```
Bamieh: maybe we can abstract the error into a function `onNotifyError` or such then use it in both places. This way the code is better refactored and easier to understand

lib/mocha.js
```diff
@@ -244,14 +244,22 @@ Mocha.prototype._growl = function (runner, reporter) {
...
+        if (err && err.code === 'ENOENT') {
+          console.warn(errMsg);
```
boneskull: If this is *not* `ENOENT`, we should not eat the error; please throw it.

lib/mocha.js
plroebuck: Definitely, DRY is better

lib/mocha.js
plroebuck: Agreed.

TheSench: The error is already being thrown elsewhere, the callback parameter in `growl()` is used to add additional error reporting.  There is no need to throw an error inside the callback (note, the default callback is a noop function).

TheSench: @plroebuck The `notify` function (`growl`) takes in 3 parameters, the third of which being a callback that is fired when growl finishes. That callback is being fired from an event on the child process created by `growl`, essentially in a separate thread. Adding logic to rethrow that error results in no change to the output. Currently, a callback is not being passed in at all, causing growl to assign a noop function to the callback, meaning the error that growl returns in the callback is already being swallowed.

plroebuck: Sench has a point, though Growl errors aren't swallowed as they do show up in Mocha's output.

However, Growl support is tangential to Mocha's purpose. Rethrowing Growl errors doesn't buy us anything. As such, I think we should do something like:

```js
Mocha.prototype._growl = function (runner, reporter) {
  var notify = require('growl');
  var sendNotification = function () {
    var image = function (name) {
      return path.join(__dirname, '../images', name + '.png');
    };
    var onNotify = function (err) {
      if (err) {
        if (err.code === 'ENOENT') {
          process.stderr.write('Growl notification support available but not installed -- disabling!\n');
          runner.removeListener('end', sendNotification);
        } else {
          process.stderr.write('growl: ' +  err.message + '\n');
        }
      }
    };
    var stats = reporter.stats;
    var msg;

    if (stats.failures) {
      msg = stats.failures + ' of ' + runner.total + ' tests failed';
      notify(msg, { name: 'mocha', title: 'Failed', image: image('error') }, onNotify);
    } else {
      msg = stats.passes + ' tests passed in ' + stats.duration + 'ms';
      notify(msg, { name: 'mocha', title: 'Passed', image: image('ok') }, onNotify);
    }
  };
  runner.on('end', sendNotification);    // should this be `.once`?
};
```

Of course this leads me to ask how many `end` events we expect to process anyway.
If that answer is one, removing the listener may be too little, too late.

It's unfortunate that the `growl` package doesn't allow for solving this problem without attempting notification. [I asked](https://github.com/tj/node-growl/issues/74)... without response. Much better solved in `Mocha#growl` than `Mocha#_growl`.

plroebuck:
Yes, it should have been autoclosed when #3542 merged...

PR #3313, opened 2018-04-05
further improve build time with aggressive caching

boneskull:
- always cache the npm cache, which is in `~/.npm`
- don’t run `npm install` if `node_modules` was restored from cache; this seems to just rebuild a bunch of native modules.
- in the case of a cache hit, this makes each job ~18s quicker, which is about how long it takes to rebuild native modules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,8 +44,7 @@ jobs:
 
     - &smoke
       stage: smoke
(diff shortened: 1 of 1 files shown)


boneskull:
this isn't quite ready.

Issue #2572, opened 2016-11-03
Setting message (reason) for test skip


ponmudivn:
Is it possible to set a message (mentioning reason) why a particular test is skipped, so that it could be used in reporters.

```
describe('xxx', function() {
 checkToSkip(1)("test1", function() {\*test goes here*\});
 checkToSkip(4)("test2", function() {\*test goes here*\});
});

function checkToSkip(now) {
    return now > 3 ? it : xit; 
   //return now > 3 ? it : it.skip; 
}
```

Here 'test1' will be skipped as 'checkToSkip' returns 'xit' (or it.skip). Is it possible to pass a message to reporter mentioning the reason for the test skip? something like below (or any other possible way).

`checkToSkip(4)("test2", function() { \\ test goes here}, "My Skip message!!!!" );`

or 

`checkToSkip(4)("test2", function() { \\ test goes here}).pending("My Skip message!!!!");`

Note: Im using mocha in webdriverIO.

Thanks.

ponmudivn:
http://stackoverflow.com/questions/31688650/is-it-possible-to-send-a-reason-for-jasmine-2-specs-skipped-with-xit-or-pending

This thread explains the solution for same problem using Jasmine. Im looking for something similar in Mocha.

vivganes:
I too would love to have something like that.

I could not find any inputs about how to enter reason for skipping in documentation [here](http://unitjs.com/guide/mocha.html#ignore-suite-s-or-test-case-s-)

May be a candidate for a feature request?  It could just be a case that the documentation is missing while feature is there.

ponmudivn:
http://stackoverflow.com/questions/40397201/mocha-setting-message-reason-for-test-skip/40407177#40407177

fyi: a workaround to achieve this is available in the above thread.

boneskull:
this would be a good usability improvement, but see @ponmudivn's [workaround](https://github.com/mochajs/mocha/issues/2572#issuecomment-259087379)

jeffvandyke:
One approach that could cover all use cases and would only be a small api change would be to allow an optional `reason` parameter to `this.skip()`. The `it.skip(...)` case could be satisfied by just calling `this.skip("reason")` in the first line of the test.

monkey-programmer:
Agreed with @jeffvandyke I would love to have this feature

Nokel81:
How about doing:

```js
    it.skip("A test").reason("Code has changed");
```

[...]

Issue #2026, opened 2015-12-27
🚀 Feature: Allow adding a description with the reason why a test is pending


basharov:
I can add pending reason in Jasmine tests with `.pend()` method and replacing `it()` with `xit()` like this:

```
xit('should save settings when proper object is passed', function(){}).pend('No local storage is supported yet');
```

Can I somehow add a pending reason message for Mocha tests?

gurdiga:
As a lightweight workaround, I would think of tweaking the test description to include the reason why it’s pending. :blush: 
```js
xit('works as expected; PENDING: waiting for a dependency to be implemented', function() {
});
```
I’m guessing this would work, but I curious to know how would such a feature be useful? :construction_worker:

tommy-p:
This feature is pretty useful, is is not a good idea to mix test description and reason why test is pending. pend() method allows to distinguish pend reason from description. And instead of meaningless reason "No reason given" in jasmine you can add more info, issue number, etc.

rdennis:
What would the syntax be for this feature? I would expect something like:
```javascript
// describe.skip = xdescribe
describe.skip('Foo', function () {
    ...
}, 'Reason Foo was skipped');

describe('Bar', function () {
    // it.skip = xit
    it.skip('#baz', function () {
        ...
    }, 'Reason #baz was skipped');
});
```

Or would a chained function be more readable?
```javascript
describe.skip('Foo', function () {
    ...
})
.because('Reason Foo was skipped');

describe('Bar', function () {
    it.skip('#baz', function () {
        ...
    })
    .because('Reason #baz was skipped');
});
```

Nokel81:
I like the chain version since that seems to be the best inline with how the rest of mocha operates

yuriy-g-eleks:
Any plans on adding this feature?

fotoflo:
Using xit.() makes tests pending - which confused me as a new user because pending means "awaiting decision or settlement." - i assumed it was some async that hadn't completed. 

seems skipped would be a better description for xit. 

```
  24 passing (690ms)
  7 pending
```

JoshuaKGoldberg:
Related to the discussion: #1815 tracks separating "pending" and "skipped" categories.

This needs discussion on how the feature should look. Maybe it should be the fluent API (`test(...).describe("...")`)? Maybe rolled into #928 (test metadata)?

JoshuaKGoldberg:
[...]

PR #3324, opened 2018-04-13
Add a description with the reason why a test is pending

rdennis:
### Requirements

### Description of the Change

Added a `reason` property to `Runnable` and an optional `reason` parameter for `it`, `specify`, `xit`, `xspecify`, and `it.skip`. Also added the reason to reporters that made sense.

### Alternate Designs

I considered a `.because` or `.reason` method that would accept the reason a test is marked pending, but it would require a larger change and can still be done after these changes.

It is adding functionality to the core system.

### Benefits

It will allow users to give a reason tests are pending. This functionality is helpful in that it lets teams track the status of pending tests in the test report. Right now, if a test is pending, it has to be documented in the test itself, or by some other external means. 

### Possible Drawbacks

It is possible that this implementation breaks existing tests in some case I have not considered. If those are found, it may be better to use a `.reason` or `.because` method instead.

### Applicable issues

I have seen this functionality requested in #2026 and #2572. 
This is an enhancement (minor release).


lib/reporters/xunit.js
```diff
@@ -155,7 +155,8 @@ XUnit.prototype.test = function (test) {
     var err = test.err;
     this.write(tag('testcase', attrs, false, tag('failure', {}, false, escape(err.message) + '\n' + escape(err.stack))));
   } else if (test.isPending()) {
-    this.write(tag('testcase', attrs, false, tag('skipped', {}, true)));
+    var reasonTag = test.reason ? tag('reason', {}, false, '<![CDATA[' + test.reason + ']]>') : '';
```
boneskull: I'm going to assume "reason" is a real thing per the XUnit spec?  (I don't use XUnit, so I don't know)

lib/reporters/json.js
```diff
@@ -83,7 +83,8 @@ function clean (test) {
     fullTitle: test.fullTitle(),
     duration: test.duration,
     currentRetry: test.currentRetry(),
-    err: cleanCycles(err)
+    err: cleanCycles(err),
+    reason: test.reason
```
boneskull: 1.  Can we suppress addition of the `reason` property in the case of a **non-pending** test?
1.  Can we coerce `reason` to a `String` (e.g. `reason: String(test.reason)`)

lib/interfaces/bdd.js
```diff
@@ -78,12 +78,16 @@ module.exports = function (suite) {
      * acting as a thunk.
      */
 
-    context.it = context.specify = function (title, fn) {
+    context.it = context.specify = function (title, fn, reason) {
```
boneskull: Please see my review summary regarding this API.

lib/runnable.js
```diff
@@ -43,8 +43,9 @@ module.exports = Runnable;
  * @class
  * @param {String} title
  * @param {Function} fn
+ * @param {String} reason
  */
-function Runnable (title, fn) {
+function Runnable (title, fn, reason) {
```
boneskull: Please revert changes to this file (see review summary)

lib/test.js
```diff
@@ -20,12 +20,13 @@ module.exports = Test;
  * @api private
  * @param {String} title
  * @param {Function} fn
+ * @param {String} reason
```
boneskull: Please revert changes to this constructor.  Even if `reason` must be a property of a `Test` or `Runnable` (I'm not entirely convinced), it should be renamed `pendingReason` and settable/gettable via methods.

Please investigate using the reason as the `message` of a thrown `Pending` object instead.  I see at least one place where the `Runner` emits `pending` where we have access to the `Pending` instance itself... that might be all we need.  Then, we can do something like `this.emit('pending', test, pending.message)` and the reporter can handle it.

test/integration/pending.spec.js
```diff
@@ -32,6 +46,25 @@ describe('pending', function () {
         done();
       });
     });
+    it('should accept a reason when used via shorthand methods', function (done) {
```
boneskull: tests look great 👍

boneskull:
(Also this needs to be rebased onto `master` w/ conflicts resolved)

lib/reporters/xunit.js
Bamieh: @boneskull yes, in XUnit.net `[Fact(Skip="reason")]` or ` [Ignore("reason")]` are used to skip a `Fact/Test` and provide a reason for skipping.

rdennis:
Thanks for the feedback, I'll make these changes.

pdulapalli:
Hello, I was curious about the status of this PR. This is something I would really like implemented in Mocha. 

Would it be possible for me to make some of the changes you've recommended @rdennis to make?

rdennis:
I think I've made all of the requested changes. The best source I could find for the `xunit` XML format (https://xunit.github.io/docs/format-xml-v2) implies a completely different format from that of the current `xunit` reporter, so I didn't bother adding `reason` to it this time. If anyone can find me an authoritative source for what the current reporter should output to properly display `reason`, I'll gladly add it.

lib/runnable.js
```diff
@@ -116,9 +116,14 @@ Runnable.prototype.enableTimeouts = function(enabled) {
...
  * @api public
+ * @param {string} reason - Reason the Runnable is being skipped.
  */
-Runnable.prototype.skip = function() {
-  throw new Pending('sync skip');
+Runnable.prototype.skip = function(reason) {
+  if (arguments.length > 0) {
+    reason = String(reason);
+    this.reason = reason;
+  }
```
outsideris: Why just check `if (reason) {`?

In a case of`it.skip()`, `arguments` is `{ '0': undefined }` and `arguments.length` is `1`, because of `this.runnable().skip(reason)` in `lib/context.js`. 
Therefore, `reason` is `'undefiled'` as string and it print like thi:
```
- should skip immediately with reason (undefined)
```

outsideris:
I didn't use xunit as well. 
[Jenkins's junit xsd](https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd) looks same with our format.

I guess we can use [`message` in `skipped` element](https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd#L41) for skip reason.

```
<xs:element name="skipped">
    <xs:complexType mixed="true">
        <xs:attribute name="type" type="xs:string" use="optional"/>
        <xs:attribute name="message" type="xs:string" use="optional"/>
    </xs:complexType>
</xs:element>
```

outsideris:
Additionally, it will be awesome if you add about reason in document.
Maybe [here](https://github.com/mochajs/mocha/blob/master/docs/index.md#inclusive-tests)?

lib/runnable.js
rdennis: Ah, good catch. Originally I has `if (reason) {`, but thought this would be better for some reason. I'll change it back.

rdennis: I also considered using `utils.isString` to validate the input, but wasn't sure if that was needed here.

rdennis:
Ok, I added support for `before` and `beforeEach`, but it required not passing a default `message` to `new Pending` in `Runnable::skip` (sync and async). I'm not sure if this is an acceptable change as I'm not sure what side effects it may have. Also, it just feels a little weird having `before` give the `reason` to every test in the suite. Maybe it just shouldn't be supported, but it seems counter intuitive to allow `this.skip(reason)` in spec, but not in `before` or `beforeEach`.

craigtaub:
@rdennis have u had any time to look at this again?

lib/reporters/xunit.js
boneskull: @Bamieh happy to have someone around that actually uses XUnit :smile:

JoshuaKGoldberg:
👋 coming back to this @rdennis, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

PR #3328, opened 2018-04-14
Replace documentation with jsdoc with mocha flavor

Munter:
Switched api docs generation from documentation to jsdoc with the new mocha-docdash theme applied.

Also removed a bunch of error inducing comments in a lot o files.

The jsdoc output it not entirely useful yet, but more than the documentation output was.

There's still some work needed to get the jsdoc comments to actually produce really good output


boneskull:
I rebased all this crap onto `master`, updated the template package to `@mocha/docdash` and added `docs/API.md`.  [see result](https://deploy-preview-3328--mocha.netlify.com/api/) (or just append `/api/` to the netlify preview URL)

PR #3333, opened 2018-04-18
Nested before error using wrong test name in error message

### LINKED ISSUE #3291: Mocha shows error in inner `before` function as error for outer `it`
### Prerequisites

### Description

Mocha shows error in inner `before` function as error for outer `it`, but this outer `it` passed successfully.

### Steps to Reproduce
```
describe('Top-level describe', () => {

  it('Top-level it', () => { });

  describe('Nested describe', () => {
    before(() => {
      throw new Error();
    });

    it('Nested it', () => { });
  });
});
```

**Expected behavior:** Info about error in inner `before` statement for `Nested describe`

**Actual behavior:** Mocha shows this error: 
```
  Top-level describe
    ✓ Top-level it
    Nested describe
      1) "before all" hook for "Top-level it"
```

**Reproduces how often:** always

### Versions

- The output of `mocha --version`  - not installed 
- `node node_modules/.bin/mocha --version`: 5.0.5
- The output of `node --version`: v8.10.0
- The version and architecture of your operating system: Ubuntu 16.04 LTS 64 bit

chinchiheather:
### Description of the Change

When an error is encountered in a hook, set the current test on `hook.ctx` to the first one in the current suite

### Benefits

The error message shown for a test will use the correct test name
Previously if you had a nested before function which errored, it would use the previous test's name

```
describe('Top-level describe', () => {

  it('Top-level it', () => { });

  describe('Nested describe', () => {
    before(() => {
      throw new Error();
    });

    it('Nested it', () => { });
  });
});
```
would show:
```
Top-level describe
    ✓ Top-level it
    Nested describe
      1) "before all" hook for "Top-level it"
```

### Applicable issues

Fixes #3291

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -324,8 +324,8 @@ Runner.prototype.hook = function(name, fn) {
             hook.pending = true;
           }
         } else {
-          // set the current test to the next one to run so the error message is correct
           if (!self.suite.bail()) {
+            // set the current test to the next one to run so the error message is correct
             hook.ctx.currentTest = suite.tests[0];
           }
           self.failHook(hook, err);
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -324,6 +324,8 @@ Runner.prototype.hook = function(name, fn) {
             hook.pending = true;
           }
         } else {
+          // set the current test to the next one to run so the error message is correct
+          hook.ctx.currentTest = suite.tests[0];
           self.failHook(hook, err);
 
           // stop executing hooks, notify callee of hook err
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -325,7 +325,9 @@ Runner.prototype.hook = function(name, fn) {
           }
         } else {
           // set the current test to the next one to run so the error message is correct
-          hook.ctx.currentTest = suite.tests[0];
+          if (!self.suite.bail()) {
+            hook.ctx.currentTest = suite.tests[0];
+          }
           self.failHook(hook, err);
 
           // stop executing hooks, notify callee of hook err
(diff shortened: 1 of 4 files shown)


lib/runner.js
```diff
@@ -322,6 +322,8 @@ Runner.prototype.hook = function (name, fn) {
             hook.pending = true;
           }
         } else {
+          // set the current test to the next one to run so the error message is correct
+          hook.ctx.currentTest = suite.tests[0];
```
boneskull: what happens if this is an "after" hook?

chinchiheather: Sorry for the late response, I missed your comment.
I have added a test case for the after hook, it seems to be working as expected.
I also had to make an additional change as a test for the bail option was failing.
Please have another look :)

lib/runner.js
```diff
@@ -331,6 +331,10 @@ Runner.prototype.hook = function(name, fn) {
             hook.pending = true;
           }
         } else {
+          // set the current test to the next one to run so the error message is correct
```
plroebuck: Shouldn't comment be inside the `if` condition?

plroebuck:
Nested `beforeEach()` does not trigger the problem, right? **Only `before()`**.

plroebuck:
@chinchiheather ping!

chinchiheather:
@plroebuck sorry I had kind of forgotten about this, I have moved the comment.
I am getting a failing test for node v9, I can try and fix, but as this PR was left open for so long I'm not sure if anyone is actually wanting to merge it in?

craigtaub:
You can click on the coverage report and see which line and file has the issue. If nothing obvious perhaps try a rebase.

chinchiheather:
hi @craigtaub @plroebuck i have rebased and resolved all the merge checks now

plroebuck:
@juergba ?

juergba:
@chinchiheather Thanks for your contribution and patience with us. Please understand we have chosen an alternative solution and I'm therefore closing this PR.

PR #3340, opened 2018-04-21
#3336: Make `--watch-extensions` default to 'js'

### LINKED ISSUE #3336: --watch-extensions needs to default to 'js'
The change in c580294893e1bf2f9e51adafd73e6cac71690d19 to `--watch-extensions` (published accidentally in v5.1.0) fails to set a default value.

If `--watch-extensions` is not specified, it should default to `.js`; otherwise use whatever was specified.

PRs should target `master` as usual.

plroebuck:
### Description of the Change

The change in c580294 to `--watch-extensions` failed to set a default value.
This change specifies that default value.

### Benefits

`--watch` default functionality restored.

### Applicable issues

Fixes #3336
(patch release)


boneskull:
This is actually going to be part of v6.0.0 since the functionality that it fixes will land then. `master` is bleeding-edge code and v5.1.1 doesn’t contain it

boneskull:
Thank! landed in 9d407deb77d9f6bdc44928e25dd8a739069d6a0d.

outsideris:
It's weird. `eslint` should find this kind of errors.

boneskull:
Part of the problem is likely the lack of filename extension.

PR #3346, opened 2018-04-22
fix to exit correctly when using bail flag

### LINKED ISSUE #3303: When using bail flag, before hook failures do not exit as failed
### Prerequisites

### Description

When using the bail flag before hook failures do not exit the test as a failure.

### Steps to Reproduce

sample test:
``` js
//index.js

describe('some test', () => {
    before(async () => {
        throw new Error('should die');
    });

    it('should not get here', () => {
        assert(false);
    });
});

```

command:
```
mocha --bail index.js
```

output:
```
  some test

  0 passing (6ms)

    1) "before all" hook
```

Although its hard to see here, the test is exiting with a 0 code.  You can see it better by running:
```
mocha --bail index.js && echo 'hello'

  some test

  0 passing (9ms)

    1) "before all" hook

hello
```

The `echo 'hello'` should never get hit but it does.

Similarly, running this without `--bail` works as expected, the process exits with a non-zero code:
```
mocha index.js && echo 'hello'
```

output:
```
  some test
    1) "before all" hook


  0 passing (7ms)
  1 failing

  1) some test
       "before all" hook:
     Error: should die
      at Context.<anonymous> (index.js:5:15)
```
**Note: Regardless of before block structure this bug still occurs**
``` js

describe('some test', function() {
    before(function() {
        throw new Error('should die');
    });

    it('should not get here', function() {
        assert(false);
    });
});
// still bad
```

**Expected behavior:** Test should fail if there is an error in before hook 

**Actual behavior:** Test does not fail

### Versions
mocha v5.0.5 global and local (tested version 4 through 5 and 5.0.5 seems to be the only affected version)
macOS
zsh

### Additional Information

Related to: https://github.com/mochajs/mocha/issues/3096

outsideris:
### Description of the Change
The bail flag fixed in #3278, but it doesn't exit correctly when `before` hook throw an error.
This PR fix to exit correctly whenever test or hook failed.

### Alternate Designs
I don't have any idea.

Now, when `before` hook failed, mocha doesn't exit correctly. It makes bail flag useless.

### Benefits
Users believe the result of mocha with bail flag.

### Possible Drawbacks
I'm not sure.

### Applicable issues
Fix #3303


medikoo:
When can we expect the fix to be published?

medikoo:
On our side it hidden failed builds for weeks, so bug seems critical (for now we reverted to `5.0.4`)

spalger:
sorry for the unnecessary ping, but also anxious for this fix

boneskull:
published as [v5.2.0](https://github.com/mochajs/mocha/releases/tag/v5.2.0)

PR #3348, opened 2018-04-23
update docs with globbing and shell expansion details; closes #3136

### LINKED ISSUE #3136: Documentation around glob patterns and shell expansion
_From @boneskull on July 12, 2015 4:15_

from mochajs/mocha#1115

> Not sure if it is needed, but it may be worth noting that due to shell expansion, glob patterns should be wrapped in double quotes on unix shells.
> 
> ```
> mocha -w -r should "test/**/*.js"
> ```
> 
> Using that would render the --recursive flag useless.

_Copied from original issue: mochajs/old-mochajs-site#8_

**UPDATE** Double quotes for portability--don't want to say single quotes in one place and double quotes in the other, so DOUBLE QUOTES (single quotes fine on *nix if you don't want variable expansion to happen, but let's assume there are no variables in use)

akrawchyk:
Add documentation around shell expansion and globbing for passing files to mocha

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1273,19 +1273,15 @@ $ mocha --reporter list --growl
 
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
 
 To configure where `mocha` looks for tests, you may pass your own glob:
 
-```
-$ mocha ./spec/*.js
-```
-
-If you use shell expansion for this, you must wrap your glob pattern in double quotes (note, using the `--recursive` flag here is useless):
-
-```
+```sh
 $ mocha "./spec/**/*.js"
 ```
+
+*Note*: Double quotes around the glob are recommended for portability.
 
 ## Editor Plugins
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -1273,9 +1273,15 @@ $ mocha --reporter list --growl
 
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option.
 
 To configure where `mocha` looks for tests, you may pass your own glob:
+
+```sh
+$ mocha --recursive "./spec/*.js"
+```
+
+Some shells support recursive matching by using the `**` wildcard in a glob. Bash >= 4.3 supports this with the [`globstar` option](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html) which [must be enabled](https://github.com/mochajs/mocha/pull/3348#issuecomment-383937247) to get the same results as passing the `--recursive` option ([ZSH](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Recursive-Globbing) and [Fish](https://fishshell.com/docs/current/#expand-wildcard) support this by default). With recursive matching enabled, the following is the same as passing `--recursive`:
 
 ```sh
 $ mocha "./spec/**/*.js"
--- a/docs/index.md
+++ b/docs/index.md
@@ -1275,6 +1275,18 @@ $ mocha --reporter list --growl
 
 By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
 
+To configure where `mocha` looks for tests, you may pass your own glob:
+
+```
+$ mocha ./spec/*.js
+```
+
+If you use shell expansion for this, you must wrap your glob pattern in double quotes (note, using the `--recursive` flag here is useless):
+
+```
+$ mocha "./spec/**/*.js"
+```
+
 ## Editor Plugins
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -1275,6 +1275,18 @@ $ mocha --reporter list --growl
 
 By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
 
+To configure where `mocha` looks for tests, you may pass your own glob:
+
+```
+$ mocha ./spec/*.js
+```
+
+If you use shell expansion for this, you must wrap your glob pattern in double quotes (note, using the `--recursive` flag here is useless):
```
boneskull: A lot of people probably doesn't know what "shell expansion" is.

In fact, I would probably remove L1279-1283 and just give the example below with double quotes.  Then, explain below the example how using `**` means `--recursive` is useless.  Recommend using double quotes for "portability"; otherwise it'll need a pretty long-winded explanation.

boneskull (changes requested):
Thanks for this!  I have a suggestion.

Also, these changes failed the Markdown lint.  run `npm start lint.markdown` to check the Markdown locally.

akrawchyk:
@boneskull I made the updates you recommended, except for "using ** means --recursive is useless". I think we might need to mention shell support/differences for the `**` glob, otherwise these changes could be confusing.

The lines above my changes read:

> ...since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.

This is true by default in Bash, but [the `globstar` option](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html) was [introduced in Bash 4.3](https://tiswww.case.edu/php/chet/bash/CHANGES) (so you need a recent version just to use this) and must be enabled for `**` == `--recursive`.

I believe ZSH and Fish support recursive matching for `**` by default.

docs/index.md
```diff
@@ -1273,7 +1273,15 @@ $ mocha --reporter list --growl
...
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
+
+To configure where `mocha` looks for tests, you may pass your own glob:
+
+```sh
+$ mocha "./spec/**/*.js"
+```
```
outsideris: I didn't know about the globstar in Bash 4.3.
I feel that current explanation is a little bit confused. If we explain the globstar explicitly, users can understand clearer because the behavior of `**` is depend on a shell.

akrawchyk:
@boneskull @outsideris I've updated the changes to reflect your feedback, please give it another look when you can!

outsideris:
hmmm.. Why did builds failed for this PR?

outsideris:
Aha! The failure is related with [this](https://github.com/mochajs/mocha/pull/3350#issuecomment-384703961).

akrawchyk:
@outsideris @boneskull I rebased from develop with the CI fixes, let me know if you need me to adjust anything else

outsideris (approved):
LGTM

akrawchyk:
@boneskull do you have more feedback on this?

akrawchyk:
@boneskull @outdooricon let's get this merged!

akrawchyk:
@boneskull @outsideris let's get this merged! (please 😄 )

outsideris:
Thank you.

docs/index.md
```diff
@@ -1273,7 +1273,21 @@ $ mocha --reporter list --growl
...
+$ mocha --recursive "./spec/*.js"
+```
+
+Some shells support recursive matching by using the `**` wildcard in a glob. Bash >= 4.3 supports this with the [`globstar` option](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html) which [must be enabled](https://github.com/mochajs/mocha/pull/3348#issuecomment-383937247) to get the same results as passing the `--recursive` option ([ZSH](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Recursive-Globbing) and [Fish](https://fishshell.com/docs/current/#expand-wildcard) support this by default). With recursive matching enabled, the following is the same as passing `--recursive`:
+
+```sh
+$ mocha "./spec/**/*.js"
+```
+
+*Note*: Double quotes around the glob are recommended for portability.
```
plroebuck: Disagree with this statement. 
If you want _portability_, then _single quotes_ are required; using double quotes allows possibility of shell expansion, etc. Using single quotes tells the shell "hands off", and Mocha's `lookupFiles` would handle the glob expansion.

akrawchyk: @plroebuck thanks for the feedback. My understanding is that glob expansion would only happen when using parameter expansion inside double quotes, see https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html#Shell-Parameter-Expansion.

fwiw the context for specifying double quotes is here https://github.com/mochajs/mocha/issues/3136#issue-280703663

plroebuck: Yes, you're correct that double quotes are required, if you want the _shell_ to do the job. Perhaps I misunderstood the intent of this update, assuming it was to explain how to target files recursively (via glob) in general. While knowing _how_ to do this via UNIX shell is useful (but unnecessary here), the _portable_ recommendation would be to use single quotes, sidestepping platform differences and characteristics of various shells with various options disabled/enabled.

docs/index.md
```diff
@@ -1273,7 +1273,21 @@ $ mocha --reporter list --growl
 
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, use `--recursive`, since `./test/*.js` only matches files in the first level of `test` and `./test/**/*.js` only matches files in the second level of `test`.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option.
+
+To configure where `mocha` looks for tests, you may pass your own glob:
+
+```sh
+$ mocha --recursive "./spec/*.js"
```
plroebuck: In this example, the glob is unnecessary.
```bash
$ mocha --recursive spec
```

docs/index.md
Dustin4444, line 1: #5704

Dustin4444 (commented):
#6243

PR #3350, opened 2018-04-24
add Node.js v10 to matrix

boneskull:
AFAIK Node.js v9.x is no longer maintained as of today.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -16,9 +16,6 @@ rules:
   strict:
     - error
     - safe
-  linebreak-style:
-    - error
-    - unix
 overrides:
   - files:
     - scripts/**/*.js
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,9 @@ jobs:
       node_js: '8'
 
     - <<: *node
+      node_js: '9'
+
+    - <<: *node
       node_js: '6'
 
     - <<: *node
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,9 @@ jobs:
           - node_modules # npm install, unlike npm ci, doesn't wipe node_modules
 
     - <<: *smoke
+      node_js: '9'
+
+    - <<: *smoke
       node_js: '8'
 
     - <<: *smoke
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ stages:
 
 # defaults
 language: node_js
-node_js: '9'
+node_js: '10'
 # `nvm install` happens before the cache is restored, which means
 # we must install our own npm elsewhere (`~/npm`)
 before_install: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ node_js: '10'
 before_install: |
   [[ ! -x ~/npm/node_modules/.bin/npm ]] && {
     # caching feature creates `~/npm` for us
-    cd ~/npm && npm install npm
+    cd ~/npm && npm install npm@^5
     cd -
   } || true
   # avoids bugs around https://github.com/travis-ci/travis-ci/issues/5092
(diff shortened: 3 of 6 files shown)


plroebuck:
@Bamieh wasn't Node9's raison d'être the development/EA release of Node10? If Node10 is available, why bother?

plroebuck:
[This](https://travis-ci.org/mochajs/mocha/jobs/370776294#L627) was kinda humorous though...

plroebuck:
Maybe consider using a single fake environment variable (e.g., MOCHA_TEST) in ".travix.yaml" for each job to annotate what/why. For instance, the last job appears to be running browser tests where others are testing Node. So maybe the last job has `MOCHA_TEST=bundle,browser` and the others `MOCHA_TEST=node`? Maybe the COVERAGE variable could be in the environment as well. Smoke tests could annotate `MOCHA_TEST=production`. Just a thought.

boneskull:
@Bamieh Noted; I'm not sure what "end of life" vs. "not maintained" means, exactly.  I'll try to clarify with the release team.

boneskull:
@plroebuck We used to do this, but the reason we no longer do so is because it breaks caching (see my [blog post](https://boneskull.com/mocha-and-travis-ci-build-stages/) about it).  We run 3 builds using the *same pre-cached dependency trees*: linting, Node.js v<latest>, and finally browser tests.  If, say, the browser test had an environment variable that the other two did not, we wouldn't gain the speed benefit of the pre-cached dependencies.

The fact that the builds look the is simply due to poor UX.  Each build should allow for a unique name, but no such thing exists.  Environment variables are displayed (because they want to display "something"), but those are functional and have other implications.  The differentiator between our builds is the `script`, but the UI doesn't display it (for good reason).

We were kicking around the idea of migrating the build to Jenkins, and I'm still up for it, since that'd give us more control.  Just don't have time to configure it.  The server would be paid for by the JS Foundation.

boneskull:
OK, I have confirmation that Node.js v9 will receive security updates through June.

At that time, we should drop Node.js v9.

We can still add Node.js v10, so I'll modify this PR.

plroebuck:
BTW, believe this line in your patch should be fast-tracked (.travis.yaml#L16) to master.
```bash
     cd ~/npm && npm install npm@^5
```

npm-6 dropped support for Node-4, so **all submitted PRs will fail TravisCI on Node-4 test**.

boneskull:
I'd fast-track it if appveyor didn't start breaking

plroebuck:
Was about to look into that failure myself. It started failing symlink tests, and also seemed to have not installed Windows growl-substitute runtime dependency.

boneskull:
since AppVeyor doesn't seem to just build against `master`, it's maybe the changes to `package-lock.json`.  I'm not sure they're backwards compatible (potentially npm bug).  I'll see what happens when I revert a7b113938f0086a438329aeecdde52defd71886c

boneskull:
adding `[ci skip]` to the commit message when modifying `package-lock.json` is a mistake which I won't repeat

plroebuck:
Here's the link for Windows Growl:
<http://www.growlforwindows.com/gfw/d.ashx?f=growlnotify.zip>

boneskull:
It's hard to tell, but I'm not confident growl is the problem.

boneskull:
I'll try just ignoring npm...

boneskull:
(the growl notification happens *after* the failure)

boneskull:
I wonder if AppVeyor changed something on their side.  those two failing tests should be skipped.

boneskull:
alright, I investigated this in windows, and while I found some other issues related to the dev environment, I couldn't reproduce this problem.  I'm just going to change the logic to "skip if windows"--instead of feature-checking--and call it good

boneskull:
alright, LGTM.

plroebuck:
To  ease anyone's OCD sensibilities when viewing Travis CI's build page, maybe these two lines:

https://github.com/mochajs/mocha/blob/e256a7a6b8b019c4de4260d3fbb7c53593c41e7a/.travis.yml#L35
https://github.com/mochajs/mocha/blob/e256a7a6b8b019c4de4260d3fbb7c53593c41e7a/.travis.yml#L38

could be swapped so the jobs run in descending order?

boneskull:
@plroebuck 08dbaa3

Issue #3354, opened 2018-04-25
Annotate when exceptions are caught but ignored 


plroebuck:
### Description

There's an [existing convention](https://duncan99.wordpress.com/2013/05/20/try-catch-ignore/) of naming intentionally ignored errors in try/catch as `ignore`. Modify codebase to follow it.

PR #3344, opened 2018-04-22
format code with Prettier

boneskull:
This adds Prettier w/ ESLint integration.  Formatting is managed via [prettier-eslint-cli](https://npm.im/prettier-eslint-cli) which formats with Prettier then applies ESLint fixes thereafter.

The only noticeable conflict with our semistandard-based ESLint config is the lack of a space after the `function` keyword; this particular rule is suppressed by `eslint-plugin-prettier` and its ilk.

The motivations here:

1. Consistency
2. Automation
3. Avoid debate about code style

[lint-staged](https://npm.im/lint-staged) and [husky](https://npm.im/husky) are here to *automatically format* staged JS files as a *pre-commit* hook.  I personally use this setup in other places and it hasn't caused any friction.

PR #3356, opened 2018-04-26
#3354: Annotate when exceptions are caught but ignored

plroebuck:
### Description of the Change

Annotates intentionally ignored errors in [try] catch blocks by naming convention.
Ignored error is named `ignore` and catch body consists of comment explaining why;
no other code allowed when following the convention. Ideally, there should be **very**
few of these in your codebase.

### Benefits

Easier to locate. Reduction of technical debt?
Found several other probable errors in codebase while doing this one.

### Possible Drawbacks

None.

### Applicable issues

Fixes: #3354 
(patch release)


bin/options.js
```diff
@@ -30,8 +30,8 @@ function getOptions() {
       : process.argv[process.argv.indexOf('--opts') + 1];
 
   try {
-    const opts = fs
-      .readFileSync(optsPath, 'utf8')
+    // prettier-ignore
+    const opts = fs.readFileSync(optsPath, 'utf8')
```
outsideris: Why did you ignore prettier here?

bin/options.js
```diff
@@ -40,8 +40,8 @@ function getOptions() {
     process.argv = process.argv
       .slice(0, 2)
       .concat(opts.concat(process.argv.slice(2)));
-  } catch (err) {
-    // ignore
+  } catch (ignore) {
+    // NOTE: should console.error() and throw the error
```
outsideris: What did you mean by this comment?

bin/options.js
plroebuck: Because the first function call is to `fs.readFileSync()`; there is no good reason to split the module from the function call -- not to mention it looks stupid. The rest of the chain processes its result via various FP methods (indented as expected for chained methods).

bin/options.js
plroebuck: Exactly what it says. Believe it's a shortcoming of existing code. If anything goes wrong parsing the "mocha.opts" file or splicing its pieces/parts back into the cmdline arguments, the error is silently discarded. This just seems wrong, so I made a note; figured it should at least be discussed (and hopefully fixed in different issue).

bin/options.js
outsideris: I also like that no splitting the module and function call.
However, I believe that we should minimalize inline ignore because [we decided to follow Prettier](https://github.com/mochajs/mocha/pull/3344).

bin/options.js
outsideris: @plroebuck I got it. Thank you for the explanation.

bin/options.js
boneskull: please take it out, thanks

bin/options.js
boneskull: Yeah, I think we should be emitting a warning here, at minimum--unless there's some Git or issue history that makes me think otherwise.

That env var should be set anyway, because it's only used to check whether or not `getOptions()` should be called.  

But that's another issue.

boneskull:
fixed a conflict.. hopefully CI will like it

bin/options.js
outsideris: Agreed.

plroebuck:
Isn't this one done?

boneskull:
this is ready to merge after conflicts get resolved.  Thanks!!

boneskull:
@plroebuck you might want to join us in the [contributors' chat](https://gitter.im/mochajs/contributors) if you plan on contributing more

plroebuck:
> this is ready to merge after conflicts get resolved. Thanks!!

@boneskull, uh, what conflicts are you referring to here?

outsideris:
@boneskull My UI shows `This branch has no conflicts with the base branch`.
I think I can squash/merge it. Can I?

boneskull:
Oh, I see, I can squash, but not rebase.

PR #3361, opened 2018-04-28
Clarify docs for watch that it watches files in the CWD

benglass:
### Description of the Change

The --watch option for mocha watches for changes in any files in the CWD. The website mentions this. The mocha --help and docs/index.md dont. This is an important detail because it is reasonable that some users may expect it watches only the specified test files and may not think to check the docs on the website.

Makes the docs from mocha --help consistent with those on the website.

### Possible Drawbacks

If this behavior of --watch changed the docs would need to be updated in multiple places.


bin/_mocha
```diff
@@ -180,7 +180,7 @@ program
     `specify user-interface (${interfaceNames.join('|')})`,
     'bdd'
   )
-  .option('-w, --watch', 'watch files for changes')
+  .option('-w, --watch', 'watch files in the CWD for changes')
```
boneskull: please expand CWD to "current working directory" in both cases

benglass: @boneskull good idea, done

plroebuck:
@benglass would you do the required `prettier` indentation to pass `lint` check so this can get merged?

plroebuck:
@outsideris, how about you just reformat it for him and merge it? Too trivial to bother.

PR #3362, opened 2018-04-29
fix report in a case of afterEach conditioanl fail

### LINKED ISSUE #1635: afterEach halts suite

### LINKED ISSUE #3345: afterEach conditional error hook breaks stats and error reporting
### Prerequisites

### Description

Trying to fail a test with this.test.error() causes the test to be reported twice, as failed and as success

### Steps to Reproduce

Run the test at:

https://github.com/mochajs/mocha/wiki/HOW-TO:-Conditionally-fail-a-test-after-completion

**Expected behavior:**
```
  something
    ✓ should one
    1) should two

  1 passing (16ms)
  1 failing
```

**Actual behavior:**
```
  something
    ✓ should one
    ✓ should two
    1) should two

  2 passing (11ms)
  1 failing

```

outsideris:
### Description of the Change
In our [wiki](https://github.com/mochajs/mocha/wiki/HOW-TO:-Conditionally-fail-a-test-after-completion), describe  "Conditionally fail a test after completion".
But it behaves like the explanation. 
If a test failed after it already passed, the final report says both the test passed and the test failed. I referred #1944 .

### Alternate Designs
I don't have an idea now.

Users could get a correct report.

### Benefits
We can report correctly. It is confused if same test place on both pass and failure.

### Possible Drawbacks
It seems a broken change because report order could be changed in some edge case.
And this PR is inspired by #1944 , the PR's owner didn't sign CLA. I'm not sure whether we can make this code change or we should get permission from the one.

### Applicable issues
close #3345
close #1635


test/integration/hooks.spec.js
```diff
@@ -49,4 +50,33 @@ describe('hooks', function() {
       done();
     });
   });
+
+  it('can fail a test in an afterEach hook', function(done) {
```
boneskull: wouldn't it be easier to just look at the output of `runMochaJSON()`?

test/integration/hook-err.spec.js
```diff
@@ -39,7 +39,7 @@ describe('hook error handling', function() {
   describe('after each hook error', function() {
     before(run('hooks/afterEach-hook-error.fixture.js'));
     it('should verify results', function() {
-      assert.deepEqual(lines, ['test 1', 'after', bang + 'test 3']);
+      assert.deepEqual(lines, ['test 1', 'after', bang, 'test 3']);
```
boneskull: is this related?

outsideris: https://github.com/mochajs/mocha/blob/8368722830ed80e03a52d29155310cdc7b6fc123/test/integration/hook-err.spec.js#L159 
In here,`splitRegExp` is `[\n.]+`.
So, prev result is `.!test 3`, but now the result is `!.test 3`.
Therefore, `!` and `test 3` is separated.

test/integration/hooks.spec.js
outsideris: I changed to use it.

outsideris:
How about `semver-minor`?
I think it is correct behavior, but it works as previous behavior for a long time even if our wiki describes it.
Now, checking test passed ran in `afterEach`, not `run`.
This change pass count/failure count.

plroebuck:
Can we fix spelling in PR title?

outsideris:
I fixed type in commit message as well and linting error. CI should be passed.

plroebuck:
@outsideris, @boneskull,

issue #3569 may be implicating this PR. Can you confirm?
If so, we may need to back this out...

plroebuck:
@boneskull, this needs to get unmerged until the order is corrected. See #3569.

boneskull:
I'll reverse this.

papb:
@outsideris Hello, I am facing this issue... You seem to have an idea on how to solve it:

> We should check `afterEach` hooks to support conditional fail and report it in a correct order.

How can I help?

PR #3374, opened 2018-05-03
avoid global timers by way of ESLint; closes #3342

### LINKED ISSUE #3342: the Runnable test is bonkers
in the `Runnable` suite (`test/unit/runnable.spec.js`) we are poisoning the global `setTimeout`, `setInterval`, etc., to assert that we don't call those functions ([read the comments](https://github.com/mochajs/mocha/blob/master/test/unit/runnable.spec.js) for further explanation of why this is necessary).

but, really, that problem is better solved with linting.  **write a one-off eslint plugin/rule that ensures Mocha never calls these functions internally, once the test files begin loading**.

In other words, this is allowed:

```js
var setTimeout = global.setTimeout;
setTimeout(something);
// or
var setTimeout = require('timers').setTimeout;
setTimeout(something);
```

This is **not** allowed:

```js
global.setTimeout(something);
// or
setTimeout(something);
// or
require('timers').setTimeout(something);
```

boneskull:
- Remove unused variables to avoid problems with #237 (and their associated inline ESLint directives)
- Use existing ESLint rules (`no-restricted-globals`, `no-restricted-modules` and `no-restricted-syntax`) to disallow global timer access.
- Removed "poison pill" tests in `test/unit/runnable.spec.js`.

For those rules, our config applies only to `bin/*` and `lib/**/*.js`.

- `no-restricted-globals`: Disallow bare references (including calls) to timers like `setTimeout`.
- `no-restricted-modules`: Disallow use of Node.js' `timers` module to get around the above restriction.
- `no-restricted-syntax`: 
  - Disallow calls of `global.setTimeout` and other timers
  - Disallow `new global.Date()`
  - Disallow any expressions involving `global.Date.*`, `global.setTimeout.*` and other timers


plroebuck:
So I suggested this, but a little lost on the details. Why are the remaining `global.Date` references not causing errors? And commit can't close without rest of the changes (you can't institute the change without "package.json", "package-lock.json" and ".eslintrc" changing as well).

boneskull:
@plroebuck 

> Why are the remaining global.Date references not causing errors? 

Which references?  You can reference `global.Date`, but not e.g. `global.Date.now`, nor can you `new global.Date()`.

> And commit can't close without rest of the changes (you can't institute the change without "package.json", "package-lock.json" and ".eslintrc" changing as well).

I don't understand what you mean.

boneskull:
@plroebuck added full URL
@outsideris fixed unused vars in `bin/_mocha`

boneskull:
@plroebuck fwiw, I thought we needed a custom rule too, but while I was writing such a rule and looking at existing rules as examples, I found that this wasn't necessary.

.eslintrc.yml
```diff
@@ -38,4 +38,37 @@ overrides:
...
       node: no
-
+  - files:
+    - bin/*
+    - lib/**/*.js
+    rules:
+      no-restricted-globals:
+        - error
+        - name: setTimeout
+          message: &GH-237 See https://github.com/mochajs/mocha/issues/237
```
Bamieh: can you explain the `*` and the `&` in the message?

plroebuck:
BTW, what's up with the "{:class="screenshot"}" immediately above the Usage section of docs?

.eslintrc.yml
plroebuck: `&` defines and names an anchor, and `*` uses it. In this case, you don't have to repeat the same message over and over (DRY).

outsideris:
@plroebuck `{:class="screenshot"}` is from `string diffs{:class="screenshot"}`.
So, it is for specifying css class for the image. (but GitHub doesn't support it)

.eslintrc.yml
Bamieh: @plroebuck never knew that! awesome

plroebuck:
@outsideris , rephrased, I knew what it's intent was - but as GH doesn't support it, why is it there?

boneskull:
@plroebuck 

- re the class business: because `docs/index.md` becomes https://mochajs.org by way of Jekyll.
- re "handled?": afaict this code specifically allows Sinon and other utils to screw with those globals w/o triggering a global leak error.
- re `timeout()` ahh, good catch.  I'll add that to try to make coveralls happy.

boneskull:
actually, I'll just modify that one test I wrote to use `1s`.  `test/unit/ms.spec.js` covers everything there.  I'll find some other coverage gap here.

test/unit/runnable.spec.js
```diff
@@ -109,11 +115,11 @@ describe('Runnable(title, fn)', function() {
 
   describe('#globals', function() {
     it('should allow for whitelisting globals', function(done) {
-      var test = new Runnable('foo', function() {});
-      assert(test.async === 0);
-      assert(test.sync === true);
-      test.globals(['foobar']);
-      test.run(done);
+      var runnable = new Runnable('foo', function() {});
+      assert(runnable.async === 0);
```
plroebuck: Why have two properties (`sync` and `async`) representing a single state. And why aren't they both boolean?

boneskull:
I thought this got merged and spaced on it.  Oops.

boneskull:
we can ignore the coverage here because anything I'd do to increase it would be tangential to this PR.

plroebuck:
So TravisCI started [failing my builds](https://travis-ci.org/mochajs/mocha/jobs/471442055#L501-L508) now...

```console
nps is executing `lint.code` : eslint . "bin/*"
/home/travis/build/mochajs/mocha/lib/browser/growl.js
  145:16  error  Unexpected use of 'Date'. See https://github.com/mochajs/mocha/issues/237        no-restricted-globals
  151:3   error  Unexpected use of 'setTimeout'. See https://github.com/mochajs/mocha/issues/237  no-restricted-globals
✖ 2 problems (2 errors, 0 warnings)
```

As this file is _specifically_ for the browser, should this lint check apply?

PR #3376, opened 2018-05-04
3363: Throw error if unable to parse Mocha options file

### LINKED ISSUE #2576: Should not silently fail given --opts option

### LINKED ISSUE #3363: Throw error if unable to parse Mocha options file
### Description

The current code silently ignores any problems parsing the options file and continues on.

https://github.com/mochajs/mocha/blob/99c2f8aa7bf4184fdf694887ef740f31fe29f1f2/bin/options.js#L32-L47

Code should instead throw an error for any error encountered, except for a nonexistent default options file (i.e., user did not specify `--opts` and "test/mocha.opts" does not exist).

### Downside?

People who have errors in their "mocha.opts" file will experience failures until corrected.

plroebuck:
### Description of the Change

Code now throws an error for any problems loading the Mocha options file (except for nonexistent default "test/mocha.opts").

Builds off of previous discussions and work done in PR #2716 and PR #3352 .

Silently ignoring problems loading this file is just wrong.

### Benefits

Users will now be told if the file failed to load.

### Possible Drawbacks

Users with invalid Mocha options files (or permissions problems, etc.) will need to correct the problem.

### Applicable issues

Fixes #3363
Fixes #2576 
(server-major) [since Mocha could throw an error when it didn't previously]

### DIFF
--- a/bin/options.js
+++ b/bin/options.js
@@ -24,10 +24,12 @@ function getOptions() {
     return;
   }
 
-  const optsPath =
-    process.argv.indexOf('--opts') === -1
-      ? 'test/mocha.opts'
-      : process.argv[process.argv.indexOf('--opts') + 1];
+  const optsIndex = process.argv.indexOf('--opts');
+  const optsPathSpecified = optsIndex !== -1;
+  const defaultOptsPath = 'test/mocha.opts';
+  const optsPath = optsPathSpecified
+    ? process.argv[optsIndex + 1]
+    : defaultOptsPath;
 
   try {
     const opts = fs
@@ -37,12 +39,19 @@ function getOptions() {
       .filter(Boolean)
       .map(value => value.replace(/%20/g, ' '));
 
-    process.argv = process.argv
-      .slice(0, 2)
-      .concat(opts.concat(process.argv.slice(2)));
-  } catch (ignore) {
-    // NOTE: should console.error() and throw the error
+    if (opts.length > 0) {
+      process.argv = process.argv
+        .slice(0, 2)
+        .concat(opts.concat(process.argv.slice(2)));
+    }
+  } catch (err) {
+    // Default options file may not exist - rethrow anything else
+    if (optsPathSpecified || err.code !== 'ENOENT') {
+      console.error(`failed to load Mocha options file: ${optsPath}`);
+      throw err;
+    }
+  } finally {
+    // Despite its name, signifies loading was attempted and should not be again
+    process.env.LOADED_MOCHA_OPTS = '1';
   }
-
-  process.env.LOADED_MOCHA_OPTS = true;
 }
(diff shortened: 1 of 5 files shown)


plroebuck:
BTW, anyone run across [this ChromeHeadless problem](https://travis-ci.org/plroebuck/mocha/jobs/375058670) before? Same code ran file for all test jobs __except__ this last one. Restarting made no difference. MochaJS PR run afterwards was fine.

outsideris:
@plroebuck You didn't set saucelab in your travisci, see`Please set SAUCE_USERNAME and SAUCE_ACCESS_KEY env variables`.

plroebuck:
@outsideris, add "server-major" tag, pls!

boneskull:
will review some stuff tomorrow

outsideris:
@Bamieh I think this is a breaking change because users will get the errors when there is no opts file. Before this change, users don't care about the existence of the opts file.

Bamieh:
@outsideris makes sense, but looking at the code, it will only throw if there is an error in the existing file. if it does *not* exist, no errors will be throws. So i have a feeling that this is more of a bug fix rather than a breaking change.


- Users with no opt file will not be affected.
- Users with working opt files will not be affected.
- Users with broken opt files will have a new exception thrown that did not throw in the past.

Do you think there are people running mocha with a broken opt file? I believe its not that case because mocha would not be running as they expect anyways.

plroebuck:
Not exactly, @outsideris... It's a breaking change because an Error _could_ now occur where it does not in the current version. Even though the existing code is broken (due to masking), I consider it major since it _might_ disrupt the CI process by causing run failures.

Mocha now throws when user-specified `--opt` options file doesn't exist, cannot be read, or if a parsing error occurs due to its content; the default "test/mocha.opts" not existing will *not* trigger an error.

plroebuck:
| Do you think there are people running mocha with a broken opt file? I believe its not that case
| because mocha would not be running as they expect anyways.

Point being that they may not even know it's broken yet...
 #2576

plroebuck:
it's server-major if upgrading _can_ break something.
<https://blog.greenkeeper.io/introduction-to-semver-d272990c44f2>

boneskull:
This would be a bug if Mocha had *intended* to throw an error upon an invalid opts file, but failed to do so correctly.  

Since Mocha's current behavior is intended (which is to say, *nothing*) this isn't a bug fix.  But it *is* a breaking change.

plroebuck:
Had problems getting this branch updated post-UnexpectedJS. Opening new PR instead.

Issue #2989, opened 2017-09-06
server side version of init


tansaku:
### Prerequisites

* [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)

### Description

This is a feature suggestion - I was wondering about creating a version of `mocha init` that would work for server side code rather than client side, e.g. just create a test dir with a single test file that might have some boilerplate code in it ...

It could also update package.json to have:

`"test": "mocha test/**/*.js"`

I'm thinking about similar server side init things I've seen for the ruby projects bundler and rspec.
I've looked and I don't seem to see the existence or discussion of such a thing.  Would this be desirable for anyone?  I'd find it handy to have when starting up a new project to speed things up a little ...

boneskull:
I would be open to this idea.   Eventually this should expand to use a config file in place of `mocha.opts`. Thanks!

(I don't suggest creating `mocha.opts` with any `init` command)

outsideris:
See #3385 .

PR #3385, opened 2018-05-20
add a command to initialize node.js tests

outsideris:
### Description of the Change
add `init-test` command to initialize node.js test environments.

1. create `test` directory
    * if `test` directory already exists and it is not empty, skip it.
    * Otherwise, create `test` directory and first test file.
2. replace `npm test` script in `package.json`.
    * If there is no `package.json`, guide `npm init` first.

### Alternate Designs
I named it as `init-test`. `init` is client initialization. So, I don't want to break backward compatibility. I didn't think `server` is not fit for here because it is for node.js tests not only server-side.

The first test file's test is from mocha.org. It is a very simple test so it can be useless. Should we add more tests case as an example like promise?

I change `test` script of `package.json` via RegExp. It is not good readability and could not cover many cases of `package.json`. I don't want to add new dependencies. Is there another solution?

### Benefits
Newbies can adopt it more easier than now.
And we do it whenever making new js project with mocha.

### Applicable issues
#2989

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -284,6 +284,48 @@ program
     process.exit(0);
   });
 
+program
+  .command('init-test')
+  .description('initialize a node.js mocha setup at current working directory')
+  .action(() => {
+    let hasTestDir = false;
+    const defaultTestDIr = 'test';
+    try {
+      const testFiles = fs.readdirSync(join(cwd, defaultTestDIr));
+      if (testFiles.length < 1) {
+        hasTestDir = true;
+      }
+    } catch (ignore) {}
+
+    if (!hasTestDir) {
+      const mkdir = require('mkdirp');
+      mkdir.sync(defaultTestDIr);
+      const test = fs.readFileSync(
+        join(__dirname, '..', 'lib/template.js.tmpl')
+      );
+      fs.writeFileSync(join(defaultTestDIr, 'index.js'), test);
+    } else {
+      console.log('test directory is not empty. Skip to create test files');
+    }
+
+    try {
+      const pkg = fs.readFileSync('package.json', 'utf8');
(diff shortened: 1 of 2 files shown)


Bamieh:
I do not favor adding this to mocha to be honest.

- The documentation should be enough of a guidance.
- The steps are very tiny to require a bootstrap command, or a learning curve.
- Is it worth the extra maintenance? (docs, code, issue support, and bugs)

Bamieh:
@outsideris lets hear the team's opinion first

bin/_mocha
```diff
@@ -284,6 +284,48 @@ program
...
+      const pkg = fs.readFileSync('package.json', 'utf8');
+      const pkgRegExp = /"scripts"\s*:\s*{\n*(.*\n)*?(\s*"test"\s*:\s*"(.*)")(.*\n)?.*}/gm;
+      let testScript = pkgRegExp.exec(pkg);
+      if (testScript && testScript.length > 3) {
+        testScript = testScript[3];
+      }
+      const newPkg = pkg.replace(
+        testScript,
+        `mocha --recursive \\"./${defaultTestDIr}/*.js\\"`
+      );
```
Munter: This should do the same, but in a much more predictable way
```js
const package = require('./package.json');
package.scripts = package.scripts || {};
package.scripts.test = `mocha --recursive \\"./${defaultTestDIr}/*.js\\"`;
fs.writeFileSync('package.json', JSON.stringify(package, undefined, 2);
```

Munter:
I'd prefer if scaffolding like this was in a separate package. It's really none of mocha core's business.

Scaffolding has become really easy to run with npx these days. So `npx mocha-scaffold` is as easy to copy/paste as `mocha init-test`

outsideris:
I believe that we get a consensus that this scaffolding is not needed in our core. A separate package is a good idea, but I don't think we should make it now because it is too simple.
We can make scaffolding package later when users want to a couple of features about scaffolding.

PR #3388, opened 2018-05-22
Added unloading files feature from require cache

WithoutCaps:
### Description of the Change

Added a feature which allows unloading test files from `require` cache. 
This feature allows reusing data/objects since there is no need to create a new Mocha process.
(Feature is orientated to those that are accessing Mocha as a module, not CLI)

### Alternate Designs

1. The user would have to implement same logic in his own code
2. The user would have to create a new Mocha process on each dataset modification

1. Because it allows reuse of objects
2. Because it already has `loadFiles` function, so I would think that there should be a function that does the opposite.

### Benefits

Better reusability, fewer headaches for programmers

### Possible Drawbacks

None as far as I know

### Applicable issues

This is just a small code enhancement, has no impact on current production code.

### DIFF
--- a/test/node-unit/mocha.spec.js
+++ b/test/node-unit/mocha.spec.js
@@ -1,37 +1,43 @@
 'use strict';
 
 var Mocha = require('../../lib/mocha');
+var Path = require('path');
 
 describe('.unloadFile()', function() {
   it('should load and unload file to/from cache', function() {
     var mocha = new Mocha({});
     var filePath = __filename;
+
     mocha.addFile(filePath);
     mocha.loadFiles();
 
-    expect(
-      require.cache[require.resolve(filePath)] !== undefined,
-      'to be',
-      true
-    );
+    expect(require.cache, 'to have property', require.resolve(filePath));
     mocha.unloadFile(filePath);
-    expect(require.cache[require.resolve(filePath)], 'to be', undefined);
+    expect(require.cache, 'not to have property', require.resolve(filePath));
   });
 });
 
 describe('.unloadFiles()', function() {
   it('should unload all test files from cache', function() {
     var mocha = new Mocha({});
-    var filePath = __filename;
-    mocha.addFile(filePath);
-    mocha.loadFiles();
+    var testFiles = [
+      __filename,
+      Path.join(__dirname, 'fs.spec.js'),
+      Path.join(__dirname, 'color.spec.js')
+    ];
 
-    expect(
-      require.cache[require.resolve(filePath)] !== undefined,
-      'to be',
-      true
-    );
+    testFiles.forEach(function(file) {
+      mocha.addFile(file);
+    });
+
+    mocha.loadFiles();
+    testFiles.forEach(function(file) {
+      expect(require.cache, 'to have property', require.resolve(file));
+    });
+
     mocha.unloadFiles();
-    expect(require.cache[require.resolve(filePath)], 'to be', undefined);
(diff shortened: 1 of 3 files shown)


WithoutCaps:
@plroebuck Fixed.
I did it in that way initially because I prefer to jump around the _dynamic_ code as little as possible

test/node-unit/mocha.spec.js
```diff
@@ -0,0 +1,37 @@
...
+var Mocha = require('../../lib/mocha');
+
+describe('.unloadFile()', function() {
+  it('should load and unload file to/from cache', function() {
+    var mocha = new Mocha({});
+    var filePath = __filename;
+    mocha.addFile(filePath);
+    mocha.loadFiles();
+
+    expect(
```
boneskull: I think you can write something like this, which IMO is more appropriate:

```js
expect(require.cache, 'to have property', require.resolve(filePath));
```

then later

```js
expect(require.cache, 'not to have property', require.resolve(filePath));
```

test/node-unit/mocha.spec.js
```diff
@@ -0,0 +1,37 @@
...
+      'to be',
+      true
+    );
+    mocha.unloadFile(filePath);
+    expect(require.cache[require.resolve(filePath)], 'to be', undefined);
+  });
+});
+
+describe('.unloadFiles()', function() {
+  it('should unload all test files from cache', function() {
```
boneskull: Perhaps this test would be more meaningful if we added multiple files, then made the assertion they were all removed by looping through `mocha.files`.

boneskull:
Hi, thanks for the PR.

The intent here seems to be to create a public API; methods shouldn't be marked private, and they should have complete JSDoc docstrings and be marked public.

Moreover, Mocha must consume this API somehow, or it simply adds maintenance burden with little benefit.  The only place currently where it *could* and *should* be used is in `bin/_mocha` L607.

What's the use case for this, anyhow?

WithoutCaps:
Unfortunately, I don't think that I would be able to use this feature in existing codebase, but what it does is: it improves public API and allows to reuse existing Mocha instance (In terms of when you need to change Mocha dataset)

plroebuck:
@boneskull wrote:
> Moreover, Mocha must consume this API somehow, or it simply adds maintenance burden with little benefit.

While I _definitely_ understand this sentiment, `Mocha.prototype.unloadFiles()` would appear to be the missing half of the public API referenced [here](https://github.com/mochajs/mocha/blob/d1902a20663a6766ffaae646f155a884fcf3e141/lib/mocha.js#L579-L585) that allows programmatic reuse of the Mocha instance.

@WithoutCaps:
Having `Mocha.prototype.unloadFiles()` would seem perfectly adequate for retesting by itself.
I find the "per file" unload method much less useful for the public API myself; partial reloads
seem unnecessary. Is there a _strong_ use case for `Mocha.prototype.unloadFile(file)`?

It would seem appropriate make `Mocha.prototype.unloadFiles()`, `Mocha.prototype.loadFiles()`, and `Mocha.prototype.run()` documentation reference each other.

```js
/**
 * Unloads registered files.
 * This will remove them from Node's `require` cache, allowing them to be safely reloaded.
 *
 * @public
 * @see Mocha#loadFiles
 * @see Mocha#run
 */
Mocha.prototype.unloadFiles = function() {
```

plroebuck:
See PR #3534.

PR #3395, opened 2018-05-28
fix runner to emit start/end event

### LINKED ISSUE #2753: unable to listen to start event during programmatic run
The 'start' event is emitted synchronously downstream of the `mocha.run` method. Thus, if you are attaching a listener to that event after calling `mocha.run` (to get the runner instance), you miss the event:

```js
var runner = mocha.run(function (failures) {
  process.on('exit', function () {
    process.exit(failures);  // exit with non-zero status if there were failures
  });
});
runner.on('start', function (Evt) {
  //never called
});
```

Maybe there is some other way to attach event listeners to the runner prior to calling run, but I haven't found it yet.

outsideris:
### Description of the Change
Using mocha programmatically, the mocha runner doesn't emit start/end event correctly.
Because both events emitted before event listeners are registered. So, this change is fixed to emit events in event loop.

### Alternate Designs
I have no idea.

This is a bug.

### Benefits
Users can get start/end event correctly.

### Applicable issues
close #2753

### DIFF
--- a/test/unit/mocha.spec.js
+++ b/test/unit/mocha.spec.js
@@ -34,6 +34,16 @@ describe('Mocha', function() {
         });
       }
     );
+
+    it('should emit start event', function(done) {
+      var mocha = new Mocha(blankOpts);
+      mocha.run().on('start', done);
+    });
+
+    it('should emit end event', function(done) {
+      var mocha = new Mocha(blankOpts);
+      mocha.run().on('end', done);
+    });
   });
 
   describe('.addFile()', function() {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -851,10 +851,15 @@ Runner.prototype.run = function(fn) {
       filterOnly(rootSuite);
     }
     self.started = true;
-    self.emit('start');
+    Runner.immediately(function() {
+      self.emit('start');
+    });
+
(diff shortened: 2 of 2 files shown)


test/unit/mocha.spec.js
```diff
@@ -34,6 +34,20 @@ describe('Mocha', function() {
         });
       }
     );
+
+    it('should emit start event', function(done) {
+      var mocha = new Mocha(blankOpts);
+      mocha.run(function() {}).on('start', function() {
+        done();
+      });
+    });
```
plroebuck: Couldn't this simplify?
```js
      mocha.run().on('start', done);
```

test/unit/mocha.spec.js
```diff
@@ -34,6 +34,20 @@ describe('Mocha', function() {
...
+
+    it('should emit end event', function(done) {
+      var mocha = new Mocha(blankOpts);
+      mocha.run(function() {}).on('end', function() {
+        done();
+      });
+    });
```
plroebuck:
```js
      mocha.run().on('end', done);
```

plroebuck:
@outsideris, issue #3569 may be implicating your PR. Can you confirm?

Issue #2657, opened 2017-01-03
enhancement: new interface to support arrow functions


alexreardon:
Currently functions that use `() =>` cannot take advantage of the mocha context: http://mochajs.org/#arrow-functions. 

However, arrow function usage could be improved by passing the context as an argument:

```js
it('should do something cool', (done, context) => {
  context.timeout = 5000;
  ...
});
```

The complicating factor is the `done` argument. The presence of which makes the test async. Perhaps if you want to use the `context` argument then you need to also call `done` when appropriate.

laggingreflex:
Came here for this, but I've also found https://github.com/skozin/arrow-mocha

eight04:
Here is a version which can work with mocha's `--require` option:
https://www.npmjs.com/package/mocha-context

boneskull:
Replacing the default `bdd` interface with this is a non-starter.

It's possible to create an interface which passes the context as the first parameter, but would likely require modifications to Mocha's core to support, due to hardcoded assumptions.

Would accept a PR that would allow the "async check" of an interface to be configurable.  Once that's ready, then it should be fairly trivial to provide an alternate interface.

eight04:
> allow the "async check" of an interface to be configurable.

What does it mean?

I think this feature could be implemented in 2 ways:

1. Create a new interface e.g. `bdd-arrow` that would [decorate the arrow function](https://github.com/eight04/mocha-context/blob/84715db337ec5fd690a190091c995e5774f76050/index.js#L11-L20) before [sending `fn` to `new Test`](https://github.com/mochajs/mocha/blob/0d95e3f41a0859be4e7225541575e78b6d8926ee/lib/interfaces/bdd.js#L85).
2. Add a new option e.g. `mocha --context-argument` that would change [how `runnable.fn` is called](https://github.com/mochajs/mocha/blob/0d95e3f41a0859be4e7225541575e78b6d8926ee/lib/runnable.js#L372).

boneskull:
@eight04 see #3399

JoshuaKGoldberg:
Yeah, this would be a pretty big change to Mocha that is already solved by at least one userland integration mentioned in this thread. Per #5027 we're not looking to do big shakeups any time soon.

Issue #1856, opened 2015-09-02
Bind ES6 arrow functions to Context


michielbdejong:
disclaimer: I'm quite new to Mocha and to arrow functions, so feel free to close this if it's nonsense. :)

Consider these three tests:
````js
test('Mocha Context available in normal ES5 functions', function() {
  expect(this.slow).to.be.a('function');
});
test('Mocha Context available ES6 arrow functions', () => {
  // Will fail, because it's bound to `window`:
  expect(this.slow).to.be.a('function');
});
test('Mocha Context available in nested, bound ES6 arrow functions', function() {
  (() => {
    expect(this.slow).to.be.a('function');
  })();
});
````
afaics this is because [Mocha uses `fn.call(ctx);`](https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L289) to execute the function passed to `test`, which does not have the desired effect in case of ES6 arrow functions, as you can [read here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Invoked_through_call_or_apply) and also see if you run this in your browser console:
````js
(() => { console.log(this); }).call({ foo: 'bar' });
// prints the Window object
(function() { console.log(this); }).call({ foo: 'bar' });
// prints { foo: 'bar' }
(function() { (() => { console.log(this); })(); }).call({ foo: 'bar' });
// prints { foo: 'bar' }
````
Looking especially at how that third test (arrow function wrapped inside a normal function) does pass, I think it could be fixed by wrapping the variable `fn` in a normal ES5 function before calling its `.call` method, i.e. replacing that `fn.call(ctx);` with `(function() { fn(); }).call(ctx);` on [line 289 of Runnable.js](https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L289)

michielbdejong:
Or the easier option would of course be to add a note to the documentation saying: Don't put your test in an arrow function if you plan to use `this.slow` or other methods from the Context. :)

developit:
@michielbdejong The documentation suggestion works, or Mocha could pass the test context as an argument in addition to setting it as `this`.

danielstjules:
I believe instead of `(function() { fn(); }).call(ctx);` you would need `(function() { return fn(); }).call(ctx);`

danielstjules:
@michielbdejong Your suggested fix doesn't seem to work in practice

``` javascript
var assert = require('assert');

function it(str, fn) {
[...]

PR #3399, opened 2018-05-30
initial implementation of "functional" interface

boneskull:
This is an initial attempt at a "functional" interface; those of us who prefer using arrow functions will hopefully appreciate this.  I've called it the `func` UI; here's how it looks:

# Usage

```js
'use strict';

describe('behavior', suite => {
  // just like this.timeout()
  suite.timeout(200);

  it('should work synchronously', test => {
    expect(1 + 1, 'to be', 2);
    expect(2 + 2, 'to be', 4);
    // test.done() must be called unless a Promise is returned
    test.done();
  });

  it('should work asynchronously', test => {
    expect(1 - 1, 'to be', 0);
    expect(2 - 1, 'to be', 1);
    process.nextTick(function() {
      test.done();
    });
  });

  it('should work with a Promise', async () => {
    expect(1 - 1, 'to be', 0);
    expect(2 - 1, 'to be', 1);
    // test.done() needn't be called
  });

  it('should work with context methods', async test => {
    test.timeout(400);
    expect(1 - 1, 'to be', 0);
    expect(2 - 1, 'to be', 1);
  });
  
  afterEach(hook => {
    // hooks have same behavior
    hook.timeout(200);
    hook.done();
  });

  afterEach(async () {
    // and also support Promises
  });
});
```

# Implementation Details

Mocha makes assumptions about how Runnables (Tests, Hooks) are executed; what function context they are called with, and what parameters.  To address this, I needed to decouple some parts of Mocha's core.

So, I've introduced a new concept: the Behavior.  We can think of this as a collection of mixins which may apply (currently) to Suites and Runnables.  These mixins introduce alternative behavior for certain methods within those objects.  This is *not* the mythical plugin API, but it could be a step in that direction.

There are currently a select few places where the Behaviors are delegated to, and I've broken out the default functionality into a Behavior (the `Default` one, fwiw).  

Due to the way Mocha's Suites configure their children (Suites, Hooks, and Tests), in which all children "inherit" the configuration of their parent Suite, an Interface (`bdd`, `tdd`, `qunit`, etc.) only need configure the Behavior on the *Root* Suite.  This configuration propagates throughout all Suites and children thereof.  

This means an Interface *can* provide an alternate Behavior.  It's not the *only* way to provide an alternate Behavior, but seems like a reasonable one; see the simplicity of `lib/interfaces/func.js`.  (Maybe the interface is where the `Functional` Behavior should live?)

A Behavior may have one or more methods and needn't implement all of them.  The default methods will always be present, and if multiple behaviors are configured, then the last one wins (think of how `Object.assign()` works).

Ultimately, I'm not married to this API.  I'm open to considering an event-driven model, as long as the scope doesn't blow up.  But once it's merged, we *are* married to it, so it behooves us to be confident about what we're doing.  

See #1856, #2657, probably others.

### DIFF
--- a/lib/behavior.js
+++ b/lib/behavior.js
@@ -1,0 +1,170 @@
+'use strict';
+
+/**
+ * Built-in behaviors.
+ * These provide hooks into modifying Mocha's behavior for different use cases.
+ */
+
+var utils = require('./utils');
+
+/**
+ * Default Mocha behavior
+ */
+exports.Default = {
+  Suite: {
+    /**
+     * Runs the suite; called by `Suite.create()`.  Calls its callback fn
+     * with the Suite as its context.
+     * @this {Suite}
+     */
+    run: function() {
+      // FUTURE: async suites
+      this.fn.apply(this);
+    }
+  },
+  Runnable: {
+    /**
+     * Determines whether or not we should provide a nodeback parameter and
+     * expect it to be called
+     * @this {Runnable}
+     * @returns {boolean} `true` if we should run the test fn with a callback
+     */
+    shouldRunWithCallback: function() {
+      return Boolean(this.fn && this.fn.length);
+    },
+    /**
+     * Runs the Runnable synchronously, or potentially returning a Promise
+     * @param {Function} done - Callback
+     */
+    run: function(done) {
+      var result = this.fn.call(this.ctx);
+      if (result && typeof result.then === 'function') {
+        this.resetTimeout();
+        result.then(
+          function() {
+            done();
+            // Return null so libraries like bluebird do not warn about
+            // subsequently constructed Promises.
+            return null;
+          },
+          function(reason) {
+            done(
+              reason || new Error('Promise rejected with no or falsy reason')
+            );
+          }
+        );
+      } else {
+        if (this.asyncOnly) {
+          return done(
+            new Error(
+              '--async-only option in use without declaring `done()` or returning a promise'
+            )
+          );
+        }
+        done();
+      }
+    },
+    /**
+     * Runs the Runnable, passing a nodeback function, which must be called to
+     * complete the test
+     * @param {Function} done - Callback
+     */
+    runWithCallback: function(done) {
+      var result = this.fn.call(this.ctx, function(err) {
+        if (
+          err instanceof Error ||
+          Object.prototype.toString.call(err) === '[object Error]'
+        ) {
+          return done(err);
+        }
+        if (err) {
+          if (Object.prototype.toString.call(err) === '[object Object]') {
+            return done(
+              new Error('done() invoked with non-Error: ' + JSON.stringify(err))
+            );
+          }
+          return done(new Error('done() invoked with non-Error: ' + err));
+        }
+        if (result && utils.isPromise(result)) {
+          return done(
+            new Error(
+              'Resolution method is overspecified. Specify a callback *or* return a Promise; not both.'
+            )
+          );
+        }
+        done();
+      });
+    }
+  }
+};
+
+/**
+ * Provides a test context in a functional manner for use with
+ * lambdas/arrow functions.
+ * All Runnables must either return a promise or call `runnable.done()`, where
+ * `runnable` is the first parameter to the Runnable's callback (`fn`)
+ */
+exports.Functional = {
+  Suite: {
+    /**
+     * Runs the Suite.  Calls its `fn` with NO context and the suite itself
+     * as the first parameter.
+     * @this {Suite}
+     */
+    run: function(opts) {
+      this.fn.call(null, this);
+    }
+  },
+  Runnable: {
+    /**
+     * Determines whether or not we should provide a nodeback parameter and
+     * expect it to be called; always false
+     * @this {Runnable}
+     * @returns false
+     */
+    shouldRunWithCallback: function() {
+      return false;
+    },
+
+    /**
+     * Runs the Runnable expecting a call to ctx.done() or a Promise
+     * @param {Function} done - Callback
+     */
+    run: function(done) {
+      this.ctx.done = function(err) {
+        if (
+          err instanceof Error ||
+          Object.prototype.toString.call(err) === '[object Error]'
+        ) {
+          return done(err);
+        }
+        if (err) {
+          if (Object.prototype.toString.call(err) === '[object Object]') {
+            return done(
+              new Error('done() invoked with non-Error: ' + JSON.stringify(err))
+            );
+          }
+          return done(new Error('done() invoked with non-Error: ' + err));
+        }
+        done();
+      };
+      var result = this.fn.call(null, this.ctx);
+      if (result && typeof result.then === 'function') {
+        this.resetTimeout();
+        result.then(
+          function() {
+            done();
+            // Return null so libraries like bluebird do not warn about
+            // subsequently constructed Promises.
+            return null;
+          },
+          function(reason) {
+            done(
+              reason || new Error('Promise rejected with no or falsy reason')
+            );
+          }
+        );
+      }
+    }
+  }
+};
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -4,6 +4,7 @@ var Pending = require('./pending');
 var debug = require('debug')('mocha:runnable');
 var milliseconds = require('./ms');
 var utils = require('./utils');
+var behavior = require('./behavior');
 
 /**
  * Save timer references to avoid Sinon interfering (see GH-237).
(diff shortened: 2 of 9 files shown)


boneskull:
lol, guess I can't use `Object.assign`... I'll fix that.

boneskull:
this needs

- [ ] docs
- [ ] an integration test to see what happens when e.g. `test.done()` is called in a `Promise`-returning Test
- [ ] more unit tests, probably

wavebeem:
imo this is exactly what i would expect an arrow-func-centric Mocha to look like!

eight04:
Is it possible to omit explicit `ctx.done()` for sync tests/hooks? Suppose that the test is sync if the return value is not a promise? If `ctx.done()` is required for sync tests, all existed tests have to be modified to switch from `bdd` to `func`:
```diff
describe("...", () => {
-  it("...", () => {
+  it("...", t => {
    // some asserts...
+    t.done();
  });
});
```

boneskull:
if we didn’t require it then how would we know a test is async? if we supported a `test, done` signature, then every async test that didn’t need to use methods on `test` would have `test` anyway. 

probably just as effective to use`async` keyword instead of `test.done()` ..

eight04:
Something like this?
```js
describe("...", () => {
  it("...", t => {
    t.async(true);
    setTimeout(t.done, 200);
  });
});
```

lib/behavior.js
```diff
@@ -0,0 +1,170 @@
...
+ * Built-in behaviors.
+ * These provide hooks into modifying Mocha's behavior for different use cases.
+ */
+
+var utils = require('./utils');
+
+/**
+ * Default Mocha behavior
+ */
+exports.Default = {
```
Bamieh: Can we rename this from `Default` to something else?
- It is a little confusing with es6 default modules.
- Whenever we want to change the default behavior to another one, we have to refactor 2 places now.
- A semantic name indicating what it does is more meaningful IMO. `noCtxAsArg` or something in this context (pun intended 😄 ).

lib/behavior.js
```diff
@@ -0,0 +1,170 @@
...
+    shouldRunWithCallback: function() {
+      return Boolean(this.fn && this.fn.length);
+    },
+    /**
+     * Runs the Runnable synchronously, or potentially returning a Promise
+     * @param {Function} done - Callback
+     */
+    run: function(done) {
+      var result = this.fn.call(this.ctx);
+      if (result && typeof result.then === 'function') {
```
Bamieh: I know this code is just moved, but if we wrap the result in `Promise.resolve(result).then(...)` instead of doing this check, we'll always have the returned value treated as a promise, without the need for the extra hassle.

lib/runnable.js
```diff
@@ -439,3 +380,48 @@ Runnable.prototype._timeoutError = function(ms) {
...
+};
+
+/**
+ * Calls a behavior method by name with positional args, `Function.prototype.call` style
+ * @param {string} name - Behavior method name
+ * @param {...*} [arg] - Arguments
+ * @private
+ * @returns {*} Whatever the behavior method returns
+ */
+Runnable.prototype.callBehavior = function(name) {
```
Bamieh: not sure about backward compatibility but can't we use `(name, ...args)`?

test/interfaces/func.spec.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+var Promise = require('es6-promise');
```
Bamieh: why is this needed? Promises are supported for all node versions we support (anything above 4)

Bamieh:
I wonder if I can use this feature to pass a `client` to the functions (wd client for native testing in my projects).

lib/behavior.js
boneskull: This would require adding a Promise polyfill to the production bundle.

lib/runnable.js
boneskull: No, Mocha is written in ES5 :smile:

test/interfaces/func.spec.js
boneskull: IE11

boneskull:
@wavebeem We *do* check for a `Promise` return value, but not all async behavior is `Promise`-based.

@Bamieh I'm not entirely sure what you're asking, but if I had to guess, it's [here](https://github.com/mochajs/mocha/pull/3399/files#diff-29f7114119fa36dd12e0dffc6097337eR73).  We call `shouldRunWithCallback` to determine which of the two methods (`run` or `runWithCallback`) we call.  Mind you, this is *not* applicable to the `func` interface, which (currently) will always pass a single parameter.

@eight04 Is there a precedent for that--does another test framework do it that way?  I'm basing my implementation on stuff like [tape](https://npm.im/tape), which requires `t.end()` (or `t.plan()`, which we can't support) to be called unless a `Promise` is returned.

This is why I mentioned it'd actually be easier to use the `async` keyword for otherwise-synchronous tests:

```js
it('should synchronously assert', t => {
  assert(true);
  t.done(); // might as well alias t.end() to this if we do it
});
```

is effectively:

```js
it('should synchronously assert', async t => {
  assert(true);
});
```

FWIW, the Jasmine/Jest strategy seems to actually make timers `Promise`-returning or synchronous (not sure which) by monkeypatching them. I don't want that to be a choice Mocha makes for its users; libraries like [Sinon](https://npm.im/sinon) can be added to make timers synchronous, if needed.

lib/behavior.js
boneskull: yeah, probably a good idea to rename this, or just use `module.exports` and put the `Functional` behavior within the `func` interface... what do you think of that?

lib/behavior.js
boneskull: (I'm not necessarily offering an opinion on that; I'm just telling you why Mocha historically does it this way)

lib/behavior.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+  },
+  Runnable: {
+    /**
+     * Determines whether or not we should provide a nodeback parameter and
+     * expect it to be called
+     * @this {Runnable}
+     * @returns {boolean} `true` if we should run the test fn with a callback
+     */
+    shouldRunWithCallback: function() {
```
boneskull: So, IMO, this is too specific.

What may be better is supporting a function which answer the question "what arguments should we pass to the execution function?", and another "what context should the execution function run with?"

Obviously we're going to need to document this specification...

test/interfaces/func.spec.js
boneskull: though, I don't think this test actually runs in the browser.  it should.

lib/behavior.js
boneskull: I'm actually really annoyed about IE11/ES5, so maybe we can just write Mocha in ES6 going forward and pipe the distfile thru Babel...

eight04:
> Is there a precedent for that--does another test framework do it that way?

No. Actually, I think `(test, done)` is better. It matches the old interface nicely.

The original purpose of *passing the context as the first argument* is to solve the problem that *users can't access `.skip()`, `.timeout()` when using arrow functions*. I don't think people would like to change all tests into async function just to gain the ability to access `.skip()`, `.timeout()` for maybe one or two sub-tests.

I prefer `(test, done)`. Reasons:

1. All existed sync tests don't have to be modified.
2. All existed promise-based async tests don't have to be modified.
3. All existed callback-based async tests just need to change the signature. (`done` -> `(t, done)`)
4. [arrow-mocha](https://www.npmjs.com/package/arrow-mocha) users can migrate to the new UI without modification.

If we still want to keep async as the default, it is probably worth adding an [`autoend` option](https://www.node-tap.org/api/#tautoendvalue) that would automatically `done()` the test when the return value is not a promise, just like current mocha.

lib/behavior.js
```diff
@@ -0,0 +1,170 @@
...
+              new Error('done() invoked with non-Error: ' + JSON.stringify(err))
+            );
+          }
+          return done(new Error('done() invoked with non-Error: ' + err));
+        }
+        done();
+      };
+      var result = this.fn.call(null, this.ctx);
+      if (result && typeof result.then === 'function') {
+        this.resetTimeout();
```
outsideris: I think this line should ve moved above `if` statement.
Because tests just exit if `test.done()` is not called if synchronous tests. Also we should fix [the error message](https://github.com/mochajs/mocha/blob/5fc58458395d8a9bc75daae97042af78dc94cc7a/lib/runnable.js#L436) for func ui.

lib/behavior.js
```diff
@@ -0,0 +1,170 @@
...
+    shouldRunWithCallback: function() {
+      return false;
+    },
+
+    /**
+     * Runs the Runnable expecting a call to ctx.done() or a Promise
+     * @param {Function} done - Callback
+     */
+    run: function(done) {
+      this.ctx.done = function(err) {
```
boneskull: I don't love this monkeypatching; `Context` should have a method which accepts a callback and assigns it to its own `done` method.

noelmace:
@boneskull do you think there is a chance you work on this approach again? Could you explain why it was postponed / what's currently blocking it? I would be happy to help if I can.

JoshuaKGoldberg:
🤖 Closing out old PRs to keep the review queue small. Cheers!

PR #3400, opened 2018-05-30
update packages per "npm audit fix"

boneskull:
there were vulernabilities... now there are fewer.


boneskull:
Something is busted on Netlify...

boneskull:
It's not related to this, though.

Issue #3369, opened 2018-05-02
Remove bower installation in docs


outsideris:
In [docs/index.md](https://github.com/mochajs/mocha/blob/master/docs/index.md#installation), we should remove installing mocha via bower because mocha no longer supports bower; see [the comments](https://github.com/mochajs/mocha/pull/3364#discussion_r185073819).

fesebuv:
It seems like this issue was addressed by this pr https://github.com/mochajs/mocha/pull/3405/files. Perhaps this issue should be resolved?
@outsideris

outsideris:
@fesebuv You're right. Thanks.

PR #3364, opened 2018-04-30
Drop node.js v4 support

outsideris:
### Description of the Change
As [Node.js announcement](https://medium.com/the-node-js-collection/april-2018-release-updates-from-the-node-js-project-71687e1f7742) and [Node.js release schedule](https://github.com/nodejs/Release), Node.js v4 is end of life April 30th.
Mocha can drop Node.js v4 support.

I checked ["Breaking changes between v4 LTS and v6 LTS"](https://github.com/nodejs/node/wiki/Breaking-changes-between-v4-LTS-and-v6-LTS), but I didn't find anything to change in code level.

### Alternate Designs
N/A

### Why should this be in core?
It's minimum Node.js version.

### Benefits
Mocha doesn't care issues with Node.js v4 anymore.

### Possible Drawbacks
N/A

### Applicable issues
close #3149

PR #3405, opened 2018-06-06
Remove bower install instructions

goteamtim:
Remove installing mocha via bower instruction because mocha no longer supports bower installs.

### Description of the Change

Remove installing mocha via bower instruction because mocha no longer supports bower installs.  See the [comments](https://github.com/mochajs/mocha/pull/3364#discussion_r185073819)

### Alternate Designs

N/A

Mocha no longer supports bower installs

### Benefits

N/A

### Possible Drawbacks

N/A

### Applicable issues

#3369


boneskull:
Thanks!

PR #3408, opened 2018-06-06
adding a "retry" event for a test that fails but can be retried; closes #2529

### LINKED ISSUE #2529: Can't see how retried test failed before passing.

### LINKED ISSUE #2592: split CI matrix on browser
Currently we're running all browsers in a single build; let's split that up.

boneskull:
Continuation of #2926.  

Addresses issue #2529 by allowing reporters to take action on tests that have failed but can be retried

cc @catdad


outsideris:
I think #2529 is the right issue.

outsideris:
BTW, is there documents for events? 
I couldn't find them in [our API document](https://mochajs.org/api/).

lud2k:
Hiii! I'm really excited by this commit. Any chance it will be merged soon?

plroebuck:
Even if our reporters _don't_ support the event, shouldn't the statistics collector do so?

plroebuck:
Perhaps the event name should also be changed - maybe "test retry" is more appropriate?

plroebuck:
@outsideris wrote:
> BTW, is there documents for events?
> I couldn't find them in [our API document](https://mochajs.org/api/).

Have work in progress for API event documentation.
Working with @Munter on some Jekyll (??) issues generating event-related JSDoc for site correctly.

Munter:
Looks like jsdoc generates broken output when using their `@events` syntax :'(

catdad:
@plroebuck I am not familiar with the statistics collector. I'd be happy to update it if you can point me in the right direction.

As for the event name, I tried following the convention of some other names I saw, but this was a very long time ago. I'd be happy to change that as well. I don't know that "test retry" totally represents what the event is. The event specifies that a test failed, but will be retried. To me, "test retry" means that a retry has occurred (which isn't false), rather than that a retryable test has failed and is providing me with the failure information.

shridharkalagi:
Any update on this feature? Is this getting merged?

boneskull:
renamed this event to `retry` and fiddled with the tests

boneskull:
the coverage decrease is nonsensical

PR #3417, opened 2018-06-15
Adding support for --ignore-timeout flag

patrick-rodgers:
### Description of the Change

This change adds support for a new flag, --ignore-timeouts, which when provided will not treat timeouts as fails, but rather as a new category of "safe timeouts". This flag can also be supplied in any of the scopes (hook, test, suite) using the ignoreTimeouts method:

```js
describe('a suite of tests', function() {
  beforeEach(function(done) {
    this.timeout(3000); // A very long environment setup.
    this.ignoreTimeout(true);
    setTimeout(done, 2500);
  });
});
```

When set a different emit event is sent, "safe timeout". This PR also updates all reporters to understand this new emit and behave appropriately for their given usage.

### Alternate Designs

Considered disabling timeouts, but in some cases I have a need for an operation to have a timeout (perhaps it hangs) but can in those cases ignore this to not fail the entire test suite, potentially blocking release. I.e. there are some known long running operations that very occasionally fail in testing and can be safely ignore.

Functionality was not present, I would like it to be. I believe it can apply to others as well.

### Benefits

New capability to still allow tests/suites/hooks to timeout, but continue testing and not fail an entire build due to one possibly long running test.

### Possible Drawbacks

Could mask a real fail if the timeout is due to a bug. Any reporters not in the core package would not understand the new emit event.

### Applicable issues

Came from me asking a question on gitter and looking through the code and determining the option I wanted was not available.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -223,6 +223,7 @@ program
     'force shutdown of the event loop after test run: mocha will call process.exit'
   )
   .option('--no-timeouts', 'disables timeouts, given implicitly with --debug')
+  .option('--ignore-timeouts', 'do not treat timeouts as fails')
   .option('--no-warnings', 'silence all node process warnings')
   .option('--opts <path>', 'specify opts path', 'test/mocha.opts')
   .option('--perf-basic-prof', 'enable perf linux profiler (basic support)')
@@ -409,6 +410,12 @@ if (program.timeout) {
   mocha.suite.timeout(program.timeout);
 }
 
+// --ignore-timeouts
+
+if (program.ignoreTimeouts) {
+  mocha.suite.ignoreTimeouts(program.ignoreTimeouts);
+}
+
 // --bail
 
 mocha.suite.bail(program.bail);
@@ -488,8 +495,8 @@ if (program.forbidPending) mocha.forbidPending();
 // custom compiler support
 
 if (program.compilers.length > 0) {
-  require('util').deprecate(() => {},
-  '"--compilers" will be removed in a future version of Mocha; see https://git.io/vdcSr for more info')();
+  require('util').deprecate(() => { },
+    '"--compilers" will be removed in a future version of Mocha; see https://git.io/vdcSr for more info')();
 }
 const extensions = ['js'];
 program.compilers.forEach(c => {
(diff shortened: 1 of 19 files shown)


patrick-rodgers:
I think I did get the CLA signed? There was a hickup with my email I guess not being public or something. That is all cleared up now.

boneskull:
Hi, thanks for the PR.  I'm not sure I'm convinced this is necessary, and worried it may cause more harm than good, so I have some questions:

1. What is the purpose of the code that has the timeout problem?  
1. Where in your test suite(s) is this code being called?
1. Does something like this work?
   ```js
   // n is however long you want to wait in ms
   before(function(done) {
     this.timeout(0);
     let t = setTimeout(done, n);
     doSomethingSlow(() => {
       if (t) { 
         clearTimeout(t);
         done();
       }
     });
   });
   ```

The problem is that when a test "times out", the execution environment becomes unstable.  There will be one or more tasks remaining in the queue which will be executed if/when the test code actually finishes.  This code may throw an exception or otherwise do literally anything--including cause a false positive in a subsequent test!

In other words, we can't cleanly recover from a timeout; even if *you* have no further tests, Mocha still has to run its own code in the unstable environment.  This is why (IMO) a timeout should *always* be considered an error.

patrick-rodgers:
Howdy,

Its your project so I will leave it up to you on making this call if this is necessary. It is something I thought was useful so I submitted the work back in this PR.

1. Here is [one example where I am using it](https://github.com/pnp/pnpjs/blob/dev/packages/sp/tests/web.test.ts#L214). In this test we create a SharePoint site and then delete that same site. This is a case where I am 99% sure that code will be working, but every so often it times out. I was getting a lot of failed CI builds on this and a few other long running cases.
2. https://github.com/pnp/pnpjs/blob/dev/packages/sp/tests/web.test.ts#L214
3. Sure, I can find other ways to work around the timeout issue but thought this might help others and add something to the library.

You (obviously :)) know the library and how it works better than me, so if there are cases where this would cause problems that is no good. It doesn't seem to have any side effects in my tests, but they are by no means representative of what everyone in the world is doing. IF the model is a timeout has to be an error this PR shouldn't be merged.

No worries whichever way you want to go with it.

patrick-rodgers:
Accidentally hit close.

boneskull:
It's not causing problems (in part) because you may have an unhandled `Promise` rejection in the case of a post-timeout failure; this isn't the same as an asynchronously thrown uncaught exception, which should always be unrecoverable. 

If we could somehow restrict this to `Promise`-returning functions, then it may be safe to do (depending on Node.js' future treatment of unhandled rejections), but I don't think we can do that in a way which wouldn't be too confusing for the user.

While I appreciate the PR, and understand the strategy works for your use case, I don't think we can add this to Mocha in a way which wouldn't create a new footgun for users.

patrick-rodgers:
Right on, thanks for all your work, mocha is fantastic, looking forward to learning the new features in the upcoming release. Cheers.

PR #3418, opened 2018-06-18
remove play from bin/_mocha

Bamieh:
I dont see this function used anywhere, also stop now just spits `\u001b[2K` to the stdout for the rerun, I am not sure if it is needed as well.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -99,26 +99,12 @@ const showCursor = () => {
 };
 
 /**
- * Stop play()ing.
+ * Stop.
  */
 const stop = () => {
   process.stdout.write('\u001b[2K');
-  clearInterval(play.timer);
 };
 
-/**
- * Play the given array of strings.
- */
-const play = (arr, interval) => {
-  const len = arr.length;
-  interval = interval || 100;
-  let i = 0;
-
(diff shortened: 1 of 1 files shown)


bin/_mocha
```diff
@@ -99,26 +99,12 @@ const showCursor = () => {
...
-const play = (arr, interval) => {
-  const len = arr.length;
-  interval = interval || 100;
-  let i = 0;
-
-  play.timer = setInterval(() => {
-    const str = arr[i++ % len];
-    process.stdout.write(`\u001b[0G${str}`);
-  }, interval);
-};
```
outsideris: There are two blank lines. So, linting is failed.

Bamieh: fixed

Bamieh:
@outsideris do you have any idea what `\u001b[2K` does? i know is unicode but im unable to find it out.

outsideris:
@Bamieh 
I didn't know about that.

As [SO](https://stackoverflow.com/a/13801345), `\u001b` means `ESC`. 
As [ANSI Escape sequences](http://ascii-table.com/ansi-escape-sequences-vt-100.php), `Esc[2K` means `Clear entire line`.

Issue #2578, opened 2016-11-10
Make CLI more reusable


rijnhard:
Hi 

In our use case mocha is basically the underlying engine for a behavioural framework we wrote that tests a bunch of async systems we use in-house. 

We did a lot, but for brevity we write the behaviours in mocha specs, and use it to assert the systems output during a test case run. This allows us to test that the system is behaving the way that it should since it's basically a giant async state machine constantly spewing out state change notifications.

In our test case we want our framework to work as much as is possible with other mocha tools.
The problem is that using the `Runner` class you basically lose all of the third party tool support unless you want to reimplement the entire mocha cli on your own. 

I have used `Commander` extensively in many other projects, so I have figured out how to make the CLI more reusable. 

More details are in the PR (https://github.com/mochajs/mocha/pull/2553) 

I am willing to make any changes requested myself. And the functionality of the cli has not changed, so if you choose not to expose a CLI component for reuse then this can be looked at as just a code cleanup. 

I know I shouldn't have made these changes without speaking to you first, but as the PR says I didn't even know if it would be possible or work, so I dug into it face first and by the time I proved it could be done, I had it working. 

FYI I am using my PR in said testing framework and its working well.

sgen:
I would be a little weary of Commander.js at this point. There are [nasty](https://github.com/tj/commander.js/issues/404) [bugs](https://github.com/tj/commander.js/issues/552) in it right now. The original author seems to be quite busy with other projects and hasn't touched it in quite some time. The other contributors seem quite busy as well with other projects.

rijnhard:
@sgen I'm aware of https://github.com/tj/commander.js/issues/404 ironically i ran into it with the `opts` option overriding the `Command.opts()` function, had to do some magic to get it to work as expected in this PR, I did open an issue in commander (https://github.com/tj/commander.js/issues/584) for that. 

But you raise a valid point, the other option is yargs. Which is fairly similar and the maintainers are more active but not as fast in some cases due to their use of `read-pkg-up`.

rijnhard:
It probably won't be toooo hard for me to swap one out for the other.

sgen:
I've never used yargs so I cannot comment.

I fully support migrating the some of the mocha cli logic to the main library. Ideally all tool functionality should be included in the library, with the cli tool being a simple wrapper to decode cli input.

But Im not a Mocha dev so you should wait for them to weigh in.

sgen:
[...]

Issue #2292, opened 2016-06-01
Merge bin/mocha and bin/_mocha


segrey:
Tools providing integration with mocha need to use `bin/_mocha` sometimes. For example, debugging mocha test is possible with `bin/_mocha` only, because it's impossible to pass `--debug-brk=<port>` option to _mocha child process from `bin/mocha`.
Well, it's easy for IDE to use `bin/_mocha` under the hood, but it might be not so obvious for users who debug mocha tests using node-inspector. Though, there is a [hint](https://www.npmjs.com/package/node-inspector#how-do-i-debug-mocha-unit-tests), one need to take it into account. Also, debugging npm script that runs mocha tests leads to the same confusion (e.g. this WebStorm [issue](https://youtrack.jetbrains.com/issue/WEB-21853)).

Related to https://github.com/mochajs/mocha/issues/1969#issuecomment-161473791 //cc @boneskull
Please merge `bin/_mocha` and `bin/mocha` to have single public main script file in mocha3. Thanks.

boneskull:
Yes, having one (1) executable is definitely on the todo list.

ScottFreeCode:
See also #2517, the `--node-arg` flag in `npx` (which is bundled with newer versions of `npm`), and the ability to just say `node <arguments> node_modules/mocha/bin/mocha`.

segrey:
Would be great to have a single executable.

PR #3420, opened 2018-06-20
split _mocha file

Bamieh:
## The change
`bin/_mocha` file is responsible for both configuring mocha from the cli flags and running mocha. This causes an issue, every tool that wants to add on top of the mocha cli must copy-paste the `bin/_mocha` code in their own repos.

This change separates the `bin/_mocha` file to two files.
- `bin/base.js` file which collects the cli flags and creates a configured mocha instance.
- `bin/_mocha` file which requires `bin/base.js` and runs the exported mocha instance.

## Advantages:
- Libraries now can require `mocha/bin/base` instead of copy-pasting the code.
- I also needed this for `mocha-watcher`.
- separate concerns for easier code comprehension and maintenance.
- non-breaking change, just a win-win


outsideris:
Is this related with #2292 or #2578 ?

outsideris:
Why the coverage decreated like that?

Bamieh:
@outsideris i'm not sure! i added the file to the ignore but still, i'll look into it tomorrow

boneskull:
I'm going to guess the coverage drop (the Travis PR build shows many tests skipped) is because the `files` array is simply discarded in `bin/base.js`.

I'm not sure why some stuff is in `bin/base.js` and other stuff remains in `bin/_mocha`.

Would prefer `bin/base.js` had a more descriptive name and actually lived in `lib/` (as well as `bin/options.js`).

Bamieh:
@boneskull the `base.js` is responsible only for creating a mocha instance and configuring it. The `_mocha` file requires this file and runs the mocha instance, it also adds a custom behavior to what happens on exit and what to do in the watch mode.

I am +1 for moving these `.js` files to `lib/bin/`

craigtaub:
@Bamieh does this change still apply now we have the separation inside /cli!?

boneskull:
It's likely invalid, but I'd like to know if the refactors around yargs and such solve the problem about wrapping the Mocha CLI...

boneskull:
closed by #3556

PR #3423, opened 2018-06-21
PoC for process.allowedEnvironmentNodeFlags; DO NOT MERGE

boneskull:
This shows how Mocha can leverage [`process.allowedEnvironmentNodeFlags`](https://github.com/nodejs/node/pull/19335) to
support future Node.js flags.  It also adds support for arbitrary V8 flags
via the `--v8-` flag prefix.

See nodejs/node#19335 and cc @Chalker

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -11,6 +11,7 @@ const spawn = require('child_process').spawn;
 const path = require('path');
 const getOptions = require('./options');
 const args = [path.join(__dirname, '_mocha')];
+const envFlags = process.allowedEnvironmentNodeFlags;
 
 // Load mocha.opts into process.argv
 // Must be loaded here to handle node-specific options
@@ -19,12 +20,20 @@ getOptions();
 process.argv.slice(2).forEach(arg => {
   const flag = arg.split('=')[0];
 
+  // every flag UNSHIFTED onto the `args` array is
+  // an flag for the `node` executable; every flag
+  // PUSHED onto the `args` array is a Mocha flag.
+
   switch (flag) {
+    // non-standard debug flags.
+    // any debug-related flag needs Mocha's --no-timeouts
+    // or Mocha will timeout while the debugger is paused.
     case '-d':
+    case 'debug':
+      // rename flag
       args.unshift('--debug');
       args.push('--no-timeouts');
       break;
-    case 'debug':
     case '--debug':
     case '--debug-brk':
     case '--inspect':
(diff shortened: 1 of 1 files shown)


boneskull:
(CI will fail as the feature does not yet exist in Node.js)

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
     case '--debug':
     case '--debug-brk':
     case '--inspect':
     case '--inspect-brk':
       args.unshift(arg);
       args.push('--no-timeouts');
       break;
+    // non-standard flag
     case '-gc':
-    case '--expose-gc':
```
ChALkeR: Why is this removed? It's not exposed in https://github.com/nodejs/node/pull/19335

ChALkeR: Why is this removed?

boneskull: oversight

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
-    case '--gc-global':
-    case '--es_staging':
-    case '--no-deprecation':
-    case '--no-warnings':
-    case '--prof':
-    case '--log-timer-events':
-    case '--throw-deprecation':
-    case '--trace-deprecation':
-    case '--trace-warnings':
-    case '--use_strict':
```
ChALkeR: By the way, why is `--use-strict` removed? Is it exposed in https://github.com/nodejs/node/pull/19335?

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
       args.push('--no-timeouts');
       break;
+    // non-standard flag
     case '-gc':
-    case '--expose-gc':
+      // rename flag
       args.unshift('--expose-gc');
       break;
-    case '--gc-global':
-    case '--es_staging':
```
ChALkeR: Why is this removed?

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
-    case '--no-deprecation':
-    case '--no-warnings':
-    case '--prof':
```
ChALkeR: why is this removed?

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
-    case '--log-timer-events':
```
ChALkeR: Why is this removed?

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
       args.unshift(arg);
       args.push('--no-timeouts');
       break;
+    // non-standard flag
     case '-gc':
-    case '--expose-gc':
+      // rename flag
       args.unshift('--expose-gc');
       break;
-    case '--gc-global':
```
ChALkeR: Why is this removed?

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
-    case '--es_staging':
-    case '--no-deprecation':
-    case '--no-warnings':
-    case '--prof':
-    case '--log-timer-events':
-    case '--throw-deprecation':
-    case '--trace-deprecation':
-    case '--trace-warnings':
-    case '--use_strict':
-    case '--allow-natives-syntax':
```
ChALkeR: Why is this removed?

ChALkeR:
Same with `--perf_basic_prof` and `--perf-basic-prof` — note that with this change you previously allowed _only_ the second one and now seem to allow only the first one, am I correct?

bin/mocha
```diff
@@ -19,50 +20,52 @@ getOptions();
...
-    case '--perf-basic-prof':
```
ChALkeR: Am I correct that with this change you previously allowed only `--perf-basic-prof` and now allow only `--perf_basic_prof` form, am I correct?
This PR should implement the `_`/`-` logic, see https://github.com/nodejs/node/pull/19335/files.

ChALkeR:
Note that this also needs a fallback to support LTS Node.js versions like v6.x that are unlikely to receive the new API. Perhaps it would be worth to move this logic into a separate module that could be updated independently of Mocha and reused? With a fallback list.

bin/mocha
boneskull: this and the next one should be a special case for backwards compat.

bin/mocha
boneskull: these are now within `process.allowedNodeEnvironmentFlags`.

bin/mocha
boneskull: needs to be restored; will update

bin/mocha
ChALkeR: `--prof` is a valid flag (from `--v8-options`) and it doesn't look to be within `process.allowedNodeEnvironmentFlag`.

boneskull: Yep, missed this one.  They all look alike.  :smile:  Again, if I had a do-over, I wouldn't hardcode any of these.

boneskull:
Yes, it would make sense to provide this as a userland shim.

bin/mocha
```diff
@@ -19,50 +20,60 @@ getOptions();
...
-      } else if (arg.indexOf('--preserve-symlinks') === 0) {
+      if (
+        envFlags.indexOf(arg) > -1 ||
+        arg.indexOf('--harmony') === 0 ||
+        arg.indexOf('--preserve-symlinks') === 0 ||
+        arg === '--gc-global' ||
+        arg === '--es_staging' ||
+        arg === '--expose-gc' ||
+        arg === '--log-timer-events' ||
+        arg === '--use_strict' ||
```
ChALkeR: Does this support `--use-strict`?

boneskull: No; it never did.  `--v8-use-strict` is allowed, however.

bin/mocha
```diff
@@ -19,50 +20,60 @@ getOptions();
...
+        arg === '--allow-natives-syntax' ||
+        arg === '--perf-basic-prof'
```
ChALkeR: Why only `--perf-basic-prof` is hardcoded here? That's a part of `process.allowedEnvironmentNodeFlags` in `--perf_basic_prof` form. Why aren't `--perf-prof` and `--max-old-space-size` here, for example?

To consume `process.allowedEnvironmentNodeFlags` in the form in which that PR introduces it, you have to do the following:
1. Check if `arg` is present there.
2. Check if `arg.replace(/-/g, '_')` is present there.

Then this line won't be needed, it's duplicating what is already present in `process.allowedEnvironmentNodeFlags`.

boneskull: > Why only --perf-basic-prof is hardcoded here? 

Backwards compat.  This is how it currently exists in Mocha.

> Why aren't --perf-prof and --max-old-space-size

because they never were.  These hardcoded flags were added because someone wanted to use them.  As new flags appear in Node.js or V8, someone may want them, and the only way to get that done is to send a PR to Mocha with the hardcoded flags.  Mocha never had an exhaustive list of hardcoded flags.

PR #3428, opened 2018-06-26
Show diff in xunit

mlucool:
### Description of the Change
XUnit reporter should show diffs as output

XUnit reporter was already in core

### Benefits
You can tell actual vs. expected which is very helpful and the default in the default reporter.

### Possible Drawbacks
People wrote parsers.

### Applicable issues
This is an enhancement.

### DIFF
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -151,6 +151,8 @@ XUnit.prototype.write = function(line) {
  * @param {Test} test
  */
 XUnit.prototype.test = function(test) {
+  Base.useColors = false;
+
   var attrs = {
     classname: test.parent.fullTitle(),
     name: test.title,
@@ -159,6 +161,10 @@ XUnit.prototype.test = function(test) {
 
   if (test.state === 'failed') {
     var err = test.err;
+    var diff =
+      Base.hideDiff || !err.actual || !err.expected
+        ? ''
+        : '\n' + Base.generateDiff(err.actual, err.expected);
     this.write(
       tag(
         'testcase',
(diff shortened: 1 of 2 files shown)


mlucool:
@outsideris All requested changes have been fixed. Please let me know if anything else is required.

mlucool:
@outsideris any thoughts?

outsideris:
@mlucool I will review it this weekend.

mlucool:
@outsideris did you get a chance to look at this?

mlucool:
@boneskull can you take a look? It seems @outsideris is caught up IRL :)

plroebuck:
@mlucool, you never responded...

Does this still function as desired if `Base.inlineDiffs = true;` XML output-wise?

mlucool:
@plroebuck this will not work well for inlineDiffs as there is no color in XML. I verified that inline diffs don't work in anywhere without colors (via `xterm -cm`), so this maintains and extends status quo for that feature. What would you like to do here?

plroebuck:
@mlucool, your change brings this question up -- please add your opinion to #3537, but it's not so much directed at you...

plroebuck:
@mlucool, whether it "looks good" is beside the point; is it "functional" (i.e., can you discern differences without color) using inline diffs or not? If so, that's fine. If not... what's plan b?

Up until this change, `--color`/`--no-color` was moot.

mlucool:
@plroebuck I apologize, I should have been more clear above. Let me try again. If you open up a terminal which does not support colors (e.g. via `xterm -cm`) then run the default reporter, `spec`, you also cannot understand the diff correctly. I am happy to update this commit to throw or warn in this case, which is likely the best thing if treated in isolation, but I was suggesting above that we do something a little more globally than just for xunit.

plroebuck:
@mlucool my apologies for having missed this reply here. Given an incredibly short time frame, would you agree the PR is adequate for normal usage, and color can be resolved in bugfix release if required?

mlucool:
@plroebuck totally agree because the default is not the inline version

PR #3437, opened 2018-07-23
[Feature] Logging values from functions/values @ test runtime in code block

cjm771:
### Description of the Change

Idea is to simply be more explicit in code block about values when not hardcoded in, the code inside the `it(){ }` is not helpful as the test was dynamically created. So by storing executed functions / expressions / vars / etc inside the test doing `this.test.LOG_myInfo = getDynamicValue()`, the results get saved in test and will replace this line of code when outputted to code block.

**Example**
```js
it(message + ' find a ' + conflictType + ' conflict', function() { 
        
        //doesn't tell me much
        expect(conflictDetected).to.be.equal(conflictExpected);

        //log some useful information from dynamically created content
        this.test.LOG_funcRan = 'board.hasAny' + capitalize(conflictType) + 'Conflicts()\nExpected: ' + 
        conflictDetected + '\nGot: ' + conflictExpected;
        this.test.LOG_Board = boardVisualizer(board);
]});
```

**Resulting code block in html specrunner**
```js
//doesn't tell me much
expect(conflictDetected).to.be.equal(conflictExpected);

//log some useful information from dynamically created content

// LOG/funcRan
// ------------
// board.hasAnyRowConflicts()
// Expected: false
// Got: false

// LOG/Board
// ------------
// 0 0 0 0
// 0 1 0 0
// 0 0 1 0
// 0 0 0 0
```

### Alternate Designs

Ideally there would be a `this.log()`  function or something. Also perhaps a log window dropdown can be used instead of including inside code block.

Super handy for dynamically created tests

### Benefits

Testmakers can write dynamic tests, but still expose variable values / useful information in their code blocks. 

### Possible Drawbacks

Exposing user to internal `this.test` might open up other issues.

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -224,24 +224,24 @@ function HTML(runner) {
     updateStats();
   });
 
-    /*  `LOG` feature: exposing stored values
-    * 
-    * how can i expose a variables VALUE / executed code in my it() test? 
-    * within the function, use this.test.LOG
-    * 
-    * anything attatched to this.test.LOG_myVar will be output as it's value to the code preview
-    * 
-    * ex: this.test.LOG_myObj = JSON.stringify(myObj)
-    * 
-    * code text will be replace with a comment
-    * 
-    * // LOG/myObj
-    * // ----------
-    * // {a: b, etc}
-    * 
-    *  @test {Test} test object
-    */
-    function logify(test) {
+  /*  `LOG` feature: exposing stored values
+  * 
+  * how can i expose a variables VALUE / executed code in my it() test? 
+  * within the function, use this.test.LOG
+  * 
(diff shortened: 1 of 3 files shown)


cjm771:
a seperate logger function would require more customizing of the default specrunner.html template. this would allow users to log pertinent data without this. 

-Use case could be "check that every possibility of a tic tac toe game determines the correct winner" . A useful thing to log (that would be automated hopefully) is the 3 x 3 game board for each possibility.

- Another might be make sure a random 50, 500-character hashes share not 5 characters in sequence (or something very annoying to harde code in)..printing these hashes would be nice to see maybe.

plroebuck:
Agree with @outsideris. This doesn't really belong in test logic.
Propose rejection of this PR without _strong_ objection from core maintainer.

PR #3438, opened 2018-07-23
Whitelist `--experimental-modules` Node flag

demurgos:
### Requirements

### Description of the Change

This commit adds `--experimental-modules` to the list of Node flag recognized by Mocha. This flag is passed directly to Node and enables experimental native support for ESM modules.

### Alternate Designs

- **Do nothing**: We could not add this option. This would require people to keep using `NODE_OPTIONS` for `--experimental-modules`.

 This PR allows to use `mocha --experimental-modules foo.spec.js` instead of requiring the workaround `NODE_OPTIONS="--experimental-modules" mocha foo.spec.js`.

This addition is in line with the various `--harmony` flags that were added while some ES6 features were still experimental.

### Benefits

Allowing the use of the flag directly improves the ability to compose the command invocation. The flag has only a local impact. Requiring to set an environment variable increases the complexity of the applications, especially if environment variables are not needed otherwise.

It also improves the ergonomics of invoking Mocha when the user wants to use experimental modules.

### Possible Drawbacks

No side-effects or negative impact (perhaps maintenance, but there are already many other such flags).

### Applicable issues

See <https://github.com/mochajs/mocha/issues/3006#issuecomment-406897900>: merging this PR would improve the use of Mocha for pure Node+Mocha tests for native ESM.

This is an enhancement (semver minor version). It can be a patch since it's very minor and does not affect prod.

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -38,6 +38,7 @@ process.argv.slice(2).forEach(arg => {
       break;
     case '--gc-global':
     case '--es_staging':
+    case '--experimental-modules':
     case '--no-deprecation':
     case '--no-warnings':
     case '--prof':


demurgos:
A typical case when using only CLI args without env variables is better is when using coverage tools such as nyc or c8. If you only have CLI args, you just have to concatenate the mocha command after the coverage command.

I'd like to experiment a bit with coverage. I already wrote some code to correctly invoke all the tools involved so there's no rush for me but having this PR merged and published would allow to clean up some logic.

I'm not really aware of the release process of Mocha: does it need more approvals or the action of specific members for this change to be merged/published?

PR #3440, opened 2018-07-26
prevent to set bdd ui always

### LINKED ISSUE #3421: BDD and TDD files can be mixed when using "--ui tdd" but not when using "--ui bdd"
As far as I understand you are not supposed to mix testing style i.e. have some files using BDD and some using TDD.
I mean, then what would the point be to specify the style with the ui flag if it would work automatically anyway?
However, if you do use "mocha --ui tdd" then indeed it seems to work with bdd files too.
But when you try with the opposite ("mocha --ui bdd") it does not work with tdd files, which does not make sense to me.

Example with two files (using mocha 5.2.0):

test/test_bdd.js
```
  describe('#indexOf()', function() {
    it('should return -1 when the value is not present', function() {
		assert.equal([1,2,3].indexOf(4), -1);
    });
  });
});
```

test/test_tdd.js
```
suite('Array', function() {
  suite('#indexOf()', function() {
    test('should return -1 when not present', function() {
      assert.equal(-1, [1,2,3].indexOf(4));
    });
  });
});
```

mocha --ui tdd
Result: 
2 passing tests

mocha --ui bdd
Result:
ReferenceError: suite is not defined

outsideris:
### Description of the Change
`bdd` ui is set anyway. So, `bdd` ui called twice if you specified `--ui bdd`. `bdd` and `tdd` uis called if you specified `--ui tdd`. It is because of [new Mocha()](https://github.com/mochajs/mocha/blob/master/bin/_mocha#L22) and [setting ui](https://github.com/mochajs/mocha/blob/master/bin/_mocha#L501).

Thererfore, I fixed to call ui only when `option.ui` specified.(for using `new Mocha()` directly)

Additionally, I named UI function to avoid anonymous function name. BDD and qUnit already has [named function](https://github.com/mochajs/mocha/blob/master/lib/interfaces/bdd.js#L22).

### Alternate Designs
N/A

As #3421, it is a bug.

### Benefits
Users can specify UI as their intention.

### Possible Drawbacks
If users uses mixed UIs (e.g bdd and tdd), it will be broken.

### Applicable issues
Fix #3421

### DIFF
--- a/lib/interfaces/exports.js
+++ b/lib/interfaces/exports.js
@@ -19,8 +19,98 @@ var Test = require('../test');
  *
  * @param {Suite} suite Root suite.
  */
-module.exports = function(suite) {
+module.exports = function exportsInterface(suite) {
   var suites = [suite];
+
+  suite.on('pre-require', function(context, file, mocha) {
+    var common = require('./common')(suites, context, mocha);
+
+    context.before = common.before;
+    context.after = common.after;
+    context.beforeEach = common.beforeEach;
+    context.afterEach = common.afterEach;
+    context.run = mocha.options.delay && common.runWithSuite(suite);
+    /**
+     * Describe a "suite" with the given `title`
+     * and callback `fn` containing nested suites
+     * and/or tests.
+     */
+
+    context.describe = context.context = function(title, fn) {
+      return common.suite.create({
+        title: title,
+        file: file,
+        fn: fn
+      });
+    };
+
+    /**
+     * Pending describe.
+     */
+
+    context.xdescribe = context.xcontext = context.describe.skip = function(
+      title,
+      fn
+    ) {
+      return common.suite.skip({
+        title: title,
+        file: file,
+        fn: fn
+      });
+    };
+
+    /**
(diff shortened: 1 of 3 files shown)


lib/interfaces/exports.js
```diff
@@ -19,9 +19,99 @@ var Test = require('../test');
  *
  * @param {Suite} suite Root suite.
  */
-module.exports = function(suite) {
+module.exports = function exportsInterface(suite) {
   var suites = [suite];
 
+  suite.on('pre-require', function(context, file, mocha) {
```
outsideris: I guess `exports` UI assumed `bdd` style already set. So, I copied and pasted from `bdd` UI.

juergba: I don't understand the changes you applied to lib/interfaces/exports.js. This file is the interface for `ui: exports` and has nothing to do with the fact that mocha.ui is called twice.
The changes to mocha.js are working fine, even when --ui is set twice (mocha.opts and command line).

boneskull:
this was resolved in b5c0fb0a

PR #3441, opened 2018-07-26
Prevent to empty default value in CLI options

### LINKED ISSUE #3433: Empty default value in cli options
```
--globals <names>                       allow the given comma-delimited global [names] (default: )
```

If default value is an empty list(`[]`), it shows as `(default: )`. It is confusing to users. No default or `(default: [])` is better.

outsideris:
### Description of the Change
If empty array is default value([]), it displayed as `default: `

With `commander` and a default value is an empty array(`[]`), it shows like this:

```
--globals <names>                       allow the given comma-delimited global [names] (default: )
```

`(default: )` is confusing to users.

So, this PR doesn't use a default value with an empty array. Instead, it set an empty array manually.

### Alternate Designs
Let it as `(default: )` or fix it in [commander.js](https://github.com/tj/commander.js).

### Possible Drawbacks
I'm not sure.

### Applicable issues
Fix #3433

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -763,9 +763,9 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
     -w, --watch                             watch files for changes
     --check-leaks                           check for global variable leaks
     --full-trace                            display the full stack trace
-    --compilers <ext>:<module>,...          use the given module(s) to compile files (default: )
+    --compilers <ext>:<module>,...          use the given module(s) to compile files
     --debug-brk                             enable node's debugger breaking on the first line
-    --globals <names>                       allow the given comma-delimited global [names] (default: )
+    --globals <names>                       allow the given comma-delimited global [names]
     --es_staging                            enable all staged features
     --harmony<_classes,_generators,...>     all node --harmony* flags are available
     --preserve-symlinks                     Instructs the module loader to preserve symbolic links when resolving and caching modules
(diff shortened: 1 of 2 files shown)


plroebuck:
Certainly understand having fix on our side. But why was there no corresponding issue created for "tj/commander"? Would prefer to leave maintenance of their bugs to them.

outsideris:
@plroebuck As I said that in Alternate Designs, I'm not sure it's a bug in commander.js because it's related with how to use default vault and we can fix it on our side easily. 
If we agree it is not our responsibility, I will report an issue or open the PR on commander.js.

plroebuck:
Think it's their problem - opened tj/commander#855

docs/index.md
```diff
@@ -763,9 +763,9 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
     -w, --watch                             watch files for changes
     --check-leaks                           check for global variable leaks
     --full-trace                            display the full stack trace
-    --compilers <ext>:<module>,...          use the given module(s) to compile files (default: )
+    --compilers <ext>:<module>,...          use the given module(s) to compile files
```
plroebuck: So for all of these README changes, why did you remove the default text altogether rather than just adding '[]' where it was missing?

plroebuck:
Tangentially related only.
Why are [these default values]([here](https://github.com/mochajs/mocha/blob/e130e1471aec93c9f320e37bd4269dc4dea97dd9/bin/_mocha#L166-L167)) hardcoded as strings?

outsideris:
I see. I saw [the opened PR](https://github.com/tj/commander.js/pull/856) in commander.js. 
We can wait for a new release of commander.js.

plroebuck:
Progress!

PR tj/commander.js#856

outsideris:
@plroebuck As I leave comment in #3433 , we can open new PR after commander.js fixed it.
We can handle it in this PR. However, many opened PRs is not good to manage and new changes will be very different with this.

Issue #1257, opened 2014-07-03
Tap reporter can produce invalid output due to stack trace


jreichenberg:
The tap reporter does the following on fail:

`console.log(err.stack.replace(/^/gm, '  '));`

Depending on the stack trace, this can produce invalid tap output.  For example, a tap file with a stack trace from WebDriverJs is below...invalid (try it in http://instanttap.appspot.com/ ).

Ideas for sanitizing the stack trace?
Or providing an option to leave it on err.stack and not print it (so it can be potentially written as a file or written at the end)?

```
1..5
Selenium Server started on port 4444
Local http server listening at http://localhost:8080
Webdriver navigating to http://localhost:8080/test/ui/fixtures/index.html
ok 1 Instancing Toggle Switch from provided DOM data attribute should automatically instance a toggle button from the expected DOM
ok 2 Instancing Toggle Switch from dynamically-added DOM should not automatically instance a toggle button without the expected DOM
not ok 3 Instancing Toggle Switch from dynamically-added DOM should be instanced via JavaScript
  AssertionError: expected true to be false
      at <anonymous>
  ==== async task ====
  WebDriver.findElements(By.xpath("//input[last()]/ancestor::div[contains(@class, 'has-switch')]"))
      at webdriver.WebDriver.schedule (/Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/webdriver/webdriver.js:267:15)
      at webdriver.WebDriver.findElements (/Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/webdriver/webdriver.js:825:17)
      at webdriver.WebDriver.isElementPresent (/Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/webdriver/webdriver.js:787:29)
      at /Users/212362413/projects/dx-components/toggle-switch/test/ui/spec/toggle-switch-ui-spec.js:74:20
      at /Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/goog/base.js:1178:15
      at webdriver.promise.ControlFlow.runInNewFrame_ (/Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/webdriver/promise.js:1445:20)
      at notify (/Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/webdriver/promise.js:328:12)
      at notifyAll (/Users/212362413/npm/lib/node_modules/selenium-webdriver/lib/webdriver/promise.js:297:7)
[...]

Issue #492, opened 2012-07-03
Turn reporters into seperate module


Raynos:
The mocha reporters are sweet. However they are hard coupled to the mocha interface.

It would be elegant if they were a seperate module.

Consider mocha just outputting TAP format for the tests to stdout

And then doing

`mocha test | mocha-reporters --dot`

The mocha-reporters module could just be a thing that takes TAP input and outputs to the reporter to stdout.

You could also consider using it programatically as

`mocha.output.pipe(mochaReporter.spec).pipe(process.stdout)`

This allows people to re-use the reporters in any test framework that outputs TAP.

If you think this is a sensible idea we can draft out how the refactoring would go and I can try migrating the reporters into seperate modules. I've pretty much ported the HTML reporter to something that consumes TAP already so that feels like duplicated effort.

PR #3452, opened 2018-08-08
resolves #2149, #1257: Make TAP reporter conform to TAP13

### LINKED ISSUE #2149: Update TAP reporter to TAP 13
I noticed that mocha is using the older TAP specification and wondering if you would be open to updating it to [TAP 13](https://testanything.org/tap-version-13-specification.html)? It is pretty similar. I think the main differences are that version 13 starts with a line that says `TAP version 13`, and encodes test output in YAML.

General format:

``` tap
TAP version 13
1..N
ok 1 Description # Directive
# Diagnostic
  ---
  message: 'Failure message'
  severity: fail
  data:
    got:
      - 1
      - 3
      - 2
    expect:
      - 1
      - 2
      - 3
  ...
ok 47 Description
ok 48 Description
more tests....
```

mollstam:
### Description of the Change

The TAP reporter has been made to conform with [TAP13](https://testanything.org/tap-version-13-specification.html).

Changes to the actual reporter are small: a version string is emitted as the first line, and error `message` and `stack` are both emitted inside a YAML block to be parsed by any TAP Consumer.

A quite large integration test has been added that encodes the expectations of the TAP13 spec, and the existing TAP reporter unit tests have been updated to be compatible with the new spec.

Finally 'reporters.fixture.js' was added to have something to run that generates test results with varying properties.

### Alternate Designs

It was considered to add either a new reporter (e.g. `tap13`) or hide the new spec under a reporter option but since there are no breaking changes in TAP13 this was deemed unnecessary.

Mocha relies on a default selection of usable reporters. A bunch of TAP Consumers that I've surveyed are poorly written and _expect_ TAP13 or will otherwise break thus rendering the current TAP reporter in mocha less useful.

### Benefits

A larger ecosystem of tools that are only compatible with TAP13 are made available to use together with mocha. Taking a step back, benefits in the TAP13 spec vs v12 are inherited, such as being able to encode richer error information in the now allowed YAML block in the TAP report is a Good Thing.

### Possible Drawbacks

Many TAP Consumers I've encountered are not as flexible as one might wish so it is possible they don't fully conform to the old v12 spec, thus not being compatible with the TAP13 spec resulting in undefined behaviour. For example, currently in mocha the error stack is simply inserted into the report following the failed test, something that is allowed by the v12 spec as an "unknown" line that shouldn't be an error and may be handled by the test harness in any way. If a test harness handles mocha's stacks by e.g. using regex it will probably break now that the stack is wrapped in what TAP13 calls a "YAML block".

### Applicable issues

https://github.com/mochajs/mocha/issues/2149
https://github.com/mochajs/mocha/issues/1257

#### Note

I've decided to not address issue https://github.com/mochajs/mocha/issues/2410 with this PR as it involves communicating with third party tools (e.g. tap-spec) over TAP13 in a new way which I feel is outside the scope of simply bringing the TAP reporter up to spec, whereas 2410 is more about establishing a new spec regarding the structure of the YAML blocks (unspecified by TAP13).

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -53,7 +53,7 @@ function TAP(runner, options) {
   });
 
   runner.on('pending', function(test) {
-    console.log('ok %d %s # SKIP -', n, title(test));
+    self.producer.writePending(n, test);
   });
 
   runner.on('pass', function(test) {
@@ -104,6 +104,10 @@ TAP12Producer.prototype.writeFail = function(n, test, err) {
   }
 };
 
+TAP12Producer.prototype.writePending = function(n, test) {
+  console.log('ok %d %s # SKIP -', n, title(test));
+};
+
 function TAP13Producer() {}
 
 TAP13Producer.prototype.yamlIndent = function(level) {
@@ -131,4 +135,8 @@ TAP13Producer.prototype.writeFail = function(n, test, err) {
     }
     console.log(this.yamlIndent(1) + '...');
   }
+};
+
+TAP13Producer.prototype.writePending = function(n, test) {
+  console.log('ok %d %s # SKIP -', n, title(test));
 };
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -57,14 +57,12 @@ function TAP(runner) {
     if (emitYamlBlock) {
       console.log('  ---');
       if (err.message) {
-        console.log('    message: ' + err.message);
+        console.log('    message: |-');
+        console.log(err.message.replace(/^/gm, '      '));
       }
-      // stack needs to be last in yaml due to literal style indicator (|)
       if (err.stack) {
         console.log('    stack: |-');
-        if (err.stack) {
-          console.log(err.stack.replace(/^/gm, '      '));
-        }
+        console.log(err.stack.replace(/^/gm, '      '));
       }
       console.log('  ...');
     }
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -32,6 +32,7 @@ function TAP(runner) {
   var failures = 0;
 
   runner.on('start', function() {
+    console.log('TAP version 13');
     var total = runner.grepTotal(runner.suite);
     console.log('%d..%d', 1, total);
   });
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -34,7 +34,7 @@ function TAP(runner, options) {
   var tapSpec = '12';
   if (options && options.reporterOptions) {
     if (options.reporterOptions.spec) {
-      var tapSpec = options.reporterOptions.spec;
+      tapSpec = options.reporterOptions.spec;
     }
   }
 
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -31,10 +31,6 @@ function TAP(runner, options) {
   var self = this;
   var n = 1;
 
-  function yamlIndent(level) {
-    return Array(level + 1).join('  ');
-  }
-
   var tapSpec = '12';
   if (options && options.reporterOptions) {
     if (options.reporterOptions.spec) {
@@ -65,20 +61,7 @@ function TAP(runner, options) {
   });
 
   runner.on('fail', function(test, err) {
-    console.log('not ok %d %s', n, title(test));
-    var emitYamlBlock = err.message != null || err.stack != null;
-    if (emitYamlBlock) {
-      console.log(yamlIndent(1) + '---');
-      if (err.message) {
-        console.log(yamlIndent(2) + 'message: |-');
-        console.log(err.message.replace(/^/gm, yamlIndent(3)));
-      }
-      if (err.stack) {
-        console.log(yamlIndent(2) + 'stack: |-');
-        console.log(err.stack.replace(/^/gm, yamlIndent(3)));
-      }
-      console.log(yamlIndent(1) + '...');
-    }
+    self.producer.writeFail(n, test, err);
   });
 
   runner.once('end', function() {
@@ -111,10 +94,41 @@ TAP12Producer.prototype.writeStart = function(runner) {
   console.log('%d..%d', 1, total);
 };
 
+TAP12Producer.prototype.writeFail = function(n, test, err) {
+  console.log('not ok %d %s', n, title(test));
+  if (err.message) {
+    console.log(err.message.replace(/^/gm, '  '));
+  }
+  if (err.stack) {
+    console.log(err.stack.replace(/^/gm, '  '));
+  }
+};
+
 function TAP13Producer() {}
+
+TAP13Producer.prototype.yamlIndent = function(level) {
+  return Array(level + 1).join('  ');
+};
 
 TAP13Producer.prototype.writeStart = function(runner) {
   console.log('TAP version 13');
   var total = runner.grepTotal(runner.suite);
   console.log('%d..%d', 1, total);
+};
+
+TAP13Producer.prototype.writeFail = function(n, test, err) {
+  console.log('not ok %d %s', n, title(test));
+  var emitYamlBlock = err.message != null || err.stack != null;
+  if (emitYamlBlock) {
+    console.log(this.yamlIndent(1) + '---');
+    if (err.message) {
+      console.log(this.yamlIndent(2) + 'message: |-');
+      console.log(err.message.replace(/^/gm, this.yamlIndent(3)));
+    }
+    if (err.stack) {
+      console.log(this.yamlIndent(2) + 'stack: |-');
+      console.log(err.stack.replace(/^/gm, this.yamlIndent(3)));
+    }
+    console.log(this.yamlIndent(1) + '...');
+  }
 };
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -31,6 +31,10 @@ function TAP(runner) {
   var passes = 0;
   var failures = 0;
 
+  function yamlIndent(level) {
+    return Array(level + 1).join('  ');
+  }
+
   runner.on('start', function() {
     console.log('TAP version 13');
     var total = runner.grepTotal(runner.suite);
@@ -55,16 +59,16 @@ function TAP(runner) {
     console.log('not ok %d %s', n, title(test));
     var emitYamlBlock = err.message != null || err.stack != null;
     if (emitYamlBlock) {
-      console.log('  ---');
+      console.log(yamlIndent(1), '---');
       if (err.message) {
-        console.log('    message: |-');
-        console.log(err.message.replace(/^/gm, '      '));
+        console.log(yamlIndent(2), 'message: |-');
+        console.log(err.message.replace(/^/gm, yamlIndent(3)));
       }
       if (err.stack) {
-        console.log('    stack: |-');
-        console.log(err.stack.replace(/^/gm, '      '));
+        console.log(yamlIndent(2), 'stack: |-');
+        console.log(err.stack.replace(/^/gm, yamlIndent(3)));
       }
-      console.log('  ...');
+      console.log(yamlIndent(1), '...');
     }
   });
 
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -59,16 +59,16 @@ function TAP(runner) {
     console.log('not ok %d %s', n, title(test));
     var emitYamlBlock = err.message != null || err.stack != null;
     if (emitYamlBlock) {
-      console.log(yamlIndent(1), '---');
+      console.log(yamlIndent(1) + '---');
       if (err.message) {
-        console.log(yamlIndent(2), 'message: |-');
+        console.log(yamlIndent(2) + 'message: |-');
         console.log(err.message.replace(/^/gm, yamlIndent(3)));
       }
       if (err.stack) {
-        console.log(yamlIndent(2), 'stack: |-');
+        console.log(yamlIndent(2) + 'stack: |-');
         console.log(err.stack.replace(/^/gm, yamlIndent(3)));
       }
-      console.log(yamlIndent(1), '...');
+      console.log(yamlIndent(1) + '...');
     }
   });
 
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -7,7 +7,6 @@
  */
 
 var Base = require('./base');
-var inherits = require('../utils').inherits;
 
 /**
  * Expose `TAP`.
@@ -28,7 +27,6 @@ exports = module.exports = TAP;
 function TAP(runner) {
   Base.call(this, runner);
 
-  var self = this;
   var n = 1;
   var passes = 0;
   var failures = 0;
@@ -76,11 +74,6 @@ function TAP(runner) {
     console.log('# fail ' + failures);
   });
 }
-
-/**
- * Inherit from `Base.prototype`.
- */
-inherits(TAP, Base);
 
 /**
  * Return a TAP-safe title of `test`
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -7,6 +7,7 @@
  */
 
 var Base = require('./base');
+var inherits = require('../utils').inherits;
 
 /**
  * Expose `TAP`.
@@ -27,6 +28,7 @@ exports = module.exports = TAP;
 function TAP(runner) {
   Base.call(this, runner);
 
+  var self = this;
   var n = 1;
   var passes = 0;
   var failures = 0;
@@ -74,6 +76,11 @@ function TAP(runner) {
     console.log('# fail ' + failures);
   });
 }
+
+/**
+ * Inherit from `Base.prototype`.
+ */
+inherits(TAP, Base);
 
 /**
  * Return a TAP-safe title of `test`
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -53,11 +53,20 @@ function TAP(runner) {
   runner.on('fail', function(test, err) {
     failures++;
     console.log('not ok %d %s', n, title(test));
-    if (err.message) {
-      console.log(err.message.replace(/^/gm, '  '));
-    }
-    if (err.stack) {
-      console.log(err.stack.replace(/^/gm, '  '));
+    var emitYamlBlock = err.message != null || err.stack != null;
+    if (emitYamlBlock) {
+      console.log('  ---');
+      if (err.message) {
+        console.log('    message: ' + err.message);
+      }
+      // stack needs to be last in yaml due to literal style indicator (|)
+      if (err.stack) {
+        console.log('    stack: |-');
+        if (err.stack) {
+          console.log(err.stack.replace(/^/gm, '      '));
+        }
+      }
+      console.log('  ...');
     }
   });
 
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -7,6 +7,7 @@
  */
 
 var Base = require('./base');
+var inherits = require('../utils').inherits;
 
 /**
  * Expose `TAP`.
@@ -89,3 +90,8 @@ function TAP(runner) {
 function title(test) {
   return test.fullTitle().replace(/#/g, '');
 }
+
+/**
+ * Inherit from `Base.prototype`.
+ */
+inherits(TAP, Base);
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -25,19 +25,31 @@ exports = module.exports = TAP;
  * @api public
  * @param {Runner} runner
  */
-function TAP(runner) {
+function TAP(runner, options) {
   Base.call(this, runner);
 
+  var self = this;
   var n = 1;
 
   function yamlIndent(level) {
     return Array(level + 1).join('  ');
   }
 
+  var tapSpec = '12';
+  if (options && options.reporterOptions) {
+    if (options.reporterOptions.spec) {
+      var tapSpec = options.reporterOptions.spec;
+    }
+  }
+
+  if (tapSpec === '13') {
+    this.producer = new TAP13Producer();
+  } else {
+    this.producer = new TAP12Producer();
+  }
+
   runner.on('start', function() {
-    console.log('TAP version 13');
-    var total = runner.grepTotal(runner.suite);
-    console.log('%d..%d', 1, total);
+    self.producer.writeStart(runner);
   });
 
   runner.on('test end', function() {
@@ -91,3 +103,18 @@ function title(test) {
  * Inherit from `Base.prototype`.
  */
 inherits(TAP, Base);
+
+function TAP12Producer() {}
+
+TAP12Producer.prototype.writeStart = function(runner) {
+  var total = runner.grepTotal(runner.suite);
+  console.log('%d..%d', 1, total);
+};
+
+function TAP13Producer() {}
+
+TAP13Producer.prototype.writeStart = function(runner) {
+  console.log('TAP version 13');
+  var total = runner.grepTotal(runner.suite);
+  console.log('%d..%d', 1, total);
+};
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -65,9 +65,7 @@ function TAP(runner, options) {
   });
 
   runner.once('end', function() {
-    console.log('# tests ' + (runner.stats.passes + runner.stats.failures));
-    console.log('# pass ' + runner.stats.passes);
-    console.log('# fail ' + runner.stats.failures);
+    self.producer.writeEnd(runner);
   });
 }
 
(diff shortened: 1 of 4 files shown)


mollstam:
Hi y'all, thought I'd get the ball rolling on having mocha's tap reporter output v13 spec compliant reports since I need it. :grimacing: 

Any thoughts?

Technically v13 is not backwards compatible with v12 so this is a breaking change for any TAP Consumers that don't also support v13 but I'd defer to a maintainer to deem this being a patch or minor update.

(Kind of wished bots ignored WIP PRs)

Bamieh:
@mollstam thanks for taking your time to work on this here. Just a thought, we can provide the new version under a new flag, until we eventually deprecate the old one.

I looked at the [v13 spec](https://testanything.org/tap-version-13-specification.html) But it does not list/mention the breaking changes, can you provide the breaking changes to give us a better idea about what we're dealing with? Also if this gets pushed, we'll include those breaking changes in the release notes.

mollstam:
All right, I've at this point created a couple of integration tests to ensure that the output conforms to spec v13 and also some other Nice Things regarding the structure of the YAML block containing error message and stack.

The YAML test in itself is a bit hairy but I feel like it's good enough for now. I'd like to create a new fixture that generates a report with "a little bit of everything" so I can make sure that it looks good, and so that I can also stop using uncaught.fixture.js.

I'd like to also keep looking a bit at putting more data about errors into the YAML block for some nice reporting by [tap-diff](https://github.com/axross/tap-diff), [tap-spec](https://github.com/scottcorgan/tap-spec) and other tape/node-tap related consumers. The TAP spec gives implementors free reign of content in YAML block would be nice to inventory what most consumers expect and see if we can provide it.

mollstam:
All right, I feel like I'm done here and I'd like to open up for review.

Also the semver-major label is no longer required.

plroebuck:
Since there's no _huge_ major change between 12, 13, & 14, any consideration given to providing the TAP version via `reporterOptions` - making the TAP producer (reporter) both forward and backward compatible?

lib/reporters/tap.js
```diff
@@ -52,11 +53,18 @@ function TAP(runner) {
...
     failures++;
     console.log('not ok %d %s', n, title(test));
-    if (err.message) {
-      console.log(err.message.replace(/^/gm, '  '));
-    }
-    if (err.stack) {
-      console.log(err.stack.replace(/^/gm, '  '));
+    var emitYamlBlock = err.message != null || err.stack != null;
+    if (emitYamlBlock) {
+      console.log('  ---');
```
plroebuck: Since TAP14 is standardizing on 2 space indention, using something like [this](https://github.com/mochajs/mocha/blob/d401ee198bd5026bc61641ad26eab31e6746713f/lib/reporters/spec.js#L36-L38) might add to maintainability.

lib/reporters/tap.js
```diff
@@ -32,6 +32,7 @@ function TAP(runner) {
   var failures = 0;
 
   runner.on('start', function() {
+    console.log('TAP version 13');
```
plroebuck: Consider using `self.write()` and `self.writeln()` so that file-baseed output could be a possibility.

test/integration/fixtures/reporters.fixture.js
```diff
@@ -0,0 +1,63 @@
...
+        });
+      });
+    });
+    describe('Birds', function() {
+      it('should have feathers', function(done) { done(); });
+      it('should lay hard-shelled eggs', function(done) { done(); });
+    });
+  });
+
+  describe('Tardigrades', function() {
```
plroebuck: Want to approve this simply because you figured out how to reference "water bears" in a unit test... :)

test/integration/reporters.spec.js
```diff
@@ -99,4 +99,135 @@ describe('reporters', function() {
       });
     });
   });
+
+  describe('tap', function() {
+    var not = function(predicate) {
```
plroebuck: Loved your use of predicates!

plroebuck:
As a (hopefully non-digressive) follow-on question, what TAP consumers were used to verify output from this PR? While I don't think these should be added to Mocha's build toolchain, any related test scripts you used for such can be preserved here as comment attachments.

lib/reporters/tap.js
```diff
@@ -52,11 +53,18 @@ function TAP(runner) {
   runner.on('fail', function(test, err) {
     failures++;
     console.log('not ok %d %s', n, title(test));
-    if (err.message) {
-      console.log(err.message.replace(/^/gm, '  '));
-    }
-    if (err.stack) {
-      console.log(err.stack.replace(/^/gm, '  '));
+    var emitYamlBlock = err.message != null || err.stack != null;
```
plroebuck: I see "diff"-able `Error`s in Mocha codebase with `actual` and `expected` properties. Would they be considered "interesting" as output?

plroebuck:
On `suite end` event, should Mocha emit a timing comment line?

A la this unbuffered subtest reference output [here](https://www.node-tap.org/subtests/).
```tap
TAP version 13
# Subtest: foo
    # Subtest: bar
        1..1
        ok 1 - this is fine
    ok 1 - bar # time=1.831ms

    1..1
ok 1 - foo # time=11.106ms

1..1
# time=19.039ms

```

mollstam:
Hello friend, thanks for having a look! :)

> Since there's no _huge_ major difference between [TAP12 and YAP13/14](https://github.com/TestAnything/testanything.github.io/issues/126) output, any consideration given to providing the TAP version via `reporterOptions` - making the TAP producer (reporter) both forward and backward compatible?
> 
> Add `TAPProducer(version)` class instance to `TAP` reporter, and defer all writing of TAP output to it.
> 
> As TAP12 is apparently ancient nowadays, perhaps TAP13 should be unspecified default. Or leave the default "as is" and allow this to be added to a "semver-minor" Mocha release.

As the TAP13 spec is both forward and backward compatible I'm not sure keeping the TAP12 output around is necessary, what's your thoughts?

> Have these changes been run against the [TAP 14 draft](https://github.com/TestAnything/testanything.github.io/pull/36)?

I had a look at the TAP14 draft back when implementing this and decided not to implement it since it is only a draft and I wanted to keep the footprint of the PR as small as possible.

> Without personally really diving into TAP, does this reporter's output match that which would be produced by Node-based TAP producers `node-tap` and `tape`?

I did recreate the fixture in `tape`, mostly to test how they handle subtests/suites (as discussed in issue #2410). [Here is a gist of the tape version](https://gist.github.com/mollstam/44f7d3a314e0a4b22765f9ce445cafb1) and [here is a diff from tape report to this version of the mocha tap reporter](https://gist.github.com/mollstam/3cb828e93cb6a1ee3dd7d4e983794d16/revisions#diff-229da9bbc37f8a37f858791aa590ad53). The main difference is the tape report having more structured data (`operator`, `expected`, `actual`) in the YAML rather than simply "baking it" into the `message` property. I investigated reaching similar formatting but the footprint of the fix grew and I wanted to keep things simple.

As you can see the running theme is me wanting to make as little of change as possible to reach what I set out to do. This iteration of the TAP reporter is not by any means a "final form" but a lot of the future work relies on the YAML format being standardized (consumers vary _wildly_) and for that I think TAP14 needs to get ratified, or more effort to accomodate the popular consumers out there.

Let me know your thoughts and thanks again.

lib/reporters/tap.js
mollstam: xunit reporter has a `write()` and I find nothing about `writeln()`. Can you point me in the right direction? `console.log` is what most reporters seem to be doing. :scream:

lib/reporters/tap.js
mollstam: I think this is a great feature to have but it doesn't really have much to do with TAP13, rather a "while we're at it" thing and if I remember correctly I ran into some problems investigating it, but this was I while back. I think this feature shouldn't ride on this PR tbh.

mollstam:
> On `suite end` event, should Mocha emit a timing comment line?
> 
> A la this [unbuffered subtest reference output](https://www.node-tap.org/subtests/).
> 
> ```
> TAP version 13
> # Subtest: foo
>     # Subtest: bar
>         1..1
>         ok 1 - this is fine
>     ok 1 - bar # time=1.831ms
> 
>     1..1
> ok 1 - foo # time=11.106ms
> 
> 1..1
> # time=19.039ms
> ```

IDK, its a very small fix so it might be useful? I can't find any consumer that use it though…

lib/reporters/tap.js
```diff
@@ -7,7 +7,6 @@
  */
 
 var Base = require('./base');
```
plroebuck: Disagree with this change. IMO, **all** reporters should extend `Base`.

lib/reporters/tap.js
```diff
@@ -28,7 +27,6 @@ exports = module.exports = TAP;
 function TAP(runner) {
   Base.call(this, runner);
 
-  var self = this;
   var n = 1;
   var passes = 0;
   var failures = 0;
```
plroebuck: L30-L32 seem unnecessary. How is this different from `runner.stats` (from `Base`)?

lib/reporters/tap.js
plroebuck: Sorry, using functions from another proposed PR.
Until that's ready, you could implement those as:
```js
/**
 * Write out the given string.
 *
 * @param {string} s
 */
TAP.prototype.write = function(s) {
  if (this.fileStream) {
    this.fileStream.write(s);
  } else {
    process.stdout.write(s);
  }
};

/**
 * Write out the given string and terminate the line.
 *
 * @param {string} s
 */
TAP.prototype.writeln = function(s) {
  this.write(s + '\n');
};
```

plroebuck: Glancing your `console.log` usage, will also need to add an `sprintf()` function to handle your string substitutions.

lib/reporters/tap.js
plroebuck: But wasn't that _exactly_ part of the `tape` output?

plroebuck:
> > Since there's no _huge_ major difference between [TAP12 and YAP13/14](https://github.com/TestAnything/testanything.github.io/issues/126) output, any consideration given to providing the TAP version via `reporterOptions` - making the TAP producer (reporter) both forward and backward compatible?
> > Add `TAPProducer(version)` class instance to `TAP` reporter, and defer all writing of TAP output to it.
> > As TAP12 is apparently ancient nowadays, perhaps TAP13 should be unspecified default. Or leave the default "as is" and allow this to be added to a "semver-minor" Mocha release.
> 
> As the TAP13 spec is both forward and backward compatible I'm not sure keeping the TAP12 output around is necessary, what's your thoughts?

Thoughts/Questions:
* We have **no** idea who (or how many) people depend on Mocha's archaic TAP12 output format.
* Do `TAP::Spec::Parser` and `TAP::Parser::Grammar` modules accept this PR's TAP13 output?
* Does PR's TAP13 output format match that of other notable producers (e.g., `node-tap`)?
* Outside minor tweaks, Mocha's TAP12 format hasn't changed in about 6 years.
* Mocha releases occur glacially already -- would changing report format be restricted to `semver-major` releases?

While I think this is a fantastic upgrade, an alternate strategy would be to retain existing TAP12 as current default. Using the `TAPProducer` class to produce the output, we could add "experimental" support for TAP13 without being "married" to the format (if it doesn't substantially match). Updates could continue on it in `semver-minor` releases. Or perhaps document Mocha's output via a diagnostic line (e.g., #Mocha TAP13 format <number|date>) , enabling consumers to adjust (since [they never standardized YAML block formatting](https://github.com/TestAnything/testanything.github.io/issues/127#issuecomment-421392681)).

> > Have these changes been run against the [TAP 14 draft](https://github.com/TestAnything/testanything.github.io/pull/36)?
> 
> I had a look at the TAP14 draft back when implementing this and decided not to implement it since it is only a draft and I wanted to keep the footprint of the PR as small as possible.
>
> > Without personally really diving into TAP, does this reporter's output match that which would be produced by Node-based TAP producers `node-tap` and `tape`?
> 
> I did recreate the fixture in `tape`, mostly to test how they handle subtests/suites (as discussed in issue #2410). [Here is a gist of the tape version](https://gist.github.com/mollstam/44f7d3a314e0a4b22765f9ce445cafb1) and [here is a diff from tape report to this version of the mocha tap reporter](https://gist.github.com/mollstam/3cb828e93cb6a1ee3dd7d4e983794d16/revisions#diff-229da9bbc37f8a37f858791aa590ad53). The main difference is the tape report having more structured data (`operator`, `expected`, `actual`) in the YAML rather than simply "baking it" into the `message` property. I investigated reaching similar formatting but the footprint of the fix grew and I wanted to keep things simple.
> 
> As you can see the running theme is me wanting to make as little of change as possible to reach what I set out to do. This iteration of the TAP reporter is not by any means a "final form" but a lot of the future work relies on the YAML format being standardized (consumers vary _wildly_) and for that I think TAP14 needs to get ratified, or more effort to accommodate the popular consumers out there. 

Not holding my breath over [TAP14's ratification](https://github.com/TestAnything/testanything.github.io/issues/127#issuecomment-421342474). But changing Mocha's output _may_ be significant to the TAP ecosystem.

plroebuck:
> > On `suite end` event, should Mocha emit a timing comment line?
> IDK, its a very small fix so it might be useful? I can't find any consumer that use it though…

All other (non-toy) reporters report timing.

plroebuck:
> I looked at the [v13 spec](https://testanything.org/tap-version-13-specification.html) But it does not list/mention the breaking changes, can you provide the breaking changes to give us a better idea about what we're dealing with?

[So I asked...](https://github.com/TestAnything/testanything.github.io/issues/126).

plroebuck:
Suggest we ask for guidance from [TestAnything maintainers](https://github.com/orgs/TestAnything/people).

@jonathanKingston, @kinow, @mblayman: Care to way in on discussion?

kinow:
I don't know much about mocha, so sorry if I say anything here that sounds silly or not pertinent at all. 

>Do TAP::Spec::Parser and TAP::Parser::Grammar modules accept this PR's TAP13 output?

This is a very good question, that would help others to assess the compatibility with TAP13 (whether mocha users or not). Though I believe there are certain features from Perl modules that are not in TAP13 (IIRC, something with subtests, or the diagnostics).

>Does PR's TAP13 output format match that of other notable producers (e.g., node-tap)?

Another excellent question.

Having both question answered would be helpful for reviewers I think. 

While the output of TAP13 should be backward compatible with TAP12, from what I understand from this discussion, looks like there is a concern that consumers of the current output of mochajs could have their environments broken due to this different output for TAP13.

Implementers of TAP12 and TAP13 consumers and producers would all work OK, but there could be users consuming Mochajs TAP output (not necessarily TAP12 or TAP13).

So unless mochajs is planning a major release, that would allow for this change to happen, I think keeping the current status and adding TAP13 output as an alternative would be a sensible approach. Later, it could be marked/announced as deprecated, and during a major release it could be dropped (or enabled via a parameter), while TAP13 became the default (or maybe leave as is and aim at TAP14+).

mollstam:
I think adding a spec version as reporter option that defaults to current (12) is the right way to go, and hopefully at some major release the default can be changed to the latest stable implemented spec (13 or even 14 who knows). Will look into it.

plroebuck:
[How many of these consumers](https://github.com/mochajs/mocha/issues/492#issuecomment-45822546) you think this PR will break?

plroebuck:
> Or perhaps document Mocha's output via a diagnostic line (e.g., #Mocha TAP13 format <number|date>) , enabling consumers to adjust...

@kinow, do other producers follow this approach (or something like it)?

plroebuck:
We should document that we don't support [YAML diagnostics](https://www.node-tap.org/tap-format/). Your subtest output is considered unbuffered, right?

mollstam:
> Plan:
> 
> * Leave Mocha TAP12 as is (default).
> * Use this PR's format as Mocha TAP13 (via `reporterOptions.tap.version`).
> * Future TAP14 format would add support for `expected`/`actual` in YAML block.
> 
> @mollstam, what do you think?

Sounds good! I named the option `spec` before reading this, would you prefer version? I'm indifferent. :slightly_smiling_face:

mollstam:
@plroebuck I think the `expected`/`actual` stuff can be added to the TAP13 producer since it's just extra data in the YAML, we wouldn't destroy anything.

I _would_ like to also clean up the `stack` property to not include the message, but I'll have a look at that when I try to reach parity with other popular Producers, which will have to wait for next week.

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
  * @api public
  * @param {Runner} runner
  */
-function TAP(runner) {
+function TAP(runner, options) {
   Base.call(this, runner);
 
+  var self = this;
   var n = 1;
```
plroebuck: `n` is same as `runner.stats.tests`.

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+TAPProducer.prototype.writePending = function(n, test) {
+  console.log('ok %d %s # SKIP -', n, title(test));
+};
+
+TAPProducer.prototype.writePass = function(n, test) {
+  console.log('ok %d %s', n, title(test));
+};
+
+TAPProducer.prototype.writeEnd = function(runner) {
+  console.log('# tests ' + (runner.stats.passes + runner.stats.failures));
```
plroebuck: Just noticed this, but think this is probably an unreported existing bug.
`runner.stats.tests` will not agree with this math if pending tests occurred.

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+  console.log('# pass ' + runner.stats.passes);
+  console.log('# fail ' + runner.stats.failures);
+};
```
plroebuck: Should this include a summary line for pending tests?

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+
+/**
+ * Initialize a new `TAPProducer`. Should only be used as a base class.
+ *
+ * @private
+ * @class
+ * @api private
+ */
+function TAPProducer() {}
+
```
plroebuck: `TAP12Producer.prototype.writeStart` should be reparented as `TAPProducer.prototype.writeStart`.

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
...
-function TAP(runner) {
+function TAP(runner, options) {
   Base.call(this, runner);
 
+  var self = this;
   var n = 1;
-  var passes = 0;
-  var failures = 0;
+
+  var tapSpec = '12';
```
plroebuck: Should this be numeric instead?

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
...
 
   runner.on('start', function() {
-    var total = runner.grepTotal(runner.suite);
-    console.log('%d..%d', 1, total);
+    self.producer.writeStart(runner);
   });
 
   runner.on('test end', function() {
     ++n;
   });
```
plroebuck: Remove `test end` event handler.

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+ */
+function TAP12Producer() {}
+
+TAP12Producer.prototype.writeStart = function(runner) {
+  var total = runner.grepTotal(runner.suite);
+  console.log('%d..%d', 1, total);
+};
+
+TAP12Producer.prototype.writeFail = function(n, test, err) {
+  console.log('not ok %d %s', n, title(test));
```
plroebuck: `TAPProducer.prototype.writeFail.call(this, n, test, err)`

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+};
+
+TAP13Producer.prototype.writeStart = function(runner) {
+  console.log('TAP version 13');
+  var total = runner.grepTotal(runner.suite);
+  console.log('%d..%d', 1, total);
+};
+
+TAP13Producer.prototype.writeFail = function(n, test, err) {
+  console.log('not ok %d %s', n, title(test));
```
plroebuck: `TAPProducer.prototype.writeFail.call(this, n, test, err)`

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+ */
+function TAP13Producer() {}
+
+TAP13Producer.prototype.yamlIndent = function(level) {
+  return Array(level + 1).join('  ');
+};
+
+TAP13Producer.prototype.writeStart = function(runner) {
+  console.log('TAP version 13');
+  var total = runner.grepTotal(runner.suite);
```
plroebuck: `TAPProducer.prototype.writeStart.call(this, runner)`

lib/reporters/tap.js
plroebuck:
```js
TAPProducer.prototype.writeFail = function(runner, test, err) {
  console.log('not ok %d %s', runner.stats.tests, title(test));
}
```

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+inherits(TAP, Base);
+
+/**
+ * Initialize a new `TAPProducer`. Should only be used as a base class.
+ *
+ * @private
+ * @class
+ * @api private
+ */
+function TAPProducer() {}
```
plroebuck: Should constructor take a `runner` argument and skip passing runner to prototype methods?

test/integration/fixtures/reporters.fixture.js
```diff
@@ -0,0 +1,63 @@
...
+    it('should answer to "water bear"', function(done) { done(); });
+    it('should be able to survive global mass extinction events', function(done) {
+      throw new Error("How do we even test for this without causing one?")
+      done();
+    });
```
plroebuck: Can we add a pending test here?
```
it('just might one day');
```

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
...
-      console.log(err.stack.replace(/^/gm, '  '));
-    }
+    self.producer.writeFail(n, test, err);
   });
 
   runner.once('end', function() {
-    console.log('# tests ' + (passes + failures));
-    console.log('# pass ' + passes);
-    console.log('# fail ' + failures);
+    self.producer.writeEnd(runner);
```
plroebuck: `    self.producer.writeEnd(runner.stats);`

lib/reporters/tap.js
```diff
@@ -77,3 +79,95 @@ function TAP(runner) {
...
+
+TAPProducer.prototype.writePending = function(n, test) {
+  console.log('ok %d %s # SKIP -', n, title(test));
+};
+
+TAPProducer.prototype.writePass = function(n, test) {
+  console.log('ok %d %s', n, title(test));
+};
+
+TAPProducer.prototype.writeEnd = function(runner) {
```
plroebuck: `TAPProducer.prototype.writeEnd = function(stats) {`

lib/reporters/tap.js
plroebuck: `TAPProducer.prototype.writeStart = function(ntests) {`

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
...
+    }
+  }
+
+  if (tapSpec === '13') {
+    this.producer = new TAP13Producer();
+  } else {
+    this.producer = new TAP12Producer();
+  }
 
   runner.on('start', function() {
```
plroebuck:
```js
  runner.on('start', function() {
    var ntests = runner.grepTotal(runner.suite);
    self.producer.writeStart(ntests);
  });
```

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
  * @api public
  * @param {Runner} runner
  */
-function TAP(runner) {
+function TAP(runner, options) {
   Base.call(this, runner);
 
+  var self = this;
```
plroebuck: `  var stats = runner.stats;`

lib/reporters/tap.js
```diff
@@ -24,46 +25,47 @@ exports = module.exports = TAP;
...
+    self.producer.writeStart(runner);
   });
 
   runner.on('test end', function() {
     ++n;
   });
 
   runner.on('pending', function(test) {
-    console.log('ok %d %s # SKIP -', n, title(test));
+    self.producer.writePending(n, test);
```
plroebuck: `    self.producer.writePending(stats.tests, test);`

plroebuck:
In "test/reporters/tap.spec.js", let's make use of this function:

```js
  var showOutput = false;

  /**
   * Run reporter using stream reassignment to capture output.
   *
   * @param {Object} stubSelf - Reporter-like stub instance
   * @param {Runner} runner - Mock instance
   * @param {boolean} [tee=false] - If `true`, echo captured output to screen
   */
  function runReporter(stubSelf, runner, tee) {
    // Reassign stream in order to make a copy of all reporter output
    var stdoutWrite = process.stdout.write;
      stdout.push(string);
      if (tee) {
      }
    };

    // Invoke reporter
    TAP.call(stubSelf, runner);

    // Revert stream reassignment here so reporter output
    // can't be corrupted if any test assertions throw
  }
```

PR #3458, opened 2018-08-15
Refactor statistics

### LINKED ISSUE #3256: refactor statistics
Stat gathering (fails, passes, errors, duration) is too closely coupled to the reporter.  The reporter shouldn't be the single source of truth for this information, yet it is.  Statistics should be *queried by the reporters* instead of just dumped into the base reporter.  The result may be as simple as an abstraction between the runner and reporter.

This makes it difficult to modify the stats we gather, or how we gather them.   

related: #3223, #3250

craigtaub:
### Description of the Change

- a very small abstraction of stats collecting into a dedicated module
- "runner" now initiates collecting
- presentation is still handled by reporter
- no context is maintained internally, so invoked as normal function

### Alternate Designs

- an API for querying could be introduced but felt overkill unless others thought it worthwhile
- currently called from Base reporter but if it is useful the reporters could call it directly themselves (so we could interchange collectors).
- I was not sure if this should be called by Base reporter or the runner, so I chose reporter. If anyone else has opinions please let me know

### Benefits

- stats handling is not so closely tied to the Base reporter,
- easier to extend, test, isolate, enhance stats collecting

### Applicable issues

Fixes #3256
(semver-minor)

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -286,7 +286,7 @@ function Base(runner) {
   if (!runner) {
     throw new TypeError('Missing runner argument');
   }
-  this.stats = runner.stats; // attached so Reporters have a reference
+  this.stats = runner.stats; // assigned so Reporters keep a closer reference
   this.runner = runner;
 
   runner.on('pass', function(test) {
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -271,9 +271,7 @@ exports.list = function(failures) {
  * Initialize a new `Base` reporter.
  *
  * All other reporters generally
- * inherit from this reporter, providing
- * stats such as test duration, number
- * of tests passed / failed etc.
+ * inherit from this reporter.
  *
  * @memberof Mocha.reporters
  * @public
@@ -283,39 +281,15 @@ exports.list = function(failures) {
  */
 
 function Base(runner) {
-  var stats = (this.stats = {
-    suites: 0,
-    tests: 0,
-    passes: 0,
-    pending: 0,
-    failures: 0
-  });
   var failures = (this.failures = []);
 
   if (!runner) {
-    return;
+    throw new TypeError('Missing runner argument');
   }
+  this.stats = runner.stats; // attached so Reporters have a reference
   this.runner = runner;
 
-  runner.stats = stats;
-
-  runner.on('start', function() {
-    stats.start = new Date();
-  });
-
-  runner.on('suite', function(suite) {
-    stats.suites = stats.suites || 0;
-    suite.root || stats.suites++;
-  });
-
-  runner.on('test end', function() {
-    stats.tests = stats.tests || 0;
-    stats.tests++;
-  });
-
   runner.on('pass', function(test) {
-    stats.passes = stats.passes || 0;
-
     if (test.duration > test.slow()) {
       test.speed = 'slow';
     } else if (test.duration > test.slow() / 2) {
@@ -323,27 +297,14 @@ function Base(runner) {
     } else {
       test.speed = 'fast';
     }
-
-    stats.passes++;
   });
 
   runner.on('fail', function(test, err) {
-    stats.failures = stats.failures || 0;
-    stats.failures++;
     if (showDiff(err)) {
       stringifyDiffObjs(err);
     }
     test.err = err;
     failures.push(test);
-  });
-
-  runner.once('end', function() {
-    stats.end = new Date();
-    stats.duration = stats.end - stats.start;
-  });
-
-  runner.on('pending', function() {
-    stats.pending++;
   });
 }
 
--- a/lib/stats-collector.js
+++ b/lib/stats-collector.js
@@ -1,0 +1,70 @@
+'use strict';
+
+/**
+ * Test statistics collector.
+ *
+ * @typedef {Object} StatsCollector
+ * @property {number} suites - integer count of suites run.
+ * @property {number} tests - integer count of tests run.
+ * @property {number} passes - integer count of passing tests.
+ * @property {number} pending - integer count of pending tests.
+ * @property {number} failures - integer count of failed tests.
+ * @property {Date} start - time when testing began.
+ * @property {Date} end - time when testing concluded.
+ * @property {number} duration - number of msecs that testing took.
+ */
+
+/**
+ * Provides stats such as test duration,
+ * number of tests passed / failed etc.
+ *
+ * @public
+ * @memberof Mocha
+ * @param {Runner} runner
+ */
+function createStatsCollector(runner) {
+  var stats = {
+    suites: 0,
+    tests: 0,
+    passes: 0,
+    pending: 0,
+    failures: 0
+  };
+
+  if (!runner) {
+    throw new TypeError('Missing runner argument');
+  }
+
+  runner.stats = stats;
+
+  runner.once('start', function() {
+    stats.start = new Date();
+  });
+
+  runner.on('suite', function(suite) {
+    suite.root || stats.suites++;
+  });
+
+  runner.on('pass', function() {
+    stats.passes++;
+  });
+
+  runner.on('fail', function() {
+    stats.failures++;
+  });
+
+  runner.on('pending', function() {
+    stats.pending++;
+  });
+
+  runner.on('test end', function() {
+    stats.tests++;
+  });
+
+  runner.once('end', function() {
+    stats.end = new Date();
+    stats.duration = stats.end - stats.start;
+  });
+}
+
+module.exports = createStatsCollector;
--- a/test/reporters/list.spec.js
+++ b/test/reporters/list.spec.js
@@ -21,6 +21,7 @@ describe('List reporter', function() {
     duration: expectedDuration,
     slow: function() {}
   };
+
   beforeEach(function() {
     useColors = Base.useColors;
     Base.useColors = false;
@@ -126,7 +127,11 @@ describe('List reporter', function() {
       test = {};
       runner = createMockRunner('fail', 'fail', null, null, test);
       runner.on = runner.once = function(event, callback) {
-        if (!checked && event === 'fail') {
+        if (
+          !checked &&
+          event === 'fail' &&
+          callback.toString().includes('stringifyDiffObjs') // target correct fail event callback
+        ) {
           err = new Error('fake failure object with actual/expected');
           err.actual = actual;
           err.expected = expected;
--- a/test/reporters/xunit.spec.js
+++ b/test/reporters/xunit.spec.js
@@ -4,12 +4,14 @@ var fs = require('fs');
 var mkdirp = require('mkdirp');
 var path = require('path');
 var assert = require('assert');
+var createStatsCollector = require('../../lib/stats-collector');
 var reporters = require('../../').reporters;
 var XUnit = reporters.XUnit;
 
 describe('XUnit reporter', function() {
   var stdout;
   var stdoutWrite;
+  // the runner parameter of the reporter
   var runner;
 
   var callbackArgument = null;
@@ -26,6 +28,7 @@ describe('XUnit reporter', function() {
   beforeEach(function() {
     stdout = [];
     runner = {on: function() {}, once: function() {}};
+    createStatsCollector(runner);
   });
 
   describe('if reporter options output is given', function() {
@@ -322,8 +325,6 @@ describe('XUnit reporter', function() {
   describe('custom suite name', function() {
     // capture the events that the reporter subscribes to
     var events;
-    // the runner parameter of the reporter
-    var runner;
     // capture output lines (will contain the resulting XML of the xunit reporter)
     var lines;
     // the file stream into which the xunit reporter will write into
@@ -332,13 +333,10 @@ describe('XUnit reporter', function() {
     beforeEach(function() {
       events = {};
 
-      runner = {
-        on: function(eventName, eventHandler) {
-          // capture the event handler
-          events[eventName] = eventHandler;
-        }
+      runner.on = runner.once = function(eventName, eventHandler) {
+        // capture the event handler
+        events[eventName] = eventHandler;
       };
-      runner.once = runner.on;
 
       lines = [];
       fileStream = {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -12,6 +12,7 @@ var utils = require('./utils');
 var inherits = utils.inherits;
 var debug = require('debug')('mocha:runner');
 var Runnable = require('./runnable');
+var createStatsCollector = require('./stats-collector');
 var stackFilter = utils.stackTraceFilter();
 var stringify = utils.stringify;
 var type = utils.type;
@@ -81,6 +82,7 @@ function Runner(suite, delay) {
   this._defaultGrep = /.*/;
   this.grep(this._defaultGrep);
   this.globals(this.globalProps().concat(extraGlobals()));
+  createStatsCollector(this);
 }
 
 /**
--- a/test/reporters/helpers.js
+++ b/test/reporters/helpers.js
@@ -8,6 +8,8 @@
   scope of this function for the tests to run properly.
 */
 
+var createStatsCollector = require('../../lib/stats-collector');
+
 function createMockRunner(runStr, ifStr1, ifStr2, ifStr3, arg1, arg2) {
   var runnerFunction = createRunnerFunction(
     runStr,
@@ -17,10 +19,12 @@ function createMockRunner(runStr, ifStr1, ifStr2, ifStr3, arg1, arg2) {
     arg1,
     arg2
   );
-  return {
+  var mockRunner = {
     on: runnerFunction,
     once: runnerFunction
   };
+  createStatsCollector(mockRunner);
+  return mockRunner;
 }
 
 function createRunnerFunction(runStr, ifStr1, ifStr2, ifStr3, arg1, arg2) {


craigtaub:
Sorry about failing test, will investigate.

plroebuck:
Basic question: why is missing runner argument (to both Base and StatsCollector ctors) not considered an error?

plroebuck:
Why is 'statsCollector' class not 'StatsCollector'? [Class naming convention](http://snowdream.github.io/javascript-style-guide/javascript-style-guide/en/naming-conventions.html).

plroebuck:
"lib/stats-collector.js#L29":
  Realize your code just migrated existing, but why is `start` event not using `runner.once`?

"lib/stats-collector.js#L58":
Would also be nice to move the `pending` handler prior to the `end` handler.
(i.e., let `start` and `end` be bookend events).

craigtaub:
@plroebuck thanks for the help here

- regarding "test/reporters/list.spec.js", it required a specific event callback handler to run. Now that I have introduced a new callback handler on the same event it fails if the wrong handler is triggered. Its really not nice but I could not think of a better way to target it without touching live code. Will have another think if there is a cleaner way.
- your right will look at adding a throw for "missing runner argument" error.
- with naming followed conventions of camelCase as its a function (and is called as such), not a constructor or class. Let me know if you have any thoughts on this.
- I will investigate `start` using `runner.once` and let you know
- Completely agree with bookending, will change that

craigtaub:
@plroebuck have made some updates based on your comments, please have a look. Thanks

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+    stats.start = new Date();
+  });
+
+  runner.on('suite', function(suite) {
+    stats.suites = stats.suites || 0;
+    suite.root || stats.suites++;
+  });
+
+  runner.on('test end', function() {
+    stats.tests = stats.tests || 0;
```
plroebuck: So why exactly is this line needed?

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+    stats.tests++;
+  });
+
+  runner.on('pass', function(test) {
+    stats.passes = stats.passes || 0;
```
plroebuck: So why exactly is this line needed?

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+    stats.passes++;
+  });
+
+  runner.on('fail', function(test, err) {
+    stats.failures = stats.failures || 0;
```
plroebuck: So why exactly is this line needed?

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+    stats.failures++;
+  });
+
+  runner.on('pending', function() {
```
plroebuck: Notice it doesn't exist for pending.

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+/**
+ * Stats Collector
+ *
+ * Provides stats such as test duration,
+ * number of tests passed / failed etc.
+ *
+ * @public
+ * @api public
```
plroebuck: Delete `@api` line - just use standard JSDoc - this was referenced by one of TJ's old packages.

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
```
plroebuck:
```js
/**
 * Reporter statistics.
 *
 * @typedef {Object} StatsCollector
 * @property {number} suites - integer count of suites run.
 * @property {number} tests - integer count of tests run.
 * @property {number} passes - integer count of passing tests.
 * @property {number} pending - integer count of pending tests.
 * @property {number} failures - integer count of failed tests.
 */
```

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+ * Stats Collector
+ *
+ * Provides stats such as test duration,
+ * number of tests passed / failed etc.
+ *
+ * @public
+ * @api public
+ * @memberof Mocha
+ * @param {Runner} runner
+ * @return {stats}
```
plroebuck: `@return {StatsCollector}`

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+/**
+ * Stats Collector
```
plroebuck: Initializes test statistics collector object.

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+    stats.tests = stats.tests || 0;
+    stats.tests++;
+  });
+
+  runner.on('pass', function(test) {
+    stats.passes = stats.passes || 0;
+    stats.passes++;
+  });
+
+  runner.on('fail', function(test, err) {
```
plroebuck: Maybe just since no need for args?
`runner.on('fail', function() {`

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+    stats.suites = stats.suites || 0;
+    suite.root || stats.suites++;
+  });
+
+  runner.on('test end', function() {
+    stats.tests = stats.tests || 0;
+    stats.tests++;
+  });
+
+  runner.on('pass', function(test) {
```
plroebuck: Maybe just this since no need for args?
`runner.on('pass', function() {`

test/reporters/list.spec.js
```diff
@@ -138,10 +138,15 @@ describe('List reporter', function() {
       var actual = {a: 'actual'};
       var expected = {a: 'expected'};
       var checked = false;
+      var stringifyDiffs = 'stringifyDiffObjs';
       var err;
       test = {};
       runner.on = runner.once = function(event, callback) {
```
plroebuck: If this is set to both `.on` and `.once`, aren't we calling it twice for same issue?

test/reporters/list.spec.js
```diff
@@ -138,10 +138,15 @@ describe('List reporter', function() {
...
-        if (!checked && event === 'fail') {
+        if (
+          callback.toString().includes(stringifyDiffs) && // handle correct callback
```
plroebuck: Sorry. Still hate this. Are we sure this is the **only** reporter affected? No other way to make this work correctly?

plroebuck:
Modify your PR  above as follows:
* Replace existing contents of the "Applicable Issues" section with the text below (so the issue will be closed automatically upon merge and label this as a minor enhancement).
```text
Fixes #3256
(semver-minor)
```

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+/**
+ * Stats Collector
+ *
+ * Provides stats such as test duration,
+ * number of tests passed / failed etc.
+ *
+ * @public
```
plroebuck: Is this being exported somewhere? Otherwise, should be `@private`

lib/stats-collector.js
```diff
@@ -0,0 +1,65 @@
...
+  runner.once('start', function() {
+    stats.start = new Date();
+  });
+
+  runner.on('suite', function(suite) {
+    stats.suites = stats.suites || 0;
+    suite.root || stats.suites++;
+  });
+
+  runner.on('test end', function() {
```
plroebuck: Move to before `runner.once('end', ` such that events listed in order of occurrence.

plroebuck: @boneskull, why does Mocha count "suites" as soon as they start, yet count "tests"  only when they complete?

lib/stats-collector.js
craigtaub: I assumed this was flagging the module/function itself as public (is that right?). Its exported and imported by base reporter.

lib/stats-collector.js
craigtaub: Yep not needed will remove them all.

test/reporters/list.spec.js
craigtaub: I dont think so. `once` touches `start`+`end` events and `on` touches everything else.
As the assertion here doesn't need `start` or `end` I believe I can remove the `once`.

lib/stats-collector.js
craigtaub: Added property docs below just below section on module.

lib/reporters/base.js
```diff
@@ -11,6 +11,7 @@ var diff = require('diff');
 var ms = require('../ms');
 var utils = require('../utils');
 var supportsColor = process.browser ? null : require('supports-color');
+var statsCollector = require('../stats-collector');
```
plroebuck: `var initStatsCollector = require('../stats-collector');`

test/reporters/list.spec.js
```diff
@@ -138,10 +138,16 @@ describe('List reporter', function() {
       var actual = {a: 'actual'};
       var expected = {a: 'expected'};
       var checked = false;
+      var stringifyDiffs = 'stringifyDiffObjs';
       var err;
       test = {};
-      runner.on = runner.once = function(event, callback) {
-        if (!checked && event === 'fail') {
+      runner = createMockRunner('fail', 'fail', null, null, test);
```
plroebuck: Maybe append `runner = null;` to the `afterEach()` function.

lib/reporters/base.js
plroebuck: `var createStatsCollector = require('../stats-collector');`

lib/reporters/base.js
```diff
@@ -283,68 +282,31 @@ exports.list = function(failures) {
...
-    pending: 0,
-    failures: 0
-  });
   var failures = (this.failures = []);
 
   if (!runner) {
-    return;
+    throw new Error('Missing runner argument');
   }
+  this.stats = statsCollector(runner);
```
plroebuck: `  this.stats = createStatsCollector(runner);`

lib/stats-collector.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+/**
+ * Initializes test statistics collector object.
```
plroebuck: * Creates test statistics collector object.

lib/stats-collector.js
```diff
@@ -0,0 +1,72 @@
...
+ * @property {number} suites - integer count of suites run.
+ * @property {number} tests - integer count of tests run.
+ * @property {number} passes - integer count of passing tests.
+ * @property {number} pending - integer count of pending tests.
+ * @property {number} failures - integer count of failed tests.
+ * @property {Date} start - time when testing began.
+ * @property {Date} end - time when testing concluded.
+ * @property {number} duration - number of msecs that testing took.
+ */
+function initStatsCollector(runner) {
```
plroebuck: `function createStatsCollector(runner) {`

Sorry for the naming waffling here...

lib/stats-collector.js
```diff
@@ -0,0 +1,72 @@
...
+
+  runner.once('end', function() {
+    stats.end = new Date();
+    stats.duration = stats.end - stats.start;
+  });
+
+  return stats;
+}
+
+module.exports = initStatsCollector;
```
plroebuck: `module.exports = createStatsCollector;`

test/reporters/list.spec.js
```diff
@@ -138,10 +138,16 @@ describe('List reporter', function() {
...
+      runner.on = function(event, callback) {
+        if (
+          callback.toString().includes(stringifyDiffs) && // target correct fail event callback
```
plroebuck: What is actual intent here? Are you lazily trying to see if `stringifyDiffObjs` _is_ the callback, or if the callback invokes `stringifyDiffObjs`?

lib/stats-collector.js
plroebuck: One last thing here. I gave the object description as "Reporter statistics.", but since we refer to it as "Test statistics" everywhere else, this should probably be changed to match.

test/reporters/list.spec.js
```diff
@@ -138,10 +138,16 @@ describe('List reporter', function() {
       var actual = {a: 'actual'};
       var expected = {a: 'expected'};
       var checked = false;
+      var stringifyDiffs = 'stringifyDiffObjs';
```
plroebuck: As this is used nowhere else, migrate it inside the event handler callback.

test/reporters/list.spec.js
plroebuck: Might also migrate this check to the end of the `if` condition. It's more expensive than the other two checks.

plroebuck: Might also migrate this to the end of the `if` condition. It's much more expensive than the other two conditionals.

lib/stats-collector.js
craigtaub: Sorry I find this super confusing, the function itself is exported though (i.e. `module.exports = createStatsCollector`). Its publicly available for anything else to import from anywhere.
Do you have an example perhaps of something similar which is private?

test/reporters/list.spec.js
craigtaub: Yep will do

test/reporters/list.spec.js
craigtaub: But then the same value is declared and assigned "every single time" a callback runs (and with this test it happens quite a lot). Isn't it better its assigned beforehand and then re-used?

test/reporters/list.spec.js
craigtaub: The latter...callback invokes `stringifyDiffObjs`. Will update var name.

lib/stats-collector.js
craigtaub: Ok have now separated and updated them.

craigtaub:
@plroebuck made some updates from your comments

test/reporters/list.spec.js
```diff
@@ -138,15 +139,15 @@ describe('List reporter', function() {
...
       var err;
       test = {};
       runner = createMockRunner('fail', 'fail', null, null, test);
       runner.on = function(event, callback) {
         if (
-          callback.toString().includes(stringifyDiffs) && // target correct fail event callback
           !checked &&
-          event === 'fail'
+          event === 'fail' &&
+          callback.toString().includes(callbackWhichStringifies) // target correct fail event callback
```
plroebuck: How about just ditch the variable altogether?
```js
          callback.toString().includes('stringifyDiffObjs') // target correct fail event callback
```

lib/stats-collector.js
plroebuck: Mea culpa. I was essentially documenting `@api private` rather than `@public`. Sorry for the confusion.

lib/stats-collector.js
```diff
@@ -1,19 +1,21 @@
...
  * Provides stats such as test duration,
  * number of tests passed / failed etc.
  *
  * @public
  * @memberof Mocha
  * @param {Runner} runner
- * @return {StatsCollector}
- *
- * Reporter statistics.
+ * @return {stats}
```
plroebuck: Revert this change.

lib/stats-collector.js
```diff
@@ -1,19 +1,21 @@
...
+ */
+
+/**
+ * {stats} object properties
  *
- * @typedef {Object} StatsCollector
+ * @typedef {Object} stats
```
plroebuck: Revert this change. We're documenting what we're creating. As the function is named `createStatsCollector()`, the `@typedef` should match - hence:
`@typedef {Object} StatsCollector`.

lib/stats-collector.js
```diff
@@ -1,19 +1,21 @@
...
  * @memberof Mocha
  * @param {Runner} runner
- * @return {StatsCollector}
- *
- * Reporter statistics.
+ * @return {stats}
+ */
+
+/**
+ * {stats} object properties
```
plroebuck:
```text
 * Test statistics
```

lib/stats-collector.js
plroebuck: `@typedef` docblock should **preceed** function docblock.

Invert them so function docblock is paired with the function.

plroebuck:
Before merge, we should update the [third-party reporter documentation](https://github.com/mochajs/mocha/wiki/Third-party-reporters) page as follows:
1. Remove the `passes` and `failures` variables.
1. Change the `end` event handler to get pass/fail information from new `stats` object.
1. Remove `process.exit` call (which should not have been there to begin with).

Bamieh:
@plroebuck @craigtaub can you ping me when the code is ready 🙏 ?

plroebuck:
@craigtaub, have you had a chance to look at the requested changes yet?

craigtaub:
@Bamieh yep no problem.

@plroebuck unfortunately not yet (just started new job), will let you know once I have (hoping to within few days).

plroebuck:
After going through something else, I no longer think the stats collector should belong to the reporter. Rather it belongs to the runner (though `Base` reporter can keep a reference for convenience).

Could be placed in the `Runner` [constructor](https://github.com/mochajs/mocha/blob/d401ee198bd5026bc61641ad26eab31e6746713f/lib/runner.js#L66-L84), somewhere after the existing event handlers (or at the end of function).

Alternatively, it could be placed in `Mocha.prototype.run()` immediately after runner instantiation [here](https://github.com/mochajs/mocha/blob/d401ee198bd5026bc61641ad26eab31e6746713f/lib/mocha.js#L581), but the former seems more appropriate.

lib/stats-collector.js
craigtaub: @plroebuck  sorry I don't follow this comment, it is not a `Test` right?

test/reporters/list.spec.js
craigtaub: Sure will do.

test/reporters/list.spec.js
```diff
@@ -32,6 +32,7 @@ describe('List reporter', function() {
   });
```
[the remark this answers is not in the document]
craigtaub: Done

lib/stats-collector.js
craigtaub: @plroebuck I have now reverted this change.

craigtaub:
@plroebuck regarding the Runner, it's a fair point. I did initially think about reporter vs runner but decided for simplicity sake (and I could not think of a better reason) to keep it in reporter.

How strongly do you feel that change needs to come in this PR?

We have abstracted the stats which feels like a good first step. I am happy to create a new issue referencing your comment which would move it to the Runner.

lib/stats-collector.js
plroebuck: [Like this](https://github.com/mochajs/mocha/blob/fe623c745c44085ee79b1978e98969454d662848/lib/stats-collector.js#L4):
```js
/**
 * Test statistics collector.
 *
 * @typedef {Object} StatsCollector
...
```

craigtaub: My query is around the term "Test". What is this module testing?

plroebuck:
> Regarding the Runner, it's a fair point. I did initially think about reporter vs runner but decided for simplicity sake (and I could not think of a better reason) to keep it in reporter.

Had done the same.

> How strongly do you feel that change needs to come in this PR?
>
> We have abstracted the stats which feels like a good first step. I am happy to create a new issue referencing your comment which would move the required functionality to the Runner. Personally I am a fan of splitting work into chunks and if you agree at least we could move this PR on.

Would really prefer to finish in this PR; thus far, test statistics collection was refactored into its own file, but not really moved from the `Base` reporter. Moving the stats collector to the `Runner` (and leaving a reference in `Base` reporter) is pretty trivial.

craigtaub:
@plroebuck wasn't the notion to remove the connection from Base reporter only introduced 2 days ago? In't this why best practice is to aim for increments not big bang?

Honestly there is a real frustration to chasing PR approvals in such a way (for me feels like this PR has been open forever). Specially for volunteer work which has to be done in spare time. Changes in small chunks is surely ideal.

plroebuck:
Moving it would appear to be less than 10 lines of code.
But if you're tired of it, I'll do it.

craigtaub:
@plroebuck I have not tried it, that was not the point of my comment.
I won't have time for a few weeks but if you leave it I will get to it eventually. Feel free to make the changes yourself.

craigtaub:
@plroebuck commits squashed.

lib/reporters/base.js
```diff
@@ -283,68 +282,31 @@ exports.list = function(failures) {
...
-    pending: 0,
-    failures: 0
-  });
   var failures = (this.failures = []);
 
   if (!runner) {
-    return;
+    throw new Error('Missing runner argument');
   }
+  this.stats = getStats(); // attached so Reporters have a reference
```
plroebuck: Why isn't this
`this.stats = runner.stats;`

craigtaub: Yep good call, have made change + build passing. 
If theres nothing else could you approve it please and I will "Rebase and merge".

lib/reporters/base.js
```diff
@@ -283,68 +281,31 @@ exports.list = function(failures) {
...
-    tests: 0,
-    passes: 0,
-    pending: 0,
-    failures: 0
-  });
   var failures = (this.failures = []);
 
   if (!runner) {
-    return;
+    throw new Error('Missing runner argument');
```
plroebuck: Should _technically_ be `TypeError`

test/reporters/xunit.spec.js
```diff
@@ -104,6 +105,7 @@ describe('XUnit reporter', function() {
           callback();
         }
       };
+      runner.stats = createStatsCollector(runner);
```
plroebuck: Better here or L29.5?

test/reporters/xunit.spec.js
```diff
@@ -339,6 +341,7 @@ describe('XUnit reporter', function() {
         }
       };
       runner.once = runner.on;
+      runner.stats = createStatsCollector(runner);
```
plroebuck: Leave this to your call as partially outside your scope (don't remember "xunit.spec.js" being in previous commits).

The `runner` variable at L328 masks the one at L14. Would it be better to remove it and change L337-L344 to this?
```js
runner.on = runner.once = function(eventName, eventHandler) {
  // capture the event handler
  events[eventName] = eventHandler;
};
```

craigtaub: Yes I agree its better to define `runner` (and attach stats) once and just re-use. I need to keep an eye out for that in these tests.
Have now made that change.

BTW yep the original commit didn't have this change, it came from the ownership changing from Reporter to Runner.

lib/stats-collector.js
```diff
@@ -0,0 +1,73 @@
...
+  var stats = {
+    suites: 0,
+    tests: 0,
+    passes: 0,
+    pending: 0,
+    failures: 0
+  };
+
+  if (!runner) {
+    throw new Error('Missing runner argument');
```
plroebuck: TypeError. Sorry, missed this one last go-round.

test/reporters/xunit.spec.js
plroebuck: > BTW yep the original commit didn't have this change, it came from the ownership changing from Reporter to Runner.

Any idea why? Both Base and Runner had a reference to the same "object". Who created it shouldn't have mattered, unless I'm really missing something.

lib/stats-collector.js
craigtaub: Ah ok, done.

test/reporters/xunit.spec.js
craigtaub: The Runner in this test is a mocked object. So the Reporter + Base require the mock to have a "stats" property they can attach to.
Does that make sense? I hope I haven't misunderstood what you meant.

craigtaub: Hmm did you mean does it need the assignment? I think your right + no assignment is needed.

craigtaub:
No worries @plroebuck . Apologies this one took so long. 
Hopefully my last set of changes made sense.

PR #3466, opened 2018-08-21
update slow description

finfin:
Update mocha doc to have more details on the "slow" configuration


plroebuck:
Nitpicky, but suggest moving the default value reference from end of line to its inline reference.
Reworded slightly for verb agreement.

```text
Many reporters will display test duration and flag tests that are slow (default: 75ms), as shown here with the "spec" reporter:
```

docs/index.md
```diff
@@ -665,10 +665,16 @@ $ mocha
...
 <h2 id="test-duration">Test duration</h2>
 
-Many reporters will display test duration, as well as flagging tests that are slow, as shown here with the "spec" reporter:
+Many reporters will display test duration and flag tests that are slow (default: 75ms), as shown here with the "spec" reporter:
 
 ![test duration](images/reporter-spec-duration.png?withoutEnlargement&resize=920,9999){:class="screenshot"}
 
+There are three levels of test duration:
+
+1. FAST: Tests run under half of the "slow" threshold is considered "FAST". The duration will be shown in green or not shown at all.
```
plroebuck: ...considered "fast".

docs/index.md
```diff
@@ -665,10 +665,16 @@ $ mocha
...
+2. REASONABLE (in yellow): Tests duration under threshold but above half of threshold will be shown in yellow.
```
plroebuck: MEDIUM: Tests that run above half of threshold (but under the threshold) will show the duration in yellow.

docs/index.md
```diff
@@ -665,10 +665,16 @@ $ mocha
...
+3. SLOW (in red): Tests that runs over the threshold.
```
plroebuck: SLOW: Tests that run exceeding the threshold will show the duration in red.

plroebuck:
Honestly think most of this textual description should be replaced with a single line bar graph image with markers [0, slow/2, slow, timeout] with the colors [green, yellow, red]. It would communicate same thing far more clearly.

boneskull:
thanks!

PR #3467, opened 2018-08-22
Added custom error objects including code

### LINKED ISSUE #3125: enhancement: error codes!
Every `Error` generated by Mocha should have a `code` property, like [Node.js now has](https://nodejs.org/api/errors.html#nodejs-error-codes).

We should keep a nice wiki page or table of the error codes alongside more helpful information, and  provide a URL to look this up.

These needn't be implemented all at once.

craigtaub:
### Description of the Change

- introduced error factories which append a `code` property
- the output/stack of an error printed in the browser/terminal should be near identical.
- added details of them to Documentation
- did not add the new error objects to test failures, tests themselves and configs.

### Alternative design

- could use a single custom error object and pass code `new MochaError('arg missing message', 'CODE')`.
- could use "custom errors classes" - nightmare getting ES5 call stacks to work on different browsers

### Benefits of custom error objects

- Keep errors concise (still 1 liners) + allows us to maintain terse error flow (i.e. less noise)
- Errors less brittle (get `code` out the box)
- using codes is a mandated best practice (following NodeJS i.e. https://nodejs.org/dist/latest-v10.x/docs/api/errors.html#errors_node_js_error_codes)
- names easier to infer about which to use

### Applicable issues

Fixes #3125
(semver-minor)


craigtaub:
Can't replicate that AppVeyor error locally + didnt happen on previous commit.

craigtaub:
@plroebuck so do u think it should throw TypeError and append code everytime or use custom but inherit TypeError? 

Felt nasty having custom MochaError but more nasty adding prop.

plroebuck:
Was any consideration given to creating specific error classes? Personally, I'd find "InvalidReporterError" a far more descriptive error than "MochaError" (also fine with the former being subclass of latter).

plroebuck:
Think "lib/error.js" should be renamed "lib/errors.js". Error codes should be exported as constants and those constants used instead. Think the following would be more suitable constant names:
* ERR_MOCHA_INVALID_ARG_VALUE
* ERR_MOCHA_INVALID_INTERFACE 
* ERR_MOCHA_INVALID_REPORTER
* ERR_MOCHA_MISSING_ARGUMENT 
* ERR_MOCHA_MISSING_CALLBACK
* ERR_MOCHA_NO_FILES_MATCH_PATTERN
* ERR_MOCHA_NOT_SUPPORTED
* ERR_MOCHA_UNDEFINED_ERROR

plroebuck:
"lib/error.js#L7-9":

why can't that just be `Error.captureStackTrace(this, MochaError);`

craigtaub:
@plroebuck thanks for the review.

- Nope I gave no consideration to specific error classes, but I like the idea. Could tie the `code` with the error itself then. Might not make constants in that case as feels superfluous (as can just check error type). Let me know if you don't agree.
- Thanks for help with the naming
- Didn't know about `Error.captureStackTrace` so will try that instead.
- Feel the custom errors are more preferable here to a wrapper function, let me know if you don't agree (we can always see how it looks once its done before deciding).

Might take me a while to get these changes done I hope thats ok.

craigtaub:
@plroebuck had a go at adding custom error objects with constant codes. Have not fully fleshed out the JSDocs for it yet, was not sure if there was a cleaner approach.

bin/_mocha
```diff
@@ -19,6 +19,9 @@ const interfaceNames = Object.keys(Mocha.interfaces);
 const join = path.join;
 const cwd = process.cwd();
 const getOptions = require('./options');
+const errors = require('../lib/errors');
+const ERR_MOCHA_NO_FILES_MATCH_PATTERN =
```
plroebuck: Maybe this instead?
```js
const NoFilesMatchPatternError = errors.NoFilesMatchPatternError;
```

bin/_mocha
```diff
@@ -515,7 +518,7 @@ args.forEach(arg => {
   try {
     newFiles = utils.lookupFiles(arg, extensions, program.recursive);
   } catch (err) {
-    if (err.message.indexOf('cannot resolve path') === 0) {
+    if (err.code === ERR_MOCHA_NO_FILES_MATCH_PATTERN) {
```
plroebuck: Better as:
```js
    if (err instanceof NoFilesMatchPatternError {
```

bin/_mocha
```diff
@@ -515,7 +518,7 @@ args.forEach(arg => {
...
       console.error(
```
plroebuck:
```js
console.warn(...);
```

docs/index.md
```diff
@@ -1302,6 +1303,44 @@ $ mocha "./spec/**/*.js"
...
+List of codes associated with Errors thrown inside Mocha. Following NodeJS practices.
+
+| Code        | Meaning |
+| ------------- | ------------- |
+| ERR_MOCHA_NO_FILES_MATCH_PATTERN | file/s of test could not be found |
+| ERR_MOCHA_MISSING_CALLBACK | callback for a running suite was not found |
+| ERR_MOCHA_INVALID_REPORTER | reporter specified in options not found |
+| ERR_MOCHA_INVALID_INTERFACE | interface specified in options not found |
+| ERR_MOCHA_NOT_SUPPORTED | type of output specified was not supported |
+| ERR_MOCHA_INVALID_ARG_VALUE | an invalid type of argument was found |
```
plroebuck:
```
ERR_MOCHA_INVALID_ARG_VALUE | an invalid argument value was given 
```

docs/index.md
```diff
@@ -1302,6 +1303,44 @@ $ mocha "./spec/**/*.js"
...
+| ERR_MOCHA_MISSING_ARGUMENT | an argument was missing |
+| ERR_UNDEFINED_ERROR | an error was thrown but no details were specified |
+
+To throw a custom Error please follow below:
```
plroebuck: Not sure this section is relevant to user of Mocha...

docs/index.md
```diff
@@ -1302,6 +1303,44 @@ $ mocha "./spec/**/*.js"
...
+
+```javascript
+// import custom error object
+const errors = require('./errors');
+const ErrorInvalidArgValue = errors.ErrorInvalidArgValue;
+
+throw new ErrorInvalidArgValue('Our error message');
+```
+
+To use the error object match its code against those available like below:
```
plroebuck: While correct, perhaps better described with custom error instead...

lib/errors.js
```diff
@@ -0,0 +1,99 @@
...
+ * Custom Errors objects
+ */
+
+function NoFilesMatchPattern(message) {
+  this.name = 'NoFilesMatchPattern';
+  this.code = ERR_MOCHA_NO_FILES_MATCH_PATTERN;
+  this.message = message;
+  Error.captureStackTrace(this, NoFilesMatchPattern);
+}
+NoFilesMatchPattern.prototype = Error.prototype;
```
plroebuck: Had always seen this done like:
```js
MyError.prototype = Object.create(Error.prototype);
MyError.prototype.constructor = MyError;
```

lib/errors.js
```diff
@@ -0,0 +1,99 @@
...
+module.exports.ERR_MOCHA_INVALID_INTERFACE = ERR_MOCHA_INVALID_INTERFACE;
+module.exports.ERR_MOCHA_NOT_SUPPORTED = ERR_MOCHA_NOT_SUPPORTED;
+module.exports.ERR_MOCHA_INVALID_ARG_VALUE = ERR_MOCHA_INVALID_ARG_VALUE;
+module.exports.ERR_MOCHA_MISSING_ARGUMENT = ERR_MOCHA_MISSING_ARGUMENT;
+module.exports.ERR_MOCHA_UNDEFINED_ERROR = ERR_MOCHA_UNDEFINED_ERROR;
+
+/**
+ * Custom Errors objects
+ */
+
```
plroebuck: As question for all concerned, should design be to have everything be completely independent or to subclass from primary? Could see advantage for user in having latter capability being able to test caught error as Mocha-related or not.

lib/suite.js
```diff
@@ -50,7 +52,7 @@ exports.create = function(parent, title) {
  */
 function Suite(title, parentContext) {
   if (!utils.isString(title)) {
-    throw new Error(
+    throw new InvalidArgumentValue(
```
plroebuck: Should be `TypeError`

lib/test.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 var Runnable = require('./runnable');
 var utils = require('./utils');
+var MochaError = require('./errors');
```
plroebuck: This doesn't work...

lib/test.js
```diff
@@ -15,10 +16,11 @@ module.exports = Test;
  */
 function Test(title, fn) {
   if (!isString(title)) {
-    throw new Error(
+    throw new MochaError(
```
plroebuck: `TypeError`

lib/utils.js
```diff
@@ -516,7 +520,9 @@ exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
     } else {
       files = glob.sync(filepath);
       if (!files.length) {
-        throw new Error("cannot resolve path (or pattern) '" + filepath + "'");
+        throw new NoFilesMatchPattern(
```
plroebuck: This one could have an extra field `pattern`.

lib/utils.js
```diff
@@ -547,7 +553,7 @@ exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
       return;
     }
     if (!extensions) {
-      throw new Error(
+      throw new MissingArgument(
```
plroebuck: Maybe this one too. `argument`

test/integration/suite.spec.js
```diff
@@ -12,7 +12,7 @@ describe('suite w/no callback', function() {
         done(err);
         return;
       }
-      var result = res.output.match(/no callback was supplied/) || [];
+      var result = res.output.match('MissingCallback') || [];
```
plroebuck:
```js
      var pattern = /MissingCallbackError/g;
      var result = res.output.match(pattern) || [];
```

boneskull:
Can't believe I missed this PR, thanks @craigtaub.

boneskull:
I'm concerned these custom error classes will break in IE11, or maybe any browser.  I'd need to see a test that runs in the browser before I believe it works properly.

That said, I'm *generally philosophically opposed* to custom error classes in ES5.  ES6 makes this more palatable as it's easy to just extend `Error`, but in other cases, you have to worry about the constructor implementation to ensure the stack trace shows up and works like it should.

For example, `Error.captureStackTrace()` is a node-ism... or maybe V8-specific.  

The alternative, then, is to create factory functions which instantiate (*not* throw-and-instantiate) plain-ass-`Error`s and give them error code properties.

boneskull:
If we can get this working well very soon, I'd love to have it in v6.0.0

craigtaub:
@boneskull its a real shame but i think your right the custom errors will have to be dumped due to browser issues
- Chrome + Safari work 100%
- IE11 works 100% but only if I omit V8-specific `Error.captureStackTrace` part (easy enough)
- FF works with message but don't get stack at all, which we obviously need

😞

Will have a look at using a factory to handle it.

craigtaub:
@boneskull could you have another look please. Its a more simple approach now using a factory not attempting error inheritance.

The biggest negative to this approach is that the top of the error stack now points to our factories `Error` instantiation, and second to the code which `throw`'s. 
Also worth noting they aren't constructors but followed normal error convention + kept them PascalCase.
Thoughts?

craigtaub:
@plroebuck do u have any thoughts on the latest change?
Essentially on different browsers had issues with call stack, so stripped it back to the basics.

boneskull:
@craigtaub 	conflicts need resolving

lib/cli/run-helpers.js
```diff
@@ -143,10 +143,8 @@ exports.handleFiles = ({
     try {
       newFiles = utils.lookupFiles(arg, extension, recursive);
     } catch (err) {
-      if (err.message.indexOf('cannot resolve path') === 0) {
-        console.error(
-          `Warning: Could not find any test files matching pattern: ${arg}`
-        );
+      if (err.code === 'ERR_MOCHA_NO_FILES_MATCH_PATTERN') {
+        console.warn(`Warning: ${err.message}: ${err.pattern}`);
```
plroebuck: `console.error`. this isn't a warning

lib/errors.js
```diff
@@ -0,0 +1,102 @@
...
+  return errorWithCode;
+}
+
+/**
+ * @summary
+ * An argument was missing.
+ *
+ * @public
+ */
+function MissingArgumentError(message, argument) {
```
plroebuck: All calls to `MissingArgumentError` should be converted to [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/More_arguments_needed)

lib/interfaces/common.js
```diff
@@ -147,7 +149,7 @@ module.exports = function(suites, context, mocha) {
           }
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
-          throw new Error(
+          throw MissingCallbackError(
```
plroebuck: seems hinky **not** to be using `new` when _creating_ error objects...

test/integration/file-utils.spec.js
```diff
@@ -75,11 +75,11 @@ describe('file utils', function() {
       var dirLookup = function() {
         return utils.lookupFiles(tmpDir, undefined, false);
       };
-      expect(
-        dirLookup,
-        'to throw',
-        'extensions parameter required when filepath is a directory'
-      );
+      expect(dirLookup, 'to throw', {
+        message: 'Parameter required when filepath is a directory',
```
plroebuck: Should we remove the `message`? The _point_ was to interrogate the `code` so spelling, etc. could safely be ignored.

lib/mocha.js
```diff
@@ -249,7 +252,7 @@ Mocha.prototype.reporter = function(reporter, reporterOptions) {
       );
```
plroebuck: At lines (L226, L231), be nice to add the reciprocal processing...
```js
 if (err.code !== 'MODULE_NOT_FOUND' ||
     err.message.indexOf('Cannot find module') !== -1) {
```

lib/errors.js
```diff
@@ -0,0 +1,102 @@
+'use strict';
+/**
+ * @module Errors
+ */
+/**
+ * Factory functions to create throwable error objects
+ */
+
+/**
+ * @summary
```
plroebuck: `@summary` is _implied_. You only really need to specify this tag if you also provide a description.

lib/suite.js
```diff
@@ -49,7 +49,7 @@ exports.create = function(parent, title) {
  */
 function Suite(title, parentContext) {
   if (!utils.isString(title)) {
-    throw new Error(
+    throw new TypeError(
```
plroebuck:
```js
var err = new TypeError(`Suite argument 'title' must be a string. Received type ${typeof title}`);
err.code = 'ERR_INVALID_ARG_TYPE'
throw err;
```

lib/test.js
```diff
@@ -15,7 +15,7 @@ module.exports = Test;
  */
 function Test(title, fn) {
   if (!isString(title)) {
-    throw new Error(
+    throw new TypeError(
```
plroebuck: see previous

lib/cli/run-helpers.js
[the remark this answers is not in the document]
boneskull: Something like `console.warn('Warning: %s: %O', err.message, err.pattern)` would also work, since Node.js' console methods run parameters thru `util.format`.  This would save the trouble of calling `JSON.stringify()` in a template string, anyhow.

lib/cli/run.js
```diff
@@ -190,7 +192,7 @@ exports.builder = yargs =>
             const pair = opt.split('=');
 
             if (pair.length > 2 || !pair.length) {
-              throw new Error(`invalid reporter option '${opt}'`);
+              throw InvalidReporterError(`invalid reporter option '${opt}'`);
```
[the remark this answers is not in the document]
boneskull: definitely needs a factory function though to save the trouble

lib/mocha.js
boneskull: In Node.js v6 you can use `String.prototype.includes`; `err.message.includes('Cannot find module')`.

lib/errors.js
```diff
@@ -0,0 +1,102 @@
...
+ * File/s of test could not be found.
+ *
+ * @public
+ */
+function NoFilesMatchPatternError(message, pattern) {
```
boneskull: Assuming we're not going to use custom constructors and instead we're just going to be assigning properties on the `Error` object... I would prefer we don't add *additional* properties (e.g., `pattern`, `argument`) to the error object, for sake of simplicity.  We're probably not going to need them.  Let's just stuff these in the `message` and call it good?  What's the use case?

This contradicts what @plroebuck commented on, so would like to understand why you and he feel this is needed.

lib/errors.js
boneskull: I disagree that we need to consume `TypeError`, `ReferenceError`, etc., as a feature of a `code` prop is that Mocha nor its consumers needn't worry about what the constructor was.  They would have to check the `code` anyway, and are unlikely to care whether the constructor was a specialized built-in `Error` subclass.

test/integration/suite.spec.js
```diff
@@ -13,7 +13,8 @@ describe('suite w/no callback', function() {
         if (err) {
           return done(err);
         }
-        var result = res.output.match(/no callback was supplied/) || [];
+        var pattern = /MissingCallbackError/g;
```
boneskull: You can use `expect` here; `assert` is just leftover and should be replaced anyway

lib/errors.js
plroebuck: There is _no_ point in us creating a specialized error class when the existing `TypeError` with `ERR_INVALID_ARG_TYPE'` 'is _exactly_ what should be thrown for missing arguments.

test/integration/file-utils.spec.js
plroebuck: Thanks for the "how to" assertion example.

lib/cli/run.js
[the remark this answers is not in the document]
plroebuck: `InvalidValueError` perhaps...

lib/cli/run-helpers.js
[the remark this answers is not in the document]
plroebuck: My concern was to ensure pattern would be displayed _exactly_ without evaluation. As long as that is addressed, don't care about method.

lib/errors.js
plroebuck: The whole point of using error codes was to stop having to parse the error message. Only by adding additional properties (as well as the code) does i18n/l10n become possible. Not to mention being easier to use programmatically. This style is used heavily by Node itself already for [system errors](https://nodejs.org/api/errors.html#errors_class_systemerror)

lib/suite.js
plroebuck: Even better...
```js
var err = new TypeError(`Suite argument 'title' must be a string. Received type ${typeof title}`);
err.code = 'ERR_INVALID_ARG_TYPE';
err.argument = 'title';
err.expected = 'string';
err.actual = typeof title;

throw err;
```

lib/interfaces/common.js
craigtaub: @plroebuck I will update the names (+ make camelCased), but plan to leave as factories without `new`. Let me know if you feel strongly about using the other way round.

lib/suite.js
craigtaub: I'll add this as a new factory for `suite` + `test`, `createArgumentTypeError`.

test/integration/file-utils.spec.js
craigtaub: For this test its important to check the `argument` property here too so can still use the default
```
expect(dirLookup, 'to throw', {
  code: 'ERR_MOCHA_MISSING_ARG_TYPE',
  argument: 'extensions'
});
```

That new assertion type looks useful though.

lib/cli/run.js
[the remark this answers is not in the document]
craigtaub: Have used `.reporter` and not `.value` property, inside this factory.

lib/errors.js
craigtaub: I thought this only applied if we utilised custom error classes (for MochaError), something we are now not doing (see comment https://github.com/mochajs/mocha/pull/3467#issuecomment-445924038) . Let me know if you have any further thoughts on this.

craigtaub: I thought this only applied if we utilised custom error classes (for MochaError), something we are now not doing (see comment https://github.com/mochajs/mocha/pull/3467#issuecomment-445924038) . Let me know if you have any further thoughts on this.

lib/cli/run.js
```diff
@@ -190,7 +192,7 @@ exports.builder = yargs =>
             const pair = opt.split('=');
 
             if (pair.length > 2 || !pair.length) {
-              throw new Error(`invalid reporter option '${opt}'`);
+              throw createInvalidReporterError('invalid reporter option', opt);
```
plroebuck: Said this last time. This is not an `InvalidReporterError`, but a `TypeError` with code `ERR_INVALID_ARGUMENT_VALUE`. This is CLI parsing code...

lib/errors.js
```diff
@@ -0,0 +1,113 @@
...
+ * @module Errors
+ */
+/**
+ * Factory functions to create throwable error objects
+ */
+
+/**
+ * File/s of test could not be found.
+ *
+ * @public
```
plroebuck: Incomplete documentation on each function...
```js
  * @param {string} message - Error message to be displayed.
  * @param {string} pattern - User-specified argument value.
  * @returns {Error} instance detailing the error condition 
```

lib/errors.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+function createArgumentTypeError(message, argument, expected) {
+  var err = new TypeError(message);
+  err.code = 'ERR_MOCHA_INVALID_ARG_TYPE';
+  err.argument = argument;
+  err.expected = expected;
+  err.actual = typeof argument;
+  throw err;
```
plroebuck:
```js
  return err;
```

lib/errors.js
plroebuck: No further thoughts. There was no comment why at the time.

lib/errors.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Callback for a running suite was not found.
+ *
+ * @public
+ */
+function createMissingCallbackError(message) {
+  var err = createMissingArgumentError(message, 'callback');
+  err.code = 'ERR_MOCHA_MISSING_CALLBACK';
+  return err;
+}
```
plroebuck:
```js
 function createMissingCallbackError(message) {
   return createMissingArgumentError(message, 'callback', 'function');
 }
```

lib/errors.js
```diff
@@ -0,0 +1,113 @@
...
+module.exports = {
+  createNoFilesMatchPatternError: createNoFilesMatchPatternError,
+  createMissingCallbackError: createMissingCallbackError,
+  createInvalidReporterError: createInvalidReporterError,
+  createInvalidInterfaceError: createInvalidInterfaceError,
+  createNotSupportedError: createNotSupportedError,
+  createMissingArgumentError: createMissingArgumentError,
+  createUndefinedError: createUndefinedError,
+  createArgumentTypeError: createArgumentTypeError
+};
```
plroebuck: Alphabetize these.

lib/interfaces/common.js
plroebuck: Well, using `create<xxx>Error` _somewhat_ implies `new`, so naming-wise okay.
**Very** disappointed we couldn't use _actual_ Error objects instead since now the stacks will be a level deeper than the actual problem...

lib/mocha.js
plroebuck: You fixed the first one, but skipped the second.

lib/utils.js
```diff
@@ -546,8 +553,9 @@ exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
       return;
     }
     if (!extensions) {
-      throw new Error(
-        'extensions parameter required when filepath is a directory'
+      throw createMissingArgumentError(
+        'Parameter required when filepath is a directory',
```
plroebuck:
```js
        'Argument "extensions" required when argument "filepath" is a directory',
```

test/integration/suite.spec.js
```diff
@@ -30,7 +30,7 @@ describe('skipped suite w/no callback', function() {
       }
       var pattern = new RegExp('Error', 'g');
```
plroebuck: Shouldn't this be `TypeError`?

test/integration/suite.spec.js
```diff
@@ -44,7 +44,7 @@ describe('skipped suite w/ callback', function() {
...
```
plroebuck: same as above

lib/errors.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Interface specified in options not found.
+ *
+ * @public
+ */
+function createInvalidInterfaceError(message, badInterface) {
+  var err = new Error(message);
+  err.code = 'ERR_MOCHA_INVALID_INTERFACE';
+  err.interface = badInterface;
```
plroebuck:
```js
   err.interface = interface;
```

test/unit/errors.spec.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+var errors = require('../../lib/errors');
+
+describe('Errors', function() {
+  var expectedMessage = 'some message';
+  it('should include expected code in thrown reporter errors', function() {
+    var throwError = function() {
+      throw errors.createInvalidReporterError(expectedMessage);
```
plroebuck: Lacking second argument which additionally should be checked in the expect.

test/unit/errors.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    };
+    expect(throwError, 'to throw', {
+      message: expectedMessage,
+      code: 'ERR_MOCHA_INVALID_REPORTER'
+    });
+  });
+
+  it('should include expected code in thrown interface errors', function() {
+    var throwError = function() {
+      throw errors.createInvalidInterfaceError(expectedMessage);
```
plroebuck: same as previous

.eslintrc.yml
```diff
@@ -22,6 +22,7 @@ overrides:
       - karma.conf.js
       - .wallaby.js
       - bin/*
+      - lib/*
```
plroebuck: Don't think this is correct. You're now making everything ES6 rather than just the Node-specific JS.

.eslintrc.yml
```diff
@@ -36,8 +37,8 @@ overrides:
     globals:
       expect: no
   - files:
-    - bin/*
-    - lib/**/*.js
+      - bin/*
+      - lib/**/*.js
```
plroebuck: Why are these being double indented now?

lib/cli/run.js
craigtaub: Ok will add `createInvalidArgumentTypeError`. Also have been using mocha prefix even for type errors so `ERR_MOCHA_INVALID_ARG_VALUE `.

lib/errors.js
craigtaub: Will update. Do you mean `@public` here? They are exported so my assumption would be `@public`?

.eslintrc.yml
craigtaub: Well all the js at top level of `lib` (i.e. not browser/interfaces/reporters/cli). This was due to lint error using template literals. 
Do you think its safer just to add the 3 files using it here?

.eslintrc.yml
craigtaub: Have removed all lint config changes.

craigtaub:
I'll fix that coverage issue

lib/errors.js
plroebuck: Only exported variables are public. You roundabout export your functions via the `module.exports` object whose properties you assign them to, but those properties are the _real_ `@public`s. But I could see documenting these as such... your call.

docs/index.md
```diff
@@ -1658,6 +1659,20 @@ $ mocha "./spec/**/*.js"
 
 *Note*: Double quotes around the glob are recommended for portability.
 
+## Error Codes
+
+List of codes associated with Errors thrown inside Mocha. Following NodeJS practices.
+
+| Code        | Meaning |
+| ------------- | ------------- |
+| ERR_MOCHA_INVALID_ARG_VALUE | an argument used did not use the supported value |
```
plroebuck: `an invalid or unsupported value was passed for a given argument`

docs/index.md
```diff
@@ -1658,6 +1659,20 @@ $ mocha "./spec/**/*.js"
...
+| ERR_MOCHA_INVALID_ARG_TYPE | an argument used did not use the supported type |
```
plroebuck: `An argument of the wrong type was passed to Mocha's API`

docs/index.md
plroebuck: Alphabetize, pls

lib/cli/run.js
plroebuck: Note that this one would use `createInvalidArgumentValueError`.

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+function createMissingArgumentError(message, argument, expected) {
+  return createInvalidArgumentTypeError(message, argument, expected);
+}
+
+/**
+ * Creates an error object used when an argument did not use the supported type
+ *
+ * @public
+ * @param {string} message - Error message to be displayed.
+ * @param {string} argument - User-specified actual argument type.
```
plroebuck:
```js
  * @param {string} argument - Function argument name.
```

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {string} expected - User-specified expected argument type.
```
plroebuck:
```js
  * @param {string} expected - Expected argument type.
```

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+  err.actual = typeof argument;
+  return err;
+}
+
+/**
+ * Creates an error object used when an argument did not use the supported value
+ *
+ * @public
+ * @param {string} message - Error message to be displayed.
+ * @param {string} argument - User-specified actual argument value.
```
plroebuck:
```js
  * @param {string} argument - Argument name.
```

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+ * @param {string} expected - User-specified expected argument value.
```
plroebuck:
```js
  * @param {string} expected - Expected argument value.
```

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+ * @returns {Error} instance detailing the error condition
+ */
+function createInvalidArgumentValueError(message, argument, expected) {
```
plroebuck:
```js
 function createInvalidArgumentValueError(message, argument, actual, expected) {
```

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+  var err = new TypeError(message);
+  err.code = 'ERR_MOCHA_INVALID_ARG_VALUE';
+  err.argument = argument;
+  err.expected = expected;
+  err.actual = typeof argument;
```
plroebuck:
```js
   err.value = value;
   err.reason = (typeof reason !== 'undefined')
     ? reason
     : 'is invalid';
```

plroebuck: BTW, I'm not just making these up arbitrarily. Mimicking Node [errors](https://github.com/nodejs/node/blob/master/lib/internal/errors.js#L737).

lib/errors.js
```diff
@@ -0,0 +1,138 @@
...
+ */
+function createUndefinedError(message) {
+  var err = new Error(message);
+  err.code = 'ERR_MOCHA_UNDEFINED_ERROR';
+  return err;
+}
+
+module.exports = {
+  createInvalidArgumentValueError: createInvalidArgumentValueError,
+  createInvalidArgumentTypeError: createInvalidArgumentTypeError,
```
plroebuck: Move this one above the first

lib/errors.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Creates an error object used when an argument did not use the supported value
+ *
+ * @public
+ * @param {string} message - Error message to be displayed.
+ * @param {string} argument - Argument name.
+ * @param {string} value - Argument value.
+ * @param {string} reason - Why value is invalid.
```
plroebuck:
```js
  * @param {string} [reason] - Why value is invalid.
```

craigtaub: Why square brackets?

lib/errors.js
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Creates an error object used when no files to be tested could be found using specified pattern.
+ *
+ * @public
+ * @param {string} message - Error message to be displayed.
+ * @param {string} pattern - User-specified argument value.
+ * @returns {Error} instance detailing the error condition
+ */
+function createNoFilesMatchPatternError(message, pattern) {
+  var err = new Error(message);
```
plroebuck: In a later revision (PR), we should generate the messages using additional arguments rather than requiring the caller to do so. But let's get this in place first...

lib/errors.js
plroebuck: optional argument

craigtaub:
@boneskull is there anything outstanding from you? I'd really like to get this in if nothing else.

lib/cli/run-helpers.js
```diff
@@ -143,10 +143,8 @@ exports.handleFiles = ({
     try {
       newFiles = utils.lookupFiles(arg, extension, recursive);
     } catch (err) {
-      if (err.message.indexOf('cannot resolve path') === 0) {
-        console.error(
-          `Warning: Could not find any test files matching pattern: ${arg}`
-        );
+      if (err.code === 'ERR_MOCHA_NO_FILES_MATCH_PATTERN') {
```
boneskull: future PR: use variables instead of hardcoded strings

test/integration/suite.spec.js
```diff
@@ -28,9 +28,9 @@ describe('skipped suite w/no callback', function() {
       if (err) {
         return done(err);
       }
-      var pattern = new RegExp('Error', 'g');
+      var pattern = new RegExp('TypeError', 'g');
       var result = res.output.match(pattern) || [];
-      assert.strictEqual(result.length, 0);
+      expect(result, 'to have length', 0);
```
boneskull: for future reference, this could be `expect(result, 'to be empty')`

lib/errors.js
boneskull: Alright.  I feel it's kind of a "YAGNI" situation, but won't die on that hill.

boneskull:
👍 @craigtaub

lib/cli/run-helpers.js
plroebuck: Regrettably, it's the standard Node error code processing pattern.
[This](https://github.com/cswl/nodejs-error-codes) seems more like what I would have preferred -- exported constants.

craigtaub: Yep, I had it like that on first iteration of PR but the change was lost in the ether. Will look at updating.

Issue #3445, opened 2018-08-01
Silence Logs in debug mode of Karma browser


chrisdel101:
### Description
In the Karma debug console there are lines of useless output. This is traced to the module. I`d like to know how to fix this.

### Steps to Reproduce
Once Karma is running, open debug console in browser and logs appear (see image)

It seems to be this piece in moch.js. If comment out the logs, things work as expected. (line 6959 for me)
```
BrowserStdout.prototype._write = function(chunks, encoding, cb) {
  var output = chunks.toString ? chunks.toString() : chunks
  if (this.label === false) {
    console.log(output)
  } else {
    console.log(this.label+':', output)
  }
  process.nextTick(cb)
}
```

How would I deal with fixing this permanently so I don`t see these logs? I'm not sure how to do this, or if it is possible, or if this is an issue on my end how I can solve it.

![image 1](https://user-images.githubusercontent.com/14151974/43527678-2bc742a8-9575-11e8-80ae-b867c36021b7.png)

plroebuck:
When #3477 is merged, the `stdout:` prefix will disappear from output.
But he is running the `NyanCat` reporter above which spews Terminal escape codes for a living.

@chrisdel101, does your problem improve if you specify `--reporter min` or `--reporter spec`?
Use of `--no-color` option _might_ reduce output further.

chrisdel101:
I added the additional `NyanCat` reporter flag to try and see if it did anything, or it the reporters did anything to the output. They did not, that I could see. 
It was still giving those output logs before I added that though.

 I've commented out the line in in question in mocha.js to fix this for now, and it makes it useable.

I checked with my teammates and we have these logs happening so it's something that needs investigation here. I am closing this as I cannot provide the investigation necessary now (and apologies for not providing enough info about OS etc to begin with). Thanks!

plroebuck:
Full Disclosure: never used Karma

From quick glance, these "karma.conf.js" configuration options might be applicable.
* `browserConsoleLogOptions`
* `client.captureConsole`
* `detached` 

Your "patch" disables all output sent to `stdout` when run in browser.

Good luck and let us know if one of these options resolves your problem (without requiring you to comment out any code).

PR #3477, opened 2018-09-18
Remove stdout: prefix from browser reporter logs

Bamieh:
As noted by @plroebuck `level` is not used, should be `label` instead.

https://github.com/mochajs/mocha/commit/37193a432119be643b9476f97da557a6f60b7d0f#commitcomment-30530038

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -7,7 +7,7 @@
  * Shim process.stdout.
  */
 
-process.stdout = require('browser-stdout')({level: false});
+process.stdout = require('browser-stdout')({label: false});
 
 var Mocha = require('./lib/mocha');
 


plroebuck:
Do Mocha's current browser tests demonstrate the issue?

plroebuck:
Maybe rename the PR as "Remove stdout: prefix from browser reporter logs"?

Bamieh:
@plroebuck i don't think so, not sure how we'd test it though

outsideris:
Now, I doubt #3445 is related this problem. In my test, I didn't found the `stdout:` problem.

plroebuck:
Can now confirm problem exists in Mocha-5.2.0.
Here are my [steps to reproduce it](https://github.com/mochajs/mocha/issues/3521#issuecomment-431554817).

plroebuck:
### Description of the Change

Removes stdout: prefix from reporter output in browser console.

N/A

### Benefits

Less onscreen noise for people who decide to run reporters _other_ than `HTML` reporter.

### Possible Drawbacks

Well, it's been around for a while now. People _may_ have written tests that work around the problem.

### Applicable issues

Fixes #2095
Fixes #3521 
semver-patch

PR #3486, opened 2018-09-25
fix slow test for dot reporter

outsideris:
### Description of the Change
Although there is [a slow test case](https://github.com/mochajs/mocha/blob/master/test/reporters/dot.spec.js#L98-L107) for dot reporter, 
[coverage](https://coveralls.io/builds/19145852/source?filename=lib/reporters/dot.js#L51) is missed.
Since slow duration is 2(`slow: function() { return 2; }`) and test duration is 2 (`test.duration = 2;`), 
this test is not check slow test actually.

### Alternate Designs
N/A

Tests should be run as intended.

### DIFF
--- a/test/reporters/dot.spec.js
+++ b/test/reporters/dot.spec.js
@@ -8,35 +8,60 @@ var createMockRunner = require('./helpers.js').createMockRunner;
 
 describe('Dot reporter', function() {
   var stdout;
-  var stdoutWrite;
   var runner;
   var useColors;
   var windowWidth;
+  var color;
+  var showOutput = false;
+
+  /**
+   * Run reporter using stream reassignment to capture output.
+   *
+   * @param {Object} stubSelf - Reporter-like stub instance
+   * @param {Runner} runner - Mock instance
+   * @param {boolean} [tee=false] - If `true`, echo captured output to screen
+   */
+  function runReporter(stubSelf, runner, tee) {
+    // Reassign stream in order to make a copy of all reporter output
+    var stdoutWrite = process.stdout.write;
+    process.stdout.write = function(string, enc, callback) {
+      stdout.push(string);
+      if (tee) {
+        stdoutWrite.call(process.stdout, string, enc, callback);
+      }
+    };
+
+    // Invoke reporter
+    Dot.call(stubSelf, runner);
+
+    // Revert stream reassignment here so reporter output
+    // can't be corrupted if any test assertions throw
+    process.stdout.write = stdoutWrite;
+  }
 
   beforeEach(function() {
     stdout = [];
-    stdoutWrite = process.stdout.write;
-    process.stdout.write = function(string, enc, callback) {
-      stdout.push(string);
-      stdoutWrite.call(process.stdout, string, enc, callback);
-    };
     useColors = Base.useColors;
     windowWidth = Base.window.width;
+    color = Base.color;
     Base.useColors = false;
     Base.window.width = 0;
+    Base.color = function(type, str) {
+      return type.replace(/ /g, '-') + '_' + str;
+    };
   });
 
   afterEach(function() {
     Base.useColors = useColors;
     Base.window.width = windowWidth;
-    process.stdout.write = stdoutWrite;
+    Base.color = color;
+    runner = undefined;
   });
 
   describe('on start', function() {
-    it('should return a new line', function() {
+    it('should write a newline', function() {
       runner = createMockRunner('start', 'start');
-      Dot.call({epilogue: function() {}}, runner);
-      process.stdout.write = stdoutWrite;
+      runReporter({epilogue: function() {}}, runner, showOutput);
       var expectedArray = ['\n'];
       expect(stdout, 'to equal', expectedArray);
     });
@@ -46,20 +71,18 @@ describe('Dot reporter', function() {
       beforeEach(function() {
         Base.window.width = 2;
       });
-      it('should return a new line and then a coma', function() {
+      it('should write a newline followed by a comma', function() {
         runner = createMockRunner('pending', 'pending');
-        Dot.call({epilogue: function() {}}, runner);
-        process.stdout.write = stdoutWrite;
-        var expectedArray = ['\n  ', Base.symbols.comma];
+        runReporter({epilogue: function() {}}, runner, showOutput);
+        var expectedArray = ['\n  ', 'pending_' + Base.symbols.comma];
         expect(stdout, 'to equal', expectedArray);
       });
     });
     describe('if window width is equal to or less than 1', function() {
-      it('should return a coma', function() {
+      it('should write a comma', function() {
         runner = createMockRunner('pending', 'pending');
-        Dot.call({epilogue: function() {}}, runner);
-        process.stdout.write = stdoutWrite;
-        var expectedArray = [Base.symbols.comma];
+        runReporter({epilogue: function() {}}, runner, showOutput);
+        var expectedArray = ['pending_' + Base.symbols.comma];
         expect(stdout, 'to equal', expectedArray);
       });
     });
@@ -76,32 +99,42 @@ describe('Dot reporter', function() {
         Base.window.width = 2;
       });
       describe('if test speed is fast', function() {
-        it('should return a new line and then a dot', function() {
+        it('should write a newline followed by a dot', function() {
           runner = createMockRunner('pass', 'pass', null, null, test);
-          Dot.call({epilogue: function() {}}, runner);
-          process.stdout.write = stdoutWrite;
-          var expectedArray = ['\n  ', Base.symbols.dot];
+          runReporter({epilogue: function() {}}, runner, showOutput);
+          expect(test.speed, 'to equal', 'fast');
+          var expectedArray = ['\n  ', 'fast_' + Base.symbols.dot];
           expect(stdout, 'to equal', expectedArray);
         });
       });
     });
     describe('if window width is equal to or less than 1', function() {
       describe('if test speed is fast', function() {
-        it('should return a dot', function() {
+        it('should write a grey dot', function() {
           runner = createMockRunner('pass', 'pass', null, null, test);
-          Dot.call({epilogue: function() {}}, runner);
-          process.stdout.write = stdoutWrite;
-          var expectedArray = [Base.symbols.dot];
+          runReporter({epilogue: function() {}}, runner, showOutput);
+          expect(test.speed, 'to equal', 'fast');
+          var expectedArray = ['fast_' + Base.symbols.dot];
+          expect(stdout, 'to equal', expectedArray);
+        });
+      });
+      describe('if test speed is medium', function() {
+        it('should write a yellow dot', function() {
+          test.duration = 2;
+          runner = createMockRunner('pass', 'pass', null, null, test);
+          runReporter({epilogue: function() {}}, runner, showOutput);
+          expect(test.speed, 'to equal', 'medium');
+          var expectedArray = ['medium_' + Base.symbols.dot];
           expect(stdout, 'to equal', expectedArray);
         });
       });
       describe('if test speed is slow', function() {
-        it('should return a dot', function() {
-          test.duration = 2;
+        it('should write a bright yellow dot', function() {
+          test.duration = 3;
           runner = createMockRunner('pass', 'pass', null, null, test);
-          Dot.call({epilogue: function() {}}, runner);
-          process.stdout.write = stdoutWrite;
-          var expectedArray = [Base.symbols.dot];
+          runReporter({epilogue: function() {}}, runner, showOutput);
+          expect(test.speed, 'to equal', 'slow');
+          var expectedArray = ['bright-yellow_' + Base.symbols.dot];
           expect(stdout, 'to equal', expectedArray);
         });
       });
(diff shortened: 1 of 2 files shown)


outsideris:
@plroebuck  It [detected "slow"](https://coveralls.io/builds/19172171/source?filename=lib/reporters/dot.js#L52) because duration is 3 and slow is 2.
And it doesn't print `bright yellow` since [this if statement](https://github.com/mochajs/mocha/blob/d21f8681dce2045d81725556ee640f9bcc1829a1/lib/reporters/base.js#L111).

plroebuck:
These tests seem hinky to begin with. Why was the [test duration](https://github.com/mochajs/mocha/blob/d21f8681dce2045d81725556ee640f9bcc1829a1/test/reporters/dot.spec.js#L69) ever allowed to be `1`? Doubt any of our tests can execute that fast. But after you turn off colors, all passing speeds return a dot. So asserting for a dot can't prove test speed.

* Should your change apply to all "on pass" tests (L69) and the line you fixed (L100) be removed instead?
* Should colors be reinstated so test speed can actually be checked?

outsideris:
That's the reason that we didn't find this issue even if this slow test was meaningless.
I got what you point out.

plroebuck:
Although I applaud your effort in having tracked this down, I see no point in merging this PR in the name of increased coverage - without meaningful changes, it only helps further mask the underlying problems.

outsideris:
* I turned on `useColors` since comparing dots without colors is meanless for dot reporter.
* All tests for dot reporter now assert result with color.

Also, I added a `function color()` for test purpose. At first, I used `Base.color` in `base.js`, but all test still passed if I turn on `useColors` or not. I want to force to check dot reporter print colors properly.

test/reporters/dot.spec.js
```diff
@@ -13,6 +13,10 @@ describe('Dot reporter', function() {
   var useColors;
   var windowWidth;
 
+  function color(type, str) {
+    return '\u001b[' + Base.colors[type] + 'm' + str + '\u001b[0m';
+  }
+
```
plroebuck: For simplicity of comparison, it might be easier _not_ to be so literal.
```js
function speedString(type, str) {
  return '##' + type + '-' + str;
};

Base.color = speedString;
```

outsideris:
I replaced `Base.color`.

plroebuck:
Come to think of it, given:
```js
if (test.duration > test.slow()) {
  test.speed = 'slow';
} else if (test.duration > test.slow() / 2) {
  test.speed = 'medium';
} else {
  test.speed = 'fast';
}
```
we could do this assertion for passing tests.
```js
expect(test.speed, 'to equal', 'fast');
```

outsideris:
However, they are not only for speed test. We should check report type, dot in here.

test/reporters/dot.spec.js
```diff
@@ -80,28 +86,38 @@ describe('Dot reporter', function() {
           runner = createMockRunner('pass', 'pass', null, null, test);
           Dot.call({epilogue: function() {}}, runner);
           process.stdout.write = stdoutWrite;
-          var expectedArray = ['\n  ', Base.symbols.dot];
+          var expectedArray = ['\n  ', 'fast_' + Base.symbols.dot];
```
plroebuck:
```js
          expect(test.speed, 'to equal', 'fast');
```

test/reporters/dot.spec.js
```diff
@@ -80,28 +86,38 @@ describe('Dot reporter', function() {
...
         });
       });
     });
     describe('if window width is equal to or less than 1', function() {
       describe('if test speed is fast', function() {
-        it('should return a dot', function() {
+        it('should return a grey dot', function() {
           runner = createMockRunner('pass', 'pass', null, null, test);
           Dot.call({epilogue: function() {}}, runner);
           process.stdout.write = stdoutWrite;
```
plroebuck:
```js
          expect(test.speed, 'to equal', 'fast');
```

test/reporters/dot.spec.js
```diff
@@ -80,28 +86,38 @@ describe('Dot reporter', function() {
...
```
plroebuck:
```js
          expect(test.speed, 'to equal', 'fast');
```

test/reporters/dot.spec.js
```diff
@@ -80,28 +86,38 @@ describe('Dot reporter', function() {
...
         });
       });
-      describe('if test speed is slow', function() {
-        it('should return a dot', function() {
+      describe('if test speed is medium', function() {
+        it('should return a yellow dot', function() {
           test.duration = 2;
           runner = createMockRunner('pass', 'pass', null, null, test);
           Dot.call({epilogue: function() {}}, runner);
           process.stdout.write = stdoutWrite;
```
plroebuck:
```js
          expect(test.speed, 'to equal', 'medium');
```

test/reporters/dot.spec.js
```diff
@@ -80,28 +86,38 @@ describe('Dot reporter', function() {
...
+          var expectedArray = ['medium_' + Base.symbols.dot];
+          expect(stdout, 'to equal', expectedArray);
+        });
+      });
+      describe('if test speed is slow', function() {
+        it('should return a bright yellow dot', function() {
+          test.duration = 3;
+          runner = createMockRunner('pass', 'pass', null, null, test);
+          Dot.call({epilogue: function() {}}, runner);
+          process.stdout.write = stdoutWrite;
```
plroebuck:
```js
          expect(test.speed, 'to equal', 'slow');
```

test/reporters/dot.spec.js
```diff
@@ -50,7 +56,7 @@ describe('Dot reporter', function() {
         runner = createMockRunner('pending', 'pending');
         Dot.call({epilogue: function() {}}, runner);
         process.stdout.write = stdoutWrite;
```
plroebuck: Aren't we doing stream reassignment in `afterEach()` already? Why repeat here?

test/reporters/dot.spec.js
outsideris: I don't understand what you mean and why we check only speed.

test/reporters/dot.spec.js
outsideris: I will test what you provide.

test/reporters/dot.spec.js
```diff
@@ -22,13 +23,18 @@ describe('Dot reporter', function() {
...
+    Base.color = function(type, str) {
+      return type.replace(/ /g, '-') + '_' + str;
+    };
   });
 
   afterEach(function() {
     Base.useColors = useColors;
     Base.window.width = windowWidth;
+    Base.color = color;
     process.stdout.write = stdoutWrite;
```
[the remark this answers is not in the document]
outsideris: Did you mean that reset `runner` here before every test?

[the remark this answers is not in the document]
plroebuck: yes. runner recreated per-test. reset ensures you can't forget to do so.

test/reporters/dot.spec.js
plroebuck: the field is set for `pass` events by `runner` in `Base` ctor; it's the only field set there.
`test.speed` assertions are _in addition to_ your `stdout` assertions, not in lieu of them.

outsideris:
I fixed them.

test/reporters/dot.spec.js
```diff
@@ -8,35 +8,50 @@ var createMockRunner = require('./helpers.js').createMockRunner;
...
-  var stdoutWrite;
   var runner;
   var useColors;
   var windowWidth;
+  var color;
 
-  beforeEach(function() {
-    stdout = [];
-    stdoutWrite = process.stdout.write;
+  function runReporter(stubSelf, runner) {
```
plroebuck:
```js
  function runReporter(stubSelf, runner, tee) {
```

test/reporters/dot.spec.js
```diff
@@ -8,35 +8,50 @@ var createMockRunner = require('./helpers.js').createMockRunner;
...
+    // Reassign stream in order to make a copy of all reporter output
+    var stdoutWrite = process.stdout.write;
     process.stdout.write = function(string, enc, callback) {
       stdout.push(string);
       stdoutWrite.call(process.stdout, string, enc, callback);
```
plroebuck:
```js
    if (tee) {
    }
```

test/reporters/dot.spec.js
```diff
@@ -8,35 +8,50 @@ var createMockRunner = require('./helpers.js').createMockRunner;
 
 describe('Dot reporter', function() {
   var stdout;
```
plroebuck:
```js
  var showOutput = false;
```

test/reporters/dot.spec.js
```diff
@@ -8,35 +8,50 @@ var createMockRunner = require('./helpers.js').createMockRunner;
...
+    Base.color = color;
+    runner = undefined;
   });
 
   describe('on start', function() {
     it('should return a new line', function() {
       runner = createMockRunner('start', 'start');
-      Dot.call({epilogue: function() {}}, runner);
-      process.stdout.write = stdoutWrite;
+      runReporter({epilogue: function() {}}, runner);
```
plroebuck: Essentially, do this for _all_ `runReporter` invocations.

```js
     runReporter({epilogue: function() {}}, runner, showOutput);
```

plroebuck:
It looked good. But glancing job results on Travis, wondered why we would be displaying the test output (which intermixes with the job "spec" report).

So my final requested modification:
* Add argument `tee` to `runReport` function
* Add a `describe()`-level global `showOutput` which defaults to `false`
* Make all invocations of `runReport` add `showOutput` as final parameter.

Now, debugging a test is still simple, but output is turned off by default.

outsideris:
> Now, debugging a test is still simple, but output is turned off by default.

It is a good idea.
We can apply this idea to other reporter tests. Modification for other reporters is in another PR is better.

test/reporters/dot.spec.js
```diff
@@ -119,18 +152,16 @@ describe('Dot reporter', function() {
       });
       it('should return a new line and then an exclamation mark', function() {
```
plroebuck: "should return a newline followed by an exclamation mark"

test/reporters/dot.spec.js
```diff
@@ -78,30 +101,40 @@ describe('Dot reporter', function() {
...
+          expect(test.speed, 'to equal', 'fast');
+          var expectedArray = ['\n  ', 'fast_' + Base.symbols.dot];
           expect(stdout, 'to equal', expectedArray);
         });
       });
     });
     describe('if window width is equal to or less than 1', function() {
       describe('if test speed is fast', function() {
-        it('should return a dot', function() {
+        it('should return a grey dot', function() {
```
plroebuck: "should write a grey dot"

test/reporters/dot.spec.js
```diff
@@ -78,30 +101,40 @@ describe('Dot reporter', function() {
...
+          runReporter({epilogue: function() {}}, runner, showOutput);
+          expect(test.speed, 'to equal', 'fast');
+          var expectedArray = ['fast_' + Base.symbols.dot];
           expect(stdout, 'to equal', expectedArray);
         });
       });
-      describe('if test speed is slow', function() {
-        it('should return a dot', function() {
+      describe('if test speed is medium', function() {
+        it('should return a yellow dot', function() {
```
plroebuck: "should write a yellow dot"

test/reporters/dot.spec.js
```diff
@@ -78,30 +101,40 @@ describe('Dot reporter', function() {
...
-          process.stdout.write = stdoutWrite;
-          var expectedArray = [Base.symbols.dot];
+          runReporter({epilogue: function() {}}, runner, showOutput);
+          expect(test.speed, 'to equal', 'medium');
+          var expectedArray = ['medium_' + Base.symbols.dot];
+          expect(stdout, 'to equal', expectedArray);
+        });
+      });
+      describe('if test speed is slow', function() {
+        it('should return a bright yellow dot', function() {
```
plroebuck: "should write a bright yellow dot"

test/reporters/dot.spec.js
```diff
@@ -119,18 +152,16 @@ describe('Dot reporter', function() {
...
-        Dot.call({epilogue: function() {}}, runner);
-        process.stdout.write = stdoutWrite;
-        var expectedArray = ['\n  ', Base.symbols.bang];
+        runReporter({epilogue: function() {}}, runner, showOutput);
+        var expectedArray = ['\n  ', 'fail_' + Base.symbols.bang];
         expect(stdout, 'to equal', expectedArray);
       });
     });
     describe('if window width is equal to or less than 1', function() {
       it('should return an exclamation mark', function() {
```
plroebuck: "should write an exclamation mark"

outsideris:
I updated the test names.

plroebuck:
On the other hand, I am starting to rethink the design of some of the code in the `Base` reporter. Think many of these standalone exports (e.g., `Base.color()`) should have been subclass instance properties/methods rather than module exports. Assuming a goal of being able to have multiple reporters for a Mocha execution, this would need change to prevent one reporter from (even inadvertently) affecting any others. For example, `Dot.colors.slow = 'bright yellow';` in its ctor would have eliminated the `if` condition within the onPassCB event handler.

outsideris:
>  For example, Dot.colors.slow = 'bright yellow'; in its ctor would have eliminated the if condition within the onPassCB event handler.

what `ctor` means?

outsideris:
I will merge this if one more core members approved.

plroebuck:
@outsideris [ctor](https://en.wikipedia.org/wiki/Constructor_(object-oriented_programming))

outsideris:
I see. Thank you.

Issue #3480, opened 2018-09-18
🚀 Feature: Make reporter options standard


plroebuck:
### Description

1. Make `reporterOptions` a standard object within `Mocha.options`.
1. Reparent the following `Mocha.options` as `reporterOptions` properties:
    * `useColors`
    * `useInlineDiffs`
    * `hideDiff`
1. Update **all** reporter constructors to accept second argument `options`.
1. Update `Mocha.run()` to pass `options.reporterOptions` to reporter constructor. Currently passes `options`, but most Mocha option properties are not applicable/unused by reporter.
   * Update `XUnit` reporter constructor to expect `options.reporterOptions`
1. Migrate `Base` [reporter settings](https://github.com/mochajs/mocha/blob/d401ee198bd5026bc61641ad26eab31e6746713f/lib/mocha.js#L598-L602) portion of `Mocha.run()` to `Base` reporter constructor (or somewhere more related).

plroebuck:
Perhaps simpler to have a `--settings` (JSON?) configuration file used with Mocha ctor. It would be applied prior to "mocha.opts" processing (so cmdline still overrides).

An extension of the settings file would turn `options.reporterOptions` into an ever-present part of `options`. Each specific reporter would have its own options object, referenced by its name.
Prior to reporter instance creation, the options object passed to `Base` ctor would be a meld of shared defaults, reporter defaults from settings file, and cmdline overrides.

```js

settings = {
  reporterOptions: {
    _common: {
      //useColors
      //useInlineDiffs
      //hideDiff
    },
    spec: {
      stdout: '-'
    },
    // other reporters
  },
  timeout: 75,
  // etc.
}
```

johanblumenberg:
I think it is a good idea to add a structure that allows to specify options to individual reporters, and allow for multiple reporters.

It might also be a good idea to only pass the specific reporters options to a reporter when constructing it. Maybe even take it one step further and only pass `options.reporterOptions.xunit` to the `xunit` constructor. There is no need to pass options for all reporters to all reporters.
A drawback could be that you don't get the mocha options. You might want to output something like `\`Test failed due to timeout. The test took ${time} ms, and the timeout is ${options.timeout}\``. But maybe you can get those things, and others, from the runner object passed to the constructor.

I don't think it is a good idea to have a set of common options to all reporters. For example `useColor` doesn't make sense to `xunit`, because you are typically writing xml to a file for some CI system to read.

plroebuck:
[...]

PR #3487, opened 2018-10-04
Support for nested options in reporterOptions

johanblumenberg:
### Description of the Change

Supporting nested reporter options to be passed on the command line. Example:

```bash
mocha --reporter mocha-multi --reporter-options xunit.stdout=-,xunit.options.suiteName=MySuiteName
```

This is useful when using the mocha-multi reporter, to be able to specify multiple reporters and pass options to each of them.

### Alternate Designs

No other designs considered

Since mocha owns the reporter-options argument, it also has to own the parsing of it.

### Benefits

It will be possible to pass options to reporters, when using the mocha-multi reporter.

### Possible Drawbacks

None that I can think of

### Applicable issues

This should be a minor release

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -149,6 +149,12 @@ program
   .option('-C, --no-colors', 'force disabling of colors')
   .option('-G, --growl', 'enable growl notification support')
   .option('-O, --reporter-options <k=v,k2=v2,...>', 'reporter-specific options')
+  .option(
+    '-o, --reporter-option <k=v>',
+    'reporter-specific option',
+    collect,
+    []
+  )
   .option('-R, --reporter <name>', 'specify the reporter to use', 'spec')
   .option('-S, --sort', 'sort test files')
   .option('-b, --bail', 'bail after first test failure')
@@ -347,8 +353,9 @@ function addReporterOption(object, name, value) {
   }
 }
 
-if (program.reporterOptions !== undefined) {
-  program.reporterOptions.split(',').forEach(opt => {
+program.reporterOption
+  .concat(program.reporterOptions ? program.reporterOptions.split(',') : [])
+  .forEach(opt => {
     const L = opt.split('=');
     if (L.length > 2 || L.length === 0) {
       throw new Error(`invalid reporter option '${opt}'`);
@@ -358,7 +365,6 @@ if (program.reporterOptions !== undefined) {
       addReporterOption(reporterOptions, L[0].split('.'), true);
     }
   });
-}
 
 // reporter
 
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -335,15 +335,27 @@ Error.stackTraceLimit = Infinity; // TODO: config
 // reporter options
 
 const reporterOptions = {};
+
+function addReporterOption(object, name, value) {
+  if (name.length === 1) {
+    object[name[0]] = value;
+  } else {
+    if (!object[name[0]]) {
+      object[name[0]] = {};
+    }
+    addReporterOption(object[name[0]], name.slice(1), value);
+  }
+}
+
 if (program.reporterOptions !== undefined) {
   program.reporterOptions.split(',').forEach(opt => {
     const L = opt.split('=');
     if (L.length > 2 || L.length === 0) {
       throw new Error(`invalid reporter option '${opt}'`);
     } else if (L.length === 2) {
-      reporterOptions[L[0]] = L[1];
+      addReporterOption(reporterOptions, L[0].split('.'), L[1]);
     } else {
-      reporterOptions[L[0]] = true;
+      addReporterOption(reporterOptions, L[0].split('.'), true);
     }
   });
 }
--- a/test/integration/fixtures/options-reporter.js
+++ b/test/integration/fixtures/options-reporter.js
@@ -1,0 +1,10 @@
+'use strict';
+
+var baseReporter = require('../../../lib/reporters/base');
+module.exports = optionsreporter;
+
+function optionsreporter(runner, options) {
+  baseReporter.call(this, runner, options);
+
+  console.log(JSON.stringify(options.reporterOptions));
+}
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -540,6 +540,48 @@ describe('options', function() {
         done
       );
     });
+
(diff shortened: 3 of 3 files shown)


plroebuck:
Disagree with this approach, at least the aspect of passing/parsing an increasingly complex cmdline argument. Feel it kinda flies in the face of the project's description, at least as far as "simple" goes.
A slightly modified version of your idea might change `--reporter-options` into a repeatable argument, which would be simpler both logically and logistically.

But I have been researching approaches for essentially having implicit "mocha-multi" capability as part of #3480.

johanblumenberg:
@plroebuck The idea that you can only pass a single `--reporter-options` argument, and that it is a list of values separated by comma, is already a part of `mocha`, so that can't be changed easily.

But I could add another `--reporter-option` repeatable argument:

```bash
mocha --reporter mocha-multi --reporter-option xunit.stdout=- --reporter-option xunit.options.suiteName=MySuiteName
```

Is that what you mean?

johanblumenberg:
@plroebuck My example usage becomes a bit more complex because `mocha` only supports a single reporter, and the only way to have multiple reporters is to use another library such as `mocha-multi`.

Having `mocha` support multiple reporters would remove one layer of objects. So that would look something like this, since it removes the `mocha-multi` layer of config:

```bash
mocha --reporter spec,xunit --reporter-option xunit.output=- --reporter-option xunit.suiteName=MySuiteName
```

This would match the settings structure you propose in https://github.com/mochajs/mocha/issues/3480, but it would allow the same structure to be passed from cmdline, you don't have to write it to a file.

plroebuck:
@johanblumenberg  wrote:
> The idea that you can only pass a single `--reporter-options` argument, and that it is a list of 
> values separated by comma, is already a part of `mocha`, so that can't be changed easily.

You're right. No idea why they implemented it that way instead of using [collect](https://github.com/mochajs/mocha/blob/04469bf0e3a008be9d066835fad278a542d8232a/bin/_mocha#L85).
But having multiple `--reporter-option` options doesn't appear to be an incompatible change.

> But I could add another `--reporter-option` repeatable argument:
> 
> ```shell
> mocha --reporter mocha-multi --reporter-option xunit.stdout=- --reporter-option xunit.options.suiteName=MySuiteName
> ```
> 
> Is that what you mean?

Basically. Thought here was that it would be simpler (assuming use of "mocha.opts") to enable/disable a single option (using comment) than to have to edit a list.

plroebuck:
@johanblumenberg wrote:
> My example usage becomes a bit more complex because `mocha` only supports a single
> reporter, and the only way to have multiple reporters is to use another library such as
> `mocha-multi`.

Working on changing that, but not simple due to existing code, etc.

> Having `mocha` support multiple reporters would remove one layer of objects.
> So that would look something like this, since it removes the `mocha-multi` layer of config:
> 
> ```shell
> mocha --reporter spec,xunit --reporter-option xunit.output=- --reporter-option xunit.suiteName=MySuiteName
> ```
> 
> This would match the settings structure you propose in #3480, but it would allow
> the same structure to be passed from cmdline, you don't have to write it to a file.

Pretty close. Current thought was this:
```bash
$ mocha --reporter spec \
           --reporter-option spec.output=- \
        --reporter xunit \
           --reporter-option xunit.output="xunit.xml" \
           --reporter-option xunit.suiteName=MySuiteName
```

Details:
* How best to specify which reporter owns terminal (there can be only one)
* Passing generic options (e.g., `suiteName`) used by all reporters
* Filename templates?

plroebuck:
@johanblumenberg wrote:
> > How best to specify which reporter owns terminal (there can be only one)
> 
> I don't think there is anything that prevents you from having multiple reporters writing to stdout. 
> For example, I'm using the spec reporter to report test progress, and the istanbul reporter to 
> report code coverage. Both go to stdout. I don't think there should be anything that prevents you 
> from doing that. If you configure both `xunit` and `spec` to report to stout, then that is probably 
> what you want, although it doesn't make any sense.

OK. So maybe there _is_ a use case for more than one reporter writing to stdout.
Was attempting to head off bug reports due to "stupid" configurations.

> > Passing generic options (e.g., suiteName) used by all reporters
> 
> I think that the `suiteName` parameter to `xunit` is just bad design of the `xunit` reporter.
> The test suite name is a property of the test cases, it's not a property to be decided by the 
> reporter. For example, it could be the top `describe()` name, or it could be passed as a property 
> to `mocha` itself, and not to a reporter.

Root suite in Mocha was originally identified by its lack of title.
#2676 allowed overriding the default "Mocha Tests" root suite title in `XUnit`.

> > Filename templates?
> 
> Wdym?

In case you maybe wanted something like a timestamp as part of the filename?

johanblumenberg:
I added another commit that adds support for `--reporter-option`.

Now it's possible to do, which is a bit simpler syntax:

```bash
mocha --reporter mocha-multi --reporter-option xunit.stdout=- --reporter-option xunit.options.suiteName=MySuiteName
```

johanblumenberg:
This PR is adding support for `--reporter-option` and the possibility to give nested options. Let's not mix it with the task of supporting multiple reporters.

I can try to come up with a separate PR that implements support for multiple reporters.

plroebuck:
Hesitant to _quickly_ move forward on this PR.

Speed of implementation is not an issue because:
1. "mocha-multi" reporter has an existing mechanism to deal with this (i.e., its "config.json" file). 
1. Fits the "would be nice if" category.
1. Little to no clamoring for this feature.

### Possible problems (even in current release):
#### `--reporter-options` issue:
* Deal with value containing commas:
   1. Set "mocha-multi" reporter's `reporterEnabled` property to `"spec,xunit,tap"`

#### `--reporter-option` issue:
* Extend an existing property that isn't an object:
   1. Set `foo` to `"bar"`
   1. Set `foo.bar` to `1`
* Disable an existing boolean property:
   1. Set `foo` to `false`, not `"false"`

craigtaub:
Any movement on this? 
Seems like a breaking change now that --reporter-options` is an alias for `--reporter-option`.

plroebuck:
@craigtaub, it's something we will need in future, but aren't ready for yet. Issues listed above would need be resolved via discussion.

juergba:
Closing due to inactivity, dates back to ancient times before _yargs_ / _yargs-parser_.

PR #3488, opened 2018-10-04
Ensure all after hooks are executed, even if one of them fail

johanblumenberg:
Reopening this PR: https://github.com/mochajs/mocha/pull/3281
See the details and discussion in the referenced PR.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -313,13 +313,14 @@ Runnable.prototype.run = function(fn) {
     self.clearTimeout();
     self.duration = new Date() - start;
     finished = true;
+    delete self.callback;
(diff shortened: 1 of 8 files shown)


johanblumenberg:
The first commit in this PR also fixes the issue pointed out by this comment: https://github.com/mochajs/mocha/blob/6535965e8655a66de54fc0ad9465c2eb825f13f8/lib/runner.js#L841

plroebuck:
We're not going back to `assert` -- use `unexpected` in your tests.
Won't recommend for review until merge conflicts are fixed, just so we're not wasting anyone's time.

PR #3515, opened 2018-10-14
docs(lib/mocha.js): Update API documentation

plroebuck:
### Description of the Change

Replaced all method headers with more complete documentation, better following recommended styles. Eliminated `@api` tags, as `@access` _should_ imply same. Used `@chainable` to denote class methods returning `this` instance.

Renamed a handful of variables, but **no logic changes**!

### Alternate Designs

N/A

N/A

### Benefits

API documentation will be more useful if more complete.

### Possible Drawbacks

N/A

### Applicable issues

semver-patch

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -8,7 +8,7 @@
 
 var escapeRe = require('escape-string-regexp');
 var path = require('path');
-var reporters = require('./reporters');
+var builtinReporters = require('./reporters');
 var utils = require('./utils');
 
 exports = module.exports = Mocha;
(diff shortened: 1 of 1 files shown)


plroebuck:
No idea how _adding_ documentation could _decrease_ code coverage...

\*\*facepalm\*\*

lib/mocha.js
```diff
@@ -110,11 +121,13 @@ function Mocha(options) {
...
 
 /**
- * Enable or disable bailing on the first failure.
+ * Enables or disables bailing on the first failure.
  *
  * @public
- * @api public
- * @param {boolean} [bail]
+ * @see {@link https://mochajs.org/#-b---bail|CLI option}
+ * @param {boolean} [bail=true] - Whether to bail on first error.
```
Munter: Is `bail` really true by default? This doesn't match the default CLI behavior

plroebuck: It doesn't?
```js
Mocha.prototype.bail = function(bail) {
  if (!arguments.length) {
    bail = true;
  }
  this.suite.bail(bail);
  return this;
};
```

plroebuck: Bear in mind, I was documenting _intent_, not buggy behavior. And yes, there are plenty I found while documenting just this file.

PR #3518, opened 2018-10-15
make isPromise support null as a parameter

fabiosantoscode:
### Description of the Change

isPromise would crash if given `null`, so I made it check for null too.

---
This is an enhancement.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -688,7 +688,7 @@ exports.stackTraceFilter = function() {
  * @returns {boolean} Whether or not `value` is a Promise
  */
 exports.isPromise = function isPromise(value) {
-  return typeof value === 'object' && typeof value.then === 'function';
+  return typeof value === 'object' && value && typeof value.then === 'function';
 };
 
 /**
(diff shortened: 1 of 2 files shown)


plroebuck:
Not saying won't accept, but don't see how this could have caused a problem for Mocha.
Only usage I find is [here](https://github.com/mochajs/mocha/blob/01171b1e7acec133363b50d5baacc9719af6bf1d/lib/runnable.js#L399-L421), which was already guarded in the `if` condition.

boneskull:
@fabiosantoscode 

Thanks for the PR.  This would need the CLA to be signed as well a corresponding unit test.

boneskull:
@plroebuck 	This is the problem with public APIs. 😝

fabiosantoscode:
Good thing that I added a test, if the value was null, the function would return null!

fabiosantoscode:
Signed CLA, tested and fixed :) and yes, I got this crash because of a public API.

boneskull:
LGTM, thanks

PR #3522, opened 2018-10-18
Provide ASCII fallbacks for text-only displays

Alhadis:
### Description of the change

When running in a text-only environment (e.g., [Linux framebuffer](https://en.wikipedia.org/wiki/Linux_framebuffer), `wscons`, boot consoles, etc), Mocha's reporters emit unrecognisable symbols because fonts are rendered using single-byte bitmaps ([here's the font](https://github.com/illumos/illumos-gate/blob/898c3fec4c1cbcd45754042f6635b76a9067abd6/usr/src/uts/common/font/12x22.c) used in the examples below).

I've addressed this by providing an ASCII fallback for environments incapable of displaying multibyte sequences in an intelligible fashion. I also upgraded `supports-color` to add `vt220` emulation support (see [`chalk/supports-color#83`](https://github.com/chalk/supports-color/pull/83)).

#### Before

<details>

`spec` reporter:

`dot` reporter:

`landing` reporter:

`progress` reporter:

</details>

#### After

<details>

`spec` reporter:

`dot` reporter:

`landing` reporter:

The `progress` reporter looks better, but it's still broken on my screen because the [`\e[0G`](https://github.com/mochajs/mocha/blob/01171b1e7acec133363b50d5baacc9719af6bf1d/lib/reporters/base.js#L149) sequence isn't supported by all terminals. I left it because it's kind of separate issue: to move the cursor to the beginning of a line, a simple `\r` should do. :-)
</details>

### Alternate Designs

I considered using `×` (U+00D7) instead of `✖` (U+2716), but didn't because extended ASCII still makes assumptions about an environment's character set support.

I chose `Y` as a fallback for `✓`, but `OK` may have been a better choice – particularly since errors aren't flagged with `N` in the `spec` reporter: 

~~~
  TTYRenderer
    Basic procedures
      Y prints text with track-kerning
      Y draws horizontal and vertical lines
    Manpage formatting
      Y formats groff_char(7) correctly
      Y formats perlre(1) correctly
      1) can be used as a command-line postprocessor
~~~

I stuck with `Y` because changing the length of `Base.symbols.ok` might screw with third-party reporters which expect to always be a single character.

Test results should be intelligible without Unicode support. Even on systems which natively support UTF8, character-based displays will be using bitmap fonts with single-byte encodings.

### Benefits

Clearer output for users running `mocha` in a non-graphical environment.

### Possible drawbacks

This relies on sniffing environment variables, so if a user's tampered with `$TERM` and `$DISPLAY`, they might see plainer-looking symbols when running tests.

Even in the worst-case scenario though, reporters will still emit readable feedback, so I can't see this being much of an issue.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -29,7 +29,8 @@ module.exports = config => {
     browserify: {
       debug: true,
       configure: function configure(b) {
-        b.ignore('glob')
+        b
+          .ignore('glob')
           .ignore('fs')
           .ignore('path')
           .ignore('supports-color')
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -29,8 +29,7 @@ module.exports = config => {
     browserify: {
       debug: true,
       configure: function configure(b) {
-        b
-          .ignore('glob')
+        b.ignore('glob')
           .ignore('fs')
           .ignore('path')
           .ignore('supports-color')
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -46,6 +46,19 @@ exports.useColors =
   (supportsColor.stdout || process.env.MOCHA_COLORS !== undefined);
 
 /**
+ * Use 7-bit ASCII approximations instead of Unicode
+ * symbols. Disabled if env isn't a hardware console.
+ */
+
+exports.forceASCII =
+  !process.browser &&
+  process.platform !== 'win32' &&
+  process.env &&
+  !process.env.CI &&
+  !process.env.DISPLAY &&
+  /^(?:dumb|console|con80x25|linux.*|vt\d{2,3})$/i.test(process.env.TERM);
+
+/**
  * Inline diffs instead of +/-
  */
 
@@ -93,6 +106,10 @@ exports.symbols = {
 if (process.platform === 'win32') {
   exports.symbols.ok = '\u221A';
   exports.symbols.err = '\u00D7';
+  exports.symbols.dot = '.';
+} else if (exports.forceASCII) {
+  exports.symbols.ok = 'Y';
+  exports.symbols.err = 'N';
   exports.symbols.dot = '.';
 }
 
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -46,11 +46,10 @@ exports.useColors =
   (supportsColor.stdout || process.env.MOCHA_COLORS !== undefined);
 
 /**
- * Use 7-bit ASCII approximations instead of Unicode
- * symbols. Disabled if env isn't a hardware console.
+ * Indicate to reporters when Unicode symbols should be avoided.
  */
 
-exports.forceASCII =
+exports.asciiOnly =
   !process.browser &&
   process.platform !== 'win32' &&
   process.env &&
@@ -107,7 +106,9 @@ if (process.platform === 'win32') {
   exports.symbols.ok = '\u221A';
   exports.symbols.err = '\u00D7';
   exports.symbols.dot = '.';
-} else if (exports.forceASCII) {
+}
+// With 7-bit displays: use symbols guaranteed to render consistently
+else if (exports.asciiOnly) {
   exports.symbols.ok = 'Y';
   exports.symbols.err = 'N';
   exports.symbols.dot = '.';
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -233,7 +233,7 @@ function HTML(runner) {
 
   function updateStats() {
     // TODO: add to stats
-    var percent = (stats.tests / runner.total * 100) | 0;
+    var percent = (stats.tests / (runner.total * 100)) | 0;
     if (progress) {
       progress.update(percent).draw(ctx);
     }
--- a/lib/reporters/landing.js
+++ b/lib/reporters/landing.js
@@ -52,7 +52,7 @@ function Landing(runner) {
   var width = (Base.window.width * 0.75) | 0;
   var total = runner.total;
   var stream = process.stdout;
-  var plane = color('plane', '✈');
+  var plane = color('plane', Base.forceASCII ? '>' : '✈');
   var crashed = -1;
   var n = 0;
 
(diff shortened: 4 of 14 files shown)


Alhadis:
I'm not sure what to do about coverage. I tried adding specs for this, but they gave false positives when testing them locally.

Open to suggestions here.

plroebuck:
Wondering whether this is worth the effort. Why would you be running Mocha test on "console" to begin with? "wscons" -> "uwscons". Well, no answer for LFB, but...
The far more likely case for Unicode issues is that the terminal is fine, but misconfigured.
Might be simpler to just mandate UTF8 support required in documentation.

* `TERM`-sniffing is _hardly_ indicative of a terminal's capability.
   * What happens with MinGW since it's running on Windows
* None of this code (or dependent package) verifies lack of Unicode support.
* Would expect `--unicode`/`--no-unicode` cmdline options
* `Mocha.prototype.unicode()` to pass parsed cmdline options through to `Base` reporter
* Only worries about symbol output in reporters. Does nothing about textual output.
  * Assumes UTF8 overlap with ASCII adequate to display most output (assumes Western)
* Did you consider using dash as fallback for "pass" and `X` for "fail"?

Maybe we just call this what it is: `--use-ascii-symbols`.

Alhadis:
> Why would you be running Mocha test on "console" to begin with?

Is this a serious question?

> The far more likely case for Unicode issues is that the terminal is fine, but misconfigured.

You know we're talking about character-based displays, not terminal emulators, right?  You don't "reconfigure" a system console's font without recompiling the kernel.

> "wscons" -> "uwscons"

Not sure how you know of that, but `uwscons` isn't developed anymore and it isn't used on OpenBSD. And you don't get a choice what console driver an OS uses without recompiling it.

> `TERM`-sniffing is *hardly* indicative of a terminal's capability.

That's what `supports-color` and numerous other libraries do.

> What happens with MinGW since it's running on Windows?

Nothing. Windows is skipped when doing the check.I assume that was evident from the code changes.

> Assumes UTF8 overlap with ASCII adequate to display most output (assumes Western)

I'm *assuming* you mean *"suite and test titles with Unicode symbols still get garbled"*, and that you're not implying codepoints 0-127 are ever going to change...

> Did you consider using dash as fallback for "pass" and X for "fail"?

Pending tests are already marked with a dash.

> Would expect `--unicode`/`--no-unicode` cmdline options

An environment variable would probably be more consistent with user expectations, given the existence of `MOCHA_FORCE_COLORS`.

Alhadis:
Let me stress that I'm not attempting to make output look perfect by guarding against all Unicode in dumbed-down displays. If test titles contain Unicode, it's not Mocha's place to filter them.

This isn't about Unicode at large; it's simply an enhancement to improve readability when it's possible to do so. If you consider that too niche, then I'll close this PR and happily never bother you again.

> I looked around to see if there were any "supports-unicode" packages, but found nothing.

You won't find such a thing, because it's impossible. The closest you'll get is a crude hack that measures the horizontal offset reported by an addressable cursor after printing a multibyte sequence. Which does nothing if not connected to a (smart) TTY, and is only useful to detect *multibyte* sequence handling.

> I was asking why running a "dumb" terminal from an MinGW environment under Windows shouldn't get the treatment?

Because we're already dealing with Windows by serving them a different set of characters. Which, BTW, isn't covered in the tests either (hence why I wasn't sure how I was supposed to assert them).

> Bummer. How about 'asterisk'?

That could work, though at this point I'm feeling like this PR is headed for rejection. :p

> Why would you be using the system console to run Mocha tests?

Because that's what most VirtualBox machines boot to by default, and I'd rather not install a window manager on each *nix box I'm running tests on.

And, of course, an excuse to read text displayed with Sun Gallant Demi. :-)

plroebuck:
@ Alhadis  wrote:
> Let me stress that I'm not attempting to make output look perfect by guarding against all Unicode in dumbed-down displays. If test titles contain Unicode, it's not Mocha's place to filter them.
> 
> This isn't about Unicode at large; it's simply an enhancement to improve readability when it's possible to do so. If you consider that too niche, then I'll close this PR and happily never bother you again.

The opening sentence of second paragraph implied otherwise (Unicode at large, that is) to me.
Mocha's symbol/color exports are kinda a mess already, so what's a little more, I guess.

> > I looked around to see if there were any "supports-unicode" packages, but found nothing.
> You won't find such a thing, because it's impossible. The closest you'll get is a crude hack that measures the horizontal offset reported by an addressable cursor after printing a multibyte sequence. Which does nothing if not connected to a (smart) TTY, and is only useful to detect _multibyte_ sequence handling.

Yeah, that matches what I found. Write multibyte char and ask curses where cursor is afterwards.

> > I was asking why running a "dumb" terminal from an MinGW environment under Windows shouldn't get the treatment?
> Because we're already dealing with Windows by serving them a different set of characters.

Honestly, I have no idea what happens in that case. It's Windows in one sense, Unix in another.

>Which, BTW, isn't covered in the tests either (hence why I wasn't sure how I was supposed to assert them).

Mocha used to just accept PRs without tests. Yeah, disaster. Dunno when the Windows special treatment of symbols happened. Could also be caused by initial CI environments all being Linux, so no way to automatically test -- but all just speculation.

> > Bummer. How about 'asterisk'?
> That could work, though at this point I'm feeling like this PR is headed for rejection. :p

Not at all. But this PR _does_ hit a lot of reporter source files that have multiple changes already, not counting others already in the pipeline. 

> > Why would you be using the system console to run Mocha tests?
> Because that's what most VirtualBox machines boot to by default, and I'd rather not install a window manager on each *nix box I'm running tests on.

`uwm` isn't _that_ big...

> And, of course, an excuse to read text displayed with Sun Gallant Demi. :-)

LOL

Alhadis:
> But this PR does hit a lot of reporter source files that have multiple changes already, not counting others already in the pipeline.

I'll just leave it. I don't want to get under anybody's feet, especially since this PR is already pushing the limits of what's acceptable.

Guess I'll just write another custom reporter then. 😝 It [wouldn't be the first time](https://github.com/Alhadis/Atom-Mocha).

PR #3528, opened 2018-10-21
Extract runReporter to capture output under tests

outsideris:
### Description of the Change
As refactoring [dot reporter tests](https://github.com/mochajs/mocha/pull/3486), capturing output for all reporter tests moves to [`test/reporters/helpers.js`](https://github.com/mochajs/mocha/pull/3528/files#diff-04e8353a17acf7c3591b1b83eef5d6b8)(`createMockReporter`). And it has `tee` to show reporter's output in running tests.

### Alternate Designs
N/A

### Benefits
* Greatly simplifies setup/teardown related to capturing screen output, reducing possibility of error.
* Reporter tests are more maintainable and easier to follow.
* Reporter test screen output is no longer echo'd by default, pruning individual test output minutiae.

### Possible Drawbacks
New contributors will be lulled into thinking writing new tests for Mocha is pretty trivial.  **LOL**

### Applicable issues
semver-patch

### DIFF
--- a/test/reporters/doc.spec.js
+++ b/test/reporters/doc.spec.js
@@ -4,22 +4,15 @@ var reporters = require('../../').reporters;
 var Doc = reporters.Doc;
 
 var createMockRunner = require('./helpers.js').createMockRunner;
+var makeRunReporter = require('./helpers.js').createRunReporterFunction;
 
 describe('Doc reporter', function() {
-  var stdout;
-  var stdoutWrite;
   var runner;
-  beforeEach(function() {
-    stdout = [];
-    stdoutWrite = process.stdout.write;
-    process.stdout.write = function(string, enc, callback) {
-      stdout.push(string);
-      stdoutWrite.call(process.stdout, string, enc, callback);
-    };
-  });
+  var options = {};
+  var runReporter = makeRunReporter(Doc);
 
   afterEach(function() {
-    process.stdout.write = stdoutWrite;
+    runner = undefined;
   });
 
   describe('on suite', function() {
@@ -32,8 +25,7 @@ describe('Doc reporter', function() {
       };
       it('should log html with indents and expected title', function() {
         runner = createMockRunner('suite', 'suite', null, null, suite);
-        Doc.call(this, runner);
-        process.stdout.write = stdoutWrite;
+        var stdout = runReporter(this, runner, options);
         var expectedArray = [
           '    <section class="suite">\n',
           '      <h1>' + expectedTitle + '</h1>\n',
@@ -48,8 +40,7 @@ describe('Doc reporter', function() {
         };
         expectedTitle = '&#x3C;div&#x3E;' + expectedTitle + '&#x3C;/div&#x3E;';
         runner = createMockRunner('suite', 'suite', null, null, suite);
-        Doc.call(this, runner);
-        process.stdout.write = stdoutWrite;
+        var stdout = runReporter(this, runner, options);
         var expectedArray = [
           '    <section class="suite">\n',
           '      <h1>' + expectedTitle + '</h1>\n',
@@ -64,8 +55,7 @@ describe('Doc reporter', function() {
       };
       it('should not log any html', function() {
         runner = createMockRunner('suite', 'suite', null, null, suite);
-        Doc.call(this, runner);
-        process.stdout.write = stdoutWrite;
+        var stdout = runReporter(this, runner, options);
         expect(stdout, 'to be empty');
       });
     });
@@ -78,8 +68,7 @@ describe('Doc reporter', function() {
       };
       it('should log expected html with indents', function() {
         runner = createMockRunner('suite end', 'suite end', null, null, suite);
-        Doc.call(this, runner);
-        process.stdout.write = stdoutWrite;
+        var stdout = runReporter(this, runner, options);
         var expectedArray = ['  </dl>\n', '</section>\n'];
         expect(stdout, 'to equal', expectedArray);
       });
@@ -90,8 +79,7 @@ describe('Doc reporter', function() {
       };
       it('should not log any html', function() {
         runner = createMockRunner('suite end', 'suite end', null, null, suite);
-        Doc.call(this, runner);
-        process.stdout.write = stdoutWrite;
+        var stdout = runReporter(this, runner, options);
         expect(stdout, 'to be empty');
       });
     });
@@ -109,8 +97,7 @@ describe('Doc reporter', function() {
     };
     it('should log html with indents and expected title and body', function() {
       runner = createMockRunner('pass', 'pass', null, null, test);
-      Doc.call(this, runner);
-      process.stdout.write = stdoutWrite;
+      var stdout = runReporter(this, runner, options);
       var expectedArray = [
         '    <dt>' + expectedTitle + '</dt>\n',
         '    <dd><pre><code>' + expectedBody + '</code></pre></dd>\n'
@@ -128,8 +115,7 @@ describe('Doc reporter', function() {
       var expectedEscapedBody =
         '&#x3C;div&#x3E;' + expectedBody + '&#x3C;/div&#x3E;';
       runner = createMockRunner('pass', 'pass', null, null, test);
-      Doc.call(this, runner);
-      process.stdout.write = stdoutWrite;
+      var stdout = runReporter(this, runner, options);
       var expectedArray = [
         '    <dt>' + expectedEscapedTitle + '</dt>\n',
         '    <dd><pre><code>' + expectedEscapedBody + '</code></pre></dd>\n'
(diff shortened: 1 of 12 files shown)


test/reporters/helpers.js
```diff
@@ -155,9 +155,44 @@ function makeExpectedTest(
...
+}
+
 module.exports = {
   createMockRunner: createMockRunner,
   makeTest: makeTest,
   createElements: createElements,
-  makeExpectedTest: makeExpectedTest
+  makeExpectedTest: makeExpectedTest,
+  createMockReporter: createMockReporter
 };
```
plroebuck: Put exports in alphabetical order.

test/reporters/helpers.js
```diff
@@ -155,9 +155,44 @@ function makeExpectedTest(
...
+    reporter.call(stubself, runner, options);
+
+    // revert stream reassignment here so reporter output
+    // can't be corrupted if any test assertions throw
+    process.stdout.write = stdoutwrite;
+
+    return stdout;
+  };
+}
+
```
plroebuck:
```js
/**
 * Creates closure with reference to the reporter class constructor.
 *
 * @param {Function} ctor - Reporter class constructor
 * @return {createRunReporterFunction~runReporter} 
 */
function createRunReporterFunction(ctor) {
  /**
   * Run reporter using stream reassignment to capture output.
   *
   * @param {Object} stubSelf - Reporter-like stub instance
   * @param {Runner} runner - Mock instance
   * @param {Object} [options] - Reporter configuration settings
   * @param {boolean} [tee=false] - Whether to echo output to screen
   * @return {string[]} Lines of output written to `stdout`
   */
  var runReporter = function(stubSelf, runner, options, tee) {
    var stdout = [];

    // Reassign stream in order to make a copy of all reporter output
    var stdoutWrite = process.stdout.write;
      stdout.push(string);
      if (tee) {
      }
    };

    // Invoke reporter
    ctor.call(stubSelf, runner, options);

    // Revert stream reassignment here so reporter output
    // can't be corrupted if any test assertions throw

    return stdout;
  };

  return runReporter;
}
```

test/reporters/dot.spec.js
```diff
@@ -5,42 +5,17 @@ var Dot = reporters.Dot;
 var Base = reporters.Base;
 
 var createMockRunner = require('./helpers.js').createMockRunner;
+var createMockReporter = require('./helpers.js').createMockReporter;
```
plroebuck: Renamed since it's not a mock at all.

test/reporters/json-stream.spec.js
```diff
@@ -5,12 +5,12 @@ var JSONStream = reporters.JSONStream;
 
 var createMockRunner = require('./helpers').createMockRunner;
 var makeExpectedTest = require('./helpers').makeExpectedTest;
+var createMockReporter = require('./helpers.js').createMockReporter;
 
 describe('Json Stream reporter', function() {
```
plroebuck:
```js
describe('JSONStream reporter', function() {
```

outsideris:
> Idea was to default to not echo, but make it simple to get output from single test. No need to specify this in the closure.

I still leave `showOutput` in the closure. Currently, only progress reporter tests use "Reporter configuration settings".  So, other tests should call `runReporter` as `runReporter({epilogue: function() {}}, runner, null, showOutput);`. Specifying `options` as `null` is too verbose for me. IMO, it's enough to handle output in suite level by passing `tee` into `createRunReporterFunction`, because we can run a single test.

plroebuck:
> > Idea was to default to not echo, but make it simple to get output from single test. No need to specify this in the closure.
> 
> I still leave `showOutput` in the closure. Currently, only progress reporter tests use "Reporter configuration settings". So, other tests should call `runReporter` as `runReporter({epilogue: function() {}}, runner, null, showOutput);`. Specifying `options` as `null` is too verbose for me. IMO, it's enough to handle output in suite level by passing `tee` into `createRunReporterFunction`, because we can run a single test.

No. Put `tee` back in `runReporter` where it belongs. No need of it in closure since we **don't** echo by default anyway. We could have a constant `var ECHO = true;`, rather than bother with `showOutput` (which is closer to new usage anyway).
`options` are coming to _all_ reporters, just not there yet.

plroebuck:
Besides, you could temporarily use a [partialRight](https://lodash.com/docs/4.17.10#partialRight) to default those arguments if you wanted to echo output for the _entire_ suite.

test/reporters/doc.spec.js
```diff
@@ -4,22 +4,16 @@ var reporters = require('../../').reporters;
 var Doc = reporters.Doc;
 
 var createMockRunner = require('./helpers.js').createMockRunner;
+var createRunReporterFunction = require('./helpers.js')
+  .createRunReporterFunction;
```
plroebuck: Is `prettier` causing this to be on two lines? If so,
```js
var makeRunReporter = require('./helpers.js').createRunReporterFunction;
```

outsideris:
Fixed.

test/reporters/list.spec.js
```diff
@@ -98,7 +101,7 @@ describe('List reporter', function() {
       );
 
       Base.cursor = cachedCursor;
-      Base.symbols = cachedSymbols;
+      Base.symbols.ok = cachedOkSymbol;
```
outsideris: Additionally, I found original code didn't restore ok symbol because `Base.symbols.ok = expectedOkSymbol;` is changed refrenced `cachedSymbols`.



Now it fixed.

test/reporters/doc.spec.js
```diff
@@ -4,22 +4,15 @@ var reporters = require('../../').reporters;
...
   var runner;
-  beforeEach(function() {
-    stdout = [];
-    stdoutWrite = process.stdout.write;
-    process.stdout.write = function(string, enc, callback) {
-      stdout.push(string);
-      stdoutWrite.call(process.stdout, string, enc, callback);
-    };
-  });
+  var showOutput = false;
```
plroebuck: How about rename `showOutput` and add `options` here?
```js
  var options = {};
  var echo = false;
```

test/reporters/doc.spec.js
```diff
@@ -32,8 +25,7 @@ describe('Doc reporter', function() {
       };
       it('should log html with indents and expected title', function() {
         runner = createMockRunner('suite', 'suite', null, null, suite);
-        Doc.call(this, runner);
-        process.stdout.write = stdoutWrite;
+        var stdout = runReporter(this, runner, null, showOutput);
```
plroebuck:
```js
        var stdout = runReporter(this, runner, options, echo);
```

test/reporters/helpers.js
```diff
@@ -155,9 +155,51 @@ function makeExpectedTest(
...
 module.exports = {
-  createMockRunner: createMockRunner,
-  makeTest: makeTest,
   createElements: createElements,
-  makeExpectedTest: makeExpectedTest
+  createMockRunner: createMockRunner,
+  createRunReporterFunction: createRunReporterFunction,
+  makeExpectedTest: makeExpectedTest,
+  makeTest: makeTest
 };
```
plroebuck: Let's add `noop`.
```js
module.exports = {
  // everything already there
  noop: function() {}
};
```

test/reporters/dot.spec.js
```diff
@@ -61,7 +36,12 @@ describe('Dot reporter', function() {
   describe('on start', function() {
     it('should write a newline', function() {
       runner = createMockRunner('start', 'start');
-      runReporter({epilogue: function() {}}, runner, showOutput);
+      var stdout = runReporter(
+        {epilogue: function() {}},
```
plroebuck:
```js
        {epilogue: noop},
```

test/reporters/dot.spec.js
```diff
@@ -5,42 +5,17 @@ var Dot = reporters.Dot;
 var Base = reporters.Base;
 
 var createMockRunner = require('./helpers.js').createMockRunner;
+var makeRunReporter = require('./helpers.js').createRunReporterFunction;
```
plroebuck:
```js
var noop = function() {};
```

test/reporters/dot.spec.js
```diff
@@ -61,7 +36,12 @@ describe('Dot reporter', function() {
   describe('on start', function() {
```
plroebuck:
```js
var noEpilogueStub = { epilogue: noop };
```

test/reporters/list.spec.js
plroebuck: I had noticed this as well. But your implementation only fixes one aspect of a larger problem. Please remove your "fix" and let's open another issue for this. For example, the `Base.cursor` functions are also "twiddled" with and not restored properly. It just goes unnoticed because it's never _actually_ used again. All tests messing around with any `Base.exports.<symbol>` should be deep cloning prior to any "twiddling".

plroebuck:
I'll write new issue for discovered cloning problem with reporter tests.
Keep your checkmark fix here if you like, or reapply it after this gets merged with new issue. 

But it would be nice to resolve the last two issues, and get this merged before Mocha-6.0.
Have other changes going on in _these same files_.

outsideris:
Done! Sorry for late work. I also choose the latter case.
And I removed a commit to fix Symbol problem.

PR #3535, opened 2018-10-26
Add ability to query Mocha version programmatically

plroebuck:
### Description of the Change

Added `version` property so current Mocha version can easily be determined programmatically.

### Alternate Designs

Original design used a method.

N/A

### Benefits

Often useful to include such meta-information when comparing output during upgrades.
Quick check for version confirmation (if debugging).

### Possible Drawbacks

None.

### Applicable issues

semver-minor


lib/mocha.js
```diff
@@ -677,6 +677,20 @@ Mocha.prototype.forbidPending = function() {
...
+ *
+ * @public
+ * @returns {string} module version as specified by "package.json".
+ * @example
+ *
+ * mocha.version();
+ * // => '5.3.0'
+ */
+Mocha.prototype.version = function() {
+  return require('../package').version;
```
boneskull: huh, I always thought this wanted the extension unless it was a `.js` file.

anyway: what does browserify do with this code?

If I'm being honest, I would rather this be a property, something like:

```js
Object.defineProperty(Mocha.prototype, 'version', {
  value: require('../package').version
});

// or writable, for whatever reason
Mocha.prototype.version = require('./package.version');
```

plroebuck: > anyway: what does browserify do with this code?

After new commit, works same on both node and browser. Used a `browserify` transform that _only_ keeps the version from "package.json" (for better security).

> If I'm being honest, I would rather this be a property, something like:

Converted to use `Object.defineProperty`. I like this better too.

plroebuck:
@boneskull anything else?

lib/mocha.js
```diff
@@ -677,6 +677,20 @@ Mocha.prototype.forbidPending = function() {
...
+/**
+ * Mocha version as specified by "package.json".
+ *
+ * @name Mocha#version
+ * @type string
+ * @readonly
+ */
+Object.defineProperty(Mocha.prototype, 'version', {
+  value: require('../package').version,
+  configurable: false,
```
boneskull: FWIW these are the default values for `configurable` and `writable` and could be omitted.

Is there a reasoning behind `enumerable: true`?

boneskull: *ping* @plroebuck

plroebuck: Defaults were explicit to imply intent (as there were only three boolean properties) rather than oversight. Found various examples (like [this](https://github.com/nodejs/node/blob/e7f710c1d4729890806610e5450d7a0edfb4ac11/lib/perf_hooks.js#L224)) within Node.js itself that did same.
`enumerable: true` allowed it to show up in `for` loop over properties.

You feel otherwise concerning the style?

plroebuck: @ORESoftware, don't know how to interpret your "thumbs up"...

Issue #2483, opened 2016-09-18
audit slow tests


boneskull:
some tests may need new `slow` values (e.g. `this.slow(1000);`) since they consistently run at a certain speed.
1.  run all mocha tests
2.  find the ones reporting slowness
3.  increase this value to something reasonable if the test can't be made faster

coderbyheart:
What's the difference to #2597?

Bamieh:
Mocha has this feature now, reopen if i am misunderstanding this issue.

Issue #2837, opened 2017-05-27
normalize test timeouts


boneskull:
Various tests generally produce "slow" warnings.  The "slow" threshold is based on a fraction of the timeout setting, unless it is otherwise supplied.

The timeout setting is set to a default of 200ms as per [mocha.opts](https://github.com/mochajs/mocha/blob/master/test/mocha.opts#L7).  This is *generally* OK for unit tests, but inappropriate for anything invoking Mocha itself.

However, there are various overrides in `Makefile` and the test files themselves; let's find a single place to put them all, and avoid one-off settings, if possible.

It's important to note that our tests run much more slowly on AppVeyor.  That should be the baseline for timeout values--not your MacBook Pro. :wink:

ScottFreeCode:
Seems like #2597 and this are duplicates of each other?

juergba:
closing, see #3539.

PR #3539, opened 2018-10-29
Normalize Mocha tests to have realistic timings

### LINKED ISSUE #2597: make test timeouts closer to reality
Many tests in our suite regularly output "slowness" warnings:

image

Let's increase the `--timeout` and/or `--slow` setting so the information it produces is actually helpful.

WaleedAshraf:
### Description of the Change

Add `.slow()` to tests which are normally taking longer time.

### Benefits

More clear view of actual slow tests.

### Possible Drawbacks

Should keep an eye on these test for later -- if they are getting even slower, then something's 🐟 !

### Applicable issues
Fixes #2597
semver-patch

See also #2837

### DIFF
--- a/test/integration/compiler-globbing.spec.js
+++ b/test/integration/compiler-globbing.spec.js
@@ -5,6 +5,7 @@ var path = require('path');
 
 describe('globbing like --compilers', function() {
   it('should find a file of each type', function(done) {
+    this.slow(900);
     exec(
       '"' +
         process.execPath +
--- a/test/integration/glob.spec.js
+++ b/test/integration/glob.spec.js
@@ -6,6 +6,7 @@ var path = require('path');
 var node = '"' + process.execPath + '"';
 
 describe('globbing', function() {
+  this.slow(600);
   describe('by the shell', function() {
     it('should find the first level test', function(done) {
       testGlob.shouldSucceed(
--- a/test/integration/hooks.spec.js
+++ b/test/integration/hooks.spec.js
@@ -7,6 +7,7 @@ var splitRegExp = require('./helpers').splitRegExp;
 var args = ['--reporter', 'dot'];
 
 describe('hooks', function() {
+  this.slow(900);
   it('are ran in correct order', function(done) {
     runMocha('cascade.fixture.js', args, function(err, res) {
       var lines, expected;
--- a/test/integration/no-diff.spec.js
+++ b/test/integration/no-diff.spec.js
@@ -4,6 +4,7 @@ var helpers = require('./helpers');
 var run = helpers.runMocha;
 
 describe('no-diff', function() {
+  this.slow(900);
   describe('when enabled', function() {
     it('should not display a diff', function(done) {
       run('no-diff.fixture.js', ['--no-diff'], function(err, res) {
--- a/test/integration/only.spec.js
+++ b/test/integration/only.spec.js
@@ -4,6 +4,7 @@ var run = require('./helpers').runMochaJSON;
 var assert = require('assert');
 
 describe('.only()', function() {
+  this.slow(900);
   describe('bdd', function() {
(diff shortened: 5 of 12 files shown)


plroebuck:
Haven't reviewed the rest, but definitely revert 6b93c95. Changes to test specifications do not require _any_ changes to "package.json".

WaleedAshraf:
@plroebuck done.

plroebuck:
Could you explain the process you used to come up with your slow threshold values?

WaleedAshraf:
I ran tests for 3,4 times and added  (~3 * average value e of duration)

If a test takes 100ms
- slow(100) would be read
- slow (200) would orange/yellow
- slow (300) would be green
- slow (>300) would be no warning


Also, if it's a **semver-patch**, then we need to update package.json with version bump.

plroebuck:
Were the tests run on AppVeyor?

`semver-patch` annotates its ability to be merged directly to master here; the package version itself will change at next release.

WaleedAshraf:
@plroebuck no. I ran them locally.
Is AppVeyor possible on local? Or maybe I can re-run the AppVeyor job here?

Got it 👍

plroebuck:
> @plroebuck no. I ran them locally.
> Is AppVeyor possible on local? Or maybe I can re-run the AppVeyor job here?

Yeah, local timings won't be as much help. Need to run against the 80386 machines Appveyor seems to be powered by... :)

You can use timings from your commits. Try to get some spread between when you collect results -- certain times of day are worse than others.

plroebuck:
BTW, you are using the _actual_ [formula](https://github.com/mochajs/mocha/pull/3466#issuecomment-417144359), right?

boneskull:
this also resolves #2483... lots of dupes.

test/integration/suite.spec.js
```diff
@@ -4,49 +4,53 @@ var assert = require('assert');
 var run = require('./helpers').runMocha;
 var args = [];
 
-describe('suite w/no callback', function() {
+describe('suit', function() {
```
boneskull: typo

plroebuck:
#2483 already closed.

WaleedAshraf:
@plroebuck ok. I'll cross check with previous builds and will update values where needed.
And yes, I used the same formula.

@boneskull good catch. Will push updates by this weekend. 👍

craigtaub:
@WaleedAshraf any movement on this?

WaleedAshraf:
Hi @craigtaub, Sorry for leaving it here for long.
I'll push changes in few days.

WaleedAshraf:
@here, I think it's not needed now.
I check some of the last builds at https://travis-ci.org/mochajs/mocha/pull_requests
and I don't see many slow tests. Only one or two at max, which is fine.

We can also this PR with #2597 and #2837

craigtaub:
Have you checked on [AppVeyor](https://ci.appveyor.com/project/boneskull/mocha/history)?

To quote issue.
> It's important to note that our tests run much more slowly on AppVeyor. That should be the baseline for timeout values--not your MacBook Pro.

WaleedAshraf:
@craigtaub I did, but it's not that clear because logs are not colored and it's difficult to identify slow (red) tests.
Maybe we need to fix that also.

craigtaub:
Ah yes, they don't support colours so that won't be possible.

WaleedAshraf:
Hi,
Let's get done with this.

I checked the lastest build on AppVeyor and there is only one test which hits the limit of [3750ms](https://github.com/mochajs/mocha/blob/master/package-scripts.js#L118)
 
```
regressions
    √ issue-1991: Declarations do not get cleaned up unless you set them to `null` - Memory Leak (16234ms)
```

But this is way too over the limit. Rest of tests are under the limit.

outsideris:
I agreed to close this issue. Any thoughts?

PR #2962, opened 2017-08-15
Add Notifications for Mocha in browser

BuonOmo:
This PR reference the thread started on PR #2938. It allows user to use `growl` option even if he is in browser, by replacing growl notifications for [Web API Notifications](https://developer.mozilla.org/en-US/docs/Web/API/notification).

Unit tests have not been made yet, I'd like to see first if this PR suits you. However I've already tried it in a personal project and it works like a charm.

PR #3517, opened 2018-10-15
3483: Squelch CI Growl-related spawn errors

plroebuck:
### Description of the Change

Installed binary needed for Linux desktop notification support as part of TravisCI job setup.
Installed binary needed for Windows desktop notification support as part of Appveyor job setup.

### Alternate Designs

N/A

### Why should this be in core?

N/A

### Benefits

Misleading Growl-related spawn errors (unrelated to what is being tested) will no longer appear
in TravisCI nor Appveyor job output.

### Possible Drawbacks

Tests will take slightly longer (<10s) to complete due to APT package install on TravisCI.
Timing not provided for Appveyor.

### Applicable issues

Fixes #3483
semver-patch

Issue #11, opened 2011-09-06
optional parallelism via processes


tj:
for cpu-bound tests

PR #3542, opened 2018-11-01
Growl: Web notifications & Desktop prerequisite software check

### LINKED ISSUE #3111: Warn on incorrect Growl usage

plroebuck:
### Description of the Change

Web notifications are now implemented as part of Mocha's browser bundle,
given a relatively modern browser implementing Notifications API.

Added (minorly brittle) means to verify prerequisite software is installed.
Migrated Growl support code to its own file. This implementation's checks
are required to enable Growl; failure will write notice to `stderr`.

Other modifications based on discussion from Chad Rickman's PR #3311.
This also checks for errors from Growl completion callback.

### Alternate Designs

Hmmm.

N/A

### Benefits

Notification support for browser now available.
Will no longer attempt to use desktop notifications if prerequisites not installed.

### Possible Drawbacks

Not all browsers are capable... yes, I'm referring to MSIE.

Minor brittle scheme to make its determination that _really_ belongs in `growl` package.
But they haven't changed the prerequisite software since 2012 so...

### Applicable issues

Closes #2962
Closes #3111

semver-patch (browser: implements the existing public API)
semver-patch (CLI: what it fixes is nuisance output from external process.)


lib/growl.js
```diff
@@ -0,0 +1,151 @@
...
+   *     process. It will include timestamp and executed command with arguments.
+   */
+
+  function onCompletion(err, stdout, stderr) {
+    if (err) {
+      if (err.code === 'ENOENT') {
+        console.error(
+          'Growl notification support not installed... listener removed!'
+        );
+        runner.removeListener('end', sendNotification);
```
plroebuck: So we had this chunk of code from #3311, but I now wonder about its utility.
L74 should now only occur if the user uninstalls the software _during_ the run (unlikely). Even so, Growl is only invoked _after_ all tests are done. Does it buy anything to bother removing a listener that has completed a one-time event?

lib/growl.js
```diff
@@ -0,0 +1,151 @@
...
+      } else {
+        console.error('growl: ', err.message);
+      }
+    }
+  }
+
+  // :TBD: should this be `.once`?
+  runner.on('end', sendNotification);
```
plroebuck: Someone care to elaborate why this wasn't using `.once`?

plroebuck:
@chadrickman, @TheSench, please add your thoughts.

lib/mocha.js
```diff
@@ -290,32 +283,6 @@ Mocha.prototype.loadFiles = function(fn) {
...
-  runner.on('end', function() {
-    var stats = reporter.stats;
-    if (stats.failures) {
-      var msg = stats.failures + ' of ' + runner.total + ' tests failed';
-      notify(msg, {name: 'mocha', title: 'Failed', image: image('error')});
-    } else {
-      notify(stats.passes + ' tests passed in ' + stats.duration + 'ms', {
-        name: 'mocha',
-        title: 'Passed',
-        image: image('ok')
```
plroebuck: Was left wondering why this wasn't triggering browser test errors. Via `browserify`, the `require('growl')` was translated to a `noop()` and `path` module was set to `false`. Yet in order to invoke `notify` (a.k.a. `noop`), its arguments would need to be assembled first. When creating the options, the `image` function should have blown up attempting to `path.join`, no?

lib/growl.js
Bamieh: I believe maybe it was related to running mocha in watch mode? I think its safe to use `once` now.

lib/mocha.js
```diff
@@ -9,6 +9,7 @@
 var escapeRe = require('escape-string-regexp');
 var path = require('path');
 var builtinReporters = require('./reporters');
+var growl = require('./growl');
```
boneskull: should we be telling the browser to stub this file?

lib/mocha.js
boneskull: gonna guess it's because `growl: true` was never set, which means this function would never get called.

lib/growl.js
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @description
+ * Takes a list of command names and searches the path for each executable
+ * file that would be run had these commands actually been invoked.
+ *
+ * @private
+ * @param {string[]} binaries - Names of binary files to search for.
+ * @return {string} absolute path of first binary found, or empty string if none
+ */
+function which(binaries) {
```
boneskull: would strongly prefer that if we need this functionality, we get it from a purpose-built module, e.g. [which](https://npm.im/which).

lib/growl.js
```diff
@@ -0,0 +1,150 @@
...
+ * @description
+ * Somewhat brittle dependency on `growl` package implementation, but it
+ * rarely changes.
+ *
+ * @private
+ * @see {@link https://github.com/tj/node-growl/blob/master/lib/growl.js#L28-L126|setupCmd}
+ * @return {string[]} names of Growl support binaries
+ */
+function getSupportBinaries() {
+  switch (os.type()) {
```
boneskull: would prefer a key/value lookup instead of a `switch`

lib/mocha.js
plroebuck: Don't think so. Part of Mocha's public API previously. This continues same (with warning/error).

lib/growl.js
plroebuck: Okay. `switch` is faster for smaller #choices, but KV is more maintainable.

plroebuck:
> 1. my main issue is the homebrew `which` implementation

Hope my explanation above addresses your concern. My primary goal was mirroring _exactly_ what the `growl` package was going to do.

> 1. pretty sure "lib/growl.js" should be excluded/stubbed from bundled code

Perhaps. But it _was_ available before, so I left it as such.
It should be made into a Mocha reporter plug-in (assuming we supported such).

> 1. would like to see some more tests against "lib/growl.js"

Not much more I _can_ test from Mocha. The `growl` package is essentially a black box; from our side, we can attempt to send a message. Since TJ's package invokes an implementation binary via child process to do the actual notification, we have no access. CI-wise, we _just_ got the underlying support software installed on Travis and Appveyor (#3517) so it would _quit_ throwing child process errors.

lib/mocha.js
```diff
@@ -440,18 +407,49 @@ Mocha.prototype.fullTrace = function() {
...
  * @return {Mocha} this
  * @chainable
  */
 Mocha.prototype.growl = function() {
-  this.options.growl = true;
+  this.options.growl = this.isGrowlCapable();
+  if (!this.options.growl) {
+    var detail = process.browser
+      ? 'Growl notification support not available in browser... cannot enable!'
+      : 'Growl notification support not installed... cannot enable!';
```
outsideris: How about adding [installation guide](https://github.com/tj/node-growl#installation) in this error message?

plroebuck: Think that would be better suited for website than source code. Although now that you mention it, the user guide should link to the [wiki](https://github.com/mochajs/mocha/wiki/Growl-Notifications).

outsideris: Users can see this message when they are using `--growl` without growl.
So, it's better that let users know how to install growl notification when they try to use growl notification than website or wiki.

plroebuck: They see every other error message Mocha generates as well, but you didn't add links for any of those.

Although it might be interesting to add a page to our website called "So I Got this Error..."

We have a website for this. I don't mind converting the existing wiki page into a user guide addition. TJ's instructions are _outdated_ for macOS (I rewrote them 6mos ago). Since half a year went by before anyone noticed, why would we want to redirect _anyone_ there?

plroebuck:
I updated Mocha's Wiki to add [installation instructions](https://github.com/mochajs/mocha/wiki/Growl-Notifications).

lib/mocha.js
plroebuck: I would prefer to leave support as an option for both environments. Although not yet available in IE, it is possible to provide similar notifications within browser using the [Notification API](https://developer.mozilla.org/en-US/docs/Web/API/notification). But it would need to come in a different PR to allow time for sufficient testing.

plroebuck:
@boneskull have I addressed your concerns?

boneskull:
@plroebuck 	I'm kind of confused about this.  Can't we just upgrade to the newest growl which doesn't eat errors?

plroebuck:
@boneskull, Rickman's node-growl [patch](https://github.com/tj/node-growl/pull/72) made it so child process errors are reported if they occurred; we switched to that update 6 months ago. This PR allows determining if prerequisite software is installed (i.e., don't enable Growl if it won't work). Opposite ends of the problem. Mine checks before bothering to use, existing code uses and errors.

The PR contains a client-side implementation of tj/node-growl#74. Plan is to migrate it there, then use it here. Would have done it the other way around, but their maintainer just responded to my 6 month old issue last week.

lib/mocha.js
plroebuck: Or not?

plroebuck:
@boneskull ping.

The [Windows Growl installer site](http://www.growlforwindows.com/gfw/default.aspx) is not responding (and hasn't for about a week), which is why the AppVeyor build won't pass.

plroebuck:
@boneskull, I'll merge this PR assuming you have no further objections.

docs/index.md
```diff
@@ -1295,6 +1298,69 @@ The "HTML" reporter is what you see when running Mocha in the browser.  It looks
...
 
+## Desktop Notification Support
+
+Desktop notifications allow asynchronous communication of events without
+forcing you to react to a notification immediately. Their appearance
+and specific functionality vary across platforms. They typically disappear
+automatically after a short delay, but their content is often stored in some
+manner that allows you to access past notifications.
+
+[Growl][] was an early notification system implementation for OS X and Windows,
```
boneskull: I think you can even omit `[]`

docs/index.md
```diff
@@ -1295,6 +1298,69 @@ The "HTML" reporter is what you see when running Mocha in the browser.  It looks
...
+<html>
+  <head>
+    <title>Mocha</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="https://unpkg.com/mocha@5.3.0/mocha.css" />
+  </head>
+  <body>
+    <div id="mocha"></div>
+    <script src="https://unpkg.com/mocha@5.3.0/mocha.js"></script>
```
boneskull: weird mocha version?

lib/browser/growl.js
```diff
@@ -1,5 +1,212 @@
...
+ * @see {@link Mocha#isGrowlPermitted}
+ * @returns {Promise<boolean>} promise determining if browser notification
+ *     permissible when fulfilled.
+ */
+function isPermitted() {
+  var permitted = {
+    granted: Promise.resolve(true),
+    denied: Promise.resolve(false),
+    default: (function ask() {
+      Notification.requestPermission().then(function(permission) {
```
boneskull: what's the IIFE for?

lib/browser/growl.js
```diff
@@ -1,5 +1,212 @@
...
+ *     notification permissible when fulfilled.
+ * @returns {Growl~CanNotifyPromise} Should notification proceed?
+ */
+function canNotify(value) {
+  if (!value) {
+    var why = value === false ? 'blocked' : 'unacknowledged';
+    var reason = 'not permitted by user (' + why + ')';
+    return Promise.reject(new Error(reason));
+  }
+  return Promise.resolve(true);
```
boneskull: is resolving `true` necessary, or can we just resolve w/ void, e.g., `Promise.resolve()`?

if we're never going to resolve with `false`--if resolution means success--then any sort of parameter is unneeded

lib/browser/growl.js
```diff
@@ -1,5 +1,212 @@
...
+ * @param {Runner} runner - Runner instance.
+ * @private
+ */
+function display(runner) {
+  var stats = runner.stats;
+  var symbol = {
+    cross: '\u274C',
+    tick: '\u2705'
+  };
+  var logo = require('../../package').logo;
```
boneskull: I'm not sure if browserify's resolution algorithm is going to be able to tell that this is the same thing as `package.json`.  can you please ensure that there aren't *two* `package.json`'s inlined in the browser bundle?

lib/growl.js
boneskull: I mean, there are cases here that we're not considering including permissions and execution bits... would rather just have that all taken care of than have to implement it ourselves.

Hand-rolling stuff like this is why our `--watch` functionality is such a mess.

lib/growl.js
```diff
@@ -0,0 +1,161 @@
...
+ * @see {@link https://github.com/tj/node-growl/blob/master/lib/growl.js#L28-L126|setupCmd}
+ * @return {string[]} names of Growl support binaries
+ */
+function getSupportBinaries() {
+  var binaries = {
+    Darwin: ['terminal-notifier', 'growlnotify'],
+    Linux: ['notify-send', 'growl'],
+    Windows_NT: ['growlnotify']
+  };
+  return binaries[os.type()] || [];
```
boneskull: should we be breaking earlier than this if the platform is unsupported?  SunOS, AIX, etc

scripts/package-json-minify.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+/**
+ * This is a transform stream used to minify "package.json" results
```
boneskull: please rename this script.  when we think minification, we think uglification... I'm not sure what it should be called (distillation?) but it's a misnomer as is.

docs/index.md
plroebuck: Had assumed a minor release was next when written. You want `6.0.0`?

lib/growl.js
plroebuck: unnecessary micro-optimization.

lib/growl.js
plroebuck: The **only** thing being done here is to check if _something_ is installed. TJ's package will handle **all** other issues when it spawns the notifier on our behalf.

Since all the notification binaries are put there via installers, permissions and such are _very_ unlikely to be a problem anyway.

scripts/package-json-minify.js
plroebuck: Other possibilities:
* quantizify
* remandraify
* vestigify
* basilify
* cuttotheboneify

boneskull: I apologize--I shouldn't have used "we" in my original comment, since you and I are in disagreement.

But I can *also* find a source documenting my definition if we want to take the pedantry to another level.

*I* would expect "minification" of a JSON to be "not prettified"; in other words, the original JSON less any formatting.

docs/index.md
plroebuck: Took some time to find documentation on [shortcut reference link](https://github.github.com/gfm/#shortcut-reference-link), but prefer the [collapsed reference link](https://github.github.com/gfm/#collapsed-reference-link) style; . Didn't know the former even existed, and I've been writing markdown for years. As I sometimes use square brackets in my text, I'm not sure I care for this _feature_.

docs/index.md
boneskull: ah, yeah, I think the next will be v6.0.0, sorry I wasn't clear.  if you feel like we need a minor before then, then we could cherry-pick our way to into it, but I'd rather just cut `master`

docs/index.md
boneskull: agree it's unexpected; I didn't discover it until I saw it being used in Mocha's CHANGELOG.  might be a peculiarity of GFM

lib/browser/growl.js
plroebuck: Leftover from testing. Replaced.

lib/browser/growl.js
plroebuck: It does. There's only one (search for "homepage" to find the _cullified_ version).

lib/browser/growl.js
plroebuck: OK.

docs/index.md
plroebuck: Think we _should_ have but can't now... The engine specification changed long ago.

plroebuck: Done

docs/index.md
```diff
@@ -1451,6 +1517,13 @@ $ REPORTER=nyan npm test
 
 In addition to chatting with us on [Gitter], for additional information such as using spies, mocking, and shared behaviours be sure to check out the [Mocha Wiki](https://github.com/mochajs/mocha/wiki) on GitHub. For discussions join the [Google Group](https://groups.google.com/group/mochajs). For a running example of Mocha, view [example/tests.html](example/tests.html). For the JavaScript API, view the [API documentation](api/) or the [source](https://github.com/mochajs/mocha/blob/master/lib/mocha.js#L51).
 
+[//]: # (Cross reference section)
+
+[Growl]: http://growl.info/
+[Growl-install]: https://github.com/mochajs/mocha/wiki/Growl-Notifications
+[Mocha-gitter]: https://gitter.im/mochajs/mocha
+[Mocha-website]: https://mochajs.org/
+
```
plroebuck: @boneskull,
I have a different PR out for updating this documentation file to use cross references throughout. I had added a couple in this PR, but was caught off-guard when asked to fix merge conflict. Where did the `AUTO-GENERATED-CONTENT` part below come from? I merged compatibly retaining both (for the moment).

boneskull: `AUTO-GENERATED-CONTENT` is [markdown-magic](https://npm.im/markdown-magic) stuff.  I was showing it's easy to use to automatically pull data out of `package.json`, knowing that we'd want to do more of it... give it a shot; you can insert the link to the mochajs homepage in the same manner as the gitter one by modifying the template (or adding a new `<!-- AUTO-GENERATED:START --> .. <!-- AUTO-GENERATED:END -->` block)

boneskull: run `npm start docs.preprocess` to apply our set modifications.  see `scripts/markdown-magic.config.js`

lib/browser/growl.js
```diff
@@ -63,23 +63,29 @@ exports.notify = function(runner) {
...
-  return permitted[Notification.permission];
+  return permitted[Notification.permission]();
 }
 
 /**
  * A promise determining if notification can proceed.
  *
  * @promise Growl~CanNotifyPromise
- * @fulfill {boolean} Notification allowed (<code>true</code>).
+ * @fulfill {undefined} Notification allowed.
```
boneskull: 1.  Does jsdoc do anything with these?  Last I saw it wasn't implemented.  Would it be problematic to add a `@returns Promise<void> Notification allowed` if the API docs don't catch this?
1.  Please mark new methods `@private` if they are not intended to be consumed externally

lib/browser/growl.js
```diff
@@ -63,23 +63,29 @@ exports.notify = function(runner) {
...
-      Notification.requestPermission().then(function(permission) {
-        return Promise.resolve(permission === 'granted');
+    granted: function allow() {
+      return Promise.resolve(true);
+    },
+    denied: function deny() {
+      return Promise.resolve(false);
+    },
+    default: function ask() {
+      return new Promise(function(resolve, reject) {
```
boneskull: the `new Promise` seems unneeded.  if you're going to eat a rejection, then just use `.catch()`

```js
default: function ask() {
  return Notification.requestPermission()
    .then(function(permission) {
      return permission === 'granted';
    });
```

lib/browser/growl.js
```diff
@@ -63,23 +63,29 @@ exports.notify = function(runner) {
  */
 function isPermitted() {
   var permitted = {
-    granted: Promise.resolve(true),
-    denied: Promise.resolve(false),
-    default: (function ask() {
-      Notification.requestPermission().then(function(permission) {
-        return Promise.resolve(permission === 'granted');
+    granted: function allow() {
+      return Promise.resolve(true);
```
boneskull: I totally spaced on noticing this, but you'll need to grab a promise shim for IE.  from what I can tell, [es6-promise](https://npm.im/es6-promise) looks ok

docs/index.md
```diff
@@ -1378,11 +1378,11 @@ tests as shown below:
     <title>Mocha</title>
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <link rel="stylesheet" href="https://unpkg.com/mocha@5.3.0/mocha.css" />
+    <link rel="stylesheet" href="https://unpkg.com/mocha@6.0.0/mocha.css" />
   </head>
   <body>
     <div id="mocha"></div>
-    <script src="https://unpkg.com/mocha@5.3.0/mocha.js"></script>
+    <script src="https://unpkg.com/mocha@6.0.0/mocha.js"></script>
```
boneskull: we can actually use markdown-magic here to insert the current version of mocha by pulling `version` out of `package.json`

plroebuck: Opened #3603 for this. Mocha wasn't using `markdown-magic` until 4 days ago.

boneskull:
@plroebuck 	I'm basically ready to cut v6.0.0-0 which will be our first prerelease.  Are you looking to get this in still?  If so, please add the milestone and I'll wait for it.

lib/browser/growl.js
plroebuck: That would go in the documentation, not the code. I don't have any means of testing old Windows browsers though.

boneskull: We support IE11 out-of-the-box and those tests run as job #11 on every build.  We don't require users of IE11 to pull in shims.

If job #11 passes--assuming there's coverage, which needs to live in `test/unit`--you're fine.

It's possible to launch these tests from a local dev environment--I'll message you on Gitter w/ the details.

docs/index.md
```diff
@@ -1487,20 +1488,22 @@ A typical setup might look something like the following, where we call `mocha.se
...
   <title>Mocha Tests</title>
-  <link href="https://unpkg.com/mocha@5.2.0/mocha.css" rel="stylesheet" />
+  <link href="https://unpkg.com/mocha@6.0.0/mocha.css" rel="stylesheet" />
 </head>
 <body>
   <div id="mocha"></div>
 
   <script src="https://unpkg.com/chai/chai.js"></script>
-  <script src="https://unpkg.com/mocha@5.2.0/mocha.js"></script>
+  <script src="https://unpkg.com/mocha@6.0.0/mocha.js"></script>
```
boneskull: if I were you I'd actually just eliminate the version altogether.  it should pull the latest one.  since we are doing a prerelease, this number will be `v6.0.0-0` anyway, and I'm pretty sure we don't want to keep changing it

docs/index.md
```diff
@@ -1547,6 +1550,69 @@ The "HTML" reporter is what you see when running Mocha in the browser.  It looks
...
+
+### Node-based notifications
+
+In order to use desktop notifications with the command-line interface (CLI),
+you **must** first install some platform-specific prerequisite software.
+Instructions for doing so can be found [here](https://github.com/mochajs/mocha/wiki/Growl-Notifications).
+
+Enable Mocha's desktop notifications as follows:
+
+```bash
```
boneskull: we need to normalize our fenced code blocks, because some of them are `sh` and some are `bash`.  should not be part of this PR--just mentioning it

lib/browser/growl.js
```diff
@@ -1,5 +1,207 @@
...
+ */
+exports.notify = function(runner) {
+  var promise = isPermitted();
+
+  /**
+   * Attempt notification.
+   */
+  var sendNotification = function() {
+    // If user hasn't responded yet... "No notification for you!" (Seinfeld)
+    Promise.race([promise, Promise.resolve(undefined)])
```
boneskull: Hey, there you have it, an actual use case for `Promise.race()`!  never seen one in the wild before.

`Promise.resolve(undefined)` is essentially the same thing as `Promise.resolve()` but with a nonempty `arguments` array.  It looks like `promise` will resolve with a boolean... maybe this should be `Promise.resolve(false)` or just `Promise.resolve()`?

not demanding this change; I'm happy to tweak it afterwards and save you the work, since you've 
 already done plenty of iteration here.

lib/browser/growl.js
boneskull: actually, you can probably avoid adding the shim if you can ensure we won't ever attempt to call `Promise.*`.  I think `isCapable()` is our escape hatch.  can you confirm?

lib/browser/growl.js
```diff
@@ -1,5 +1,207 @@
...
+ */
+function notPermitted(err) {
+  console.error('notification error:', err.message);
+}
+
+/**
+ * Used to help distinguish notifications within a root suite run.
+ * @type {number}
+ */
+var ntags = 0;
```
boneskull: literally anything that has a docstring needs `@private` or it'll show in the documentation.

we can also address this post-merge

lib/browser/growl.js
```diff
@@ -1,5 +1,207 @@
...
+ *
+ * @private
+ * @returns {string} tag unique within a root suite execution
+ * @example
+ *
+ * generateTag();
+ * // => "mocha_results-20181117-1"
+ */
+function generateTag() {
+  return ['mocha_results', datestamp(), ++ntags].join('-');
```
boneskull: if the implementation doesn't matter, why not just use `Date.now()` instead of the `datestamp()` impl below?

scripts/package-json-cullify.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
boneskull: i hate it when you want to cullify but you actually just omitify 🤷‍♂️

test/unit/mocha.spec.js
```diff
@@ -129,12 +129,26 @@ describe('Mocha', function() {
   });
 
   describe('.growl()', function() {
-    it('should set the growl option to true', function() {
-      var mocha = new Mocha(opts);
-      mocha.growl();
-      expect(mocha.options, 'to have property', 'growl', true);
+    describe('if capable of notifications', function() {
```
boneskull: OK, so, if this passes job 11, we're fine on the IE11 question.

docs/index.md
boneskull: (https://unpkg.com/mocha/mocha.js is just a redirect to the latest version)

package-scripts.js
```diff
@@ -23,7 +23,7 @@ function test(testName, mochaParams) {
 module.exports = {
   scripts: {
     build: {
-      script: `browserify ./browser-entry --plugin ./scripts/dedefine --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' > mocha.js`,
+      script: `browserify -e browser-entry.js --plugin './scripts/dedefine' --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' -o mocha.js`,
```
boneskull: the AppVeyor failure is caused by this change; you may want to revert.

docs/index.md
plroebuck: done. would have removed the version earlier, but thought it might be intentional.

lib/browser/growl.js
plroebuck: It must **not** become `Promise.resolve(false)`, as this is the third possible state (which is used when creating the error message). Think the explicit `undefined` provides better context of _where_ the third state value comes from, but whatever.

lib/browser/growl.js
plroebuck: PR has passed all CI tests since originally pushed. If your browser doesn't support the feature (which IE doesn't), `isCapable` won't enable. If I can find someone brave enough to have it installed, I'll see if adding shims for _both_ promises and notifications will allow it to work.

lib/browser/growl.js
plroebuck: Since based on msecs, it could _never_ be used to match up with previous notifications. This scheme allows for replacing previous notifications with newer ones.

package-scripts.js
plroebuck: It had worked fine for months, until I merged with master two days ago. And only fails when run on Windows.

scripts/package-json-cullify.js
plroebuck: Don't follow.

**cull**: (verb)
1. to select from a group : choose
// culled the best passages from the poet's work
// Damaged fruits are culled before the produce is shipped.
2. to reduce or control the size of (something, such as a herd) by removal (as by hunting) of especially weaker animals 
// The town issued hunting licenses in order to cull the deer population.

package-scripts.js
boneskull: it’s the path-as-string I think that’s the windows problem. remove quotes I imagine

lib/browser/growl.js
boneskull: it sounds like then, the limitation is that only results that arrive on the same *day* can be replaced?  this is fine, but please mention the intent/reasoning in the comments, b/c it sounded like it didn't need to be anything in particular.  alternatively, just merge it and we'll fix it later.

boneskull:
@plroebuck 	Are you still trying to get this into v6?  AFAIK this is the only thing we're waiting on (though we might squeeze in a couple more PRs).

docs/index.md
plroebuck: changed to `sh`

PR #3547, opened 2018-11-02
update list of published files in package.json

boneskull:
- this will fix a problem with npm (potentially) publishing certain files
which should always be ignored; see npm/npm-packlist#14.
- also fixes issue with growl images not getting published!

PRO TIP: use `npm pack --dry-run` to see what would be included in
published tarball.

### DIFF
--- a/package.json
+++ b/package.json
@@ -510,14 +510,18 @@
     "watchify": "^3.7.0"
   },
   "files": [
-    "bin",
-    "images",
-    "lib",
+    "bin/{*mocha,options.js}",
+    "assets/growl/*.png",
+    "lib/**/*.{js,html}",
     "index.js",
     "mocha.css",
     "mocha.js",
     "browser-entry.js"
   ],
+  "directories": {
+    "lib": "lib",
+    "test": "test"
+  },
   "browser": {
     "growl": "./lib/browser/growl.js",
     "tty": "./lib/browser/tty.js",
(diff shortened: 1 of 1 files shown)


plroebuck:
as long as you're here, we should be documenting:
* `directories.lib`
* `directories.test`
* `directories.doc`

boneskull:
@plroebuck specifying `directories` doesn't actually *do* anything, but might as well for metadata purposes.

package.json
```diff
@@ -510,14 +510,20 @@
     "watchify": "^3.7.0"
   },
   "files": [
-    "bin",
-    "images",
-    "lib",
+    "bin/{*mocha,options.js}",
+    "assets/growl/*.png",
+    "lib/**/*.{js,html}",
     "index.js",
```
plroebuck: Does the increase in specificity buy us anything here? Obviously, `images` is wrong (why was directory renamed to begin with?) and should be corrected. Wonder why we didn't get bug reports on that...

plroebuck: Myself included. Since Apple took away the ability to show a custom icon, I don't pay the Notification Center notices much mind anymore.

boneskull: @plroebuck The increased specificity avoids problems described in npm/npm-packlist#14.  I had published another package the other day which used `files` but wound up publishing a `.DS_Store` file... alas.

boneskull:
@plroebuck 	please see changes in 0f26b58

package.json
plroebuck: @boneskull, that's [not supposed to be possible](https://docs.npmjs.com/files/package.json#files)...

package.json
```diff
@@ -510,14 +510,18 @@
...
     "mocha.css",
     "mocha.js",
     "browser-entry.js"
   ],
+  "directories": {
+    "lib": "lib",
+    "test": "test"
```
plroebuck: Decide not to express these as explicit relative paths?
`    "lib": "./lib",`
`    "test": "./test"`

boneskull: not sure why this matters...

PR #3550, opened 2018-11-04
Warn that suites cannot return a value

### LINKED ISSUE #2975: Describe block with async function behaving weirdly

plroebuck:
### Description of the Change

As this is not supported, the behavior is confusing. It can produce a suite that is not properly initialized. Providing a deprecation warning will make it clear that this is not supported.

For example, the test suite below does nothing.

```javascript
describe("my suite", async function() {
  await Promise.resolve();
  it("does a thing", function() {
    assert(false);
  });
});
```

Other usage could possibly produce a partially-initialized test suite that fails for unknown reasons; this change will make it clear by alerting the user.

### Alternate Designs

Original proposal #3243 was to throw, but settled for a deprecation warning in the accepted version.
This PR will need to be revisited in next major release to throw an Error instead.

N/A

### Benefits

Reduction in user confusion/expectations attempting to use `describe` function as anything but a namespace.

### Possible Drawbacks

Users will get deprecation warning for unsupported behavior.
No _real_ downside to knowledge.

### Applicable issues

Fixes #2975
Closes #3243
semver-patch


plroebuck:
Approving reviews done on original PR.

Merging as is and revisit to throw error (for Mocha-7.0)

PR #3556, opened 2018-11-07
WIP: add config file support; switch to yargs

### LINKED ISSUE #3037: consider using something else instead of commander
there are a couple problems with commander:

- the upgrade from 2.9.x to 2.11.x introduced breaking changes
- we're fiddling with commander's internals in at least one place in `bin/_mocha`
- no out-of-box support for passing node flags

Ideally whatever module we consume instead should adhere more strictly to semver and allow us to consolidate the executables in `bin/`.

boneskull:
Ideally I'd like to get this into v6.0.0, since it's something people might get mildly excited about.

Prompted by too many issues to mention, I've been wanting to replace `commander` with `yargs` for a very long time.

But `yargs` makes it pretty easy to grab "arguments" out of configuration files, so it wasn't too difficult to make it part of the change.

What's done:

- [x] Replace `commander` with `yargs`
- [x] Add support for specifying arguments/options in `.mocharc.js`, `.mocharc.json`, `.mocharc.yaml`, and a `mocha` property of `package.json`.  We also get config file extension [for free](http://yargs.js.org/docs/#api-configobject-extends-keyword).
- [x] Add support for user-defined path to config file
- [x] Retain support for `mocha.opts` (`mocha.opts` is loaded first, then options are *merged* with any config files)
- [x] Update `--help` output
- [x] Refactor much CLI-specific code and options into `lib/cli/`
- [x] Existing tests pass (or they did last time I checked, anyway... surprise me, Travis)
- [x] For any boolean flag, e.g., `--foo`, `--no-foo` is *also* supported.  This means `--no-exit` is now valid (and is in fact the default behavior)
- [x] Consume `.mocharc.yml` instead of `test/mocha.opts` in Mocha's own test suites! (this necessitated moving a `mocha.opts`-specific test into its own suite)

TODO LIST:

- [x] ~~Docs and~~ examples for config files
- [x] Ensure coverage doesn't dip
- [x] Pass-through(s) for Node.js/V8 options - derive from `process.allowedNodeEnvrionmentFlags` or as-of-yet-nonexistent shim thereof (will publish such a package asap)
- [x] Ensure `.eslintrc.yml` isn't stale

Given the impact of these changes, I fully expect to see a patch soon after v6.0.0... it's bound to touch some edge cases.  These *should not be breaking changes*, but they probably are.

~~**UPDATE**: a thing that I need to figure out is this:  `bin/mocha` loads `mocha.opts`, and it allows `node` flags.  people are going to want `node` flags in config files too.  that means we need to parse config files in `bin/mocha`.  ...but we *also* don't want to parse the config files twice, just like we don't parse `mocha.opts` twice.~~

**UPDATE**: 
- [x] Ensure defaults stored in `lib/mocharc.json` work in a browser

### Applicable issues

Closes #3037
semver-minor (should be semver-major due to amount of change)


lib/cli/options.js
```diff
@@ -64,14 +64,16 @@ function parseOptions(content) {
 function getOptions() {
   if (
     process.argv.length === 3 &&
-    (process.argv[2] === '-h' || process.argv[2] === '--help')
+    ['--help', '-h', '--version', '--reporters', '--interfaces'].indexOf(
```
boneskull: short-circuit more stuff... essentially a performance fix.

lib/suite.js
```diff
@@ -131,7 +131,7 @@ Suite.prototype.timeout = function(ms) {
  * @return {Suite|number} for chaining
  */
 Suite.prototype.retries = function(n) {
-  if (!arguments.length) {
+  if (!arguments.length || typeof n === 'undefined') {
```
boneskull: this *could* break something. even though it's "private", it's called by `Mocha#retries` which is public.

package.json
```diff
@@ -461,16 +461,22 @@
   },
   "dependencies": {
     "browser-stdout": "1.3.1",
+    "chalk": "^2.4.1",
```
boneskull: need to use exact versions here

package.json
```diff
@@ -532,5 +540,8 @@
   "prettier": {
     "singleQuote": true,
     "bracketSpacing": false
+  },
+  "yargs": {
```
boneskull: see https://github.com/yargs/yargs/blob/master/docs/advanced.md#customizing-yargs-parser

test/integration/helpers.js
```diff
@@ -130,7 +130,7 @@ function invokeSubMocha(args, fn, cwd) {
 
 function _spawnMochaWithListeners(args, fn, cwd) {
   var output = '';
-  var mocha = spawn(process.execPath, args, {cwd: cwd});
+  var mocha = spawn(process.execPath, args, {cwd: cwd, env: {}});
```
boneskull: this avoids a problem when using `DEBUG`

test/integration/glob.spec.js
```diff
@@ -179,7 +179,7 @@ function execMochaWith(validate) {
       node +
         ' "' +
         path.join('..', '..', '..', '..', 'bin', 'mocha') +
-        '" -R json-stream ' +
+        '" -R json-stream --config /dev/null ' +
```
boneskull: I can't remember why I had to do this.

test/integration/hooks.spec.js
```diff
@@ -53,7 +53,7 @@ describe('hooks', function() {
   it('can fail a test in an afterEach hook', function(done) {
     helpers.runMochaJSON(
       'hooks/afterEach-hook-conditionally-fail.fixture.js',
-      args,
+      [],
```
boneskull: this was running the `dot` reporter.  that doesn't make sense with `runMochaJSON()`; it *does* make sense with `runMocha()`.

test/integration/retries.spec.js
```diff
@@ -2,56 +2,60 @@
...
 var helpers = require('./helpers');
-var args = ['--reporter', 'dot'];
+var args = [];
 var bang = require('../../lib/reporters/base').symbols.bang;
 
 describe('retries', function() {
   it('are ran in correct order', function(done) {
-    helpers.runMocha('retries/hooks.fixture.js', args, function(err, res) {
-      var lines, expected;
+    helpers.runMocha(
```
boneskull: thanks prettier

boneskull:
[here's the output of `mocha --help`](https://gist.github.com/boneskull/1babfa5c0d3c70c2826488ea5f066772)

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+ * @param {yargs.Arguments} argv - Arguments
+ * @private
+ */
+exports.handleExtensions = argv => {
+  const compilers = argv.compilers || [];
+  if (compilers.length) {
+    require('util').deprecate(() => {},
+    '"--compilers" will be removed in a future version of Mocha; see https://git.io/vdcSr for more info')();
+  }
+
```
plroebuck: Shouldn't `--compilers` be removed?

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+exports.runShowInterfaces = () => {
+  showKeys(Mocha.interfaces);
+};
+
+/**
+ * Dump list of built-in reporters
+ * @private
+ */
+exports.runShowReporters = () => {
+  showKeys(Mocha.reporters);
```
plroebuck: I see what you're doing here. But it only really shows _our_ built-in reporters. Could we provide some means to allow installed external reporters/interfaces to register themselves and show up in `--help`?

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+            }
+
+            acc[pair[0]] = pair.length === 2 ? pair[1] : true;
+            return acc;
+          }, {}),
+        description: 'Reporter-specific options (<k=v,[k1=v1,..]>)',
+        group: GROUPS.OUTPUT,
+        requiresArg: true,
+        type: 'string'
+      },
```
plroebuck: For what it's worth, this needs to be replaced with `--reporter-option` that can be used multiple times (collected). Current version doesn't allow for passing arguments with commas in the value. Given druthers, the routine would allow the value to be hierarchially-capable (#3487), beginning with the affected reporter's name. Would also simplify "commenting out" a single reporter option without having to "edit" the existing.

Example:
```bash
$ mocha --reporter xunit --reporter-option xunit.outfile="xunit.xml" --reporter-option xunit.suiteName="Surely Thou Jest" 
```
equivalent to
```bash
$ mocha --reporter xunit --reporter-options outfile="xunit.xml",suiteName="Surely Thou Jest" 
```

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+
+      validatePlugin(argv, 'reporter', Mocha.reporters);
+      validatePlugin(argv, 'ui', Mocha.interfaces);
+
+      return true;
+    });
+
+exports.handler = argv => {
+  debug('post-yargs config', argv);
+
```
plroebuck: One'n'Dones can be handled within `yargs`. See [here](https://github.com/cybertk/abao/blob/eb850ef47cbe14964c5951746f5485109cd57caf/lib/cli.coffee#L32-L60).

bin/_mocha
```diff
@@ -7,646 +7,35 @@
...
-}
-
-process.on('SIGINT', () => {
-  runner.abort();
-
-  // This is a hack:
-  // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
-  // The amount of failures will be emitted as error code later
-  runner.failures = 130;
-});
```
plroebuck: Even sweeter would be to reduce "bin/_mocha" to something like [this](https://github.com/cybertk/abao/blob/master/bin/abao)...

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+  process.stdout.write('\u001b[?25h');
+};
+
+/**
+ * Stop cursor business
+ * @private
+ */
+const stop = () => {
+  process.stdout.write('\u001b[2K');
+};
```
plroebuck: Shouldn't we create a cursor handling module (more in `Base.cursor`) and reuse it?

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+ */
+const showKeys = obj => {
+  console.log();
+  Object.keys(obj)
+    .filter(value => /^[a-z]/.test(value))
+    .sort()
+    .forEach(key => {
+      console.log(`    ${key}`);
+    });
+  console.log();
```
plroebuck: Maybe we drop `console.log` in favor of `process.stdout.write`?

lib/cli/run-helpers.js
plroebuck: Any thoughts for an `--extensions` flag to add extensions to the lookup pattern to find tests? Given the retirement of `--compilers`, it would make life easier for CoffeeScript/TypeScript users.

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+/**
+ * Used for `--reporter` and `--ui`.  Ensures there's only one, and asserts
+ * that it actually exists.
+ * @todo XXX This must get run after requires are processed, as it'll prevent
+ * interfaces from loading.
+ * @param {yargs.Arguments} argv - Arguments
+ * @param {string} option - Resolvable module name or path
+ * @param {Object} map - An object perhaps having key `option`
+ * @private
+ */
```
plroebuck: This looks interesting. Is there a corresponding `registerPlugin`?

lib/cli/init.js
```diff
@@ -0,0 +1,32 @@
...
+  yargs.positional('path', {
+    type: 'string',
+    normalize: true
+  });
+
+exports.handler = argv => {
+  const dest = argv.path;
+  const mkdir = require('mkdirp');
+  mkdir.sync(dest);
+  const css = fs.readFileSync(path.join(__dirname, '..', '..', 'mocha.css'));
```
plroebuck:
```js
  const srcdir = path.join(__dirname, '..', '..');
  const css = fs.readFileSync(path.join(srcdir, 'mocha.css'));
```

lib/cli/init.js
```diff
@@ -0,0 +1,32 @@
...
+exports.description = 'create a client-side Mocha setup at <path>';
+
+exports.builder = yargs =>
+  yargs.positional('path', {
+    type: 'string',
+    normalize: true
+  });
+
+exports.handler = argv => {
+  const dest = argv.path;
```
plroebuck:
```js
  const destdir = argv.path;
```

lib/cli/init.js
```diff
@@ -0,0 +1,32 @@
...
+  const mkdir = require('mkdirp');
+  mkdir.sync(dest);
+  const css = fs.readFileSync(path.join(__dirname, '..', '..', 'mocha.css'));
+  const js = fs.readFileSync(path.join(__dirname, '..', '..', 'mocha.js'));
+  const tmpl = fs.readFileSync(path.join(__dirname, '..', 'template.html'));
+  fs.writeFileSync(path.join(dest, 'mocha.css'), css);
```
plroebuck:
```js
  fs.writeFileSync(path.join(destdir, 'mocha.css'), css);
```

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+};
+
+/**
+ * Exits Mocha when Mocha itself has finished execution, regardless of
+ * what the tests or code under test is doing.
+ * @param {number} code - Exit code; typically # of failures
+ * @private
+ */
+const exit = code => {
+  const clampedCode = Math.min(code, 255);
```
plroebuck: `process.exitCode` should be one of:
* `0` for _success_
* `1` for _failure_ (e.g., internal error, if _any_ tests fail, no tests were run)
* `2` for _failure_ due to invalid command line usage
* `126` if "bin/_mocha" not executable
* `128+n` where `n` is signal# (e.g., `SIGINT`=2, so rc=`130`)

This should be _explicitly_ added to the documentation.

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+      config: {
+        config: true,
+        default: CONFIG_PATH_PLACEHOLDER,
+        defaultDescription: '(nearest rc file)',
+        description: 'Load config from file',
+        global: false,
+        group: GROUPS.CONFIG
+      },
+      debug: {
+        alias: 'd',
```
plroebuck: Can we drop the `d` alias?

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+  if (argv.interfaces) {
+    debug('interfaces:');
+    runShowInterfaces();
+    process.exit();
+  }
+  if (argv.reporters) {
+    debug('reporters:');
+    runShowReporters();
+    process.exit();
+  }
```
plroebuck: These could be handled by `yargs.check`.

lib/suite.js
plroebuck: Think this change is incorrect. If we _were_ passed anything then it was meant for the mutator portion of the method. 

In this case, we should verify we were passed a number and `floor` it.
* if it's not the right type, then `TypeError` code=`ERR_INVALID_ARG_TYPE`
* if right type but bad value (e.g., `Number.NaN`), `TypeError` code= `ERR_INVALID_ARG_VALUE`.

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+  RULES: 'Rules & Behavior',
+  CONFIG: 'Config Files'
+};
+
+/**
+ * Placeholder value to force config parsing function to run.
+ */
+const CONFIG_PATH_PLACEHOLDER = '__MOCHA_CONFIG_PATH__';
+
+exports.command = '$0 [spec..]';
```
plroebuck:
```js
const prog = path.basename(require('../package').bin.mocha);
exports.command = `Usage: ${prog} [options] [spec..]`;
```

If you use `$0`, then path to binary becomes part of usage (indeterminate).
Also, does it need to document `mocha debug`?

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+        group: GROUPS.FILES,
+        type: 'boolean'
+      }
+    })
+    .positional('spec', {
+      default: ['test/'],
+      defaultDescription: 'test',
+      description: 'One or more files, directories, or globs to test',
+      type: 'array'
+    })
```
plroebuck: Why doesn't positional show up last?

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"'
+        );
+      }
+
+      validatePlugin(argv, 'reporter', Mocha.reporters);
+      validatePlugin(argv, 'ui', Mocha.interfaces);
+
+      return true;
+    });
+
```
plroebuck:
```
    .wrap(80)
    .help('help', 'Show usage information and exit')
    .version('version', 'Show version number and exit', pkg.version)
    .epilog(`Website:\n  ${pkg.homepage}`)
```

package.json
```diff
@@ -461,16 +461,22 @@
...
+    "findup-sync": "^2.0.0",
     "glob": "7.1.2",
     "growl": "1.10.5",
     "he": "1.1.1",
+    "js-yaml": "^3.12.0",
+    "log-symbols": "^2.2.0",
     "minimatch": "3.0.4",
     "mkdirp": "0.5.1",
-    "supports-color": "5.4.0"
+    "strip-json-comments": "^2.0.1",
```
plroebuck: Since you're adding dependencies anyway, how about we drop our out-of-date internal version of `ms` and use the npm package instead?

test/integration/glob.spec.js
plroebuck: But would be nice to know...

test/integration/options.spec.js
```diff
@@ -433,7 +433,7 @@ describe('options', function() {
           if (error) {
             return done(error);
           }
-          expect(result.output, 'to contain', 'Usage:');
+          expect(result.output, 'to contain', 'Run tests with Mocha');
```
plroebuck: Would prefer to revert this one. The "Usage:" part _should be there_. Fixed above.

test/integration/helpers.js
plroebuck: But aren't there _other_ envvars that we _do_ need?

test/integration/retries.spec.js
plroebuck: LOL. So it doesn't _just_ happen to me after all...

test/node-unit/cli/config.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+        it('should use the JSON parser', function() {
+          loadConfig('foo.json');
+          expect(parsers.json, 'to have calls satisfying', [
+            {args: [filepath], returned: config}
+          ]).and('was called times', 1);
+        });
+      });
+    });
+
```
plroebuck: How about an identity function check (i.e., does YAML give back same as JSON)?
Save result of each and add `deep.equal` assertion.

lib/cli/options.js
plroebuck: `-V` missing

boneskull: - [ ] add -V

lib/suite.js
boneskull: > Think this change is incorrect. If we were passed anything then it was meant for the mutator portion of the method.

That's what the code does... but I'm not convinced that was the intent.  The API is inconsistent for setter/getter methods across the codebase, which is problematic.  Some methods check for `arguments.length`, others check if the first parameter is `undefined`, and probably other strategies or combinations thereof.

The use case is:

```js
suite.retries(foo.bar)
```

Where Object `foo` *has no property* `bar`.

I'll just hack around this elsewhere.  The API needs to be made consistent, and that means settling on a strategy.  I don't love setters/getters *anyway*, so maybe that strategy is "remove them where possible."

- [ ] revert and fix elsewhere

test/integration/helpers.js
boneskull: > But aren't there other envvars that we do or might need?

It's *possible* we *might* need this in the future, but I can't think of any reason.  

If we can get away with ignoring `STDERR`, that's a better fix.  I'll try it.

- [ ] see if we can ignore `STDERR` entirely when spawning for integration tests

test/integration/glob.spec.js
boneskull: - [ ] see if I can't revert this; if I can't, then explain why

test/integration/retries.spec.js
boneskull: "works as intended"

lib/cli/run-helpers.js
boneskull: `--extensions` is essentially an undocumented flag at this point, so that's probably a good idea.

I have FUD around removing `--compilers`.  But you're right, we should probably do that.

- [ ] expose `--extensions`
- [ ] hard-deprecate `--compilers`

lib/cli/run-helpers.js
boneskull: - [ ] do this

I thought about this but decided to just short-circuit execution instead.  I suppose it could be helpful to show external reporters & interfaces.  That means processing requires as the first step.

lib/cli/run.js
boneskull: Agree. This is pretty trivial since yargs does a bunch of this for us.  Let's soft-deprecate `--reporter-options`. Sound good?

- [ ] implement `--reporter-option`; soft-deprecate `--reporter-options`

We could go a step further and soft-deprecate anything else that wants a comma-separated parameter..

lib/cli/run.js
boneskull: I don't understand.

bin/_mocha
boneskull: - [ ] destroy `_mocha` more

lib/cli/run-helpers.js
boneskull: browser doesn't need it.

lib/cli/run-helpers.js
boneskull: What's the reasoning?

lib/cli/run-helpers.js
boneskull: Maybe "Plugin" is a misnomer.

A "Plugin" in this case is a module which, when `--require`'d, adds itself to `Mocha.interfaces` or `Mocha.reporter`.

We could provide a nice API to do that sort of thing, but there's already plenty of scope creep happening

lib/cli/init.js
boneskull: - [ ] rename vars

lib/cli/init.js
boneskull: - [ ] rename more vars

bin/_mocha
```diff
@@ -7,646 +7,35 @@
...
+  .updateStrings({
+    'Positionals:': 'Positional Arguments',
+    'Options:': 'Other Options'
+  })
+  .fail((msg, err, yargs) => {
+    yargs.showHelp();
+    console.error(`\n${symbols.error} ${chalk.red('ERROR:')} ${msg}`);
+  })
+  .epilog(`Read the docs at ${chalk.yellow('https://mochajs.org')}`)
+  .parse();
```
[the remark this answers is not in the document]
boneskull: I don't understand where to draw the line.

lib/cli/run-helpers.js
boneskull: if you want to argue for that, please create an issue, but this would break certain CI systems which rely on exit code abuse.

lib/cli/run.js
boneskull: I realize also that `debug` (not `--debug`, but does same thing) is a valid command which I failed to implement.

- [ ] drop `-d` alias for `--debug`
- [ ] note that `debug` is no longer supported somewhere

lib/cli/run.js
boneskull: `$0` is set to `mocha` via `yargs.scriptName()`

see above about `debug`.

lib/cli/run.js
boneskull: In the help output?  The ungrouped options always show last...  I don't know if we have control over that.

lib/cli/run.js
boneskull: - [ ] do this

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+exports.command = '$0 [spec..]';
+
+exports.describe = 'Run tests with Mocha';
+
+exports.builder = yargs =>
+  yargs
+    .showHidden(
+      'show-node-options',
+      'Show supported Node.js & V8 options; use with "--help"'
```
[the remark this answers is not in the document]
boneskull: - [ ] do this

package.json
boneskull: seems tangential, but a good idea.  see #3557

test/integration/options.spec.js
boneskull: - [ ] do this

test/node-unit/cli/config.spec.js
boneskull: there are plenty of tests that need writing, but this is a good idea.

- [ ] do this

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+exports.list = str =>
+  Array.isArray(str) ? exports.list(str.join(',')) : (str || '').split(/ *, */);
+
+/**
+ * Dump list of built-in interfaces
+ * @private
+ */
+exports.runShowInterfaces = () => {
+  showKeys(Mocha.interfaces);
+};
```
plroebuck: So we lose the reporter and interface descriptions?

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+exports.handleRequires = argv => {
+  argv.require.forEach(mod => {
+    let modpath = mod;
+    if (fs.existsSync(mod, {cwd}) || fs.existsSync(`${mod}.js`, {cwd})) {
+      modpath = path.resolve(mod);
+      debug(`resolved ${mod} to ${modpath}`);
+    }
+    require(modpath);
+  });
+};
```
plroebuck: For my money, I'd prefer these `handleXXX` methods **not** be written in terms of `yargs` implementation.

```js
/*
 *  @param {string[]} modules - Modules to be `require`d.
 */
exports.handleRequires = modules => {
  modules.forEach(mod => {
```
then
```js
  handleRequires(argv.require);
```

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+ * Actually run tests
+ * @param {Mocha} mocha - Mocha instance
+ * @param {yargs.Arguments} argv - Arguments
+ * @param {string[]} files - Array of test files
+ * @private
+ */
+exports.runMocha = (mocha, argv, files) => {
+  let runner;
+
+  if (argv.watch) {
```
plroebuck: Contemplate exporting the watchable `runMocha`?

test/integration/glob.spec.js
boneskull: The reason is that our config requires `test/setup.js`, relative from the root, but the cwd used is `test/integration/fixtures/glob`, so `test/setup.js` won't be found.

lib/cli/run-helpers.js
boneskull: yeah, I could split this up, I think.

- [ ] split up `runMocha`

lib/cli/run-helpers.js
boneskull: Hmm, yeah, the reporters had descriptions, but not the interfaces.  Should probably keep them.  That's probably going to be a `description` property on each reporter class...

boneskull: - [ ] restore descriptions

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+    debug('reporters:');
+    runShowReporters();
+    process.exit();
+  }
+
+  // load requires first, because it can impact subsequent operations.
+  handleRequires(argv);
+
+  const mocha = new Mocha();
+
```
plroebuck: Why are we not using a sensible defaults `options` object with the ctor?

lib/cli/run-helpers.js
boneskull: ah, hmm.  I don't feel strongly about it either way, ok.

- [ ] don't pass `argv` around here

lib/cli/run.js
boneskull: It seems like there should be a reason, right?

If we're going to change it, I'd prefer to understand why it was done this way, even if the answer is "no particular reason."  Care to peruse the Git history?

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+    .reporter(argv.reporter, argv.reporterOptions)
+    .retries(argv.retries)
+    .slow(argv.slow)
+    .timeout(argv.timeout)
+    .ui(argv.ui)
+    .useColors(argv.colors)
+    .useInlineDiffs(argv.inlineDiffs);
+
+  // pesky inverted options
+  mocha.hideDiff(!argv.diff).enableTimeouts(!argv.noTimeouts);
```
plroebuck: We should add `showDiff` method and `timeouts (default: true)` option to allow us to use positive logic.

lib/cli/run-helpers.js
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * Actually run tests
+ * @param {Mocha} mocha - Mocha instance
+ * @param {yargs.Arguments} argv - Arguments
+ * @param {string[]} files - Array of test files
+ * @private
+ */
+exports.runMocha = (mocha, argv, files) => {
+  let runner;
```
plroebuck: Want to consider adding requested changes from #1810 here? Or is it just too much with all the rest of this?

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
+'use strict';
+
+/**
+ * Command module for default ("run") command
+ */
+
+const path = require('path');
+const {loadConfig, findConfig} = require('./config');
```
plroebuck: This aspect of `prettier` drives me nuts. When doing multi-imports via `require`, I _always_ want them listed vertically (like the L11-L21 version).

lib/mocha.js
```diff
@@ -330,6 +330,9 @@ Mocha.prototype._growl = function(runner, reporter) {
  * mocha.fgrep('foo.');
  */
 Mocha.prototype.fgrep = function(str) {
+  if (!str) {
+    return this;
```
plroebuck: or throw `TypeError`

plroebuck:
Is the go-forward recommendation:
* to configure via ".mocharc.yaml" 
* use "mocha.opts" for Node flags and such?

Think it would be too much to have a reporter config file as well?

test/opts/opts.spec.js
```diff
@@ -0,0 +1,21 @@
...
+
+var fs = require('fs');
+var path = require('path');
+
+// as of v6.0.0 we no longer use `mocha.opts` by default, which means
+// we must provide `--opts` on command line to run this test file
+describe('mocha.opts', function() {
+  it('should not duplicate mocha.opts args in process.argv', function() {
+    var processArgv = process.argv.join('');
+    var mochaOpts = fs
```
plroebuck: Uh, couldn't you have left this in "regressions.spec.js" and passed `--opts` manually via `process.argv`? For that matter (unless I'm misunderstanding this test), shouldn't we be able to manually set `argv`, invoke `getOptions`, and `expect` as is?

lib/cli/run.js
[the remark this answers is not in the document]
boneskull: Actually, no, "implies" means "fail unless this other thing is present", which doesn't work like I'd think it should work.

What I want is "if this is set, set this other thing too"... 

Essentially what I'll do is make `--show-node-options` invoke help (with hidden opts shown)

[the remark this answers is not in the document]
boneskull: otherwise:

```plain
✖ ERROR: Implications failed:
 show-node-options -> help
```

That's not very helpful.

test/integration/options.spec.js
boneskull: This doesn't make sense in yargs:

```plain
mocha [options..] [spec..]

Usage:

Commands
  mocha [options..] [spec..]     Usage:                                [default]
  mocha init <path>              create a client-side Mocha setup at <path>
```

Ideally we'd have a bit more control over how and where the default command displays.  We can ask yargs to do something about it (or propose something) but IMO it shouldn't hold this up.

lib/cli/run-helpers.js
boneskull: this also implies that external reporters could use such a thing

test/opts/opts.spec.js
boneskull: No, because the test relies on `--opts` being set when both the test AND fixture are run.  We don't use it on the test any longer (by default), so we need to run it special-like.

lib/mocha.js
boneskull: maybe, but hard to justify breaking it for that reason.

lib/cli/run.js
```diff
@@ -0,0 +1,403 @@
...
+    'forbidOnly',
+    'forbidPending',
+    'fullTrace',
+    'growl',
+    'invert'
+  ].forEach(arg => {
+    if (argv[arg]) {
+      mocha[arg]();
+    }
+  });
```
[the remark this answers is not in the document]
boneskull: I would be interested in that, but maybe not for another major.  this is going to be disruptive enough and I don't want to break half of the API surface as well

lib/cli/run-helpers.js
boneskull: did the best I could with this

lib/cli/run.js
plroebuck: one'n'done: cmdline option treated like `--version`. If option present, execute its function and terminate. For Mocha, `--reporters` and `--interfaces`.

bin/_mocha
plroebuck: Well, if gets down to just a couple lines, all coverage is basically tracked within library.

lib/cli/run-helpers.js
plroebuck: True, but both CLI and Base reporter do... they could share it, and stub it in "browserify"

lib/cli/run-helpers.js
plroebuck: All output uses same stream, not same underlying implementation. It's easier to mentally track given the routines directly above are using `process.stdout.write`.

lib/cli/run-helpers.js
plroebuck: `mocha.use(myReporter)`

lib/cli/run-helpers.js
plroebuck: Done. #3559

lib/cli/run.js
plroebuck: Guess that's a new feature...

plroebuck:
```js
exports.command = `Usage: $0 [options] [spec..]`;
```

lib/cli/run.js
plroebuck: Shove them in a group?

lib/cli/run.js
boneskull: agree, but tangential

lib/cli/run.js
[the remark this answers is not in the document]
plroebuck: Yeah, their `implies` error message sucks. If you can make it invoke help, that's all I was after anyway.

test/integration/options.spec.js
plroebuck: @bcoe, @nexdrew, how do we do this?

lib/cli/run-helpers.js
plroebuck: mais bien sûr!

We should add descriptions to the interfaces as well, even if it's only expanding the acronym

lib/cli/run-helpers.js
plroebuck: Easier to test in isolation.

lib/cli/run.js
boneskull: you mean put `spec` in a group, e.g., "File Handling"?

lib/cli/run-helpers.js
boneskull: let's deal w/ that later

lib/cli/run-helpers.js
boneskull: We tried something like this before, but there were problems, because `console.log()` actually does a hell of a lot more than `process.stdout.write()` (including trapping certain exception classes).  I agree about consistency, but it's just one of those things I'd rather not do without a stronger reason b/c of my experience w/ it.

There's actually some cool stuff in `console` (Node.js v8.5+) that we could be using...

lib/cli/run-helpers.js
boneskull: feel free to make refactors... might be reasonable to extract the fiddly cursor stuff, or pull a module down from npm that does it better.  I'm going to pass for this PR though

lib/cli/run-helpers.js
boneskull: @plroebuck I'm unsure how to describe `--extensions`.  Ideas?

lib/cli/run.js
boneskull: ok, so in `check()`, essentially.  

- [ ] do this

lib/cli/run-helpers.js
boneskull: see d7f5d1c3]

lib/interfaces/bdd.js
```diff
@@ -112,3 +112,5 @@ module.exports = function bddInterface(suite) {
     };
   });
 };
+
+module.exports.description = '[default] BDD or RSpec style';
```
plroebuck: `module.exports.description = 'BDD or RSpec style [default]';`

lib/ms.js
```diff
@@ -17,7 +17,7 @@ var y = d * 365.25;
  *
  * @memberof Mocha
  * @public
- * @api public
+ * @public
```
plroebuck: delete this duplicate line

plroebuck: For what it's worth, since you're now pulling the "shitcan the @api" task into this PR, to effectively make use of the API documentation, it will need be generated with `--private` cmdline arg to `jsdoc`.

lib/interfaces/exports.js
```diff
@@ -56,3 +56,5 @@ module.exports = function(suite) {
     }
   }
 };
+
+module.exports.description = 'Node.js module ("exports") style';
```
plroebuck: Not a fan of this change.

Would not a better approach to have been to export a `meta` object which included a `description` property? That way we could extend it. For example, reporters could annotate their stream compatibility (e.g.,`isBrowserOnly`, `isFileCapable`).

plroebuck:
Tried inline a dozen times, but it wouldn't take comment:

["lib/cli/run.js#L379-L414"](https://github.com/mochajs/mocha/blob/a7b587dc142faede788d92a4c2cec3323b1568ff/lib/cli/run.js#L379-L414)
Why is this _reimplementing_ the ctor externally?

lib/reporters/base.js
```diff
@@ -105,7 +105,7 @@ if (process.platform === 'win32') {
  * @param {string} type
```
plroebuck: As long as you're at it, we should remove the `MOCHA_COLORS` environment variable  
[here](https://github.com/mochajs/mocha/blob/a7b587dc142faede788d92a4c2cec3323b1568ff/lib/reporters/base.js#L46). Seems to be yet another anachronism used to "solve" the same problem as `FORCE_COLOR`.

lib/interfaces/bdd.js
boneskull: heh, I had it that way then changed it...

- [ ] revert

lib/ms.js
boneskull: whoops, thought I got the dupes.  

- [ ] zap dupes

boneskull: but I don't actually *want* docs about Mocha's private functions on its public API site.  in other words stuff marked `@private` isn't part of Mocha's API.

It's also kind of a way to cover our asses so someone doesn't see something in the docs and think they can or should use it.

lib/interfaces/exports.js
boneskull: I don't think I understand why this is problematic.  There's no reason we can't  iterate over keys or add arbitrary stuff to `exports`?

The point of this change was to restore the descriptions.  I don't really want to get into what a nice plugin convention should look like...

lib/reporters/base.js
boneskull: hmm, OK.

- [ ] remove `MOCHA_COLORS`

boneskull:
I'll just use the constructor.

lib/cli/run.js
boneskull: I'm not going to do this right now.  It should get done though.

lib/cli/run.js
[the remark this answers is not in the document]
plroebuck: Why would it break anything?

These could get a boolean argument that defaults `true` (to retain compatibility), but _could_ be used to explicitly disable it:
* `Mocha#allowUncaught()`
* `Mocha#asyncOnly()`
* `Mocha#forbidOnly()`
* `Mocha#forbidPending()`
* `Mocha#fullTrace()`
* `Mocha#growl()`
* `Mocha#invert()`    # could have just been a 2nd argument to `Mocha#{f}grep`

This should become wrapper routine for a replacement accepting boolean:
* `Mocha#noHighlighting` wraps [new method] `Mocha#highlighting(bool)`

lib/mocha.js
plroebuck: Disagree. I think all _public_ Mocha instance methods should throw when given bad input. Otherwise, no one knows something didn't work because of bad input.

lib/ms.js
plroebuck: Private routines are explicitly labeled _private_ in JSDoc output. 
We could always add an `.onclick()` handler to [warn them](https://www.youtube.com/watch?v=RG0ochx16Dg)...
Since API documentation includes the source code, it's not really that hidden to begin with.

But I understand your reasoning...

lib/interfaces/exports.js
plroebuck: Made comment here, but was really more concerning reporters.

I have a NRFPT reporter PR where I used an object to do the same things I've explained. While I didn't need `description` in mine, but integrating with this PR would require unnecessary rewrite upon merge for whoever gets there last.

It's not that big a change now to use an object property vs. string (e.g., `meta.description` vs. `description`).

lib/mocha.js
```diff
@@ -95,29 +96,47 @@ function Mocha(options) {
...
+  // root suite
   this.suite = new exports.Suite('', new exports.Context());
-  this.ui(options.ui);
-  this.bail(options.bail);
-  this.reporter(options.reporter, options.reporterOptions);
-  if (typeof options.timeout !== 'undefined' && options.timeout !== null) {
-    this.timeout(options.timeout);
+
+  this.grep(options.grep)
+    .fgrep(options.fgrep)
```
plroebuck: `Mocha#grep` and `Mocha#fgrep` are really an "either or" choice.
`options.fgrep` has priority.

bin/mocha
```diff
@@ -7,10 +7,9 @@
  * when found, before invoking the "real" _mocha(1) executable.
  */
 
-const spawn = require('child_process').spawn;
-const path = require('path');
-const getOptions = require('./options');
-const args = [path.join(__dirname, '_mocha')];
+const {spawn} = require('child_process');
+const getOptions = require('../lib/cli/options');
+const args = [require.resolve('./_mocha')];
```
plroebuck: or `require('./package').bin._mocha`

lib/cli/config.js
```diff
@@ -0,0 +1,73 @@
...
+  const ext = path.extname(filepath);
+  let config = {};
+  try {
+    if (/\.ya?ml/.test(ext)) {
+      config = parsers.yaml(filepath);
+    } else if (ext === '.json') {
+      config = parsers.json(filepath);
+    } else if (ext === '.js') {
+      config = parsers.js(filepath);
+    }
```
plroebuck: Not sure if `yargs` already provides but, could be handy from `debug` PoV to have means to trace what pathnames _were_ tried but didn't exist.

package.json
```diff
@@ -545,5 +554,9 @@
   "prettier": {
     "singleQuote": true,
     "bracketSpacing": false
-  }
+  },
+  "yargs": {
+    "combine-arrays": true
```
plroebuck: don't see `combine-arrays` listed as a parser option.

lib/cli/run.js
```diff
@@ -0,0 +1,392 @@
...
+        description: 'Disable color output',
+        group: GROUPS.OUTPUT,
+        type: 'boolean'
+      },
+      'no-timeouts': {
+        conflicts: ['timeout'],
+        description: 'Disable timeouts',
+        group: GROUPS.RULES,
+        type: 'boolean'
+      },
```
plroebuck: Had no problems with either of the two above and the default handling of options beginning with `--no-` (`yargs-parser` default settings)?

lib/cli/run.js
```diff
@@ -0,0 +1,392 @@
...
+      }
+
+      // yargs.implies() isn't flexible enough to handle this
+      if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
+        throw new Error(
+          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"'
+        );
+      }
+
+      if (argv.compilers) {
```
plroebuck: Do you need an unseen group `DEPRECATED` for retired options? How would `compilers` end up as an `argv` argument if it doesn't exist first?

lib/cli/run.js
```diff
@@ -0,0 +1,392 @@
...
+      validatePlugin(argv, 'ui', Mocha.interfaces);
+
+      // post-process options
+      if (argv.noTimeouts) {
+        argv.enableTimeouts = !argv.noTimeouts;
+      }
+
+      if (argv.retries === 0) {
+        delete argv.retries;
+      }
```
plroebuck: Don't we still need to initialize this value in ctor?

lib/cli/run.js
```diff
@@ -0,0 +1,392 @@
...
+
+      // load requires first, because it can impact "plugin" validation
+      handleRequires(argv.require);
+      validatePlugin(argv, 'reporter', Mocha.reporters);
+      validatePlugin(argv, 'ui', Mocha.interfaces);
+
+      // post-process options
+      if (argv.noTimeouts) {
+        argv.enableTimeouts = !argv.noTimeouts;
+      }
```
plroebuck: `delete argv.noTimeouts`?

lib/cli/run.js
```diff
@@ -368,50 +368,22 @@ exports.builder = yargs =>
...
-    if (argv[arg]) {
-      mocha[arg]();
-    }
-  });
-
-  // more pain: retries() is nonsensical
-  if ('retries' in argv) {
-    mocha.retries(argv.retries);
-  }
+  const mocha = new Mocha(argv);
```
plroebuck: Shouldn't you split the `argv` object up into relevant parts before use? And why would `runMocha` need `argv`?
<https://github.com/mochajs/mocha/blob/3f046723e7171c0646a065315cbef078fb03617f/lib/cli/run.js#L383-L392>

lib/mocha.js
```diff
@@ -95,29 +96,47 @@ function Mocha(options) {
...
+    .ui(options.ui || 'bdd')
+    .bail(options.bail)
+    .reporter(options.reporter, options.reporterOptions)
+    .useColors(options.useColors)
+    .timeout(options.timeout || 200)
```
plroebuck: Isn't the `timeout` default value part of the `yargs` setup already.

lib/mocha.js
```diff
@@ -95,29 +96,47 @@ function Mocha(options) {
...
+    .slow(options.slow || 75)
```
plroebuck: same as above

lib/mocha.js
boneskull: this would assume the only thing invoking the constructor is Yargs...

boneskull: ...but that suggests the default values should exist in a single place, and whatever needs them should pull them out.

A "default" JSON config file would make sense, IMO; we'd pull this in when constructed, and it could be inlined when bundling for the browser.

lib/cli/config.js
boneskull: good idea.

- [ ] add debugs here

package.json
boneskull: - [ ] audit [yargs-parser](https://npm.im/yargs-parser) options

lib/cli/run.js
boneskull: hm, the only reason this was here was because it was in the help output before.

it'd probably be better to remove stuff like this and write something about how all boolean options can be prefixed with `--no-` to disable them.

- [ ] do that.

lib/cli/run.js
plroebuck: Well, what the Mocha ctor takes as an argument is a plain object containing properties. Not what you gave it. Same with `handleFiles`, which has no need of `argv` save its positional arguments.

_actually_

lib/cli/run-helpers.js
plroebuck: --extensions <exts> : Add extensions for files to be tested for alternative languages.

boneskull:
yeah I was thinking about doing a prerelease.

I have significant changes yet to be pushed to this branch... there's more here than I originally thought, given that `mocha.opts` supports `node` flags, and it only makes sense that config files should as well.  

Currently running into an issue with how `yargs-parser` treats repeated options (e.g., `--require foo --require bar`); a `--require` on the command line will overwrite a `require` in a config file, which isn't what we want--we want to merge all `require`s into a single array.  Will probably need to send a PR to `yargs-parser` to address it.

I could yank out the config file support, but I really don't want to.

lib/cli/config.js
boneskull: this is actually handled by `findup-sync`.  there isn't any place we could hook in to.

lib/mocha.js
boneskull: this is true, but probably undocumented and untested.

- [ ] make these conflict

docs/index.md
```diff
@@ -1291,13 +1291,69 @@ The "HTML" reporter is what you see when running Mocha in the browser.  It looks
 
 [Mochawesome](https://www.npmjs.com/package/mochawesome) is a great alternative to the default HTML reporter.
 
+## Configuring Mocha (Node.js)
```
bcoe: this is a really cool feature, and I'm excited you're leaning on some of yargs' functionality for it.

Bamieh:
Closes #2578

lib/cli/run.js
boneskull: FWIW `handleFiles` needs quite a bit more than the positional args:

```js
/**
 * Smash together an array of test files in the correct order
 * @param {Object} [opts] - Options
 * @param {string[]} [opts.extension] - File extensions to use
 * @param {string[]} [opts.spec] - Files, dirs, globs to run
 * @param {string[]} [opts.exclude] - Files, dirs, globs to exclude
 * @param {boolean} [opts.recursive=false] - Find files recursively
 * @param {boolean} [opts.sort=false] - Sort test files
 * @returns {string[]} List of files to test
 * @private
 */
```

> I did this same transition a year ago. I learned it becomes a PITA to simulate stuff generating fake yargs structures for testing instead of just pruning down and passing functions what they really required (and no more).

I still don't quite understand.  

They don't actually have to *be* `yargs.Arguments` objects, that's true.  There's nothing really special about this object other than it has a convention for positional arguments--which we're not consuming anywhere (see `spec` in the docstring above).

What it *sounds* like you're suggesting is extracting all of the properties that are relevant to each function and passing *only those* instead of the whole thing.  If, for some reason, we needed a new option passed to one of these functions, we'd have it already.  If we didn't, as suggested, it'd create churn in the code that filters properties...

What problems were you encountering in your tests?

plroebuck:
I'll make the unpopular comment here. This PR is simply "too beaucoup" -- it's really like 10 different PRs all rolled into one. It changes 88 files thus far, and we will have no way of knowing how many fixes were actually incorporated since everything in the PR queue will possibly no longer match to anything. Change history will basically be rendered useless afterwards. It should be broken back out into digestable pieces. No reviewer can possibly do due diligence against this many files...

boneskull:
@plroebuck 

> I'll make the unpopular comment here. This PR is simply "too beaucoup" -- it's really like 10 different PRs all rolled into one. It changes 88 files thus far, and we will have no way of knowing how many fixes were actually incorporated since everything in the PR queue will possibly no longer match to anything. Change history will basically be rendered useless afterwards. It should be broken back out into digestable pieces. No reviewer can possibly do due diligence against this many (unrelated) changes in that many files...

I don't disagree.  I'm happy to break it up into multiple logical commits after I've knocked out what's left to do.

I should be able to pull some of it out, as well.  Let's hold off on further review until I'm ready

boneskull:
Barring any needed fixes for CI failures, this is now ready to review...

I pulled several PRs out of this, but there's still many more changed files than I would have liked.  Many of these changes are test fixes related to version upgrades (Karma) or needed due to removing our `test/mocha.opts` and thus eliminating some assumptions the tests were making.  In a couple places I've sped up integration tests by invoking `bin/_mocha` directly instead of `bin/mocha`.

Even small modifications to files require a complete reformat, which is unfortunate, but would happen eventually.

The commit message has a better summary of changes, which I'll paste:

* * *

use yargs for option parsing and support config files

## BREAKING CHANGES

- `-d` is no longer an alias for `--debug`
- `--grep` and `--fgrep` are now mutually exclusive
- The option formerly known as `--compilers` is now removed from Mocha
- `lib/template.html` moved to `lib/browser/template.html`
- An error is thrown if a path to `mocha.opts` is specified by the user and the

## Soft Deprecations

### CLI

- `-gc` users should use `--gc-global`

### Programmatic API

- Public API method `getOptions()` in `bin/options.js` is deprecated and currently has no alternative
- Users of the `enableTimeouts` option of the `Mocha` constructor should use `timeout` instead.  Specify `false` for no timeout
- Users of the `useColors` option of the `Mocha` constructor should use `color` instead

## Enhancements

- Mocha now supports JS, JSON, YAML, or `package.json`-based config.
- Any/all configuration, including `mocha.opts` can be used in addition to command-line arguments, and everything will be merged as applicable
- Node/V8 flag support:
  - Support of all available `node` flags on a *per-version* basis
  - Support of any V8 flag by prepending `--v8-` to the option
  - These are supported within `mocha.opts` or config files
- More flexible command-line parsing including negation of any boolean flag by prepending `--no-` to the name
- Better `--help`
- Descriptions of interfaces in `--help` text
- A useful `Mocha` constructor
- Debug-related flags (e.g., `--inspect`) now *imply* `--no-timeouts`

## Fixes

- Many bug fixes around CLI option handling, e.g., closes #3475
- Fixes #3363, #2576
- `--no-timeouts` works

## Documentation

- Added new, updated, or rewrote documentation for all non-self-explanatory options
- Updated usage text, TOC
- Added section on configuration
- Added example configuration files in `example/config/`

## Etc.

- Updated many dev deps, mostly within existing semver ranges
- Removed `commander`
- Added production deps:
  - ansi-colors - color terminal output
  - findup-sync - find package.json and other config files
  - js-yaml - parse YAML
  - log-symbols - visual symbols for success/fail etc.
  - node-environment-flags - per-version allowed Node.js env flags
  - object.assign - for IE11
  - strip-json-comments - allow comments in JSON config files
  - wide-align - terminal formatting
  - yargs - option parser
  - yargs-parser - for compatible, lightweight parsing of config files
  - yargs-unparser - for passing config to `bin/_mocha` from `bin/mocha`

lib/cli/run.js
boneskull: I think I meant `check()`. but this is done.

docs/index.md
```diff
@@ -902,73 +937,209 @@ To ensure your tests aren't leaving messes around, here are some ideas to get st
...
+> *Updated in v6.0.0.  Previously `--watch-extensions`, but now expanded to affect general test file loading behavior. `--watch-extensions` is now an alias*
 
-### `-t, --timeout <ms>`
+Files having this extension will be considered test files.  Defaults to `js`.
 
-Specifies the test-case timeout, defaulting to 2 seconds. To override you may pass the timeout in milliseconds, or a value with the `s` suffix, ex: `--timeout 2s` or `--timeout 2000` would be equivalent.
+Affects `--watch` behavior.
 
-### `--no-timeouts`
+Specifying `--extension` will *remove* `.js` as a test file extension; use `--extension js` to re-add it.  For example, to load `.mjs` and `.js` test files, you must supply `--extension mjs --extension js`.
```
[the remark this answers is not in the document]
boneskull: Yeah, not sure why I didn't do this--it's a breaking change--but I'll enable it and add test.

- [ ] (re-)enable comma-separated `extension` and add a test

bin/options.js
```diff
@@ -1,88 +1,5 @@
...
-    process.argv = process.argv
-      .slice(0, 2)
-      .concat(opts.concat(process.argv.slice(2)));
-  } catch (ignore) {
-    // NOTE: should console.error() and throw the error
-  }
-
-  process.env.LOADED_MOCHA_OPTS = true;
-}
+module.exports = require('../lib/cli/options');
```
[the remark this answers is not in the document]
boneskull: I know it's being used by at least one dependency, so I didn't want to kill it outright.

Yeah, it needs to call `utils.deprecate`, which I overlooked.

- [ ] call `deprecate()` when `getOptions` used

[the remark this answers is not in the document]
boneskull: looks like I already did that.  see L120 of `lib/cli/options.js`

boneskull:
I decided *not* to remove `MOCHA_FORCE_COLOR` since we don't seem to have a workaround atm.

boneskull:
I'm going to merge this very soon if there are no further reviews. cc @mochajs/core

plroebuck:
You merged this with a single approved review?

docs/index.md
```diff
@@ -809,88 +810,120 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
...
+--help, -h     Show usage information & exit                         [boolean]
+--version, -V  Show version number & exit                            [boolean]
+--interfaces   List built-in user interfaces & exit
+--reporters    List built-in reporters & exit
+
+Mocha Resources
+Chat: https://gitter.im/mochajs/mocha/
+GitHub: https://github.com/mochajs/mocha.git
+Docs: https://mochajs.org/
+```
```
plroebuck: Not sure where it would go, but shouldn't `--debug` be in Usage?

dogancelik:
I get this error after updating to `mocha@next`:

```
# npm i --save-dev mocha@next
+ mocha@6.0.0-0
updated 1 package and audited 1283 packages in 5.509s
found 0 vulnerabilities

# npx mocha --debug
Cannot find module './mocharc.json'

# npx mocha --help
Cannot find module './mocharc.json'

# npx mocha --version
Cannot find module './mocharc.json'
```

plroebuck:
@dogancelik,
Looks like you didn't create a settings file. Copy [".mocharc.yml"](https://github.com/mochajs/mocha/blob/master/example/config/mocharc.yml) into your project root directory and edit accordingly.

See the [docs](https://next.mochajs.org/#configuring-mocha-nodejs).

@boneskull, this seems wrong though -- a settings file shouldn't be _required_. Mocha shouldn't even be checking for one if one'n'dones given from cmdline.

dogancelik:
@plroebuck I don't think any other Node tool I use forces me to create a RC file, do I need a RC file to do `--version`?

boneskull:
`.mocharc.json` is likely `lib/mocharc.json` which may not have been published properly.  I'll need to check.

lib/mocharc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "diff": true,
+  "extension": ["js"],
+  "opts": "./test/mocha.opts",
+  "package": "./package.json",
+  "reporter": "spec",
+  "slow": 75,
+  "timeout": 2000,
+  "ui": "bdd"
+}
```
plroebuck: Just noticed this file. Why was it placed in "/lib" rather than in "/config" (since it's not source code)?

boneskull: because it didn’t seem like it needed a new directory containing a single file

lib/cli/run.js
boneskull: https://github.com/mochajs/mocha/issues/3564#issuecomment-458740928

johansteffner:
After updating to mocha 6, our tests consisting of a file `test.js` has stopped working. I guess this is since the default value for `spec` is test/ and not test. This is not described in the breaking changes for v6.

I'm wondering if this should be considered a bug or if this is the new expected behaviour.

boneskull:
@johansteffner likely a regression

boneskull:
@johansteffner please provide error output

johansteffner:
@boneskull thanks for the response, here is an example of the issue https://github.com/johansteffner/mocha-spec-issue

The expected outcome is for the tests to run be i receive the error `Error: No test files found: "test/"`.

outsideris:
@johansteffner Please open it as new issue. It is better to let people help your problem than leaving the comment. Anyway it is fixed in v6.0.1, see #3754 .

PR #3563, opened 2018-11-10
ms module added instead of internal ms js #3557

### LINKED ISSUE #3557: use ms module instead of internal implementation
It's good to reduce LOC and use purpose-built modules.  

Prompted by comments on #3556.

Remove internal `ms.js` and use https://npm.im/ms instead.

gizemkeser:
### Requirements

Replace internal "ms.js" file with npm [ms](https://npm.im/ms) package.

### Description of the Change

- Dependency for the `ms` package has been added to "package.json".
- All references to the old "ms.js" file has been changed to new `ms` module.
- Internal "ms.js" file and its test specification has been removed.
- The test that checks the negative time has been enabled in "runnable.spec.js".

### Benefits

- Reduces lines of code to maintain.
- Using purpose build npm package

### Applicable issues

Fixes #3557
(semver-patch)


package.json
```diff
@@ -479,7 +479,8 @@
     "he": "1.1.1",
     "minimatch": "3.0.4",
     "mkdirp": "0.5.1",
-    "supports-color": "5.4.0"
+    "supports-color": "5.4.0",
+    "ms": "2.1.1"
```
plroebuck: Swap these two lines so deps are in alphabetical order.

gizemkeser: thanks for your comment :+1:

Issue #2134, opened 2016-02-29
Mocha reports the wrong message for a before() failure


jdmarshall:
With nested test suites, a failure will be reported for the last 'it' clause that ran, regardless of whether it has any relationship to the before() that caused the failure.

```
it('does something or other', function() {
    // Do something or other
});

describe('Outer suite', function() {
    before(function() {
        throw new Error("Something went wrong");
    });

    describe('when nothing goes wrong', function() {
        before(function() {
            //...
        });
        it('should have done something interesting', function() {
            //Do something interesting.
        });
    });
});
```

Expected behavior:  
    "before all" hook for "should have done something interesting"

Actual behavior:  
    "before all" hook for "does something or other"

mekdev:
We are seeing the same too and it does not have to be a nested suite, just something that runs sequential. See below

**a.js** 
```
'use strict';
describe('Suite A', function() {
    before(function () {
        console.log('before A')
    });

    it('A should do something', function*() {
        console.log('Test A')
    });

});
```


File **b.js**

```
'use strict';
describe('Suite B', function() {

    before(function () {
        console.log('before B')
        throw new Error("Something went wrong in before B");
    });

    it('B should do something', function*() {
        console.log('Test B')
    });
});
```

**test.js**

```
require('./a.js');
require('./b.js');
```

Result 

```
 mstittri@mstittri-osx node_modules/.bin/_mocha test/test.js


  Suite A
before A
Test A
    ✓ A should do something

  Suite B
before B
    1) "before all" hook for "A should do something"


  1 passing (13ms)
  1 failing

  1) Suite B "before all" hook for "A should do something":
     Error: Something went wrong in before B
      at Context.<anonymous> (/.../test/b.js:7:15)
      at callFn (/.../node_modules/mocha/lib/runnable.js:286:21)
      at Hook.Runnable.run (/.../node_modules/mocha/lib/runnable.js:279:7)
      at Hook.Runnable.run (/.../node_modules/co-mocha/lib/co-mocha.js:43:16)
      at next (/.../node_modules/mocha/lib/runner.js:297:10)
      at Immediate._onImmediate (/.../node_modules/mocha/lib/runner.js:319:5)
      at processImmediate [as _immediateCallback] (timers.js:383:17)
```

The correct behavior should be 

```
 1) "before all" hook for "Suite B"
```

Or something that would point user to B.js not A.js

mekdev:
[...]

Issue #1638, opened 2015-04-02
setting currentTest in a hook is incorrect


brian-mann:
This line:

https://github.com/mochajs/mocha/blob/41a21192ff288958b81654aeea3805190e9304b3/lib/runner.js#L263

This is completely wrong based on the when `self.test` is defined.

Currently `self.test` is defined when a test emits its `test` event here:

https://github.com/mochajs/mocha/blob/41a21192ff288958b81654aeea3805190e9304b3/lib/runner.js#L462

This has the following problems:
- A `beforeAll` hook will not have `hook.ctx.currentTest` defined, since `self.test` has not been assigned to the runner yet.
- Nested suites which have a `beforeAll` hook will actually set the `hook.ctx.currentTest` to the PREVIOUSLY run test, based on the same logical problem as the previous bullet point.
- `beforeEach` hooks do not suffer from this problem because tests emit their `test` event prior to the hooks being invoked, thus setting `self.test` on the runner.

The structure I'm referring to looks like this:

``` coffeescript
before ->
  ## this.currentTest is undefined

describe "outer suite", ->
  it "foo", ->

  context "nested [0me]", ->
    before ->
      ## this.currentTest will refer to "foo" test and not "bar" test

    it "bar", ->
```

danielstjules:
So, if I understand correctly, you'd like currentTest to be set for the `before` hook. To demonstrate how this behavior does not currently exist:

``` javascript
describe('example', function() {
  before(function() {
    console.log('before', this.currentTest);
  });

  beforeEach(function() {
    console.log('beforeEach', this.currentTest.title)
  });

  afterEach(function() {
    console.log('afterEach', this.currentTest.title)
  });

  after(function() {
    console.log('after', this.currentTest.title)
  });

  it('test1', function() { });

  it('test2', function() { });
});
```

```
$ ./bin/mocha test.js

before undefined
beforeEach test1

  ․afterEach test1
beforeEach test2
․afterEach test2
after test2
```

brian-mann:
Yes, you've demonstrated the bug exactly as described.

The one thing you're missing is that if you were to nest a new `describe` block, and then issue a new `before` hook, instead of seeing `undefined` as `this.currentTest`, you would actually see a reference to the previous test.

That makes this bug twice as bad.  In a top level `before` hook, `this.currentTest` is usually `undefined`, but in a nested `before` hook, `this.currentTest` will actually refer to the **wrong** test.

ORESoftware:
[...]

PR #3573, opened 2018-11-15
Fix Hook Test Names

juergba:
### Description of bug
Mocha shows an incorrect test-title if a nested `before` hook fails. For `after` hooks the second part of the message is completely missing.
```js
describe('Top-level describe', () => {
  it('Top-level it', () => { });       // the failing before-hook is pointing to this test
  describe('Nested describe', () => {
    before(() => {
      throw new Error();
    });
    it('Nested it', () => { });        // but it should point to this test
  });
});
```
```
Top-level describe
    ✓ Top-level it
    Nested describe
      1) "before all" hook for "Top-level it"       <<<incorrect
```
Root problem: see [#1638](https://github.com/mochajs/mocha/issues/1638). `hook.ctx.currentTest` is set incorrectly in the hook before emitting the hook event:
- `before` hook: currentTest is undefined or points to the outer test already past successfully.
The runner hasn't reached the next test to come yet, so it's too early to derive currentTest from the runner. Instead the first test of hook's parent is used.
- `after` hook: currentTest is undefined.
The runner has already completed the latest test and runner.currentTest is deleted. Instead the last test of hook's parent is used.

#### 1. Setting currentTest correctly
- `before` hook: hook.ctx.currentTest is set to the first test of the parent suite
- `after hook`: hook.ctx.currentTest is set to the last test of the parent suite
- for nested suites just the current suite is searched for tests to come. When no tests can be found, currentTest remains undefined.

#### 2. Creating title in hook failure message
A correct `hook.ctx.currentTest` renders a correct title. 
When no currentTest can be found, the title of the parent suite is used for the hook failure message.

### Alternate Designs
PR [#3333](https://github.com/mochajs/mocha/pull/3333)

### Benefits

This is a rather old bug back to 2015.

### Applicable issues
closes [#1638](https://github.com/mochajs/mocha/issues/1638)
closes [#3291](https://github.com/mochajs/mocha/issues/3291)
closes [#2134](https://github.com/mochajs/mocha/issues/2134)

**EDITED by @boneskull**: added js syntax highlighting, comments

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -265,10 +265,18 @@ Runner.prototype.fail = function(test, err) {
  * @param {Error} err
  */
 Runner.prototype.failHook = function(hook, err) {
+  hook.originalTitle = hook.originalTitle || hook.title;
   if (hook.ctx && hook.ctx.currentTest) {
-    hook.originalTitle = hook.originalTitle || hook.title;
     hook.title =
       hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
+  } else {
+    var parentTitle;
+    if (hook.parent.title) {
+      parentTitle = hook.parent.title;
+    } else {
+      parentTitle = hook.parent.root ? '{root}' : '';
+    }
+    hook.title = hook.originalTitle + ' in "' + parentTitle + '"';
   }
 
   this.fail(hook, err);
@@ -294,7 +302,13 @@ Runner.prototype.hook = function(name, fn) {
     }
     self.currentRunnable = hook;
 
-    hook.ctx.currentTest = self.test;
+    if (name === 'beforeAll') {
+      hook.ctx.currentTest = hook.parent.tests[0];
+    } else if (name === 'afterAll') {
+      hook.ctx.currentTest = hook.parent.tests[hook.parent.tests.length - 1];
+    } else {
+      hook.ctx.currentTest = self.test;
+    }
 
     self.emit('hook', hook);
 
--- a/test/integration/fixtures/current-test-title.fixture.js
+++ b/test/integration/fixtures/current-test-title.fixture.js
@@ -1,0 +1,47 @@
+'use strict';
+var assert = require('assert');
+
+function getTitle(ctx) {
+  return ctx.currentTest && ctx.currentTest.title;
+};
+
+before(function () {
+  assert.equal(getTitle(this), undefined);
+});
+
+describe('suite A', () => {
+
+  before(function () {
+    assert.equal(getTitle(this), undefined);
+  });
+
+  describe('suite B', () => {
+
+    it('test1 B', () => {});
+
+    describe('suite C', function () {
+      var lap = 0;
+
+      before(function () {
+        assert.equal(getTitle(this), 'test1 C');
+      });
+      beforeEach(function () {
+        assert.equal(getTitle(this), ++lap === 1 ? 'test1 C' : 'test2 C');
+      });
+
+      it('test1 C', function () {});
+      it('test2 C', function () {});
+
+      afterEach(function () {
+        assert.equal(getTitle(this), lap === 1 ? 'test1 C' : 'test2 C');
+      });
+      after(function () {
+        assert.equal(getTitle(this), 'test2 C');
+      });
+    });
+  });
+});
+
+after(function () {
+  assert.equal(getTitle(this), undefined);
+});
--- a/test/integration/fixtures/hooks/before-hook-deepnested-error.fixture.js
+++ b/test/integration/fixtures/hooks/before-hook-deepnested-error.fixture.js
@@ -1,0 +1,13 @@
+'use strict';
+
+describe('spec 1', function () {
+  it('should pass', function () { });
+  describe('spec 2 nested - this title should be used', function () {
+    before(function() {
+      throw new Error('before hook nested error');
+    });
+    describe('spec 3 nested', function () { 
+      it('it nested - this title should not be used', function () { });
+    });
+  });
+});
--- a/test/integration/fixtures/hooks/before-hook-root-error.fixture.js
+++ b/test/integration/fixtures/hooks/before-hook-root-error.fixture.js
@@ -1,0 +1,9 @@
+'use strict';
+
+before(function() {
+  throw new Error('before hook root error');
+});
+
+describe('spec 1', function () {
+  it('should not be called', function () { });
+});
--- a/test/integration/fixtures/options/bail-with-after.fixture.js
+++ b/test/integration/fixtures/options/bail-with-after.fixture.js
@@ -50,7 +50,7 @@ describe('suite2', function () {
   before('before suite2', function () {});
   beforeEach('beforeEach suite2', function () {});
   it('test suite2', function () {
-    runOrder.push('test suite2 - should not run');
+    console.log('test suite2 - should not run');
   });
   afterEach('afterEach suite2', function () {});
   after('after suite2', function () {});
--- a/test/integration/fixtures/options/bail-with-before.fixture.js
+++ b/test/integration/fixtures/options/bail-with-before.fixture.js
@@ -37,7 +37,7 @@ describe('suite2', function () {
   before('before suite2', function () {});
   beforeEach('beforeEach suite2', function () {});
   it('test suite2', function () {
-    runOrder.push('test suite2 - should not run');
+    console.log('test suite2 - should not run');
   });
   afterEach('afterEach suite2', function () {});
   after('after suite2', function () {});
--- a/test/integration/hook-err.spec.js
+++ b/test/integration/hook-err.spec.js
@@ -2,6 +2,7 @@
 
 var helpers = require('./helpers');
 var runMocha = helpers.runMocha;
+var runMochaJSON = require('./helpers').runMochaJSON;
 var splitRegExp = helpers.splitRegExp;
 var bang = require('../../lib/reporters/base').symbols.bang;
 
@@ -25,6 +26,59 @@ describe('hook error handling', function() {
     });
   });
 
+  describe('before hook root error', function() {
+    it('should verify results', function(done) {
+      var fixture = 'hooks/before-hook-root-error.fixture.js';
+      runMochaJSON(fixture, [], function(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res, 'to have failed with error', 'before hook root error')
+          .and('to have failed test', '"before all" hook in "{root}"')
+          .and('to have passed test count', 0);
+        done();
+      });
+    });
+  });
+
+  describe('before hook nested error', function() {
+    it('should verify results', function(done) {
+      var fixture = 'hooks/before-hook-nested-error.fixture.js';
+      runMochaJSON(fixture, [], function(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res, 'to have failed with error', 'before hook nested error')
+          .and(
+            'to have failed test',
+            '"before all" hook for "it nested - this title should be used"'
+          )
+          .and('to have passed test count', 1)
+          .and('to have passed test', 'should pass');
+        done();
+      });
+    });
+  });
+
+  describe('before hook deepnested error', function() {
+    it('should verify results', function(done) {
+      var fixture = 'hooks/before-hook-deepnested-error.fixture.js';
+      runMochaJSON(fixture, [], function(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res, 'to have failed with error', 'before hook nested error')
+          .and(
+            'to have failed test',
+            '"before all" hook in "spec 2 nested - this title should be used"'
+          )
+          .and('to have passed test count', 1)
+          .and('to have passed test', 'should pass');
+        done();
+      });
+    });
+  });
+
   describe('before each hook error', function() {
     before(run('hooks/beforeEach-hook-error.fixture.js'));
     it('should verify results', function() {
@@ -36,6 +90,53 @@ describe('hook error handling', function() {
     before(run('hooks/after-hook-error.fixture.js'));
     it('should verify results', function() {
       expect(lines, 'to equal', ['test 1', 'test 2', 'after', bang + 'test 3']);
+    });
+  });
+
+  describe('after hook nested error', function() {
+    it('should verify results', function(done) {
+      var fixture = 'hooks/after-hook-nested-error.fixture.js';
+      runMochaJSON(fixture, [], function(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res, 'to have failed with error', 'after hook nested error')
+          .and(
+            'to have failed test',
+            '"after all" hook for "it nested - this title should be used"'
+          )
+          .and('to have passed test count', 3)
+          .and(
+            'to have passed test order',
+            'should pass',
+            'it nested - this title should be used',
+            'it nested - not this title'
+          );
+        done();
+      });
+    });
+  });
+
+  describe('after hook deepnested error', function() {
+    it('should verify results', function(done) {
+      var fixture = 'hooks/after-hook-deepnested-error.fixture.js';
+      runMochaJSON(fixture, [], function(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res, 'to have failed with error', 'after hook nested error')
+          .and(
+            'to have failed test',
+            '"after all" hook in "spec 2 nested - this title should be used"'
+          )
+          .and('to have passed test count', 2)
+          .and(
+            'to have passed test order',
+            'should pass',
+            'it nested - this title should not be used'
+          );
+        done();
+      });
     });
   });
 
--- a/test/integration/hook-err.spec.js
+++ b/test/integration/hook-err.spec.js
@@ -18,7 +18,10 @@ describe('hook error handling', function() {
   describe('before hook error tip', function() {
     before(run('hooks/before-hook-error-tip.fixture.js', onlyErrorTitle()));
     it('should verify results', function() {
-      expect(lines, 'to equal', ['1) spec 2', '"before all" hook:']);
+      expect(lines, 'to equal', [
+        '1) spec 2',
+        '"before all" hook for "skipped":'
+      ]);
     });
   });
 
--- a/test/integration/multiple-done.spec.js
+++ b/test/integration/multiple-done.spec.js
@@ -90,7 +90,10 @@ describe('multiple calls to done()', function() {
     });
 
     it('correctly attributes the error', function() {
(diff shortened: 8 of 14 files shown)


lib/runner.js
```diff
@@ -273,6 +273,10 @@ Runner.prototype.failHook = function(hook, err) {
     hook.originalTitle = hook.originalTitle || hook.title;
     hook.title =
       hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
+  } else {
+    var parent = hook.parent ? hook.parent.title : 'root';
```
boneskull: this is incorrect; your original code in 5ea6e653 was fine.  The test failures in `test/unit/runner.spec.js` are due to bad test fixtures--`parent` is a required property of any `Runnable`.

You can fix the tests by ensuring every newly-instantiated `Test` belongs to a `Suite`, like this:

```js
it('should increment .failures', function() {
  expect(runner.failures, 'to be', 0);
  var test1 = new Test('fail hook 1', noop);
  var test2 = new Test('fail hook 2', noop);
  suite.addTest(test1);
  suite.addTest(test2);
  runner.failHook(test1, {});
  expect(runner.failures, 'to be', 1);
  runner.failHook(test2, {});
  expect(runner.failures, 'to be', 2);
});
```

test/integration/fixtures/current-test-title.fixture.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+function getTitle() {
+  return this.currentTest ? this.currentTest.title : undefined;
```
boneskull: maybe more straightforward to skip `Function.prototype.call` altogether:

```js
function getTitle(hook) {
  return hook.currentTest && hook.currentTest.title;
}

before(function() {
  getTitle(this);
});
```

when coerced to a string you're looking at `'undefined'` in your implementation or `false` in mine; either way it should cause a test failure. you'll need to update your test using this of course

test/integration/hook-err.spec.js
```diff
@@ -18,7 +18,34 @@ describe('hook error handling', function() {
   describe('before hook error tip', function() {
     before(run('hooks/before-hook-error-tip.fixture.js', onlyErrorTitle()));
     it('should verify results', function() {
-      assert.deepEqual(lines, ['1) spec 2', '"before all" hook:']);
+      assert.deepEqual(lines, [
```
boneskull: I'm not sure why (can't remember?) we're using `assert` in these tests, but please use `expect` like in the other tests.  if that fails for some reason, then we'll know why :smile:

You don't need to convert all tests in these files to use `expect` (unless you want to)

test/integration/multiple-done.spec.js
```diff
@@ -84,7 +84,10 @@ describe('multiple calls to done()', function() {
     });
 
     it('correctly attributes the error', function() {
-      assert.equal(res.failures[0].fullTitle, 'suite "before all" hook');
+      assert.equal(
```
boneskull: pls try to use `expect`

lib/runner.js
boneskull: that said, you will need to address the fact that the root suite a title of an empty string. :smile:

plroebuck: But suite `title` having an empty string value isn't set in stone, just the default. The root suite _should_ be identified by presence of truthy `root` property.

boneskull: my comment was in reference to how the error message is getting concatenated, anyway.  an empty  `title` would need to be taken into consideration.  

the root suite having an empty string title sounds like one of those things that other code makes assumptions about.

test/integration/hook-err.spec.js
juergba: @boneskull I'm trying to change "hook-err.spec.js" and to use `expect` as shown in f.e. "options.spec.js". As I understood so far, the key difference is: the "options.\*.fictures.js" don't use `console.log()` while "hooks.\*.fictures.js" do.
In order to use `expect` I need to call (helpers.js)runMochaJSON which results in a parsing error when using `console.log`. Probably I can change most of the tests accordingly, but there will be some depending on the `console.log` output.
Since JSON does not accept comments `//` I don't see a possibility to get around.
Do you have any tips?

lib/runner.js
```diff
@@ -272,7 +272,11 @@ Runner.prototype.failHook = function(hook, err) {
     hook.title =
       hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
   } else {
-    var parent = hook.parent ? hook.parent.title : 'root';
+    var parent = hook.parent.title
+      ? hook.parent.title
+      : hook.parent.root && !hook.parent.parent
+        ? 'root'
+        : '';
```
juergba: `hook.parent.root` is not reliable since any suite with an empty string title "" is marked as root. That's why I'm testing for `hook.parent.parent`, also

juergba:
@boneskull Could you have another look, please? Thanks.

lib/runner.js
plroebuck: Sounds like YA bug...

plroebuck: > `hook.parent.root` is not reliable since any suite with an empty string title "" is marked as root.

Sounds like YA bug...

juergba: see PR #3632

test/integration/fixtures/hooks/after-hook-async-error.fixture.js
```diff
@@ -2,20 +2,13 @@
...
+      done(new Error('after hook error'));
     });
   });
-  it('should be called because error is in after hook', function () {
-    console.log('test 1');
-  });
-  it('should be called because error is in after hook', function () {
-    console.log('test 2');
-  });
+  it('test 1 - should be called', function () {});
```
plroebuck: why was all the context from the test descriptions removed?

lib/runner.js
plroebuck: Should the root suite be output as maybe `'{root}'` instead to designate its implied title (rather than an explicit one).

plroebuck:
What's the status on this PR now that the "just one root" PR is done?

juergba:
@plroebuck I will rebase it and revert all changes to test files not directly linked to this PR ("pls try to use expect").

test/integration/fixtures/hooks/before-hook-deepnested-error.fixture.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+describe('spec 1', function () {
+  it('should pass', function () { });
+  describe('spec 2 nested - this title should be used', function () {
+    before(function() {
+      throw new Error('before hook nested error');
+    });
+    describe('spec 3 nested', function () { 
+        it('it nested - this title should not be used', function () { });
```
plroebuck: dedent 2 spaces

test/integration/fixtures/hooks/multiple-hook-async-error.fixture.js
```diff
@@ -1,141 +1,160 @@
...
-after(function () {
-  console.log('root after');
+afterEach('root after each', function () {
+  runOrder.push('root after each');
 });
-afterEach(function () {
-  console.log('root after each');
+after('root after', function () {
+  runOrder.push('root after');
+  assert.deepEqual(runOrder, [
```
plroebuck: Do your assertion check in a separate `after`.

test/integration/fixtures/hooks/multiple-hook-error.fixture.js
```diff
@@ -1,131 +1,150 @@
...
-after(function () {
-  console.log('root after');
+afterEach('root after each', function () {
+  runOrder.push('root after each');  
 });
-afterEach(function () {
-  console.log('root after each');
+after('root after', function () {
+  runOrder.push('root after');
+  assert.deepEqual(runOrder, [
```
plroebuck: same as previous comment

test/unit/runner.spec.js
```diff
@@ -344,9 +344,13 @@ describe('Runner', function() {
   describe('.failHook(hook, err)', function() {
     it('should increment .failures', function() {
       expect(runner.failures, 'to be', 0);
-      runner.failHook(new Test('fail hook 1', noop), {});
+      var test1 = new Test('fail hook 1', noop);
+      var test2 = new Test('fail hook 2', noop);
+      suite.addTest(test1);
+      suite.addTest(test2);
+      runner.failHook(test1, {});
```
plroebuck: Shouldn't second argument be an `Error`?

juergba: It's an existing test, I did not change that part. It works fine like this.

juergba:
@plroebuck thanks for your feedback. 
@boneskull 
I rebased this branch, adapted some hook titles in existing tests and added my new tests. I reverted many changes to existing tests in order to keep this PR lean.
Could you review, please?

lib/runner.js
```diff
@@ -265,10 +265,17 @@ Runner.prototype.fail = function(test, err) {
...
-    hook.originalTitle = hook.originalTitle || hook.title;
     hook.title =
       hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
+  } else {
+    var parent = hook.parent.title
+      ? hook.parent.title
+      : hook.parent.root
+      ? '{root}'
+      : '';
+    hook.title = hook.originalTitle + ' in "' + parent + '"';
```
plroebuck: Venturing into ternary op abuse...
```js
    var parent = hook.parent.title
      ? hook.parent.title
      : hook.parent.root
        ? '{root}'
        : '';
```

I indented the latter -- if prettier did this, it's borken...

test/unit/runner.spec.js
plroebuck: Does this work?
```js
runner.failHook(test1, new Error('bummer'));
```

test/unit/runner.spec.js
```diff
@@ -375,6 +380,7 @@ describe('Runner', function() {
 
     it('should not emit "end" if suite bail is not true', function(done) {
       var hook = new Hook();
+      hook.parent = suite;
       var err = {};
```
plroebuck: Why is this faking `Error`s instead of making them?

lib/runner.js
```diff
@@ -265,10 +265,17 @@ Runner.prototype.fail = function(test, err) {
  * @param {Error} err
  */
 Runner.prototype.failHook = function(hook, err) {
+  hook.originalTitle = hook.originalTitle || hook.title;
   if (hook.ctx && hook.ctx.currentTest) {
-    hook.originalTitle = hook.originalTitle || hook.title;
     hook.title =
       hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
+  } else {
+    var parent = hook.parent.title
```
plroebuck: `parentTitle`?

test/integration/hook-err.spec.js
```diff
@@ -18,7 +19,72 @@ describe('hook error handling', function() {
...
+        '1) spec 2',
+        '"before all" hook for "skipped":'
+      ]);
+    });
+  });
+
+  describe('before hook root error', function() {
+    it('should verify results', function(done) {
+      runJSON('hooks/before-hook-root-error.fixture.js', [], function(
+        err,
```
plroebuck: Suggest minor change to avoid ^%*^&% prettier reformatting...
```js
      var fixture = 'hooks/before-hook-root-error.fixture.js';
      runJSON(fixture, [], function(err, res) {
```

lib/runner.js
juergba: I had to use `// prettier-ignore`

test/unit/runner.spec.js
juergba: It works both ways.

lib/runner.js
juergba: `hook.parent.title` is correct, it's the suite title.

test/integration/hook-err.spec.js
juergba: I inserted `// eslint-disable-next-line`

lib/runner.js
plroebuck: Was in reference to `var parent`

test/unit/runner.spec.js
plroebuck: Then for clarity of intent, create an `Error` instance rather than relying on internal processing to do it for you...

lib/runner.js
plroebuck: Chris won't go for that (turning off prettier).
By reformulate, I meant use two variables or `if` logic -- combining two (or more) ternary operators on a single variable is too much added complexity (metrics-wise) for a single variable (IMO).

test/integration/hook-err.spec.js
plroebuck: Please remove the `eslint` directive -- we use them for enforcement, not formatting

test/unit/runner.spec.js
plroebuck: Same as comment below -- create `Error`.

test/integration/fixtures/hooks/multiple-hook-async-error.fixture.js
plroebuck: Where did this go?

lib/runner.js
boneskull: guilty as charged. 

the idea with prettier is to just let it do whatever it's going to do.  sometimes it might format something a way you don't like, or I don't like, or maybe *nobody* likes, but the intent is to make formatting a non-issue by complete, final, *mindless* adherence to whatever it does.  when we pray at the unholy altar of prettier, it demands no less than total submission.

We *do* have agency when using nested ternaries though.  I try to avoid them in my own code.  but if I've requested people eliminate them in the past, I think I'm over it now.  there are more important things to worry about.

lib/runner.js
```diff
@@ -294,7 +302,13 @@ Runner.prototype.hook = function(name, fn) {
     }
     self.currentRunnable = hook;
 
-    hook.ctx.currentTest = self.test;
+    if (name === 'beforeAll') {
```
boneskull: note to self: replace with constant, see #3655

test/integration/hook-err.spec.js
```diff
@@ -36,6 +96,55 @@ describe('hook error handling', function() {
...
+            'it nested - not this title'
+          );
+        done();
+      });
+    });
+  });
+
+  describe('after hook deepnested error', function() {
+    it('should verify results', function(done) {
+      // eslint-disable-next-line
```
boneskull: what's avoided here?

test/unit/runner.spec.js
boneskull: I was going to say it wasn't worth requesting, but @plroebuck changed my mind.

It *is* irrelevant to the test, but the test becomes more tightly coupled to the internal implementation if we don't pass an `Error`.  probably best to change it.

test/integration/hook-err.spec.js
boneskull: I think you figured this out?

test/integration/fixtures/hooks/multiple-hook-async-error.fixture.js
juergba: I reversed my changes to this file, out of topic.

lib/runner.js
```diff
@@ -265,10 +265,17 @@ Runner.prototype.fail = function(test, err) {
...
   if (hook.ctx && hook.ctx.currentTest) {
-    hook.originalTitle = hook.originalTitle || hook.title;
     hook.title =
       hook.originalTitle + ' for "' + hook.ctx.currentTest.title + '"';
+  } else {
+    var parentTitle = hook.parent.title
+      ? hook.parent.title
+      : hook.parent.root
+      ? '{root}'
+      : '';
```
plroebuck: I asked before... please reformulate this. Since prettier denies us the ability to indent the nested ternary, we should not use them. Nesting them is considered a code smell in many languages. [This article](https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340) says they're great, but almost _all_ the commenters disagree.

```js
    var parentTitle;
    if (hook.parent.title) {
      parentTitle = hook.parent.title;
    } else {
      parentTitle = hook.parent.root
        ? '{root}'
        : '';
    }
```

juergba: @plroebuck I usually expect more tolerance against other's coding style, but I will make this change for you. This PR was my first one and you have supported me from beginning. So thanks for that!
Tomorrow I'm going to merge this PR.

boneskull: agree. @mochajs/core 	please dial it back on code style comments.  our linters and formatters are more than up to the task.  if they fall short, we can create new rules to extend functionality.

test/integration/fixtures/options/bail-with-after.fixture.js
```diff
@@ -50,7 +50,7 @@ describe('suite2', function () {
   before('before suite2', function () {});
   beforeEach('beforeEach suite2', function () {});
   it('test suite2', function () {
-    runOrder.push('test suite2 - should not run');
+    console.log('test suite2 - should not run');
```
plroebuck: Shouldn't this be an assertion?

test/integration/fixtures/options/bail-with-before.fixture.js
```diff
@@ -37,7 +37,7 @@ describe('suite2', function () {
   before('before suite2', function () {});
   beforeEach('beforeEach suite2', function () {});
   it('test suite2', function () {
-    runOrder.push('test suite2 - should not run');
+    console.log('test suite2 - should not run');
```
plroebuck: Shouldn't this be an assertion?

plroebuck:
Could you check this against the test cases from #3333?
Believe your PR should cover Heather's tail patch, but would like to be sure.

test/integration/fixtures/hooks/before-hook-root-error.fixture.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+before(function() {
+  throw new Error('before hook root error');
+});
+
+describe('spec 1', function () {
+  it('should not be called', function () { });
```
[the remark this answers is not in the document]
juergba: The assertions are done in "hook-err.spec.js":
```
expect(res, 'to have failed with error', 'before hook root error')
          .and('to have failed test', '"before all" hook in "{root}"')
          .and('to have passed test count', 0);
```

test/integration/fixtures/options/bail-with-after.fixture.js
juergba: `describe('suite2', function () {...} )` is skipped by the `--bail` flag. So this blog does never run.

test/integration/fixtures/options/bail-with-before.fixture.js
juergba: describe('suite2', function () {...} ) is skipped by the --bail flag. So this blog does never run.

juergba:
I checked the tests of #3333, they are covered with this PR and their output is correct.

Issue #3562, opened 2018-11-09
Use "package.json" to determine test directory


plroebuck:
### Description

Read "package.json" field `directories.test` if it exists. If it does, use its value instead of `"test"` as Mocha's default for test directory. 

Inspired from discussion in #1275 where someone had named his test directory "tests" and then Mocha wouldn't find his "mocha.opts" file (without specifying it via `--opts`).

juergba:
`package.json` is not updated dynamically. When creating/renaming directories, the field `directories.test` would have to be maintained manually. `package.json` is not a good place to define the default directory. There is already a place to set such values, which is the `mocha.opts` file.

For the test files `scripts.test` in `package.json` can already be used for defining the test directory.
```
"scripts": {
    "test": "mocha ./tests/"
  },
```

plroebuck:
> `package.json` is not a good place to define the default directory.

Unsure why you think this, since it's [specifically documented](https://docs.npmjs.com/files/package.json#directoriestest) as the means to communicate a package's structure.

The `directories` field in "package.json" _does not exist_ by default; therefore, if `directories.test` does exist, the user purposefully set it and we should respect that value, without making them re-specify the same thing in "mocha.opts" or the `scripts.test` field of "package.json".

It should just work...

juergba:
> The directories field in "package.json" does not exist by default; [...]

When you do `npm init` and the directory test already exists, the `directories` field is created automatically. This works at least with VSCode, no other directories unless `test` are recognised and inserted.

> It should just work...

You have made your decision, so close the `discussion` label.
And I would like to implement this.

PR #3381, opened 2018-05-12
3363: Throw error if unable to parse Mocha options file

plroebuck:
### Description of the Change

Code now throws an error for any problems loading the Mocha options file (except nonexistent default "test/mocha.opts").

Builds off of previous discussions and work done in PR #2716 and PR #3352 .

### Why should this be in core?

Silently ignoring problems loading this file is just wrong.

### Benefits

Users will now be told if the file failed to load.

### Possible Drawbacks

Users with invalid Mocha options files (or permissions problems, etc.) will need to correct the problem.

### Applicable issues

Fixes #3363
Fixes #2576
(server-major) [since Mocha could throw an error when it didn't previously]

PR #3582, opened 2018-11-27
WIP: Use "package.json" to determine test directory

juergba:
### Description

By default, the test folder is set to "test".
Set the field `directories.test` of "package.json" to use an alternate folder.

#### "bin/options.js"
- get `directories.test` of "package.json"
- if the field exists, read "mocha.opts" from that directory. The `--opt` flag has higher priority.
#### "bin/_mocha.js"
- if the field exists, use that directory as the default location for test files.

nice-to-have

### Applicable issues

#3562

### Additional information
Finding "package.json" in user's project root (usr/root):
- `npm test`: it's easy to find, since `npm` throws an error or sets `process.pwd` always to usr/root.
-- usr/root: found successfully, cwd == usr/root
-- subdirs of usr/root: found successfully, cwd == usr/root
-- anywhere else: throws ENOENT error: no such file or directory
- `node mocha`: can be started from anywhere in the filesystem, even from another partition. `process.cwd` is set to directory where the Node process was started. It isn't possible to determine usr/root exactly, Mocha doesn't know it (f.e. environment variable).
-- search first in cwd (which can be totally wrong), then in cwd's parent directories.
-- ~~`mocha "./spec/*.js"`: when specific test files are passed, use this location.~~
-- ~~when no "package.json" is found, throw an error, since later the test files won't be found neither.~~

The current release 5.2.0 also relies on `process.cwd` and fails finding the test files in described situation, however no error is thrown.


bin/options.js
```diff
@@ -69,9 +70,26 @@ function getOptions() {
...
+      fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf8')
+    ).directories['test'];
+
+    if (testDir && testDir !== 'test') {
+      process._testDirectory = testDir;
+    }
+  } catch (ignore) {
+    // NOTE: should console.error() and throw the error
+  }
+
```
plroebuck: So all of this should go into a new exported `getTestDirectory` function. Your implementation above should only be done once, caching the result.


Don't add properties to objects you don't own.

bin/options.js
```diff
@@ -69,9 +70,26 @@ function getOptions() {
     return;
   }
 
+  try {
+    const testDir = JSON.parse(
+      fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf8')
+    ).directories['test'];
+
```
plroebuck: This finds the user's "package.json", regardless of where Mocha was installed?
Would think you would need to use "read-pkg-up" npm package.
For example, see [here](https://github.com/yargs/yargs/blob/497194d0cf4253182aa32eed6cadfe6bc8507413/yargs.js#L358-L372)

docs/index.md
```diff
@@ -1326,7 +1326,13 @@ $ mocha --reporter list --growl
 
 ## The `test/` Directory
 
-By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option.
+By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. Use the field `directories.test` in package.json to overwrite the default folder. If you want to include sub directories, pass the `--recursive` option.
+
```
plroebuck: Confusing the way you added into the middle of existing paragraph here. Just make a new paragraph instead. Put `package.json` in double quotes and append the word 'file'. "Overwrite" is incorrect (and _scary_) -- maybe just say "use an alternate directory".

bin/options.js
juergba: I will have a look at 'read-pkg-up' tomorrow. What do you think of this solution?
`const testDir = require('package.json').directories.test;`
Are there several package.json files? For node and for mocha, globally/locally installed?

bin/mocha
```diff
@@ -9,7 +9,7 @@
 
 const spawn = require('child_process').spawn;
 const path = require('path');
-const getOptions = require('./options');
+const getOptions = require('./options').getOptions;
```
juergba: I should not have changed this file, right?
But I had to, since I adapted the module.exports object of options.js

bin/options.js
juergba: ok, done.

docs/index.md
juergba: ok, done

bin/options.js
juergba: ok, done.

juergba:
I inserted `read-pkg-up` as a dependency into package.json. Bad?

bin/_mocha
```diff
@@ -18,7 +18,7 @@ const utils = Mocha.utils;
 const interfaceNames = Object.keys(Mocha.interfaces);
 const join = path.join;
 const cwd = process.cwd();
-const getOptions = require('./options');
+const options = require('./options');
```
plroebuck: Add the functions here as well.
```js
const getOptions = options.getOptions;
const getTestDirectory = options.getTestDirectory;
```

bin/_mocha
```diff
@@ -524,10 +524,11 @@ mocha.ui(program.ui);
 
 const args = program.args;
 
-// default files to test/*.{js,coffee}
+// sets directory for test files with default to test/*.{js,coffee}
 
 if (!args.length) {
-  args.push('test');
+  const testDirectory = options.getTestDirectory();
+  testDirectory ? args.push(testDirectory) : args.push('test');
```
plroebuck: Huh? Why isn't this simply
```js
if (!args.length) {
  args.push(getTestDirectory());
}
```

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
 const defaultPathname = 'test/mocha.opts';
 
+/**
+ * Reads directories.test of package.json.
+ *
+ * @public
+ * @returns {string} test directory, if it exists
+ */
+function getTestDirectory() {
+  if (getTestDirectory.cash) {
```
plroebuck: did you mean
```js
if (getTestDirectory.cache) {
```

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+    return getTestDirectory.cash;
+  }
+
+  try {
+    const packPath = readPkgUp.sync('__dirname').path;
```
plroebuck:
```js
    const pkg = readPkgUp.sync('__dirname').path;
```

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+    const testDir = require(packPath).directories.test;
+    if (testDir) {
+      return (getTestDirectory.cash = testDir);
```
plroebuck:
```js
    if (testDir) {
      getTestDirectory.cache = testDir;
    }
```

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+    } else {
+      return (getTestDirectory.cash = 'test');
```
plroebuck: Remove.

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+    }
+  } catch (ignore) {
+    return (getTestDirectory.cash = 'test');
```
plroebuck: Remove

bin/_mocha
```diff
@@ -524,10 +526,10 @@ mocha.ui(program.ui);
 
 const args = program.args;
 
-// default files to test/*.{js,coffee}
+// sets directory for test files with default to test/*.{js,coffee}
```
plroebuck: Change this comment (since it's wrong anyway)
`// Set default directory for test files`

docs/index.md
```diff
@@ -1328,6 +1328,14 @@ $ mocha --reporter list --growl
 
 By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include sub directories, pass the `--recursive` option.
```
plroebuck: Fix spelling: "subdirectories" without space

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
  */
 const defaultPathname = 'test/mocha.opts';
```
plroebuck: Change this to incorporate `getTestDirectory()`.
See [this code](https://github.com/mochajs/mocha/blob/fcb6fb4972f1a9feeb0b3ceff72a6dac906adc3e/bin/options.js) from #3381. Feel free to incorporate...

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+/**
+ * Reads directories.test of package.json.
+ *
+ * @public
+ * @returns {string} test directory, if it exists
+ */
+function getTestDirectory() {
+  if (!getTestDirectory.cache) {
+    try {
+      const pkg = readPkgUp.sync('__dirname').path;
```
plroebuck: Actually, to be consistent with expectation that `pkg` will be an object:
```js
      const pkg = require(readPkgUp.sync('__dirname').path);
```

bin/options.js
```diff
@@ -69,9 +98,16 @@ function getOptions() {
...
+
   const optsPath =
     process.argv.indexOf('--opts') === -1
-      ? defaultPathname
+      ? testDirectory
+        ? path.join(
+            testDirectory,
+            defaultPathname.slice(defaultPathname.lastIndexOf('/') + 1)
+          )
+        : defaultPathname
```
plroebuck: None of this needs to change if `defaultPathname` set appropriately above.

plroebuck:
Assume you have created multiple sample projects to test this against?

* One with `directories.test` set with existing alternate directory (global Mocha)
* One with `directories.test` set with existing alternate directory (local Mocha)
* One with `directories.test` set with alternate directory that does not exist (local Mocha)
* One with `directories` set but not `directories.test` (local Mocha)
* One without `directories` set (local Mocha)

Additionally, with second one above, initiate `npm test` from different directories within project.

bin/options.js
boneskull: I mentioned in the chat room that I'd prefer using [findup-sync](https://npm.im/findup-sync), since I need to use that in #3556.

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+ * Reads directories.test of package.json.
+ *
+ * @public
+ * @returns {string} test directory, if it exists
+ */
+function getTestDirectory() {
+  if (!getTestDirectory.cache) {
+    try {
+      const pkg = readPkgUp.sync('__dirname').path;
+      const testDir = require(pkg).directories.test;
```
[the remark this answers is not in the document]
boneskull: yes, but we can safely abort at that point

bin/options.js
```diff
@@ -21,6 +25,31 @@ module.exports = getOptions;
...
+      if (testDir) {
```
boneskull: might as well use a ternary here, but that's just an opinion

juergba:
@plroebuck thanks for your feedback! I will wait for #3556 to be implemented first. 
> [...] getOptions() in bin/options.js is deprecated [...]

@boneskull finding user's root: I would love to have an environment variable which 
(can)/must be set by the user. Like in Java, Python, Node etc...

bin/options.js
plroebuck: @boneskull, used this since "yargs" does as well and provides us with a debugged, ready-to-go recipe (see example link)? Since they _will_ be a dependency, "read-pkg-up" package adds no extra cost...

bin/options.js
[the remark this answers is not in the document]
plroebuck:
```js
const testDir = pkg.directories && pkg.directories.test
  ? pkg.directories.test
  : undefined;
```

bin/options.js
```diff
@@ -3,14 +3,18 @@
...
+ * Export `getTestDirectory`.
  */
-
-module.exports = getOptions;
+module.exports = {
+  getOptions: getOptions,
+  getTestDirectory: getTestDirectory
+};
 
 /**
```
plroebuck: const DEFAULT_TEST_DIRECTORY = 'test';
const DEFAULT_OPTS_FILENAME = 'mocha.opts';

bin/options.js
plroebuck:
```js
const defaultPathname = path.join(getTestDirectory(), DEFAULT_OPTS_FILENAME);
```

bin/options.js
juergba: @plroebuck No, user's "package.json" can't be found in any situation, that's not realistic. I added above some "Additional information".

docs/index.md
```diff
@@ -1328,7 +1328,7 @@ $ mocha --reporter list --growl
 
 By default, `mocha` looks for the glob `./test/*.js`, so you may want to put your tests in `test/` folder. If you want to include subdirectories, pass the `--recursive` option.
 
-Mocha will honor the [test directory](https://docs.npmjs.com/files/package.json#directoriestest) setting in your "package.json" if it exists. 
+Mocha will honor the [test directory](https://docs.npmjs.com/files/package.json#directoriestest) setting in your "package.json" if it exists.
 
 Set this field's value to `"./test"` to match Mocha's default, or specify an alternate directory to where your Mocha tests. For example, if an example project used an alternate directory named "myTests", its "package.json" setting would look like this:
```
plroebuck: "to where" => "containing"

Sorry, rewrote several times, but mixed two thoughts here.

bin/options.js
```diff
@@ -17,37 +17,54 @@ module.exports = {
...
+      let pkgPath = readPkgUp.sync(process.cwd()).path;
+
+      if (!pkgPath) {
+        // PIW: get directory of specific test location
+        const specTest = path.dirname(process.argv[process.argv.length - 1]);
+        pkgPath = readPkgUp.sync(specTest).path;
+      }
+
+      if (pkgPath) {
+        const pkg = require(pkgPath);
```
plroebuck: Confused with _all_ of the above. Since `readPkgUp.sync` will return the package contents via `.pkg`, why go through all this?

bin/options.js
plroebuck: I did several other edits that have vanished. The biggest was this change:
`const pkg = readPkgUp.sync('__dirname').pkg;`

bin/options.js
juergba: @plroebuck in most cases "package.json" is located in `cwd` or one of its parent directories. `npm test` throws an error otherwise. With `node mocha` there is no guaranty that `cwd` is user's project root, please see above in "Additional information". So in a second step I'm trying to use the last argv element, if it is a path with/without glob, to get a second entry point for the search of "package.json". This is more complicated, since argv has not been processed at that point.

It would be far easier to just use `cwd`/parent dirs and throw an error otherwise as `npm`does.
And no, user's project root can't be located in any case.
I would prefer the way of `npm`, no second entry point and throw an error.

docs/index.md
juergba: ok, thanks

bin/options.js
juergba: @plroebuck this is a quote out of the chat room:
> mocha’s behavior is really just “best effort". we can’t know if the root is cwd, but most people are probably running mocha with the cwd of the root, so we can try ./test.

So, Mocha does one search with cwd as a starting point, there is no second search and no error will be thrown. Mocha will behave as in the current release, just the fix default value `./test` can be replaced by `directories.test` of "package.json".

boneskull:
PROTIP: `npx npm-merge-driver install --global` will install a thing that helps w/ `package-lock.json` conflicts.

bin/_mocha
```diff
@@ -18,7 +18,9 @@ const utils = Mocha.utils;
 const interfaceNames = Object.keys(Mocha.interfaces);
 const join = path.join;
 const cwd = process.cwd();
-const getOptions = require('./options');
+const options = require('./options');
```
boneskull: fair warning: changes to this file (and `bin/options.js`) are incompatible with #3556 due to the default test directory defined via other means (there's a `lib/mocharc.json` with Mocha's default settings)

bin/options.js
```diff
@@ -3,23 +3,58 @@
...
  * @type {string}
  * @default
  */
-const defaultPathname = 'test/mocha.opts';
+const DEFAULT_OPTS_FILENAME = 'mocha.opts';
+
+/**
+ * Reads directories.test of package.json.
+ *
+ * @public
```
boneskull: I think we should make a rule that every new method should have tag `@private` unless there's a good reason to make it `@public`, it's been discussed.

heck... it'd be nice to have an ESLint rule to enforce this (require `@public` or `@private` in docstrings).. maybe #3583 will help there.

bin/options.js
```diff
@@ -3,23 +3,58 @@
...
+ * @returns {string} test directory
+ */
+let cachedTestDirectory;
+
+function getTestDirectory() {
+  if (!cachedTestDirectory) {
```
boneskull: know what though?  if you use `require('/path/to/some/package.json')`, Node.js *will cache it for you.*  that's another reason why `findup-sync` makes more sense here, IMO--it just gives you a path.

test/integration/fixtures/package/subdir1/subdir1-1/package-with-testdir.fixture.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+var getTestDirectory = require('../../../../../../bin/options').getTestDirectory;
```
boneskull: 👍 🥇 💯 💥 🎉 😄 😃

PR #3588, opened 2018-11-29
remove unsupported @api tag in docstrings

boneskull:
This replaces `@api private` with `@private` and `@api public` (or just `@api`) with `@public`.

This will necessarily remove many functions from the automatically-generated API documentation, which you should be able to see on the netlify build below

By not publicly documenting Mocha's private API, we reduce the risk of consumers using it.


lib/context.js
```diff
@@ -11,14 +11,14 @@ module.exports = Context;
 /**
  * Initialize a new `Context`.
  *
- * @api private
+ * @private
```
plroebuck: Why would a module exported function be private?

plroebuck: Uh, don't think that's how JSDoc access is _supposed_ to be used though. It's done from the aspect of the module, not user.

plroebuck: Geez, you already merged this too?

PR #3589, opened 2018-11-29
add script to automatically dump CLI usage into documentation

boneskull:
this looks for `<!-- usage -->` and `<!-- usagestop -->` in `docs/index.md` and replaces everything between these two comments with the current output of `bin/mocha --help`.


docs/index.md
```diff
@@ -779,73 +779,77 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
 
 ## Usage
 
-```console
+<!-- usage -->
+
+```plain
```
plroebuck: `console` is more appropriate, no?

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * The CLI included with markdown-toc doesn't support `bullets`
+ * and `maxdepth` options, so that's why this exists.
+ */
+
+'use strict';
+
+const toc = require('markdown-toc');
+const miss = require('mississippi');
```
plroebuck: Have plans for the other `mississippi` dependencies? Know we use `through2` for browser bundle. Used `pump` and `concat-stream` below. Other 7 unused dependencies worth it?

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+              maxdepth: 2
+            })
+            .replace(/\n\n$/, '\n')
+        )
+      ),
+      err => {
+        if (err) {
+          return reject(err);
+        }
+        console.error(`Updated TOC in ${docsFilepath}.`);
```
plroebuck: If we're in error section, did we update the TOC or fail to do so?

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+  });
+}
+
+/**
+ * Inserts/updates command-line usage info into the documentation
+ * @returns Promise<void>
+ */
+function updateUsage() {
+  return new Promise((resolve, reject) => {
+    console.error(`Updating command-line usage in ${docsFilepath}...`);
```
plroebuck: `console.log` more appropriate for non-error messages

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+            if (m) {
+              trailingNewlines = m[0];
+            }
+            const sections = input
+              .split(/(?:<!--\s+usage(?:\s*stop)?\s+-->)/g)
+              .map(v => v.trim());
+            if (sections.length > 3) {
+              throw new Error('invalid; multiple "usage" sections found');
+            }
+            if (sections.length === 3) {
```
plroebuck: Be nice to have comment as to why invoking `--help` might have a different number of sections.

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+            } else {
+              throw new Error(
+                `could not find placeholder to update usage in ${docsFilepath}`
+              );
+            }
+
+            fs.writeFileSync(
+              docsFilepath,
+              sections.join('\n\n') + trailingNewlines
+            );
```
plroebuck: Would have expected the "Updated usage" message here...

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+          }),
+          err => {
+            if (err) {
+              return reject(err);
+            }
+            console.error(`Updated command-line usage in ${docsFilepath}`);
```
plroebuck: Again, did the update occur. If so, why here?

docs/index.md
```diff
@@ -779,73 +779,77 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
...
+--forbid-only                           causes test marked with only to fail the suite
+--forbid-pending                        causes pending tests and test marked with skip to fail the suite
+--file <file>                           adds file be loaded prior to suite execution (default: [])
+--exclude <file>                        adds file or glob pattern to ignore (default: [])
+-h, --help                              output usage information
 
 Commands:
-  init <path>                             initialize a client-side mocha setup at <path>
+init <path>                             initialize a client-side mocha setup at <path>
 ```
```
plroebuck: Why was indentation removed? Running `--help` on most programs displays cmdline options with indentation.

docs/index.md
boneskull: No:

image

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,144 @@
...
+
+const docsFilepath = path.join(__dirname, '..', 'docs', 'index.md');
+
+/**
+ * Inserts/updates a table of contents into the documentation
+ * @returns Promise<void>
+ */
+function updateTOC() {
+  return new Promise((resolve, reject) => {
+    console.error(`Updating TOC in ${docsFilepath}...`);
```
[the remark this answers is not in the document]
boneskull: it doesn't really matter much here, but STDOUT is for program output and STDERR is for logs, errors, etc.

the litmus test is "does it make sense to pipe this to another program?" 

the answer in this case is no, but there's no output otherwise.

scripts/preprocess-docs.js
boneskull: we use `through` like you said, then `concat`, and `pipe` here, and `pipe` uses `finish`.

it's a dev dep.  guess I don't care too much about adding new dev deps.

scripts/preprocess-docs.js
boneskull: we're not in an error-handler at this point?

scripts/preprocess-docs.js
boneskull: that is a good point, because I had to figure it out from the code in `markdown-toc` that I ripped off to do this.

scripts/preprocess-docs.js
boneskull: the last function to `mississippi.pipe()` is a "finish" handler, which is not necessarily an error condition.

docs/index.md
boneskull: probably has to do with the lack of a tty somewhere.  

the actual `bin/mocha --help` output has indents.

docs/index.md
plroebuck: why? you're displaying console output.

docs/index.md
plroebuck: That's what I thought this was supposed to do -- run `mocha --help` and paste its _exact_ content.

scripts/preprocess-docs.js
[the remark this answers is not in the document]
plroebuck: so `process.stderr.write` is your actual intent...

scripts/preprocess-docs.js
plroebuck: Guess I misunderstood what (L46) `err => {` meant. We appear to be in an error handler.

scripts/preprocess-docs.js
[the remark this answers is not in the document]
boneskull: no.. `console.error` which writes to STDERR  is preferable unless you really need direct stream access. among other reasons, if you want to play with control characters (like terminal spinners) or just don’t want the new line console gives you.  but that’s not the case here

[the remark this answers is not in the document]
boneskull: fwiw  I find `console.error` to be semantically confusing at best. but there’s no alias for it unless we create our own `Console` subclass

scripts/preprocess-docs.js
boneskull: I added some comments here

scripts/preprocess-docs.js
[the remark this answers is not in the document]
plroebuck: IMO, just seems poor form to use the semantically named `console.error` for
non-sematic purpose (i.e., to write to `stderr` rather than `stdout` rather than log an error).

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,149 @@
...
+            const sections = input
+              .split(/(?:<!--\s+usage(?:\s*stop)?\s+-->)/g)
+              .map(v => v.trim());
+
+            if (sections.length > 3) {
+              throw new Error('invalid; multiple "usage" sections found');
+            }
+
+            if (sections.length === 3) {
+              // this is triggered if we already have stuff in between the comments
```
plroebuck: So if a usage codefence is already present? When would that _not_ be the case?

boneskull: no, meaning, if we have

```
<!-- usage -->
STUFF
<!-- usagestop -->
```

we hit the first branch, otherwise if we have this (whitespace or nothing):

```
<!-- usage -->

<!-- usagestop -->
```

we hit the second.

scripts/preprocess-docs.js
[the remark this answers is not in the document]
boneskull: also this is important to understand:  https://nodejs.org/api/process.html#process_a_note_on_process_i_o

so it's become a habit to use `console.error()` when I want immediate output.

that being said, it doesn't really matter, so I'll just use `console.log`

docs/index.md
boneskull: because of the wacko highlighting (my console doesn't highlight like that).  using `plain`:

image

boneskull:
merging this; if there are further concerns, please send edits

scripts/preprocess-docs.js
plroebuck: > no, meaning, if we have
> 
> ```
> 
> STUFF
> 
> ```
> 
> we hit the first branch, otherwise if we have this (whitespace or nothing):
> 
> ```
> 
> 
> 
> ```
> 
> we hit the second.

Well, that _was_ what I assumed, where STUFF=code fence. Just not following when  the second case would ever occur.

scripts/preprocess-docs.js
```diff
@@ -0,0 +1,149 @@
...
+            console.log(`Updated command-line usage in ${docsFilepath}`);
+            resolve();
+          }
+        );
+      }
+    );
+  });
+}
+
+updateTOC().then(updateUsage());
```
plroebuck: Don't you need a `catch` here?
```js
updateTOC().then(updateUsage()).catch((err) => {
  const script = path.basename(
    require.main.filename, path.extname(require.main.filename)
  );
  console.error(`${script}:`, err);
});
```

package.json
```diff
@@ -512,6 +513,7 @@
     "markdownlint-cli": "^0.9.0",
     "nps": "^5.7.1",
     "nyc": "^11.7.3",
+    "mississippi": "^3.0.0",
```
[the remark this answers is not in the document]
plroebuck: Why not done alphabetically?

PR #3597, opened 2018-12-05
use markdown-magic for processing documentation

boneskull:
Sorry about the churn here, as I recently updated these scripts.

I happened across [markdown-magic](https://npm.im/markdown-magic) which looked like an easier way to do the stuff we're doing to transform `docs/index.md`.

- adopted `markdown-magic` for preprocessing
- added `scripts/markdown-magic.config.js` with transforms
- renamed some scripts in `package-scripts.js`
- removed `scripts/preprocess-docs.js`
- removed `common-tags` dep
- removed `mississippi` dep
- re-added `through2` dep
- added `markdown-magic-package-json` to easily dump stuff from `package.json` into the document; added an example of this at the end of `docs/index.md` with the `gitter` property

Also, @plroebuck, whatever was causing the indents in the usage output to not appear stopped happening; the indents are back!


docs/index.md
```diff
@@ -93,7 +93,7 @@ or as a development dependency for your project:
 $ npm install --save-dev mocha
 ```
 
-> Mocha currently requires Node.js v6.x or newer.
+> As of v6.0.0, Mocha requires Node.js v6.0.0 or newer.
```
boneskull: this reminds me; we really need to not publish the site on commits to master.  I have no idea how to do that

scripts/markdown-magic.config.js
```diff
@@ -0,0 +1,45 @@
...
+ * Add autogenerated stuff to our docs (`docs/index.md`)
+ * @see https://npm.im/markdown-magic
+ */
+
+const {execSync} = require('child_process');
+const markdownToc = require('markdown-toc');
+const path = require('path');
+
+exports.transforms = {
+  usage: (content, options) => {
```
boneskull: I could probably extract this into its own plugin.

PR #3602, opened 2018-12-08
modified SIGINT handling #3570

Jayasankar-m:
### Requirements
* Fix for 'Unable to terminate tests manually, e.g via Ctrl+C when running tests in a terminal.'
* refer issue #3570 

### Description of the Change
* Updated the SIGINT handler.
* A function property set to check if 'SIGINT' was called first time.
* if 'SIGINT" is called first time,keep the current behavior. 
* invoke 'SIGINT' again after 2 seconds which will kill the current process if still running. 

### Applicable issues
see #3570

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -196,19 +196,15 @@ exports.singleRun = (mocha, {files = [], exit = false} = {}) => {
   mocha.files = files;
   const runner = mocha.run(exit ? exitMocha : exitMochaLater);
 
-  process.on('SIGINT', function forceStop() {
-    if (forceStop.lock === undefined) {
-      forceStop.lock = true;
-      debug('aborting runner');
-      runner.abort();
+  process.once('SIGINT', () => {
+    debug('aborting runner');
+    runner.abort();
 
-      // This is a hack:
-      // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
-      // The amount of failures will be emitted as error code later
-      runner.failures = 130;
-      // Reraising the signal
-      setImmediate(() => process.kill(process.pid, 'SIGINT'));
-    }
+    // This is a hack:
+    // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+    // The amount of failures will be emitted as error code later
+    runner.failures = 130;
+    setImmediate(() => process.kill(process.pid, 'SIGINT'));
   });
 };
 
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -196,14 +196,19 @@ exports.singleRun = (mocha, {files = [], exit = false} = {}) => {
   mocha.files = files;
   const runner = mocha.run(exit ? exitMocha : exitMochaLater);
 
-  process.on('SIGINT', () => {
-    debug('aborting runner');
-    runner.abort();
+  process.on('SIGINT', function forceStop() {
+    if (forceStop.lock === undefined) {
+      forceStop.lock = true;
+      debug('aborting runner');
+      runner.abort();
 
-    // This is a hack:
-    // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
-    // The amount of failures will be emitted as error code later
-    runner.failures = 130;
+      // This is a hack:
+      // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+      // The amount of failures will be emitted as error code later
+      runner.failures = 130;
+      // Reraising the signal
+      setImmediate(() => process.kill(process.pid, 'SIGINT'));
+    }
   });
 };
 


boneskull:
Thanks for the help, but I think we have a fix incoming.  I'll leave this open until I can confirm.

bin/_mocha
```diff
@@ -642,11 +642,19 @@ if (program.watch) {
...
+
+    // This is a hack:
+    // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+    // The amount of failures will be emitted as error code later
+    runner.failures = 130;
+    forceStop.notCalled = false;
+    // invoke SIGINT after 2 seconds
+    setImmediate(() => setTimeout(() => process.emit('SIGINT'), 2000));
+  } else {
+    console.log('    forcefully killing process : ' + process.pid);
```
plroebuck: Unnecessary to log this; remove.

plroebuck: Add:
```js
forceStop.reraised = true;
```

bin/_mocha
```diff
@@ -642,11 +642,19 @@ if (program.watch) {
...
 
-process.on('SIGINT', () => {
-  runner.abort();
-
-  // This is a hack:
-  // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
-  // The amount of failures will be emitted as error code later
-  runner.failures = 130;
+process.on('SIGINT', function forceStop() {
+  if (forceStop.notCalled === undefined) {
```
plroebuck: Maybe `forceStop.lock`
If you're going here, this should be:
```js
  if (forceStop.lock === undefined) {
    forceStop.lock = true;
    ...
```

bin/_mocha
```diff
@@ -642,11 +642,19 @@ if (program.watch) {
...
+    // This is a hack:
+    // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+    // The amount of failures will be emitted as error code later
+    runner.failures = 130;
+    forceStop.notCalled = false;
+    // invoke SIGINT after 2 seconds
+    setImmediate(() => setTimeout(() => process.emit('SIGINT'), 2000));
+  } else {
+    console.log('    forcefully killing process : ' + process.pid);
+    process.kill(process.pid, 'SIGINT');
```
plroebuck: Actually, the entire`else` seems unnecessary.

bin/_mocha
```diff
@@ -642,11 +642,19 @@ if (program.watch) {
...
+process.on('SIGINT', function forceStop() {
+  if (forceStop.notCalled === undefined) {
+    runner.abort();
+
+    // This is a hack:
+    // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+    // The amount of failures will be emitted as error code later
+    runner.failures = 130;
+    forceStop.notCalled = false;
+    // invoke SIGINT after 2 seconds
```
plroebuck: So why wait? Mocha processing is done -- timeout `0`.

Jayasankar-m: Thank you!!
Seems there is a lot of change happening in codebase..
Have modified based on the comments..

lib/cli/run-helpers.js
```diff
@@ -204,6 +204,7 @@ exports.singleRun = (mocha, {files = [], exit = false} = {}) => {
     // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
     // The amount of failures will be emitted as error code later
     runner.failures = 130;
+    setImmediate(() => process.kill(process.pid, 'SIGINT'));
```
boneskull: This fires the signal recursively?

It doesn't work for me, FWIW...

plroebuck: If you read the GNU signal handling link I posted, reraising the signal is correct.

Jayasankar-m: @boneskull 
@plroebuck  -  Updated to reraise the signal once as specified in the comment

boneskull: @plroebuck That's fine, but nowhere does it say you're supposed to trigger your custom *handler* again.  Which is what's happening if you debug this.  ... at least on my machine.

lib/cli/run-helpers.js
```diff
@@ -196,14 +196,19 @@ exports.singleRun = (mocha, {files = [], exit = false} = {}) => {
...
-    // This is a hack:
-    // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
-    // The amount of failures will be emitted as error code later
-    runner.failures = 130;
+      // This is a hack:
+      // Instead of `process.exit(130)`, set runner.failures to 130 (exit code for SIGINT)
+      // The amount of failures will be emitted as error code later
+      runner.failures = 130;
+      // Reraising the signal
+      setImmediate(() => process.kill(process.pid, 'SIGINT'));
```
boneskull: you don't need `lock`, just stop listening:

```js
process.on('SIGINT', function forceStop() {
  debug('aborting runner');
  runner.abort();
  runner.failures = 130;
  setImmediate(() => process.kill(process.pid, 'SIGINT'));
  process.removeListener('SIGINT', forceStop);
});
```

lib/cli/run-helpers.js
boneskull: @Jayasankar-m has addressed this in 956dc59, but I have [a suggestion](https://github.com/mochajs/mocha/pull/3602#pullrequestreview-184921208)

lib/cli/run-helpers.js
boneskull: or even `process.once`!

Jayasankar-m: Updated as per the suggestion

boneskull:
LGTM.  I'll add a test.

bin/_mocha
plroebuck: Take it back. This is Node, not C -- unneeded.

lib/cli/run-helpers.js
boneskull: alright, I apologize, this is coming back to me now.

The intent here is to cleanly exit, which it seems we just ditched with the PR.

We call `runner.abort()`, which simply sets a flag in the runner.  *Once the current test completes*, the runner encounters this flag, and exits the process with an exit code reflecting the number of test failures.

This, of course, conflicts with the intent of *this* PR, which was to *not* let the currently-running test complete, and just exit.

I made this change so long ago, I'm not sure about the reasoning behind the "hack"; pretty sure that was me.  Nor am I sure of the reasons behind the intent of a clean exit, but that's the intent nonetheless.

It's likely we've broken the "exit code with # of failures" feature here (which I realize you do not like), but only if we interrupt the process.  

This seems OK to me, not recalling the original reasons for doing this, but *also* means the handler itself is useless and should be removed.

boneskull: OK, it comes from #2438 by @Munter.

We're exiting with `130` to ensure the expected behavior--exiting with code `130`--is retained while the handler disrupts the default behavior.

But because we're now actually invoking the default behavior, we don't need to call `runner.abort()` (because the process will immediately exit), and we *don't* need to set `runner.failures` (because the process will immediately exit) and we *don't* need to set `process.exitCode` if we just remove the damn `SIGINT` handler.

I'll send a PR to this effect.

PR #3609, opened 2018-12-12
add DCO 1.1; remove reference to CLA

boneskull:
We're able to ditch the CLA (yay!) for a DCO, which (IMO) is a lot less painful for developers to deal with.

I've already configured [probot/dco](/probot/dco) to handle the checks; once this is ready, it will be a requirement to merge any PR across the org.

The most straightforward way I've found to signoff automatically is just to create an alias, e.g., `git config alias.ci 'commit -s'`, then use `git ci` to commit.

Often you can configure GUIs such as SourceTree to do this automatically, as well.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -64,12 +64,13 @@ Follow these steps to get going.  If you are having trouble, don't be afraid to 
 1. Navigate to the source repository.  You should see a notification about your recent changes in your fork's branch, with a button to create a pull request.  Click it.
 1. Describe your changes in detail here.  Once you're satisfied, submit the form.
     - *PRO TIP*: If you've used a multi-line commit message, Github will pre-fill the PR's description with it.
-1. If you have not signed our Contributor License Agreement, a friendly robot will prompt you to do so.  A [CLA](https://cla.js.foundation/mochajs/mocha) (electronic) signature is **required** for all contributions of code to Mocha.
+1. You must accept the [DCO](#developers-certificate-of-origin-11).  To do this, either manually append `Signed-off-by: <your@email.com>` as the last line of your commit message, or use the `-s` flag with `git commit`, e.g., `git commit -s -m 'a commit message'".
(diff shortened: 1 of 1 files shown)


boneskull:
cc @mochajs/core since this is kind of Important

boneskull:
This should affect *new* PRs going forward.  If the CLA is already signed, I don't want to require the DCO as well.  Contributors of existing PRs are welcome to sign off on their commits, of course, but we should not require this.

craigtaub:
Mind if i ask why you think its less painful? 
From trust perspective it seems the better choice, but from ease-of-use isn't it more work that first time with a DCO? 
Just trying to understand the reason behind this.

outsideris:
Should we remove CLA check when we merge this PR?

boneskull:
@outsideris 	that's the idea, yeah.

still would appreciate more input esp. from @plroebuck 	and @munter.

plroebuck:
So if someone commits something that would have been illegal, under CLA they'd have been responsible. Under DCO, who is? Can both of them be used without conflict?

I prefer CLA scheme we already have myself, but don't care excessively either way.
Also seems easier to press a website button once than have to sign every git commit.

boneskull:
> So if someone commits something that would have been illegal, under CLA they'd have been responsible. Under DCO, who is? Can both of them be used without conflict?

Who is ultimately responsible is an open question, as there's no legal precedent yet (to my knowledge).

> I prefer CLA scheme we already have myself, but don't care excessively either way.
> Also seems easier to press a website button once than have to sign every git commit.

Let's keep in mind that the people who would find CLAs to be an impediment to contributing are likely *not* seeing this PR.

I do see the benefits of a CLA if you've already signed it.

That said, I don't want to flip-flop back and forth.  Maybe it'd be workable to allow for both?  I'll follow up on that with JSF legal.

JoshuaKGoldberg:
Closing to keep our review queue small. The current CLA system works and nobody from OpenJS has yelled at us to change it. 🙂

PR #3613, opened 2018-12-13
make lib/stats-collector a private API

boneskull:
Unless we have a good reason to do so, I think new APIs should default to private.

@craigtaub what are your thoughts?  Is there any valid use case for this (outside of how we're using it) that you can think of?

`semver-minor` since it's new stuff anyway, but doesn't really matter.

### DIFF
--- a/lib/stats-collector.js
+++ b/lib/stats-collector.js
@@ -1,8 +1,15 @@
 'use strict';
 
 /**
+ * Provides a factory function for a {@link StatsCollector} object.
+ * @private
+ * @module
+ */
+
+/**
  * Test statistics collector.
  *
+ * @private
  * @typedef {Object} StatsCollector
  * @property {number} suites - integer count of suites run.
  * @property {number} tests - integer count of tests run.
@@ -15,14 +22,16 @@
  */
 
 /**
- * Provides stats such as test duration,
- * number of tests passed / failed etc.
+ * Provides stats such as test duration, number of tests passed / failed etc., by listening for events emitted by `runner`.
  *
- * @public
- * @memberof Mocha
- * @param {Runner} runner
+ * @private
+ * @param {Runner} runner - Runner instance
+ * @throws {TypeError} If falsy `runner`
  */
 function createStatsCollector(runner) {
+  /**
+   * @type StatsCollector
+   */
   var stats = {
     suites: 0,
     tests: 0,


lib/stats-collector.js
```diff
@@ -1,8 +1,15 @@
...
+/**
+ * Provides a factory function for a {@link StatsCollector} object.
+ * @private
+ * @module
+ */
+
 /**
  * Test statistics collector.
  *
+ * @private
```
plroebuck: This would be `@public` since this is how reporters (both ours and 3rd party) get the statistics from the run. But the tag is not meaningful in a @typedef.

lib/stats-collector.js
```diff
@@ -15,14 +22,16 @@
...
 
 /**
- * Provides stats such as test duration,
- * number of tests passed / failed etc.
+ * Provides stats such as test duration, number of tests passed / failed etc., by listening for events emitted by `runner`.
  *
- * @public
- * @memberof Mocha
- * @param {Runner} runner
+ * @private
```
plroebuck: I have attempted to ask in multiple places now. Why are you changing everything to `@private`? Attempting to make this mean "user no touchy" is **not** what the tag is for -- it's meant to document the _module_, not the _user_ access.

For example, it's considered incorrect for _anything_ exported from a module to be private [Sorry, lost my reference], and there was an issue to make this official.

Closure Compiler documents the module POV as well. [If tagged `@private`, it's **unavailable** outside _this_ file](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#private).

boneskull: assigning @plroebuck for response

boneskull: though I suppose it's actually closed.

craigtaub: Perhaps we should think of both from now on? (E.g stats-collector @public and @ignore).
It's usful to have code documented even if it doesn't show in generated source.

boneskull: @craigtaub if we do that then it's fair game as a public API, even if it's not documented on the site as such.

What I like about the `@private` solution is that you can generate "complete" API docs with a flag.

regardless of any other solutions we decide on, I feel `@ignore` is appropriate for functions that aren't exported.

craigtaub: Fair point. I don't really mind what we decide as long as we can be consistent.

PR #3616, opened 2018-12-14
add tests for SIGINT handling

boneskull:
See #3602. cc @Jayasankar-m

### DIFF
--- a/test/integration/sigint.spec.js
+++ b/test/integration/sigint.spec.js
@@ -1,0 +1,45 @@
+'use strict';
+
+var helpers = require('./helpers');
+var invokeMocha = helpers.invokeMocha;
+var invokeSubMocha = helpers.invokeSubMocha;
+
+var fixturePath = require.resolve('./fixtures/sigint.fixture');
+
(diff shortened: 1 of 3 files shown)


test/integration/sigint.spec.js
```diff
@@ -0,0 +1,45 @@
...
+var helpers = require('./helpers');
+var invokeMocha = helpers.invokeMocha;
+var invokeSubMocha = helpers.invokeSubMocha;
+
+var fixturePath = require.resolve('./fixtures/sigint.fixture');
+
+describe('SIGINT handling', function() {
+  describe('using "mocha" wrapper', function() {
+    describe('when SIGINT received', function() {
+      it('should terminate itself w/ SIGTERM', function(done) {
```
plroebuck: How/why did we switch from SIGINT to SIGTERM?
This by itself seems wrong...

boneskull: I don't actually remember why other than it was that `SIGINT` just wasn't cutting it.

FWIW, I haven't heard complaints about it since...

PR #3617, opened 2018-12-14
Behavior of after/afterEach hooks with --bail flag

juergba:
### Description

It hasn't been clearly defined yet wether `after`/`afterEach` hooks should run after a failing test. Since merging issue #3278 `after`/`afterEach` hooks are not beeing executed anymore.
This was a big change in event behavior and quite a few users complained about.
Reading many posts (of maintainers also) I dare to claim that those hooks should run in order to guarantee a correct cleanup.

current pattern with failing test:
- --bail: the `end` event is triggered twice (which is bug). The first time `after`/`afterEach` hooks are not executed, with the second `end` event they are actually called.
- --bail and --exit: none of the `after`/`afterEach` hooks is called.

proposed pattern: (required changes are marked **bold**)
- `before` hook:
  - `bail: false`: failed `before` hook skips all tests in current suite/subsuites, and jumps to corresponding `after` hook. Subsequent suites/tests are executed.
  - `bail: true`: **as above**, but all subsequent suites/tests are skipped.
- `before each` hook: 
  - `bail: false`: failed `before each` hook skips remaining tests in current suite/subsuites, and jumps to corresponding `after each` hook, then `after`hook. Subsequent suites/tests are executed.
  - `bail: true`: **as above**, but all subsequent suites/tests are skipped.
- `test`: 
  - `bail: false`: failed `test` does not alter execution order.
  - `bail: true`: failed `test` **jumps to corresponding `after each` hook, then `after`hook**. All remaining tests in current suite, subsuites  and subsequent suites are skipped.
- `after` hook:
  - `bail: false`: failed `after` hook does not alter execution order.
  - `bail: true`:  failed `after` hook executes remaining `after` hooks, then all subsequent suites are skipped.
- `after each` hook:
  - `bail: false`: failed `after each` hook executes remaining `after each` and `after` hooks, but skips remaining tests in current suite/subsuites. Subsequent suites/tests are executed.
  - `bail: true`: **as above**, but all subsequent suites/tests are skipped.

#### 1. remove added `end` event: 
Emitting an `end` event just somewhere in a test tree doesn't work correctly. The runner is not aborted and keeps going, until a second `end` event (the real one) is emitted.
```
describe('level A', () => {
    beforeEach(() => {
      console.log('beforeEach A');
    });  
    it('test A should fail', () => {
      throw new Error();
    });
    describe('level A2', () => {
        it('should log "test A2', () => {
            console.log('test A2');
        });
    });
    afterEach(() => {
        console.log('afterEach A');
      });
});
```
with --bail flag and "debug":
```
mocha:runner start +1ms
  mocha:runner run suite  +1ms
  mocha:runner run suite level A +4ms
  level A
beforeEach A
    1) test A should fail

  0 passing (7ms)
  1 failing
  1) level A
       test A should fail:
  Error
      at Context.it [...]

  mocha:runner end +10ms                      <<<<<< first end: runner doesn't abort

afterEach A
  mocha:runner run suite level A level A2 +1ms
  mocha:runner finished running +2ms
  mocha:runner end +0ms                       <<<<<< second end
```
#### 2. Correct hook pattern after failing `after each` and `after` hooks:
All remainig `after each` and `after` hooks - the entire suite tree upwards - have to be executed.

### Applicable issues

#3398 
#3598 
#3457

### DIFF
--- a/test/integration/fixtures/options/bail.fixture.js
+++ b/test/integration/fixtures/options/bail.fixture.js
@@ -13,11 +13,11 @@ describe('suite1', function () {
 });
 
 describe('suite2', function () {
-  before(function (done) {
+  before(function () {
     throw new Error('this hook should not be displayed');
   });
 
-  it('should not display this error', function (done) {
+  it('should not display this error', function () {
     throw new Error('this should not be displayed');
   });
 });
--- a/docs/index.md
+++ b/docs/index.md
@@ -907,7 +907,7 @@ Enforce a rule that tests must be written in "async" style, meaning each test pr
 
 ### `--bail, -b`
 
-Causes Mocha to stop running tests after the first test failure it encounters.
+Causes Mocha to stop running tests after the first test failure it encounters. Corresponding `after()` and `afterEach()` hooks are executed for potential cleanup.
 
 `--bail` does *not* imply `--exit`.
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -241,16 +241,14 @@ Runner.prototype.fail = function(test, err) {
   }
 
   this.emit('fail', test, err);
-  // if (this.suite.bail()) {
-  //   this.emit('end');
-  // }
 };
 
 /**
  * Fail the given `hook` with `err`.
  *
  * Hook failures work in the following pattern:
- * - If bail, then exit
+ * - If bail, run corresponding `after each` and `after` hooks,
+ *   then exit
  * - Failed `before` hook skips all tests in a suite and subsuites,
  *   but jumps to corresponding `after` hook
  * - Failed `before each` hook skips remaining tests in a
@@ -494,7 +492,7 @@ Runner.prototype.runTests = function(suite, fn) {
   function next(err, errSuite) {
     // if we bail after first err
     if (self.failures && suite._bail) {
-      return fn();
+      tests = [];
     }
 
     if (self._abort) {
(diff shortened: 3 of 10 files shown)


test/integration/fixtures/options/bail.fixture.js
```diff
@@ -13,11 +13,11 @@ describe('suite1', function () {
 });
 
 describe('suite2', function () {
-  before(function (done) {
+  before(function () {
```
boneskull: was this causing a problem?  perhaps your changes should actually be additions; two tests that use `done` and two that don't.  this would give us confidence that the same thing works properly regardless of the Runnable's strategy (async or sync)

boneskull:
This is `semver-patch` since it effectively reverts `semver-major`-type changes introduced in #3278, so both are bug fixes.

After this is accepted, let's please evaluate if #3398, #3598, and #3457 are no longer reproducible.

test/integration/fixtures/options/bail.fixture.js
juergba: no, no problem arose, the 'suite2' describe is skipped (correctly) anyway. I added a second test, one sync and one async.

boneskull:
closes #3398, closes #3598, closes #3457

boneskull:
@juergba 	thanks!  awesome.

PR #3580, opened 2018-11-26
Fix an error when fn===undefined

Lana-Light:
This change fixes an error that occurs when reporter is 'xunit' and run is called without callback: " /rbd/pnpm-volume/a49b9a7c-c3f0-42ba-ae66-612ead86e96c/node_modules/.registry.npmjs.org/mocha/5.2.0/node_modules/mocha/lib/reporters/xunit.js:126
 fn(failures);  ^
TypeError: fn is not a function "

### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions.

PR #3620, opened 2018-12-15
Add test for `mocha.reporter('xunit').run()`

Lana-Light:
### Description of the Change

It will test function `mocha.run` with reporter `xunit`.

### Applicable issues
see #3580

### DIFF
--- a/test/unit/mocha.spec.js
+++ b/test/unit/mocha.spec.js
@@ -55,6 +55,15 @@ describe('Mocha', function() {
       mocha.run(done);
     });
   });
+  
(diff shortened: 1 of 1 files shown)


boneskull:
This build cannot succeed.  Please move this changeset into your `patch-1` branch (PR #3580).  You can do this by cherry-picking the changeset, OR reimplementing this PR, then committing on top of `patch-1`.  Then push `patch-1`.  That should update #3580, and hopefully we'll have a passing build.  I'll close this PR.

thanks

PR #3623, opened 2018-12-18
various dependency upgrades pre-release

boneskull:
I've upgraded everything here *except* `karma-browserify` and `karma-saucelabs`, both of which were giving me fits recently.  I'll create an issue for those two.

Production deps should all be installed via `--save-exact`; should have a linter or something for that!

new "features" include:

- YAML linting
- remove `prettier-eslint-cli` and use `prettier` (JSON, YAML) and `eslint` w/ `eslint-plugin-prettier` for JS files, which rids us of a somewhat-inactive dependency.
- automatic stashing of unstaged changes on a *per-file basis* with Husky upgrade

`debug` might not make it to v4.x if it doesn't work properly in IE11.


test/integration/hooks.spec.js
```diff
@@ -42,10 +42,10 @@ describe('hooks', function() {
       ];
 
       expected.forEach(function(line, i) {
-        assert.equal(lines[i], line);
+        assert.strictEqual(lines[i], line);
```
outsideris: Is there a reason that you keep `assert` rather than `unexpected`.

boneskull: too lazy to use `expect`.  these need to be converted.

boneskull:
As I suspected, `debug` @ v4 is not compatible with IE11

PR #3628, opened 2018-12-18
add more tests for utils.deprecate

boneskull:
Mainly trying to get coverage up.

Also a tweak to Wallaby config, and one to coverage computation which will zap `.nyc_output/` between `npm test` runs to ensure the data isn't stale.


boneskull:
blah, gotta revert some of the refactors.

PR #3631, opened 2018-12-21
execute test after "before each" hook asynchronously; closes #3009

### LINKED ISSUE #3009: Impact of `setImmediate` between hooks on Promises
### Prerequisites

### Description

I'm opening this issue as a result of an investigation into https://github.com/domenic/chai-as-promised/issues/173. However, there's a lot of unrelated noise in that thread, so I'll demonstrate the issue here independently of `chai-as-promised`.

Hooks typically wait until the next tick of the event loop before proceeding, per [these lines](https://github.com/mochajs/mocha/blob/82c9cb4855052f09d44e62ade32263e0514bf0cf/lib/runner.js#L338-L340). As a result, if a `Promise` with no rejection handler is declared in a hook prior to the test in which the rejection handler is added to the `Promise`, then a warning is typically generated about an unhandled promise rejection, followed by another warning about asynchronous handling of the promise rejection.

I used the word "typically" twice above because what I described doesn't apply to the final `beforeEach` prior to the test; unlike the other hooks, the final `beforeEach` proceeds immediately to the test, without waiting for the next tick of the event loop.

### Steps to Reproduce

**Example 1:**

```js
let promise;

describe("mySuite", function () {
  before(function () {
    promise = Promise.reject(new Error());
  });

  it("myTest", function () {
    promise.catch(function () {});
  });
});
```

**Example 2:**

```js
let promise;

describe("mySuite", function () {
  beforeEach(function () {  // This is the only different line
    promise = Promise.reject(new Error());
  });

  it("myTest", function () {
    promise.catch(function () {});
  });
});
```

**Expected behavior:** Although the tests are meaningless, ideally I'd expect neither of the examples to generate any unhandled promise rejection warnings, or at least for both examples to result in the same behavior.

**Actual behavior:** Example 1 generates an unhandled promise rejection and Example 2 doesn't.

**Example 1 output:**

> > mocha
> 
>   mySuite
> (node:14219) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error
> (node:14219) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
>     ✓ myTest
> (node:14219) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)
> 
>   1 passing (11ms)

**Example 2 output:**

> > mocha
> 
>   mySuite
>     ✓ myTest
> 
>   1 passing (7ms)

### Versions

Node v8.5.0
Mocha v3.5.3

### Additional Information

Changing [this line](https://github.com/mochajs/mocha/blob/82c9cb4855052f09d44e62ade32263e0514bf0cf/lib/runner.js#L92) to favor `process.nextTick` over `global.setImmediate` fixes the issue so that neither example generates a warning, and the Mocha test suite still passes. But I don't know what other consequences there are.

I think that in most situations this issue isn't a problem. Most promises being tested in the wild have (or should have) rejection handlers already registered, so no warning would be generated in those cases, despite the next tick between hooks. I just started digging in order to understand what was happening with https://github.com/domenic/chai-as-promised/issues/173. The example in that issue is completely contrived, as are the examples I provided here.

boneskull:
This provides consistent, though not necessarily ideal, behavior when a
`Promise` is rejected within a "before" or "before each" hook and *not*
caught synchronously or within a `process.nextTick()` callback.

Previously, the `Promise` would be considered handled if it was rejected
in a "before each" hook and caught within the test, because the test
would be executed synchronously.  This is not the same case for a
"before" hook, which would consider the `Promise` unhandled.  Now, both
result in an unhandled rejection warning from node.

An alternative implementation would be to execute tests within a
`process.nextTick()` callback (not sure what happens in a browser).
This would mean rejected `Promise`s in both cases would be considered
handled, because `process.nextTick()` would execute *before* whatever considers the `Promise` rejection to be unhandled.

Mocha executes tests recursively, and needs to do so
asynchronously in order to avoid exceeding the maximum call stack.  As
per the Node.js documentation [here](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/),
this is *not* a valid use case of `process.nextTick()`.

Yet another alternative implementation which may avoid a decision
entirely would be to execute tests using a generator.  This could only
be *simulated* in an environment such as IE11, and thus would
potentially result in inconsistent or undefined behavior when running
the same tests in different browsers (that being said, generators are
probably a Good Idea for Mocha to consume in the long run).

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -332,7 +332,9 @@ Runner.prototype.hook = function(name, fn) {
       }
       self.emit('hook end', hook);
       delete hook.ctx.currentTest;
-      next(++i);
+      Runner.immediately(function() {
+        next(++i);
+      });
     });
   }
 
--- a/test/integration/fixtures/hooks/before-hook-promise-rejection.fixture.js
+++ b/test/integration/fixtures/hooks/before-hook-promise-rejection.fixture.js
@@ -1,0 +1,13 @@
+'use strict';
+
+describe('mySuite', function() {
+  var promise;
+
+  before(function() {
+    promise = Promise.reject(new Error());
+  });
+
+  it('myTest', function() {
+    promise.catch(function() {});
+  });
+});
--- a/test/integration/fixtures/hooks/beforeEach-hook-promise-rejection.fixture.js
+++ b/test/integration/fixtures/hooks/beforeEach-hook-promise-rejection.fixture.js
@@ -1,0 +1,13 @@
+'use strict';
+
+describe('mySuite', function() {
+  var promise;
+
+  beforeEach(function() {
+    promise = Promise.reject(new Error());
+  });
+
+  it('myTest', function() {
+    promise.catch(function() {});
+  });
+});
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -30,6 +30,12 @@ module.exports = {
    * @param {Object} [opts] - Options for `spawn()`
    */
   runMocha: function(fixturePath, args, fn, opts) {
+    if (typeof args === 'function') {
+      opts = fn;
+      fn = args;
+      args = [];
+    }
+
     var path;
 
     path = resolveFixturePath(fixturePath);
@@ -61,6 +67,12 @@ module.exports = {
    * @returns {*} Parsed object
    */
   runMochaJSON: function(fixturePath, args, fn, opts) {
+    if (typeof args === 'function') {
+      opts = fn;
+      fn = args;
+      args = [];
+    }
+
     var path;
 
     path = resolveFixturePath(fixturePath);
@@ -208,6 +220,9 @@ function _spawnMochaWithListeners(args, fn, opts) {
 }
 
 function resolveFixturePath(fixture) {
+  if (path.extname(fixture) !== '.js') {
+    fixture += '.fixture.js';
+  }
   return path.join('test', 'integration', 'fixtures', fixture);
 }
 
--- a/test/integration/hook-err.spec.js
+++ b/test/integration/hook-err.spec.js
@@ -1,7 +1,8 @@
 'use strict';
 
-var runMocha = require('./helpers').runMocha;
-var splitRegExp = require('./helpers').splitRegExp;
+var helpers = require('./helpers');
+var runMocha = helpers.runMocha;
+var splitRegExp = helpers.splitRegExp;
 var bang = require('../../lib/reporters/base').symbols.bang;
 
 describe('hook error handling', function() {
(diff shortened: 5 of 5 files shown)


boneskull:
Would appreciate any sort of feedback from anybody on this, since it has the ability to fail previously-passing tests.  That's always a risky change.

boneskull:
Compounding the problem is that Node.js' behavior is to warn the user, but the tests will still pass.

Here's the output of a test fixture:

```
  mySuite
(node:40959) UnhandledPromiseRejectionWarning: Error
    at Context.<anonymous> (/Users/boneskull/projects/mocha/test/integration/fixtures/hooks/beforeEach-hook-promise-rejection.fixture.js:7:30)
    at callFn (/Users/boneskull/projects/mocha/lib/runnable.js:383:21)
    at Hook.Runnable.run (/Users/boneskull/projects/mocha/lib/runnable.js:375:7)
    at next (/Users/boneskull/projects/mocha/lib/runner.js:307:10)
    at Immediate.<anonymous> (/Users/boneskull/projects/mocha/lib/runner.js:342:5)
    at processImmediate (timers.js:632:19)
(node:40959) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or byrejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:40959) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
    ✓ myTest
(node:40959) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)


  1 passing (9ms)
```

craigtaub:
Really sorry if these are bad questions, i'm missing a load of context.
Is there a reason we would not want to fail the suite, like we do if a hook returned a rejected promise?
If its because there are valid scenarios for using a promise this way (or that its not possible from the code in its current form), thats fair enough.

juergba:
The behavior of "before" and "beforeEach" hooks should be different as it was before. The context/`this` is the same for "beforeEach", "it" and "afterEach", so a promise is handled within the same unit and there should be no warning or error.
Whereas for "before" and "it" the context/`this` is different.

juergba:
In nested tests (current master) even a "beforeEach" hook results in an unhandled rejection warning:
```
use strict';
let promise;
describe("mySuite", function () {
  beforeEach(function () {
    promise = Promise.reject(new Error());
    //  promise.catch(()=>{});          <<<<<<< avoids rejection warning
  });
  describe("mySuite nested", function () {
    it("myTest", function () {
      promise.catch(function () {console.log('catched')});
    });
  });
});
```
We are talking about an edge case, which isn't worth that risky change. The easiest way is: leave everything like it is, and in the documentation we instruct the user how to avoid this warning (in future probably unhandled error). Like inserted above.

boneskull:
I would expect @juergba's example code to work, but given that it doesn't (and my changes have no impact on it), this change is not a complete solution.

Happy to revisit, but whatever change we make here--if we make a change--it must fall in line with Mocha's philosophy around contexts (whatever that is).

boneskull:
this branch is available in `issue/3009` of [boneskull/mocha](/boneskull/mocha) for future reference.

boneskull:
Ref: #3671

PR #3632, opened 2018-12-21
Set "root" property on Suite correctly

### LINKED ISSUE #2755: Decouple the "root" property on Suite from title length.
Right now, Mocha deems a `Suite` to be a `root` if the title for it is empty, as can be seen [here](https://github.com/mochajs/mocha/blob/2bb2b9fa35818db7a02e5068364b0c417436b1af/lib/suite.js#L60):

```
function Suite (title, parentContext) {
  if (!utils.isString(title)) {
    throw new Error('Suite `title` should be a "string" but "' + typeof title + '" was given instead.');
  }
  this.title = title;
  // skipping non-pertinent stuff...
  this.root = !title;
```

Mocha creates a first `Suite` with `''` for title, which marks that suite as the root suite. This is a problem because, it seems to me when the user uses an empty string to name a suite, this should not result in Mocha building a structurally deficient tree of tests. (A tree with two "root" suites.)

Being able to name suites with the empty string (without at the same time marking them as "root") is useful. Sometimes I have suites where testing what is logically a single piece of functionality is best done by dividing the tests into two groups: simple tests that can be performed with super fast setup code, more complex tests that need a costlier setup. Oh, I could use the setup code for the 2nd group for all tests, but that would increase the total run time. I end up with something like:

```
// Testing the method foo on class Bar.
describe("#foo", () => {
  describe("", () => { 
    beforeEach(() => {
       // create a trivial data structure that is quick to create and sufficient for the tests in this group.
    });

    // tests...
  });

  describe("", () => { 
    beforeEach(() => {
       // create a complex structure that is costlier to create but needed for these tests.
    }); 

    // tests...
  });
});
```

The fact that the suite uses different setups for the two groups of tests is an internal detail that is not useful to know in test reports. The need for different test setups does not always correlate with divisions that are meaningful from the point of view of reporting successes or failures. It *often does*, but not always.

So I would suggest that the code be changed to determine the value of `root` through some other test than the length of the suite's title. There could be a unique object that serves as a marker to indicate "this suite I'm building is a root suite". It could be "statically" added to `Suite` (e.g. `Suite.Root` so doing `new Suite(Suite.Root, ...)` would result in a root suite).

Ultimately, though, the `root` property seems redundant to me: a `Suite` which has no `parent` is a root suite, no? So "root"-ness should correlate with the absence of a set `parent`. But maybe there's some scenario I'm missing? At any rate, removing `root` cannot be done without breakage. The `karma-mocha` plugin, for instance, relies on it to produce reports (which is how I discovered the problem).

juergba:
Currently every suite with an empty string title "" has set its property `suite.root === true`, see [here](https://github.com/mochajs/mocha/blob/master/lib/suite.js#L69). This is a bug since there can be only one root suite and a deficient test tree would be the result.
```js
'use strict';
var emptyTitleSuite1;

describe('suite1', function() {
  var rootSuite = this.parent;
  console.log('title rootSuite: %s, root: %s', rootSuite.title || '""', rootSuite.root);
  console.log('title suite1: %s, root: %s', this.title, this.root);

  describe('', function() {                 <<<<< empty title
    emptyTitleSuite1 = this;
    console.log('title emptyTitleSuite1: %s, root: %s', this.title || '""', this.root);
    console.log('rootSuite === emptyTitleSuite1:', rootSuite === this);
  });
});

describe('', function() {                   <<<<< empty title
  var rootSuite = this.parent;
  console.log('title emptyTitleSuite2: %s, root: %s', this.title || '""', this.root);
  console.log('rootSuite === emptyTitleSuite2:', rootSuite === this);
  console.log('emptyTitleSuite1 === emptyTitleSuite2:', emptyTitleSuite1 === this);
});
```
Above we have four different suites (incl. root), three of them with `suite.root === true`. 
```
title rootSuite: "", root: true                <<<< 1. root
title suite1: suite1, root: false
title emptyTitleSuite1: "", root: true         <<<< 2. root
rootSuite === emptyTitleSuite1: false

title emptyTitleSuite2: "", root: true         <<<< 3. root
rootSuite === emptyTitleSuite2: false
emptyTitleSuite1 === emptyTitleSuite2: false
```

### Description of the Change

Add an optional parameter "root" with default value `false` to the Suite constructor function. 

### Alternative

Don't allow empty string suite title, except for root suite.

### Applicable issues

closes #2755

### DIFF
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -41,16 +41,18 @@ exports.create = function(parent, title) {
 };
 
 /**
- * Initialize a new `Suite` with the given `title`, `ctx` and `root`. Derived from [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter)
+ * Constructs a new `Suite` instance with the given `title`, `ctx`, and `isRoot`.
  *
- * @memberof Mocha
  * @public
  * @class
- * @param {string} title
- * @param {Context} parentContext
- * @param {boolean} root
+ * @extends EventEmitter
+ * @memberof Mocha
+ * @see {@link https://nodejs.org/api/events.html#events_class_eventemitter|EventEmitter}
+ * @param {string} title - Suite title.
+ * @param {Context} parentContext - Parent context instance.
+ * @param {boolean} [isRoot=false] - Whether this is the root suite.
  */
-function Suite(title, parentContext, root) {
+function Suite(title, parentContext, isRoot) {
   if (!utils.isString(title)) {
     throw createInvalidArgumentTypeError(
       'Suite argument "title" must be a string. Received type "' +
@@ -71,7 +73,7 @@ function Suite(title, parentContext, root) {
   this._beforeAll = [];
   this._afterEach = [];
   this._afterAll = [];
-  this.root = root || false;
+  this.root = isRoot === true;
   this._timeout = 2000;
   this._enableTimeouts = true;
   this._slow = 75;
@@ -327,6 +329,7 @@ Suite.prototype.afterEach = function(title, fn) {
  */
 Suite.prototype.addSuite = function(suite) {
   suite.parent = this;
+  suite.root = false;
   suite.timeout(this.timeout());
   suite.retries(this.retries());
   suite.enableTimeouts(this.enableTimeouts());
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -41,15 +41,16 @@ exports.create = function(parent, title) {
 };
 
 /**
- * Initialize a new `Suite` with the given `title` and `ctx`. Derived from [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter)
+ * Initialize a new `Suite` with the given `title`, `ctx` and `root`. Derived from [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter)
  *
  * @memberof Mocha
  * @public
  * @class
  * @param {string} title
  * @param {Context} parentContext
+ * @param {boolean} root
  */
-function Suite(title, parentContext) {
+function Suite(title, parentContext, root) {
   if (!utils.isString(title)) {
     throw createInvalidArgumentTypeError(
       'Suite argument "title" must be a string. Received type "' +
@@ -70,7 +71,7 @@ function Suite(title, parentContext) {
   this._beforeAll = [];
   this._afterEach = [];
   this._afterAll = [];
-  this.root = !title;
+  this.root = root || false;
   this._timeout = 2000;
   this._enableTimeouts = true;
   this._slow = 75;
--- a/test/unit/suite.spec.js
+++ b/test/unit/suite.spec.js
@@ -390,7 +390,7 @@ describe('Suite', function() {
     describe('when there is a parent', function() {
       describe('the parent is the root suite', function() {
         it('returns the suite title', function() {
-          var parentSuite = new Suite('');
+          var parentSuite = new Suite('', {}, true);
           parentSuite.addSuite(this.suite);
           expect(this.suite.titlePath(), 'to equal', ['A Suite']);
         });
@@ -492,6 +492,16 @@ describe('Suite', function() {
       expect(function() {
         new Suite('Bdd suite', 'root');
       }, 'not to throw');
+    });
+
+    it('should be root suite', function() {
+      var rootSuite = new Suite('', {}, true);
+      expect(rootSuite.root, 'to be', true);
+    });
+
+    it('should not be root suite', function() {
+      var suite = new Suite('');
+      expect(suite.root, 'to be', false);
     });
   });
 
--- a/test/unit/suite.spec.js
+++ b/test/unit/suite.spec.js
@@ -302,6 +302,27 @@ describe('Suite', function() {
     });
   });
 
+  describe('.create()', function() {
+    before(function() {
+      this.first = new Suite('Root suite', {}, true);
+      this.second = new Suite('RottenRoot suite', {}, true);
+      this.first.addSuite(this.second);
+    });
+
+    it('does not create a second root suite', function() {
+      expect(this.second.parent, 'to be', this.first);
+      expect(this.first.root, 'to be', true);
+      expect(this.second.root, 'to be', false);
+    });
+
+    it('does not denote the root suite by being titleless', function() {
+      var emptyTitleSuite = Suite.create(this.second, '');
+      expect(emptyTitleSuite.parent, 'to be', this.second);
+      expect(emptyTitleSuite.root, 'to be', false);
+      expect(this.second.root, 'to be', false);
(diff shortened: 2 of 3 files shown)


boneskull:
there's only one root suite ever... it's the same one across *n* different files.

boneskull:
that said #2755 is perfectly valid.  just wanted to clear up any confusion

juergba:
@boneskull Thanks, but I'm confused now. Should I stop this PR, or go on?
In the sense of: "there **should be** only one root suite ever...

boneskull:
The PR is fine, I was just responding to your “can result in a tree with more than one root suite.” There can only be one root suite regardless of what reporters do with the title.

lib/suite.js
```diff
@@ -39,15 +39,16 @@ exports.create = function(parent, title) {
...
- * Initialize a new `Suite` with the given `title` and `ctx`. Derived from [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter)
+ * Initialize a new `Suite` with the given `title`, `ctx` and `root`. Derived from [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter)
  *
  * @memberof Mocha
  * @public
  * @class
  * @param {string} title
  * @param {Context} parentContext
+ * @param {boolean} root
  */
```
plroebuck: Let's fix the JSDoc while we're here...

```js
/**
 * Constructs a new `Suite` instance with the given `title`, `ctx`, and `isRoot`.
 *
 * @public
 * @class
 * @extends EventEmitter 
 * @memberof Mocha
 * @see {@link https://nodejs.org/api/events.html#events_class_eventemitter|EventEmitter} 
 * @param {string} title - Suite title.
 * @param {Context} parentContext - Parent context instance.
 * @param {boolean} isRoot - Whether this is the root suite.
 */
```

lib/suite.js
```diff
@@ -39,15 +39,16 @@ exports.create = function(parent, title) {
...
-function Suite(title, parentContext) {
+function Suite(title, parentContext, root) {
```
plroebuck: Would prefer boolean-like name `isRoot`

test/unit/suite.spec.js
```diff
@@ -390,7 +390,7 @@ describe('Suite', function() {
     describe('when there is a parent', function() {
       describe('the parent is the root suite', function() {
         it('returns the suite title', function() {
-          var parentSuite = new Suite('');
+          var parentSuite = new Suite('', {}, true);
```
plroebuck: How about the more specific `rootSuite` instead of `parentSuite`?

test/unit/suite.spec.js
```diff
@@ -493,6 +493,16 @@ describe('Suite', function() {
...
+
+    it('should be root suite', function() {
+      var rootSuite = new Suite('', {}, true);
+      expect(rootSuite.root, 'to be', true);
+    });
+
+    it('should not be root suite', function() {
+      var suite = new Suite('');
+      expect(suite.root, 'to be', false);
+    });
```
plroebuck: Maybe it's just me, but these two tests don't really add much value since they just verify your ability to set a property correctly (which admittedly has been done wrong in the past). Wouldn't a more meaningful test be to verify that only a single Suite can have the `root` property set `true`?

lib/suite.js
```diff
@@ -66,7 +67,7 @@ function Suite(title, parentContext) {
   this._beforeAll = [];
   this._afterEach = [];
   this._afterAll = [];
-  this.root = !title;
+  this.root = root || false;
```
plroebuck: Prefer insurance myself...
```js
   this.root = isRoot === true;
```

juergba: @plroebuck thanks, an alternative way could be:
1. empty title "" results in `root = true` (no changes so far)
2. in [Suite.create](https://github.com/mochajs/mocha/blob/master/lib/suite.js#L33): test for empty title and existing parent
```
exports.create = function(parent, title) {
  var suite = new Suite(title, parent.ctx);
  suite.parent = parent;
  if (!title && parent) {       <<< insert
    suite.root = false;         <<< insert
  }                             <<< insert
  title = suite.fullTitle();    // ???
  parent.addSuite(suite);
  return suite;
};
```
It would still be possible to manually create a suite-tree with several roots via `Suite.addSuite`, f.e. in external interfaces or test fixtures. As per now I would not prevent this, but could move `if (!title && parent) {...}` to `Suite.addSuite`.

Do you prefer this alternative solution?

plroebuck: Don't see how `Suite.create` would be helpful as "lib/mocha.js" invokes the `Suite` ctor directly. `Context` ctor from "lib/context.js" is _only_ used by `Mocha` ctor; if "flag" set there, then Mocha ctor could designate root suite based on it.

juergba: Mocha constructor is not changed in my last proposition, just `Suite.create`.

plroebuck: If it's still possible to create multiple root suites, it's still broken -- [there can be only one](https://www.youtube.com/watch?v=sqcLjcSloXs) -- the suite created in the `Mocha` ctor!

juergba: ok, done.
I added a line in `Suite.prototype.addSuite` which is actually the only change required for this PR.

test/unit/suite.spec.js
```diff
@@ -302,6 +302,27 @@ describe('Suite', function() {
...
   });
 
+  describe('.create()', function() {
+    before(function() {
+      this.first = new Suite('Root suite', {}, true);
+      this.second = new Suite('RottenRoot suite', {}, true);
+      this.first.addSuite(this.second);
+    });
+
+    it('sets second root property to false', function() {
```
plroebuck:
```js
     it('does not create a second root suite', function() {
```

test/unit/suite.spec.js
```diff
@@ -302,6 +302,27 @@ describe('Suite', function() {
...
+      expect(this.second.parent, 'to be', this.first);
+      expect(this.first.root, 'to be', true);
+      expect(this.second.root, 'to be', false);
+    });
+
+    it('only one root property is set', function() {
```
plroebuck:
```js
     it('does not denote the root suite by being titleless', function() {
```

boneskull:
This seems to be waiting on @plroebuck 	to approve changes

juergba:
@boneskull @plroebuck 
I'm going to merge this tomorrow, unless you have any additional feedback. Thanks.

boneskull:
fine by me. squash when you do.

plroebuck:
I still don't care for the new `isRoot` Suite ctor argument; maybe we'd have been better off to just always set `this.root = false;` within "lib/suite.js", and override it in the Mocha ctor as:
```js
  this.suite = new exports.Suite('', new exports.Context());
  this.suite.root = true;
```
But I won't hold this PR up over it.

PR #3633, opened 2018-12-21
upgrade supports-color; closes #1304

### LINKED ISSUE #1304: Color codes output to non-ttys
Like the writer of #1296 I am piping the output of Mocha tests to a non-tty (file in this case), and am getting control characters (specifically those for colors) in my text. This is nasty as I am trying to support unicode characters, and that makes life difficult.

The `reports` code appears correct, but I tracked the problem down to a combination of two places in `lib/mocha.js`:
1. in `Mocha.prototype.useColors` the default if there is no setting is to be `true`, if you have selected neither `-c/--colors` or `-C/--no-colors` this gets called called by `mocha.js` with `undefined` in the line `this.useColors(options.useColors)`, so `this.options.useColors` is `true`.
2. In `Mocha.prototype.run` there is the line:

``` python
exports.reporters.Base.useColors = options.useColors;
```

Since this does not get run until the Mocha object is instantiated, it is run later than the code in the top-level of the the `reports` modules. So this one always overrides what is set in `reporters` modules, and always sets it to `true`, unless you explicitly call `-C/--no-colors`.

A correction would be to change that last line to:

``` python
if (exports.reporters.Base.useColors == undefined) exports.reporters.Base.useColors = options.useColors;
```

boneskull:
At least, I *think* it does.  @plroebuck seems to know more about this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,6 +84,17 @@
             "ansi-styles": "^3.2.1",
             "escape-string-regexp": "^1.0.5",
             "supports-color": "^5.3.0"
+          },
+          "dependencies": {
+            "supports-color": {
+              "version": "5.5.0",
+              "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-5.5.0.tgz",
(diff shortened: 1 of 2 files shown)


outsideris:
ref: https://github.com/chalk/supports-color/pull/90

PR #3640, opened 2018-12-28
use userland which; closes #3639

### LINKED ISSUE #3639: growl should use userland "which" package
We should consume userland modules for nontrivial operations to avoid maintenance burden.  continuation of my comments from #3542, which introduced a hand-rolled `which` implementation.

boneskull:
also refactor to use modern JS in `lib/growl.js`; update ESLint rules accordingly

### DIFF
--- a/lib/growl.js
+++ b/lib/growl.js
@@ -5,9 +5,9 @@
  * @module Growl
  */
 
-var fs = require('fs');
-var os = require('os');
-var path = require('path');
+const os = require('os');
+const path = require('path');
+const {sync: which} = require('which');
 
 /**
  * @summary
@@ -23,8 +23,14 @@ var path = require('path');
  * @see {@link Mocha#isGrowlCapable}
  * @return {boolean} whether Growl notification support can be expected
  */
-exports.isCapable = function() {
-  return !process.browser && which(getSupportBinaries()) !== '';
+exports.isCapable = () => {
+  if (!process.browser) {
+    return getSupportBinaries().reduce(
+      (acc, binary) => acc || Boolean(which(binary, {nothrow: true})),
+      false
+    );
+  }
+  return false;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


boneskull:
I'm going to ignore the coverage drop here.  not my fault!!!

lib/growl.js
```diff
@@ -88,15 +95,11 @@ function display(runner) {
  * @private
  * @callback Growl~growlCB
  * @param {*} err - Error object, or <code>null</code> if successful.
- * @param {string} stdout - <code>stdout</code> from notification delivery
- *     process.
- * @param {string} stderr - <code>stderr</code> from notification delivery
- *     process. It will include timestamp and executed command with arguments.
  */
-function onCompletion(err, stdout, stderr) {
```
plroebuck: For what it's worth, I disagree with undocumenting the callback.

lib/growl.js
```diff
@@ -23,8 +23,16 @@ var path = require('path');
...
  * @return {boolean} whether Growl notification support can be expected
  */
-exports.isCapable = function() {
-  return !process.browser && which(getSupportBinaries()) !== '';
+exports.isCapable = () => {
+  if (!process.browser) {
+    const binaries = getSupportBinaries();
+    if (binaries.length) {
+      try {
+        return Boolean(which(binaries, {all: true}));
```
plroebuck: This is broken now... `binaries` is an array, but your `which` takes a string necessitating a `for` loop.
You also only want the first match, so `all` is not as helpful as `nothrow` would have been.

boneskull:
assuming this passes muster, I'll release v6.0.0-0 thereafter (maybe tomorrow?)

lib/growl.js
plroebuck: I was referring to the JSDoc (L91-L94 in original) more so than the function arguments themselves.

PR #3642, opened 2018-12-30
Add test for Runner event order

juergba:
### Description

Recently the correct order of event firing has been an issue, see [here](https://github.com/mochajs/mocha/issues/3414). 
- two `end` events with `--bail` flag (fixed)
- switched order of `start` and `suite` event

### Description of the Change

- 00ca06b needs to be backed out, it causes the switch of `start` and `suite` events. There are registered event listeners for both events.
- the order of `test`/`beforeEach` events is weird, also it's not symmetrical to `test end`/`afterEach` event order. Nevertheless the change of the `test` event firing is not part of this PR.
  - there is no event listener registered for `test` events, so as per now there are no negative implications.
  - the events `test`/`beforeEach` are called synchroniously (not caused by `setImmediate` calls).

```js
'use strict';
var Runner = require('../../../../lib/runner.js');

var realEmit = Runner.prototype.emit;
Runner.prototype.emit = function(event, ...args) {
  console.log(`emit: ${event}`);
  return realEmit.call(this, event, ...args);
};

  before('before', function() {});
  beforeEach('beforeEach', function() {});
  it('test', function() {});
  afterEach('afterEach', function() {});
  after('after', function() {});
});
```
```
emit: suite
emit: start                <<<< incorrect order
emit: suite
  suite
emit: hook
emit: hook end
emit: test                 // self.listenerCount('test') === 0
emit: hook                 <<<< why "beforeEach" hook after test?
emit: hook end
emit: pass
    √ test
emit: test end
emit: hook
emit: hook end
emit: hook
emit: hook end
emit: suite end
emit: suite end
emit: end
  1 passing (11ms)
```


boneskull:
@juergba While 00ca06b is not backed out, does this test show incorrect behavior, in your estimation?

boneskull:
btw @juergba 	please join us in the [contributors' chat room](https://gitter.im/mochajs/contributors)

test/integration/fixtures/runner/events-retries.fixture.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+var Runner = require('../../../../lib/runner.js');
+var assert = require('assert');
+
+var emitOrder = [
+  'suite'/* incorrect order*/, 'start', 'suite',
```
boneskull: This order changes when 00ca06b is removed?

boneskull:
It sounds like to me that 00ca06b shouldn't be backed out, but rather the `suite` event should also be emitted asynchronously. Adding @outsideris and @plroebuck 	as reviewers

test/integration/fixtures/runner/events-retries.fixture.js
juergba: yes, see #3414 which shows event order before and after merging 00ca06b.

juergba:
@boneskull We already have an unclear situation with emitting events asynchronously, just remember the rejected promise discussion. If the `suite` event is changed to asynch, then what about `suite end`? And `hook`, and `hook end`, ... ?

Yes, presently my tests let pass incorrect event order and have to be adapted after backing out/correcting 00ca06b.

@outsideris is already working on 00ca06b.

boneskull:
I'm going to merge this as it's just some tests.  They'll need to be changed once @outsideris addresses the problem(s).  Thanks

Issue #2312, opened 2016-06-15
Add a `--clearscreen` option to go with `--watch`


ScottFreeCode:
We should add a `--clearscreen` flag that clears the console before a test run, for use with `--watch`. This would be handy for reporters other than "min" (which has this effect built in and can be used as an example for implementation). It can currently be done with a global `before` hook, but that's clumsier than a flag would be (you have to go out of your way to set it up so that you can choose to include it or not as easily as if it were a flag) and could work incorrectly in the unlikely edge case that you have other global `before` hooks printing stuff.

Inspired by @AshCoolman on Gitter.

dNItro:
+1 for this feature

boneskull:
A neat idea.  I'm loathe to add anything to the "watch" functionality at this point though...

erikmellum:
This would be a really great flag because I want to use --reporter min with --clearscreen false (or something similar).

brennancheung:
I want this feature too.  Since I guess it has been decided that this won't be added, I just hacked it in by adding `console.log('\x1Bc');` to the `reporters` directly.

Just edit `node_modules/mocha/lib/reporters/` and add it manually where you see `runner.on('start'...`.

Obviously it's not permanent, and has to be done for each repo, but it works well enough for me.

boneskull:
@brennancheung I didn't decide it wasn't to be added :wink:  I'd just rather avoid piling features ontop of an already problematic feature.

HoltMansfield:
+1

justinmchase:
Or another option would be, if adding a flag is too complicated, to make a new reporter, call it `even-more-min`, which does the exact same thing as min except for clearing the screen and setting the cursor.

brentgreeff:
I am running `yarn mocha --recursive --watch --reporter min`

This did not clear the screen when I saved a file
I upgraded mocha from 3.2.0 to mocha@6.1.4

this had no effect. Screen output is not cleared at all.

Coming from Rails where I run Guard all the time with clearing :on - I don't know if I can live without it.

brentgreeff:
> This did not clear the screen when I saved a file

Actually it seems to add a full page of black space which is not the `clear` command

justinmchase:
Maybe yarn blocks the clearing or translates it? I don't think this feature was ever implemented in mocha so the min reporter does still clear last I looked.

Try:
```
npx mocha --recursive --watch --reporter min
```

brentgreeff:
[...]

PR #3645, opened 2019-01-02
 Remove clear and cursor move in min reporter 

### LINKED ISSUE #2761: How to prevent console clears when using --reporter min (-R min)
Right now I get lots of clears in my console when using the --reporter min flag. This is not desired because I am running tests as part of a gulpfile that does watching / transpiling / linting etc. The console clears really obfuscate some of the other important info that I get.

Is there any way to stop the console clears? I perused the mochajs homepage and couldn't find a way.

Thanks!

justinmchase:
### Description of the Change

Affecting the entire view of the console is not minimal and it prevents this from being used in conjunction with other watch scripts running in parallel in the same process, such as linters and nodemon, docker, etc.

Its not appropriate to do this in a reporter and a separate `--clearscreen` feature should be added if anyone wants that feature specifically.

### Alternate Designs

- Adding a new reporter called `even-more-min` which was identical to `min` but just didn't clear screen or move the cursor.
- Adding a `--clearscreen false` option which has been [rejected](https://github.com/mochajs/mocha/issues/2312#issuecomment-318923995) by mocha devs as being too complex in an already complicated feature.

It is a reduction in features. It removes functionality already in core that negatively impacts some scenarios, especially running in docker or in conjunction with other watch scripts.

### Benefits

It will allow the min reporter to be able to be used in the same process as other "watch" scripts. This is really important for being able to be run in docker, for example. When running in docker many processes are emitting output and being displayed into the console of a single shell, if any of those processes does a "clear screen" or a move of a cursor it can wipeout the output of other processes. It's not polite to do this and should be an opt-in only feature.

### Possible Drawbacks

The draw back is that upon success during `--watch -R min` the minimal reporter will append the epilogue rather than controlling the entire console window. Over time you will accumulate messages such as `300 passing (400ms)`.

This is necessary in a multi-child process terminal but may not be as pretty in 5 minute demos.

### Applicable issues

Relates to #2312
Fixes #2761

This is technically a breaking change but I don't think it actually warrants a major version bump. I would recommend a minor version bump.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1405,7 +1405,7 @@ The "JSON stream" reporter outputs newline-delimited JSON "events" as they occur
 
 ### Min
 
-The "min" reporter displays the summary only, while still outputting errors on failure. This reporter works great with `--watch` as it clears the terminal in order to keep your test summary at the top.
+The "min" reporter displays the summary only, while still outputting errors on failure. This reporter works great with `--watch` as it only outputs the test summary.
 
 ![min reporter](images/reporter-min.png?withoutEnlargement&resize=920,9999){:class="screenshot"}
 
--- a/lib/reporters/min.js
+++ b/lib/reporters/min.js
@@ -26,14 +26,6 @@ exports = module.exports = Min;
  */
 function Min(runner) {
   Base.call(this, runner);
-
-  runner.on('start', function() {
-    // clear screen
-    process.stdout.write('\u001b[2J');
-    // set cursor position
-    process.stdout.write('\u001b[1;3H');
-  });
-
   runner.once('end', this.epilogue.bind(this));
 }
 
--- a/test/reporters/min.spec.js
+++ b/test/reporters/min.spec.js
@@ -16,11 +16,11 @@ describe('Min reporter', function() {
   });
 
   describe('on start', function() {
-    it('should clear screen then set cursor position', function() {
+    it('should not clear screen or set cursor position', function() {
       runner = createMockRunner('start', 'start');
       var stdout = runReporter({epilogue: function() {}}, runner, options);
 
-      var expectedArray = ['\u001b[2J', '\u001b[1;3H'];
+      var expectedArray = [];
       expect(stdout, 'to equal', expectedArray);
     });
   });


markowsiak:
Generally, I also like the idea of the screen not being cleared by the min reporter.

@boneskull is there a reason why we might want to consider leaving it in, other than the scenario presented in "Possible Drawbacks"?

Otherwise LGTM 👍

justinmchase:
I believe the reason why is just so you can get that experience of doing:
```sh
mocha -r min
```

And then you get that "minimalist" output of just seeing your failing tests as you develop. Look here at the documentation, I think it explains the reasoning:

https://mochajs.org/#min
> The “min” reporter displays the summary only, while still outputting errors on failure. This reporter works great with --watch as it clears the terminal in order to keep your test summary at the top.

And now that I mention it, I will update the documentation to remove the comment about clearing the terminal also. But I believe that was the entire reasoning.

plroebuck:
Is color support needed?

justinmchase:
@plroebuck I'm sorry, I don't understand the question. What do you mean? This should not affect colors at all.

boneskull:
I have a feeling people are going to complain about this one.  It seems like a matter of opinion.

Unless anyone at @mocha/core feels strongly about this (and doesn't mind explaining the reasoning to users), I'd suggest we pass on it.  A custom reporter with these changes can be used instead; in fact, it'd be pretty simple to implement (untested):

```js
const {Base, Min} = require('mocha').reporters;
const {inherits} = require('util');

// avoid calling Min constructor; I don't think ES6 classes will let you do this
function MyMin(runner) {
  Base.call(this, runner);
  runner.once('end', () => {
    this.epilogue();
  });
}

inherits(MyMin, Min);

module.exports = MyMin;
```

justinmchase:
I made that exact reporter and published it as [`even-more-min`](https://www.npmjs.com/package/even-more-min) in case anyone wants it.

_I do feel strongly_ that it should be the default reporter however, especially since the devs previously declined to add the option to prevent it from happening, but I'm just me and I am not from @mochajs/core.

plroebuck:
I _was_ working on this, but solving it from a different angle that allowed both behaviors without new cmdline arguments. @justinmchase, if we do _your_ change, then you're happy but someone else will complain about the lost "feature".

plroebuck:
@justinmchase
My (sorry, still unimplemented) idea was to create a "terminal.js" file for use with "lib/reporters/base.js". It would provide an `isDumb()` method and the existing cursor handling implementation from `Base` would be migrated there as well. As dumb terminals (by definition) do not provide cursor handling, this would seem to cover both your `--clearscreen` behavior without any change to existing users. When fleshed out, reporters would no longer be writing raw escape codes themselves, but invoking routines within `terminal` to do so.

####  "lib/terminal.js"
```js
/**
 * @module Terminal
 */
module.exports = {
  isDumb: function() {
    return process.env.TERM === 'dumb';
  }
  // cursor handling here...
};
```
####  "lib/reporters/min.js"
```js
var terminal = require('../terminal');

function Min(runner) {
  Base.call(this, runner);

  if (!terminal.isDumb()) {
    runner.on('start', function() {
      // clear screen
      process.stdout.write('\u001b[2J');
      // set cursor position
      process.stdout.write('\u001b[1;3H');
    });
  }

  runner.once('end', this.epilogue.bind(this));
}
```

Usage:
```sh
$ TERM=dumb mocha --watch -R min
```

boneskull:
@justinmchase Yes, you'd have to pass the `--colors` flag to Mocha to force the colors, or `FORCE_COLOR=1` etc.  But colors seem kind of _extra_, don't they?  :wink:

boneskull:
Thanks for the PR anyway.  Please understand that changing this behavior will likely draw the ire of many who are happy with the reporter the way it works now.  Your custom reporter looks like a fine solution.

PR #3653, opened 2019-01-04
fix test running on windows

boneskull:
- fix prettier settings
- add `cross-env`

cc @juergba

### DIFF
--- a/package.json
+++ b/package.json
@@ -517,6 +517,7 @@
     "chai": "^4.2.0",
     "coffee-script": "^1.10.0",
     "coveralls": "^3.0.2",
+    "cross-env": "^5.2.0",
     "cross-spawn": "^6.0.5",
     "eslint": "^5.10.0",
     "eslint-config-prettier": "^3.3.0",
@@ -582,7 +583,8 @@
   },
   "prettier": {
     "singleQuote": true,
-    "bracketSpacing": false
+    "bracketSpacing": false,
+    "endOfLine": "auto"
   },
   "yargs": {
     "combine-arrays": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -3038,6 +3038,16 @@
       "integrity": "sha512-w9UZUtkaGd8MfS7eMG7Sa0lV5vCJghqQfiOnwNVrPhbZScUp5h0jwYoAF933MKlotlG1JAJOCCT3xU6r+SDKNw==",
       "dev": true
     },
+    "cross-env": {
+      "version": "5.2.0",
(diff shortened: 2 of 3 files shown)


boneskull:
cc @juergba 	this work for you?

package.json
```diff
@@ -582,7 +583,8 @@
   },
   "prettier": {
     "singleQuote": true,
-    "bracketSpacing": false
+    "bracketSpacing": false,
+    "endOfLine": "auto"
```
plroebuck: lexical order?

package.json
```diff
@@ -517,17 +517,18 @@
     "chai": "^4.2.0",
     "coffee-script": "^1.10.0",
     "coveralls": "^3.0.2",
+    "cross-env": "^5.2.0",
```
plroebuck: I understand the addition of this devDependency. Are the others being updated here related to _this_ PR?

juergba:
@boneskull this one merged would make my life bright and beautiful...

package.json
```diff
@@ -517,17 +517,18 @@
...
     "cross-spawn": "^6.0.5",
-    "eslint": "^5.10.0",
+    "eslint": "^5.11.1",
     "eslint-config-prettier": "^3.3.0",
     "eslint-config-semistandard": "^13.0.0",
     "eslint-config-standard": "^12.0.0",
```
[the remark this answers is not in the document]
boneskull: peer deps of semistandard

package.json
boneskull: unless we can lint and automatically format it, I'm not interested in enforcing lexical order anywhere.  it's nice, but also IMO a nit.

package.json
boneskull: I don't think so; didn't intend this.  I'll pull them out.  note to self: don't run `npm upgrade` unless you mean it.

package.json
plroebuck: > unless we can lint and automatically format it, I'm not interested in enforcing lexical order anywhere. it's nice, but also IMO a nit.

Wholeheartedly disagree. We shouldn't _require_ lint and autoformat just to put items in alphabetical order, and not doing so just makes maintenance harder (requiring search rather than skim). There are only two properties in this file ('prettier' and 'yargs') that really _aren't_, and both can be trivially corrected.

package.json
plroebuck: Removed from here, but "package-lock.json" not regenerated...

boneskull: it’s the same because the semver ranges install the same packages at this point.

boneskull: at least that’s my theory. I’ll double check that it doesn’t get modified if I install again

boneskull: yep, no change.

PR #3655, opened 2019-01-04
refactor: use constants for event names instead of string literals

boneskull:
- also a few other string literals got replaced, but not error messages nor codes
- a couple adjacent refactors:
  - move the suite "GC" function into the `Suite` prototype
  - move stats collector init to `Mocha#run` due to circular reference with `Runner`
  - rename a couple fixture files lacking proper extension
  - rename variable in integration test helpers
- specifically did NOT refactor event names from Node.js incl. `error` and `uncaughtException`

Did *not* rename any events.


lib/suite.js
```diff
@@ -422,6 +422,79 @@ Suite.prototype.eachTest = function(fn) {
...
+  SUITE_HOOK_BEFORE_EACH: 'beforeEach',
+  SUITE_HOOK_AFTER_ALL: 'afterAll',
+  SUITE_HOOK_AFTER_EACH: 'afterEach',
+
+  // XXX these are all unused and need to be removed.
+  // they've existed since time immemorial, so probably safe to consider
+  // removal to be a breaking change.
+  SUITE_EVENT_TEST: 'test',
+  SUITE_EVENT_SUITE: 'suite',
+  SUITE_EVENT_BEFORE_ALL: 'beforeAll',
```
craigtaub: Out of alphabetical order

lib/reporters/nyan.js
```diff
@@ -40,24 +41,24 @@ function NyanCat(runner) {
   this.trajectories = [[], [], [], []];
   this.trajectoryWidthMax = width - nyanCatWidth;
 
-  runner.on('start', function() {
+  runner.on(Runner.constants.RUNNER_EVENT_START, function() {
```
craigtaub: Quite noisy in this state. Can we re-assign them all at top of file?

boneskull: do you mean e.g. `var RUNNER_EVENT_START = Runner.constants.RUNNER_EVENT_START`?

ugh, need es6 everywhere... somebody give me babel and webpack

craigtaub: Yep thats what i mean...es6 everywhere 😍

boneskull:
Instead of the wiki, I've added a tutorial in the API documentation section.

Once this is merged & released we can redirect users to it.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,113 @@
+Mocha allows you to define and use custom reporters install via `npm`.
+
+For example, if `mocha-foo-reporter` was published to the npm registry, you could install it via `npm install mocha-foo-reporter -D`, then use it via `mocha --reporter mocha-foo-reporter`.
```
outsideris: How about we use `--save-dev` rather than `-D`. All npm users don't familiar with `-D` flag and it is hard to search even if I also use `-D`.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+    this._indents--;
+  }
+}
+
+module.exports = MyReporter;
+```
+
+To use this reporter, execute `mocha --reporter /path/to/my-reporter.js`.
+
+For further examples, the built-in reporter implementations are the [best place to look](<(https://github.com/mochajs/mocha/tree/master/lib/reporters)>). The [integration tests](https://github.com/mochajs/mocha/tree/master/test/reporters) may also be helpful.
```
outsideris: `<( )>` is needed?

docs/api-tutorials/custom-reporter.md
plroebuck: Agreed. I've never used the single letter flag.

docs/api-tutorials/custom-reporter.md
plroebuck: Agreed. Just use normal MD link.

lib/runner.js
```diff
@@ -1027,11 +1004,34 @@ function extraGlobals() {
...
+  RUNNER_EVENT_SUITE: 'suite',
+  RUNNER_EVENT_SUITE_END: 'suite end',
+  RUNNER_EVENT_TEST: 'test',
+  RUNNER_EVENT_TEST_END: 'test end',
+  RUNNER_EVENT_WAITING: 'waiting',
+  RUNNER_STATE_FAILED: 'failed',
+  RUNNER_STATE_PASSED: 'passed'
+});
+
+module.exports.constants = constants;
```
plroebuck: Consider instead:

**"lib/events.js"**
```js
module.exports = {
  Runnable: {
    EVENT_RUNNABLE_ERROR: 'error'
  },
  Suite: {
    EVENT_AFTER_ALL_HOOK_QUEUED: 'afterAll',
    EVENT_AFTER_EACH_HOOK_QUEUED: 'afterEach',
    EVENT_BEFORE_ALL_HOOK_QUEUED: 'beforeAll',
    EVENT_BEFORE_EACH_HOOK_QUEUED: 'beforeEach',

    EVENT_POST_REQUIRE: 'post-require',
    EVENT_PRE_REQUIRE: 'pre-require',
    EVENT_REQUIRE: 'require',

    EVENT_ROOT_SUITE_RUN: 'run',

    EVENT_SUITE_QUEUED: 'suite',
    EVENT_TEST_QUEUED: 'test'
  },
  Runner: {
    EVENT_HOOK_BEGIN: 'hook',
    EVENT_HOOK_END: 'hook end',

    EVENT_ROOT_SUITE_BEGIN: 'start',
    EVENT_ROOT_SUITE_END: 'end',
    EVENT_ROOT_SUITE_WAITING: 'waiting',

    EVENT_SUITE_BEGIN: 'suite',
    EVENT_SUITE_END: 'suite end',

    EVENT_TEST_BEGIN: 'test',
    EVENT_TEST_END: 'test end',

    EVENT_TEST_PASS: 'pass',
    EVENT_TEST_FAIL: 'fail',
    EVENT_TEST_PENDING: 'pending',
    EVENT_TEST_RETRY: 'retry'
  }
};
```

Then in usage:
```js
var events = require('lib/events').Runner;
...
  runner.on(events.EVENT_TEST_BEGIN, ...);
```

The events for each should be JSDoc'ed (not shown above).
```js
/**
 * Test fail event.
 * Indicates test failed.
 *
 * @event module:Runner.event:fail
 * @property {Test} test - Instance that was executed
 * @property {Error} err - Cause of failure
 */
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+| `RUNNER_EVENT_HOOK`      | `hook`      | `Hook`          | A [Hook] is about to execute                                                                |
+| `RUNNER_EVENT_HOOK_END`  | `hook end`  | `Hook`          | A [Hook] has completed execution                                                            |
+| `RUNNER_EVENT_PASS`      | `pass`      | `Test`          | A [Test] has passed                                                                         |
+| `RUNNER_EVENT_PENDING`   | `pending`   | `Test`          | A [Test] was skipped                                                                        |
+| `RUNNER_EVENT_RETRY`     | `retry`     | `Test`, `Error` | A [Test] failed, but is about to be retried; never emitted unless `retry` option is nonzero |
+| `RUNNER_EVENT_BEGIN`     | `start`     | _(n/a)_         | Execution will begin                                                                        |
+| `RUNNER_EVENT_SUITE`     | `suite`     | `Suite`         | The [Hook]s and [Test]s within a [Suite] are about to be executed                           |
```
plroebuck: Think the event names should be bookended when appropriate.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+| `RUNNER_EVENT_TEST_END`  | `test end`  | `Test`          | A [Test] has completed execution                                                            |
+| `RUNNER_EVENT_WAITING`   | `waiting`   | _(n/a)_         | Waiting for `global.run()` to be called; only emitted when `delay` option is `true`         |
+
+**Please use these constants** instead of the event names in your own reporter! This will ensure compatibility with future versions of Mocha.
+
+> It's important to understand that all suite callbacks will be run _before_ the [Runner] emits `RUNNER_EVENT_BEGIN`. Hooks and tests, however, won't run until _after_ the [Runner] emits `RUNNER_EVENT_BEGIN`.
+
+## Custom Reporter for Browser
+
+As of Mocha v6.0.0, custom reporters are _only_ "officially" supported in Node.js. [Mochify](https://npm.im/mochify) might help you. YMMV!
```
plroebuck: Say what?

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+
+[runner]: /api/mocha.runner
+[test]: /api/mocha.test
+[hook]: /api/mocha.hook
+[suite]: /api/mocha.suite
```
plroebuck: Why was the [Wiki page](https://github.com/mochajs/mocha/wiki/Third-party-reporters) poo-poo'd in favor of this?

jsdoc.conf.json
```diff
@@ -1,32 +1,33 @@
...
   },
-  "plugins": ["plugins/markdown"],
   "opts": {
-    "encoding": "utf8",
-    "template": "node_modules/@mocha/docdash",
     "destination": "docs/api",
+    "encoding": "utf8",
     "recurse": true,
+    "template": "node_modules/@mocha/docdash",
+    "tutorials": "./docs/api-tutorials",
```
plroebuck: outside this single line, I can't say I understand why any other changes were made here...

lib/mocha.js
```diff
@@ -759,6 +761,7 @@ Mocha.prototype.run = function(fn) {
   var options = this.options;
   options.files = this.files;
   var runner = new exports.Runner(suite, options.delay);
+  createStatsCollector(runner);
```
plroebuck: No issue with moving it here, but could you explain the problem it caused where it was.

test/integration/fixtures/simple-reporter.fixture.js
```diff
@@ -0,0 +1,35 @@
...
+var RUNNER_EVENT_END = constants.RUNNER_EVENT_END;
+var RUNNER_EVENT_TEST_END = constants.RUNNER_EVENT_TEST_END;
+
+module.exports = SimpleReporter;
+
+function SimpleReporter(runner) {
+  Base.call(this, runner);
+
+  runner.on(RUNNER_EVENT_SUITE, function(suite) {
+    console.log("on('suite') called");
```
plroebuck: Why aren't the event names being used in console messages?

test/integration/regression.spec.js
```diff
@@ -31,7 +31,7 @@ describe('regressions', function() {
       'fixtures',
       'regression',
       '1794',
-      'simple-ui.js'
+      'simple-ui.fixture.js'
```
plroebuck: This unrelated change should have been done in separate PR...

test/integration/reporters.spec.js
```diff
@@ -66,7 +66,7 @@ describe('reporters', function() {
   describe('loader', function() {
     it('loads a reporter from a path relative to the current working directory', function(done) {
       var reporterAtARelativePath =
-        'test/integration/fixtures/simple-reporter.js';
+        'test/integration/fixtures/simple-reporter.fixture.js';
```
plroebuck: See previous comment

test/jsapi/index.js
```diff
@@ -21,10 +21,6 @@ mocha.addFile('test/unit/duration.spec.js');
 mocha.addFile('test/unit/globals.spec.js');
 mocha.addFile('test/unit/timeout.spec.js');
 
-mocha
-  .run(function() {
-    console.log('done');
-  })
-  .on('pass', function(test) {
```
plroebuck: Curious. Why event handler removed?

test/unit/throw.spec.js
```diff
@@ -111,7 +112,7 @@ describe('a test that throws', function() {
       });
       suite.addTest(test);
       runner = new Runner(suite);
-      runner.on('end', function() {
+      runner.on(Runner.constants.RUNNER_EVENT_END, function() {
```
plroebuck: Didn't use short var.

lib/runner.js
boneskull: I'm following Node.js' lead on how they export constants.  Each module contains a `constants` object which contains constants relevant to that module.

docs/api-tutorials/custom-reporter.md
boneskull: I am not sure what you're referring to.

docs/api-tutorials/custom-reporter.md
boneskull: To what are you referring?

docs/api-tutorials/custom-reporter.md
boneskull: I would like to consolidate tutorials on the documentation site.  It is painful to have to go to separate sites for information.

I am not thrilled about how the tutorial is buried in the navigation, but #3663 should allow us to move the "Tutorials" nav section to the top.

jsdoc.conf.json
boneskull: I think I sorted the object keys.

lib/mocha.js
boneskull: Circular dependency.  Runner would require stats-collector which would in turn require runner.  Can create problems with various tools (browserify, test runners, etc)

test/integration/fixtures/simple-reporter.fixture.js
boneskull: You mean the constants?  What the reporter actually does here is irrelevant to the existing integration tests, which just check that the reporter can load without error.

test/integration/regression.spec.js
boneskull: I felt like that was too much overhead for a rename, given that I'm in the file already.

test/jsapi/index.js
boneskull: Irrelevant; see [comment](https://github.com/mochajs/mocha/pull/3655#discussion_r247232710)

test/unit/throw.spec.js
boneskull: - [ ] use short var

lib/runner.js
plroebuck: [Such as](https://github.com/search?q=constants++repo%3Anodejs%2Fnode+path%3A%2Flib+language%3AJavaScript&type=Code&ref=advsearch&l=JavaScript&l=)?

docs/api-tutorials/custom-reporter.md
plroebuck: For example, `RUNNER_EVENT_SUITE` and `RUNNER_EVENT_SUITE_END` are unmatched.
Better named `RUNNER_EVENT_SUITE_BEGIN`.

docs/api-tutorials/custom-reporter.md
plroebuck: When did that happen?

lib/runner.js
boneskull: `fs.constants`, `os.constants`, `zlib.constants`, `http2.constants`, etc.

docs/api-tutorials/custom-reporter.md
boneskull: - [ ] rename.  also `RUNNER_EVENT_TEST` should be `RUNNER_EVENT_TEST_BEGIN` and `RUNNER_EVENT_HOOK` should be `RUNNER_EVENT_HOOK_BEGIN`.  reference corresponding event

docs/api-tutorials/custom-reporter.md
boneskull: they've never been supported, as far as I'm aware.  might as well make that clear... I suppose I can remove the "as of.."

test/integration/fixtures/simple-reporter.fixture.js
boneskull: I'm going to replace this with the "example" reporter.

lib/runner.js
boneskull: - [ ] add docstrings for all events

lib/reporters/list.js
```diff
@@ -8,6 +8,12 @@
 
 var Base = require('./base');
 var inherits = require('../utils').inherits;
+var constants = require('../runner').constants;
+var RUNNER_EVENT_PASS = constants.RUNNER_EVENT_PASS;
+var RUNNER_EVENT_FAIL = constants.RUNNER_EVENT_FAIL;
+var RUNNER_EVENT_BEGIN = constants.RUNNER_EVENT_BEGIN;
```
plroebuck: Still dislike many of these constant names. For example, think the four root suite-related events should express that in their constant names (e.g., `RUNNER_EVENT_BEGIN` vs. `EVENT_ROOT_SUITE_BEGIN`. Same for test events (e.g., `RUNNER_EVENT_PASS` vs. `EVENT_TEST_PASS`). Both the `RUNNER_` and `SUITE_` prefixes don't add anything (given the `require`) [IMO], making the names unnecessarily longer without providing additional context.

lib/reporters/list.js
```diff
@@ -8,6 +8,12 @@
 
 var Base = require('./base');
 var inherits = require('../utils').inherits;
+var constants = require('../runner').constants;
```
plroebuck: Since these are being for the event names...
```js
var events = require('../runner').constants;
```

lib/reporters/min.js
```diff
@@ -8,6 +8,9 @@
 
 var Base = require('./base');
```
plroebuck: This is a question, not request for change:
As _all_ reporters besides `html` are Node-based, are they required to retain ES5-compatiblity?

lib/runner.js
```diff
@@ -227,7 +227,7 @@ Runner.prototype.fail = function(test, err) {
   }
 
   ++this.failures;
-  test.state = 'failed';
+  test.state = constants.RUNNER_STATE_FAILED;
```
plroebuck: As these two are unused outside "lib/runner.js" itself, why not just let them be module variables?

lib/suite.js
```diff
@@ -420,6 +420,79 @@ Suite.prototype.eachTest = function(fn) {
...
+    cleanArrReferences(this._afterAll);
+  }
+
+  if (Array.isArray(this._afterEach)) {
+    cleanArrReferences(this._afterEach);
+  }
+
+  for (var i = 0; i < this.tests.length; i++) {
+    delete this.tests[i].fn;
+  }
```
plroebuck: Realize you just migrated the existing code. Question unrelated to this PR.
* Should we be using `delete` here instead of `this.tests[i].fn = null;`?

While `delete` is better semantically, (re)setting the value is [far more performant](https://jsperf.com/delete-vs-undefined-vs-null/16) assuming no aliasing. As this is being done _only_ for GC purposes, I'm glossing over the _actual_ difference between the two and only looking at the performance.

See also:
* [Does using delete keyword effect v8 optimizations of an object?]
(https://groups.google.com/forum/#!topic/v8-users/zE4cOHBkAnY)
* [Trash talk: the Orinoco GC](https://v8.dev/blog/trash-talk)
* [How To Write Fast, Memory-Efficient JavaScript](https://www.smashingmagazine.com/2012/11/writing-fast-memory-efficient-javascript/)

test/integration/fixtures/simple-reporter.fixture.js
plroebuck:
```js
  runner.on(RUNNER_EVENT_SUITE, function(suite) {
    console.log("on('" + RUNNER_EVENT_SUITE + "') called");
  });
```

test/integration/regression.spec.js
plroebuck: It just makes it that much harder to use GitHub history when all the changes keep happening in unrelated PRs. For simple file rename, assuming checks still work afterwards, I wouldn't care if you had just merged it first.

Anyhow...

lib/runner.js
plroebuck: Given your examples, create `Runner.constants.events` mirroring how they handle `Error.codes`.

docs/api-tutorials/custom-reporter.md
boneskull: What I was trying to say was this:

- Mocha supports 3rd-party reporters insofar as it provides an API for those reporters to use, and an option (`--reporter`) to run tests using one.
- The `Base` reporter--which Node.js reporters should extend--is not intended for use in a browser, so it would be difficult to make one portable.

But looking more closely at the code: while the HTML reporter is used by default in the browser, I'm not seeing anything that would prevent a user from using a custom reporter in a browser.  So I'll probably just remove this note and add another about `Base` & the browser.

- [ ] rewrite note about browser testing

lib/reporters/list.js
boneskull: I want to ensure the constant names are all unique.  This will help code completion, and if you need to use two similarly-named constants from two different modules.  For example, `SUITE_BEGIN` could be a problem, as currently written.

I'll go back through them and ensure all the constant names are unique and see if it makes sense to remove the prefixes.

- [ ] audit names; attempt to remove prefixes

lib/reporters/min.js
boneskull: I think, simply because they're all pulled in to the bundle.  *Not* pulling them in to would break people using certain headless browser setups.

Ideally we shouldn't have to manage the language level on a file-by-file or directory-by-directory basis, and would have a transpilation step.  I don't want to have to think about it, and neither should our contributors.

lib/reporters/list.js
boneskull: if/when this becomes ES6 code, it can be easily search/replace using a regex as `const {constants} = require('../runner')`.

lib/reporters/min.js
boneskull: (yes, to pile on w/ the confusion, the node.js-based reporters *can and do* run in a browser on the command-line)

lib/runner.js
boneskull: custom reporters which use the `failed` event.

lib/suite.js
boneskull: I don't know.. I didn't write this stuff.  if you want to profile it and have a better solution, please send a PR

lib/reporters/min.js
plroebuck: Wish I could say I understand what you just wrote, but I don't. Many are terminal-based (e.g., Nyan) and _should_ fail if attempted to run in browser. I was about to add more error checking which would cause `Error` to be thrown if terminal cursor positioning was unsupported in a new PR. You saying that won't work due to browser?

lib/runner.js
plroebuck: Move to `Runner.constants.states`.

lib/reporters/list.js
boneskull: Also, I want to ensure each module has its own constants.  If the event names are the same strings in two different modules, we shouldn't conflate these into a single constant.

test/integration/fixtures/regression/1794/simple-ui.fixture.js
```diff
@@ -1,15 +1,22 @@
 'use strict';
 
-var path = '../../../../../lib/';
-var Mocha = require(path + 'mocha');
-var Test = require(path + 'test');
+var Mocha = require('../../../../../lib/mocha');
```
plroebuck: [module-alias](https://www.npmjs.com/package/module-alias)?

lib/reporters/list.js
boneskull: I've done the best I can with the names.

lib/reporters/min.js
boneskull: See what [mochify](https://npm.im/mochify) does.  Essentially it takes Mocha and pipes it thru a headless browser, but it looks like it's running as it would with Node.js.  That means you can use console-based reporters with it.

It's unclear to me if what you're doing would cause a problem.  My suggestion is to download Mochify and `npm link` it against your `mocha` working copy & see what happens.

lib/reporters/list.js
boneskull: Eh, I have a couple more to change...

lib/runner.js
boneskull: Most custom reporters probably wouldn't use this, but some may want to listen for e.g., `EVENT_TEST_END` and check `test.state === TEST_STATE_FAILED`, like `xunit`.

test/integration/fixtures/regression/1794/simple-ui.fixture.js
boneskull: we can probably tackle that when we start transpiling stuff, but yes.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,119 @@
...
+
+## The `Base` Reporter Class
+
+[Base] is an "abstract" class.  It contains console-specific settings and utilities, commonly used by built-in reporters.  Browsing the built-in reporter implementations, you may often see static properties of [Base] referenced.
+
+It's often useful--but not necessary--for a custom reporter to extend [Base].
+
+## Statistics
+
+Mocha adds a `stats` property of type [StatsCollector](/api/module-lib_stats-collector.html) to each reporter instance immediately following instantiaton.
```
plroebuck: ...to each reporter instance derived from `Base`...

instantiation - sp

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Events
+
+A reporter should listen for events emitted from the `runner` (a singleton instance of [Runner]).
+
+The event names are exported from the `constants` property of `Mocha.Runner`:
+
+| Constant             | Event Name  | Event Arguments | Description                                                                                                                                                   |
+| -------------------- | ----------- | --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `EVENT_RUN_BEGIN`    | `start`     | _(n/a)_         | Execution will begin. Corresponding "end" event is `EVENT_RUN_END`.                                                                                           |
```
plroebuck: Don''t think the 'Corresponding "xxx" event is `EVENT_XXX`' are necessary now that the event names are paired via their constant names.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,119 @@
...
+| `EVENT_RUN_END`      | `end`       | _(n/a)_         | All [Suite]s, [Test]s and [Hook]s have completed execution. Corresponding "begin" event is `EVENT_RUN_BEGIN`.                                                 |
+| `EVENT_RUN_DELAY`    | `waiting`   | _(n/a)_         | Waiting for `global.run()` to be called; only emitted when `delay` option is `true`.                                                                          |
```
plroebuck: While I understand what you're saying here, a link to [Delayed Root Suite](https://mochajs.org/#delayed-root-suite) might be appropriate...

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,119 @@
...
+| `EVENT_SUITE_BEGIN`  | `suite`     | `Suite`         | The [Hook]s and [Test]s within a [Suite] are about to be executed. Any nested [Suite]s will also be executed. Corresponding "end" event is `EVENT_SUITE_END`. |
+| `EVENT_SUITE_END`    | `suite end` | `Suite`         | The [Hook]s, [Test]s, and nested [Suite]s within a [Suite] have completed execution. Corresponding "begin" event is `EVENT_SUITE_BEGIN`.                      |
+| `EVENT_HOOK_BEGIN`   | `hook`      | `Hook`          | A [Hook] will execute. Corresponding "end" event is `EVENT_HOOK_END`.                                                                                         |
```
plroebuck: A [Hook] will be executed.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,119 @@
...
+| -------------------- | ----------- | --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `EVENT_RUN_BEGIN`    | `start`     | _(n/a)_         | Execution will begin. Corresponding "end" event is `EVENT_RUN_END`.                                                                                           |
+| `EVENT_RUN_END`      | `end`       | _(n/a)_         | All [Suite]s, [Test]s and [Hook]s have completed execution. Corresponding "begin" event is `EVENT_RUN_BEGIN`.                                                 |
+| `EVENT_RUN_DELAY`    | `waiting`   | _(n/a)_         | Waiting for `global.run()` to be called; only emitted when `delay` option is `true`.                                                                          |
+| `EVENT_SUITE_BEGIN`  | `suite`     | `Suite`         | The [Hook]s and [Test]s within a [Suite] are about to be executed. Any nested [Suite]s will also be executed. Corresponding "end" event is `EVENT_SUITE_END`. |
```
plroebuck: The [Hook]s and [Test]s within a [Suite] will be executed.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,119 @@
...
+| `EVENT_SUITE_END`    | `suite end` | `Suite`         | The [Hook]s, [Test]s, and nested [Suite]s within a [Suite] have completed execution. Corresponding "begin" event is `EVENT_SUITE_BEGIN`.                      |
+| `EVENT_HOOK_BEGIN`   | `hook`      | `Hook`          | A [Hook] will execute. Corresponding "end" event is `EVENT_HOOK_END`.                                                                                         |
+| `EVENT_HOOK_END`     | `hook end`  | `Hook`          | A [Hook] has completed execution. Corresponding "begin" event is `EVENT_HOOK_BEGIN`.                                                                          |
+| `EVENT_TEST_BEGIN`   | `test`      | `Test`          | A [Test] will execute. Corresponding "end" event is `EVENT_TEST_END`.                                                                                         |
```
plroebuck: A [Test] will be executed.

lib/suite.js
```diff
@@ -24,10 +24,10 @@ exports = module.exports = Suite;
 /**
  * Create a new `Suite` with the given `title` and parent `Suite`.
  *
- * @memberof Mocha
  * @public
  * @param {Suite} parent
  * @param {string} title
+ * @constructs Suite
```
plroebuck: Move to L27.5 (i.e., after `@public`)

lib/test.js
```diff
@@ -28,6 +29,15 @@ function Test(title, fn) {
...
   this.type = 'test';
+
+  /**
+   * @summary State of Test after execution
+   * @desc One of `Test.constants.STATE_FAILED` or `Test.constants.STATE_PASSED`.
+   * @see Test.constants
+   * @type {string}
+   * @public
+   */
+  this.state = void 0;
```
plroebuck: Why not `undefined` or `null`, especially as its supposedly a string?

lib/test.js
```diff
@@ -48,3 +58,23 @@ Test.prototype.clone = function() {
...
+var constants = {
+  /**
+   * Value of {@link Test#state} when a `Test` has failed
+   */
+  STATE_FAILED: 'failed',
+  /**
+   * Value of {@link Test#state} when a `Test` has passed
+   */
+  STATE_PASSED: 'passed'
+};
```
plroebuck: Are there other states?
* `STATE_INIT` (maybe `null` or `undefined`)
* `STATE_PENDING`?

test/integration/fixtures/regression/1794/simple-ui.fixture.js
```diff
@@ -1,15 +1,22 @@
...
+var Test = Mocha.Test;
+var Suite = Mocha.Suite;
```
plroebuck:
```js
var EVENT_FILE_PRE_REQUIRE = Suite.constants.EVENT_FILE_PRE_REQUIRE;
```

test/integration/helpers.js
```diff
@@ -90,7 +90,8 @@ module.exports = {
           fn(
             new Error(
               format(
-                'Failed to parse JSON reporter output from result:\n\n%O',
+                'Failed to parse JSON reporter output.\nArgs: %O\nResult:\n\n%O',
+                args,
```
plroebuck: Discards the `SyntaxError.message` which tells you what was wrong with the JSON, n'est-ce pas?

test/integration/helpers.js
```diff
@@ -212,7 +213,8 @@ function _spawnMochaWithListeners(args, fn, opts) {
   mocha.on('close', function(code) {
     fn(null, {
       output: output.split('\n').join('\n'),
```
plroebuck: Isn't this a noop?

test/unit/runner.spec.js
```diff
@@ -191,7 +195,7 @@ describe('Runner', function() {
 
       global.foo = 'bar';
       global.bar = 'baz';
-      runner.on('fail', function(test, err) {
+      runner.on(EVENT_TEST_FAIL, function(test, err) {
         expect(test.title, 'to be', 'im a test about lions');
         expect(err.message, 'to be', 'global leak detected: bar');
         delete global.foo;
```
plroebuck: Why is not deleting `global.bar` _not_ causing issues for tests that follow?

lib/mocha.js
```diff
@@ -14,6 +14,8 @@ var utils = require('./utils');
 var mocharc = require('./mocharc.json');
 var assign = require('object.assign').getPolyfill();
 var errors = require('./errors');
+var Suite = require('./suite');
```
plroebuck:
```js
var EVENT_FILE_PRE_REQUIRE =  Suite.constants.EVENT_FILE_PRE_REQUIRE;
var EVENT_FILE_REQUIRE = Suite.constants.EVENT_FILE_REQUIRE;
var EVENT_FILE_POST_REQUIRE = Suite.constants.EVENT_FILE_POST_REQUIRE;
```

lib/interfaces/qunit.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var Test = require('../test');
+var Suite = require('../suite');
```
plroebuck:
```js
var EVENT_FILE_PRE_REQUIRE =  Suite.constants.EVENT_FILE_PRE_REQUIRE;
```

lib/runner.js
```diff
@@ -1027,11 +989,80 @@ function extraGlobals() {
...
+   */
+  EVENT_HOOK_END: 'hook end',
+  /**
+   * Emitted when Root {@link Suite} execution begins (all files have been parsed and hooks/tests are ready for execution)
+   */
+  EVENT_RUN_BEGIN: 'start',
+  /**
+   * Emitted when Root {@link Suite} execution has been delayed via `delay` option
+   */
+  EVENT_RUN_DELAY: 'waiting',
```
plroebuck: Think this makes the pairing any more obvious?
`EVENT_RUN_DELAY` => `EVENT_RUN_DELAY_BEGIN`
`EVENT_ROOT_SUITE_RUN` => `EVENT_RUN_DELAY_END`

lib/runner.js
```diff
@@ -1,8 +1,5 @@
 'use strict';
 
-/**
- * @module Runner
- */
```
plroebuck: Why was this removed?

lib/stats-collector.js
```diff
@@ -2,14 +2,22 @@
 
 /**
  * Provides a factory function for a {@link StatsCollector} object.
- * @private
  * @module
```
plroebuck:
```js
 * @module StatsCollector
```

docs/api-tutorials/custom-reporter.md
boneskull: - [ ] fix

docs/api-tutorials/custom-reporter.md
boneskull: - [ ] remove "corresponding" stuff

docs/api-tutorials/custom-reporter.md
boneskull: - [ ] link to delayed root suite docs

docs/api-tutorials/custom-reporter.md
boneskull: - [ ] passive voice

lib/suite.js
boneskull: we have not defined the order in which tags appear in docstrings, so I'm going to decline to do this unless we have some way to lint and automate this.

lib/suite.js
```diff
@@ -420,6 +419,120 @@ Suite.prototype.eachTest = function(fn) {
...
+  EVENT_SUITE_ADD_HOOK_BEFORE_EACH: 'beforeEach',
+  /**
+   * Emitted after a child `Suite` has been added to a `Suite`
+   */
+  EVENT_SUITE_ADD_SUITE: 'suite',
+  /**
+   * Emitted after a `Test` has been added to a `Suite`
+   */
+  EVENT_SUITE_ADD_TEST: 'test'
+};
```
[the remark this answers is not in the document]
boneskull: I can change these, but I think they need to be deprecated.  The usage of these events is unknown to me (Github search isn't giving me much).  We can listen on the [`newListener` event](https://nodejs.org/api/events.html#events_class_eventemitter) to print a deprecation notice.

- [ ] rename
- [ ] deprecate

lib/test.js
boneskull: habit.  `undefined` can be redefined in older browsers.  we don't support any of those anymore (IE11 disallows this), so `undefined` is safe to use.

- [ ] use `undefined`

`null` needs to die.

lib/test.js
boneskull: unfortunately, `pending` is not a state, it's a prop.  it *should* be a state.  there are no other states.

test/integration/helpers.js
boneskull: does it?  not related to these changes... I'm not removing anything.

lib/test.js
plroebuck: To me, the code gets messy here, having other states possible but not accounted for. Think both of them would be good to describe states when not (yet) pass/fail.

test/integration/helpers.js
plroebuck: Saying you ignored the `err.message` (from `SyntaxError` thrown by `JSON.parse()`) when handcrafting your own `Error` for display.

lib/test.js
boneskull: I'm not going to refactor the pending stuff in this PR, but I can add an initial state.

- [ ] add "initial" state

boneskull: this would necessarily replace the `undefined` business

plroebuck: I was only in reference to adding the pending state constant `XXX.state = STATE_PENDING` when `XXX.pending === true`, not changing the pending processing logic. But if that's hard, let's do it in a different PR...

lib/runner.js
plroebuck: Had an exact example but browser crashed.
But `os` module provides similar division of `constants`.

```js
var os = require('os')
os.constants.errno
os.constants.signals
os.constants.priority
```

boneskull: I see

lib/suite.js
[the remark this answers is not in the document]
boneskull: They're undocumented, Mocha doesn't use them, and there are no *known* use cases.

If there *are* use cases, we'll find out when people complain that they're getting deprecation messages.  😝

lib/test.js
boneskull: hmm.   a problem with a truthy default value for `Test#state` is that some 3rd party reporter might have code that does this:

```js
runner.on('test end', test => {
  if (test.state && test.state !== 'failed') {
    // something
  }
});
```

adding another state, e.g. `test.state = 'init'` would break this assumption.

the reason for setting `test.state = something` initially is really just... not a good one.  I'm unsure how to document an instance variable without an actual assignment statement.  that's the reason.  maybe best to punt on it until/if that becomes clear.

lib/test.js
boneskull: right, yes, generally nothing is easy.

test/integration/helpers.js
boneskull: oh, right, you're referring to the original code, pre-PR.  usually I don't find  JSON syntax errors to be helpful at all in this case.

it's something like "unexpected token blah at position whatever" which means ... it doesn't actually matter what it means, because execution failed and mocha output non-JSON.  so, we don't need to fix broken JSON--we need to ensure the fixture and arguments are correct.

test/integration/helpers.js
boneskull: yes, I think so. that is odd indeed.

lib/runner.js
boneskull: OK, so I can do that, but what I still don't want to do is stuff everything in one file.  that should be a reasonable compromise

boneskull: but we still need the prefixes for autocompletion.

boneskull: bah, actually, some modules do it `os`-style, and others do it `fs`-style, and I'm tired of working on this PR, so I'm going to leave it.  if we nest them it'll be more annoying for users to reference them anyway.

lib/suite.js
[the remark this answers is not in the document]
boneskull: I'll add a message to that effect.  kind of unprecedented, but whatever:

> Event "foo" is deprecated.  Please let the Mocha team know about your use case: https://git.io/v6Lwm'

lib/test.js
boneskull: *punting*

lib/test.js
boneskull: *punting*

docs/api-tutorials/custom-reporter.md
plroebuck: Many of these are now marked as fixed, but without the fixes... why?

lib/interfaces/bdd.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var Test = require('../test');
+var Suite = require('../suite');
```
plroebuck:
```js
var EVENT_FILE_PRE_REQUIRE = Suite.constants.EVENT_FILE_PRE_REQUIRE;
```

lib/interfaces/exports.js
```diff
@@ -22,7 +22,7 @@ var Test = require('../test');
 module.exports = function(suite) {
```
plroebuck:
```js
var EVENT_FILE_REQUIRE = Suite.constants.EVENT_FILE_REQUIRE;
```

lib/interfaces/tdd.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var Test = require('../test');
+var Suite = require('../suite');
```
plroebuck:
```js
var EVENT_FILE_PRE_REQUIRE = Suite.constants.EVENT_FILE_PRE_REQUIRE;
```

lib/runner.js
```diff
@@ -12,7 +9,10 @@ var utils = require('./utils');
 var inherits = utils.inherits;
 var debug = require('debug')('mocha:runner');
 var Runnable = require('./runnable');
-var createStatsCollector = require('./stats-collector');
+var Suite = require('./suite');
+var Test = require('./test');
```
plroebuck: Destructured events are missing...

lib/runner.js
```diff
@@ -873,8 +833,8 @@ Runner.prototype.run = function(fn) {
   if (this._delay) {
     // for reporters, I guess.
     // might be nice to debounce some dots while we wait.
-    this.emit('waiting', rootSuite);
-    rootSuite.once('run', start);
+    this.emit(constants.EVENT_RUN_DELAY, rootSuite);
+    rootSuite.once(Suite.constants.EVENT_ROOT_SUITE_RUN, start);
```
plroebuck: Still think these two would work better as `EVENT_RUN_DELAY_BEGIN` and `EVENT_RUN_DELAY_END`...

test/integration/helpers.js
plroebuck: After having just gone through debugging with this exact code, `SyntaxError` is definitely not the only error that comes through here. The original `err.message` should be part of the new `Error` message.

lib/runner.js
boneskull: hmm, ok

boneskull: actually I can't move it, because the runner needs it, but I can add a new event to the runner which repeats it, so reporters can use it. ok.

lib/runner.js
boneskull: because it conflicts with the class name.  we're using `@module` wrong, basically.

when referring to e.g. `Runner#run`, we want to write `Runner#run`; when we use modules this way we have to do `module:Runner~Runner#run` or some such nonsense.

since all/most of these modules are exported via `Mocha` itself (e.g. `Mocha.Runner`), and the default export is the class (`Runner`) we don't need to use `@module` _in these cases_.

the docstrings are screwy anyway, but since I needed to modify API docs I thought I'd try to make it consistently output for constant-defining classes, which it is now.

lib/stats-collector.js
boneskull: ...`StatsCollector` is a typedef.  we don't want to conflicting names.  it seems like it should refer to what you'd actually call require with: `require('mocha/lib/stats-collector')`, so `stats-collector`.

it defaults to `lib/stats-collector`, which is fine for now, I guess.

test/unit/runner.spec.js
boneskull: I don't know... please make an issue

test/integration/helpers.js
boneskull: please make an issue to explore if this can safely be removed

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+
+To use this reporter, execute `mocha --reporter /path/to/my-reporter.js`.
+
+For further examples, the built-in reporter implementations are the [best place to look](https://github.com/mochajs/mocha/tree/master/lib/reporters). The [integration tests](https://github.com/mochajs/mocha/tree/master/test/reporters) may also be helpful.
+
+## The `Base` Reporter Class
+
+[Base] is an "abstract" class. It contains console-specific settings and utilities, commonly used by built-in reporters. Browsing the built-in reporter implementations, you may often see static properties of [Base] referenced.
+
+It's often useful--but not necessary--for a custom reporter to extend [Base].
```
plroebuck: Add spaces around your dbl-hyphens so spelling can be checked more easily

lib/reporters/spec.js
```diff
@@ -36,11 +43,11 @@ function Spec(runner) {
...
-  runner.on('start', function() {
+  runner.on(EVENT_RUN_BEGIN, function() {
     console.log();
   });
 
-  runner.on('suite', function(suite) {
+  runner.on(EVENT_SUITE_BEGIN, function(suite) {
     ++indents;
     console.log(color('suite', '%s%s'), indent(), suite.title);
   });
```
plroebuck: Why was `'suite end'` not done (L55)?

lib/runner.js
boneskull: I think I also removed `@module Suite`.

lib/stats-collector.js
boneskull: I mean, the docstrings need cleanup.  I am not convinced we're all on the same page about what belongs in the API docs and how best to present it.  ...or even how to come to that decision.  Maybe someone can just propose something and throw it in a PR and we can go from there?

docs/api-tutorials/custom-reporter.md
boneskull: I've never heard of that issue before, can you explain more?

I can say with certainty that I don't love double-hyphens!  Maybe there's a smart hyphen transformation...

lib/reporters/spec.js
boneskull: missed it.

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+      .on(EVENT_TEST_PASS, test => {
+        // Test#fullTitle() returns the suite name(s)
+        // prepended to the test title
+        console.log(`${this.indent()}pass: ${test.fullTitle()}`);
+      })
+      .on(EVENT_TEST_FAIL, (test, err) => {
+        console.log(
+          `${this.indent()}fail: ${test.fullTitle()} - error: ${err.message}`
+        );
+      })
```
plroebuck: `.on(EVENT_TEST_PENDING,`?

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+  EVENT_RUN_END,
+  EVENT_TEST_FAIL,
+  EVENT_TEST_PASS,
+  EVENT_SUITE_BEGIN,
+  EVENT_SUITE_END
+} = Mocha.Runner.constants;
+
+// this reporter outputs test results, indenting two spaces per suite
+class MyReporter {
+  constructor(runner) {
```
plroebuck:
```js
    const self = this;
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+    this._indents = 0;
+
```
plroebuck:
```js
    this.stats = runner.stats;
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+    runner
+      .once(EVENT_RUN_BEGIN, () => {
+        console.log('start');
+      })
+      .on(EVENT_SUITE_BEGIN, () => {
+        this.increaseIndent();
```
plroebuck:
```js
       self.increaseIndent();
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+      })
+      .on(EVENT_SUITE_END, () => {
+        this.decreaseIndent();
```
plroebuck:
```js
       self.decreaseIndent();
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+      })
+      .on(EVENT_TEST_PASS, test => {
+        // Test#fullTitle() returns the suite name(s)
+        // prepended to the test title
+        console.log(`${this.indent()}pass: ${test.fullTitle()}`);
```
plroebuck:
```js
        console.log(`${self.indent()}pass: ${test.fullTitle()}`);
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+          `${this.indent()}fail: ${test.fullTitle()} - error: ${err.message}`
+        );
+      })
+      .once(EVENT_RUN_END, function() {
+        // this is Runner
+        console.log(
+          `end: ${this.stats.passes}/${this.stats.passes +
+            this.stats.failures} ok`
+        );
+      });
```
plroebuck:
```js
        console.log(
          `end: ${self._stats.passes}/${self._stats.tests} ok`
        );
```

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+        // prepended to the test title
+        console.log(`${this.indent()}pass: ${test.fullTitle()}`);
+      })
+      .on(EVENT_TEST_FAIL, (test, err) => {
+        console.log(
+          `${this.indent()}fail: ${test.fullTitle()} - error: ${err.message}`
+        );
+      })
+      .once(EVENT_RUN_END, function() {
+        // this is Runner
```
plroebuck: Beyond confusing as L48 implies `this` changes between event handlers; made all handler code reference `self` variable to remove any confusion.

docs/api-tutorials/custom-reporter.md
plroebuck: Added `_stats` field (as all `Base`-derived classes would have).

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+          `end: ${this.stats.passes}/${this.stats.passes +
+            this.stats.failures} ok`
+        );
+      });
+  }
+
+  indent() {
+    return Array(this._indents)
+      .fill('  ')
+      .join('');
```
plroebuck: Simplified: `Array(this._indents).join('  ');`

docs/api-tutorials/custom-reporter.md
plroebuck: Would prefer:
`It's often useful--but not necessary--for`
changed to:
`It's often useful -- but not necessary -- for`
or
`It's often useful (but not necessary) for`

docs/api-tutorials/custom-reporter.md
```diff
@@ -0,0 +1,120 @@
...
+Mocha adds a `stats` property of type [StatsCollector](/api/module-lib_stats-collector.html) to the reporter's `Runner` instance (passed as first argument to constructor).
+
+## Events
+
+A reporter should listen for events emitted from the `runner` (a singleton instance of [Runner]).
+
+The event names are exported from the `constants` property of `Mocha.Runner`:
+
+| Constant             | Event Name  | Event Arguments | Description                                                                                 |
+| -------------------- | ----------- | --------------- | ------------------------------------------------------------------------------------------- |
```
plroebuck: Possible to make `Event Name` column larger (so "suite end" fits on single line)?

docs/api-tutorials/custom-reporter.md
plroebuck: Or maybe just use `const stats = runner.stats` (instead of class member) and sidestep the issue...

docs/api-tutorials/custom-reporter.md
boneskull: I don't get why you want `self` here.  this is what lambdas are for...

docs/api-tutorials/custom-reporter.md
boneskull: yeah, let's just do that.

it's not confusing if you know the diff between lambda and function, and what `this` is bound to on `EventEmitter` listeners, *but* people tend not to notice.

docs/api-tutorials/custom-reporter.md
boneskull: I'm not sure what you mean.  every row is on its own line.  soft wraps?

docs/api-tutorials/custom-reporter.md
boneskull: meh

docs/api-tutorials/custom-reporter.md
plroebuck: yes, but if pending tests occur, the number of tests would be incorrect...

docs/api-tutorials/custom-reporter.md
plroebuck: It was my opinion that this example was meant to _teach_ about reporting in Mocha. While not technically needed, use of the `self` variable showed intent (well, moreso within the listener).

docs/api-tutorials/custom-reporter.md
plroebuck: If you view the displayed MD (vs. the raw), the `Event Name` column is one character too small and causes "suite end" to be wrapped (i.e., "suite\nend" within cell)

package-scripts.js
```diff
@@ -281,18 +281,27 @@ module.exports = {
...
       },
       preprocess: {
-        script:
-          'md-magic --config ./scripts/markdown-magic.config.js --path docs/index.md',
-        description: 'Preprocess documenation',
-        hiddenFromHelp: true
+        default: {
+          script:
+            'md-magic --config ./scripts/markdown-magic.config.js --path docs/index.md',
+          description: 'Preprocess documenation',
```
plroebuck:
```js
          description: 'Preprocess documentation',
```

docs/api-tutorials/custom-reporter.md
boneskull: oh. probably not without extra effort. this is rendered to html anyway.

package-scripts.js
boneskull: agh

PR #3658, opened 2019-01-06
Reduce CI waste

### LINKED ISSUE #3626: 🛠️ Repo: Avoid running tests in CI for documentation changes
**If you're knowledgable about Travis CI and/or GitHub integrations**, this might be a good issue for you.  It's helpful even if you just tell me it's not possible!

What I'd like to do is this:

When changes are *only* documentation updates, **only run the Markdown linter** on Travis CI.  That means no tests, no code coverage, and for the love of Pete, no AppVeyor.

The command to run these checks would be `npm start lint.markdown`.

If AppVeyor *must* be asked to run a build, it should abort immediately with a result of success.

Code coverage won't be computed if the tests don't run on Travis CI.

Munter:
These changes attempt to avoid the unnecessary work of testing the library for source code changes that don't affect the library.

## Appveyor

Appveyor has a built-in filtering mechanism that allows us to opt out based on file matching patterns in the config file. This is relatively easy to set up and has been configured in a way so we only trigger appveyor tests when we change things that affect the outcome of node tests.

## Travis

Travis doesn't have an easy way to opt out of a build or a job in the same way. Here we need to do a manual git diff and compare the changed files with a set of minimatch patterns. This has been centralized in `scripts/test-runner-filter`. The problem is that we can only run this script at a time after we have installed our npm dependencies, meaning we'll spend a lot of time booting machines, installing apt dependencies, rolling in caches, installing npm dependencies etc, before we even know to skip any work. For that reason, I've left the smoke tests alone, since the running of the test itself is just a fraction of the job runtime.

Instead, I've tried to opt out of the longest running tests that do full tests in node. The slowest ones of these being the saucelabs browser run and the coverage collector.

I'll do a few docs and library changes on this PR, which we can revert before merging, just to check that the CI setup is running as expected

Closes #3626


Munter:
Result of the docs-only change:

Reduced time spent on library testing. Total Travis runtime ~7 minutes vs ~13 minutes (excluding time waiting for worker nodes availability)

[Full test build](https://travis-ci.org/mochajs/mocha/builds/475850698)
[docs-only change](https://travis-ci.org/mochajs/mocha/builds/475859014)

Munter:
The last commit to trigger the full test suite has run as expected, Appveyor and all Travis tests.

The total runtime of the Apveyor tests was 23 minutes, Travis builds at 28 minutes. Since we're triggering both PR and branch builds at the same time on a PR branch push, the actual waiting time for feedback for a developer might not be `Math.max(...buildTimes)`, but something slower, since it's likely there is a waiting time for pod availability to run test jobs.

With a docs-only push the appveyour waiting time is completely removed and the Travis waiting time is reduced to something around 13 minutes, a halving of the runtime.

It's important to note that the Travis builds also fire both for a push and for a PR. The push file change evaluation results in only the files changed in the commit range since last build on the branch, while the PR build evaluates against all commits on the branch. So if you push a docs change to a branch that also contains library changes, the build times for the Travis PR build will be the same as a normal library change

scripts/test-runner-filter.js
```diff
@@ -0,0 +1,48 @@
...
+  const childProcess = require('child_process');
+  const gitDiffRange = (
+    process.env.TRAVIS_COMMIT_RANGE || 'master..HEAD'
+  ).replace('...', '..');
+  const changedFiles = childProcess
+    .execSync(`git diff --name-only ${gitDiffRange}`, {encoding: 'utf8'})
+    .split('\n')
+    .slice(0, -1);
+
+  const testRelatedFilePatterns = [
```
craigtaub: How come there are files in this block that arent in the AppVeyor block?
Also would we not need eslint config here?

.travis.yml
```diff
@@ -35,10 +35,12 @@ cache:
 
 jobs:
   include:
-    - script: COVERAGE=1 npm start test.node
+    - before_script: node scripts/test-runner-filter.js || travis_terminate 0
```
boneskull: can we not use `before_install`?

appveyor.yml
```diff
@@ -4,6 +4,20 @@
...
+  files:
+    - bin/
+    - lib/
+    - test/
+    - .mocharc.yml
+    - .wallaby.js
+    - .appveyor.yml
+    - index.js
+    - package.json
+    - package-lock.json
```
vkarpov15: More a question than anything: [`package-lock.json` can't be published](https://docs.npmjs.com/files/package-lock.json), so why put it here? For compiled output?

scripts/test-runner-filter.js
Munter: Currently Appveyor only runs node tests, so all browser test related files do not need to trigger an appveyor build

.travis.yml
Munter: Then the node modules are not in place to run the script. We could rewrite it to a bash script I guess

appveyor.yml
Munter: We are using `npm ci` on our CI servers to install the exact dependencies noted in `package-lock.json`. This means a build must be triggered if the file changes because changes in dependencies can affect test outcomes. This has nothing to do with what runs on a consumers installation

scripts/test-runner-filter.js
```diff
@@ -0,0 +1,48 @@
...
+    console.log('No library related files have changed. Skipping test');
+    process.exit(1);
+  }
+
+  console.log('Library related files have changed. Continuing full tests');
+} catch (err) {
+  console.log(
+    'Caught an error while checking git diff against test file patterns:'
+  );
+  console.log(err);
```
plroebuck: `console.error(err);`

scripts/test-runner-filter.js
```diff
@@ -0,0 +1,48 @@
...
+  if (!shouldRunTest) {
+    console.log('No library related files have changed. Skipping test');
+    process.exit(1);
+  }
+
+  console.log('Library related files have changed. Continuing full tests');
+} catch (err) {
+  console.log(
+    'Caught an error while checking git diff against test file patterns:'
+  );
```
plroebuck: Don't think this `console.log` is necessary.

scripts/test-runner-filter.js
```diff
@@ -0,0 +1,48 @@
...
+    'index.js',
+    'package.json',
+    'package-lock.json',
+    'package-scripts.json',
+    '.nycrc',
+    '.travis.yml',
+    'browser-entry.js',
+    'karma.conf.js',
+    'mocha.*'
+  ];
```
plroebuck: Should these be ordered by likelihood of being a hit, then maybe lexically?
Would think `'lib/**/*'`, `'test/**/*'`, `'bin/**/*'` should be first three (in that order). Alphabetize the rest.

scripts/test-runner-filter.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  console.log('Library related files have changed. Continuing full tests');
+} catch (err) {
+  console.log(
+    'Caught an error while checking git diff against test file patterns:'
+  );
+  console.log(err);
+  console.log('Running the full test suite as a consequence');
+}
```
plroebuck: Wouldn't this file be better named "scripts/travis-run-filter.js" (or something that includes "travis" in the name)?

plroebuck:
Nice job!

.travis.yml
boneskull: that's probably reasonable, but doesn't need to happen in this PR.

scripts/test-runner-filter.js
boneskull: agree

craigtaub:
@Munter do you have any time to fix conflicts? If not i can.

craigtaub:
Once this is green will be merging in a couple of days.

appveyor.yml
```diff
@@ -4,6 +4,20 @@
 
 ## General configuration
 version: '{build}'
+
+only_commits:
```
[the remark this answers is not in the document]
craigtaub: Think its actually less with `only_commits`. i.e.
```
skip_commits:
  files:
    - docs/*
    - example/*
    - assets/*
    - .github/*
    - scripts/*
    ... many more
```

.travis.yml
```diff
@@ -36,20 +36,22 @@ cache:
...
     - &node
+      before_script: node scripts/travis-run-filter.js || travis_terminate 0
       script: npm start test.node
       node_js: '10'
 
     - <<: *node
       node_js: '8'
 
-    - script: npm start test.bundle test.browser
-      # XXX: update when canvas supplies a prebuilt binary for Node.js v12.x
```
juergba: I would not remove this comment, this task is still pending.

index.js
```diff
@@ -1,3 +1,4 @@
 'use strict';
 
+// The real stuff is all in /lib
```
juergba: This change can be reverted see commit "Core library change to prove that a full CI test cycle is run"

craigtaub: Ah your right. Good spot.

craigtaub:
> What about the different linting config files, shouldn't they trigger the entire tests?

Good point, have pushed a change adding them. Let me know if you have any thoughts on it.

> Why is the Appveyor test not running? There is a code change in "index.js"

Seems to work as expected (see 2nd to last commit which touches it).

juergba:
> Let me know if you have any thoughts on it.

I'm unsure about following statements: 

> For that reason, I've left the smoke tests alone, since the running of the test itself is just a fraction of the job runtime.
[...]
Instead, I've tried to opt out of the longest running tests that do full tests in node. The slowest ones of these being the saucelabs browser run and the coverage collector.

The smoke tests seem to run independently of any filter. There is no `before_script` for the lint test neither. Do the lint test run independently as well? Then adding lint config files to our filter wouldn't make any sense. Can you do some test for this?

juergba:
Outdated, since we migrated from Travis/Appveyor to Github Actions => closing.

Issue #3065, opened 2017-10-11
Runner `end` event can’t be caught if the `grep` option matches no tests


Mr0grog:
### Prerequisites

* [x] Checked that your issue isn't already filed by cross referencing [issues with the `common mistake` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Acommon-mistake%20)
* [x] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
* [x] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
* [x] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with:
`node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description

When running Mocha programmatically or in a browser, `mocha.run()` returns the test runner instance and event handlers can be attached to it. However, when the `grep` option is used and it matches no tests, ]it emits the `end` event synchronously](https://github.com/mochajs/mocha/blob/master/lib/runner.js#L621-L623) before the runner is ever returned from `mocha.run()` and before any third-party code could attach a listener.

I know I can work around this by providing a callback to `mocha.run()`, but it still seems reasonable to expect that you’d always have an opportunity to attach an event listener before the `end` event fires.

### Steps to Reproduce

If I have a test suite at `/single-test.js` that looks like:

```js
describe('Test suite', function () {
  it('A single passing test', function() {
    // this passes
  });
});
```

And I run the following:

```js
const Mocha = require('mocha');
const mocha = new Mocha({ui: 'bdd', grep: /unmatchable/});
mocha.addFile('/single-test.js');
mocha
  .run()
  .on('end', () => console.log('End event handled!'));
```

**Expected behavior:** I expect “End event handled!” to be logged to the console.

**Actual behavior:** The test suite runs, but my handler does not.

**Reproduces how often:** 100%

### Versions

Mocha 4.0.1

ScottFreeCode:
I wonder if fixing #2753 would automatically fix this too.

Mr0grog:
Ah! Yes, that definitely would (assuming `end` is guaranteed to come after `start` :P )

boneskull:
duplicate of #2753

PR #3660, opened 2019-01-06
Fix: listen to 'start' and 'end' events when using Mocha programmatically

### LINKED ISSUE #2753: unable to listen to start event during programmatic run

juergba:
### Description

`mocha.run()` returns a `runner` instance. Since the `EVENT_RUN_BEGIN` (`start`) event is emitted synchronously, it's impossible to register an `EVENT_RUN_BEGIN` listener in time. The event is emitted, but never catched.

When no tests are provided - eg. by a too restrictive `grep` expression - then neither `EVENT_RUN_BEGIN` nor `EVENT_RUN_END` (`end`) events can be listened to.

```js
var mocha = new Mocha();

mocha.addFile("../eventStart.js");

mocha.run()
  .on(EVENT_RUN_BEGIN, () => console.log('start'))    // never runs
  .on(EVENT_RUN_END, () => console.log('end'));       // never runs when no tests are provided
```

### Description of Change

We delay the start of the root Suite by `Runner.immediately`. The order of the event emissions remains unchanged.

### Applicable issues

#3395 (backed out)
#3331 
#3065
closes #2753

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -939,7 +939,9 @@ Runner.prototype.run = function(fn) {
     this.emit(constants.EVENT_DELAY_BEGIN, rootSuite);
     rootSuite.once(EVENT_ROOT_SUITE_RUN, start);
   } else {
-    start();
+    Runner.immediately(function() {
+      start();
+    });
   }
 
   return this;
--- a/test/unit/mocha.spec.js
+++ b/test/unit/mocha.spec.js
@@ -487,6 +487,16 @@ describe('Mocha', function() {
       }, 'not to throw');
     });
 
+    it('should catch the `start` event if no tests are provided', function(done) {
+      var mocha = new Mocha(opts);
+      mocha.run().on('start', done);
+    });
+
+    it('should catch the `end` event if no tests are provided', function(done) {
+      var mocha = new Mocha(opts);
+      mocha.run().on('end', done);
+    });
+
     describe('#reporter("xunit")#run(fn)', function() {
       // :TBD: Why does specifying reporter differentiate this test from preceding one
       it('should not raise errors if callback was not provided', function() {


juergba:
I messed up my branch, now everything is ok. Hopefully this will reopen my PR.

plroebuck:
> There is no way to register EventListener to the runner instance when Mocha runs programmatically with `mocha.run()`.

Isn't this already possible with [delayed root suite](https://mochajs.org/#delayed-root-suite)?

juergba:
@plroebuck 
I don't expect the user to know the answer, if I (we) don't know it without analyzing code and try-and-error testing. It wouldn't be an user-friendly solution.

yes, but... : if there are test files found, which include the (delay) run() function. (no: `waiting`)
no: if there are no test files found (grep)
no: `waiting` event doesn't work at all, which is emitted before `start` event

`--delay` does alter the event firing. `start` and `end` events are emitted twice, probably one additional set while the files get loaded.

boneskull:
(may be semver-patch, dunno)

plroebuck:
Impact of user's event handlers now run prior to Mocha's own?

juergba:
I have to do more testing.
When user calls `mocha.run()` twice on the same `Mocha` instance, the `Runner` instance is created only once now.

PR #3665, opened 2019-01-08
Removed MOCHA_COLORS environment variable

### LINKED ISSUE #3641: Remove MOCHA_COLORS environment variable
### Description

This environment variable was introduced in Sep 2013 by #965.
Later, we updated to use [supports-color](https://github.com/chalk/supports-color/) module for color support in Feb 2015 by #1545.

This module [later gained its own pseudo-standard environment variable](https://github.com/chalk/supports-color/pull/31) `FORCE_COLOR` which provides the same function in Jan 2017.

As `MOCHA_COLORS` provides _redundant_ functionality, it should be removed.

#### lib/reporters/base.js
```js
exports.useColors =
  !process.browser &&
  (supportsColor.stdout || process.env.MOCHA_COLORS !== undefined);
```
 would become
```js
exports.useColors = !process.browser && supportsColor.stdout;
```

waseemahmad31:
### Description of the Change

Remove MOCHA_COLORS environment variable

The [supports-color](https://github.com/chalk/supports-color/) module's environment variable `FORCE_COLOR` provides the same functionality.

### Benefits

Redundant. One less thing to maintain.

### Applicable issues

Fixes #3641
patch release

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -28,9 +28,7 @@ var isatty = tty.isatty(1) && tty.isatty(2);
  * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors =
-  !process.browser &&
-  (supportsColor.stdout || process.env.MOCHA_COLORS !== undefined);
+exports.useColors = !process.browser && supportsColor.stdout;
 
 /**
  * Inline diffs instead of +/-
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -27,7 +27,13 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 /**
  * Enable coloring by default, except in the browser interface.
  */
-
+if (process.env.MOCHA_COLORS) {
+  utils.deprecate(
+    '"MOCHA_COLORS" is deprecated and will be removed from a future version of Mocha. Use "FORCE_COLOR" instead.'
+  );
+  delete process.env.MOCHA_COLORS;
+  process.env.FORCE_COLOR = 1;
+}
 exports.useColors = !process.browser && supportsColor.stdout;
 
 /**
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -27,13 +27,6 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 /**
  * Enable coloring by default, except in the browser interface.
  */
-if (process.env.MOCHA_COLORS) {
-  utils.deprecate(
-    '"MOCHA_COLORS" is deprecated and will be removed from a future version of Mocha. Use "FORCE_COLOR" instead.'
-  );
-  delete process.env.MOCHA_COLORS;
-  process.env.FORCE_COLOR = 1;
-}
 exports.useColors = !process.browser && supportsColor.stdout;
 
 /**
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -17,6 +17,7 @@ const {loadOptions} = require('./options');
 const commands = require('./commands');
 const ansi = require('ansi-colors');
 const {repository, homepage, version, gitter} = require('../../package.json');
+const {deprecate} = require('../utils');
 
 /**
  * - Accepts an `Array` of arguments
@@ -31,6 +32,16 @@ exports.main = (argv = process.argv.slice(2)) => {
   module.paths.push(process.cwd(), path.resolve('node_modules'));
 
   Error.stackTraceLimit = Infinity; // configurable via --stack-trace-limit?
+
+  if (process.env.MOCHA_COLORS) {
+    deprecate(
+      'Support for deprecated "MOCHA_COLORS" environment variable was removed from Mocha. Use "FORCE_COLOR" instead.'
+    );
+    if (argv.indexOf('--colors') < 0 && argv.indexOf('-c') < 0) {
+      argv.push('--colors');
+    }
+    delete process.env.MOCHA_COLORS;
+  }
 
   yargs
     .scriptName('mocha')


lib/reporters/base.js
```diff
@@ -28,9 +28,7 @@ var isatty = tty.isatty(1) && tty.isatty(2);
  * Enable coloring by default, except in the browser interface.
  */
 
-exports.useColors =
-  !process.browser &&
-  (supportsColor.stdout || process.env.MOCHA_COLORS !== undefined);
+exports.useColors = !process.browser && supportsColor.stdout;
```
outsideris: How about this?

```js
exports.useColors = supportsColor && supportsColor.stdout;
```

Because `process.browser` is already tested in [L13](https://github.com/mochajs/mocha/pull/3665/files#diff-5c8a5788ecd4b4af2a29bc27ac4ae986R13).

waseemahmad31: That's a good suggestion 👍 – thanks @outsideris

waseemahmad31: In test cases value of ```useColors``` is considered false when value of ```process.browser``` is not null. And if we use ```exports.useColors = supportsColor && supportsColor.stdout; ``` then value of the ```useColors``` will be null when value of ```process.browser``` is not null. And that will be not correct behavior. So either we can  write ```exports.useColors = supportsColor != null && supportsColor.stdout;``` or leave it as it is.

plroebuck: Leave it as is then -- more important _not_ to break anything...

boneskull:
Wait.  I don't think we should merge this.  

[This is why](https://github.com/search?q=MOCHA_COLORS&type=Code).

Can we instead detect the use of `MOCHA_COLORS` and print a deprecation notice and migration info?  This will hopefully avoid a torrent of issues.

boneskull:
(The reason why we never bothered to do this is that it interferes w/ our integration tests, apparently.)

boneskull:
We could even go ahead and print the deprecation notice, then remove `MOCHA_COLORS` from the environment and set `FORCE_COLOR=1`... too invasive?

waseemahmad31:
I think it will be good if we show the deprecation message and then remove `MOCHA_COLORS` from the environment and set `FORCE_COLOR`=1.

plroebuck:
> We could even go ahead and print the deprecation notice, then remove `MOCHA_COLORS` from the environment and set `FORCE_COLOR=1`... too invasive?

Far less invasive would be to check for `process.env.MOCHA_COLORS` at startup (i.e., Mocha-5.2 "/bin/_mocha"). If it is defined, display deprecation notice and **append** `--colors` to user-provided Mocha cmdline. `FORCE_COLOR` is used by many packages; we treat color support as a boolean state, but others may process differently depending on the number of colors available.

lib/reporters/base.js
```diff
@@ -27,7 +27,13 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 /**
  * Enable coloring by default, except in the browser interface.
  */
-
+if (process.env.MOCHA_COLORS) {
+  utils.deprecate(
+    '"MOCHA_COLORS" is deprecated and will be removed from a future version of Mocha. Use "FORCE_COLOR" instead.'
```
plroebuck: Uh, you already removed it...
```js
  utils.deprecate(
    'Support for deprecated "MOCHA_COLORS" environment variable was removed from Mocha. Use "FORCE_COLOR" instead.');
```

lib/reporters/base.js
```diff
@@ -27,7 +27,13 @@ var isatty = tty.isatty(1) && tty.isatty(2);
...
+  );
+  delete process.env.MOCHA_COLORS;
+  process.env.FORCE_COLOR = 1;
```
plroebuck: Environment variables should _always_ be strings...

plroebuck: But this is a rather [heavy-handed](https://www.youtube.com/watch?v=QGc-iPc-9dE) approach...

lib/reporters/base.js
```diff
@@ -27,7 +27,13 @@ var isatty = tty.isatty(1) && tty.isatty(2);
...
+}
```
plroebuck: Think all deprecation-related code belongs [here](https://github.com/mochajs/mocha/blob/master/lib/cli/cli.js) instead.

waseemahmad31:
Added code to show deprication message of `MOCHA_COLORS`  and handled `--colors` argv in the cli.js.

boneskull:
@plroebuck 

> Far less invasive would be to check for `process.env.MOCHA_COLORS` at startup (i.e., Mocha-5.2 "/bin/_mocha"). If it is defined, display deprecation notice and append `--colors` to user-provided Mocha cmdline arguments. `FORCE_COLOR` is used by many packages; we treat color support as a boolean state, but others may process differently depending on the number of colors available.

I don't think `--colors` is equivalent to `FORCE_COLOR` (it still defers the final decision to `supports-color`, which considers `FORCE_COLOR`), but I'll need to confirm that.  It would be great if @waseemahmad31 could confirm, actually.

waseemahmad31:
yes, @boneskull. `--colors` do the same work  as `FORCE_COLOR = true,  FORCE_COLOR = 1 or FORCE_COLOR  = []`

boneskull:
OK, I don't think we should do this at all actually.

I've spent another 30m or so looking at this, and it's a rabbit hole.

1.  We shouldn't append to `process.argv` to do this, as this would potentially conflict with `--no-colors`, `--no-color` or `-C`.  We could special-case it, but that's just replacing code rot with tech debt.
1. We need to decide the priority environment flags take.  I believe environment flags should be secondary only to command-line arguments, meaning they should override configuration files (`{"color": false}`) and `mocha.opts`.
1. We *should* support some form of environment flags; Lerna monorepos in particular  seem to want to use this behavior to avoid maintaining the same settings across packages.
1. We *can* support environment flags "easily" (I said that before, of course) because [yargs supports it](http://yargs.js.org/docs/#api-envprefix); we haven't enabled it.  I *do* want to enable it, but maybe in a near-future minor release.  This takes all our options and allows them to be defined via environment flags having a prefix, e.g., `MOCHA_BAIL=1` would be equivalent to `--bail`.
1. If we enable environment flag support via yargs, **`MOCHA_COLORS=1` is no longer actually deprecated**; it's a valid configuration option, equivalent to `--colors`.

boneskull:
@waseemahmad31 I apologize for the confusion here.

plroebuck:
>     1. We shouldn't append to `process.argv` to do this, as this would potentially conflict with `--no-colors`, `--no-color` or `-C`.  We could special-case it, but that's just replacing code rot with tech debt.
> 
>     2. We need to decide the priority environment flags take.  I believe environment flags should be secondary only to command-line arguments, meaning they should override configuration files (`{"color": false}`) and `mocha.opts`.

Mea culpa. I wrote "append", but meant "prepend" `--no-colors`, which requires no checking of existing `argv` and covers the argument processing order issue.

>     3. We _should_ support some form of environment flags; Lerna monorepos in particular seem to want to use this behavior to avoid maintaining the same settings across packages.
>     4. We _can_ support environment flags "easily" (I said that before, of course) because [yargs supports it](http://yargs.js.org/docs/#api-envprefix); we haven't enabled it.  I _do_ want to enable it, but maybe in a near-future minor release.  This takes all our options and allows them to be defined via environment flags having a prefix, e.g., `MOCHA_BAIL=1` would be equivalent to `--bail`.
>     5. If we enable environment flag support via yargs, **`MOCHA_COLORS=1` is no longer actually deprecated**; it's a valid configuration option, equivalent to `--colors`.

Potential future maintenance nightmare, given the effort to remove this single redundant envvar.

boneskull:
It's not usually trivial to remove things from Mocha, no.

@waseemahmad31 Thanks for your effort on this

waseemahmad31:
@boneskull, @plroebuck thanks for the feedback. Initially, it seems like the easy one to start the first contribution. Anyway thanks for the detailed analysis of the impact of this change.

Issue #3656, opened 2019-01-05
Ensure all errors have code


craigtaub:
We now have error codes for errors generated by mocha (via issue https://github.com/mochajs/mocha/issues/3125) but there are still some missing.
e.g. https://github.com/mochajs/mocha/blob/master/lib/interfaces/common.js#L130

Make sure the others have factory functions with appropriate code, arguments and documentation.

PR #3666, opened 2019-01-08
Fix couple errors missing codes

vkarpov15:
### Requirements

### Description of the Change

#3656 added a couple utils for setting `code` on errors. I fixed a couple here and wanted to make sure I was on the right track before making more changes.

### Alternate Designs

N/A

Codes help identify errors as mocha errors at a glance without using custom error types, and better DX is always good.

### Benefits

:point_up: 

### Possible Drawbacks

None that I can see

### Applicable issues

#3656, #3125. semver-patch

### DIFF
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -11,6 +11,7 @@ const Mocha = require('../mocha');
 const ansi = require('ansi-colors');
 const errors = require('../errors');
 const createInvalidArgumentValueError = errors.createInvalidArgumentValueError;
+const createMissingArgumentError = errors.createMissingArgumentError;
 
 const {
   list,
@@ -258,15 +259,19 @@ exports.builder = yargs =>
 
       // yargs.implies() isn't flexible enough to handle this
       if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
-        throw new Error(
-          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"'
+        throw createMissingArgumentError(
+          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"',
+          '--invert',
+          argv.invert
         );
       }
 
       if (argv.compilers) {
-        throw new Error(
+        throw createInvalidArgumentValueError(
           `--compilers is DEPRECATED and no longer supported.
-          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`
+          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`,
+          '--compilers',
+          argv.compilers
         );
       }
 
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -9,9 +9,10 @@
 
 const Mocha = require('../mocha');
 const ansi = require('ansi-colors');
-const errors = require('../errors');
-const createInvalidArgumentValueError = errors.createInvalidArgumentValueError;
-const createMissingArgumentError = errors.createMissingArgumentError;
+const {
+  createInvalidArgumentValueError,
+  createMissingArgumentError
+} = require('../errors');
 
 const {
   list,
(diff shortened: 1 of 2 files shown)


lib/cli/run.js
```diff
@@ -258,15 +259,19 @@ exports.builder = yargs =>
 
       // yargs.implies() isn't flexible enough to handle this
       if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
-        throw new Error(
-          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"'
+        throw createMissingArgumentError(
+          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"',
+          '--invert',
+          argv.invert
```
craigtaub: Think this should be the "expected argument datatype" (see https://github.com/mochajs/mocha/blob/master/lib/errors.js#L73) ..so I think "string" would be fine

lib/cli/run.js
```diff
@@ -258,15 +259,19 @@ exports.builder = yargs =>
...
         );
       }
 
       if (argv.compilers) {
-        throw new Error(
+        throw createInvalidArgumentValueError(
           `--compilers is DEPRECATED and no longer supported.
-          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`
+          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`,
+          '--compilers',
```
craigtaub: `--` not needed.

test/integration/options.spec.js
```diff
@@ -346,6 +346,26 @@ describe('options', function() {
...
+          'options/grep.fixture.js',
+          args,
+          function(err, res) {
+            if (err) {
+              done(err);
+              return;
+            }
+            expect(res, 'to satisfy', {
+              code: 1,
+              output: /fgrep/
```
craigtaub: Is this the error message? If so can we cover more of the string here? Be really good to be able to check error code (if possible)

test/integration/options.spec.js
```diff
@@ -346,6 +346,26 @@ describe('options', function() {
           done();
         });
       });
+
+      it('fails if no --grep', function(done) {
```
craigtaub: `should throw an error ...`.

lib/cli/run.js
```diff
@@ -258,15 +259,19 @@ exports.builder = yargs =>
 
       // yargs.implies() isn't flexible enough to handle this
       if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
-        throw new Error(
-          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"'
+        throw createMissingArgumentError(
+          '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"',
+          '--invert',
```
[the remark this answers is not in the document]
plroebuck: Actually, it does since it refers to the CLI argument (same as your PR for `--reporter-option`). But its value should be `--fgrep | --grep`

lib/cli/run.js
plroebuck: yep. `string|regexp`

lib/cli/run.js
plroebuck: Perhaps more to the point:
```js
        throw createNotSupportedError(
          `--compilers is DEPRECATED and no longer supported.
          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`
        );
```

test/integration/options.spec.js
plroebuck:
```js
      it('should throw an error when `--invert` used in isolation', function(done) {
```

test/integration/options.spec.js
plroebuck: Using `runMocha` means we're scraping output from stdout|stderr -- we should be able to see `Error.name` and `Error.message`. `Error.code` shouldn't be output if that's what you're asking for.

plroebuck: @craigtaub, where's the corresponding test for invalid reporter option from previous PR?

craigtaub: Not sure one was added, might have been missed. If not we should add one.

lib/cli/run.js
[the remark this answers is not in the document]
craigtaub: Ah your right my mistake

lib/cli/run.js
plroebuck: Not addressed

boneskull: I'll fix it after merge.

PR #3669, opened 2019-01-12
set allowUncaught on hooks

### LINKED ISSUE #2302: allowUncaught doesn't work in hooks

givanse:
### Description of the Change

When passing `--allow-uncaught`, uncaught exceptions thrown inside hooks are not surfaced. This PR enables that. Original issue: https://github.com/mochajs/mocha/issues/2302

### Alternate Designs

The hook-runner is the best place to set this flag, since it has to be read from the parent context (Runner). I can't see any other way of doing given that strong dependency.

It is a missing feature, already marked as a confirmed bug (#2302 )

### Benefits

Tests authors can be more strict about not accepting code that throws errors as side effects. Debugging failing hooks becomes easier.

### Possible Drawbacks

There shouldn't be drawbacks. Tests that throw errors within their hooks are most likely not setting `allowUncaught`, otherwise the browser would hang. So it remains an opt-in feature.

### Applicable issues

closes #2302 , patch release

### DIFF
--- a/test/unit/runner.spec.js
+++ b/test/unit/runner.spec.js
@@ -443,7 +443,20 @@ describe('Runner', function() {
   });
 
   describe('allowUncaught', function() {
-    it('should allow unhandled errors to propagate through', function(done) {
+    var immediately;
+
+    before(function() {
+      immediately = Runner.immediately;
+      Runner.immediately = function(fn) {
+        fn();
+      };
+    });
+
+    after(function() {
+      Runner.immediately = immediately;
+    });
+
+    it('should allow unhandled errors to propagate through', function() {
       var newRunner = new Runner(suite);
       newRunner.allowUncaught = true;
       newRunner.test = new Test('failing test', function() {
@@ -453,7 +466,31 @@ describe('Runner', function() {
         newRunner.runTest();
       }
       expect(fail, 'to throw', 'allow unhandled errors');
-      done();
+    });
+
+    it('should allow unhandled errors in sync hooks to propagate through', function() {
+      suite.beforeEach(function() {
+        throw new Error('allow unhandled errors in sync hooks');
+      });
+      var runner = new Runner(suite);
+      runner.allowUncaught = true;
+      function throwError() {
+        runner.hook('beforeEach', function() {});
+      }
+      expect(throwError, 'to throw', 'allow unhandled errors in sync hooks');
+    });
+
+    it('async - should allow unhandled errors in hooks to propagate through', function() {
+      suite.beforeEach(function(done) {
+        // having `done` triggers the async path
+        throw new Error('allow unhandled errors in async hooks');
+      });
+      var runner = new Runner(suite);
+      runner.allowUncaught = true;
+      function throwError() {
+        runner.hook('beforeEach', function() {});
+      }
+      expect(throwError, 'to throw', 'allow unhandled errors in async hooks');
     });
   });
 
--- a/test/unit/runner.spec.js
+++ b/test/unit/runner.spec.js
@@ -468,6 +468,19 @@ describe('Runner', function() {
       expect(fail, 'to throw', 'allow unhandled errors');
     });
 
+    it('should not allow unhandled errors in hooks to propagate through', function() {
+      suite.beforeEach(function() {
+        throw new Error('this error will not propagate');
+      });
+      var runner = new Runner(suite);
+      try {
+        runner.hook('beforeEach', function() {});
+        expect(true, 'to be', true);
+      } catch (err) {
+        expect(false, 'to be', true);
+      }
+    });
+
     it('should allow unhandled errors in sync hooks to propagate through', function() {
       suite.beforeEach(function() {
         throw new Error('allow unhandled errors in sync hooks');
@@ -481,8 +494,8 @@ describe('Runner', function() {
     });
 
     it('async - should allow unhandled errors in hooks to propagate through', function() {
+      // having `done` triggers the async path
       suite.beforeEach(function(done) {
-        // having `done` triggers the async path
         throw new Error('allow unhandled errors in async hooks');
       });
       var runner = new Runner(suite);
--- a/test/unit/runner.spec.js
+++ b/test/unit/runner.spec.js
@@ -9,6 +9,7 @@ var Test = Mocha.Test;
 var Runnable = Mocha.Runnable;
 var Hook = Mocha.Hook;
 var noop = Mocha.utils.noop;
+var EVENT_HOOK_BEGIN = Runner.constants.EVENT_HOOK_BEGIN;
 var EVENT_TEST_FAIL = Runner.constants.EVENT_TEST_FAIL;
 var EVENT_TEST_RETRY = Runner.constants.EVENT_TEST_RETRY;
 var EVENT_RUN_END = Runner.constants.EVENT_RUN_END;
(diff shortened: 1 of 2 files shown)


test/unit/runner.spec.js
```diff
@@ -416,7 +416,20 @@ describe('Runner', function() {
   });
 
   describe('allowUncaught', function() {
-    it('should allow unhandled errors to propagate through', function(done) {
+    var immediately;
+
+    before(function() {
+      immediately = Runner.immediately;
+      Runner.immediately = function(fn) {
```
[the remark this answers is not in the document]
givanse: Using `before` or `beforeAll` for this suite of tests should make no difference nor affect in any way the one test we already had. I've opted for `before` because it executes less code, granted the gains are marginal. If you have a strong preference, I'm cool with switching to `beforeAll`.

test/unit/runner.spec.js
```diff
@@ -426,7 +439,44 @@ describe('Runner', function() {
...
+    });
+
+    it('should not allow unhandled errors in hooks to propagate through', function() {
+      suite.beforeEach(function() {
+        throw new Error('this error will not propagate');
+      });
+      var runner = new Runner(suite);
+      try {
+        runner.hook('beforeEach', function() {});
+        expect(true, 'to be', true);
```
boneskull: it's difficult to tell what we're trying to express in these tests.  

if you want to ensure that `runner.hook('beforeEach', function() {})` does not throw, you can do:

```js
expect(function() {
  runner.hook('beforeEach', function() {});
}, 'not to throw');
```

which would replace L449-L454 here.  if you *do* expect that call to throw, of course, remove the "not" :wink:

test/unit/runner.spec.js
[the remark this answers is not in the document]
boneskull: no, just wanted to know if there was an intent there, regarding "before all".

I believe you re: `immediately`; I just wonder if there's a better way to do this.  I'm not sure what that is yet.  we have to do this because the code under test is wonky, of course.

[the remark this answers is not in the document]
boneskull: BTW, the conflict in `lib/runner.js` might make this unnecessary.

givanse:
Before you merge I'll resolve the conflicts and address the feedback on the assertions.

test/unit/runner.spec.js
givanse: thanks, will update

givanse:
@boneskull I resolved the conflict and took another approach, instead of `immediately` now I'm using the `hook` event. If you are ok with these changes, then this is ready to merge.

boneskull:
@givanse 	Looks like this is failing some tests

test/unit/runner.spec.js
```diff
@@ -473,37 +460,56 @@ describe('Runner', function() {
...
-      expect(throwError, 'to throw', 'allow unhandled errors in sync hooks');
+
+      var EVENT_HOOK_BEGIN = 'hook';
+      runner.on(EVENT_HOOK_BEGIN, function(hook) {
+        var _run = hook.run;
+        hook.run = function() {
+          function throwError() {
+            _run.call(hook);
+          }
+          var expected = 'allow unhandled errors in async hooks';
```
juergba: should be "sync", not "async": `var expected = 'allow unhandled errors in sync hooks';`

juergba:
@givanse I will review within the next few days, I haven't understood this PR completely, yet.

juergba:
```js
    beforeEach('beforeEach', function() {
        throw new Error('uncaught error');
    })
    it('test', function() {
        console.log('test runs');
    });
});
describe('suite2', function() {    
    it('test2', function() {
        console.log('test runs');
    });
});
```
I would have expected a different result, when run with
- `--no-allow-uncaught`  (default)
- `--allow-uncaught`

Both ways the second `describe()` does run and the result is:
```
suite
    1) "before each" hook: beforeEach for "test"

  suite2
test runs
    √ test2

  1 passing (11ms)
  1 failing

  1) suite
       "before each" hook: beforeEach for "test":
     Error: uncaught error     // Uncaught Error
```
The error type changes from `Error` to `Uncaught Error` though.
What is the sense of allowing uncaught exceptions to propagate? Is there an `uncaughtException` event emitted?

test/unit/runner.spec.js
givanse: ah! probably a typo introduced during conflicts resolution, thanks

givanse:
`Error: uncaught error` vs `Uncaught Error: uncaught error`

The difference is that `Error` is just an entry in the log. Mocha captures the error and logs it. An `Uncaught Error` is an actual JS error that can stop JS execution, this is useful for:
 - Testem `bail_on_uncaught_error`
 - Chrome dev tools `Pause on exceptions`

You can verify this by executing your code snippet with the flag `--inspect-brk`.

test/unit/runner.spec.js
```diff
@@ -453,7 +453,63 @@ describe('Runner', function() {
...
+
+    it('should not allow unhandled errors in hooks to propagate through', function() {
+      suite.beforeEach(function() {
+        throw new Error('this error will not propagate');
+      });
+      var runner = new Runner(suite);
+      expect(function() {
+        runner.hook('beforeEach', function() {});
+      }, 'not to throw');
+    });
```
juergba: I don't trust this test, since `expect(...)` passes even with `runner.allowUncaught = true;`. First it passes, then the test fails. Please use the same logic as of the following two tests.
To be consistent with the existing code, use `noop` instead of `function() {}`.

test/unit/runner.spec.js
```diff
@@ -453,7 +453,63 @@ describe('Runner', function() {
...
+
+    it('should allow unhandled errors in sync hooks to propagate through', function(done) {
+      suite.beforeEach(function() {
+        throw new Error('allow unhandled errors in sync hooks');
+      });
+      var runner = new Runner(suite);
+      runner.allowUncaught = true;
+
+      var EVENT_HOOK_BEGIN = 'hook';
+      runner.on(EVENT_HOOK_BEGIN, function(hook) {
```
juergba: I guess `runner.once` is safer.

test/unit/runner.spec.js
```diff
@@ -453,7 +453,63 @@ describe('Runner', function() {
...
+          expect(throwError, 'to throw', expected);
+          done();
+        };
+      });
+
+      runner.hook('beforeEach', function() {});
+    });
+
+    it('async - should allow unhandled errors in hooks to propagate through', function(done) {
+      // having the `done` argument triggers the async path
```
juergba: Remove this comment.
But please add a short comment with the first test, why we need this monkey patching. In some months nobody will know otherwise.

test/unit/runner.spec.js
```diff
@@ -453,7 +453,63 @@ describe('Runner', function() {
...
+      suite.beforeEach(function(done) {
+        throw new Error('allow unhandled errors in async hooks');
+      });
+      var runner = new Runner(suite);
+      runner.allowUncaught = true;
+
+      var EVENT_HOOK_BEGIN = 'hook';
+      runner.on(EVENT_HOOK_BEGIN, function(hook) {
```
juergba: as above.

juergba:
@givanse Thanks for this PR! It does work, however with the tests I'm not really happy. But I admit I haven't found a better solution so far. I tried with catching the `uncaughtException` event, then patching `Runner.prototype.uncaught()`... all just weird solutions.

I requested some small changes and will merge after that. If you don't agree, please say so, you seem to know what you are doing.

test/unit/runner.spec.js
givanse: Strange, I saw the test fail reliably when I added `runner.allowUncaught = true;`. Anyway, the pattern used in the other two tests is definitely 100% reliable and would provide more consistency.

PR #3671, opened 2019-01-14
WIP: predictable "setImmediate" between hooks

juergba:
### Description

see #3009 

Currently between every hook there is a `setImmediate` call, actually there is even one for non-existing hooks. The following simple test without any hook causes eight `setImmediate` calls, two suites each with four possible hooks:
```js
describe("synchSuite", function () {
    it("myTest", function () { });
  });
```
My first efforts to eliminate all `setImmediate` between hooks and insert instead one between each test failed with many "timeout" errors and huge error stacks with hundreds of lines. I decided therefore to take small steps.

### Description of the Change

The smallest synchronous unit should be:
[nearest `beforeEach` hook + `test` + nearest `afterEach` hook]

The following example should run without `PromiseRejectionHandledWarning`.
```js
describe("synchSuite", function () {
    beforeEach(function () {
      promise = Promise.reject(new Error());
    });
    it("myTest", function () {
      // return promise.catch(function () {console.log('catched it()')});
    });
    afterEach(function () {
      return promise.catch(function () {console.log('catched after() nested')});
    });
  });
```

required changes:
- `beforeEach`: no changes required, since [nearest `beforeEach` hook + `test`] do run synchronously.
- `afterEach`: [`test` + nearest `afterEach` hook] change from asynch to synch call.

### Alternate Designs

- `setImmediate` calls only for existing hooks
- eliminate all `setImmediate` between hooks and insert instead one between each test.

Behavior is more predictable when testing with promises.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -336,9 +336,14 @@ Runner.prototype.hook = function(name, fn) {
     });
   }
 
-  Runner.immediately(function() {
+  // nearest 'beforeEach' + 'it' + nearest 'afterEach' run synchronously
+  if (name === 'afterEach' && suite === this.synchSuite && hooks.length) {
     next(0);
-  });
+  } else {
+    Runner.immediately(function() {
+      next(0);
+    });
+  }
 };
 
 /**
@@ -385,6 +390,7 @@ Runner.prototype.hooks = function(name, suites, fn) {
  */
 Runner.prototype.hookUp = function(name, fn) {
   var suites = [this.suite].concat(this.parents()).reverse();
+  this.synchSuite = this.suite;
   this.hooks(name, suites, fn);
 };
 


boneskull:
What if instead of `setImmediate(next)` we did `Promise.resolve().then(next)`?

This should not cause a tick in either Node.js or a browser (to my knowledge).

The `process.nextTick()` shim uses `setTimeout(fn, 0)` which isn't (also, to my knowledge) going to reproduce the same behavior, even if we did use `process.nextTick()` instead.

A problem is IE11, of course (no `Promise`), so we'd have to pull a polyfill in--and depending on the polyfill, it may be unsuccessful.  Also, I've heard Edge has issues with the microtask queue...

juergba:
It's a complex topic, at the moment I don't see an alternative to `setImmediate`.

The callback of a native JS `promise` is a microtask and is run directly after the `nextTick` (also microtask) queue. These microtasks run several times during a loop, before each of the four main queues as expired timers/intervals, IO events, Immediates, close handlers.

The callbacks of each hook have to be executed before calling the next hook, this can only be garantueed with `setImmediate(next)`. Otherwise, executing hook after hook and just queuing (not running) their callbacks, could starve the IO queue and therefore result in a timeout exceeded error.

So no, up to now, `Promise.resolve`, `nextTick`, `setTimeout(fn, 0)` are not working.

PR #3674, opened 2019-01-16
Reorder cmdline options top-level test suites lexically

plroebuck:
### Description of the Change
~~Reordered each cmdline option's test suite alphabetically.~~
~~**No code changes to test suites.**~~
Implemented the alternative design, creating individual spec files for the various cmdline options
being tested. 

### Alternate Designs
Arguably, it would be _much_ easier to deal with if _each_ option was in its own file.

### Benefits
With this file at almost 850 dense [LOC](https://en.wikipedia.org/wiki/Source_lines_of_code), it will be _far_ easier to work on a particular cmdline option now.
Less likely that same file will be in contention by multiple PRs.
Developers with [OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder) will have an easier time with Mocha maintenance (in this file anyway).

### Possible Drawbacks
Previous commit history changes more difficult to track.

### Applicable issues
semver-patch


plroebuck:
Noted this:
```js
  if (process.platform !== 'win32') {
    // Windows: Feature works but SIMULATING the signal (ctr+c), via child process, does not work
    // due to lack of *nix signal compliance.
    describe('--watch', function() {
      describe('with watch enabled', function() {
        it('should show the cursor and signal correct exit code, when watch process is terminated', function(done) {
          // ...
        });
      });
    });
  }
```
Think this should be instead, documenting the skipped platform-specific test...

```js
  describe('--watch', function() {
    // Windows: Feature works but SIMULATING the signal (ctr+c), via child process,
    // does not work due to lack of POSIX signal compliance.
    (process.platform !== 'win32' ? describe : describe.skip)('with watch enabled', function() {
      it('should show the cursor and signal correct exit code, when watch process is terminated', function(done) {
        // ...
      });
    });
  });
```

boneskull:
> Think this should be instead, documenting the skipped platform-specific test...

I'd even prefer a `this.skip()` if it's windows in a before-all hook.  IIRC skipping at the suite level completely hides it from reporter output; using `this.skip()` would cause the individual tests to all become pending, and reported as such

boneskull:
@plroebuck 	I'll let you fix the conflict

plroebuck:
@boneskull care to look this over? Plan to merge it sometime tomorrow unless I hear objections...

boneskull:
LGTM.

boneskull:
ty!!

Issue #3676, opened 2019-01-17
createInvalidArgument function, other renames


boneskull:
In #3666 we're tossing an error because `--compilers` can no longer be used.  But we're using [`createInvalidArgumentValueError`](https://github.com/mochajs/mocha/blob/414bc1e184358cac886a3649fa2a610b7cb35cd8/lib/errors.js#L108), which doesn't seem to be appropriate based on that function's documentation.

It looks like what we want is a `createInvalidArgumentError`, where the argument itself is disallowed (I'm going to create a PR for this imminently).

Somewhat related--this probably wants to be a separate PR--I'm thinking in these function names, "argument" should be replaced with "option", because what we're really dealing with are options.  They can *be* command-line arguments, but command-line arguments are just a means to configure options.  The same error should be thrown whether `--compilers` was used on the command-line, or `.mocharc.json` has a `compilers` option.

cc @vkarpov15

boneskull:
closed by 186ca36.

boneskull:
any other related refactors should be handled in a diff issue and PR

Issue #3564, opened 2018-11-11
soft-deprecate --reporter-options and replace with --reporter-option


boneskull:
This assumes #3556 lands.  This issue came out of code review there.

Yargs has great support for "array"-type options.  Generally, any option can be specified multiple times.  In addition, options can accept multiple, *space-delimited* values, e.g., `--foo bar baz quux`.

This issue has two parts:

1.  Soft-deprecate (using `utils.deprecate()`) `--reporter-options`
2.  Implement `--reporter-option` which can be repeated multiple times.

It'd be cool to be able to specify a minimum of one value and a maximum of two, e.g.:

```
--reporter-option someBooleanFlag --reporter-option someString stringValue
```

Not sure if Yargs can to that, but if it can't, then require a single value and split the result on `=`:

```
--reporter-option someBooleanFlag --reporter-option someString=stringValue
```

plroebuck:
See discussion from #3487. @johanblumenberg provided the PR's implementation.

Would strongly prefer that reporter name be prefixed to key in `--reporter-option key=value`.
For example, if specifying "xunit" reporter output file, then:
   `--reporter-option xunit.output=mocha-xunit.xml`

We should decide how to address [these potential problems](https://github.com/mochajs/mocha/pull/3487#issuecomment-427459008) first.

boneskull:
I'm closing this; I've since gotten some backchannel feedback that comma-sep is useful, and it's really not much trouble to support both styles.  as of v6, `--reporter-option` and `--reporter-options` are the same thing, and can be described using commas (`--reporter-option foo=bar,baz=quux`) or repetition (`--reporter-option foo=bar --reporter-option baz=quux`)

plroebuck:
Wait. I don't care whether `--reporterOptions` is deprecated or not.

But need `--reporterOption` to be able to process comma-separated arguments as the value
(e.g., `--reporterOption meta=foo,bar,baz`) which _cannot_ currently be done if using same code.

PR #3677, opened 2019-01-17
add createInvalidArgumentError(); see #3676

boneskull:
- updated error code table in docs
- all other changes in docs are from prettier
- updated `--compilers` test

@plroebuck, please take some tylenol before reviewing `docs/index.md` :wink:


boneskull:
> Why changed from more specific createNotSupportedError?

In the docs this is "Creates an error object used when the type of output specified was not supported".  It seems to be a reporter thing; only used in `xunit` atm.

plroebuck:
> > Why changed from more specific createNotSupportedError?
> 
> In the docs this is "Creates an error object used when the type of output specified was not supported". It seems to be a reporter thing; only used in `xunit` atm.

The docs are wrong (but the PR was still in work). It's supposed to be for _any_ unsupported behavior and/or argument (catch-all).

docs/index.md
```diff
@@ -223,17 +224,17 @@ $ ./node_modules/.bin/mocha mocha.test.js
 
 ## Assertions
 
-Mocha allows you to use any assertion library you wish.  In the above example, we're using Node.js' built-in [assert](https://nodejs.org/api/assert.html) module--but generally, if it throws an `Error`, it will work! This means you can use libraries such as:
+Mocha allows you to use any assertion library you wish. In the above example, we're using Node.js' built-in [assert](https://nodejs.org/api/assert.html) module--but generally, if it throws an `Error`, it will work! This means you can use libraries such as:
```
plroebuck: ...if it throws an `AssertionError`,

docs/index.md
```diff
@@ -223,17 +224,17 @@ $ ./node_modules/.bin/mocha mocha.test.js
...
 
 - [should.js](https://github.com/shouldjs/should.js) - BDD style shown throughout these docs
```
plroebuck: We're using "should.js" in the docs?

docs/index.md
```diff
@@ -268,36 +269,34 @@ Alternately, instead of using the `done()` callback, you may return a [Promise](
...
-    });
+  return db.clear().then(function() {
+    return db.save([tobi, loki, jane]);
+  });
 });
 
 describe('#find()', function() {
   it('respond with matching records', function() {
-    return db.find({ type: 'User' }).should.eventually.have.length(3);
+    return db.find({type: 'User'}).should.eventually.have.length(3);
```
plroebuck: is the `return` needed?

docs/index.md
```diff
@@ -268,36 +269,34 @@ Alternately, instead of using the `done()` callback, you may return a [Promise](
...
-In Mocha v3.0.0 and newer, returning a `Promise` *and* calling `done()` will result in an exception, as this is generally a mistake:
+In Mocha v3.0.0 and newer, returning a `Promise` _and_ calling `done()` will result in an exception, as this is generally a mistake:
 
 ```js
 
-it('should complete this test', function (done) {
-  return new Promise(function (resolve) {
+it('should complete this test', function(done) {
+  return new Promise(function(resolve) {
```
plroebuck: need `return`?

docs/index.md
```diff
@@ -268,36 +269,34 @@ Alternately, instead of using the `done()` callback, you may return a [Promise](
...
 
-In Mocha v3.0.0 and newer, returning a `Promise` *and* calling `done()` will result in an exception, as this is generally a mistake:
+In Mocha v3.0.0 and newer, returning a `Promise` _and_ calling `done()` will result in an exception, as this is generally a mistake:
 
 ```js
 
-it('should complete this test', function (done) {
-  return new Promise(function (resolve) {
+it('should complete this test', function(done) {
```
plroebuck: Add comment above this line denoting anti-pattern.

docs/index.md
```diff
@@ -497,7 +495,7 @@ describe('Array', function() {
 });
 ```
 
-Previous to v3.0.0, `.only()` used string matching to decide which tests to execute.  As of v3.0.0, this is no longer the case.  In v3.0.0 or newer, `.only()` can be used multiple times to define a subset of tests to run:
+Previous to v3.0.0, `.only()` used string matching to decide which tests to execute. As of v3.0.0, this is no longer the case. In v3.0.0 or newer, `.only()` can be used multiple times to define a subset of tests to run:
```
plroebuck: Previous to v3.0.0, `.only()` used string matching to decide which tests to execute, but this is no longer the case.

plroebuck:
>     * all other changes in docs are from prettier
> 
> @plroebuck, please take some tylenol before reviewing `docs/index.md` 😉

Because the documentation changes in "docs/index.md" seemingly have _nothing_ to do with this PR and should have been in one named "Make docs even more pretty"?

boneskull:
just like js, json, and yaml files, staged changes now get the prettier treatment. feel free to send a PR that reformers everything at once.

docs/index.md
boneskull: it’s not necessary to throw an `AssertionError` from a failed assertion.

docs/index.md
boneskull: I think this is a list of popular assertion libs?

docs/index.md
boneskull: yes. it’s a promise.

docs/index.md
```diff
@@ -223,17 +224,17 @@ $ ./node_modules/.bin/mocha mocha.test.js
...
 - [chai](http://chaijs.com/) - `expect()`, `assert()` and `should`-style assertions
-- [better-assert](https://github.com/visionmedia/better-assert) -  C-style self-documenting `assert()`
+- [better-assert](https://github.com/visionmedia/better-assert) - C-style self-documenting `assert()`
 - [unexpected](http://unexpected.js.org) - "the extensible BDD assertion toolkit"
 
 ## Asynchronous Code
 
-Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test. This callback accepts both an `Error` instance (or subclass thereof) *or* a falsy value; anything else will cause a failed test.
+Testing asynchronous code with Mocha could not be simpler! Simply invoke the callback when your test is complete. By adding a callback (usually named `done`) to `it()`, Mocha will know that it should wait for this function to be called to complete the test. This callback accepts both an `Error` instance (or subclass thereof) _or_ a falsy value; anything else will cause a failed test.
```
plroebuck: > To make things even easier, the `done()` callback also accepts an `Error` instance (i.e. `new Error()`), so we may use this directly:

Maybe reword as:
```md
Alternatively, just use the `done()` callback directly.
```

plroebuck:
> just like js, json, and yaml files, staged changes now get the prettier treatment. feel free to send a PR that reformats everything at once.

Or maybe pull your PR's MD changes, prettify "index.md" and commit its updates separately, _then_ reapply your changes. There's no way to follow what's going on with automated changes being applied _at the same time_ as actual updates...

boneskull:
> Or maybe pull your PR's MD changes, prettify "index.md" and commit its updates separately, then reapply your changes. There's no way to follow what's going on with automated changes being applied at the same time as actual updates...

Our pre-commit hooks are _supposed_ to lint staged files.  It's a tradeoff; one I believe to be worth it.  It's by no means perfect--linting and formatting tools work on entire files, not staged hunks.

Yes, that's imperfect, but it reduces friction for contributors who would otherwise be asked to lint stuff in PRs, or have to do so manually.  I _strongly believe_ that any potential murkiness in a Git history created by these tools is less problematic (ultimately, less time-consuming) than requiring manual intervention or extra back-and-forth on PRs.

lib/cli/run.js
```diff
@@ -268,9 +269,10 @@ exports.builder = yargs =>
       }
 
       if (argv.compilers) {
-        throw createInvalidArgumentValueError(
+        throw createUnsupportedError(
           `--compilers is DEPRECATED and no longer supported.
-          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`
+          See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`,
```
plroebuck: Maybe move the URL into a const variable, then reference that...

lib/cli/run.js
```diff
@@ -268,9 +269,10 @@ exports.builder = yargs =>
...
+          'compilers'
```
plroebuck: Delete this line

lib/errors.js
```diff
@@ -7,7 +7,7 @@
  */
 
 /**
- * Creates an error object used when no files to be tested could be found using specified pattern.
+ * Creates an error object to be thrown when no files to be tested could be found using specified pattern.
```
plroebuck: maybe better stated as "to be used" given `done(err)` or `reject(err)`.

docs/index.md
plroebuck: Maybe just misread the description "BDD style shown throughout these docs".

docs/index.md
plroebuck: Perhaps, but then diffs may not occur. All the packages listed fire `AssertionError`...

docs/index.md
```diff
@@ -902,41 +900,41 @@ Mocha Resources
...
 
 ### `--compilers`
 
-> *`--compilers` was removed in v6.0.0. See [further explanation and workarounds](https://github.com/mochajs/mocha/wiki/compilers-deprecation).*
+> _`--compilers` was removed in v6.0.0. See [further explanation and workarounds](https://github.com/mochajs/mocha/wiki/compilers-deprecation)._
 
 ### `--exit`
 
-> *Updated in v4.0.0.*
+> _Updated in v4.0.0._
```
plroebuck: delete this line?

docs/index.md
```diff
@@ -902,41 +900,41 @@ Mocha Resources
...
+_Prior to_ version v4.0.0, _by default_, Mocha would force its own process to exit once it was finished executing all tests. This behavior enables a set of potential problems; it's indicative of tests (or fixtures, harnesses, code under test, etc.) which don't clean up after themselves properly. Ultimately, "dirty" tests can (but not always) lead to _false positive_ or _false negative_ results.
 
-"Hanging" most often manifests itself if a server is still listening on a port, or a socket is still open, etc.  It can also be something like a runaway `setInterval()`, or even an errant `Promise` that never fulfilled.
+"Hanging" most often manifests itself if a server is still listening on a port, or a socket is still open, etc. It can also be something like a runaway `setInterval()`, or even an errant `Promise` that never fulfilled.
 
-The *default behavior* in v4.0.0 is `--no-exit`, where previously it was `--exit`.
+The _default behavior_ in v4.0.0 is `--no-exit`, where previously it was `--exit`.
```
plroebuck:
```md
The current _default behavior_ is `--no-exit`, where previously it was `--exit`.
```

docs/index.md
```diff
@@ -902,41 +900,41 @@ Mocha Resources
...
 
-**The easiest way to "fix" the issue is to simply pass `--exit` to the Mocha process.**  It *can* be time-consuming to debug--because it's not always obvious where the problem is--but it *is* recommended to do so.
+**The easiest way to "fix" the issue is to simply pass `--exit` to the Mocha process.** It _can_ be time-consuming to debug--because it's not always obvious where the problem is--but it _is_ recommended to do so.
```
plroebuck:
```md
**The easiest way to "ignore" the issue is to simply pass `--exit` to the Mocha process.** It _can_ be time-consuming to debug, because it's not always obvious where the problem is -- but it _is_ recommended to do so.
```

docs/index.md
```diff
@@ -1043,7 +1041,7 @@ Allows use of third-party reporters. For example, [mocha-lcov-reporter](https://
 
 ### `--reporter-option <option>, -O <option>, --reporter-options <option>`
 
-> *Updated in v6.0.0.  Can be specified multiple times.  `--reporter-options` is now an alias for `--reporter-option`.*
+> _Updated in v6.0.0. Can be specified multiple times. `--reporter-options` is now an alias for `--reporter-option`._
```
plroebuck: While the options are very related, this is **incorrect**.
`--reporter-option` is _only_ for a _single_ option; `--reporter-options` handles a comma-separated list of options.

docs/index.md
```diff
@@ -1215,7 +1210,7 @@ V8 flags can be defined in Mocha's [configuration](#configuring-mocha-nodejs).
 
 ## Interfaces
 
-Mocha's "interface" system allows developers to choose their style of DSL.  Mocha has **BDD**, **TDD**, **Exports**, **QUnit** and **Require**-style interfaces.
+Mocha's "interface" system allows developers to choose their style of DSL. Mocha has **BDD**, **TDD**, **Exports**, **QUnit** and **Require**-style interfaces.
```
plroebuck: Probably be better to use unordered list format here

docs/index.md
```diff
@@ -1540,13 +1537,13 @@ mocha.setup({
 
 ### Browser-specific Option(s)
 
-The following option(s) *only* function in a browser context:
+The following option(s) _only_ function in a browser context:
 
 `noHighlighting`: If set to `true`, do not attempt to use syntax highlighting on output test code.
```
plroebuck: This still exists? Thought the named negative logic functions were replaced...
```js
mocha.noHighlighting(true);
```
vs.
```js
mocha.highlighting(false);
```

docs/index.md
```diff
@@ -1724,21 +1721,24 @@ Some shells support recursive matching by using the `**` wildcard in a glob. Bas
...
+| Code                             | Description                                                  |
+| -------------------------------- | ------------------------------------------------------------ |
+| ERR_MOCHA_INVALID_ARG            | argument is invalid                                          |
+| ERR_MOCHA_INVALID_ARG_TYPE       | argument of the wrong type was passed to Mocha's API         |
+| ERR_MOCHA_INVALID_ARG_VALUE      | invalid or unsupported value was passed for a given argument |
+| ERR_MOCHA_INVALID_INTERFACE      | interface specified in options not found                     |
+| ERR_MOCHA_INVALID_REPORTER       | reporter specified in options not found                      |
+| ERR_MOCHA_NO_FILES_MATCH_PATTERN | file/s of test could not be found                            |
+| ERR_MOCHA_UNSUPPORTED            | requested behavior, option, or parameter is unsupported      |
+| ERR_MOCHA_INVALID_EXCEPTION      | a falsy or otherwise underspecified exception was thrown     |
```
plroebuck: out of order

docs/index.md
```diff
@@ -1724,21 +1721,24 @@ Some shells support recursive matching by using the `**` wildcard in a glob. Bas
...
-| ERR_MOCHA_INVALID_ARG_TYPE | argument of the wrong type was passed to Mocha's API |
-| ERR_MOCHA_INVALID_ARG_VALUE | invalid or unsupported value was passed for a given argument |
-| ERR_MOCHA_INVALID_INTERFACE | interface specified in options not found |
-| ERR_MOCHA_INVALID_REPORTER | reporter specified in options not found |
-| ERR_MOCHA_NO_FILES_MATCH_PATTERN | file/s of test could not be found |
-| ERR_MOCHA_NOT_SUPPORTED | type of output specified was not supported |
-| ERR_MOCHA_UNDEFINED_ERROR | an error was thrown but no details were specified |
+| Code                             | Description                                                  |
+| -------------------------------- | ------------------------------------------------------------ |
+| ERR_MOCHA_INVALID_ARG            | argument is invalid                                          |
```
plroebuck: remove

docs/index.md
```diff
@@ -1724,21 +1721,24 @@ Some shells support recursive matching by using the `**` wildcard in a glob. Bas
 $ mocha "./spec/**/*.js"
 ```
 
-*Note*: Double quotes around the glob are recommended for portability.
+_Note_: Double quotes around the glob are recommended for portability.
```
plroebuck: **Wrong!** For portability, single quotes should be used, not double...

docs/index.md
```diff
@@ -1617,20 +1614,20 @@ tests as shown below:
...
 
-> *New in v6.0.0*
+> _New in v6.0.0_
 
 In addition to supporting the legacy [`mocha.opts`](#mochaopts) run-control format, Mocha now supports configuration files, typical of modern command-line tools, in several formats:
 
 - **JavaScript**: Create a `.mocharc.js` in your project's root directory, and export an object (`module.exports = {/* ... */}`) containing your configuration.
 - **YAML**: Create a `.mocharc.yaml` (or `.mocharc.yml`) in your project's root directory.
-- **JSON**: Create a `.mocharc.json` in your project's root directory.  Comments--while not valid JSON--are allowed in this file, and will be ignored by Mocha.
+- **JSON**: Create a `.mocharc.json` in your project's root directory. Comments--while not valid JSON--are allowed in this file, and will be ignored by Mocha.
```
plroebuck: **JSON**: Create a `.mocharc.json` in your project's root directory. Comments -- while not valid JSON -- are allowed in this file, and will be ignored by Mocha.

docs/index.md
```diff
@@ -1596,16 +1593,16 @@ tests as shown below:
 <html>
   <head>
```
plroebuck: <meta charset="utf-8" />

docs/index.md
```diff
@@ -1596,16 +1593,16 @@ tests as shown below:
...
     <title>Mocha</title>
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
```
plroebuck: Don't think we need either of these meta tags

docs/index.md
```diff
@@ -1596,16 +1593,16 @@ tests as shown below:
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="stylesheet" href="https://unpkg.com/mocha/mocha.css" />
   </head>
   <body>
     <div id="mocha"></div>
```
plroebuck: <script src="https://unpkg.com/chai/chai.js"></script>

docs/index.md
boneskull: I didn't replace nor rename functions, no.  we can do that next major.

docs/index.md
boneskull: no, that doesn't work on windows.  cmd or powershell, I forget.

docs/index.md
boneskull: I'm not sure what you're reviewing here, but it's not this PR...

docs/index.md
boneskull: I'm not sure where you're getting that information.  `reporter-options` is an alias of `reporter-option`.

plroebuck: It was meant to be the replacement of `--reporter-options`, **never** an alias.

https://github.com/mochajs/mocha/pull/3556#discussion_r231503173
#3564

docs/index.md
plroebuck: Better stated, I thought you had added the renamed versions (retaining the negative-logic originals, but deprecating them).

docs/index.md
plroebuck: We should find out then. Leaving this up to each individual shell (given the varying levels of glob support) is _definitely_ not portable.

For UNIX, single quotes will be the most portable since they will prevent the shell from doing the expansion. For _maximum_ portability and reproducability, Mocha **must** handle the globs itself.

docs/index.md
```diff
@@ -977,31 +975,31 @@ Mocha does not retry test failures by default.
 
 ### `--slow <ms>, -s <ms>`
 
-Specify the "slow" test threshold in milliseconds.  Mocha uses this to highlight test cases that are taking too long.  "Slow" tests are not considered failures.
+Specify the "slow" test threshold in milliseconds. Mocha uses this to highlight test cases that are taking too long. "Slow" tests are not considered failures.
 
-Note: A test that executes for *half* of the "slow" time will be highlighted *in yellow* with the default `spec` reporter; a test that executes for entire "slow" time will be highlighted *in red*.
+Note: A test that executes for _half_ of the "slow" time will be highlighted _in yellow_ with the default `spec` reporter; a test that executes for entire "slow" time will be highlighted _in red_.
 
 ### `--timeout <ms>, -t <ms>`
```
plroebuck: BTW, `--timeouts` has been added as well. No idea why though since it's a single value not capable of multiple values.

docs/index.md
```diff
@@ -953,21 +951,21 @@ Enforce a rule that tests may not be exclusive (use of e.g., `describe.only()` o
...
 ### `--global <variable-name>`
 
-> *Updated in v6.0.0; the option is `--global` and `--globals` is now an alias.*
+> _Updated in v6.0.0; the option is `--global` and `--globals` is now an alias._
 
 Define a global variable name. For example, suppose your app deliberately exposes a global named `app` and `YUI`, you may want to add `--global app --global YUI`.
 
 `--global` accepts wildcards. You could do `--global '*bar'` and it would match `foobar`, `barbar`, etc. You can also simply pass in `'*'` to ignore all globals.
 
 `--global` can accept a comma-delimited list; `--global app,YUI` is equivalent to `--global app --global YUI`.
```
plroebuck: IMO, `--globals` is either unnecessary or should process differently.
I could see `--globals` handling the comma-separated list, and `--global` handling single value only.

docs/index.md
```diff
@@ -1053,55 +1051,55 @@ Can be specified as a comma-delimited list.
...
 
 By default, Mocha looks for a `mocha.opts` in `test/mocha.opts`; use `--no-opts` to suppress this behavior.
 
 ### `--package <path>`
 
-> *New in v6.0.0.*
+> _New in v6.0.0._
 
-Specify an explicit path to a [`package.json` file](#configuring-mocha-node-js)  (ostensibly containing configuration in a `mocha` property).
+Specify an explicit path to a [`package.json` file](#configuring-mocha-node-js) (ostensibly containing configuration in a `mocha` property).
```
plroebuck: From `mocha --help`:
>  --opts     Path to `mocha.opts`        [string] [default: "./test/mocha.opts"]
>   --package  Path to package.json for config                            [string]

"package.json" should be quoted like the preceding "mocha.opts"

lib/runnable.js
```diff
@@ -4,6 +4,8 @@ var Pending = require('./pending');
 var debug = require('debug')('mocha:runnable');
 var milliseconds = require('ms');
 var utils = require('./utils');
+var createInvalidExceptionError = require('./errors')
+  .createInvalidExceptionError;
```
plroebuck: This is beyond stupid formatting.
```js
var errors = require('./errors')
var createInvalidExceptionError = errors.createInvalidExceptionError;
```

lib/runner.js
```diff
@@ -16,7 +16,8 @@ var createStatsCollector = require('./stats-collector');
 var stackFilter = utils.stackTraceFilter();
 var stringify = utils.stringify;
 var type = utils.type;
-var undefinedError = utils.undefinedError;
+var createInvalidExceptionError = require('./errors')
+  .createInvalidExceptionError;
```
plroebuck: > This is beyond stupid formatting.
> 
> ```js
> var errors = require('./errors');
> var createInvalidExceptionError = errors.createInvalidExceptionError;
> ```

lib/runnable.js
```diff
@@ -450,3 +452,20 @@ Runnable.prototype._timeoutError = function(ms) {
...
 };
+
+/**
+ * Generate an undefined error if `err` is not defined.
+ *
+ * @param {*} [err] - Hopefully an Error
+ * @return {Error}
+ */
+
+Runnable.getError = function(err) {
```
plroebuck: This doesn't appear to be used anywhere but this file. Why export it? Seems to unnecessarily complicate calling it (L358,L381) for no benefit.

lib/runnable.js
```diff
@@ -450,3 +452,20 @@ Runnable.prototype._timeoutError = function(ms) {
   }
   return new Error(msg);
 };
+
+/**
+ * Generate an undefined error if `err` is not defined.
+ *
```
plroebuck:
```js
 * @static
```

docs/index.md
boneskull: `--timeouts` is the opposite of `--no-timeouts`

docs/index.md
boneskull: I am unsure why the behavior needs to be different, but it's also not relevant to this PR

lib/runnable.js
boneskull: it's implied; `@static` is necessary if it's not obviously static to jsdoc (which it is)

docs/index.md
boneskull: ah, right.  https://github.com/mochajs/mocha/issues/3564#issuecomment-458740928

docs/index.md
boneskull: except that our unordered list inexplicably has two columns, making it gross

docs/index.md
boneskull: mm, I think I didn't address this because it's browser-specific; not really on the radar for CLI options.  but yeah, it needs to be addressed.  I'll make an issue.

docs/index.md
boneskull: Mocha generates its own diffs unless the reporter tells Mocha not do to that...

docs/index.md
boneskull: please send a PR for further changes to this doc that are tangential to this PR

docs/index.md
plroebuck: Logically, sure... don't see your point.

Question was why was `--timeouts` added at all, given that we already have `--timeout` which was handling the task just fine. And the existing matches preconceived expectation -- singular name accepting singular value.

docs/index.md
```diff
@@ -1809,21 +1809,22 @@ $ REPORTER=nyan npm test
...
-spies, mocking, and shared behaviours be sure to check out the [Mocha Wiki][Mocha-wiki] on GitHub.
-For discussions join the [Google Group][Google-mocha]. For a running example of Mocha, view
+In addition to chatting with us on [Gitter][gitter-mocha], for additional information such as using
+spies, mocking, and shared behaviours be sure to check out the [Mocha Wiki][mocha-wiki] on GitHub.
+For discussions join the [Google Group][google-mocha]. For a running example of Mocha, view
 [example/tests.html](example/tests.html). For the JavaScript API, view the [API documentation](api/)
 or the [source](https://github.com/mochajs/mocha/blob/master/lib/mocha.js).
 
-[//]: # (Cross reference section)
-
```
plroebuck: This blank line must be put back for portability reasons.
Blank lines before **and** after the comment line are [required for portability](https://stackoverflow.com/questions/4823468/comments-in-markdown/20885980#20885980).

docs/index.md
plroebuck: Medium article "[Why you should always quote your globs in NPM scripts](https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784)".

docs/index.md
boneskull: https://github.com/prettier/prettier/issues/new/choose

docs/index.md
boneskull: please create an issue or fix elsewhere

docs/index.md
plroebuck:
```md
<!-- prettier-ignore -->
```

test/unit/runner.spec.js
```diff
@@ -544,6 +552,9 @@ describe('Runner', function() {
       it('should not hang if overlong error message is single line', function(done) {
         var hook = new Hook();
         hook.parent = suite;
+        var data = [];
+        // mock a long message
+        for (var i = 0; i < 10000; i++) data[i] = {a: 1};
```
plroebuck: These lines should not be present.
`genOverlongSingleLineMessage()` handles this...

test/integration/options.spec.js
```diff
@@ -1,16 +0,0 @@
...
-
-/**
- * Tests for cmdline options.
- * Provides umbrella for all spec files in the "options" subdirectory.
- */
-
-var path = require('path');
-var utils = require('../../lib/utils');
-
-describe('options', function() {
```
plroebuck: This was better with the `describe` grouping the cmdline options...

plroebuck: For consistency, "show-plugins.spec.js" should be broken into two files whose contents should be in the "options" directory...

boneskull: OK, I've split this up.  does it belong in this PR?  likely not, but sometimes it's just easier to do it or it won't get done...

plroebuck: That addressed the "show-plugins.spec.js" break-up. Still don't see why this file had to go...

boneskull: ohhhh OK, I think I zapped the file because I thought it was vestigial and part of a bad merge.

I don't know why, but something about it caused the code coverage numbers to drop.  I made changes `test.node.integration` in `package-scripts.js` elsewhere to address the issue.  I think the problem may be that `options.spec.js` doesn't accurately replicate what `Mocha#loadFiles` does?  Even if we did try to do it programmatically, there's still the challenge of not being able to reach the current `Mocha` instance from the test files themselves. 
 
Even though we lose the wrapper suite, the options tests are still run in alphabetical order w/o this file. 

you're welcome to take another stab at loading things this way, as it'd be nice to have an example we can point users to.

boneskull:
@plroebuck 	here's an example of how the `try/catch` block in `runMochaJSON` is incorrect (I changed it in another PR): https://travis-ci.org/mochajs/mocha/jobs/490660198#L2102

lib/cli/run.js
```diff
@@ -268,7 +269,7 @@ exports.builder = yargs =>
       }
 
       if (argv.compilers) {
-        throw createInvalidArgumentValueError(
+        throw createUnsupportedError(
           `--compilers is DEPRECATED and no longer supported.
           See ${ansi.cyan('https://git.io/vdcSr')} for migration information.`
```
plroebuck: Don't think colorized formatting has any place in thrown error message.

boneskull: I think I agree.

PR #3680, opened 2019-01-19
Tests: Avoid the 'was called with' assertion

papandreou:
### Description of the Change

Avoid the use of the `was called with` assertion from unexpected-sinon. It has been deprecated for two years and will soon display warnings or even be removed.

See discussion here: https://github.com/unexpectedjs/unexpected-sinon/issues/32

### Alternate Designs

Not relevant.

Not applicable, this is just a change to the test suite.

### Benefits

Due diligence so we don't have to deal with this later when upgrading unexpected-sinon.

### Possible Drawbacks

None.

### Applicable issues

None.

### DIFF
--- a/test/node-unit/cli/options.spec.js
+++ b/test/node-unit/cli/options.spec.js
@@ -114,7 +114,7 @@ describe('options', function() {
               try {
                 loadOptions(`--opts ${opts}`);
               } catch (ignored) {}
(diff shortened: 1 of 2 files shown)


boneskull:
LGTM.  coverage failure is unrelated

boneskull:
thanks again

PR #3684, opened 2019-01-23
do not recursively call `process.exit()`

addaleax:
### Description of the Change

When inside `process.on('exit')`, calling `process.exit()`
stops other `exit` event listeners from being invoked.

This has lead to issues with coverage generation with `nyc`
because a fallback exit detection mechanism in it that relies
on internal Node.js APIs was accidentally disabled in Node v11.7.0,
leaving `nyc` with no way to detect process exit.

### Alternate Designs

No alternatives were considered.

### Benefits

`process.on('exit')` handlers are run even if `exitMochaLater` is used. At least `nyc` relies on this behaviour.

### Possible Drawbacks

I guess running extra code while exiting could have drawbacks.

### Applicable issues

This fixes https://github.com/nodejs/node/issues/25650 (which is already closed, but because Node.js is (possibly temporarily) reintroducing monkey-patching for the internal method in question, not because the underlying cause was addressed).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ stages:
 
 # defaults
 language: node_js
-node_js: '11.6'
+node_js: '11'
 addons:
   apt:
     packages:
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -25,7 +25,7 @@ const cwd = (exports.cwd = process.cwd());
  */
 const exitMochaLater = code => {
   process.on('exit', () => {
-    process.exit(Math.min(code, 255));
+    process.exitCode = Math.min(code, 255);
   });
 };
 


addaleax:
I’ve also included a revert that resets the `.travis.yml` version to Node 11’s latest release – since this fixes the issue, that should work (it does locally for me).

boneskull:
I'm going to add a couple extra tests around this.  Our integration tests that hit this behavior (`--no-exit`) may not be taking exit codes into account.

addaleax:
> But now I'm curious how this ever worked in the first place--doesn't the `process.exit()` call cause the listener to be fired recursively?

I don’t know the reasons for this behaviour, but calling `process.exit()` while already exiting exits the process immediately rather than emitting the event again.

addaleax:
@plroebuck I’m not sure I understand the question – this reverts the Travis Node version change, because that was a workaround for the bug in question, and usually, projects want to have the latest versions of each major Node.js release in their Travis matrix?

plroebuck:
My apologies. Read diff backwards accidentally (not realizing it had been modified 2 days earlier without formal PR) as we _had_ been using the latest Node.

boneskull:
yeah I just updated w/o PR it to fix our broken builds.

Issue #3416, opened 2018-06-14
Test failure with large string is very slow.


shunjikonishi:
I have a test that compares very large string.

The case, it succeed, there is no problem. It is fast enough.
However the case, it fails, the test stops for a while. (More than 1min.)

I made a small example. It takes more than 1min on my local.
I guess building error message part is slow.
But I'm not sure why it takes so long time.

### Steps to Reproduce
Run this test.

```
"use strict";

const expect = require("chai").expect;

// Make large string
// If loop count is 10000, the test never finish.
const largeString = (function() {
    let result = "";
    for (let i=0; i<3000; i++) {
        result += "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
    }
    return result;
})();

describe("Very slow test", () => {

    it("compare large string", () => {
        expect("a").to.equal(largeString);
    });
});
```

### Versions
mocha --version = 5.2.0
node --version = v8.8.1

outsideris:
I believe it is chai's issue, not mocha.

Below test is not too long.

```
"use strict";

const assert = require('assert');

// Make large string
// If loop count is 10000, the test never finish.
const largeString = (function() {
    let result = "";
    for (let i=0; i<3000; i++) {
        result += "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
    }
    return result;
})();

describe("Very slow test", () => {

    it("compare large string", () => {
				assert.equal(largeString, "a");
    });
});
```

Please reopen if I'm miss something.

PR #3686, opened 2019-01-24
fix regex in utils.stackTraceFilter to prevent ReDoS #3416

cyjake:
if the stack trace begins with a large error message (>= 20k charactors), and user leaves `--full-trace` disabled, `utils.stackTraceFilter()` takes ages to finish. Large error messages is quite possible when user makes containment assertions such as `expect(content).to.contain(word)`.

### Description of the Change

simplified the regex used in `utils.stackTraceFilter()` to boost performance therefore prevent ReDoS if the stack contains large error message.

### Alternate Designs

It's possible to skip check error message in stack by skipping the first line but the regex fix won't change any behavior hence it is chosen.

well, it's a bug introduced by the stack filtering feature.

### Benefits

mocha won't hang even if users make assertions with large error messages, which is quite possible because containment check is used a lot.

### Possible Drawbacks

None

### Applicable issues

#3416 looks like one but it was because of an external library (still quite possible if chai makes large error message and `assert.deepEqual` don't, didn't check).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -671,8 +671,6 @@ exports.stackTraceFilter = function() {
   function isMochaInternal(line) {
     return (
       ~line.indexOf('node_modules' + slash + 'mocha' + slash) ||
-      ~line.indexOf('node_modules' + slash + 'mocha.js') ||
-      ~line.indexOf('bower_components' + slash + 'mocha.js') ||
       ~line.indexOf(slash + 'mocha.js')
     );
   }
@@ -701,7 +699,7 @@ exports.stackTraceFilter = function() {
       }
 
       // Clean up cwd(absolute)
-      if (/\(?.+:\d+:\d+\)?$/.test(line)) {
+      if (/:\d+:\d+\)?$/.test(line)) {
         line = line.replace('(' + cwd, '(');
       }
 
--- a/test/unit/runner.spec.js
+++ b/test/unit/runner.spec.js
@@ -6,6 +6,7 @@ var Runner = mocha.Runner;
 var Test = mocha.Test;
 var Hook = mocha.Hook;
 var path = require('path');
+var fs = require('fs');
 var noop = mocha.utils.noop;
 
 describe('Runner', function() {
@@ -491,6 +492,62 @@ describe('Runner', function() {
 
         runner.on('fail', function(hook, err) {
           expect(err.stack, 'to be', stack.join('\n'));
+          done();
+        });
+        runner.failHook(hook, err);
+      });
+    });
+
+    describe('hugeStackTrace', function() {
+      beforeEach(function() {
+        if (path.sep !== '/') {
+          this.skip();
+        }
+      });
+
+      it('should not hang if the error message is ridiculously long single line', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
+        var data = [];
+        // mock a long message
+        for (var i = 0; i < 10000; i++) data[i] = {a: 1};
+        var message = JSON.stringify(data);
+        var err = new Error();
+        // Fake stack-trace
+        err.stack = [message].concat(stack).join('\n');
+
+        runner.on('fail', function(hook, err) {
+          expect(
+            err.stack
+              .split('\n')
+              .slice(1)
+              .join('\n'),
+            'to be',
+            stack.slice(0, 3).join('\n')
+          );
+          done();
+        });
+        runner.failHook(hook, err);
+      });
+
+      it('should not hang if error message is ridiculously long multiple lines either', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
+        var fpath = path.join(__dirname, '../../mocha.js');
+        var message = fs.readFileSync(fpath, 'utf8');
+        var err = new Error();
+        // Fake stack-trace
+        err.stack = [message].concat(stack).join('\n');
+
+        runner.on('fail', function(hook, err) {
+          expect(
+            err.stack
+              .split('\n')
+              .slice(-3)
+              .join('\n'),
+            'to be',
+            stack.slice(0, 3).join('\n')
+          );
           done();
         });
         runner.failHook(hook, err);


lib/utils.js
```diff
@@ -647,8 +647,6 @@ exports.stackTraceFilter = function() {
   function isMochaInternal(line) {
     return (
       ~line.indexOf('node_modules' + slash + 'mocha' + slash) ||
-      ~line.indexOf('node_modules' + slash + 'mocha.js') ||
-      ~line.indexOf('bower_components' + slash + 'mocha.js') ||
       ~line.indexOf(slash + 'mocha.js')
```
cyjake: both L650 and L651 are unnecessary because of L652

test/unit/runner.spec.js
```diff
@@ -487,6 +487,32 @@ describe('Runner', function() {
...
+              .split('\n')
+              .slice(1)
+              .join('\n'),
+            'to be',
+            stack.slice(0, 3).join('\n')
+          );
+          done();
+        });
+        runner.failHook(hook, err);
+      });
```
cyjake: `./bin/mocha test/unit/runner.spec.js --grep longStackTrace --timeout 1000` will hang and then timeout if regex were reverted.

boneskull:
will wait for someone else to review, but fine by me if it's merged thereafter

lib/utils.js
```diff
@@ -677,7 +675,7 @@ exports.stackTraceFilter = function() {
       }
 
       // Clean up cwd(absolute)
-      if (/\(?.+:\d+:\d+\)?$/.test(line)) {
+      if (/:\d+:\d+\)?$/.test(line)) {
         line = line.replace('(' + cwd, '(');
```
juergba: What is the purpose of this change?

cyjake: as explained in the description of this pr, the original regex performs poorly if `line` is larger than 20k, which is quite possible if assertions on large string were made. The modified regex checks the same thing without performance issue.

The removed part in the original regex, `/\?.+/`, basically matches any character but it's actually unnecessary because we only to find lines with line and col numbers separated by `:` which is already written out.

plroebuck: Is there a timing test where this speedup is shown?
```sh
$ time mocha test.spec.js
$ time mocha-patched test.spec.js
```

plroebuck: > as explained in the description of this pr, the original regex performs poorly if `line` is larger than 20k,
> which is quite possible if assertions on large string were made. The modified regex checks the same thing 
> without performance issue.

If the `Error.message` is causing the problem, why not just regex test `Error.stacktrace.slice(1)`?

plroebuck: Or tweak the original regex so first line won't match immediately...
```js
      if (/^ \(?.+:\d+:\d+\)?$/.test(line)) {
```

test/unit/runner.spec.js
```diff
@@ -487,6 +487,32 @@ describe('Runner', function() {
...
+      it('should not hang if the error message is ridiculously long', function(done) {
+        var hook = new Hook();
+        var message = [];
+        // mock a long message
+        for (var i = 0; i < 20000; i++) {
+          var line = stack[i % stack.length];
+          message[i] = line[Math.floor(Math.random() * line.length)];
+        }
+        var err = new Error(message);
+        // Fake stack-trace
```
juergba: Error.message or Error.stack, are both causing problems when too large?

lib/utils.js
plroebuck: @boneskull, [your #2503 updates](https://github.com/mochajs/mocha/commit/8bf1f297142b02128664523ab5a9b3d55b09c365#diff-50e3aa130a4f97a42ee2cf111c7b1d9d) don't appear to be equivalent changes for these lines...

The two lines appear to need correction, not removal.

test/unit/runner.spec.js
plroebuck: Think he's saying `Error.message` is causing the problem, but it's the first line of the stacktrace...

lib/utils.js
boneskull: I think you mean `stack`, but it's not that simple: it's a string, and the message may be multi-line.

boneskull: afaik lines displayed after the message don't necessary start with a space, or whitespace at all.  even though Node.js may display stacktraces this way (is it four spaces or a tab??), `stack` is a non-standard property, and implementations differ.

test/unit/runner.spec.js
boneskull: this is not necessarily the case:

```js
throw new Error(`foo
bar
baz
`);
```

lib/utils.js
boneskull: regarding benchmarks, we have none, and I'd be happy to accept a PR with harnesses/scripts/dependencies to make something work... but it doesn't need to be in this PR.

lib/utils.js
boneskull: my changes to these two lines in 8bf1f297142b02128664523ab5a9b3d55b09c365 were incorrect, but they are also extraneous; the lines I replaced were *also* extraneous.  removing them is correct.

lib/utils.js
plroebuck: > I think you mean `stack`, but it's not that simple: it's a string, and the message may be multi-line.

Fair enough. I was just talking basic idea, but now see we only have the stacktrace, not the `Error` itself.

plroebuck: While unspecified, the stacktrace itself _is_ indented in all viable implementations.
 ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) say MS doesn't have `Error.stack`).
```js
      if (/^\s\(?.+:\d+:\d+\)?$/.test(line)) {
```

Just spitballing here...

test/unit/runner.spec.js
cyjake: @juergba `err.stack` is composed of `err.message` and the real stack. Skipping the first line is my first thought but `err.message` could be multiline. My second try is filtering the real stack by checking `/^\s+at ` first but the test cases already constructed (in the file below) is far complicated than I thought. 

Removing `/^\(?.+/` is the ideal fix AFAICT because it fixes the problem and won't introduce regression.

lib/utils.js
```diff
@@ -647,8 +647,6 @@ exports.stackTraceFilter = function() {
   function isMochaInternal(line) {
     return (
       ~line.indexOf('node_modules' + slash + 'mocha' + slash) ||
-      ~line.indexOf('node_modules' + slash + 'mocha.js') ||
```
outsideris: Why are these lines removed? As git history, they are related with [windows issues](https://github.com/mochajs/mocha/commit/8bf1f297142b02128664523ab5a9b3d55b09c365).

lib/utils.js
cyjake: @outsideris L650~651 were removed because L652 already has them covered. The windows issues were caused by not using `path.seq`, which was fixed by assigning `slash` to `path.sep` instead of `/`.

juergba:
@cyjake I feel uncomfortable with this `err.stack` stuff, since I still don't understand it clearly. My platform is windows.
- I run your test case and your "end-to-end benchmark" example, both work with and without your changes. I didn't check the time though, but no obvious difference.
- `Error.prototype.message` is a standard property which contains a brief description of the error (definition of Mozilla.org). Do assertion errors (ab)use this property to log huge diffs? And when getting too large we have those hangers?
- why not pass `err.message` as second parameter to `stackTraceFilter()` and omit that regex?

test/unit/runner.spec.js
```diff
@@ -487,6 +487,32 @@ describe('Runner', function() {
...
       });
+
+      it('should not hang if the error message is ridiculously long', function(done) {
+        var hook = new Hook();
+        var message = [];
+        // mock a long message
+        for (var i = 0; i < 20000; i++) {
+          var line = stack[i % stack.length];
+          message[i] = line[Math.floor(Math.random() * line.length)];
+        }
```
plroebuck: OK. Think I'm good with the regex change to code in "lib/utils.js". But uncomfortable with your test here, which seems at odds with your issue. Why doesn't it just create a multi-line `AssertionError.message` and replace `stack[0]`? The `for` loop code seems... excessive-? to that task, and its contents are non-reproducible.

test/unit/runner.spec.js
```diff
@@ -487,6 +487,32 @@ describe('Runner', function() {
         });
```
plroebuck: On L450 and L471, can we change the skip condition to be more apparent?
```js
 if (path.sep !== '/') {
```
would become
```js
 if (process.platform === 'win32') {
```

test/unit/runner.spec.js
cyjake: @plroebuck ah, good catch. There actually are two factors that might slow down the regex test, a long single line (>= 40k takes ~1600ms), or a large message with multiple lines. The test case I constructed was trying to mimic the first case but it was done incorrectly. I've pushed the fix.

The `for` loop is the simplest way I can come up with to create a long and non-repeating text. Using `fs.readFileSync(path.join(__dirname, '../mocha.js'), 'utf8')` suffices the task but I'm not sure this unit test is node.js only or not.

cyjake: ha, turns out I wasn't very wrong. 

```bash
> a = [1,2]
[ 1, 2 ]
> [a]
[ [ 1, 2 ] ]
> [a].concat(3).join('\n')
'1,2\n3'
```

`[message].concat(stack).join(\n')` just yields a large `err.message` with lots of `,`. Anyway, I've updated the test code.

lib/utils.js
boneskull: they are extraneous. the next line catches them:

```js
(~line.indexOf(slash + 'mocha.js'))
```

the windows-specific changes have to do with the value of `slash`.

lib/utils.js
boneskull: @plroebuck 	please add a comment about what you changed when editing comments/descriptions

test/unit/runner.spec.js
boneskull: yes, but it doesn't need to happen in this PR.

cyjake:
@juergba Sorry for not catching your comment. Here's my two cents. 

- You might've got a computer faster than mine. It should fail if you bump the length if the `for` loop of the mock function. Run the test with shorter timeout shall suffice too, such as `./bin/mocha.js test/unit/runner.spec.js --timeout 500`. 
- Yes, `AssertionError` use `err.message` to display failure message which might be (accidentally) very large because `expect(content).to.contain(string)` is quite common, at least in my projects.
- I did consider the approach of separating `err.message` from `err.stack`, but the changes involved is disproportionate compared to this one. The regex needs fix anyway, `/\(?.+/` simply matches any character which makes no point.

lib/utils.js
cyjake: @boneskull The diff can be found by clicking the `edited by ...` link. He fixed a typo in my comment, gracias 😆

test/unit/runner.spec.js
```diff
@@ -487,6 +487,32 @@ describe('Runner', function() {
...
         runner.failHook(hook, err);
       });
+
+      it('should not hang if the error message is ridiculously long', function(done) {
```
plroebuck:
```js
it('should not hang if the error message is ridiculously long single line',
```

test/unit/runner.spec.js
plroebuck: Could you add a test case for a large multi-line message to cover _both_ triggers?

test/unit/runner.spec.js
cyjake: Well, `err.message` with multiple lines can also cause mocha to hang. What's the point with this wording?

test/unit/runner.spec.js
cyjake: ah, guess I missed this one. I'm on it.

test/unit/runner.spec.js
```diff
@@ -496,6 +496,62 @@ describe('Runner', function() {
...
+          done();
+        });
+        runner.failHook(hook, err);
+      });
+
+      it('should not hang if error message is ridiculously long multiple lines either', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
+        var fpath = require('path').join(__dirname, '../../mocha.js');
+        var message = require('fs').readFileSync(fpath, 'utf8');
```
cyjake: I've failed to mock any long but concise text to re-produce regex block. Using `mocha.js` is the easiest I can come up with. @plroebuck

test/unit/runner.spec.js
```diff
@@ -496,6 +496,62 @@ describe('Runner', function() {
...
+        }
+      });
+
+      it('should not hang if the error message is ridiculously long single line', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
+        var data = [];
+        // mock a long message
+        for (var i = 0; i < 10000; i++) data[i] = {a: 1};
+        var message = JSON.stringify(data);
```
cyjake: Stolen from a recently reported issue, which is much simpler.

test/unit/runner.spec.js
cyjake: Sorry for my misunderstanding, I've pushed specific test case for multilines.

test/unit/runner.spec.js
```diff
@@ -496,6 +496,62 @@ describe('Runner', function() {
...
+          );
+          done();
+        });
+        runner.failHook(hook, err);
+      });
+
+      it('should not hang if error message is ridiculously long multiple lines either', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
+        var fpath = require('path').join(__dirname, '../../mocha.js');
```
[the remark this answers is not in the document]
cyjake: https://nodejs.org/api/path.html#path_path_sep path actually accepts both back and forward slash, manually separating them into parameters is unnecessary.

test/unit/runner.spec.js
cyjake: updated both `fs` and `path`

test/unit/runner.spec.js
```diff
@@ -496,6 +497,62 @@ describe('Runner', function() {
...
+        err.stack = [message].concat(stack).join('\n');
+
+        runner.on('fail', function(hook, err) {
+          expect(
+            err.stack
+              .split('\n')
+              .slice(1)
+              .join('\n'),
+            'to be',
+            stack.slice(0, 3).join('\n')
```
plroebuck: assertion is comparing `err.stack` sans `message` against the first three lines of `stack`?

test/unit/runner.spec.js
```diff
@@ -496,6 +497,62 @@ describe('Runner', function() {
...
+          );
+          done();
+        });
+        runner.failHook(hook, err);
+      });
+
+      it('should not hang if error message is ridiculously long multiple lines either', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
```
plroebuck:
```js
function genMultiLineMessage() {
    var fpath = path.join(__dirname, '..', '..', 'mocha.js');
    return fs.readFileSync(fpath, 'utf8');
}
var message = genMultiLineMessage();
```

test/unit/runner.spec.js
```diff
@@ -496,6 +497,62 @@ describe('Runner', function() {
...
+        }
+      });
+
+      it('should not hang if the error message is ridiculously long single line', function(done) {
+        var hook = new Hook();
+        hook.parent = suite;
+        var data = [];
+        // mock a long message
+        for (var i = 0; i < 10000; i++) data[i] = {a: 1};
+        var message = JSON.stringify(data);
```
plroebuck:
```js
function genSingleLineMessage() {
  var data = [];
  for (var i = 0; i < 10000; i++) data[i] = {a: 1};
  return JSON.stringify(data);
}
var message = genSingleLineMessage();
```

test/unit/runner.spec.js
cyjake: yes, because the reset of `stack` should be filtered out.

test/unit/runner.spec.js
cyjake: I think the test case title is self explanatory enough. The test case body isn't that long that needs this level of refactor. Can I pass on this one?

test/unit/runner.spec.js
```diff
@@ -496,6 +497,62 @@ describe('Runner', function() {
...
+        var err = new Error();
+        // Fake stack-trace
+        err.stack = [message].concat(stack).join('\n');
+
+        runner.on('fail', function(hook, err) {
```
plroebuck: How about this to make it more clear:
```js
          var prettyErrStack = err.stack.split('\n').slice(1)
          expect(
            prettyErrStack.join('\n'),
            'to be',
            stack.slice(0, 3).join('\n')
          );
```

test/unit/runner.spec.js
plroebuck: Each function would live outside the corresponding `it()` replacing the _how_ with _what_ was generated. Prefer self-documenting code (especially considering the number of questions this small PR has already brought up). But whatever...

test/unit/runner.spec.js
```diff
@@ -496,6 +497,62 @@ describe('Runner', function() {
...
+          expect(
+            err.stack
+              .split('\n')
+              .slice(-3)
+              .join('\n'),
+            'to be',
+            stack.slice(0, 3).join('\n')
```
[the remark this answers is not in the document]
plroebuck: See `prettyErrStack` comment above.

test/unit/runner.spec.js
cyjake: Let's just discuss in this thread. I believe the changes you proposed are all quite similar. Extracting three lines into a function, breaking one line into multiple or joining multiple into one, these are all personal preferences. This repo has already configured quite strict (far too strict in my opinion) rules, and I've done the changes to comply with them already. 

And about *the number of questions this small PR has already brought up* , at least half of them were because you guys didn't read my PR description or the code carefully to understand in the first place. A week has passed since I requested this PR, four reviewers have reviewed the code, more than 55 comments were made, the real change that actually matters is to remove five characters `\(?.+`, and guess what, it is still not merged.

I understand that people get busy and stuff, but I'm done. Take it or leave it.

test/unit/runner.spec.js
plroebuck: Was trying to annotate (albeit poorly) that `err.stack` had changed behind the scenes by using different variable with "pretty" in its name.

Line length probably triggers Prettier to break things up (but I don't know what column).

test/unit/runner.spec.js
plroebuck: > And about _the number of questions this small PR has already brought up_ , at least half of 
> them were because you guys didn't read my PR description or the code carefully to 
> understand in the first place. A week has passed since I requested this PR, four reviewers 
> have reviewed the code, more than 55 comments were made, the real change that actually 
> matters is to remove five characters `\(?.+`, and guess what, it is still not merged.

I _read_ your PR description. Problem is that you're making the assumption we're extremely familiar with the _entirety_ of the codebase. **Most of us aren't.** As such, we have to learn what the original code does, figure out what changes you made, whether it could have been done differently, what is tested (or wasn't), and what possible breakage could come. We also have to maintain it afterwards, typically only having gained cursory knowledge of the original problem. 
And all that takes unpaid time away from our families... Sorry we're too slow for your taste.

juergba:
@cyjake I didn't get your description at first, especially this Error.stack is non standard and quite fuzzy to catch. A few more words in your description would have helped me a lot. And it's true, most of us don't know every line of code in every detail.

@plroebuck what about one more round of reviewing, then we merge this PR and you do your adaptions later?

plroebuck:
@juergba, I've said what I was going to. You have anything else? If not, let's just merge it and I'll make my changes to the tests afterwards.

juergba:
@plroebuck I didn't follow all the latest comments, but yes let's merge it.

PR #3695, opened 2019-01-27
add zero-config support for loading ES modules in Node

jdalton:
This PR adds zero-config support for loading ES modules in Node. This PR just replaces the entry `require` with `esmRequire`. Now `--require` modules, configs, and others files can have ESM syntax. The caveat is that `.mjs` files are not supported since under the hood mocha is still using `require()` to load files. See the following `esm` doc note:
> 💡 By Node’s rules, builtin `require` cannot sideload `.mjs` files. However, with `esm`, ES modules can be sideloaded as `.js` files or `.mjs` files may be loaded with dynamic `import`.

@boneskull Feel free to push tests to this PR.

### DIFF
--- a/bin/_mocha
+++ b/bin/_mocha
@@ -1,10 +1,12 @@
 #!/usr/bin/env node
 'use strict';
 
+const esmRequire = require('esm')(module);
+
 /**
  * This file remains for backwards compatibility only.
  * Don't put stuff in this file.
  * @see module:lib/cli
  */
 
-require('../lib/cli').main();
+esmRequire('../lib/cli').main();
--- a/package-lock.json
+++ b/package-lock.json
@@ -4813,6 +4813,11 @@
       "integrity": "sha512-qzm/XxIbxm/FHyH341ZrbnMUpe+5Bocte9xkmFMzPMjRaZMcXww+MpBptFvtU+79L362nqiLhekCxCxDPaUMBQ==",
       "dev": true
     },
+    "esm": {
+      "version": "3.1.4",
+      "resolved": "https://registry.npmjs.org/esm/-/esm-3.1.4.tgz",
+      "integrity": "sha512-GScwIz0110RTNzBmAQEdqaAYkD9zVhj2Jo+jeizjIcdyTw+C6S0Zv/dlPYgfF41hRTu2f1vQYliubzIkusx2gA=="
+    },
     "espree": {
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/espree/-/espree-5.0.0.tgz",
--- a/package.json
+++ b/package.json
@@ -489,6 +489,7 @@
     "debug": "3.2.6",
     "diff": "3.5.0",
     "escape-string-regexp": "1.0.5",
+    "esm": "3.1.4",
     "findup-sync": "2.0.0",
     "glob": "7.1.3",
     "growl": "1.10.5",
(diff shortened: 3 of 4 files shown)


boneskull:
@jdalton I can push code to your branch? didn’t realize that was a thing.

jdalton:
@boneskull Yep, [it's a thing](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) :tada:

jdalton:
PS I think the one test-fail in Node 6 on travis-ci might be a false fail.

boneskull:
restarting the job... I haven't seen flake in those tests before.

boneskull:
this closes #3006

jdalton:
> restarting the job... I haven't seen flake in those tests before.

Ah ok. Thank you. It looks like that test still fails so I'll dig in.

boneskull:
sorry about the rebase, I’m stepping on your toes here. let me know when you’re done

jdalton:
(no worries, rebase away)

boneskull:
So this will get `esm` in front of a LOT of people.  I'm thinking it will probably hit some corner-cases.  I'm not too worried about those getting fixed in a timely manner on the `esm` side, but it'd be nice to not block people who do encounter an issue.

To mitigate this, I'm thinking we should do one or more of:

1. Provide an `esm` option which would default to `true`, allowing e.g., `--no-esm` to disable the integration.
1. The above, but set this to `false`.  There's a great argument for Mocha prioritizing a "real" testing environment (as opposed to Jest or other frameworks which may mock/stub/fake a bunch of stuff for whatever reason).
1. Create an exception to our rule about exact production dependencies (that rule being "they should all be exact versions") using either a tilde or caret to accept semver-patch or semver-minor upgrades to `esm` as they are released.  This would allow users encountering problems to easily get a fix in `esm` without Mocha needing to release.
1. `citgm`-like tests against userland modules which monkey with `require`, Node.js' `Module` object, ES modules, stack traces, etc.
1. new integration tests probing the behavior of the systems in the previous item

@jdalton Looking for some recommendations from you here... you will know better what we should be cautious about.

boneskull:
@mochajs/core What are your thoughts on releasing this in v6?  v6.1?  v7?

jdalton:
Thanks @boneskull!

You could check for the Node flag `--experimental-modules` and skip `esm` use in those situations as an intuitive toggle.

While the `esm` loader does mask some stack traces, I'm sure there's gaps. An easy `esm` option would be great! 

An on-by-default might help folks realize they don't need to include Babel. Though I can understand if, when flighted, there are issues that it goes to an off-by-default toggle. 

The `esm` field could also be implemented as a `loader` field so folks could make it something else too. Here is how AVA [detects `esm` in their options](https://github.com/avajs/ava/blob/6b10f2e725a29a30a923704415aa734b60e358c6/lib/worker/subprocess.js#L106-L114).

boneskull:
@jdalton 	what's `Symbol.for('esm\u200D:package')`?

boneskull:
might be premature to hook in to Node's `--loader`, but makes sense whenever that becomes stable.

oh, this PR needs to be retargeted to `master`.  I think that means it needs to be closed and reopened.  I'm not sure where `next` came from, but it's probably my fault.

boneskull:
I think this probably needs to land after v6... I don't think I have time to give it the attention it deserves rn

boneskull:
apparently deleting the target branch closes the issue.

jdalton:
Will re-open under master.

jdalton:
Ok @boneskull, retargeted. Is it possible to just re-open this PR?

jdalton:
No worries. Re-opened as #3703.

Issue #3681, opened 2019-01-19
`--inspect` flag doesn't work


outsideris:
When I tried to debug mocha with `--inspect` option, `ReferenceError: describe is not defined` error occurred. 
 It seems it is not run under `mocha` context.

```sh
$ ./node_modules/.bin/mocha --version
6.0.0-1

$ ./node_modules/.bin/mocha --inspect ex.test.js
Debugger listening on ws://127.0.0.1:9229/aca2ad85-76ab-42ac-99bc-6acb84bd1bc2
For help, see: https://nodejs.org/en/docs/inspector
/Users/outsider/temp/mocha-next/6/ex.test.js:1
(function (exports, require, module, __filename, __dirname) { describe('Suite', () => {
                                                              ^

ReferenceError: describe is not defined
    at Object.<anonymous> (/Users/outsider/temp/mocha-next/6/ex.test.js:1:63)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)
```

With `mocha@5.2.0`, it is fine.

```sh
$ ./node_modules/.bin/mocha --version
5.2.0

$ ./node_modules/.bin/mocha --inspect ex.test.js
Debugger listening on ws://127.0.0.1:9229/4b56feea-1711-43bd-82a3-5a587171122f
For help, see: https://nodejs.org/en/docs/inspector


  Suite
    ✓ test


  1 passing (5ms)
```

outsideris:
I tried to fix this issue, but I could find a good solution. Any idea? /cc @mochajs/core 

```sh
$ ./bin/mocha --inspect example.js

  mocha:cli exec /Users/outsider/bin/lib/node.js/node-v10.15.0-darwin-x64/bin/node w/ args: [ '--inspect',
  'example.js',
  '/Users/outsider/mocha/bin/_mocha',
  '--timeout',
  '200',
  '--no-config',
  '--no-package',
  '--no-opts',
  '--require',
  'test/setup',
[...]

PR #3698, opened 2019-01-28
ensure --debug becomes --inspect; closes #3697

### LINKED ISSUE #3697: --debug doesn't invoke --inspect correctly
at least, in theory, because of this typo:

https://github.com/mochajs/mocha/blob/1e305012428dafc72a664623553cc1783ceddaff/bin/mocha#L42-L46

boneskull:
- add some handiness to the integration test helpers
- update docstrings in integration test helpers

Found this when looking at #3681.

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -9,7 +9,7 @@
  * @private
  */
 
-const {deprecate} = require('../lib/utils');
+const {deprecate, warn} = require('../lib/utils');
 const {spawn} = require('child_process');
 const {loadOptions} = require('../lib/cli/options');
 const {isNodeFlag, impliesNoTimeouts} = require('../lib/cli/node-flags');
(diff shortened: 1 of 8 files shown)


boneskull:
saucelabs flake 😢

bin/mocha
```diff
@@ -39,8 +39,9 @@ Object.keys(opts).forEach(opt => {
   }
 });
 
-// allow --debug to invoke --inspect on Node.js v8 or newer  nodeOpts.inspect = childOpts.debug;
+// allow --debug to invoke --inspect on Node.js v8 or newer
 if (childOpts.debug) {
+  nodeOpts.inspect = childOpts.debug;
   childOpts.timeout = false;
   delete childOpts.debug;
   debug('--debug -> --inspect');
```
juergba: I don't understand where the distinction is made between Node >=v8 and Node v6.
We will deprecate `--debug` anyway once support of Node v6 is dropped, right? We could prepare this deprecation and pass the responsibility of choosing the correct Node flag to the user.

test/assertions.js
```diff
@@ -59,9 +59,9 @@ exports.mixinMochaAssertions = function(expect) {
       }
     )
     .addAssertion(
-      '<RawRunResult|JSONRunResult> [not] to have completed with [exit] code <number>',
+      '<RawResult|RawRunResult|JSONRunResult> [not] to have [completed with] [exit] code <number>',
       function(expect, result, code) {
```
juergba: why those additional "[ ... ]"?

test/assertions.js
```diff
@@ -290,17 +290,10 @@ exports.mixinMochaAssertions = function(expect) {
...
-      output
-    ) {
-      expect(result.output, '[not] to satisfy', output);
-    })
     .addAssertion(
-      '<RawRunResult|JSONRunResult> to have [exit] code <number>',
-      function(expect, result, code) {
-        expect(result.code, 'to be', code);
+      '<RawResult|RawRunResult> [not] to contain [output] <any>',
+      function(expect, result, output) {
```
juergba: should it be "[output]" or "output"? I don't know.

juergba: there are more examples: "test count" or "[test] count"?

bin/mocha
boneskull: Yeah, I agree, we should provide a deprecation message when debug is used outside of Node.js v6.

test/assertions.js
boneskull: This rule was duplicated above, so I consolidated it.

So this allows you to say "<var> to have completed with exit code <code>" or "<var> to have exit code <code>" or "<var> to have completed with code <code>".

I find it's nicer to provide several different ways of expressing the same thing.

bin/mocha
boneskull: actually I changed my mind.  a warning should be printed, but it's a nice convenience for users until Node.js v6 drops out of active maintenance.  at that point, we can remove it, because we'll drop v6 support as well.

PR #3699, opened 2019-01-28
 fix --inspect and its ilk; closes #3681

### LINKED ISSUE #3681: `--inspect` flag doesn't work

boneskull:
~~this is rebased onto the branch from #3698 and thus depends on it.~~

This seems to make `--inspect` work properly with or without a parameter (`--inspect=0.0.0.0:12345`).  The `=` is required, as it's required by `node`.

In v6 of Node, `--debug --port=12345` should also work as expected.

**UPDATE** #3698 was merged

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -12,10 +12,15 @@
 const {deprecate, warn} = require('../lib/utils');
 const {spawn} = require('child_process');
 const {loadOptions} = require('../lib/cli/options');
-const {isNodeFlag, impliesNoTimeouts} = require('../lib/cli/node-flags');
+const {
+  unparseNodeFlags,
+  isNodeFlag,
+  impliesNoTimeouts
+} = require('../lib/cli/node-flags');
 const unparse = require('yargs-unparser');
-const debug = require('debug')('mocha:cli');
+const debug = require('debug')('mocha:cli:mocha');
 const {aliases} = require('../lib/cli/run-option-metadata');
+const nodeEnv = require('node-environment-flags');
 
 const mochaPath = require.resolve('./_mocha');
 const mochaArgs = {};
@@ -45,13 +50,12 @@ const disableTimeouts = value => {
  * @ignore
  */
 const trimV8Option = value =>
-  value !== 'v8-options' && /^v8-/.test(value) ? value.slice(2) : value;
+  value !== 'v8-options' && /^v8-/.test(value) ? value.slice(3) : value;
 
 // sort options into "node" and "mocha" buckets
 Object.keys(opts).forEach(opt => {
-  opt = trimV8Option(opt);
   if (isNodeFlag(opt)) {
-    nodeArgs[opt] = opts[opt];
+    nodeArgs[trimV8Option(opt)] = opts[opt];
     disableTimeouts(opt);
   } else {
     mochaArgs[opt] = opts[opt];
@@ -90,9 +94,8 @@ if (/^(debug|inspect)$/.test(mochaArgs._[0])) {
 }
 
 // allow --debug to invoke --inspect on Node.js v8 or newer.
-// these show up in childOpts because they are not recognized as valid node flags in this version of node.
 ['debug', 'debug-brk']
-  .filter(opt => opt in mochaArgs)
+  .filter(opt => opt in nodeArgs && !nodeEnv.has(opt))
   .forEach(opt => {
     const newOpt = opt === 'debug' ? 'inspect' : 'inspect-brk';
     warn(
@@ -100,10 +103,10 @@ if (/^(debug|inspect)$/.test(mochaArgs._[0])) {
         process.version
       }; use "--${newOpt}" instead.`
     );
-    nodeArgs[newOpt] = mochaArgs[opt];
+    nodeArgs[newOpt] = nodeArgs[opt];
     mochaArgs.timeout = false;
     debug(`--${opt} -> ${newOpt}`);
-    delete mochaArgs[opt];
+    delete nodeArgs[opt];
   });
 
 // historical
@@ -115,8 +118,10 @@ if (nodeArgs.gc) {
   delete nodeArgs.gc;
 }
 
+debug('final node args', nodeArgs);
+
 const args = [].concat(
-  unparse(nodeArgs),
+  unparseNodeFlags(nodeArgs),
   mochaPath,
   unparse(mochaArgs, {alias: aliases})
 );
--- a/bin/mocha
+++ b/bin/mocha
@@ -37,7 +37,7 @@ debug('loaded opts', opts);
 const disableTimeouts = value => {
   if (impliesNoTimeouts(value)) {
     debug(`option "${value}" disabled timeouts`);
-    mochaArgs.timeout = false;
+    mochaArgs.timeout = 0;
     delete mochaArgs.timeouts;
     delete mochaArgs.t;
   }
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -18,6 +18,7 @@ const findup = require('findup-sync');
 const {deprecate} = require('../utils');
 const debug = require('debug')('mocha:cli:options');
 const {createMissingArgumentError} = require('../errors');
+const {isNodeFlag} = require('./node-flags');
 
 /**
  * The `yargs-parser` namespace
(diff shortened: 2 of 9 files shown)


test/integration/options/debug.spec.js
```diff
@@ -0,0 +1,62 @@
...
+          expect(res, 'to have passed').and(
+            'to contain output',
+            /Debugger listening/i
+          );
+          done();
+        },
+        {stdio: 'pipe'}
+      );
+    });
+  });
```
outsideris: How about we add node flag tests like: 

```
it('should handle node flags with equal sign', function(done) {
  invokeMocha(
    ['--inspect=127.0.0.1:9229', DEFAULT_FIXTURE],
    function(err, res) {
      if (err) {
        return done(err);
      }
      expect(res, 'to have passed').and(
        'to contain output',
        /Debugger listening/i
      );
      done();
    },
    {stdio: 'pipe'}
  );
});
```

package.json
```diff
@@ -498,7 +498,7 @@
     "minimatch": "3.0.4",
     "mkdirp": "0.5.1",
     "ms": "2.1.1",
-    "node-environment-flags": "1.0.2",
+    "node-environment-flags": "1.0.3",
```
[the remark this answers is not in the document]
juergba: `--inspect[=[host:]port]` is added early in v6.3.0
`--inspect-brk[=[host:]port]` in v7.6.0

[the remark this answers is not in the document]
boneskull: @juergba 	good catch, I've updated `node-environment-flags` appropriately and will upgrade

test/integration/options/debug.spec.js
boneskull: added this test and a couple others

boneskull:
normal boolean node flags also are subject to the positional argument munching problem, so I'll need to fix those, e.g., `mocha --prof foo.js` thinks the value of `prof` is `foo`.  we need to assume all node flags are boolean unless we know otherwise.

juergba:
My solution would be:
- Define all options that way they don't consume any subsequent parameter as value:
  - by typing them as boolean
  - or by using nargs( key, 0)    (influences dynamically created `--help`)
- define default values, f.e. boolean=true
- to assign a value in CL, the "=" has to be used

The handling of all flags - Mocha and Node - is consistent, all require nothing or "=".

`--inspect` ok, works
`--inspect=0.0.0.0:12345` ok, works
`--inspect 0.0.0.0:12345` no, does not work

boneskull:
yargs-parser (which we're consuming directly) has no option which requires `=` (though they could add one), which means we won't be able to easily restrict this.  That said...

> But v6 could be a good occasion to change that. And Node seems to parse that way (see f.e. --inspect[=[host:]port]).

IMO...

This ("yargs-parser is confused" or "node.js does it this way") is a poor reason to break backwards compatibility.  

Breaking backwards compat for such a widely-used API should be a last resort. We're not automatically justified in breaking *anything* just because it's a major version bump.  

Other projects may have the luxury of issuing frequent breaking changes, or have the resources to provide migration scripts, but Mocha is the most widely-depended-on package on npm--our justifications need to be solid.  For more unpopular APIs, it's easier to make that justification.

(It *would* be helpful to provide some information on API maturity.  APIs marked "experimental" would be subject to breaking changes; "stable" APIs will likely never change, etc.)

Obviously I'm still trying to get this one working as intended. 

AFAICT `nargs: 0` doesn't actually do anything.

lib/cli/options.js
```diff
@@ -75,22 +76,46 @@ const nargOpts = types.array
...
+  // 3. to avoid explicitly defining the set of them, we tell yargs-parser they
+  //    are ALL boolean flags.
+  // 4. we can then reapply the values after yargs-parser is done.
+  const nodeArgs = (Array.isArray(args) ? args : args.split(' ')).reduce(
+    (acc, arg) => {
+      const pair = arg.split('=');
+      const flag = pair[0].replace(/^--?/, '');
+      if (isNodeFlag(flag)) {
+        return arg.includes('=')
+          ? acc.concat([[flag, pair[1]]])
```
boneskull: (sic)

boneskull: we save ALL found node args here so we can use them in the boolean array below.
the values are saved as whatever they are; if there is no value (e.g., `--inspect`), then the value is saved as `true`, which is what yargs-unparser understands.

boneskull:
(still not ready)

boneskull:
just going to skip windows tests for now, because the behavior upon `SIGINT` differs (the child process doesn't return an exit code OR any output, afaik).  I'm not knowledgeable enough about windows processes to know how to address this myself. maybe @juergba 	can take a look at some point.

boneskull:
I did naively test this on a windows machine, and it works, it's just that it cannot be tested the same way.

boneskull:
@juergba 	Please complete a review if/when you are satisfied

juergba:
I did some quick tests and have found some issues:
- there is no error when using non-existent options
(you could add synonyms for uppercase shortcuts like -V/-v, -A/-a, -G/-g, -R/-r, -S/-s)
- the `--help` output  does not list `--inspect` / `--debug` 
- I have difficulties debugging out of my VSCode, most breakpoints are just ignored. It seems to happen when calling `bin/mocha`. With `bin/_mocha` it seems to work though.

Btw: `--timeout 5000` and `--timeout=5000` both ways do work.

boneskull:
> there is no error when using non-existent options

This is intended (for now)

> the --help output does not list --inspect / --debug

This is also intended, as they are Node flags.

> I have difficulties debugging out of my VSCode, most breakpoints are just ignored. It seems to happen when calling bin/mocha. With bin/_mocha it seems to work though.

This is expected, but not ideal. When you create a "Mocha" launch config, it points to `bin/_mocha` for this reason.  It has something to do with being unable to debug the child process, but I don't know why.

boneskull:
> I guess --inspect-brk is parsed twice. Once by Node directly (since flag follows immediatly after node.exe), then a second time by Mocha. Possibly the port gets lost on the way. In this constellation Mocha should not parse these flags a second time. (I'm not sure, just thinking loud)

I don't really follow you here, but the node flags shouldn't be passed into Mocha a second time.

boneskull:
> Calling a Mocha flag like this:
> node --bail D:\mocha3291\mocha\bin\mocha async-only-async.fixture.js
> results in Node error: node: bad option: --bail

This is expected, because `node` rejects unknown options; this is invalid usage of Mocha flags.  There's no way Mocha can intercept any arguments before `node` does.

bin/mocha
```diff
@@ -45,13 +50,12 @@ const disableTimeouts = value => {
  * @ignore
```
juergba: above L40:
`mochaArgs.timeout` is a number, should be set to 0?

bin/mocha
```diff
@@ -45,13 +50,12 @@ const disableTimeouts = value => {
...
  */
 const trimV8Option = value =>
-  value !== 'v8-options' && /^v8-/.test(value) ? value.slice(2) : value;
+  value !== 'v8-options' && /^v8-/.test(value) ? value.slice(3) : value;
 
 // sort options into "node" and "mocha" buckets
 Object.keys(opts).forEach(opt => {
```
juergba: just noted, there is a Mocha flag `-r` and also one in `nodeEnv` . In this case flag is put into mocha bucket. Could lead to conflicts?

test/integration/options/debug.spec.js
```diff
@@ -25,8 +32,116 @@ describe('--debug', function() {
...
+          done();
+        },
+        'pipe'
+      );
+
+      // debugger must be manually killed
+      setTimeout(function() {
+        proc.kill('SIGINT');
+      }, 1000);
+    });
```
juergba: change to:
```
setTimeout(function() {
  process.kill(proc.pid, 'SIGINT');
}, 2000);
```
This way the `err.code` is 1 (not undefined which crashes `to have failed`).
A timeout of 1000 is too short, `res` may not be ready, yet => `to contain output` fails.

On my windows platform the four 'Node.js v8+' tests do pass.

bin/mocha
boneskull: when it's `false` it gets set to `0`, but might as well make it `0`.

bin/mocha
boneskull: Mocha's `--require` and `-r` are both preferred over Node's, see `isNodeFlag` in `lib/cli/node-flags.js`

bin/mocha
```diff
@@ -115,8 +118,10 @@ if (nodeArgs.gc) {
...
 }
 
+debug('final node args', nodeArgs);
+
 const args = [].concat(
-  unparse(nodeArgs),
+  unparseNodeFlags(nodeArgs),
   mochaPath,
   unparse(mochaArgs, {alias: aliases})
 );
```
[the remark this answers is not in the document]
boneskull: the short answer is "yes"

test/integration/options/debug.spec.js
```diff
@@ -25,8 +32,116 @@ describe('--debug', function() {
...
+
+    it('should invoke --inspect-brk', function(done) {
+      var proc = invokeMocha(
+        ['--debug-brk', DEFAULT_FIXTURE],
+        function(err, res) {
+          if (err) {
+            return done(err);
+          }
+          expect(res, 'to have passed').and(
+            'to contain output',
```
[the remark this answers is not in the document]
juergba: The exit code (on windows) is 1, so `expect(res, 'to have passed').and(` will not work.

plroebuck:
> ...but Mocha is the most widely-depended-on package on npm --

Uh, according to npm, , that would be 'lodash'.
Mocha's in the [mid-60's](https://www.npmjs.com/browse/depended?offset=66), though Top 100 is nothing to sneeze at. 

Where does Libraries.io get their numbers?

boneskull:
> > ...but Mocha is the most widely-depended-on package on npm --
> 
> Uh, according to npm, , that would be 'lodash'.
> Mocha's in the [mid-60's](https://www.npmjs.com/browse/depended?offset=66), though Top 100 is nothing to sneeze at.
> 
> Where does Libraries.io get their numbers?

npm's list only reports `dependencies`, not `devDependencies`, where most people use Mocha.  taken as a sum--`dependencies` and `devDependencies`--Mocha has the most by a considerable margin.

PR #3700, opened 2019-01-29
Show total test cases correctly in growl

outsideris:
### Description of the Change
When I run exclusive tests with like `only`, growl notification show total tests count instead test count which I ran.

IMO, it should be test's count which I ran.

### Alternate Designs
N/A

Growl notification show total test cases which users ran.

### Possible Drawbacks
N/A

### DIFF
--- a/lib/browser/growl.js
+++ b/lib/browser/growl.js
@@ -129,7 +129,7 @@ function display(runner) {
   var title;
 
   if (stats.failures) {
-    _message = stats.failures + ' of ' + runner.total + ' tests failed';
+    _message = stats.failures + ' of ' + stats.tests + ' tests failed';
     message = symbol.cross + ' ' + _message;
     title = 'Failed';
   } else {
--- a/lib/growl.js
+++ b/lib/growl.js
@@ -64,7 +64,7 @@ const display = runner => {
   let title;
 
   if (stats.failures) {
-    _message = `${stats.failures} of ${runner.total} tests failed`;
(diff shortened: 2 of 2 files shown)


plroebuck:
While I understand what you want here, left wondering why `stats.tests` isn't adjusted by `only`, etc. Don't think reporters (which includes Growl) should need to use anything but `stats` object.

boneskull:
@plroebuck 

> While I understand what you want here, left wondering why stats.tests isn't adjusted by only, etc. Don't think reporters (including Growl) should need to use anything but stats object.

It is.  `runner.total` is a different number (and the wrong one to be using).

boneskull:
`stats.tests` is how many tests mocha ran.
`runner.total` is how many tests mocha *could* run, calculated before execution.

`runner.total` may need to be checked for accuracy, as it should change when `describe.skip()`, `describe.only()`, `it.only()` or `it.skip()` are used (in addition to `--grep/--fgrep`).

we shouldn't change `runner.total` after the runner's `start` event is emitted.

plroebuck:
Why did we merge a non-change? Shouldn't this be reverted?

outsideris:
@plroebuck What should it change to?

plroebuck:
Nevermind, your change is correct. Good catch.
I had just migrated the existing code, and hadn't noticed it was using `runner.total`.
But assumed it matched what the reporters were using.

PR #3704, opened 2019-01-30
test(unit/runner.spec.js): Minor tweeks related to #3686

plroebuck:
### Description of the Change
Refactored portions of tests for (hopefully) increased clarity.

### Alternate Designs
NA

NA

### Benefits
Easier to follow maintenance-wise... well, to me anyway.

### Possible Drawbacks
None to knowledge

### Applicable issues
#3686
semver-patch

### DIFF
--- a/test/unit/runner.spec.js
+++ b/test/unit/runner.spec.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var fs = require('fs');
 var path = require('path');
 var mocha = require('../../lib/mocha');
 var Suite = mocha.Suite;
@@ -452,13 +451,27 @@ describe('Runner', function() {
       'at processImmediate [as _immediateCallback] (timers.js:321:17)'
     ];
 
-    describe('shortStackTrace', function() {
-      beforeEach(function() {
-        if (process.platform === 'win32') {
-          this.skip();
-        }
-      });
+    before(function() {
+      // Only for Node running on Windows
+      if ('platform' in process && process.platform === 'win32') {
+        var addDrive = function(str) {
+          var drive = 'C:';
+          var pos = str.indexOf(path.posix.sep);
+          return pos !== -1 ? str.slice(0, pos) + drive + str.slice(pos) : str;
+        };
 
+        var useWinPathSep = function(str) {
+          return str.split(path.posix.sep).join(path.win32.sep);
+        };
+
+        // Fake Windows pathnames in stacktrace
+        stack = stack.map(function(line) {
+          return useWinPathSep(addDrive(line));
+        });
+      }
+    });
+
+    describe('shortStackTrace', function() {
       it('should prettify the stack-trace', function(done) {
         var hook = new Hook();
         hook.parent = suite;
@@ -475,12 +488,6 @@ describe('Runner', function() {
     });
 
     describe('longStackTrace', function() {
-      beforeEach(function() {
-        if (process.platform === 'win32') {
-          this.skip();
-        }
-      });
-
       it('should display the full stack-trace', function(done) {
         var hook = new Hook();
         hook.parent = suite;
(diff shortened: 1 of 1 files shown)


plroebuck:
Anyone know how to fix whatever's wrong with our TravisCI SauceLab jobs?
```console
The command "travis_start_sauce_connect" failed and exited with 1 during .
```

boneskull:
> Anyone know how to fix whatever's wrong with our TravisCI SauceLab jobs?

No.  Can you restart the job?

boneskull:
I restarted the jobs my damn self

test/unit/runner.spec.js
```diff
@@ -454,7 +454,7 @@ describe('Runner', function() {
 
     describe('shortStackTrace', function() {
       beforeEach(function() {
-        if (path.sep !== '/') {
+        if (process.platform === 'win32') {
```
cyjake: It looks like the failed tests weren't meant to run because `path.sep !== '/'` rules out browsers, but `process.platform == 'win32'` doesn't. Haven't investigate how this unit test runs in browser though. It seems `requrie('path')` just gives empty obejct `{}`.

cyjake: @boneskull It's because `path` and `fs` are both ignored in the browserify config from `karma.conf.js`. IMHO, several changes are needed to fix this unit test:

- calibrate test data `stack` by replacing `/` to `path.sep` if `path.sep === '\'`
- change the way multiline error message is generated in browser, using `document.body.outerHTML` perhaps? Or just remove this test instead, which is just complimentary to the single line test anyway. The issue is already covered.

boneskull: this test probably doesn’t belong lumped in with the unit tests, and/or the code under test should be refactored out of the runner. the unit tests _shouldn’t_ be platform-dependent.

good catch

test/unit/runner.spec.js
```diff
@@ -452,9 +451,29 @@ describe('Runner', function() {
       'at processImmediate [as _immediateCallback] (timers.js:321:17)'
     ];
 
+    before(function() {
+      // Only for Node running on Windows
+      if ('platform' in process && process.platform === 'win32') {
```
cyjake: `'platform' in process` is unnecessary check

test/unit/runner.spec.js
```diff
@@ -452,9 +451,29 @@ describe('Runner', function() {
...
+          return useWinPathSep(addDrive(line));
+        });
+      }
+    });
+
     describe('shortStackTrace', function() {
-      beforeEach(function() {
-        if (path.sep !== '/') {
+      before(function() {
+        if ('browser' in process) {
```
cyjake: personally I'd prefer `if (process.brower) {`

cyjake: Since we've found that this test was skipped in browser testing unintentionally, should it be turned on instead of continue skipping?

plroebuck: It doesn't currently work.
https://travis-ci.org/mochajs/mocha/jobs/486574106#L1594

test/unit/runner.spec.js
plroebuck: > It looks like the failed tests weren't meant to run because `path.sep !== '/'` rules out browsers, but `process.platform == 'win32'` doesn't. Haven't investigate how this unit test runs in browser though. It seems `requrie('path')` just gives empty obejct `{}`.

Somewhere in browserify-land, `path.sep` is hardcoded to use POSIX. `process.platform` [won't exist](https://github.com/defunctzombie/node-process/blob/master/browser.js) when browserified.

test/unit/runner.spec.js
plroebuck: > personally I'd prefer `if (process.brower) {`

Lacking an [existential operator](https://esdiscuss.org/topic/the-existential-operator), IMO this style better communicates the fact that it _may not exist_.

test/unit/runner.spec.js
plroebuck: communicates the fact that it _may not exist_.

test/unit/runner.spec.js
```diff
@@ -499,29 +512,47 @@ describe('Runner', function() {
...
           this.skip();
         }
       });
 
-      it('should not hang if the error message is ridiculously long single line', function(done) {
+      // Generate 64k string
```
plroebuck: @cyjake, did you determine the lower bound of what you termed "ridiculously long"?

test/unit/runner.spec.js
plroebuck: Actually, [it does](https://github.com/defunctzombie/node-process/blob/77caa43cdaee4ea710aa14d11cea1705293c0ef3/browser.js#L156)...

plroebuck:
> does Travis test the Windows part?

No, but AppVeyor does.
So all tests now run on UN*X and Windows.
Skips stacktrace filter-related tests in browser since they don't work correctly.

Issue #3689, opened 2019-01-24
refactor: encapsulation issues


boneskull:
There are many places throughout the codebase (mostly in the main classes) where we modify data via assignment in some object (e.g., a `Suite`) directly from some *other* class.  this includes properties which are ostensibly "private" (e.g. `Suite#_onlyTests`).

This breaks encapsulation/information-hiding. Refactor these instances by creating methods within the classes having these properties, and call those methods from whatever code is currently doing direct assignment.

Example:

https://github.com/mochajs/mocha/blob/b5d8513887e71477861db74a43b0d84fd191f313/lib/interfaces/common.js#L169-L180

Instead, `Suite` could expose an instance method (e.g., `appendOnlyTest`) which appends the test to its `_onlyTests` array.  The consuming code would call `test.parent.appendOnlyTest(test)`.  Going a step further, the `Test` class could expose an instance method (e.g. `markOnly()`) which would make the call.

These changes can be incremental; we don't have to do it all at once.  Any new methods will need unit tests.

plroebuck:
[The Paperboy, The Wallet,
and The Law Of Demeter](https://www2.ccs.neu.edu/research/demeter/demeter-method/LawOfDemeter/paper-boy/demeter.pdf)

boneskull:
yes. 
even the “bad” example in that pdf is better than what we currently have!

doniyor2109:
Hi. Can I take the issue?

shubham-rocko:
Hi, Can I take this issue if it is not fixed yet?

AbhinavHaridas:
Hi! Can I work on this issue?

JoshuaKGoldberg:
Sure, go ahead. 

Just note that per https://github.com/mochajs/mocha/issues/5027 we're still ramping up as new maintainers and might not be able to give much help or review your PR for a while.

PR #3707, opened 2019-01-31
Refactor out usages of Suite#_onlyTests and Suite#_onlyTests (#3689)

vkarpov15:
### Requirements

### Description of the Change

Re: #3689, there's a few places where `common/interface.js` and `runner.js` reach into `Suite` internals, specifically with using `_onlyTests` and `_onlySuites`. This PR refactors this logic into a helper function on `Suite`, so there's no code that accesses `_onlyTests` and `onlySuites` outside of the `Suite` class.

### Alternate Designs

@boneskull 's `Test.markOnly()` suggestion is a good next step but is not strictly necessary for the goal of improved encapsulation.

Refactoring can't be in a plugin

### Benefits

Cleaner abstraction barriers: `Suite` is responsible for modifying its own internal state.

### Possible Drawbacks

`hasOnly()` and `filterOnly()` were previously not exported. Users may start using these functions in unexpected ways.

### Applicable issues

#3689, patch release


lib/suite.js
```diff
@@ -427,3 +427,66 @@ Suite.prototype.run = function run() {
     this.emit('run');
   }
 };
+
+/**
+ * Determines whether a suite has an `only` test or suite as a descendant.
+ *
+ * @returns {Boolean}
+ */
+Suite.prototype.hasOnly = function hasOnly() {
```
boneskull: should be marked `@private`

lib/suite.js
```diff
@@ -427,3 +427,66 @@ Suite.prototype.run = function run() {
...
+    })
+  );
+};
+
+/**
+ * Filter suites based on `isOnly` logic.
+ *
+ * @returns {Boolean}
+ */
+Suite.prototype.filterOnly = function filterOnly() {
```
boneskull: should be marked `@private`

lib/suite.js
```diff
@@ -427,3 +427,66 @@ Suite.prototype.run = function run() {
...
+  // Keep the suite only if there is something to run
+  return this.tests.length || this.suites.length;
+};
+
+/**
+ * Adds a suite to the list of subsuites marked `only`.
+ *
+ * @param {Suite} suite
+ */
+Suite.prototype.appendOnlySuite = function(suite) {
```
boneskull: should be marked `@private`

lib/suite.js
```diff
@@ -427,3 +427,66 @@ Suite.prototype.run = function run() {
...
+  this._onlySuites.push(suite);
+};
+
+/**
+ * Adds a test to the list of tests marked `only`.
+ *
+ * @param {Test} test
+ */
+Suite.prototype.appendOnlyTest = function(test) {
```
boneskull: should be marked `@private`

boneskull:
@vkarpov15 	right, then they wouldn't have tests.  but they should probably have unit tests now since they're class methods.  `test/unit/suite.spec.js` would be the place to add these

vkarpov15:
@boneskull I addressed your comments, let me know if you have any more suggestions :+1:

test/unit/suite.spec.js
```diff
@@ -545,6 +545,89 @@ describe('Suite', function() {
...
+      var test = new Test('c');
+      var test2 = new Test('d');
+
+      suite.suites.push(nested);
+      suite.appendOnlyTest(test);
+      suite.tests.push(test2);
+
+      suite.filterOnly();
+
+      expect(suite.suites.length, 'to be', 0);
```
boneskull: it's best to make a single assertion where possible.  you can do this

```js
expect(suite, 'to satisfy', {
  suites: expect.it('to be empty'),
  tests: expect.it('to have length', 1).and('to contain', {title: 'c'})
});
```

similarly below

vkarpov15: @boneskull I addressed the comments, but I'm curious: why is a single assertion best?

jsdoc.conf.json
```diff
@@ -27,7 +27,7 @@
     "default": {
       "includeDate": false,
       "outputSourceFiles": true
-  },
+    },
```
plroebuck: How'd this get past prettier?

test/unit/suite.spec.js
plroebuck: I myself prefer multiple explicit assertions that I don't have to lookup within unexpected config files. Seems like violation of the KISS mentality...

jsdoc.conf.json
vkarpov15: No idea. Do you know whether my change is correct or the original was correct?

test/unit/suite.spec.js
vkarpov15: TBH I also prefer multiple explicit assertions, but that's mostly because I'm not very good with unexpected's syntax off the top of my head :)

vkarpov15:
@plroebuck what's the policy on merging this - can I do it myself or do I need to wait for another maintainer to do so?

plroebuck:
> @plroebuck what's the policy on merging this - can I do it myself or do I need to wait for another maintainer to do so?

I usually wait for two accepts (unless it's just trivial) before merge. Either the more controversial the PR or number of significantly touched files should increase the number of needed accepts.

By my thinking, merge away...

vkarpov15:
Cool in which case I'll merge this. This is a non-controversial refactor with two approvals, no sense in waiting for the PR to go stale. Feel free to revert if it causes problems.

PR #3715, opened 2019-02-04
Revert 00ca06b; closes #3414

### LINKED ISSUE #3414: --bail and --retries events: multiple "end" occurs

boneskull:
- Reverts 00ca06b
- Update event-order tests
- Added retries-and-bail case

Reopens #2753


boneskull:
cc @1999

test/integration/helpers.js
```diff
@@ -81,27 +81,30 @@ module.exports = {
...
-                'Failed to parse JSON reporter output.\nArgs: %O\nResult:\n\n%O',
-                args,
+                'Failed to parse JSON reporter output: %s\n\nResult:%O',
+                err,
                 res
               )
             )
           );
         }
+        fn(null, result);
```
plroebuck: Why move this? Original location was more to the point.
```js
try {
  return fn(null, toJSONRunResult(res));
} catch (err) {
  return fn(
    new Error(
      format(
        'Failed to parse JSON reporter output: %s\n\nResult:%O',
        err,
        res
      )
    )
  );
}
```

test/integration/options/grep.spec.js
```diff
@@ -28,8 +21,7 @@ describe('--grep', function() {
 
   describe('should run specs matching a RegExp', function() {
     it('with RegExp-like strings (pattern follow by flag)', function(done) {
```
plroebuck: ...followed...

test/integration/options/grep.spec.js
```diff
@@ -68,40 +58,45 @@ describe('--grep', function() {
...
+        ['--fgrep', 'first', '--grep', 'second'],
         function(err, res) {
           if (err) {
             return done(err);
           }
-          expect(res, 'to satisfy', {
-            code: 1,
-            output: /--invert.*--grep <regexp>/
-          });
+          expect(res, 'to have failed with output', /mutually exclusive/i);
```
plroebuck: no longer verifies failure exit code

test/integration/options/grep.spec.js
```diff
@@ -68,40 +58,45 @@ describe('--grep', function() {
...
+        function(err, res) {
+          if (err) {
+            return done(err);
+          }
+          expect(
+            res,
+            'to have failed with output',
+            /--invert.*--grep <regexp>/
+          );
+          done();
```
plroebuck: no longer verifies failure exit code

test/integration/options/compilers.spec.js
```diff
@@ -4,20 +4,18 @@ var helpers = require('../helpers');
...
-  it('should fail', function(done) {
-    invokeMocha(args, function(err, res) {
-      if (err) {
-        return done(err);
-      }
-
-      expect(res.code, 'to be', 1);
-      done();
-    });
+        expect(res, 'to have failed with output', /compilers is deprecated/i);
```
plroebuck: no longer verifies failure exit code

test/integration/options/grep.spec.js
```diff
@@ -68,40 +58,45 @@ describe('--grep', function() {
...
-      args = ['--fgrep', 'first', '--grep', 'second'];
-
-      runMocha(fixture, args, function(err, res) {
-        if (err) {
-          return done(err);
-        }
-        expect(res, 'to have failed');
-        done();
-      });
+describe('--invert', function() {
```
plroebuck: this breaks lexical ordering...

test/integration/options/compilers.spec.js
boneskull: it does, actually: 'to have failed with output' ensures the exit code is nonzero

test/integration/helpers.js
boneskull: because the logic was incorrect.  it would catch a potential error thrown out of the `fn` callback.

boneskull: I'll add a comment explaining myself

test/integration/options/grep.spec.js
boneskull: again, if we can't automate the check or reformat automatically...

[eslint-import-order-alphabetical](https://www.npmjs.com/package/eslint-plugin-import-order-alphabetical) is probably a good place to start implementing such a thing.  it might not have a `--fix` implementation, but the check is better than nothing.

test/assertions.js
```diff
@@ -100,6 +100,24 @@ exports.mixinMochaAssertions = function(expect) {
         code: expect.it('to be greater than', 0)
       });
     })
+    .addAssertion(
```
boneskull: @plroebuck new assertions for common "fail and validate output" checks

test/assertions.js
```diff
@@ -100,6 +100,24 @@ exports.mixinMochaAssertions = function(expect) {
...
+      '<RawRunResult|RawResult> [not] to have failed (with|having) output <any>',
+      function(expect, result, output) {
+        expect(result, '[not] to satisfy', {
+          code: expect.it('to be greater than', 0),
+          output: output
+        });
```
plroebuck: OK. See your rule, but don't see how it would be satisfied.
```js
// simulate output from [_spawnMochaWithListeners](https://github.com/mochajs/mocha/blob/c77d396e7607289a40c2cf5713403a2ed72c9bf9/test/integration/helpers.js#L261)
var res = {
  output: ['{ "foo": 123, "bar": true, "baz": "abc" }'],
  code: 1,
  args: '--whatever'
};
// convert run result from [toJSONRunResult](https://github.com/mochajs/mocha/blob/c77d396e7607289a40c2cf5713403a2ed72c9bf9/test/integration/helpers.js#L186)
var result = toJSONRunResult(res);
// => { foo: 123, bar: true, baz: 'abc', code: 1 }
// fn(null, result);

It still has `code`, but wasn't `output` lost in translation?
```

test/integration/options/grep.spec.js
plroebuck: The ordering was done by filename, each option in its own file with a single `describe`, merged just two weeks ago...

test/assertions.js
boneskull: hmm, I still don't really follow.  this does not apply to `JSONRunResult`; it can't be used with `runMochaJSON`.  generally when we use `runMochaJSON` we're more interested in specific information, e.g., did test _x_ pass and how many pending tests were there.

test/integration/options/grep.spec.js
boneskull: I'll move `--invert` into its own file for consistency.  my reasoning was that a) the test was already in this file and b) the option can't be used without (f)grep

plroebuck: > my reasoning was that a) the test was already in this file and b) the option can't be used without (f)grep

That's why it was previously nested...

test/assertions.js
plroebuck: I see now. Assumed this was related to your change in "test/integration/helpers.js", but it's unrelated to the other unrelated changes of reverting the original PR...

PR #3719, opened 2019-02-07
Deprecate this.skip() for "after all" hooks

juergba:
### Description

There have been long discussions about the behavior of `this.skip()` in hooks. The `before` hooks have already been clarified and fixed in #3225. The purpose of this PR is to make a decision about the behavior of `after` and `afterEach` hooks.

https://github.com/mochajs/mocha/pull/2571#issuecomment-265416756

> Skipping a test/suite after running it is a bit counter logical. The only purpose would be marking it as pending. Which is not very clean because you've already emitted `pass`/`fail` events for such test. Not sure there's value in changing the test state/outcome post-mortem.

I propose to emit a DeprecationWarning when `this.skip()` is used in `after` and `afterEach` hooks and to remove it in a future release. Besides marking an already passed/failed test as pending, there are actually no obvious effects of calling `this.skip()` in mentioned hooks.
```js
    beforeEach(function () {
        console.log("A");
    });
    describe('inner', function () {
        beforeEach(function () {
            console.log("B");
        });
        it('test case', function () {
            console.log("C");
        });
        it('test case', function () {
            console.log("C2");
        });
        afterEach(function () {          // or after()
            console.log("Y");
            this.skip();                 // has no effect
            console.log("Y2");           // indeed there is an effect, this line is skipped.
        });
    });
    afterEach(function () {
        console.log("Z");
    });
    after(function () {
        console.log("ZZ");
    });
});
```
```
  outer
      inner
A
B
C
        √ test case
Y
Z
A
B
C2
        √ test case
Y
Z
ZZ

  2 passing (13ms)
```

### Description of the Change

~~Print a DeprecationWarning when `this.skip()` is used in `after`/`afterEach` hooks~~
Print a DeprecationWarning when `this.skip()` is used in `after` hooks.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -626,7 +626,7 @@ Because this test _does nothing_, it will be reported as _passing_.
 
 > _Best practice_: Don't do nothing! A test should make an assertion or use `this.skip()`.
 
-To skip _multiple_ tests in this manner, use `this.skip()` in a "before" hook:
+To skip _multiple_ tests in this manner, use `this.skip()` in a "before all" hook:
 
 ```js
 before(function() {
@@ -661,6 +661,8 @@ describe('outer', function() {
   });
 });
 ```
+
+Skipping a test within an "after all" hook is deprecated and will throw an exception in a future version of Mocha. Use a return statement or other means to abort hook execution.
 
 > Before Mocha v3.0.0, `this.skip()` was not supported in asynchronous tests and hooks.
 
@@ -910,7 +912,7 @@ Enforce a rule that tests must be written in "async" style, meaning each test pr
 
 ### `--bail, -b`
 
-Causes Mocha to stop running tests after the first test failure it encounters. Corresponding `after()` and `afterEach()` hooks are executed for potential cleanup.
+Causes Mocha to stop running tests after the first test failure it encounters. Corresponding "after each" and "after all" hooks are executed for potential cleanup.
 
 `--bail` does _not_ imply `--exit`.
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -363,9 +363,9 @@ Runner.prototype.hook = function(name, fn) {
     }
     self.currentRunnable = hook;
 
-    if (name === 'beforeAll') {
+    if (name === HOOK_TYPE_BEFORE_ALL) {
       hook.ctx.currentTest = hook.parent.tests[0];
-    } else if (name === 'afterAll') {
+    } else if (name === HOOK_TYPE_AFTER_ALL) {
       hook.ctx.currentTest = hook.parent.tests[hook.parent.tests.length - 1];
     } else {
       hook.ctx.currentTest = self.test;
@@ -388,6 +388,12 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
+          if (name === HOOK_TYPE_AFTER_ALL) {
+            utils.deprecate(
+              'Skipping a test within an "after all" hook is DEPRECATED and will throw an exception in a future version of Mocha. ' +
+                'Use a return statement or other means to abort hook execution.'
+            );
+          }
           if (name === HOOK_TYPE_BEFORE_EACH || name === HOOK_TYPE_AFTER_EACH) {
             self.test.pending = true;
           } else {
--- a/test/integration/fixtures/pending/skip-sync-after.fixture.js
+++ b/test/integration/fixtures/pending/skip-sync-after.fixture.js
@@ -1,0 +1,18 @@
+'use strict';
+
+describe('skip in after', function () {
+  it('should run this test-1', function () {});
+
+  after('should print DeprecationWarning', function () {
+    this.skip();
+    throw new Error('never throws this error');
+  });
+
+  describe('inner suite', function () {
+    it('should run this test-2', function () {});
+  });
+});
+
+describe('second suite', function () {
+  it('should run this test-3', function () {});
+});
--- a/test/integration/pending.spec.js
+++ b/test/integration/pending.spec.js
@@ -71,6 +71,28 @@ describe('pending', function() {
       });
     });
 
+    describe('in after', function() {
+      it('should run all tests', function(done) {
+        runMocha(
+          'pending/skip-sync-after.fixture.js',
+          args,
+          function(err, res) {
+            if (err) {
+              return done(err);
+            }
+            expect(res, 'to have passed').and('to satisfy', {
+              passing: 3,
+              failing: 0,
+              pending: 0,
+              output: expect.it('to contain', '"after all" hook is DEPRECATED')
+            });
+            done();
+          },
+          'pipe'
+        );
+      });
+    });
+
     describe('in before', function() {
       it('should skip all suite specs', function(done) {
         run('pending/skip-sync-before.fixture.js', args, function(err, res) {


juergba:
@boneskull please tell me wether it makes sense to continue with this PR. 
I could add two tests and a short note in the documentation.

plroebuck:
Think this should go forward. Don't see any reason to logically allow marking a test skipped after its execution.

lib/runner.js
```diff
@@ -388,6 +395,13 @@ Runner.prototype.hook = function(name, fn) {
             });
             // a pending hook won't be executed twice.
             hook.pending = true;
+            if (name === HOOK_TYPE_AFTER_ALL) {
```
plroebuck: Possible given L380?

juergba: Yes, it's possible. This statement is part of the `else` block which handles the `before` and `after` hooks.

juergba:
> I don't see any point to a `this.skip()` in these hooks. Even if you're using something like `--retries`, the `beforeEach` hook is again run, and you could conditionally skip there.

`--retries` has no effect on pending tests, even when `it.skip()` is used.

> Skipping in "after each" marks the test as pending, which seems problematic. Has the runner already emit failed/passed at this point?

Yes, the `pass` and `test end` events have already been emitted.

> This is not _quite_ correct; `this.skip()` will abort further execution of the hook (except stuff that's already in the event loop's queue).

You are right, I will recommend the `return` statement instead in the deprecation message.

lib/runner.js
plroebuck: Sorry. Misread hook name condition. But could just do the deprecation check for both `after` hook types prior to L380.
```js
       if (err instanceof Pending) {
         // Skipping after test has run is nonsensical...
         if (name === HOOK_TYPE_AFTER_EACH || name === HOOK_TYPE_AFTER_ALL) {
            utils.deprecate(
              'this.skip() in ' +
                dquote(name + '()') +
                ' hooks is DEPRECATED and will be removed from a future version of Mocha.'
            );
         }
```

test/integration/pending.spec.js
```diff
@@ -145,6 +166,27 @@ describe('pending', function() {
...
+    describe('in afterEach', function() {
+      it('should run all tests', function(done) {
+        runMocha(
+          'pending/skip-sync-afterEach.fixture.js',
+          args,
+          function(err, res) {
+            if (err) {
+              return done(err);
+            }
+            expect(res.code, 'to be', 0);
```
boneskull: you can combine this and others like it

```js
expect(res, 'to satisfy', {
  passing: 3,
  failing: 0,
  pending: 0,
  output: expect.it('to contain', "'afterEach' hook is DEPRECATED")
});

File: lib/runner.js [hunk 3]
```diff
@@ -377,6 +378,15 @@ Runner.prototype.hook = function(name, fn) {
...
+            utils.deprecate(
+              format(
+                "skipping a test within an '%s' hook is DEPRECATED and will be removed from a future version of Mocha. " +
+                  "Use 'return' instead to exit hook abortively.",
+                name
+              )
+            );
+          }
           if (name === HOOK_TYPE_BEFORE_EACH || name === HOOK_TYPE_AFTER_EACH) {
             self.test.pending = true;
```
boneskull: OK, so given the discussion here, I don't think this should be happening for "after each".  do you want to change this, and add a test?

boneskull: ...or just leave the deprecation message and change it later?  it's not clear to me what this would break.

juergba: @boneskull yes, I will change this. To be consistent ' after` should be changed also.

lib/runner.js
```diff
@@ -377,6 +378,15 @@ Runner.prototype.hook = function(name, fn) {
...
           } else {
```
juergba: this should not be done for `after` neither, right?

lib/runner.js
```diff
@@ -378,7 +379,15 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
-          if (name === HOOK_TYPE_BEFORE_EACH || name === HOOK_TYPE_AFTER_EACH) {
+          if (name === HOOK_TYPE_AFTER_EACH || name === HOOK_TYPE_AFTER_ALL) {
+            utils.deprecate(
+              format(
+                "skipping a test within an '%s' hook is DEPRECATED and will be removed from a future version of Mocha. " +
```
boneskull: 1. `name` here is one of `afterAll` or `afterEach`, which aren't intended for humans.  Replace `afterAll` with "after all" and `afterEach` with "after each"
1. instead of "will be removed from" say "will throw an exception in"
1. and "Use a return statement or other means to abort hook execution."

docs/index.md
```diff
@@ -662,6 +662,8 @@ describe('outer', function() {
 });
 ```
 
+Skipping a test within an `after/afterEach` hook is deprecated and will be removed from a future version of Mocha. Use `return` instead to exit hook abortively.
```
boneskull: Replace this with the language I mentioned in `lib/runner.js`

lib/runner.js
```diff
@@ -378,7 +379,15 @@ Runner.prototype.hook = function(name, fn) {
...
+                  "Use 'return' instead to exit hook abortively.",
+                name
+              )
+            );
+          } else if (name === HOOK_TYPE_BEFORE_EACH) {
```
boneskull: I thought the idea with the deprecation would be to just _warn_ the user and keep the logic otherwise the same.  as I said, I'm not sure what this would necessarily break.  unless we can show that there's a reproducible bug caused by the current behavior, I think I'd prefer we just warn...

lib/runner.js
juergba: to 1.: I have chosen the format `after()` and `afterEach()` which is used mostly in Mocha user documentation. I can change it otherwise.

docs/index.md
juergba: Most used format in Mocha user documentation is `after()` and `afterEach()` hook.

lib/runner.js
```diff
@@ -386,6 +386,15 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
+          if (name === HOOK_TYPE_AFTER_EACH || name === HOOK_TYPE_AFTER_ALL) {
+            utils.deprecate(
+              util.format(
+                "skipping a test within an '%s' hook is DEPRECATED and will throw an exception in a future version of Mocha. " +
+                  'Use a return statement or other means to abort hook execution.',
+                name === HOOK_TYPE_AFTER_EACH ? 'afterEach()' : 'after()'
```
plroebuck:
```js
               sQuote(name === HOOK_TYPE_AFTER_ALL ? 'after' : name)
```

plroebuck: Why don't the expected names match? Let's fix **"lib/suite.js"** to fix _this_...
* `HOOK_TYPE_AFTER_ALL: 'after'`
* `HOOK_TYPE_BEFORE_ALL: 'before'`

lib/runner.js
```diff
@@ -386,6 +386,15 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
+          if (name === HOOK_TYPE_AFTER_EACH || name === HOOK_TYPE_AFTER_ALL) {
+            utils.deprecate(
+              util.format(
+                "skipping a test within an '%s' hook is DEPRECATED and will throw an exception in a future version of Mocha. " +
```
plroebuck: No need to quote internal to format string. Extraneous space at end.

lib/runner.js
plroebuck: On the flip side, this could be viewed as a fix to an existing **bug**, no deprecation message -- just throw, and document the condition as a breaking change. This _is_ being put into the forthcoming major release after all.

plroebuck: Out of curiousity, what _is_ the LOE of another 5.x release? Could several of these deprecations be trivially backported and released? Then we could apply the changes to 6.0 as they should be without having to drag deprecated behavior around for yet another year...

lib/runner.js
plroebuck: Or:
```js
function name2HookType(name) {
  var hookTypeFor = {};
  hookTypeFor[HOOK_TYPE_AFTER_ALL] = 'after';
  hookTypeFor[HOOK_TYPE_AFTER_EACH] = HOOK_TYPE_AFTER_EACH;
  return hookTypeFor[name];
}
...
               sQuote(name2HookType(name)),
...
};


```

lib/runner.js
juergba: With a deprecation message only and no changes to the logic, we are on the safe side. Users get opportunity to give their feedback, just in case (I don't expect any). This issue is an old one, so there is no need to push quick changes.

plroebuck:
@juergba, if you do _exactly_ what the review comment says, please comment "done" and press the related "Resolve Conversation" button. If you do something else, explain that in comment and let the other reviewer handle decide about the button.

Makes it easier each time we re-review the code to tell what issues were addressed and which are still outstanding.

lib/runner.js
boneskull: please do; this should be changed across the board, because it's unique to the `bdd` interface

lib/runner.js
juergba: ok, just deprecation message, done.

lib/runner.js
juergba: ok, just deprecation message, done.

lib/runner.js
juergba: done

docs/index.md
juergba: ok, "after each" and "after all". done.

lib/runner.js
juergba: ` ' [...] of Mocha.  ' + 'Use a return [...]' `  This space is set intentionally.

internal quotes: I prefer my code style.

lib/runner.js
```diff
@@ -386,6 +386,15 @@ Runner.prototype.hook = function(name, fn) {
       }
       if (err) {
         if (err instanceof Pending) {
+          if (name === HOOK_TYPE_AFTER_EACH || name === HOOK_TYPE_AFTER_ALL) {
+            utils.deprecate(
+              util.format(
+                'skipping a test within an "%s" hook is DEPRECATED and will throw an exception in a future version of Mocha. ' +
```
plroebuck: Could you revert the quoting style change? Make the format '%s' rather than "%s".

lib/runner.js
```diff
@@ -386,6 +386,15 @@ Runner.prototype.hook = function(name, fn) {
...
+                  'Use a return statement or other means to abort hook execution.',
+                name === HOOK_TYPE_AFTER_EACH ? 'after each' : 'after all'
```
plroebuck: Are these names (terms) being used to have a generic message for all interfaces?

* TDD uses `teardown` and `suiteTeardown`
* other provided ones use `after` and `afterEach`

plroebuck:
This isn't annotated as such, but wasn't it intended for the 6.0 milestone? If so, might want to mark it as such so it gets in before they do the release this weekend.

lib/runner.js
juergba: resolved, see below.

lib/runner.js
juergba: yes, meant as generic term for all interfaces.

juergba:
I don't want to delay the release of v6.0.0 by this PR, therefore I haven't set the milestone.
Without @boneskull 's approval, it shouldn't be merged.

plroebuck:
@boneskull ping! This should get in...

juergba:
@plroebuck @boneskull @craigtaub please review.
It's a soft deprecation without any change to the program flow. So I would like to have this in v6.1. Thanks

docs/index.md
```diff
@@ -662,6 +662,8 @@ describe('outer', function() {
 });
 ```
 
+Skipping a test within an "after each" or "after all" hook is deprecated and will throw an exception in a future version of Mocha. Use a return statement or other means to abort hook execution.
```
plroebuck: Can't say I really understand the point of the second sentence here in this context.

docs/index.md
```diff
@@ -626,7 +626,7 @@ Because this test _does nothing_, it will be reported as _passing_.
```
plroebuck: Could you add the following line to the preceding codefence at L616?
```js
// antipattern
it('should only test in the correct environment', function() {
```

docs/index.md
juergba: `this.skip()` aborts the execution of the hook, subsequent code will not be executed.

plroebuck: This new behavior? My understanding was that `this.skip` marked the test as pending. Execution of the hook continues regardless as documented on L611.

juergba: There is no new behavior, it's just a soft deprecation without any change in the logic.
So the current behavior (aborting execution, marking as pending) remains.

test/integration/fixtures/pending/skip-sync-afterEach.fixture.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+describe('skip in afterEach', function () {
+  it('should run this test-1', function () {});
+  it('should run this test-2', function () {});
+
+  afterEach('should print DeprecationWarning', function () {
+    this.skip();
+    throw new Error('never throws this error');
```
plroebuck: some sort of implicit return is definitely _not_ what I would expect.

test/integration/fixtures/pending/skip-sync-after.fixture.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+describe('skip in after', function () {
+  it('should run this test-1', function () {});
+
+  after('should print DeprecationWarning', function () {
+    this.skip();
+    throw new Error('never throws this error');
```
plroebuck: same as below

test/integration/fixtures/pending/skip-sync-afterEach.fixture.js
juergba: Really? This "implicit return" was discussed here four weeks ago. It's current state and also described in the documentation. I don't understand what you want now?

juergba:
@plroebuck I have re-read this PR and your comments several times.
I don't understand what you want and why you come up with this stuff after four weeks. It's just like you have never read this PR from beginning to the end.
I rebased and forced-pushed this PR today, without any changes. Code and test all the same, since several weeks.
So what exactly are you asking for?

test/integration/fixtures/pending/skip-sync-afterEach.fixture.js
plroebuck: Pretend I review dozens of PRs here -- I lose track of what's actual and what's proposed, especially if I'm not deeply involved with that area of the code. I do recall discussing implicit return on one of your other PRs; was unaware it was implemented.

> It's current state and also described in the documentation.

Where? This in the documentation (L611):
> _Best practice_: To avoid confusion, do not execute further instructions in a test or hook after calling `this.skip()`.

That sentence _certainly_ doesn't imply an implicit return will occur. I find it horrid that `this.skip` does without using `return`.

juergba: Climb one line up:
> It's also important to note that calling this.skip() will effectively abort the test.

If you are terrified by the "implicit return", act accordingly and open your PR.
It's current state and out of scope of this PR.

plroebuck:
@juergba, guess part of this was never having read your fixture closely enough, and not realizing existing behavior of `this.skip`.

Would requested documentation changes:
* L611: Denote that invoking `this.skip` causes implicit return
* L616: Denote codefence is antipattern

Realize neither is strictly related to your code changes, so new issue acceptable.

test/integration/fixtures/pending/skip-sync-afterEach.fixture.js
plroebuck: Effectively aborting the test within a `before` to me only implies setting a flag, not implicit return (since the test has not yet begun execution).

As already stated, the implicit return behavior itself surprised me -- "terrified" is the wrong word -- "stunned" is more accurate. _Nowhere else_ in normal JavaScript coding (to my knowledge) would invoking a function cause implicit return from the caller's function.

Well aware it is outside scope of your PR here.

plroebuck: Much of today's conversation has nothing to do with your PR. You do (to my knowledge) possess the most up-to-date knowledge on hook processing, so you're the person most likely to know the answers. Sorry if my questions came off as some sort of "last-minute-redo-this-PR" thing, as that was never my intention.

boneskull:
Thinking about this further, I don't see any use case for a `this.skip()` in an "after all" hook.  There's no further test execution that can occur at that point. 
 It should be deprecated.

I'm having a tough time with skipping in "after each", however.  Consider this:

```js
describe('some filesystem code', function() {
  afterEach(function() {
    if (require('os').type() === 'FooOS') {
      this.skip();
    }
  });

  it('should modify some file', function() {
    // this is run regardless of os
  });
  
  describe('extended file attributes', function() {
    it('should set extended file attributes', function() {
      // does not work on FooOS
    });
  });

  describe('extended permissions', function() {
    it('should set extended permissions', function() {
      // does not work on FooOS
    });
  });
});
```

Functionally equivalent code using "before each" hooks requires objectively more lines of code:

```js
describe('some filesystem code', function() {
  it('should modify some file', function() {
    // this is run regardless of os
  });
  
  describe('extended file attributes', function() {
    beforeEach(function() {
      if (require('os').type() === 'FooOS') {
        this.skip();
      }
    });

    it('should set extended file attributes', function() {
      // does not work on FooOS
    });
  });

  describe('extended permissions', function() {
    beforeEach(function() {
      if (require('os').type() === 'FooOS') {
        this.skip();
      }
    });

    it('should set extended permissions', function() {
      // does not work on FooOS
    });
  });
});
```

While this can be mitigated using another nested suite, I suppose I'd rather not force users into doing this if we can help it.

What do you think?

boneskull:
(As much as we'd like to say "don't rely on the order in which tests are executed", Mocha _does_ execute tests in a certain order and IMO it's unreasonable to believe people won't write test code with that in mind)

juergba:
In short, what you are saying is:
- already passed/failed tests are no longer set to pending retroactively
- `this.skip()` affects all subsequent tests in the current suite and all subsuites
- the hook pattern is similar to the one of failing hooks, there is an important difference though. In your example above:
  - when `afterEach` throws an error, the second and third tests just disappear. They are not run and are not listed in the report. So `afterEach` runs only once, but affects all three tests.
  - with `this.skipped`, `afterEach` runs three times and decides (conditionally) at runtime wether to skip or not. The third run is senseless since there is no test to come.

juergba:
If you agree with my previous comment, I would open a new PR for `afterEach` hook.

With this PR I would like to finish `after` hook.
The failing hook pattern for `after` hook states:
> Failed `after` hook does not alter execution order

So this definition would confirm your statement:
> It should be deprecated, 100% agree.

I need to know:
1. deprecate `this.skipped()` in `after` hooks: yes or no ?
2. soft deprecation: just an deprecation warning ?
3. changes to the logic: passed/failed test should not be marked as pending ?

boneskull:
> deprecate this.skipped() in after hooks: yes or no ?

yes

> soft deprecation: just an deprecation warning ?

yes

> changes to the logic: passed/failed test should not be marked as pending ?

no; if a test has already completed execution, it should not be marked pending.  only any _future_ tests will be marked as pending.  there may not *be* any future tests, mind you.

this is as much of a limitation of Mocha's architecture as anything, because it might be useful to retroactively mark a failed test as pending.  but we can't retroactively change test statuses once they've been emitted to reporters.

juergba:
Finally I applied the following changes:
- a soft deprecation of `this.skipped()` in `after` hooks => prints a deprecation message.
- no other changes have been applied. Tests are still marked as pending retroactively, this behavior will have to be changed later, provided we don't get any negative feedback by users.

Note: `afterEach` hooks are not affected by this PR.

test/integration/pending.spec.js
```diff
@@ -71,6 +71,28 @@ describe('pending', function() {
...
+          args,
+          function(err, res) {
+            if (err) {
+              return done(err);
+            }
+            expect(res, 'to have passed').and('to satisfy', {
+              passing: 3,
+              failing: 0,
+              pending: 0,
+              output: expect.it('to contain', '"after all" hook is DEPRECATED')
```
boneskull: this can be `output: /"after all" hook is DEPRECATED/` per the semantics of "to satisfy"

PR #3722, opened 2019-02-09
Refactor `lookupFiles` and `files`

plroebuck:
### Description of the Change
Refactored code for functions `exports.lookupFiles` and `exports.files` in **"lib/utils.js"**.

### Alternate Designs
NA

NA (already is)

### Benefits
* Makes intent clearer
* Code reuse for file extension matching function
* 2x speedup using `RegExp.test` instead of `String.match`

### Possible Drawbacks
None to knowledge.

### Applicable issues
semver-patch


lib/utils.js
```diff
@@ -9,11 +9,10 @@
  * Module dependencies.
  */
 
-var debug = require('debug')('mocha:watch');
 var fs = require('fs');
-var glob = require('glob');
 var path = require('path');
-var join = path.join;
+var debug = require('debug')('mocha:watch');
```
boneskull: maybe just `mocha:utils` in case we want to use it elsewhere in this file.  again, realize you didn't do this

lib/utils.js
```diff
@@ -506,6 +503,40 @@ exports.canonicalize = function canonicalize(value, stack, typeHint) {
...
+ * Determines if pathname has a matching file extension.
+ *
+ * @private
+ * @param {string} pathname - Pathname to check for match.
+ * @param {string[]} exts - List of file extensions (sans period).
+ * @return {boolean} whether file extension matches.
+ * @example
+ * hasMatchingExtname('foo.html', ['js', 'css']); // => false
+ */
+function hasMatchingExtname(pathname, exts) {
```
boneskull: is the regexp really necessary here?  I know you just moved this code, but:

```js
function hasMatchingExtname(pathname, exts) {
  return exts.indexOf(path.extname(pathname).slice(1)) > -1;
}
```

plroebuck: Made a [JSPerf](https://jsperf.com/hasmatchingextname) for this. Had started down this route, but a different performance check showed seesawing results across different browser-version combinations.

plroebuck: Should the matching extname check have been done case-insensitively?

lib/utils.js
plroebuck: I moved it to `exports.watch`, where it was being used.
We can add another module variable if the need arises.

lib/utils.js
boneskull: I'd say no.  never seen a `.JS` file in the wild.

boneskull: either way, it's not hugely important

plroebuck:
Should the following be migrated into (maybe) **"lib/htmlUtils.js"**:
* `escape`
* `parseQuery` (used?)
* `highlight`
* `highlightTags` (used?)

PR #3723, opened 2019-02-10
fix --reporter-option to allow comma-separated options; closes #3706

### LINKED ISSUE #3706: --reporter-option should allow comma-separated values
Add `coerce: list` to the option in `lib/cli/run.js`. maybe add an integration test.  

Reference: #3564

boneskull:
adds tests for comma-separated k/v pairs and repeated `--reporter-option` usage


boneskull:
I'm going to just go ahead and merge this; I don't think it's going to be problematic.  we can always fix it later.

Issue #3604, opened 2018-12-10
reporters should be immune to interference with STDOUT


boneskull:
Potentially related: #2943

If a reporter uses `console.log()`, it should retain the function like we do with timers.

This will theoretically allow users to stub out `console.log()` in their own tests without affecting reporter output.

Since `console.log()` expects `this` to be `console`, it likely needs to take a form of:

```js
const log = console.log.bind(console);
```

`process.stdout.write()` may need to be consumed in a similar manner.  

To reproduce the problem:

```js
// yuck.spec.js
it('should do, but it do not', function() {
  // user wants to hide output generated by console.log calls in fn 'foo'
  console.log = function() {};
  foo();
});
```

Another approach would be to instantiate our own `Console` object ([docs](https://nodejs.org/api/console.html#console_class_console)) which would bypass the global `console` entirely.  We *could* consider subclassing `Console` in order to provide a nice abstraction for reporters consuming `process.stdout` directly (or other streams).

A workaround (for the user) would be to use something like proxyquire, rewiremock, etc. in lieu of replacing the method outright, like one would with Sinon.

boneskull:
If possible, ensure this is enforced by ESLint, as we do with the timers.

plroebuck:
Or more simply, provide the ability to redirect to file.

boneskull:
I think you’re conflating two separate issues. Even if any reporter could write to a built-in or user-defined stream, that stream would need to be protected from test harness interference.

craigtaub:
I like the idea of base utility to unify reporters and make them more resilient. Happy to have a go when I get time.

craigtaub:
Annoyingly our own tests rely on this functionality (all over the helpers), a large portion of them will break if we remove this.
Going to propose adding base utility with a gradual roll-out.

craigtaub:
@boneskull with so many of our reporter tests relying on interfering with `process.stdout.write` we would have to redo most of them if we did this.
What do you think if we just move the reporters using `console.log` to use `process.stdout.write` instead? Very quick change.

Worth noting `console.log` calls `process.stdout.write` internally, and `process.stdout.write` cant be stored + used as a reference like `console.log` can (as it relies on a stream) so not sure how we could do it.

boneskull:
> @boneskull with so many of our reporter tests relying on interfering with process.stdout.write we would have to redo most of them if we did this.

[...]

PR #3725, opened 2019-02-13
Base reporter store ref to console.log

craigtaub:
### Description of the Change
Small fix so that users can silence their own code (focused on just console.log's).
```js
// yuck.spec.js
it('should do, but it do not', function() {
  // user wants to hide output generated by console.log calls in fn 'foo'
  console.log = function() {};
  foo();
});
```

We can still stub its underlying mechanism. 
Decided this small fix offers enough for now. Let me know if others feel strongly that its not.

### Alternate Designs

Thinking more about how to unify the reporters (and `process.stdout.write`). 

### Applicable Issues

#3604

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -313,7 +313,7 @@ Base.prototype.epilogue = function() {
   var stats = this.stats;
   var fmt;
 
-  println();
+  this.println();
 
   // passes
   fmt =
@@ -321,26 +321,26 @@ Base.prototype.epilogue = function() {
     color('green', ' %d passing') +
     color('light', ' (%s)');
 
-  println(fmt, stats.passes || 0, milliseconds(stats.duration));
+  this.println(fmt, stats.passes || 0, milliseconds(stats.duration));
 
   // pending
   if (stats.pending) {
     fmt = color('pending', ' ') + color('pending', ' %d pending');
 
-    println(fmt, stats.pending);
+    this.println(fmt, stats.pending);
   }
 
   // failures
   if (stats.failures) {
     fmt = color('fail', '  %d failing');
 
-    println(fmt, stats.failures);
+    this.println(fmt, stats.failures);
 
     Base.list(this.failures);
-    println();
+    this.println();
   }
 
-  println();
+  this.println();
 };
 
 /**
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -313,7 +313,7 @@ Base.prototype.epilogue = function() {
   var stats = this.stats;
   var fmt;
 
-  this.println();
+  Base.println();
 
   // passes
   fmt =
@@ -321,26 +321,26 @@ Base.prototype.epilogue = function() {
     color('green', ' %d passing') +
     color('light', ' (%s)');
 
-  this.println(fmt, stats.passes || 0, milliseconds(stats.duration));
+  Base.println(fmt, stats.passes || 0, milliseconds(stats.duration));
 
   // pending
   if (stats.pending) {
     fmt = color('pending', ' ') + color('pending', ' %d pending');
 
-    this.println(fmt, stats.pending);
+    Base.println(fmt, stats.pending);
   }
 
   // failures
   if (stats.failures) {
     fmt = color('fail', '  %d failing');
 
-    this.println(fmt, stats.failures);
+    Base.println(fmt, stats.failures);
 
     Base.list(this.failures);
-    this.println();
+    Base.println();
   }
 
-  this.println();
+  Base.println();
 };
 
 /**
@@ -492,5 +492,7 @@ var objToString = Object.prototype.toString;
 function sameType(a, b) {
   return objToString.call(a) === objToString.call(b);
 }
+
+Base.println = println;
 
 Base.abstract = true;
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -28,6 +28,11 @@ exports = module.exports = Base;
 var isatty = tty.isatty(1) && tty.isatty(2);
 
 /**
+ * Save log references to avoid tests interfering (see GH-3604).
+ */
+var println = console.log;
+
+/**
  * Enable coloring by default, except in the browser interface.
  */
 
@@ -192,7 +197,7 @@ var generateDiff = (exports.generateDiff = function(actual, expected) {
  *     Error property
  */
 exports.list = function(failures) {
-  console.log();
+  println();
   failures.forEach(function(test, i) {
     // format
     var fmt =
@@ -253,7 +258,7 @@ exports.list = function(failures) {
       testTitle += str;
     });
 
-    console.log(fmt, i + 1, testTitle, msg, stack);
+    println(fmt, i + 1, testTitle, msg, stack);
   });
 };
 
@@ -308,7 +313,7 @@ Base.prototype.epilogue = function() {
   var stats = this.stats;
   var fmt;
 
-  console.log();
+  println();
 
   // passes
   fmt =
@@ -316,26 +321,26 @@ Base.prototype.epilogue = function() {
     color('green', ' %d passing') +
     color('light', ' (%s)');
 
-  console.log(fmt, stats.passes || 0, milliseconds(stats.duration));
+  println(fmt, stats.passes || 0, milliseconds(stats.duration));
 
   // pending
   if (stats.pending) {
     fmt = color('pending', ' ') + color('pending', ' %d pending');
 
-    console.log(fmt, stats.pending);
+    println(fmt, stats.pending);
   }
 
   // failures
   if (stats.failures) {
     fmt = color('fail', '  %d failing');
 
-    console.log(fmt, stats.failures);
+    println(fmt, stats.failures);
 
     Base.list(this.failures);
-    console.log();
+    println();
   }
 
-  console.log();
+  println();
 };
 
 /**
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -197,7 +197,7 @@ var generateDiff = (exports.generateDiff = function(actual, expected) {
  *     Error property
  */
 exports.list = function(failures) {
-  println();
+  Base.log();
   failures.forEach(function(test, i) {
     // format
     var fmt =
@@ -258,7 +258,7 @@ exports.list = function(failures) {
       testTitle += str;
     });
 
-    println(fmt, i + 1, testTitle, msg, stack);
+    Base.log(fmt, i + 1, testTitle, msg, stack);
   });
 };
 
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -30,7 +30,7 @@ var isatty = tty.isatty(1) && tty.isatty(2);
 /**
  * Save log references to avoid tests interfering (see GH-3604).
  */
-var println = console.log;
+var consoleLog = console.log;
 
 /**
  * Enable coloring by default, except in the browser interface.
@@ -197,7 +197,7 @@ var generateDiff = (exports.generateDiff = function(actual, expected) {
  *     Error property
  */
 exports.list = function(failures) {
-  Base.log();
+  Base.consoleLog();
   failures.forEach(function(test, i) {
     // format
     var fmt =
@@ -258,7 +258,7 @@ exports.list = function(failures) {
       testTitle += str;
     });
 
-    Base.log(fmt, i + 1, testTitle, msg, stack);
+    Base.consoleLog(fmt, i + 1, testTitle, msg, stack);
   });
 };
 
@@ -313,7 +313,7 @@ Base.prototype.epilogue = function() {
   var stats = this.stats;
   var fmt;
 
-  Base.println();
+  Base.consoleLog();
 
   // passes
   fmt =
@@ -321,26 +321,26 @@ Base.prototype.epilogue = function() {
     color('green', ' %d passing') +
     color('light', ' (%s)');
 
-  Base.println(fmt, stats.passes || 0, milliseconds(stats.duration));
+  Base.consoleLog(fmt, stats.passes || 0, milliseconds(stats.duration));
 
   // pending
   if (stats.pending) {
     fmt = color('pending', ' ') + color('pending', ' %d pending');
 
-    Base.println(fmt, stats.pending);
+    Base.consoleLog(fmt, stats.pending);
   }
 
   // failures
   if (stats.failures) {
     fmt = color('fail', '  %d failing');
 
-    Base.println(fmt, stats.failures);
+    Base.consoleLog(fmt, stats.failures);
 
     Base.list(this.failures);
-    Base.println();
+    Base.consoleLog();
   }
 
-  Base.println();
+  Base.consoleLog();
 };
 
 /**
@@ -493,6 +493,6 @@ function sameType(a, b) {
   return objToString.call(a) === objToString.call(b);
 }
 
-Base.println = println;
+Base.consoleLog = consoleLog;
 
 Base.abstract = true;
--- a/lib/reporters/dot.js
+++ b/lib/reporters/dot.js
@@ -22,6 +22,11 @@ var EVENT_RUN_END = constants.EVENT_RUN_END;
 exports = module.exports = Dot;
 
 /**
+ * Save log references to avoid tests interfering (see GH-3604).
+ */
+var println = console.log;
+
+/**
  * Constructs a new `Dot` reporter instance.
  *
  * @public
@@ -68,7 +73,7 @@ function Dot(runner, options) {
   });
 
   runner.once(EVENT_RUN_END, function() {
-    console.log();
+    println();
     self.epilogue();
   });
 }
--- a/lib/reporters/dot.js
+++ b/lib/reporters/dot.js
@@ -73,7 +73,7 @@ function Dot(runner, options) {
   });
 
   runner.once(EVENT_RUN_END, function() {
-    println();
+    self.println();
     self.epilogue();
   });
 }
--- a/lib/reporters/dot.js
+++ b/lib/reporters/dot.js
@@ -73,7 +73,7 @@ function Dot(runner, options) {
   });
 
   runner.once(EVENT_RUN_END, function() {
-    self.println();
+    Base.println();
     self.epilogue();
   });
 }
--- a/lib/reporters/dot.js
+++ b/lib/reporters/dot.js
@@ -22,11 +22,6 @@ var EVENT_RUN_END = constants.EVENT_RUN_END;
 exports = module.exports = Dot;
 
 /**
- * Save log references to avoid tests interfering (see GH-3604).
- */
-var println = console.log;
-
-/**
  * Constructs a new `Dot` reporter instance.
  *
  * @public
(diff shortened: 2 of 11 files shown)


lib/reporters/base.js
```diff
@@ -27,6 +27,11 @@ exports = module.exports = Base;
 
 var isatty = tty.isatty(1) && tty.isatty(2);
 
+/**
+ * Save log references to avoid tests interfering (see GH-3604).
+ */
+var println = console.log;
```
boneskull: how about `log`

lib/reporters/base.js
```diff
@@ -276,6 +281,7 @@ function Base(runner) {
   }
   this.stats = runner.stats; // assigned so Reporters keep a closer reference
   this.runner = runner;
+  this.println = println;
```
boneskull: does this need to be an instance prop?  should it be a static prop?  if it was the latter, reporters wouldn't have to extend `Base` to use it.  the less `this`/`self` the better, imo

test/reporters/base.spec.js
```diff
@@ -403,4 +403,20 @@ describe('Base reporter', function() {
...
   });
+
+  it('should let you stub out console.log without effecting reporters output', function() {
+    var logCached = console.log;
+    var logRes = [];
+
+    console.log = function(line) {
+      logRes.push(line);
+    };
+    var base = new Base({on: function() {}, stats: {duration: 0, passes: 3}});
```
boneskull: I think you can just use `Base.list(stuff)` instead of instantiating `Base` and calling `epilogue()`...

another (better?) way to test it would be to just use sinon; stub `console.log` then check that it was not called.

```js
// untested
sinon.stub(console, 'log');
Base.log('something');
expect(console.log, 'was not called');
```

lib/reporters/base.js
boneskull: also, I'm not sure if `console.log` needs a context in all environments we support.  in other words, if `var log = console.log; log('foo')` works, we're ok.  

I know it's fine to do this in Node.js and chromium, but not sure about IE11 or other browsers.  `console` is completely nonstandard afaik, so we have to be careful.

boneskull: if we *do* discover this is not supported in something, then the fix would be

```js
var log = console.log.bind(console);
```

lib/reporters/base.js
craigtaub: Nope better as static. Done.

test/reporters/base.spec.js
craigtaub: Nice, done.

lib/reporters/base.js
craigtaub: I branched directly from upstream so that the PR would run the full Karma features. I also tested it locally on IE11 and it seemed to work as expected. So I think its ok.

plroebuck: Why not `consoleLog` rather than `log` (which is used for _lots_ of things)?

test/reporters/base.spec.js
```diff
@@ -403,4 +404,10 @@ describe('Base reporter', function() {
     errOut = stdout.join('\n').trim();
     expect(errOut, 'to be', '1) test title:\n     Error\n  foo\n  bar');
   });
+
+  it('should let you stub out console.log without effecting reporters output', function() {
+    sinon.stub(console, 'log');
+    Base.list([]);
+    expect(console.log, 'was not called');
+  });
```
plroebuck: Shouldn't this be done twice, once normally expecting "to call", then apply your hideaway and expect "not to call"?

test/reporters/xunit.spec.js
```diff
@@ -194,15 +194,13 @@ describe('XUnit reporter', function() {
...
-        var cachedConsoleLog = console.log;
-        console.log = function(string) {
-          stdout.push(string);
-        };
 
+        var cachedPrintln = reporters.Base.log;
+        reporters.Base.log = stdout.push.bind(stdout);
         var xunit = new XUnit({on: function() {}, once: function() {}});
         xunit.write.call({fileStream: false}, expectedLine);
+        reporters.Base.log = cachedPrintln;
```
plroebuck: Don't see why any changes should be made here... the PR was addressing user-related squelching of output, not modifying unrelated tests.

cachedPrintln?

test/reporters/base.spec.js
craigtaub: Dont quite follow the "to call" scenario. Stub console.log before importing `Base` to prove that it is called?

test/reporters/xunit.spec.js
craigtaub: So the test relies on controlling console.log. With my base changes it fails.
Do you see a better way of doing this?

Will rename `cachedConsoleLog`.

test/reporters/base.spec.js
craigtaub: @plroebuck sorry still don't quite follow what you meant above, could you explain please.

lib/reporters/dot.js
```diff
@@ -67,7 +67,7 @@ function Dot(runner) {
   });
 
   runner.once(EVENT_RUN_END, function() {
-    console.log();
+    Base.consoleLog();
```
plroebuck: Pointless use of `console.log` to begin with...
```
process.stdout.write('\n');
```

lib/reporters/progress.js
```diff
@@ -91,7 +91,7 @@ function Progress(runner, options) {
   // and the failures if any
   runner.once(EVENT_RUN_END, function() {
     cursor.show();
-    console.log();
+    Base.consoleLog();
```
plroebuck: again, pointless.
```js
process.stdout.write('\n');
```

test/reporters/base.spec.js
plroebuck: This test uses unproven assumption that `Base.list` implementation uses `console.log` to begin with. In order to prove your test stubs it out, you have to prove the opposite as well.

```js
 it('should let you stub out console.log without effecting reporters output', () => {
    sinon.stub(console, 'log');
    sinon.stub(Base, 'consoleLog');
    Base.list([]);
    expect(Base.consoleLog, 'was called');
    expect(console.log, 'was not called');
  });
```

test/reporters/xunit.spec.js
```diff
@@ -194,15 +194,13 @@ describe('XUnit reporter', function() {
       it('should call write with line', function() {
         stdoutWrite = process;
         process = false; // eslint-disable-line no-native-reassign, no-global-assign
```
plroebuck: Mentioned elsewhere, but this test should fail after `process` reassignment...

test/reporters/xunit.spec.js
```diff
@@ -194,15 +194,13 @@ describe('XUnit reporter', function() {
...
-        var cachedConsoleLog = console.log;
-        console.log = function(string) {
-          stdout.push(string);
-        };
 
+        var cachedConsoleLog = reporters.Base.consoleLog;
```
plroebuck: nit: this would seem better named as `savedConsoleLog` -- not what I think about when I see the word "cached"...

test/reporters/xunit.spec.js
craigtaub: Is there a change needed to this test to make that happen? Dont follow why `should fail after process reassignment` would be true.

test/reporters/base.spec.js
craigtaub: Yep..done.

test/reporters/xunit.spec.js
craigtaub: @plroebuck any chance you could help me with this please.

plroebuck: https://github.com/mochajs/mocha/pull/2698#issuecomment-467413307

craigtaub:
```
node bin/mocha "test/reporters/xunit.spec.js"
```

Nope, locally passes for me on master and this branch.

craigtaub: Is your concern running function on a falsy value? Seems as its re-assigned before the test finishes its back for `EVENT_RUN_END`.

lib/reporters/base.js
```diff
@@ -27,6 +27,11 @@ exports = module.exports = Base;
 
 var isatty = tty.isatty(1) && tty.isatty(2);
 
+/**
+ * Save log references to avoid tests interfering (see GH-3604).
+ */
+var consoleLog = console.log;
```
boneskull: IMO the "more correct" thing to do is to modify the eslint configuration to handle `console.*` the same way it handles timers (unless that's not desirable, for some reason?)

don't think we need to block on that, though; we can change it later.

lib/reporters/dot.js
```diff
@@ -67,7 +67,7 @@ function Dot(runner) {
   });
 
   runner.once(EVENT_RUN_END, function() {
-    console.log();
+    process.stdout.write('\n');
```
boneskull: maybe we should leave this as-is until comes a time when we do this same PR against `process.stdout` and `process.stderr`.  what do you think?

test/reporters/base.spec.js
```diff
@@ -403,4 +404,16 @@ describe('Base reporter', function() {
     errOut = stdout.join('\n').trim();
     expect(errOut, 'to be', '1) test title:\n     Error\n  foo\n  bar');
   });
+
+  it('should let you stub out console.log without effecting reporters output', function() {
+    sinon.stub(console, 'log');
```
boneskull: 1. a sinon sandbox is typically easier to use, because no matter what you do, it's only one `restore()` call:

	```js
	var sandbox;
	var myStub;
	var myOtherStub;
	
	beforeEach(function() {
	  sandbox = sinon.createSandbox();
	  myStub = sandbox.stub();
	  myOtherStub = sandbox.stub();
	})
	
	afterEach(function() {
	  sandbox.restore();
	});
	```

2. if you call `restore()` in the test itself after making assertions, the call won't happen if the assertions fail; use hooks instead

test/reporters/xunit.spec.js
```diff
@@ -194,15 +194,13 @@ describe('XUnit reporter', function() {
       it('should call write with line', function() {
         stdoutWrite = process;
         process = false; // eslint-disable-line no-native-reassign, no-global-assign
-        var cachedConsoleLog = console.log;
-        console.log = function(string) {
-          stdout.push(string);
-        };
 
+        var savedConsoleLog = reporters.Base.consoleLog;
```
boneskull: as above, stubbing & un-stubbing should happen in hooks

lib/reporters/dot.js
boneskull: (I'm thinking we probably shouldn't do that)

craigtaub: It was flagged as a pointless use of console.log to begin with and i agreed. Feels a good idea to have the reporters consistent while im here.

lib/reporters/base.js
craigtaub: Had a quick look at this. The main issue being that outside the reporters we use `console` (log and error) a fair amount. The solution was always focused on fixing the issue with reporters. We could have a rule just for reporters (seems brittle), or we can completely this solution to follow the timers practice, but AFAIK there has been no indication we actually need this.

lib/reporters/dot.js
boneskull: IMO console.log should be used unless it can’t be. it does some queuing and stream-related stuff that prevents the chance for the output to be lost before the process ends. not a big deal right here since the epilogue will be output after.

lib/reporters/base.js
boneskull: the rules that exist for timers are also specific to certain files, so there’s a precedent for this anyway. again, not a blocker.

test/reporters/base.spec.js
plroebuck:
```js
  describe('when reporter output immune to user test changes', function() {
    var sandbox;

    beforeEach(function() {
      sandbox = sinon.createSandbox();
      sinon.stub(console, 'log');
      sinon.stub(Base, 'consoleLog');
    });

    it('should let you stub out console.log without effecting reporters output', function() {
      Base.list([]);

      expect(Base.consoleLog, 'was called');
      expect(console.log, 'was not called');
    });

    afterEach(function() {
      sandbox.restore();
    });
  });
```

lib/reporters/dot.js
plroebuck: Can't find what you're in reference to [here](https://github.com/nodejs/node/blob/5e98f875b93b7556458a6a274b09a36875207d9d/lib/internal/console/constructor.js#L288).

Would appear to actually be a speed disadvantage using `console.log` for simple strings. For more complex strings, the format routine can be `require`'d and used, same as `console.log` does. Outside of use in browser, I find little use for `console.log` at all...

test/reporters/base.spec.js
craigtaub: Nice with the sandboxes. But if I don't restore the `Base.consoleLog` before the test ends it breaks the reporter.

test/reporters/xunit.spec.js
craigtaub: Similar issue to above, if `reporters.Base.consoleLog` is not back by end of test it breaks the reporter running the test. Any ideas?

lib/reporters/base.js
craigtaub: Ive added the lint rule now.

test/reporters/base.spec.js
plroebuck: Why?

craigtaub: I guess as its interfering with running which is running the test.

> node bin/mocha "test/reporters/base.spec.js"

Stub inside hooks, restore inside test
```
  Base reporter
    when reporter output immune to user test changes
      ✓ should let you stub out console.log without effecting reporters output

  1 passing (17ms)
```

Stub + restore inside hooks
```
  Base reporter
    when reporter output immune to user test changes

  1 passing (17ms)
```

Thoughts?

test/reporters/xunit.spec.js
craigtaub: That isnt going to be possible. Please see above for reason, except this cant `callThrough()`.

craigtaub:
Going to get this in as don't think anything outstanding.

test/reporters/xunit.spec.js
```diff
@@ -194,15 +194,13 @@ describe('XUnit reporter', function() {
...
+        reporters.Base.consoleLog = stdout.push.bind(stdout);
         var xunit = new XUnit({on: function() {}, once: function() {}});
         xunit.write.call({fileStream: false}, expectedLine);
+        reporters.Base.consoleLog = savedConsoleLog;
 
-        console.log = cachedConsoleLog;
         process = stdoutWrite; // eslint-disable-line no-native-reassign, no-global-assign
         expect(stdout[0], 'to be', expectedLine);
       });
```
plroebuck: This is what I'm using...
Achieves equivalent test without messing with `process` object and produces no "noise" output...

```js
    describe('when output directed to console', function() {
      var xunit;

      it("should call 'console.log' with line", function() {
        xunit = new XUnit(runner);

        var fakeThis = {preferConsole: true};
        Object.setPrototypeOf(fakeThis, XUnit.prototype);
        var consoleLog = sinon.stub(console, 'log');
        xunit.writeln.call(fakeThis, expectedLine);
        consoleLog.restore();

        expect(consoleLog.calledWith(expectedLine), 'to be true');
      });
    });
```

plroebuck: Of course the user could have used this same technique rendering this issue unnecessary.

test/reporters/xunit.spec.js
plroebuck: See below

test/reporters/base.spec.js
plroebuck: Edited sandbox code above.

plroebuck:
> I think though, before we merge it, we need to be certain that this will be more beneficial than harmful.
>
> the main benefit is that tests can stub their own use of console.log in the code under test without impacting reporter output. while the use-case is not unheard of, I suspect anyone dealing with this currently just lives with it.

While I won't block it, I don't think we should merge this (nothing to do with Craig's implementation). There are existing schemes for accomplishing the issue's goal, no changes required. Would prefer to simply add wiki page demonstrating _how_ to correctly stub `process.stdout.write` and `console.log`...

craigtaub:
> Would prefer to simply add wiki page demonstrating how to correctly stub process.stdout.write and console.log...

That is fine for us to do, but I do feel most users who do this will initially try the technique in the description, so having immediate successful feedback seems like a benefit for us and users.

> There are existing schemes for accomplishing the issue's goal, no changes required.

I have seen discussions but nothing concrete or PR'd yet. Let me know if thats not the case. 
If there is something close to being complete i am happy to let that take over this. But I had not seen anything and thought a half-measure now is better than waiting indefinitely for a larger fix.

plroebuck:
> > Would prefer to simply add wiki page demonstrating how to correctly stub process.stdout.write and console.log...
> 
> That is fine for us to do, but I do feel most users who do this will initially try the technique in the description, so having immediate successful feedback seems like a benefit for us and users.

Problem is that you likely gave them a false positive outcome though; very likely they've now screwed up their test setup if they don't reset correctly. Then their unrelated next test fails instead...

> > There are existing schemes for accomplishing the issue's goal, no changes required.
> 
> I have seen discussions but nothing concrete or PR'd yet. Let me know if that's not the case.
> If there is something close to being complete i am happy to let that take over this. But I had not seen anything and thought a half-measure now is better than waiting indefinitely for a larger fix.

No. I meant "no changes to Mocha required" existing schemes, like [this](https://github.com/mochajs/mocha/pull/3725/files#r275112753).

test/reporters/xunit.spec.js
craigtaub: But the logic inside `xunit` reporter requires `process` to have been tampered with doesnt it? 
The assertion is on the logging but the test setup will have to mess with `process` wont it?

```javascript
  if (this.fileStream) {
    // ... not this
  } else if (typeof process === 'object' && process.stdout) {
    // ... not this
  } else {
   // ... want to call this
  }
```

Happy to be corrected here if ive missed something.

plroebuck: You're right, of course.

I'm too far ahead of myself here; I adjusted that part of the code to make XUnit more easily configurable (and testable)...

plroebuck: But the scheme for stubbing `console.log` (assuming XUnit instance was thus configured) is demonstrated.

craigtaub:
> Problem is that you likely gave them a false positive outcome though; very likely they've now screwed up their test setup if they don't reset correctly

If they dont reset correctly im not sure thats our fault/problem in any way. If they do reset correctly they will benefit from this (specially if reset in hooks).

craigtaub:
UPDATE: going to leave this here for now while we continue the work to unify the reporters (i.e. https://github.com/mochajs/mocha/pull/3874)

craigtaub:
As wider attempts to fix this are taking some time, I feel this small addition is worth adding for now.

craigtaub:
Have added a small section to the Wiki https://github.com/mochajs/mocha/wiki/HOW-TO:-Correctly-stub-stdout

PR #3726, opened 2019-02-13
Add ability to unload files from `require` cache (redux)

plroebuck:
### Description of the Change
Adds a feature which allows unloading test files from `require` cache programmatically.
This feature simplifies rerunning tests since there is no need to recreate the Mocha instance.

Currently handled by one of these options:
1. The user has to implement same logic in his own code.
1. The user has to create a new Mocha instance before each run.

### Alternative Designs
A potentially more complete design could take a snapshot of the cache before testing begins and remove _everything_ added afterwards. Prefer to see how well this implementation does before going further.

1. Because it allows reuse of object instances.
1. Provides missing half of Mocha's public API referenced [here](https://github.com/mochajs/mocha/blob/d1902a20663a6766ffaae646f155a884fcf3e141/lib/mocha.js#L579-L585).
1. (Hopeful) reduction of user questions about how to do this.

### Benefits
Better reusability, fewer headaches for programmers.

### Possible Drawbacks
Still possible that even though registered files were unloaded from the `require` cache,
they may have in turn `require`d something else which _will_ remain in the cache.

### Applicable issues
Closes #3534 (original PR from Oct 2018 redone against current master)
semver-minor

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -329,10 +329,11 @@ Mocha.prototype.loadFiles = function(fn) {
  * Removes a previously loaded file from Node's `require` cache.
  *
  * @private
+ * @static
  * @see {@link Mocha#unloadFiles}
  * @param {string} file - Pathname of file to be unloaded.
  */
-Mocha.prototype.unloadFile = function(file) {
+Mocha.unloadFile = function(file) {
   delete require.cache[require.resolve(file)];
 };
 
@@ -343,17 +344,17 @@ Mocha.prototype.unloadFile = function(file) {
  * This allows files to be "freshly" reloaded, providing the ability
  * to reuse a Mocha instance programmatically.
  *
- * <strong>Not needed for (or used by) CLI.</strong>
+ * <strong>Intended for consumers &mdash; not used internally</strong>
  *
  * @public
- * @see {@link Mocha#unloadFile}
+ * @see {@link Mocha.unloadFile}
  * @see {@link Mocha#loadFiles}
  * @see {@link Mocha#run}
  * @returns {Mocha} this
  * @chainable
  */
 Mocha.prototype.unloadFiles = function() {
-  this.files.forEach(this.unloadFile, this);
+  this.files.forEach(Mocha.unloadFile);
   return this;
 };
 
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -301,7 +301,6 @@ Mocha.prototype.ui = function(name) {
 };
 
 /**
- * @summary
  * Loads `files` prior to execution.
  *
  * @description
@@ -310,6 +309,8 @@ Mocha.prototype.ui = function(name) {
  *
  * @private
  * @see {@link Mocha#addFile}
+ * @see {@link Mocha#run}
+ * @see {@link Mocha#unloadFiles}
  * @param {Function} [fn] - Callback invoked upon completion.
  */
 Mocha.prototype.loadFiles = function(fn) {
@@ -322,6 +323,38 @@ Mocha.prototype.loadFiles = function(fn) {
     suite.emit(EVENT_FILE_POST_REQUIRE, global, file, self);
   });
   fn && fn();
+};
+
+/**
+ * Removes a previously loaded file from Node's `require` cache.
+ *
+ * @private
+ * @see {@link Mocha#unloadFiles}
+ * @param {string} file - Pathname of file to be unloaded.
+ */
+Mocha.prototype.unloadFile = function(file) {
+  delete require.cache[require.resolve(file)];
+};
+
+/**
+ * Unloads `files` from Node's `require` cache.
+ *
+ * @description
+ * This allows files to be "freshly" reloaded, providing the ability
+ * to reuse a Mocha instance programmatically.
+ *
+ * <strong>Not needed for (or used by) CLI.</strong>
+ *
+ * @public
+ * @see {@link Mocha#unloadFile}
+ * @see {@link Mocha#loadFiles}
+ * @see {@link Mocha#run}
+ * @returns {Mocha} this
+ * @chainable
+ */
+Mocha.prototype.unloadFiles = function() {
+  this.files.forEach(this.unloadFile, this);
+  return this;
 };
 
 /**
@@ -741,8 +774,7 @@ Object.defineProperty(Mocha.prototype, 'version', {
  */
 
 /**
- * @summary
- * Runs tests and invokes `fn()` when complete.
+ * Runs root suite and invokes `fn()` when complete.
  *
  * @description
  * To run tests multiple times (or to run tests in files that are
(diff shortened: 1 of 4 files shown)


plroebuck:
Original PR was approved by both of you. Please give another quick look for any issues with remerged version.

lib/mocha.js
```diff
@@ -324,6 +325,38 @@ Mocha.prototype.loadFiles = function(fn) {
...
 };
 
+/**
+ * Removes a previously loaded file from Node's `require` cache.
+ *
+ * @private
+ * @see {@link Mocha#unloadFiles}
+ * @param {string} file - Pathname of file to be unloaded.
+ */
+Mocha.prototype.unloadFile = function(file) {
```
boneskull: this can be static (`Mocha.unloadFile`)

lib/mocha.js
```diff
@@ -324,6 +325,38 @@ Mocha.prototype.loadFiles = function(fn) {
...
+};
+
+/**
+ * Unloads `files` from Node's `require` cache.
+ *
+ * @description
+ * This allows files to be "freshly" reloaded, providing the ability
+ * to reuse a Mocha instance programmatically.
+ *
+ * <strong>Not needed for (or used by) CLI.</strong>
```
boneskull: you may want to qualify this as "intended for consumers" / "not used internally"

test/unit/mocha.spec.js
```diff
@@ -226,6 +226,58 @@ describe('Mocha', function() {
...
   });
 
+  describe('.unloadFile()', function() {
+    before(function() {
+      if (process.browser) {
+        this.skip();
+      }
+    });
+
+    it('should unload a specific file from cache', function() {
```
boneskull: this test isn't very...unit-y.  we can remove all of this stuff up thru the check on L244 by just calling `require()` on some file, since the method under test here is `unloadFile()`--not `Mocha#addFile` nor `Mocha#loadFiles`

alternatively, just move the tests out of `unit`

lib/mocha.js
plroebuck: Fine. Converted to static.

test/unit/mocha.spec.js
plroebuck: Migrated tests to "node-unit".

lib/mocha.js
```diff
@@ -324,6 +325,39 @@ Mocha.prototype.loadFiles = function(fn) {
...
+ *
+ * @public
+ * @see {@link Mocha.unloadFile}
+ * @see {@link Mocha#loadFiles}
+ * @see {@link Mocha#run}
+ * @returns {Mocha} this
+ * @chainable
+ */
+Mocha.prototype.unloadFiles = function() {
+  this.files.forEach(Mocha.unloadFile);
```
boneskull: 👍

test/unit/mocha.spec.js
```diff
@@ -224,6 +235,11 @@ describe('Mocha', function() {
       mocha.bail();
       expect(mocha.suite._bail, 'to be', true);
     });
+
+    it('should be chainable', function() {
```
boneskull: this looks like just adding some extra coverage around the chainable methods?  unrelated to `unloadFile`, yes?

test/node-unit/mocha.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    path.join(__dirname, 'cli', 'config.spec.js'),
+    path.join(__dirname, 'cli', 'run.spec.js')
+  ];
+  const resolvedTestFiles = testFiles.map(require.resolve);
+
+  describe('#addFile', function() {
+    it('should add the given file to the files array', function() {
+      const mocha = new Mocha(opts);
+      mocha.addFile(__filename);
+      expect(mocha.files, 'to have length', 1).and('to contain', __filename);
```
boneskull: :+1:

test/node-unit/mocha.spec.js
```diff
@@ -0,0 +1,72 @@
...
+      const mocha = new Mocha(opts);
+
+      testFiles.forEach(mocha.addFile, mocha);
+      mocha.loadFiles();
+      expect(require.cache, 'to have keys', resolvedTestFiles);
+    });
+
+    it('should execute the optional callback if given', function() {
+      const mocha = new Mocha(opts);
+      expect(cb => {
```
boneskull: I don't think I've ever used "to call the callback" before (as I usually delegate this stuff to sinon), so it's cool to learn about this.  but... would maybe be better to be more specific about whether the callback is called with an error or not? 
 "to call the callback without error" looks like the syntax.

test/node-unit/mocha.spec.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  describe('#unloadFiles', function() {
+    it('should unload all test files from cache', function() {
+      const mocha = new Mocha(opts);
+
+      testFiles.forEach(mocha.addFile, mocha);
+      mocha.loadFiles();
+      mocha.unloadFiles();
+      expect(require.cache, 'not to have keys', resolvedTestFiles);
```
boneskull: can you just do `require(__filename)` instead here, like in L49?

test/unit/mocha.spec.js
plroebuck: Correct -- just noticed several chainables were missing their tests.

test/node-unit/mocha.spec.js
plroebuck: It's a no-arg callback (Function.length === 0), so not really relevant in this case.

```js
Mocha.prototype.loadFiles = function(fn) {
  var self = this;
  var suite = this.suite;
  this.files.forEach(function(file) {
    file = path.resolve(file);
    suite.emit(EVENT_FILE_PRE_REQUIRE, global, file, self);
    suite.emit(EVENT_FILE_REQUIRE, require(file), file, self);
    suite.emit(EVENT_FILE_POST_REQUIRE, global, file, self);
  });
  fn && fn();
};
```

test/node-unit/mocha.spec.js
plroebuck: It was [your idea](https://github.com/mochajs/mocha/pull/3388#discussion_r193554473) to test against multiple files, a good one. This _is_ somewhat how I would expect it to be used in the wild, so think this test should reflect expected usage black-box style, since the other was done as a white-box test.

plroebuck:
Geez, what a tremendous amount of time (original from May 2018) and rework for 2 functions consisting of 7 SLOC...

test/unit/mocha.spec.js
```diff
@@ -195,11 +190,17 @@ describe('Mocha', function() {
   });
 
   describe('.noHighlighting()', function() {
```
plroebuck: So, should this test be here -- or in "browser-unit"?

PR #3731, opened 2019-02-15
CSS Smooth scrolling on fragment navigation

Munter:
### Requirements

### Description of the Change

Add CSS smooth scrolling to documentation page. This is a progressive enhancement.

https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior
https://caniuse.com/#feat=css-scroll-behavior

### Benefits

This will possibly make it easier to understand what's happening when you navigate page internal fragment identifiers

### Possible Drawbacks

Might be annoying. We'll have to try it out

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -1,11 +1,11 @@
 html {
   font: 16px/1.6 'Helvetica Neue', Helvetica, Arial, sans-serif;
+  scroll-behavior: smooth;
 }
 
 body {
   color: #2c2c2c;
   border-top: 2px solid #ddd;
-  scroll-behavior: smooth;
 }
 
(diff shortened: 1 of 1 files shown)


Munter:
Looks like I read the documentation wrong. Doesn't seem to have any effect on the full browser viewport, only smaller boxes with explicit scrolling behavior :'(

Munter:
There we go. I applied it to the wrong tag

Munter:
Having tried it on the deployment preview I'm not a big fan of this technique for our use case

boneskull:
@Munter Agree.  The page is just too long.  I'll close this...

PR #3735, opened 2019-02-16
Edit Hook.prototype.error to Hook.prototype.getError

sapzape:
### Description of the Change
Change error function to getError because the function to set error is not used.
There is only one place to use Hook.prototype.error. ([/lib/runner.js](https://github.com/mochajs/mocha/blob/master/lib/runner.js))
```
hook.run(function(err) {
  var testError = hook.error();
....
```
Using 'get' only, it seems good to remove 'set' code.

### Benefits
Unnecessary code is removed.

### DIFF
--- a/lib/hook.js
+++ b/lib/hook.js
@@ -28,19 +28,14 @@ function Hook(title, fn) {
 inherits(Hook, Runnable);
 
 /**
- * Get or set the test `err`.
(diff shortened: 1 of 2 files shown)


sapzape:
Sorry for late reply. I agree with you. I close.

Issue #3740, opened 2019-02-18
this.skip() - bug summary


juergba:
#### 1. `this.skip()` in `beforeEach` hooks
Incorrect hook pattern.
All `beforeEach` hooks are executed, but `afterEach` hooks are skipped completely, see #2546
=> fixed: #3741

#### 2. async `this.skip()` in `beforeEach` hooks
results in a "Uncaught async skip; aborting execution", see [test](https://github.com/mochajs/mocha/blob/master/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js)
`before` hooks seem to work async though, and as per user documentation it should work.
=> fixed: #3745 

#### 3. `this.skip()` in `it` tests
does not work correctly, since `beforeEach` hooks are executed, but `afterEach` hooks do not run.
=> fixed: #3859

#### 4. `this.skip()` in `afterEach` hooks
Incorrect, since already passed/failed tests are set to `pending` retrospectively. No tests are skipped at all.
=> definition of use case still pending

#### 5. `this.skip()` in `after` hooks
Incorrect, since already passed/failed tests are set to `pending` retrospectively. No tests are skipped at all.
=> soft deprecation: #3719
=> disallow usage: #4136

#### 6. `this.skip()` with `--allow-uncaught`
Incorrect, since silently aborts execution after first `this.skip()`.
=>fixed: #4030 

### Versions
current master

mrstux:
Would really like to see this.skip() from an AsyncStatus test result in the afterEach being called. 

In our test suite, the this.skip() is being called from a function invoked by the test which checks prerequisites, and if they fail, it calls skip(), which as you know, throws an exception which aborts the outer function. 

Except the cleanup hook doesn’t get called, and should.

marmik18:
i am facing issue with `this.skip()` in `it` tests. 
the test does not skip.

JoshuaKGoldberg:
Closing as this just points to other issues/PRs.

PR #3741, opened 2019-02-18
Fix hook pattern of this.skip() in beforeEach hooks

### LINKED ISSUE #2546: Test case skip()ed, beforeEach() hook still executed, afterEach() missed
This with Mocha 3.1.2. Source test file:

``` javascript
"use strict";
describe('outer', function() {

  beforeEach(function() {
    console.log("A");
    this.skip();
  });

    beforeEach(function() {
      console.log("B");
    });

    it('test case', function() {
      console.log("C");
    });

    afterEach(function() {
      console.log("Y");
    });
  });

  afterEach(function() {
    console.log("Z");
  });
});
```

Expected output: either "B" **and** "Y" should be printed, or neither "B" **nor** "Y" should be printed.

Actual output: "B" is printed, but not "Y":

```
  outer
    inner
A
B
      - test case


  0 passing (11ms)
  1 pending
```

In general,

* If I call `skip()`, then I would not expect that inner `beforeEach` blocks would be executed.
* If a `beforeEach` block is executed, then the corresponding `afterEach` blocks should also be executed, for symmetry.

juergba:
### Description

#3745 is required for this PR.

The current hook pattern is incorrect. Skipping a test within a `beforeEach` hook executes all inner `beforeEach` hooks, but skips `afterEach` hooks completely.

Some of the `afterEach`  hooks should run to ensure a correct cleanup. This same pattern has already been implemented with failing hooks and also while using the `--bail` option.

The correct hook pattern should be:
- `beforeEach` runs for each test individually and determines at runtime wether to skip or not.
(The failing hook pattern is different: the first hook failure skips all linked tests.)
- since `this.skip()` aborts hook execution, subsequent inner `beforeEach` should be skipped ("B").
- skipped `beforeEach` => skip corresponding `afterEach` hooks as well ("Y").
- `afterEach` hooks of the same/higher suite level than the skipping one should run ("Z").

```js
describe('outer', function() {
  beforeEach(function() {
    console.log("A");              // yes, does run
    this.skip();
  });
    beforeEach(function() {
      console.log("B");            // yes, does run  => should not run
    });
    it('test case', function() {
      console.log("C");            // pending
    });
    afterEach(function() {
      console.log("Y");            // no, does not run
    });
  });
  afterEach(function() {
    console.log("Z");              // no, does not run  => should run
  });
});
```

### Description of the Change

We abort `hookDown()` at skipping `suite` level in order to skip subsequent inner `beforeEach` hooks. The same skipping `suite` level is used as entry point for `hookUp()` to skip corresponding inner `afterEach` hooks.

### Applicable issues

closes #2546
closes #2623 PR
#3740 partially

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -315,8 +315,7 @@ Runner.prototype.fail = function(test, err) {
  * - Failed `before each` hook skips remaining tests in a
  *   suite and jumps to corresponding `after each` hook,
  *   which is run only once
- * - Failed `after` hook does not alter
- *   execution order
+ * - Failed `after` hook does not alter execution order
  * - Failed `after each` hook skips remaining tests in a
  *   suite and subsuites, but executes other `after each`
  *   hooks
@@ -386,7 +385,7 @@ Runner.prototype.hook = function(name, fn) {
       if (testError) {
         self.fail(self.test, testError);
       }
-      // conditional this.skip()
+      // conditional skip
       if (hook.pending) {
         if (name === HOOK_TYPE_AFTER_ALL) {
           utils.deprecate(
@@ -394,11 +393,20 @@ Runner.prototype.hook = function(name, fn) {
               'Use a return statement or other means to abort hook execution.'
           );
         }
-        if (name === HOOK_TYPE_BEFORE_EACH || name === HOOK_TYPE_AFTER_EACH) {
+        if (name === HOOK_TYPE_AFTER_EACH) {
+          // TODO define and implement use case
           if (self.test) {
             self.test.pending = true;
           }
+        } else if (name === HOOK_TYPE_BEFORE_EACH) {
+          if (self.test) {
+            self.test.pending = true;
+          }
+          self.emit(constants.EVENT_HOOK_END, hook);
+          hook.pending = false; // activates hook for next test
+          return fn(new Error('abort hookDown'));
         } else {
+          // TODO throw error for afterAll
           suite.tests.forEach(function(test) {
             test.pending = true;
           });
@@ -619,6 +627,7 @@ Runner.prototype.runTests = function(suite, fn) {
       return;
     }
 
+    // static skip, no hooks are executed
     if (test.isPending()) {
       if (self.forbidPending) {
         test.isPending = alwaysFalse;
@@ -634,6 +643,7 @@ Runner.prototype.runTests = function(suite, fn) {
     // execute test and hook(s)
     self.emit(constants.EVENT_TEST_BEGIN, (self.test = test));
     self.hookDown(HOOK_TYPE_BEFORE_EACH, function(err, errSuite) {
+      // conditional skip within beforeEach
       if (test.isPending()) {
         if (self.forbidPending) {
           test.isPending = alwaysFalse;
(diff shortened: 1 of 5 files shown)


plroebuck:
Curious on your thoughts -- outside scope of this PR. Think there's any value to adding a `type` field to `Pending` where values would be one of: `sync`, `async`, or `promise`? You're more experienced with that aspect of the code.

juergba:
Yes, I was thinking about extending `Pending` with additional properties, like error code or the hook causing the pending-error. I don't know yet, I'm still working on it.
I'm worried about the async version of this.skip(), it has not been working before, now getting worse as a mixture of uncaught error and multiple done calls. (The async skip() already includes a done call.)
Promises? I have just seen `sync` and `async` so far. I will have a look.

lib/runner.js
```diff
@@ -634,7 +640,12 @@ Runner.prototype.runTests = function(suite, fn) {
           self.emit(constants.EVENT_TEST_PENDING, test);
         }
         self.emit(constants.EVENT_TEST_END, test);
-        return next();
+        if (err instanceof Pending) {
```
juergba: I'm not sure wether this if() check is really needed.

lib/runner.js
```diff
@@ -634,7 +640,12 @@ Runner.prototype.runTests = function(suite, fn) {
...
+          self.suite = errSuite;
```
juergba: This line is unnice, but the same technique is used already above in line549.
IMO, by just changing one property the runner gets inconsistent.

juergba:
@boneskull please tell me wether it makes sense to continue with this PR. Thanks.

boneskull:
@juergba 	I think what you're proposing sounds good.
Would also like to hear your thoughts about whether this can land in a patch release.

boneskull:
I've been ignoring this because of the "WIP" in the title.  If it's ready to review, please let me know.

juergba:
Patch release is risky. IMO it certainly is a bug fix, but there have been very few (one?) opened issue about this topic. Either users don't use `this.skip()` frequently or they are happy with the current hook pattern. I suggest at least minor release.

Anyway there are more hook pattern fixes to come, for `this.skip()` in  `it()` and `afterEach()`.

Please review, it works, but I need to be confirmed with one or two doubts.

boneskull:
I think this needs to be semver-major then

dasilvacontin:
I'm thoroughly reviewing this one. It's taking me a bit due to having stepped away from the code for a while, and because our vars / function names are not very friendly (in runner.js, at least).

mrstux:
FWIW, this is a major issue for us. We are trying to find a solution, but other than forking mocha and fixing the bug, we're not sure what to do.

We have beforeEach functions which acquire resources, and we have afterEach(done) functions which release them. The tests are predicated on the assumption that if a beforeEach runs, and there is an afterEach, then it will be run, whether the test passes, fails or is skipped.

End of story.

If the afterEach is not run because a test is skipped via this.skip(), then the resources will not be cleaned up, and tests will eventually cascade fail.

I know this is not what this exact PR is about, but you were wondering why you haven't had any other issues reported. Most likely because there is a very good issue, which documents the exact issue, and has a number of PRs in progress. Its only when you dig into the PR that you find some debate about when or if to actually release the fix.

Yes, this will possibly be a "breaking fix", and yes, you may want to update the major semver. your call :)

But please fix!

juergba:
@mrstux It would be awesome if you could do some test runs with this branch.

@boneskull @craigtaub @outsideris 
I will merge this branch within the next two/three weeks. I would like to have this PR published with v7.0.0.

lib/runner.js
juergba: It's not needed, reviewed by myself, outdated anyway.

lib/runner.js
juergba: I improved this part, it's the same pattern used by failing `beforeEach` hooks. 
Reviewed by myself.

juergba:
@HarshaNalluru I would highly appreciate some feedback on this PR. Could you do some test runs with this branch?

HarshaNalluru:
> afterEach hooks of the same/higher suite level than the skipping one should run ("Z").

This .skip() behaviour seems to be more appropriate. 
Just tested this branch, works as expected. Thanks a lot for fixing this!! 🧡

juergba:
@HarshaNalluru thank you! 👍

PR #3745, opened 2019-02-19
Fix this.skip() async calls

juergba:
### Description

The async call of `this.skip` in a `beforeEach` hook results in an error.

```js
'use strict';

describe('skip in beforeEach', function () {
  beforeEach(function (done) {
    var self = this;
    setTimeout(function () {
      self.skip();
      done();
    }, 0);
  });

  it('should skip this test', function () {
    throw new Error('never run this test');
  });
  it('should not reach this test', function () {
    throw new Error('never run this test');
  });
  it('should not reach this test', function () {
    throw new Error('never run this test');
  });
});
```
```
skip in beforeEach
    - should skip this test
    1) "before each" hook

  0 passing (19ms)
  1 pending
  1 failing

  1) skip in beforeEach
       "before each" hook:
     Uncaught async skip; aborting execution
```
The async `this.skip` does not abort the hook/test (as the sync version does). This is just usual behavior since neither `done()`, `done(err)` nor "resolving/rejecting" a promise do abort an async running function.
To enforce this abortion Mocha throws an uncaught error. This is a very crude mechanisme which sometimes works and othertimes doesn't.

The user documentation states:
> Best practice: To avoid confusion, do not execute further instructions in a test or hook after calling this.skip().

I propose to remove this uncaught error. In order to abort a running async function, in some cases `this.skip()` has to be replaced by `return this.skip()`.

Consequences for the user:
- `this.skip()` works in async `beforeEach` hooks.
- "Bad practice" users have to insert the `return` statement in `before` hooks and `it` tests to abort execution.
- an explicit call of `done()` will result in an `Error: done() called multiple times`.

### Description of the Change

~~Remove uncaught error in "runnable.js".~~
Fix Runner.prototype.uncaught()

### Applicable issues

#3740 partially

### DIFF
--- a/test/integration/pending.spec.js
+++ b/test/integration/pending.spec.js
@@ -258,18 +258,21 @@ describe('pending', function() {
 
     describe('in beforeEach', function() {
       it('should skip all suite specs', function(done) {
-        run('pending/skip-async-beforeEach.fixture.js', args, function(
-          err,
-          res
-        ) {
+        var fixture = 'pending/skip-async-beforeEach.fixture.js';
+        run(fixture, args, function(err, res) {
           if (err) {
-            done(err);
-            return;
+            return done(err);
           }
-          assert.strictEqual(res.stats.pending, 1);
-          assert.strictEqual(res.stats.passes, 0);
-          assert.strictEqual(res.stats.failures, 1);
-          assert.strictEqual(res.code, 1);
+          expect(res, 'to have passed')
+            .and('to have passed test count', 0)
+            .and('to have pending test count', 3)
+            .and(
+              'to have pending test order',
+              'should skip this test-1',
+              'should skip this test-2',
+              'should skip this test-3'
+            )
+            .and('to have failed test count', 0);
           done();
         });
       });
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -781,6 +781,9 @@ Runner.prototype.runSuite = function(suite, fn) {
  * @private
  */
 Runner.prototype.uncaught = function(err) {
+  if (err instanceof Pending) {
+    return;
+  }
   if (err) {
     debug('uncaught exception %O', err);
   } else {
--- a/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-hooks.fixture.js
@@ -14,8 +14,7 @@ describe('outer suite', function () {
       console.log('inner before');
       var self = this;
       setTimeout(function () {
-        self.skip();
-        done();
+        self.skip();   // done() is not required
       }, 0);
     });
 
--- a/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before-nested.fixture.js
@@ -2,8 +2,7 @@ describe('skip in before with nested describes', function () {
   before(function (done) {
     var self = this;
     setTimeout(function () {
-      self.skip();
-      done();
+      self.skip();   // done() is not required
     }, 0);
   });
 
--- a/test/integration/fixtures/pending/skip-async-before.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-before.fixture.js
@@ -7,7 +7,7 @@ describe('outer describe', function () {
     before(function (done) {
       var self = this;
       setTimeout(function () {
-        self.skip();
+        self.skip();   // done() is not required
       }, 0);
     });
 
--- a/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js
@@ -4,18 +4,17 @@ describe('skip in beforeEach', function () {
   beforeEach(function (done) {
     var self = this;
     setTimeout(function () {
-      self.skip();
-      done();
+      self.skip();   // done() is not required
     }, 0);
   });
 
-  it('should skip this test', function () {
+  it('should skip this test-1', function () {
     throw new Error('never run this test');
   });
-  it('should not reach this test', function () {
+  it('should skip this test-2', function () {
     throw new Error('never run this test');
   });
-  it('should not reach this test', function () {
+  it('should skip this test-3', function () {
     throw new Error('never run this test');
   });
 });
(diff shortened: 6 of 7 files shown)


juergba:
The async `this.skip` does not abort the hook/test, therefore a subsequent call to `done()` leads to:
`Error: done() called multiple times`.

```js
beforeEach(function (done) {
    var self = this;
    setTimeout(function () {
      self.skip();
      // done(new Pending('async skip call'));
      // done(new Error('async skip call'));
      // done();
      console.log('should not print');
    }, 0);
```

Replacing `self.skip()` by one of the three options above does not prevent `console.log(...)` from beeing executed.

juergba:
A normal `done()` call obviously does not abort the hook/test, a `return done()` has to be used instead.

So a logical solution for my problem would be:
```js
beforeEach(function (done) {
    var self = this;
    setTimeout(function () {
      return self.skip();
      console.log('should not print');    // does not run
    }, 0);
```
Could this solution be acceptable?
Or: how do I change the async `skip()` function to get the same result?

juergba:
`Promise`s are working correctly before/after this PR.
```js
beforeEach(function () {
  console.log('A');
  return new Promise(function(res, rej) {
    setTimeout(function() {
      rej();                      // does not abort here
      console.log('bad practice?');
    }, 0);
  }).catch(()=>this.skip());
});
```

boneskull:
@juergba these changes (and your other open PRs) demand close review.  I'll have time to look at them perhaps next week.  in the meanwhile, I'd really appreciate if @craigtaub 	and/or @plroebuck 	could take a look.  didn't want you to think I was ignoring you, but.. I guess I'm ignoring you. until next week. :wink:

craigtaub:
Code looks good.

Not sure I've seen a case requiring both done and return w/o actually calling done, for async (does it exist?). That could confuse people. 
But if it works (unlike currently) and we can update the docs to explain it, then 🤷‍♂️.

Interested to hear what others think.

juergba:
> Not sure I've seen a case requiring both done and return w/o actually calling done [...]

No, you don't need calling `done()`, this call is already included in the async version of `this.skip()`. The example below with an async `before` hook did work before this PR, since the `done()` instruction had never been reached. Because you can't properly abort an async function dynamically at runtime in ES5 (in ES6 maybe with Generator Iterator) Mocha throws an "uncaught Error".
```js
before(function (done) {
    var self = this;
    setTimeout(function () {
      self.skip();  // before this PR: abort here
                    // after this PR: no abort, function keeps going
      done();  // before: worked, since never reached
               // after: "Error: done() called multiple times", since "this.skip() already calls "done()"
    }, 0);
  });
```
After this PR user will get an error running above example. He has two options:
- delete `done()` (see best practice above)
- `return self.skip();`

craigtaub:
> Not sure I've seen a case requiring both done and return w/o actually calling done

Sorry what I meant here was done in function argument.

juergba:
> That could confuse people. Could `this.skip(done)` work (sounds like it cant)? [...]

Yes, this works without aborting. For the user it would be clearer this way .

craigtaub:
> Yes, this works without aborting. For the user it would be clearer this way .

Ok great, yes it definitely think so too.

plroebuck:
I know you've done a _lot_ of really hard research on this, but I fail to see the need for attempting `this.skip()` asynchronously. Reminds me of [that old joke](https://pics.me.me/me-as-a-doctor-it-hurts-when-i-do-this-22316076.png)...

Why isn't the solution to point to the documentation?
Still open, but feel it really needs a convincing business case...

juergba:
The documentation says: 
> Before Mocha v3.0.0, this.skip() was not supported in asynchronous tests and hooks.

My interpretation is, that the async call of `this.skip` should work. But it doesn't. Our [fixture](https://github.com/mochajs/mocha/blob/master/test/integration/fixtures/pending/skip-async-beforeEach.fixture.js) is failing and the "spec" is just pretending everything is ok.

So I'm trying to fix a bug, because I need this PR for another pending PR of mine.
A use case could be: a db access failing asyncly in a "hook" => skip test on runtime.

lib/runnable.js
```diff
@@ -343,11 +343,7 @@ Runnable.prototype.run = function(fn) {
 
     // allows skip() to be used in an explicit async context
     this.skip = function asyncSkip() {
-      done(new Pending('async skip call'));
-      // halt execution.  the Runnable will be marked pending
```
boneskull: per these comments, the intent was that the uncaught handler will ignore the thrown `Pending`.

for whatever reason (probably my fault), the uncaught handler is not ignoring the thrown `Pending`.

The proper fix then, IMO, is to modify `Runner#uncaught` instead:

```js
Runner.prototype.uncaught = function(err) {
  if (err instanceof Pending) {
    return;
  }
  /* etc */
}
```

test/integration/pending.spec.js
```diff
@@ -258,18 +258,21 @@ describe('pending', function() {
...
           if (err) {
-            done(err);
-            return;
+            return done(err);
           }
-          assert.strictEqual(res.stats.pending, 1);
-          assert.strictEqual(res.stats.passes, 0);
-          assert.strictEqual(res.stats.failures, 1);
-          assert.strictEqual(res.code, 1);
+          expect(res, 'to have passed')
```
boneskull: This is good, yes, I agree that the spec was not making the correct assertions.

lib/runnable.js
boneskull: it also appears we need this on 390

```js
            if (self.test) {
              self.test.pending = true;
            }
```

juergba: First part: done

Second part: IMO should not be necessary.

`self.test` is set when the `test` event is emitted, before any `beforeEach` hook.
The event order is:
- test start    <- `self.test` is set
- hook `beforeEach` start
- hook `beforeEach` end
- test end

juergba:
@craigtaub @plroebuck @boneskull please review.

Finally the changes I applied:
- the uncaught Error is thrown as in the past
- Runner.prototype.uncaught() is fixed in order to swallow this `Pending` error.
- cheating spec fixed
- `before` and `beforeEach` hooks both do work
- nothing changes for the user

lib/runnable.js
boneskull: I'm not entirely sure how I got an error to be thrown there where `self.test` is `undefined`, but I'll see if I can reproduce the situation.

PR #3754, opened 2019-02-20
Revert default glob to match Mocha-5.2

### LINKED ISSUE #3750: test.js in root no longer works with v6.0.0
### Prerequisites

### Description

v6.0.0 seems not work for small projects without `test/` folder and with `test.js` in root instead/

### Steps to Reproduce

- project: https://github.com/markdown-it/uc.micro
- greenkeeper PR and Travis-CI log: https://github.com/markdown-it/uc.micro/pull/11

Everything should work as before

```
$ npm test
> uc.micro@1.0.6 test /home/travis/build/markdown-it/uc.micro
> mocha
Error: No test files found: "test/"
The command "npm test" exited with 1.
```

100%

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 6.0.0
- The output of `node --version`: v10.11.0
- The version and architecture of your operating system: Ubuntu 18.04
- Any other third party Mocha related modules (with versions):
- The code transpiler being used:

### Additional Information

I could not find any info that such feature was deprecated. Let me know if i missed something.

plroebuck:
### Description of the Change

The old glob 'test' would allow "test.js" to be used by default (rather than requiring spec file be
placed inside "test" directory).

In minimalist repos, users were simply using "test.js" rather than "test/\<file\>.js", allowing...
```sh
$ mocha
```

### Alternate Designs
NA

NA - already there

### Benefits
Restore compatibility with previous releases.
Prevent a boatload of more issues from minimalist repositories using Mocha.

### Possible Drawbacks
None

### Applicable issues

Fixes #3750
semver-patch

### DIFF
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -245,7 +245,7 @@ exports.builder = yargs =>
       }
     })
     .positional('spec', {
-      default: ['test/'],
+      default: ['test'],
       description: 'One or more files, directories, or globs to test',
       type: 'array'
     })


plroebuck:
This is just the patch itself. Do we _want_ to document this "feature", or simply treat it as a
"laissez-faire" possibility?

boneskull:
no, whatever, it's fine.

PR #3758, opened 2019-02-21
Add new cmdline option to denote implicit returns

### LINKED ISSUE #3744: CoffeeScript tests after v6.0.0 and suite return value check
### Prerequisites

- [x ] Checked that your issue hasn't already been filed by cross-referencing [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20)
- [x ] Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.
- [x ] 'Smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, your usage of Mocha, or Mocha itself
- [x ] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend avoiding the use of globally installed Mocha.

### Description

Issue #3243 shipped with v6.0.0 throw a deprecation warning or an error if `--throw-deprecation` is enabled with all CoffeeScript tests since the language support implicit returns.

### Steps to Reproduce

The following minimalist test case:

```coffee
describe 'My Suite', () ->
  it 'run a test', () ->
    console.log 'yes'
```

Throw the error "DeprecationWarning: Suites ignore return values. Suite "My Suite" in /package/test/my/test.coffee returned a value; this may be a bug in your test code"

A temporary workaround include returning `undefined`:

```coffee
describe 'system.authconfig', () ->
  it 'fuck around', () ->
    console.log 'yes'
  undefined
```

### Versions

- The output of `npx mocha --version` : 6.0.0
- The code transpiler being used: CoffeeScript
- Other information not relevant

### Additional Information

I understand the motivation behind #3243. One way to please everyone is to detect if the return value found in the suite is not the same as the one returned by `it`.

plroebuck:
### Description of the Change
Coffeescript automatically returns the result of the last expression (implicit return).
The change made to deprecate suites returning a value causes a warning for _every_ `describe`. This adds a cmdline option which can disable the deprecation warning.

### Alternate Designs
Simplest and most expedient solution.

NA - already in core

### Benefits
Allows CoffeeScript users to upgrade.

### Possible Drawbacks
None

### Applicable issues
Fixes #3744
Original PR #3243
semver-patch

### DIFF
--- a/lib/cli/run-option-metadata.js
+++ b/lib/cli/run-option-metadata.js
@@ -35,6 +35,7 @@ exports.types = {
     'forbid-pending',
     'full-trace',
     'growl',
+    'implicit-returns',
     'inline-diffs',
     'interfaces',
     'invert',
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -143,6 +143,10 @@ exports.builder = yargs =>
         description: 'Enable Growl notifications',
         group: GROUPS.OUTPUT
       },
+      'implicit-returns': {
+        description: 'Denote source language uses implicit returns',
+        group: GROUPS.RULES
+      },
       'inline-diffs': {
         description:
           'Display actual/expected differences inline within each string',
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -1,9 +1,11 @@
 'use strict';
 
+var util = require('util');
 var Suite = require('../suite');
 var utils = require('../utils');
 var errors = require('../errors');
 var createMissingArgumentError = errors.createMissingArgumentError;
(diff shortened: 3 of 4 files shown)


plroebuck:
This is quick patch only. Not married to option name, not sure which group it belongs in either.
But it does fix the problem.

boneskull:
recommend this be closed

PR #3759, opened 2019-02-21
backout deprecation of value returned from suite; closes #3744

### LINKED ISSUE #3744: CoffeeScript tests after v6.0.0 and suite return value check

boneskull:
This just kills the deprecation as I suggested [here](https://github.com/mochajs/mocha/issues/3744#issuecomment-465853188).

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var Suite = require('../suite');
-var utils = require('../utils');
 var errors = require('../errors');
 var createMissingArgumentError = errors.createMissingArgumentError;
 
@@ -138,16 +137,7 @@ module.exports = function(suites, context, mocha) {
           }
         }
         if (typeof opts.fn === 'function') {
-          var result = opts.fn.call(suite);
-          if (typeof result !== 'undefined') {
-            utils.deprecate(
-              'Suites ignore return values. Suite "' +
-                suite.fullTitle() +
-                '" in ' +
-                suite.file +
-                ' returned a value; this may be a bug in your test code'
-            );
-          }
+          opts.fn.call(suite);
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
           throw createMissingArgumentError(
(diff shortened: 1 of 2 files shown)


plroebuck:
Consider an alternative version, using something like
```js
function isAsync(fn) {
   return fn.constructor.name === 'AsyncFunction';
}
```
on the `describe`'s function and issuing the deprecation warning based on that.

boneskull:
@plroebuck 	we can revisit after this patch release.  I *would* like to just solve async suites.

PR #3760, opened 2019-02-21
Add support for configuration files with ".jsonc" extension

### LINKED ISSUE #3753: config: default support of .jsonc extension
By default, Mocha only searches for `.mocharc.json`.  

However, Mocha supports "JSON with comments".  In VSCode anyway, these files can have a `jsonc` extension.  Mocha should also look for `.mocharc.jsonc`.

(Note that Mocha *does not* support [JSON5](https://json5.org/) and currently has no plans to do so. Future support for this *may* make sense, but AFAIK JSON5 is not widely used; it would also require a different parsing strategy.)

To do:

1. Make changes in `lib/cli/config.js`
1. Update/add test in `test/integration/config.spec.js`

sstephant:
### Description of the Change
Adds support for ".mocharc.jsonc" as an alternative name to imply JSON with comments.

### Alternate Designs
Add JSON5 support.

Minor enhancement of existing capability already in core.

### Benefits
Support for ".jsonc" file extension

### Possible Drawbacks
None

### Applicable issues
Fixes #3753
semver-minor


sstephant:
Note: I suppose the user documentation should be updated too to reflect the changes but I'm not sure where to do it.

sstephant:
> 
> 
> Note: I suppose the user documentation should be updated too to reflect the changes but I'm not sure where to do it.

I think I found the appropriate location and updated the user documentation in the last commit.

plroebuck:
Color me silly, but my read on the original issue was simply to add another file extension to the set of Mocha configuration files... where did JSON5 come in?

.gitignore
```diff
@@ -79,6 +79,7 @@ Session.vim
 out/
 .idea_modules/
 atlassian-ide-plugin.xml
+mocha.iml
```
plroebuck: what is this?

package.json
```diff
@@ -411,6 +411,7 @@
     "Sulabh Bista <sul4bh@gmail.com>",
     "Sune Simonsen <sune@we-knowhow.dk>",
     "Svetlana <39729453+Lana-Light@users.noreply.github.com>",
+    "Sylvain Stéphant <sstephant+github@gmail.com>",
```
plroebuck: Remove this change. There's a script that handles this...

docs/index.md
```diff
@@ -1628,6 +1628,7 @@ In addition to supporting the legacy [`mocha.opts`](#mochaopts) run-control form
 - **JavaScript**: Create a `.mocharc.js` in your project's root directory, and export an object (`module.exports = {/* ... */}`) containing your configuration.
 - **YAML**: Create a `.mocharc.yaml` (or `.mocharc.yml`) in your project's root directory.
 - **JSON**: Create a `.mocharc.json` in your project's root directory. Comments &mdash; while not valid JSON &mdash; are allowed in this file, and will be ignored by Mocha.
+- **[JSON5](https://json5.org/)**: Create a `.mocharc.jsonc` (or `.mocharc.json5`) in your project's root directory.
```
plroebuck: Add the URL to the bottom of this file, not inline...

.gitignore
boneskull: that's a JetBrains-related file. but when I used it, it wanted to put them in `.idea`.  

@sstephant I think this is a non-standard location for the `.iml` file?  If so, maybe better to add it to your _global_ gitignore file.

docs/index.md
boneskull: @sstephant Please pull out the JSON5-related stuff.  Mocha shouldn't support it unless people explicitly ask for it.  Just "JSON with comments" and `.jsonc`, please.

package.json
boneskull: unless it matches *exactly*, the name will be duplicated.

.gitignore
sstephant: Yes it's Jetbrains related. I'm using IntelliJ IDEA which generates this in addition to the .idea folder (unlike PhpStorm for example). I can remove it from the .gitignore.

package.json
sstephant: Ok I did not know this.

sstephant:
> 
> 
> Color me silly, but my read on the original issue was simply to add another file extension to the set of Mocha configuration files... where did JSON5 come in?

🤦‍♂️. This is my mistake, I read the issue too fast. I understood JSON with comments support had to be added for files with a .jsonc extension. It's so unusual to see actual JSON files with comments I did not realize mocha was already stripping comments from JSON config files. In my opinion the main purpose of JSON5 is precisely the support of comments in JSON/JS like documents, that's why I introduced the json5 parser.

lib/cli/config.js
```diff
@@ -24,6 +24,7 @@ exports.CONFIG_FILES = [
   '.mocharc.js',
   '.mocharc.yaml',
   '.mocharc.yml',
+  '.mocharc.jsonc',
   '.mocharc.json'
 ];
```
plroebuck: Somewhat related -- asking for opinions here... not changes atm.

Think this is better?
```js
exports.loadConfig = filepath => {
  let config = {};
  const ext = path.extname(filepath);
  try {
    if (/\.ya?ml/.test(ext)) {
      config = parsers.yaml(filepath);
    } else if (ext === '.js') {
      config = parsers.js(filepath);
    } else if (/\.jsonc?/.test(ext)) {
    } else {
      throw new Error('no parser for this type file');
    }
  } catch (err) {
    throw new Error(`failed to parse "${filepath}": ${err}`);
  }
  return config;
};
```
Rationale: However stupid, if I used `--config` with a Word document argument, the code would throw a `SyntaxError` (as result of failing `JSON.parse`) which is utterly unhelpful and its error message doesn't
 address the actual problem.

test/node-unit/cli/config.spec.js
```diff
@@ -57,6 +57,17 @@ describe('cli/config', function() {
         });
       });
 
+      describe('when supplied a filepath with .jsonc extension', function() {
```
plroebuck: It would be nice if all suite titles double-quoted the file extension
(e.g., `when supplied a filepath with ".jsonc" extension`)

plroebuck:
> It's so unusual to see actual JSON files with comments I did not realize mocha was already stripping comments from JSON config files.

Yeah, wasn't crazy about using JSON files containing comments (since they won't parse naturally without stripping). Actually prefer the new file extension ".jsonc" since it at least implies something different.

example/config/.mocharc.jsonc
```diff
@@ -0,0 +1,14 @@
+// This config file contains Mocha's defaults.
+// As you can see, comments are allowed.
+// This same configuration could be provided in the `mocha` property of your
+// project's `package.json`.
+{
+  diff: true,
```
outsideris: No double quotes are allowed in jsonc?

I didn't know about JSONC.

sstephant: Good point this is a leftover of JSON5 support introduced earlier bur then dropped.
The double quotes are actually required with the JSON parser. I'll fix the example.

lib/cli/config.js
sstephant: I agree the fallback to the JSON parser can lead to an unhelpful error message.
But in your snippet the inner throw would be hidden by the upper catch was that your intention?
Another point I noticed is that the file extension is case sensitive. Was that on purpose?

plroebuck: Threw from `try` just to pass message to `catch` where it can generate higher-level `Error`.
Dislike the thought of it, but guess could throw `String` to pass along less baggage...

plroebuck: > Another point I noticed is that the file extension is case sensitive. Was that on purpose?

We support lowercase file extensions.

sstephant:
> 
>
> Yeah, wasn't crazy about using JSON files containing comments (since they won't parse naturally without stripping). Actually prefer the new file extension ".jsonc" since it at least implies something different.

Does that mean comment stripping should be disabled on ".json" file from now on? I didn't do it because it's a breaking change.

plroebuck:
> Does that mean comment stripping should be disabled on ".json" file from now on? I didn't do it because it's a breaking change.

No. Too late now... Pandora's box already opened.

I would be okay with reversing these as below, but not required:
```
- **JSON**: Create a `.mocharc.jsonc` (or `.mocharc.json`) in your ...
```

plroebuck:
@boneskull ping. LGTM

Alhadis:
You guys might wanna consider exposing a list of supported file extensions (or even just a predicate function which tells if a filepath is a recognised config file).

This can matter for editor plugins when a subset of files are passed to Mocha to prevent it scanning the entire workspace (or if implementation details prevent them giving Mocha open-slather access to the filesystem).

Case-in-point, I very recently upgraded [Atom-Mocha](https://github.com/Alhadis/Atom-Mocha) to use Mocha 6's new configs feature, and had to work off the existing kruft of predetermined filenames. I can already see this happening in Emacs too...)

boneskull:
@alhadis Why does an editor plugin need to scan the entire workspace?

Feel free to send a PR that exposes this information in a convenient manner.

Alhadis:
@boneskull In Atom's case, the editor's test-runner API is invoked with a list of paths, which may be directories or files. There's no guarantee they'll fall within the same path, and a translation layer is needed to call Mocha with the paths provided by Atom.

(It's all pretty convoluted, but the long-and-short of it all is that it isn't as simple as just throwing a bunch of paths at Mocha and expecting it to do The Right Thing™...)

Will send a PR soon-ish. :)

PR #3764, opened 2019-02-22
Display website sponsor logos on non-white background

### LINKED ISSUE #3736: make sponsors logo more readable when logo is white in homepage
In [homepage](https://mochajs.org/), some sponsors's logo is white. So, people can't realize who is support us. 
We should make them to people see the logo well like background color or shadow.

Migarve55:
### Description of the Change

I have changed the website style sheet to change the links style of the "Principal" and "TripleByte" sponsors. This way, they can be seen against a blurred grey background. 

### Alternate Designs

N/A

N/A

### Benefits

Now there is a grey background in the third and fourth sponsor's logo, so now they can be seen.

### Possible Drawbacks

None

### Applicable issues

Fixes #3736 
semver-patch

### DIFF
--- a/docs/_includes/sponsors.md
+++ b/docs/_includes/sponsors.md
@@ -4,6 +4,6 @@ Use Mocha at Work?  Ask your manager or marketing team if they'd help [support](
 
 <!-- markdownlint-disable MD034 -->
 {% for i in (0..15) %}[![](https://opencollective.com/mochajs/sponsor/{{ i }}/avatar.jpg)](https://opencollective.com/mochajs/sponsor/{{ i }}/website){: target="_blank"} {% endfor %}
-{: .image-list .faded-images}
+{: .image-list .faded-images id="_sponsors"}
 
 <script src="js/avatars.js"></script>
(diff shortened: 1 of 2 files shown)


Migarve55:
The way I see it there are only two possibilities:

- Change the logos to make them visible. I don't know if this is possible since they are stored in the "opencollective" website.
- Change the website style as I did, another idea would be to change the background where the logos are displayed. This will work for all logos, but maybe it affects the desired aesthetic.

I will try the second approach.

Migarve55:
I have tried this, but it does not look very good.

image

Migarve55:
The problem is that each logo is different, and the logos are supplied by an external source. I am not sure if a general solution is possible.

Munter:
I really think this is open collective that should handle the guarantee that a logo will always be visible no matter the background

Migarve55:
I believe you are right, the only solution would be to contact the open collective webpage.

boneskull:
agree with the general assessment from the others here.  thanks anyway though!

PR #3766, opened 2019-02-24
Make reporter ctors support optional 'options' parameter

plroebuck:
### Description of the Change
Change allows for passing settings to _any_ reporter.
The same `options` parameter is passed to abstract `Base` superctor.

Fixed/standardized some of JSDoc as well.

### Alternate Designs
Only means to pass settings to reporters.

NA

### Benefits
Immediately, none. Prerequisite step to further upgrades.

### Possible Drawbacks
None, since changes only allow possibility of future changes.

### Applicable issues
semver-patch

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -92,10 +92,10 @@ if (process.platform === 'win32') {
  * as well as user-defined color
  * schemes.
  *
+ * @private
  * @param {string} type
  * @param {string} str
  * @return {string}
- * @private
  */
 var color = (exports.color = function(type, str) {
   if (!exports.useColors) {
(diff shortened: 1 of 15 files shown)


plroebuck:
Intend to merge this if no comments by tomorrow -- not much here to contest...

craigtaub:
Change looks fine. 
Could you explain more the plans for "passing settings to reporters" please? (mainly benefit to user).

plroebuck:
> Could you explain more the plans for "passing settings to reporters" please?

You mean past the obvious "currently not possible" to pass settings for reporters except `progress`, `tap`, and `xunit`. Further, if some type of global handling were desired, without being able to push that handling into `base`, each reporter would have to reimplement it themselves (not DRY).

This PR was only to put low-level support there without fanfare -- didn't think there would be any argument about this.

PR #3767, opened 2019-02-25
fix handling of bareword args matching node flags; closes #3761

### LINKED ISSUE #3761: cli option `--require trace-something` causes `bad option`-error
### Prerequisites

### Description

When executing mocha with the option `--require trace-something`, the error `bad-option --trace-something` is output. In my case. it is `trace-and-clarify-if-possible`, but it is reproducible with other package names.

This happens in `mocha@6.0.0` and `mocha@6.0.1`, but not in `mocha@6.0.0-1`, and only if the required package-name starts with "trace-"

### Steps to Reproduce

```
# Setup dependency with empty index.js-file
mkdir trace-dependency
cd trace-dependency/
npm init -y
touch index.js
cd ..

# Setup project and add dependency
mkdir test-project
cd test-project
npm init -y
npm install ../trace-dependency
mkdir test
touch test/dash-spec.js
```

Then install mocha and run the tests:

```
npm install mocha@6.0.0
npx mocha --require trace-dependency
# Error "bad option --trace-dependency"

npm install mocha@6.0.1
npx mocha --require trace-dependency
# Error "bad option --trace-dependency"

npm install mocha@6.0.0-1
npx mocha --require trace-dependency
# No error, "0 passing"
```

I would have expected the same result for each mocha-version, i.e. the output "0 passing".

* `mocha@6.0.0` and `mocha@6.0.1` have the output "bad option --trace-dependency"
* `mocha@6.0.0-1` has the expected output

100%

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 6.0.0, 6.0.1
- The output of `node --version`: v10.15.1
- The version and architecture of your operating system: Xubuntu 18.04
- Any other third party Mocha related modules (with versions): none
- The code transpiler being used: none

boneskull:
maybe not the most elegant solution, but it seems to work.

cc @nknapp @janpot

### DIFF
--- a/lib/cli/node-flags.js
+++ b/lib/cli/node-flags.js
@@ -29,16 +29,32 @@ const debugFlags = new Set(['debug', 'debug-brk', 'inspect', 'inspect-brk']);
  *   - `--v8-*` (but *not* `--v8-options`)
  * @summary Whether or not to pass a flag along to the `node` executable.
  * @param {string} flag - Flag to test
- * @returns {boolean}
+ * @param {boolean} [bareword=true] - If `false`, we expect `flag` to have one or two leading dashes.
+ * @returns {boolean} If the flag is considered a "Node" flag.
  * @private
  */
-exports.isNodeFlag = flag =>
-  !/^(?:require|r)$/.test(flag) &&
-  (nodeFlags.has(flag) ||
-    debugFlags.has(flag) ||
-    /(?:preserve-symlinks(?:-main)?|harmony(?:[_-]|$)|(?:trace[_-].+$)|gc(?:[_-]global)?$|es[_-]staging$|use[_-]strict$|v8[_-](?!options).+?$)/.test(
-      flag
-    ));
+exports.isNodeFlag = (flag, bareword = true) => {
+  if (!bareword) {
+    // check if the flag begins with dashes; if not, not a node flag.
+    if (!/^--?/.test(flag)) {
+      return false;
+    }
+    // strip the leading dashes to match against subsequent checks
+    flag = flag.replace(/^--?/, '');
+  }
+  return (
+    // treat --require/-r as Mocha flag even though it's also a node flag
+    !(flag === 'require' || flag === 'r') &&
+    // check actual node flags from `process.allowedNodeEnvironmentFlags`,
+    // then historical support for various V8 and non-`NODE_OPTIONS` flags
+    // and also any V8 flags with `--v8-` prefix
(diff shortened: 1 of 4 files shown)


lib/cli/node-flags.js
```diff
@@ -29,16 +29,32 @@ const debugFlags = new Set(['debug', 'debug-brk', 'inspect', 'inspect-brk']);
...
+    // check if the flag begins with dashes; if not, not a node flag.
+    if (!/^--?/.test(flag)) {
+      return false;
+    }
+    // strip the leading dashes to match against subsequent regexes
+    flag = flag.replace(/^--?/, '');
+  }
+  return (
+    // treat --require/-r as Mocha flag even though it's also a node flag
+    !/^r(?:equire)?$/.test(flag) &&
```
boneskull: `/^r(?:equire)?$` is more performant than `/^(?:require|r)$/`

boneskull: (yes, the whole function could be optimized, but not worrying about it rn)

plroebuck: Yeah, cause pretty sure two `if`s would be faster than one `RegExp`
```js
  !(flag === 'require' || flag === 'r') &&
```

lib/cli/node-flags.js
```diff
@@ -29,16 +29,32 @@ const debugFlags = new Set(['debug', 'debug-brk', 'inspect', 'inspect-brk']);
...
  * @private
  */
-exports.isNodeFlag = flag =>
-  !/^(?:require|r)$/.test(flag) &&
-  (nodeFlags.has(flag) ||
-    debugFlags.has(flag) ||
-    /(?:preserve-symlinks(?:-main)?|harmony(?:[_-]|$)|(?:trace[_-].+$)|gc(?:[_-]global)?$|es[_-]staging$|use[_-]strict$|v8[_-](?!options).+?$)/.test(
-      flag
-    ));
+exports.isNodeFlag = (flag, bareword = true) => {
```
plroebuck: Since we're processing cmdline options, your default value seems... inverted.

boneskull: I agree it seems that way at first.  but this function is hit more often *with* the default value (config processing!) than it is *without* it.

lib/cli/node-flags.js
boneskull: agree.  hey, I used to have this function on _one_ line, so give me some credit

Issue #3368, opened 2018-05-02
🚀 Feature: Add pseudo-tests for timing of before()/after() hooks


plroebuck:
### Prerequisites

- [x] Checked that your issue hasn't already been filed by cross-referencing [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20)

Related issues/proposals:
#419 
#700
cybertk/abao#203

### Description

Breakdown of test timing should add setup/teardown pseudo-tests to help give better individual test timings. Timing of `before()` functions would be added to the "setup" pseudo-test, and `after()` functions to the "teardown" pseudo-test of each `describe()`.

Timing of `beforeEach()/afterEach()` would remain with each `it()`.

Reporters would display these as two additional tests named "setup" and "teardown", assuming they exceeded some threshold.

### Rationale

When using Mocha tests to perform RESTful interface testing against a resource, timing of a database setup routine in a `before()` would be added to the first test -- making its performance look terrible. When running API tests, the timing of the routine can be almost as important as it passing.

eight04:
I think it will be better to provide a way to access the timing API or a "pseudo it" that will record the timing. Currently, I have a test file like this:
```js
describe("test cases", () => {
  for (const case of cases) {
    it(case.name, () => {
      const input = fs.readFileSync(case.inputFile);
      const output = doSomething(input);
      const expectedOutput = fs.readFileSync(case.outputFile);
      assert.equal(output, expectedOutput);
    });
  }
});
```
The problem is that setup, assert, and teardown phases are all counted in the timing. What I really care is the calculation time i.e. `doSomething(input)`. Currently, I can use a workaround like:
```js
describe("test cases", () => {
  for (const case of cases) {
    describe(case.name, () => {
      let input, expectedOutput;
      before(() => {
        input = fs.readFileSync(case.inputFile);
        expectedOutput = fs.readFileSync(case.outputFile);
      });
      it("run", () => {
        const output = doSomething(input);
        assert.equal(output, expectedOutput);
      });
    });
  }
});
```
With a timing API, we can simplify this to:
```js
describe("test cases", () => {
  for (const case of cases) {
    it(case.name, function () {
      const input = fs.readFileSync(case.inputFile);
      const expectedOutput = fs.readFileSync(case.outputFile);
      
      this.timeStart();
      const output = doSomething(input);
      this.timeEnd();
      
      assert.equal(output, expectedOutput);
[...]

Issue #700, opened 2013-01-03
Add ability to add up before() runtime with it() runtime


Almad:
--slow is highlighting only long running tests, measuring only time it() took to execute.

This leads to missed slow tests if slow part is hidden in test preparation (in before()).

It would be useful to have some option (--measure-before ?) to highlight tests that are above threshold together with appropriate before(Each)().

Issue #1975, opened 2015-11-23
afterEach inside beforeEach and mocha timing issues.


tomhicks:
Just spent ages figuring out why our tests seem to slow down about half way through a test suite and continue slowly from then on.

Basically, through what looks like a git merge, an `afterEach` call had got inside a `beforeEach` call.

This seemingly causes the `afterEach` to be run in the outer context (or whatever mocha calls the `describe`-scoped `before/after` mechanism) which was adding a load of `afterEach` calls for every subsequent `it` in the suite, causing every test thereafter to run slower.

Mocha's slow test reporting was not picking this up and it was a nightmare to figure out.

This brings up two issues:
1. Should the mocha "slow" timing take into account `beforeEach` and `afterEach` calls? It doesn't seem to, and if it did it would have made debugging this a whole lot easier.
2. Should there be some warning/error about running an `afterEach` inside a `beforeEach`? I can't think of a reason why you would ever want to do this?

Example:

``` javascript

// test1.js
describe("first suite", function() {
  beforeEach(function() {
    // setup

    afterEach(function() {
      console.log("after each first file")
    })
  })

  it("first test", function () { console.log("test1") })

  it("second test", function () { console.log("test2") })

  it("third test", function() { console.log("test3") })
})

// test2.js

describe("second test", function() {
  it("fourth test", function() { console.log("test4") })
})

// output
test1
after each first file
test2
after each first file
after each first file
test 3
after each first file
after each first file
after each first file
test 4
after each first file
after each first file
after each first file

```

As you can see, this adds an extra after step each time an `it` is run. This resulted in ~10,000 extra calls being made to the `afterEach` in one test file, which slowed our tests down significantly. These tests were definitely taking more than the `--slow` timeout, but they were not being reported as slow.

Expected output would be an error being thrown that `afterEach` cannot be run inside `beforeEach`.

Obviously, this is a mistake in the test code, but an error would have been handy. This would also be avoided by using a proper language that doesn't have significant whitespace and linting the indentation, but I feel like mocha could have been more helpful here.

rodoabad:
I'd love this check to be added to mocha but I feel like an ESLint rule could also solve this one real quick.

ORESoftware:
sorry about that problem that you experienced, frankly Git is not the best with JavaScript

and contrary to what @rodoabad said I don't think ESLint is the right way to do this...

[...]

PR #3776, opened 2019-02-26
Include timing of beforeEach in test duration

### LINKED ISSUE #419: 🚀 Feature: Account for beforeEach/afterEach in timing badges
My beforeEach functions are pretty slow, and sometimes even dominate test time. (I set up test fixture data and initialize an Ember app.) But the little timing badges that show up in the browser (above 50ms or so) do not include this time, they only seem to measure the time of the test functions themselves.

I'm thinking that it would be sensible for the badges to include the beforeEach/afterEach time.

(Perhaps there are also other parts where these timings are used, like timeouts -- I'm fairly new to mocha.)

What do you think?

sgilroy:
### Description of the Change
As noted in #419 it is sometimes desirable to include the duration of any `beforeEach`/`setup` hooks in the duration of tests. This PR adds an option `--time-setup` which allows this. The `--time-setup` option is helpful when looking to discover slow tests, where the setup makes up a significant portion of the total duration. Without this option, the user has no visibility into the time being spent in setup. The `--time-setup` option can be combined with `--slow` to find tests that exceed a slow threshold.

### Alternate Designs
Alternative designs:
1. Include detailed timing about all hooks and allow plugins or reporters to decide what to do with the timing information. The existing `duration` attribute would not need to be redefined.
1. Add pseudo-tests for timing of before()/after() hooks #3368.

The design implemented in this PR has the advantage of being compatible with existing tooling and the `--slow` feature. For users who want to opt in to measuring beforeEach hooks, all existing test reporters can display this redefined duration in place of the normal tests-only duration. The alternatives described above could make this feature obsolete, but are significantly larger scope.

Alternative option names:
1. `measure-before` https://github.com/mochajs/mocha/issues/700
1. `include-before-each`
1. `include-setup-time`
1. `before-each-timing`

I'm open to renaming the option.

Mocha does not expose a mechanism for a plugin to extend or override how the duration of each test is calculated. Detailed timing information is not available for tests.

### Benefits
Users are able to detect slow beforeEach hooks and better understand the overall performance of their test suites.

### Possible Drawbacks
Increased complexity. Yet another option to document, understand, and maintain.

### Applicable issues

Fixes #419
semver-minor

### DIFF
--- a/lib/cli/run-option-metadata.js
+++ b/lib/cli/run-option-metadata.js
@@ -42,6 +42,7 @@ exports.types = {
     'recursive',
     'reporters',
     'sort',
+    'time-setup',
     'watch'
   ],
   number: ['retries', 'slow', 'timeout'],
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -233,6 +233,11 @@ exports.builder = yargs =>
         description: 'Specify test timeout threshold (in milliseconds)',
         group: GROUPS.RULES
       },
+      'time-setup': {
+        description:
+          'Include any beforeEach (or setup) time in duration of each test',
+        group: GROUPS.RULES
+      },
       ui: {
         default: defaults.ui,
         description: 'Specify user interface',
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -113,6 +113,7 @@ function Mocha(options) {
     .reporter(options.reporter, options.reporterOptions)
     .useColors(options.color)
     .slow(options.slow)
+    .timeSetup(options.timeSetup)
     .useInlineDiffs(options.inlineDiffs)
     .globals(options.globals);
 
@@ -663,6 +664,21 @@ Mocha.prototype.retries = function(n) {
  */
 Mocha.prototype.slow = function(msecs) {
   this.suite.slow(msecs);
+  return this;
+};
+
+/**
+ * Sets time setup option, which will cause time for any beforeEach (setup)
+ * methods to be included in the duration of each test.
+ *
+ * @public
+ * @see {@link https://mochajs.org/#-s---time-setup|CLI option}
+ * @param {boolean} timeSetup - Whether to enable the time setup option.
+ * @return {Mocha} this
+ * @chainable
+ */
+Mocha.prototype.timeSetup = function(timeSetup) {
+  this.suite.timeSetup(timeSetup);
   return this;
 };
 
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -290,10 +290,10 @@ Runnable.prototype.globals = function(globals) {
 Runnable.prototype.run = function(fn) {
   var self = this;
   var start;
-  if (typeof self.start === 'number') {
+  if (self.start) {
     start = self.start;
   } else {
-    start = Date.now();
+    start = new Date();
   }
   var ctx = this.ctx;
   var finished;
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -289,7 +289,12 @@ Runnable.prototype.globals = function(globals) {
  */
 Runnable.prototype.run = function(fn) {
   var self = this;
-  var start = new Date();
+  var start;
+  if (typeof self.start === 'number') {
+    start = self.start;
+  } else {
+    start = Date.now();
+  }
   var ctx = this.ctx;
   var finished;
   var emitted;
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -501,6 +501,7 @@ Runner.prototype.parents = function() {
 Runner.prototype.runTest = function(fn) {
   var self = this;
   var test = this.test;
+  test.start = self.start;
 
   if (!test) {
     return;
@@ -623,6 +624,7 @@ Runner.prototype.runTests = function(suite, fn) {
     }
 
     // execute test and hook(s)
+    self.start = Date.now();
     self.emit(constants.EVENT_TEST_BEGIN, (self.test = test));
     self.hookDown(HOOK_TYPE_BEFORE_EACH, function(err, errSuite) {
       if (test.isPending()) {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -28,6 +28,11 @@ var createInvalidExceptionError = require('./errors')
   .createInvalidExceptionError;
 
 /**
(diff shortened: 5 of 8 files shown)


sgilroy:
@kevinburkeshyp Could you please sign the mocha [Contributor License Agreement](https://cla.js.foundation/mochajs/mocha?pullRequest=3776) so that your commit can be accepted as part of this PR? Thanks!

kevinburke:
I believe I have other commits to this project already? Are they not sufficient?

kevinburke:
I signed as kevinburke / kev@inburke.com. I don't have access to the burke@shyp.com email address used for the first commit.

You have my blessing to just cherry pick the commit contents and change the author, a note (originally suggested by Kevin Burke) or something in the commit message would be appreciated!

szb512:
Make that 2/3 commiters have signed the CLA.

plroebuck:
For what it's worth, think Mocha should have a `stats.timing` object that collects durations for _all_ suites, tests, hooks implicitly.

sgilroy:
I recreated the branch by cherry picking the commit from burke@shyp.com, changing the author to kev@inburke.com. Looks like the license/cla check is passing now.

sgilroy:
@boneskull Capturing timing details on `stats` sounds desirable, and using a new or modified reporter to then see the complete timing information would potentially address many use cases. However, I think it is important to support the behavior implemented in this PR, and enable a way to actually change the `duration` of each test, so that existing reporters (and the `--slow` feature) will include the time from any setup hooks. It is valuable to be compatible with existing reporters, including the one I use in my IDE.

Given that there does not appear to be a plugin API yet for mocha https://github.com/mochajs/mocha/issues/1457, I don't know how I would accomplish this without making a change to set the `duration` in the core of mocha. Any suggestions for a way to provide existing reporters with `duration` that includes time from setup hooks?

sgilroy:
@boneskull I'm looking for some feedback on my response to your concerns so that I know how to proceed. As I stated I'd like to figure out a path to being able to include time for setup/beforeEach hooks in the duration seen by existing reporters and `--slow`.

Are you suggesting that you will not accept some form of this enhancement into mocha? Or is there some way to rework this to be compatible with the vision for `stats.timing` suggested by @plroebuck?

Perhaps it would be possible for a reporter to modify `duration`. Would this be appropriate? Is there a good/supported way to enhance mocha in this way without patching the core?

boneskull:
@sgilroy Making any built-in reporter report this information is going to be semver-major, and doesn't need to be part of *this* PR, which would be to just *collect* these statistics so a) built-in reporters could eventually use them, and b) a custom reporter could use them immediately.

To expose this information, you'll want to do something like this:

1. Create an e.g., `appendHookDuration` method on `Test`.  This function should be invoked after each `Hook` has completed execution against its `ctx.currentTest` property, passing the `Hook`'s `duration` property.  This method should increase the value of some numeric internal property within `Test`, or use some other data structure.  Careful to reset the value if a test needs to be retried.
1. Create a method on `Test` to retrieve the hook duration (a getter)
1. Tests, of course

To use this information with a reporter, you'd create a listener on the `Runner`'s `EVENT_TEST_END` event, which receives a `Test` object as its parameter.  You can then get the hook duration(s) by calling your getter.  From within this listener, you could add the hook duration(s) to the `duration` property of your `Test`.  Then output the info as you like.

Please let me know if the above is what you were looking for.

boneskull:
(See the `json` reporter for an example of how to collect information about the tests and output it when the run is complete)

sgilroy:
@boneskull I don't understand why you are asserting that "Making any built-in reporter report this information is going to be semver-major". The modified behavior is only effective if the option (currently `--time-setup`) is specified, so there is no breaking change.

You had previously implied (in 2016: https://github.com/mochajs/mocha/pull/2244#issuecomment-225394691) that putting this behavior behind a flag and adding tests (which I have done) would be sufficient to get a PR like this accepted. I appreciate that it would be desirable to have more flexible, detailed timing information available but I am not prepared to implement the solution you described right now, and it would not accomplish my stated objective:  include the duration of any `beforeEach`/`setup` hooks in the duration of tests, as seen by **all** reporters (not just a custom new one). I use multiple existing reporters for different purposes, and I like the format and content of the output; I just want to change what is included in the duration.

I believe this optional alternative definition of duration would be useful to other developers and teams as well. See #419, #700, and #1975, and articles such as [Speeding up Javascript Test Time 1000x](https://medium.com/shyp-engineering/speeding-up-javascript-test-time-1000x-460c528418e7) and [Speeding Up Your JavaScript Test Suite](https://spin.atomicobject.com/2015/06/22/speed-up-javascript-test-suite/) which talk about the importance of the duration of setup hooks. Seeing where time is being spent (including setup time) is an essential part of making tests run faster (and keeping them fast) and currently mocha provides no visibility into the time for `beforeEach`/`setup` hooks.

micdah:
For those wanting this kind of feature, I have come up with a hack way of adding it by using a hook:

```javascript
let start = performance.now()
export const mochaHooks = {
    beforeEach: async function () {
        const currentTest = this.currentTest
        const originalRun = currentTest.run
        currentTest.run = function (fn) {
            originalRun.call(currentTest, function (...args) {
                const end = performance.now()
                currentTest.duration = Math.round(end - start)
                start = end
                fn(...args)
            })
        }
    }
}
```

It's put into a `--require` file which adds a global hook.

I have made it such that `start` is set immediately and updated after each test, this ensures all time spent between tests are included, so any `before` / `beforeEach` / `after` / `afterEach` etc. is counted in the duration.

It's kind of hackish, but works.

JoshuaKGoldberg:
👋 coming back to this @sgilroy, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

I do see that your last message was discussing semver majoring of it - once we're ramped up we can pick up that thread of conversation.

JoshuaKGoldberg:
Closing as it's been ~3 months. If anybody wants to pick up this change and re-uses code from this PR, please add a [co-author attribution](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) to the body of the PR. Cheers all! ☕ 🤎

Issue #3783, opened 2019-02-28
Website HTML tweaks


plroebuck:
### Description

**"docs/_includes/default.html"**

* Fix `<title>` (solution courtesy of @XhmikosR)
```html
    <title>{{ title }}</title>
```

* Remove the `shortcut` relationship. [While allowed, it's meaningless](https://github.com/whatwg/html/commit/95978ef2b115e7dbd03b5b5e8d78c00340119588). See [here](https://mathiasbynens.be/notes/rel-shortcut-icon) also.
```html
<link rel="icon" href="favicon.ico" />
```
 
* Mocha logo
  * Change the `alt` attribute to "Mocha logo"
  * Add an `id` attribute.
  * Migrate `width` and `height` attributes to stylesheet
```html
<img
  id="mocha-logo"
  src="/images/mocha-logo.svg"
  alt="Mocha logo" />
```
**"docs/css/style.css"**
```css
#mocha-logo {
  width: 192px;
  height: 192px;
}
```

* Convert the `<span>` tag to a `<div>`. [Inline tags may **not** contain block tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Conceptual_differences).
```html
<span>
  <a href="https://mochajs.org">mochajs.org</a> is licensed under a
  <a rel="license" href="http://creativecommons.org/licenses/by/4.0/"
          >Creative Commons Attribution 4.0 International License</a>.
   <p>
     <em>Last updated: {{ 'now' | date: '%a %b %d %H:%M:%S %Y' }}</em>
  </p>
</span
```

* Add "rel=home" protocol for Mocha URL. Append slash ("/") to end of URL.
```html
  <a rel="home" href="https://mochajs.org/">mochajs.org</a> is licensed under a
```

* Use "https" protocol for CC license `<a>` "href" URL
```html
  <a rel="license" href="https://creativecommons.org/licenses/by/4.0/"
          >Creative Commons Attribution 4.0 International License</a>.
```

* Convert the`<p>/<em>` tags with the following `<dl>` embedding `<time>`
```html
  <dl class="dl-inline last-modified">
    <dt>Last updated</dt>
    <dd>
      <time
        itemprop="lastModified"
        datetime="{{ 'now' | date: '%Y-%m-%dT%H:%M:%SZ' }}">
              {{ 'now' | date: '%a %b %d %H:%M:%S %Y' }}
      </time>
    </dd>
  </dl>
```
**"docs/css/style.css"**
```css
[...]

PR #3784, opened 2019-02-28
Fixes Parts of #3783

rscotchmer:
### Description of the Change
Fixes some issues outlined in #3783.
* Modified the alt text for the mocha logo and gave it an ID attribute, which is then used in styles.css to set the logo's width and height.
* Added the `external` attribute to the link to the Creative Commons License
* Removed the `shortcut` attribute from the favicon.

### Alternate Designs
N/A

N/A

### Benefits
It improves the website.
Removes meaningless attributes while updating HTML/CSS to more modern standards.

### Possible Drawbacks
None to knowledge.

### Applicable issues
Partial #3783.
semver-patch

### DIFF
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -13,7 +13,7 @@
     <link rel="stylesheet" href="css/normalize.css" />
     <link rel="stylesheet" href="css/style.css" />
     <link rel="stylesheet" href="css/prism.css" />
-    <link rel="shortcut icon" href="favicon.ico" />
+    <link rel="icon" href="favicon.ico" />
 
     <!--[if lt IE 9]> <script src="js/html5shiv.min.js"></script> <![endif]-->
   </head>
@@ -23,10 +23,9 @@
       <h1>
         <a href="/">
           <img
+            id="mocha-logo"
             src="/images/mocha-logo.svg"
-            alt="Mocha"
-            width="192"
-            height="192"
+            alt="Mocha logo"
           />
         </a>
       </h1>
@@ -37,14 +36,14 @@
     <main id="content">{{ content }}</main>
 
     <footer>
-      <span>
+      <div>
         <a href="https://mochajs.org">mochajs.org</a> is licensed under a
-        <a rel="license" href="http://creativecommons.org/licenses/by/4.0/"
+        <a rel="license" external href="http://creativecommons.org/licenses/by/4.0/"
           >Creative Commons Attribution 4.0 International License</a
         >.
         <p>
           <em>Last updated: {{ 'now' | date: '%a %b %d %H:%M:%S %Y' }}</em>
-        </p></span
+        </p></div
       >
     </footer>
   </body>
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -37,14 +37,14 @@
 
     <footer>
       <div>
-        <a href="https://mochajs.org">mochajs.org</a> is licensed under a
-        <a rel="license" external href="http://creativecommons.org/licenses/by/4.0/"
+        <a rel="home" href="https://mochajs.org">mochajs.org</a> is licensed under a
+        <a rel="license external noopener" href="http://creativecommons.org/licenses/by/4.0/"
           >Creative Commons Attribution 4.0 International License</a
         >.
         <p>
           <em>Last updated: {{ 'now' | date: '%a %b %d %H:%M:%S %Y' }}</em>
-        </p></div
-      >
+        </p>
+      </div>
     </footer>
   </body>
 </html>
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -200,7 +200,7 @@ footer {
   border-top: 1px solid #ddd;
 }
 
-footer span {
+footer div {
   display: block;
   margin-right: 30px;
   color: #888;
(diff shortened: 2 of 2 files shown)


docs/css/style.css
```diff
@@ -39,6 +39,11 @@ header {
   padding-top: 20px;
 }
 
+#mocha-logo {
+  width:192;
+  height:192;
```
Alhadis: These need a `px` unit affixed, or else they'll be invalid CSS properties:

~~~css
#mocha-logo{
	width: 192px;
	height: 192px;
}

/* (...Or my favourite trick...) */
#mocha-logo{
	width: 1em;
	height: 1em;
	font-size: 192px;
}
~~~

Unitless values are permitted in `width` and `height` attributes largely due to history, IIRC.

plroebuck: Agreed. See original issue.
And please not "hairy"'s favourite trick...

docs/_includes/default.html
```diff
@@ -37,14 +36,14 @@ <h1>
...
+      <div>
         <a href="https://mochajs.org">mochajs.org</a> is licensed under a
-        <a rel="license" href="http://creativecommons.org/licenses/by/4.0/"
+        <a rel="license" external href="http://creativecommons.org/licenses/by/4.0/"
           >Creative Commons Attribution 4.0 International License</a
         >.
         <p>
           <em>Last updated: {{ 'now' | date: '%a %b %d %H:%M:%S %Y' }}</em>
-        </p></span
+        </p></div
```
plroebuck: Put `<div>` on its own line

docs/_includes/default.html
```diff
@@ -37,14 +36,14 @@ <h1>
     <main id="content">{{ content }}</main>
 
     <footer>
-      <span>
+      <div>
         <a href="https://mochajs.org">mochajs.org</a> is licensed under a
-        <a rel="license" href="http://creativecommons.org/licenses/by/4.0/"
+        <a rel="license" external href="http://creativecommons.org/licenses/by/4.0/"
```
plroebuck: Multiple relations are handled the same way as multiple class names.
```html
<a rel="license external noopener"
  href="http://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0 International License</a>

```

docs/_includes/default.html
```diff
@@ -37,14 +36,14 @@ <h1>
     <main id="content">{{ content }}</main>
 
     <footer>
-      <span>
+      <div>
         <a href="https://mochajs.org">mochajs.org</a> is licensed under a
```
plroebuck:
```html
       <a href="https://mochajs.org" rel="home">mochajs.org</a>
```

plroebuck:
Needs tweek to [stylesheet](https://github.com/mochajs/mocha/blob/4999b430153c1b8be2e73bfdadbf2bd73ee191ad/docs/css/style.css#L203) to match what was there... change `footer span` to `footer div` and see if that makes your changes match the [deployed site](https://mochajs.org/#more-information).
```css
footer div {
```

docs/_includes/default.html
Alhadis: This shouldn't be using `<em>` unless it's ***actually emphasis*** (e.g., like that right there).

I recommend either a plain `<i>` or simply targeting it with a structural selector.

docs/css/style.css
Alhadis: Just as an aside, the second trick I mentioned is a great way to retain aspect ratio without changing more than one property. E.g., an image with a width 150% greater than its height could could be expressed using `width: 1.5em; height: 1em`.

This only matters when media queries are being used to scale elements snugly between viewport sizes though, so it's moot here.

docs/_includes/default.html
```diff
@@ -23,10 +23,9 @@
       <h1>
         <a href="/">
           <img
+            id="mocha-logo"
```
XhmikosR: I wouldn't use an ID for targeting this. Just a `logo` class or something.

docs/_includes/default.html
```diff
@@ -23,10 +23,9 @@
...
             src="/images/mocha-logo.svg"
-            alt="Mocha"
-            width="192"
-            height="192"
```
XhmikosR: Keep the dimensions; the HTML is rendered first so it's always better to have the dimensions here.

docs/css/style.css
```diff
@@ -39,6 +39,11 @@ header {
   padding-top: 20px;
 }
 
+#mocha-logo {
+  width:192px;
```
XhmikosR: This is moot if you just keep the dimensions in the HTML, which IMO is better.

docs/_includes/default.html
```diff
@@ -13,7 +13,7 @@
     <link rel="stylesheet" href="css/normalize.css" />
     <link rel="stylesheet" href="css/style.css" />
     <link rel="stylesheet" href="css/prism.css" />
-    <link rel="shortcut icon" href="favicon.ico" />
+    <link rel="icon" href="favicon.ico" />
```
XhmikosR: BTW you can remove the `/>` from the HTML tags here. It's not needed in HTML5+.

XhmikosR: Forgot to mention, `shortcut icon` is needed for old IE. Which AFAICT since html5shiv is loaded, it is still supported.

If not, that's OK, but both should be dropped.

PR #3789, opened 2019-03-01
add Matomo headers to 11ty and jsdoc; closes #3785

### LINKED ISSUE #3785: add matomo headers to doc layouts
It will need to go in the 11ty layout *and* the API layout.

I don't want to have to update `@mocha/docdash` to make the latter happen; hopefully it allows you to add a random header via configuration.

boneskull:
- use new `headScripts` config per `@mocha/docdash` @ 2.1.0
- do not send analytics in dev environment
- add subdomain tracking

once this is published, then must remove headers from Netlify UI.

### DIFF
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -16,6 +16,8 @@
     <link rel="icon" href="favicon.ico" />
 
     <!--[if lt IE 9]> <script src="js/html5shiv.min.js"></script> <![endif]-->
+
+    <script src="js/matomo.js"></script>
   </head>
 
   <body>
--- a/docs/js/matomo.js
+++ b/docs/js/matomo.js
@@ -1,0 +1,22 @@
+// don't trigger analytics for local development environments
+if (document.domain.indexOf('mochajs.org') >= 0) {
+  // don't know if this must be a global variable
+  var _paq = _paq || [];
+  // subdomain tracking support, e.g., next.mochajs.org vs. mochajs.org
+  _paq.push(['setDocumentTitle', document.domain + '/' + document.title]);
+  _paq.push(['trackPageView']);
+  _paq.push(['enableLinkTracking']);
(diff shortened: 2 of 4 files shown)


Alhadis:
> `@mocha/docdash`

This should have been named "Docha". :wink:

docs/_includes/default.html
```diff
@@ -16,6 +16,8 @@
     <link rel="shortcut icon" href="favicon.ico" />
 
     <!--[if lt IE 9]> <script src="js/html5shiv.min.js"></script> <![endif]-->
+
+    <script src="/js/matomo.js"></script>
```
plroebuck: Why begin `src` attribute URL with slash?

jsdoc.conf.json
```diff
@@ -5,7 +5,8 @@
   },
   "mocha-docdash": {
     "sort": true,
-    "static": false
+    "static": false,
+    "headScripts": ["/js/matomo.js"]
```
plroebuck: same question from above

boneskull: This is necessary because these are in `/api/`, and the js is in `/js/`, not `/api/js/`.

docs/js/matomo.js
```diff
@@ -0,0 +1,21 @@
+// don't trigger analytics for local development environments
+if (document.domain !== 'localhost') {
```
outsideris: Shouldn't we prevent the trigger for netlify preview like https://deploy-preview-3789--mocha.netlify.com/ ?

Only trigger analytics in production site like:

```js
if  (document.domain.indexOf('mochajs.org') !== -1)
```

boneskull: good idea

JoshuaKGoldberg:
Closing out to keep our queue small. Someone please yell at me if this should be re-opened though. Cheers! 🤎

PR #3793, opened 2019-03-02
Remove "package" flag from sample config file because it can only be passes as CLI arg

cspotcode:
### Requirements

### Description of the Change

Adds comments to `.mocharc.yml` to clarify that `--package` can be specified as a CLI argument but will be ignored within a `.mocharc` file.

### Alternate Designs

Alternatively, mocha could support `package` within config files.  Then it would merge the config from package.json with the config from `.mocharc`.  But that doesn't seem useful.

N/A; this is a docs tweak.

### Benefits

Better docs.

### Possible Drawbacks

N/A

### Applicable issues

N/A; this is a docs tweak.

### DIFF
--- a/example/config/.mocharc.yml
+++ b/example/config/.mocharc.yml
@@ -29,6 +29,7 @@ inline-diffs: false
 # needs to be used with grep or fgrep
 # invert: false
 opts: './test/mocha.opts'
+# Only allowed as a CLI flag; ignored in config files.  For example `mocha --package ./manifest.json`
 package: './package.json'
 recursive: false
 reporter: spec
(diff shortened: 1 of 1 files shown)


boneskull:
I guess it doesn't make sense to me to leave this in the example config file if it's ignored.

boneskull:
@cspotcode 	What do you think?  We should just remove it from the example config, yes?

cspotcode:
Sorry I didn't notice your replies until now.  I think you're right; if the flag is documented elsewhere then we should just remove it from those sample configurations.

boneskull:
why this causes a coverage drop is beyond me.

PR #3794, opened 2019-03-02
Fix npm scripts on Windows

XhmikosR:


XhmikosR:
mkdirp is a dependency, which many dependencies/devDependencies depend on it already.

rimraf is doable, although I'm not sure if it's even worth it. Because there are already many other devDependencies that depend on rimraf too.

PR #3795, opened 2019-03-02
docs HTML fixes

XhmikosR:
Refs #3783 

Closes #3784 by superseding it.

### DIFF
--- a/docs/_includes/backers.md
+++ b/docs/_includes/backers.md
@@ -1,7 +1,8 @@
 ## Backers
 
-Find Mocha helpful?  Become a [backer](https://opencollective.com/mochajs#support) and support Mocha with a monthly donation.
+Find Mocha helpful? Become a [backer](https://opencollective.com/mochajs#support) and support Mocha with a monthly donation.
 
 <!-- markdownlint-disable MD034 -->
-{% for i in (0..29) %}[![](https://opencollective.com/mochajs/backer/{{ i }}/avatar.jpg)](https://opencollective.com/mochajs/backer/{{ i }}/website){: target="_blank" rel="noopener"}{% endfor %}
-{: .image-list id="_backers" }
+
+{% for i in (0..29) %}[![](https://opencollective.com/mochajs/backer/{{ i }}/avatar.jpg)](https://opencollective.com/mochajs/backer/{{ i }}/website){: target="\_blank" rel="noopener"}{% endfor %}
+{: .image-list class="backers" }
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width,initial-scale=1" />
-    <title>{{ page.title }}</title>
+    <title>{{ title }}</title>
     <link
       rel="preload"
       href="https://opencollective.com/static/images/user.svg"
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -37,15 +37,15 @@
     <main class="content">{{ content }}</main>
 
     <footer>
-      <span>
+      <div>
         <a href="https://mochajs.org">mochajs.org</a> is licensed under a
         <a rel="license" href="http://creativecommons.org/licenses/by/4.0/"
           >Creative Commons Attribution 4.0 International License</a
         >.
         <p>
           <em>Last updated: {{ 'now' | date: '%a %b %d %H:%M:%S %Y' }}</em>
-        </p></span
-      >
+        </p>
+      </div>
     </footer>
   </body>
 </html>
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -15,11 +15,11 @@
     <link rel="stylesheet" href="css/prism.css">
     <link rel="shortcut icon" href="favicon.ico">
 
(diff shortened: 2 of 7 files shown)


docs/_includes/default.html
```diff
@@ -1,51 +1,51 @@
-<!DOCTYPE html>
+<!doctype html>
```
XhmikosR: So, something is overwriting this change, and I think it's an old `html-minifier` version. This was fixed a few versions ago, but I don't have the time to track which dep (I think it's that assetgraph-builder one).

docs/index.md
```diff
@@ -1491,7 +1491,7 @@ A typical setup might look something like the following, where we call `mocha.se
   <head>
     <meta charset="utf-8" />
     <title>Mocha Tests</title>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
```
XhmikosR: For some reason, something that injects or formats the snippets, adds the useless end tags. Not a big deal, it's just it's useless.

docs/_includes/default.html
```diff
@@ -1,51 +1,51 @@
...
-    />
-    <link rel="stylesheet" href="css/normalize.css" />
-    <link rel="stylesheet" href="css/style.css" />
-    <link rel="stylesheet" href="css/prism.css" />
-    <link rel="shortcut icon" href="favicon.ico" />
+    >
+    <link rel="stylesheet" href="css/normalize.css">
+    <link rel="stylesheet" href="css/style.css">
+    <link rel="stylesheet" href="css/prism.css">
+    <link rel="shortcut icon" href="favicon.ico">
```
XhmikosR: Note that I didn't change this here since we don't have a clear list of supported browsers and this is needed for IE.

docs/_includes/default.html
plroebuck: Because it's [wrong](https://www.w3.org/TR/html-polyglot/#doctype)...

XhmikosR: No it's not wrong. You are confusing things. That's for XML.

https://mathiasbynens.be/notes/xhtml5#doctype

plroebuck: Even your own reference says it's incorrect.

XhmikosR: I don't think you even read what I sent. That XHTML, not HTML, which is what you are using.

plroebuck: Sure I did
> If you insist on using it, however, you should know ‘DOCTYPE’ is supposed to be written in uppercase

 but missed part
> in XML mode.

plroebuck: Nonetheless, not accepting _any_ change going against W3C...

XhmikosR: Seriously, you don't see that this is HTML, not XHTML?

You keep saying something that's wrong.

plroebuck: >  Polyglot markup uses a document type declaration (DOCTYPE) specified by section 8.1.1 of [HTML5]. In addition, the DOCTYPE conforms to the following rules:
>
>    The string DOCTYPE is in uppercase letters.

docs/index.md
plroebuck: Welcome to the prettier hate club...

docs/index.md
```diff
@@ -1598,7 +1598,7 @@ tests as shown below:
...
```
plroebuck: `initial-scale` defines a ratio -- hence, the number written in "float/double" format.

XhmikosR: ...which is fine an integer in this case.

docs/_includes/backers.md
```diff
@@ -1,7 +1,8 @@
 ## Backers
 
-Find Mocha helpful?  Become a [backer](https://opencollective.com/mochajs#support) and support Mocha with a monthly donation.
+Find Mocha helpful? Become a [backer](https://opencollective.com/mochajs#support) and support Mocha with a monthly donation.
 
 <!-- markdownlint-disable MD034 -->
-{% for i in (0..29) %}[![](https://opencollective.com/mochajs/backer/{{ i }}/avatar.jpg)](https://opencollective.com/mochajs/backer/{{ i }}/website){: target="_blank" rel="noopener"}{% endfor %}
-{: .image-list id="_backers" }
+
+{% for i in (0..29) %}[![](https://opencollective.com/mochajs/backer/{{ i }}/avatar.jpg)](https://opencollective.com/mochajs/backer/{{ i }}/website){: target="\_blank" rel="noopener"}{% endfor %}
```
Munter: What's the purpose of the added backslash?

docs/_includes/default.html
boneskull: @plroebuck The spec you're referring to is out-of-date, but importantly, "polyglot" markup is defined as markup that is both valid HTML5 and XHTML (I think this is referred to as "XHTML5") at once.  Our site is not claiming to be XHTML-compatible.

Nowadays, browser vendors work with/from the WHATWG's HTML living standard.  Here's what it says about [doctype](https://html.spec.whatwg.org/#syntax-doctype).

docs/_includes/backers.md
boneskull: this could be prettier choking on the template code.

boneskull: all 11ty template stuff needs wrapping in 

```html
<!-- prettier-ignore -->

stuff

<!-- prettier-ignore-end -->
```

PR #3796, opened 2019-03-02
mocha.css: remove unit from 0 and fix `box-shadow`.

XhmikosR:
`0` isn't a valid keyword; use `none` instead.

### DIFF
--- a/mocha.css
+++ b/mocha.css
@@ -167,9 +167,9 @@ body {
   -webkit-border-radius: 0;
   -moz-border-radius: 0;
   border-radius: 0;
-  -webkit-box-shadow: 0;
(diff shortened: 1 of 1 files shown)


mocha.css
```diff
@@ -317,7 +317,7 @@ body {
 
 @media screen and (max-device-width: 480px) {
   #mocha {
-    margin: 60px 0px;
+    margin: 60px 0;
```
plroebuck: Prefer to leave 'px' here for documentation

XhmikosR: It's useless. https://www.w3.org/TR/css-values-3/#lengths

PR #3799, opened 2019-03-02
Update devDependencies to fix the npm vulnerabilities

XhmikosR:
The semver compliant ones, i.e. no major version bump.

This fixes 68 vulnerabilities. (from 93 down to 25)

Not sure which Node.js version the contributors use, but IMO they should use >=8 to get npm 6 which has `npm audit` support.


plroebuck:
Just because a new version of a dependency has been released doesn't imply it _should_ be updated here. Checks should be done on each individually to ensure they complied with semver rules.

XhmikosR:
I didn't *just* update the version, just to update it. As you can see the dep tree is flattened **a lot**.

That's a huge thing you should care about.

XhmikosR:
Down to 25 vulnerabilities, from 93.

XhmikosR:
Also fixes those `npm i` errors coming from the zopfli wrapper, due to the assetgraph-builder update

boneskull:
LGTM, thanks.

Issue #3800, opened 2019-03-02
docs: configure svgo


XhmikosR:
Currently it's using the defaults which may or may not be optimal. This should be configured to a well know config. I can supply the config as long as I figure out how that assets-builder package works.

PR #3801, opened 2019-03-02
docs: compress images

XhmikosR:
I guess I should drop the SVG patch since those are being compressed on build time, but we need to sort out #3800 so that we get the best results.

If everyone's happy with the result, I will squash the other 2 PNG patches into 1.

The reason for this patches is that I used the best possible options and tools for PNG images, optipng, zopflipng, pngout and deflopt, with the maximum compression settings.

So, this will get us the best result, and also docs build time should decrease, in theory.

EDIT:

From a quick test I see page size has dropped 30.2 KB with these patches.


XhmikosR:
From a quick test I see page size has dropped 30.2 KB with these patches.

docs/images/mocha-logo.svg
```diff
@@ -1,125 +1,9 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
plroebuck: Don't think you should have deleted the [XML declarations](https://stackoverflow.com/a/38172170).

XhmikosR: Of course they can be deleted.

XhmikosR: But see my original comment. The production SVGs are compressed with the svgo default atm.

And you will see this stuff is removed from there too.

boneskull:
@Munter 	I'm not sure what you're requesting here... @XhmikosR 	do you understand?

XhmikosR:
Yes, but like I explain, this results in better compression that what the tool you use is doing.

XhmikosR:
I can keep the lossless compression patch, drop SVGH, as long as someone guides me to that tool that's being used image compression option. But even so, we'll never get the same results as this, because I use extensively slow settings for the best savings, that we don't want to have on each build.

PR #3803, opened 2019-03-03
docs: Add autoprefixer.

### LINKED ISSUE #3797: assetgraph-builder: specify supported browsers
Currently there are many CSS prefixes missing. Since this tool is used, its browsers option should be used. Not sure if it supports `.browserslistrc` or not, I guess this would be preferable.

I have added them manually but this is a good chance to revisit which browsers are supported and make any other changes too like dropping html5shiv, and switch to `link rel=icon`.

XhmikosR:
Fixes #3797.

The browsers I went with are the ones from Bootstrap v4 just so that we get an idea. We can adapt it as needed of course :)

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -276,7 +276,7 @@ module.exports = {
       },
       postbuild: {
         script:
-          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
+          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --browsers ">= 1%, last 1 major version, not dead, Chrome >= 45, Firefox >= 38, Edge >= 12, Explorer >= 10, iOS >= 9, Safari >= 9, Android >= 5, Opera >= 30" --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && shx cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
         description: 'Post-process docs after build',
         hiddenFromHelp: true
       },
--- a/package.json
+++ b/package.json
@@ -521,6 +521,7 @@
     "@mocha/contributors": "^1.0.3",
     "@mocha/docdash": "^2.1.0",
     "assetgraph-builder": "^6.9.0",
+    "autoprefixer": "^9.4.9",
     "browserify": "^16.2.3",
     "browserify-package-json": "^1.0.1",
     "chai": "^4.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1192,6 +1192,33 @@
       "integrity": "sha1-NCQX2PLzRhsUzwkIjV7fh5HcmDI=",
       "dev": true
     },
+    "autoprefixer": {
+      "version": "9.4.9",
+      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-9.4.9.tgz",
+      "integrity": "sha512-OyUl7KvbGBoFQbGQu51hMywz1aaVeud/6uX8r1R1DNcqFvqGUUy6+BDHnAZE8s5t5JyEObaSw+O1DpAdjAmLuw==",
+      "dev": true,
+      "requires": {
+        "browserslist": "^4.4.2",
+        "caniuse-lite": "^1.0.30000939",
+        "normalize-range": "^0.1.2",
+        "num2fraction": "^1.2.2",
+        "postcss": "^7.0.14",
+        "postcss-value-parser": "^3.3.1"
+      },
+      "dependencies": {
+        "browserslist": {
+          "version": "4.4.2",
+          "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-4.4.2.tgz",
+          "integrity": "sha512-ISS/AIAiHERJ3d45Fz0AVYKkgcy+F/eJHzKEvv1j0wwKGKD9T3BrwKr/5g45L+Y4XIK5PlTqefHciRFcfE1Jxg==",
+          "dev": true,
+          "requires": {
+            "caniuse-lite": "^1.0.30000939",
+            "electron-to-chromium": "^1.3.113",
+            "node-releases": "^1.1.8"
+          }
+        }
+      }
+    },
     "aws-sign2": {
       "version": "0.7.0",
       "resolved": "https://registry.npmjs.org/aws-sign2/-/aws-sign2-0.7.0.tgz",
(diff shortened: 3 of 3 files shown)


Munter:
This PR highlighted what we consider a flaw in the `--browsers` switch for assetgraph-builder and we are working on getting an improvement through the build pipelines.

When the update hits it should be enough to have a `.browserslistrc` and have autoprefixer installed.

We should of course run autoprefixer. I simply forgot it didn't come out of the box when I set up the build system. I'll make the adjustments for the upcoming version of assetgraph-builder on this branch and bump the version when it lands

package-scripts.js
```diff
@@ -276,7 +276,7 @@ module.exports = {
       },
       postbuild: {
         script:
-          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --browsers ">= 1%, last 1 major version, not dead, Chrome >= 45, Firefox >= 38, Edge >= 12, Explorer >= 10, iOS >= 9, Safari >= 9, Android >= 5, Opera >= 30" --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && shx cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
+          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --browsers ">= 1%, last 1 major version, not dead, Chrome >= 45, Firefox >= 38, Edge >= 12, Explorer >= 10, iOS >= 9, Safari >= 9, Android >= 5, Opera >= 30" --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
```
plroebuck: Why'd you add all that crap to the cmdline when it could have been added to "package.json" in `]browserlist](https://github.com/browserslist/browserslist-example#how-to-add-browserslist-to-your-project)`?

.browserslistrc
```diff
@@ -0,0 +1,11 @@
+>= 1%
+last 1 major version
+not dead
+Chrome >= 45
+Firefox >= 38
+Edge >= 12
+Explorer >= 10
```
plroebuck: We don't support IE10...

package-scripts.js
Munter: Turns out assetgraph-builder didn't quite behave as documented and didn't let you use autoprefixer without configuring like this. A patch for that is on the way. Then we can switch to `.browserslit`

.browserslistrc
Munter: Who gets prefixed CSS properties on the documentation page is not nessessarily the same set as who gets to run the the test runner

package-scripts.js
plroebuck: Why create external file instead of using "package.json"?

.browserslistrc
plroebuck: True, but...
> On January 12, 2016, support ended for IE10 on Windows operating systems capable of running Internet Explorer 11

plroebuck: Since IE11 has been available since Nov 2013, I just don't see the point...
Since Jan 2016, only the latest version of the browser still gets security updates...

Munter: Suggest your browser ranges then. This is kind of what this file can help us do

package-scripts.js
Munter: Because that's the convention we have followed for every other tool that has this choice, and because it's the most common convention I've seen people use, which will increase the chances of them seeing something they can relate to

.browserslistrc
plroebuck: * Safari 11+
* IE11 only
* others: most recent LTS, and 2 most recent majors

But migrate contents to "package.json" and remove this file...

Munter: I disagree on moving this to package.json. Eslint can also be configured in package.json but we don't do that. There are tonnes of configuration files in the root. Why is this one a problem?

package-scripts.js
plroebuck: I find this file confusing since my impression would be that it implies what Mocha the runtime supports, not browser is needed to view the website.

plroebuck: Was any consideration ever given to migrating the website to its own repository?

.browserslistrc
plroebuck: Suggested [what they did](https://github.com/browserslist/browserslist#queries)...

package-scripts.js
Munter: The website was migrated from its own repository into the main repository in order to version it together with the library.

If you objection to the root location is potential confusion with what browsers mocha.js supports, we can move this file into `/docs/`

plroebuck: Moving into "/docs/" would reduce confusion, IMO. Adding a comment to the top of the file indicating the setup relates to website generation would be much appreciated as well.

Trying to read the `[browserslist](https://github.com/browserslist/browserslist)` documentation to see how to support _both_ concepts; if we have settings in "package.json" to imply runtime, would that cause issues with the website build?

plroebuck: Use `"BROWSERSLIST_CONFIG"` environment variable for website generation.

Munter: I fail to see the aversion against having a config file named intuitively after what it does. An environment variable would be declared some non-obvious place in package-scripts, which is already quite noisy.

I moved the config file to `/docs/` and adjusted the browser range to be more modern

.browserslistrc
Munter: Changed the browser ranges to reflect more modern browsers

package-scripts.js
plroebuck: The browserslist config file, now moved into "/docs/", is fine.

What I was attempting to communicate was that _if_ we added "browserslist" to "package.json" to represent runtime, the environment variable _could be used_ to specify that the "/docs/.browserlistrc" be used for website generation.

Have no experience with the package -- "BROWSERSLIST_CONFIG" may not be necessary, but wanted to ensure that the default lookup order could be overridden -- not sure it would be immediately obvious which was being used otherwise.

docs/.browserslistrc
```diff
@@ -0,0 +1,2 @@
+last 2 major versions
```
plroebuck:
```
## Browsers supported for website generation
```

docs/.browserslistrc
```diff
@@ -0,0 +1,2 @@
...
+not dead
```
plroebuck: Firefox ESR

.browserslistrc
plroebuck: BTW, my apologies at having added to the confusion here. Many of the comments really belonged to runtime aspect, not website generation.

I'll blame not thinking straight on a lack of coffee.

plroebuck:
So with the addition of "autoprefixer", is [this](https://github.com/assetgraph/assetgraph-builder#specifying-which-browsers-to-support) no longer needed?

Not following how AG-B automagically runs "autoprefixer" for us using "/docs/.browserlistrc",
as "autoprefixer" is not an AG-B dependency.

package.json
```diff
@@ -520,7 +520,8 @@
     "@11ty/eleventy": "^0.7.1",
     "@mocha/contributors": "^1.0.3",
     "@mocha/docdash": "^2.0.0",
-    "assetgraph-builder": "^6.6.3",
+    "assetgraph-builder": "^6.9.0",
+    "autoprefixer": "^9.4.9",
```
plroebuck: Why isn't this an AG-B dependency? Don't see why it should be ours...

XhmikosR:
.browserslistrc doesn't work with the tool you are using thus why I had to add it here.

Not gonna comment on the other stuff, sorry.

XhmikosR:
And please don't edit my PRs from now on or I'll have to block this. You are doing things wrong.

package.json
Munter: autoprefixer is on a quite rapid release cycle to keep up with all the browser changes. So assetgraph-builder lets you bring your own dependency so you can update the version independently of assetgraph-builder releases.

Munter:
@XhmikosR I was helping you get this past review and merged. You just reverted the work I did to make that happen. As a maintainer of the docs I find this disrespectful of the time I spent here, especially with the "doing things wrong" comment.

I'm closing this PR and doing the work myself instead then

XhmikosR:
I'm sorry, but you don't know a single thing about this. I refuse to further engage to any discussion with you.

In fact I'm gonna close any PRs I made and move one. Good luck to the project with such maintainers as you.

Munter:
As a maintainer of both the mocha docs, and a maintainer of the build system we use I respectfully disagree

PR #3804, opened 2019-03-03
index.md Remove gnu makefile reference.

XhmikosR:
This seems to be a leftover.


plroebuck:
But it's not. Requires build tools for compilation of modules with native code.

XhmikosR:
Useless and wrong reply.

@boneskull  please reopen this and merge it.

boneskull:
Building the docs requires build tools and potentially other libs, yes. 

Running the tests no longer requires `make`,  however.

PR #3805, opened 2019-03-03
Stop inlining stuff.

XhmikosR:
This should make the HTML size ~-120KB (uncompressed, -90 compressed) HTML, the images can now be cached and should lead to faster TTFB.

We can adapt this of course, like still inline small SVG images in the HTML so that they can be styled etc.

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -95,7 +95,7 @@ nav.badges a + a {
 
 #_backers a img {
   width: 64px;
-  background: url(/images/backer-background.svg?inline) center center no-repeat;
+  background: url(/images/backer-background.svg) center center no-repeat;
 }
 
 h2 {
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -276,7 +276,7 @@ module.exports = {
       },
       postbuild: {
         script:
-          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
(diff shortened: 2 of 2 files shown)


docs/css/style.css
```diff
@@ -95,7 +95,7 @@ nav.badges a + a {
 
 #_backers a img {
   width: 64px;
-  background: url(/images/backer-background.svg?inline) center center no-repeat;
+  background: url(/images/backer-background.svg) center center no-repeat;
```
Munter: The inlining of this background gives a significantly improved user experience in the page rendering progress: http://webpagetest.org/video/compare.php?tests=190303_XT_8d430aac89f487867e1c34a2091db20c-r:1-c:0,190303_R9_221121187cf891367ca68538dc27e542-r:1-c:0

package-scripts.js
```diff
@@ -276,7 +276,7 @@ module.exports = {
       },
       postbuild: {
         script:
-          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --optimizeimages --svgo --inlinehtmlimage 9400 --inlinehtmlscript 0 --asyncscripts && cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
+          'buildProduction docs/_site/index.html --outroot docs/_dist --canonicalroot https://mochajs.org/ --optimizeimages --svgo --inlinehtmlimage 0 --inlinehtmlscript 0 --inlinecssimage 0 --asyncscripts && cp docs/_headers docs/_dist/_headers && node scripts/netlify-headers.js >> docs/_dist/_headers',
```
Munter: I think we can adjust some of the inlining parameters, but then I also think we need to explicitly inline the logo, seeing as this change caused a regression in hero rendering and lost us the animation of the logo: http://webpagetest.org/video/compare.php?tests=190303_XT_8d430aac89f487867e1c34a2091db20c-r:1-c:0,190303_R9_221121187cf891367ca68538dc27e542-r:1-c:0

PR #3812, opened 2019-03-06
Add autoprefixer to documentation page CSS

Munter:
This PR replaces #3803

### DIFF
--- a/docs/.browserslistrc
+++ b/docs/.browserslistrc
@@ -1,0 +1,4 @@
+## Browsers support for autoprefixing CSS for the website
+last 2 major versions
+not dead
+Firefox ESR
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,7 @@
         },
         "minimist": {
           "version": "1.2.0",
-          "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
+          "resolved": "http://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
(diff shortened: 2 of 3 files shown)


plroebuck:
> Not following how AG-B automagically runs "autoprefixer" for us using "/docs/.browserlistrc",
as "autoprefixer" is not an AG-B dependency.

Based on your reply why you might not desire to make it a direct dependency.
Maybe add "autoprefixer" as AG-B [optional dependency](https://docs.npmjs.com/files/package.json#optionaldependencies)?

Munter:
We had some of these bring-your-own-version module dependencies marked as optional at one point in time. The problem is how the default npm behavior shows that to the user at install time. First off, an optional dependency is installed at install time, but is allowed to fail while still succeeding the root level installation run. However, a failed optional install is exactly as verbose with red blinking lights as a failed installation of a dev or non-dev dependency, causing every user to assume the installation failed despite the exit code being 0.

Secondarily, in order to actually control the specific version you'd like from an optional dependency, we'd need to specify the wides possible version range, basically `*`, and you'd **still** have to save the version of your own choice as a direct dependency of your project.

So we concluded that `optionalDependency` is bollocks in its current form and went over to make a runtime check for module availability and have assetgraph log an `info` event that explains how you can install a specific optional module to gain certain improvements in your build.

We're actively working on reducing some of these more optional parts of the pipeline. I have a branch running where we've eliminated the need for two-thirds of these types of checks because the node ecosystem now provides modules that wrap certain OS-level binaries we were depending on to do the same work before. But the concept of letting developers chose their own version of specific libraries still remains

PR #3814, opened 2019-03-06
update yargs; closes #3742

### LINKED ISSUE #3742: running mocha shows yargs deprecation with latest yargs dependency
### Prerequisites

### Description
My project uses the latest `yargs@13.1.0`, which reports a deprecation warning when its parser is configured in the `package.json`. see https://github.com/yargs/yargs/pull/1262.

because mocha configures it's own yargs in its [package.json](https://github.com/mochajs/mocha/blob/master/package.json#L602-L606), running the tests will show the deprecation warning.

### Steps to Reproduce
example project: https://github.com/tripodsan/mochayargstest

`test.js`
```js
const yargs = require('yargs');
function runCli() {
  new yargs()
    .help().parse(['--help']);
}
describe('Test Yargs', () => {
  it('shows help', () => {
    runCli();
  })
});
```

1. run `npm test`
```
$ npm test

> mochayargs@1.0.0 test /Users/nobody/tmp/mochayargs
> mocha test.js

  Test Yargs
Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.
Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
```

2. run w/o mocha

```
$ node index.js
Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
```
**Expected behavior:** 
no deprecation warning is shown from yargs.

**Actual behavior:**
deprecation warning is shown then running tests.

always.

### Versions
```
    "mocha": "^6.0.0",
    "yargs": "^13.2.1"
```

### Additional Information
I think that mocha should use the new https://github.com/yargs/yargs/blob/master/docs/api.md#parserconfigurationobj instead of configuring the parser through the `package.json`

boneskull:
as of v13.2.2, yargs has un-un-supported Node.js v6, so we can now consume it, as well as avoid the deprecation by passing the parser configuration via `parserConfiguration()` method.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -13,7 +13,7 @@ const debug = require('debug')('mocha:cli:cli');
 const symbols = require('log-symbols');
 const yargs = require('yargs');
 const path = require('path');
-const {loadOptions} = require('./options');
+const {loadOptions, YARGS_PARSER_CONFIG} = require('./options');
 const commands = require('./commands');
 const ansi = require('ansi-colors');
 const {repository, homepage, version, gitter} = require('../../package.json');
@@ -59,6 +59,7 @@ exports.main = (argv = process.argv.slice(2)) => {
     Docs: ${ansi.yellow(homepage)}
       `
     )
+    .parserConfiguration(YARGS_PARSER_CONFIG)
     .parse(argv, loadOptions(argv));
 };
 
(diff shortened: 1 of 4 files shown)


boneskull:
hmm, this is not good.

PR #3710, opened 2019-02-03
closes #3702 default column CSS

crossaaron:
### Description of the Change
Changed default CSS for `<ul>` elements to single column.
Created a "two-column-display" class as well for use if desired.

### Alternate Designs
NA

### Why should this be in core?
NA (already exists)

### Benefits
Requested styling fix

### Possible Drawbacks
Longer scroll down

### Notes
Had to reinstall "wd" module to get tests to run completely and without error locally

### Applicable issues
Fixes #3702 
semver-patch

PR #3815, opened 2019-03-06
MarkdownMagic script updates

plroebuck:
### Description of the Change

* Quoted pathname used in exec, allowing for embedded spaces
* Changed codefence language to 'text'
* Saved `markdownTOC` output to variable prior to output

### Alternate Designs
NA

NA

### Benefits
Code can now run on Windows if pathname contains spaces.

### Possible Drawbacks
None to knowledge.

### Applicable issues
semver-patch

### DIFF
--- a/scripts/markdown-magic.config.js
+++ b/scripts/markdown-magic.config.js
@@ -20,11 +20,11 @@ exports.transforms = {
   usage: (content, options) => {
     const {executable} = options;
     const flag = options.flag || '--help';
-    const header = options.header || '\n```plain';
+    const header = options.header || '\n```text';
     const footer = options.footer || '```\n';
     const output = stripAnsi(
       String(
-        execSync(`${process.execPath} ${executable} ${flag}`, {
+        execSync(`"${process.execPath}" ${executable} ${flag}`, {
           cwd: path.join(__dirname, '..')
         })
       ).trim()
--- a/scripts/markdown-magic.config.js
+++ b/scripts/markdown-magic.config.js
@@ -8,10 +8,10 @@
  */
 
 const {execSync} = require('child_process');
-const stripAnsi = require('strip-ansi');
-const markdownToc = require('markdown-toc');
-const path = require('path');
 const fs = require('fs');
+const path = require('path');
+const markdownToc = require('markdown-toc');
+const stripAnsi = require('strip-ansi');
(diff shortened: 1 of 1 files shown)


scripts/markdown-magic.config.js
```diff
@@ -40,18 +40,15 @@ exports.transforms = {
...
-    );
+    const toc = markdownToc(config.outputContent, {
+      slugify: require('uslug'),
+      bullets: options.bullets,
+      firsth1: false,
+      // if filter is supplied, maxdepth is apparently ignored,
+      // so we have to do it ourselves.
+      filter: (str, ele) => ele.lvl < 2 && !IGNORED_HEADINGS_REGEXP.test(str)
+    }).content;
+    return '\n' + toc + '\n';
```
boneskull: might be worth adding a comment as to the relevance of the `\n`'s.  they ensure markdownlint doesn't barf on the rendered output.

when we ensure that the auto-generated content is no longer under VCS (and is only part of the docs build), then we don't need the extra newlines.

scripts/markdown-magic.config.js
```diff
@@ -20,11 +20,11 @@ exports.transforms = {
   usage: (content, options) => {
     const {executable} = options;
     const flag = options.flag || '--help';
-    const header = options.header || '\n```plain';
+    const header = options.header || '\n```text';
```
boneskull: does this actually matter or make a difference? curious

boneskull:
@plroebuck 	merge at will

scripts/markdown-magic.config.js
plroebuck: Not sure what it does if language is incorrect. Looked around earlier for what language was in use if none specified, but couldn't determine that either.

"text" seemed closest to your intent based on [languages](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) documentation.

scripts/markdown-magic.config.js
plroebuck: I had just considered it good style -- didn't realize an automated tool agreed with me. :)

plroebuck: [Implementation preference](https://github.com/mochajs/mocha/pull/3710#issuecomment-470178999)?

PR #3816, opened 2019-03-07
guard against undefined timeout option in constructor; closes #3813

### LINKED ISSUE #3813: Regression caused by the deprecation of `enableTimeouts`
### Description

There is a regression caused by the deprecation of `enableTimeouts`

As you cas see here: https://github.com/mochajs/mocha/pull/3556/files#diff-aa849a970cef551664c12f04a4209f6fR121
`enableTimeouts` was marked as deprecated and the use of `timeout: false` should replace it.

But in the same time, we removed:
```js
if (typeof options.timeout !== 'undefined' && options.timeout !== null) {
    this.timeout(options.timeout);
```

The bug is present with `grunt-mocha-test`. As you can see, `grunt-mocha-test` parse all options as undefined:
```js
{ grep: undefined,
  reporter: 'spec',
  bail: false,
  ui: undefined,
  timeout: undefined,
  invert: undefined,
  ignoreLeaks: undefined,
  growl: undefined,
  globals: undefined,
  require: undefined,
  colors: undefined,
  slow: undefined 
}
```

The bug is not present with `grunt-simple-mocha`

Should we kept it while `enableTimeouts` is marked as deprecated ?

### Steps to Reproduce
```sh
npm install grunt grunt-mocha-test
```

in `Gruntfile.js`

```js
module.exports = function(grunt) {

  grunt.loadNpmTasks('grunt-mocha-test');

  grunt.initConfig({
    mochaTest: {
      test: {
        src: ['test/**/*.js']
      }
    }
  });

  grunt.registerTask('default', 'mochaTest');

};
```

Then launch test
```sh
grunt mochaTest
```

### Expected behavior

Mocha v6 should work with options like {timeout: undefined} because there is only a deprecation warning in v6 for `enableTimeouts` option.

### Actual behavior:

TypeError: Cannot read property 'toString' of undefined
  at Suite.timeout (/home/fwadmin-server/smc/node_modules/mocha/lib/suite.js:124:10)
  at Mocha.timeout (/home/fwadmin-server/smc/node_modules/mocha/lib/mocha.js:625:14)
  at new Mocha (/home/fwadmin-server/smc/node_modules/mocha/lib/mocha.js:124:8)
  at new MochaWrapper (/home/fwadmin-server/smc/node_modules/grunt-mocha-test/tasks/lib/MochaWrapper.js:30:15)

### Reproduces how often:
  Always

### Versions
  6.0.2

boneskull:
IMO `Suite#timeout` should eventually do nothing if its parameter is `undefined`.  it should not be up to Mocha's constructor to guard against this.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -120,12 +120,15 @@ function Mocha(options) {
     utils.deprecate(
       'enableTimeouts is DEPRECATED and will be removed from a future version of Mocha. Instead, use "timeout: false" to disable timeouts.'
     );
+    if (options.enableTimeouts === false) {
+      this.timeout(0);
+    }
   }
-  this.timeout(
-    options.enableTimeouts === false || options.timeout === false
-      ? 0
-      : options.timeout
-  );
+
+  // this guard exists because Suite#timeout does not consider `undefined` to be valid input
+  if (typeof options.timeout !== 'undefined') {
+    this.timeout(options.timeout === false ? 0 : options.timeout);
+  }
 
   if ('retries' in options) {
     this.retries(options.retries);
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -114,6 +114,7 @@ Suite.prototype.clone = function() {
  * Set or get timeout `ms` or short-hand such as "2s".
  *
  * @private
+ * @todo Do not attempt to set value if `ms` is undefined
  * @param {number|string} ms
  * @return {Suite|number} for chaining
  */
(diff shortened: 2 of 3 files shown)


boneskull:
cleaned up the constructor unit tests a little while I was in there

plroebuck:
As `undefined` is not a valid value for _any_ Mocha option, should the user "options" object have them stripped before merged with default values?

Otherwise, won't this just turn into a game of whack-a-mole?

boneskull:
> As undefined is not a valid value for any Mocha option, should the user "options" object have them stripped before merged with default values?

> Otherwise, won't this just turn into a game of whack-a-mole?

I'm thinking no, because prior to v6, anybody trying to pass `undefined` values via options _in a way which was not previously guarded against_ would find a thrown exception.

The problem here, in particular, is that I accidentally removed such a guard.

If someone tries this with some _other_ option that doesn't handled `undefined` nicely, they will get an exception, and I suppose we can live with that for now.

boneskull:
A 0.06% decrease in coverage should be a warning.  0.08 means your IDE gets impounded

PR #3820, opened 2019-03-09
No double reloading defaults, fixes #3808

### LINKED ISSUE #3808: Extension option does not clear the default .js
### Prerequisites

- [ x Checked next-gen ES issues and syntax problems by using the same environment and/or transpiler configuration without Mocha to ensure it isn't just a feature that actually isn't supported in the environment in question or a bug in your code.

### Description

According to [Mocha's own docs](https://mochajs.org/#-extension-ext-watch-extensions-ext) one should be able to specify which extensions mocha will automatically load and that by specifying an extension the default of `*.js` should be cleared. _This does not happen in version 6_ (and I have never seen it happen earlier either).

### Steps to Reproduce
```
git clone fatso83/test-driven-learning
cd javascript/mocha-recursive-extensions-test
npm i
npm run test-bar
```

**Expected behavior:** 
Only this output should be shown, by Mocha loading only the file with the `.bar` extension
```
mocha --extension bar

  normal.bar
    ✓ should do nothing but register the test
```

**Actual behavior:** 
It loads the `normal.bar` - _in addition to_ all the files ending in `.js`.
```
mocha --extension bar

This is not a test: no-test.js


  normal.bar
    ✓ should do nothing but register the test

  normal.bar.js
    ✓ should do nothing but register the test

  normal.js
    ✓ should do nothing but register the test
```

**Reproduces how often:** 
Always

### Versions

6.0.2

dedene:
### Description of the Change

As the mocha defaults were merged early on(https://github.com/mochajs/mocha/blob/master/lib/cli/options.js#L331), it prevented the yargs parser to properly fallback to its defaults configuration for each cli argument (which are coming from the same defaults (https://github.com/mochajs/mocha/blob/master/lib/cli/run.js#L26). With this fix 'js' is no longer always included in the list of extensions.

This fixes issue #3808

### Benefits

Allow mocha to properly run i.e. in Typescript based projects.

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -322,14 +322,7 @@ const loadOptions = (argv = []) => {
     args._ = args._.concat(optsConfig._ || []);
   }
 
-  args = parse(
-    args._,
-    args,
-    rcConfig || {},
-    pkgConfig || {},
-    optsConfig || {},
-    mocharc
-  );
+  args = parse(args._, args, rcConfig || {}, pkgConfig || {}, optsConfig || {});
 
   // recombine positional arguments and "spec"
   if (args.spec) {


dedene:
The test for `loadOptions` fail on the part when no options are given or missing and they should be filled in by the defaults. But isn't it so when no options are given, the defaults as configured in the yargs config will be taken..
https://github.com/mochajs/mocha/blob/e6542538aa10c6137babf96d9ddc851fc6595ad3/lib/cli/run.js#L26
which has the same effect?

juergba:
I think the bug can be found here: [lib/utils.js](https://github.com/mochajs/mocha/blob/master/lib/utils.js#L573)

`extensions` contains the correct CL-input, but does not use it within the function. `js`is hard coded.

This is probably the reason:
> --extension <ext>, --watch-extensions <ext>
Updated in v6.0.0. Previously --watch-extensions, but now expanded to affect general test file loading behavior. --watch-extensions is now an alias

juergba:
I think we have more than one bug here. Something is wrong with the parsing of the `extension` option as well.

`extension` is of yargs option type = "array[]".  => "run-option-metadata.js"
The way we parse and coerce those "array[]"  => "YARGS_PARSER_CONFIG" = {'combine-arrays': true}
This must be the reason why the default extension which is the last member in the access sequence, is added to the extension array.

@dedene You have to get rid of this default `js` somehow. You can't just delete `mocharc` as you did before. Probably you should delete the default `js` before parsing. After parsing you don't know anymore which soure file provided `js`.

boneskull:
> extension is of yargs option type = "array[]". => "run-option-metadata.js"
> The way we parse and coerce those "array[]" => "YARGS_PARSER_CONFIG" = {'combine-arrays': true}
> This must be the reason why the default extension which is the last member in the > access sequence, is added to the extension array.


This is likely the reason.  It's a regression, then.  Essentially, I think we're going to need a special case that says "when merging the default configuration, if `extension` is defined _and not_ the default value, do not merge the default value into the array."

boneskull:
We *want* "combine arrays" to work in all other cases (AFAIK!) *except* this one.

boneskull:
@zenjoy this is kind of hairy, and there's no easy way to make `yargs-parser` do what we want here--nor is it easy to explain exactly what it is we want to happen.  If you don't want to continue with this PR, please let me know, and I'll pick it up.

dedene:
@juergba @boneskull thanks for the feedback! I don't have much time this week, so feel free to take it over. Shall I close the PR or can you edit and continue on this PR?

juergba:
@boneskull maybe you can use yargs "coerce" option.

When you do it before reducing the array to unique values, the logic is simple:
when "extension.length > 1" then delete one "js". If there is a second "js", it's one the user set intentionally and must not be deleted.

boneskull:
closed in lieu of #3834

PR #3823, opened 2019-03-11
Fix #3822

### LINKED ISSUE #3822: --config does not resolve paths relative to CWD
### Prerequisites

- [ ] Checked that your issue hasn't already been filed by cross-referencing [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20)
- [ ] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend that you _not_ install Mocha globally.

### Description

`.js` config loading with `--config` does not handle relative paths correctly, because it calls `require()` with a relative path.

Here's the problem: https://github.com/mochajs/mocha/blob/master/lib/cli/config.js#L38

### Steps to Reproduce

`cd` to an empty directory.  `touch foo.js`

```
mocha --config foo.js
```

mocha loads config from foo.js

```
/c/Users/cspotcode/Documents/Personal-dev/@mochajs/mocha/lib/cli/config.js:64
    throw new Error(`failed to parse ${filepath}: ${err}`);
    ^

Error: failed to parse foobar.js: Error: Cannot find module 'foo.js'
```

All the time

### Versions

Mocha 6.0.2
WSL WLinux bash

### Additional Information

Issue #3818

cspotcode:
### Requirements

### Applicable issues

Fixes #3822

### DIFF
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -35,7 +35,7 @@ exports.CONFIG_FILES = [
 const parsers = (exports.parsers = {
   yaml: filepath =>
     require('js-yaml').safeLoad(fs.readFileSync(filepath, 'utf8')),
-  js: filepath => require(filepath),
+  js: filepath => require(path.resolve(filepath)),
   json: filepath =>
     JSON.parse(
       require('strip-json-comments')(fs.readFileSync(filepath, 'utf8'))


boneskull:
thanks @cspotcode, but I'm going to close this, as it exchanges one problem for another (and would need tests).

PR #3824, opened 2019-03-11
Upgrade yargs-parser dependency to avoid loading 2 copies of yargs

cspotcode:
### Requirements

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -234,6 +234,18 @@
             "which-module": "^2.0.0",
             "y18n": "^3.2.1 || ^4.0.0",
             "yargs-parser": "^11.1.1"
+          },
+          "dependencies": {
+            "yargs-parser": {
+              "version": "11.1.1",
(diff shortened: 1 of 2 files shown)


cspotcode:
As far as I know, `yargs` does not expose a reference to `yargs-parser`, so if we want to use it directly -- which mocha does -- then we need to declare a dependency on it.

It might make sense for `yargs` to re-export the parser as part of its public API, to simplify these situations.

boneskull:
@plroebuck 	Yeah, that's the reason.  We don't need to "go full yargs" when people use `bin/mocha`, since `bin/_mocha` does.  We _do_ need to parse the arguments (hopefully as little as possible) in order to determine what we do with Node.js flags and tell `bin/_mocha` not to bother with config files.

plroebuck:
So this will be a recurring issue... unless there's a change so that `yargs` re-exports the parser as part of its public API?

cspotcode:
The issue will only recur if a mocha developer changes the yargs or yargs-parser dependency without bumping the other to match.

Issue #3690, opened 2019-01-24
bin/mocha: don't spawn child process unnecessarily


boneskull:
This _may_ or _may not_ be a good idea:

If the user has _not_ passed Node.js-specific or V8-specific arguments, import `main` from `lib/cli/cli.js` and execute it.

If the user _has_ passed Node.js-specific or V8-specific arguments, use current behavior.

Advantages:
- faster startup time

Disadvantages:
- increased potential for inconsistent behavior
- would require us to be much more careful about which executable we're using in integration tests
- would possibly require the same integration tests execute against both executables

This isn't related to process replacement, but that's something else to consider if the hacks aren't too bad.

juergba:
closed, see #3827

PR #3827, opened 2019-03-11
do not fork if no node flags present

boneskull:
This is *likely* a quick win, but there may be unforeseen consequences.

If we don't detect any flags that must be sent to `node`, *don't spawn a child process*; instead, pass the unparsed arguments directly to `lib/cli/cli.js` for handling by `yargs`.

This means two things:

1. Depending on the use-case, there's no extra overhead.  This decreases the time spent running our `node`-specific tests:

    ```shell
    # before
    $ nps test.node  143.66s user 33.77s system 100% cpu 2:56.31 total
    # after
    $ nps test.node  115.04s user 26.46s system 97% cpu 2:25.53 total
    ```

1. Users leveraging debuggers which invoke `node --inspect` (VSCode does this, likely Webstorm too) can now execute `bin/mocha` instead of needing to specify `bin/_mocha`.  

It also means that we're a step closer to killing `bin/_mocha` entirely.  Instead of `bin/mocha` spawning `bin/_mocha`, it can actually just spawn *itself*, since all Node.js flags will be handed off to `node` and `lib/cli/cli.js` will be invoked!

(We'll tackle that in another PR though. We need to discuss how to remove/deprecate `bin/_mocha`.)

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -120,30 +120,34 @@ if (nodeArgs.gc) {
 
 debug('final node args', nodeArgs);
 
-const args = [].concat(
-  unparseNodeFlags(nodeArgs),
-  mochaPath,
-  unparse(mochaArgs, {alias: aliases})
-);
+if (Object.keys(nodeArgs).length) {
+  const args = [].concat(
+    unparseNodeFlags(nodeArgs),
+    mochaPath,
+    unparse(mochaArgs, {alias: aliases})
+  );
 
-debug(`exec ${process.execPath} w/ args:`, args);
+  debug(`exec ${process.execPath} w/ args:`, args);
 
-const proc = spawn(process.execPath, args, {
-  stdio: 'inherit'
-});
-
-proc.on('exit', (code, signal) => {
-  process.on('exit', () => {
-    if (signal) {
-      process.kill(process.pid, signal);
-    } else {
-      process.exit(code);
-    }
+  const proc = spawn(process.execPath, args, {
+    stdio: 'inherit'
   });
-});
 
-// terminate children.
-process.on('SIGINT', () => {
-  proc.kill('SIGINT'); // calls runner.abort()
-  proc.kill('SIGTERM'); // if that didn't work, we're probably in an infinite loop, so make it die.
-});
+  proc.on('exit', (code, signal) => {
+    process.on('exit', () => {
+      if (signal) {
+        process.kill(process.pid, signal);
+      } else {
+        process.exit(code);
+      }
+    });
+  });
+
+  // terminate children.
+  process.on('SIGINT', () => {
+    proc.kill('SIGINT'); // calls runner.abort()
+    proc.kill('SIGTERM'); // if that didn't work, we're probably in an infinite loop, so make it die.
+  });
+} else {
+  require('../lib/cli/cli').main(unparse(mochaArgs, {alias: aliases}));
+}
--- a/bin/mocha
+++ b/bin/mocha
@@ -3,14 +3,14 @@
 'use strict';
 
 /**
- * This wrapper executable checks for known node flags and appends them when found, before invoking the "real" _mocha(1) executable.
+ * This wrapper executable checks for known node flags and appends them when found,
+ * before invoking the "real" executable (`lib/cli/cli.js`)
  *
  * @module bin/mocha
  * @private
  */
 
 const {deprecate, warn} = require('../lib/utils');
-const {spawn} = require('child_process');
 const {loadOptions} = require('../lib/cli/options');
 const {
   unparseNodeFlags,
@@ -22,7 +22,6 @@ const debug = require('debug')('mocha:cli:mocha');
 const {aliases} = require('../lib/cli/run-option-metadata');
 const nodeEnv = require('node-environment-flags');
 
-const mochaPath = require.resolve('./_mocha');
 const mochaArgs = {};
 const nodeArgs = {};
 
@@ -118,9 +117,12 @@ if (nodeArgs.gc) {
   delete nodeArgs.gc;
 }
 
-debug('final node args', nodeArgs);
-
 if (Object.keys(nodeArgs).length) {
+  const {spawn} = require('child_process');
+  const mochaPath = require.resolve('../lib/cli/cli.js');
+
+  debug('final node args', nodeArgs);
+
   const args = [].concat(
     unparseNodeFlags(nodeArgs),
     mochaPath,
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 'use strict';
 
 /**


plroebuck:
> Users leveraging debuggers which invoke node --inspect (VSCode does this, likely Webstorm too) can now execute bin/mocha instead of needing to specify bin/_mocha.

But you already know this is **not** the case with WebStorm per the issue with multiple `--timeout` values...

plroebuck:
After _years_ of telling people to execute "nyc" against "/bin/_mocha", getting rid of it seems more like a pipedream.

bin/mocha
```diff
@@ -120,30 +120,34 @@ if (nodeArgs.gc) {
 
 debug('final node args', nodeArgs);
 
-const args = [].concat(
-  unparseNodeFlags(nodeArgs),
-  mochaPath,
-  unparse(mochaArgs, {alias: aliases})
-);
+if (Object.keys(nodeArgs).length) {
```
plroebuck: Move `mochaPath` (L25) here.

bin/mocha
```diff
@@ -120,30 +120,34 @@ if (nodeArgs.gc) {
...
+  const args = [].concat(
+    unparseNodeFlags(nodeArgs),
+    mochaPath,
+    unparse(mochaArgs, {alias: aliases})
+  );
 
-debug(`exec ${process.execPath} w/ args:`, args);
+  debug(`exec ${process.execPath} w/ args:`, args);
```
plroebuck: Move `{spawn}` (L13) here

boneskull:
@plroebuck This PR does not get rid of `bin/_mocha`.  That's another issue entirely, which would require a deprecation notice for some time first.

boneskull:
@plroebuck 	

> But you already know this is not the case with WebStorm per the issue with multiple --timeout values...

I don't understand how that has anything to do with this pr.

boneskull:
TO BE CLEAR: this does not do anything with `bin/_mocha`; it's still there, it still works, we just don't spawn it if we don't have to.

craigtaub:
Do we lose any fault tolerance by having tests run in current and not child process?
Is keeping them for Node flags just a case of defending against scenarios/code out of our direct control? (if this was the reason it was introduced in the first place).

boneskull:
> But we should be cautious, I would tend to semver-major. Is it possible users created a concurrent wrapper around Mocha, using these child processes?

Anybody wrapping Mocha in some way is likely going to be consuming `bin/_mocha` or using it programmatically.  This change won't affect those users.

Signal handling is going to be the main area of concern; I'll need to do some manual testing with `--watch` to ensure that it's not broken.

boneskull:
> In the original issue #3690 you mentioned our intgration tests. I wouldn't like to do all tests twice, once for single main process, and a second time for parent/child szenario.

The fact is that the option parsing flow is not the same.  If you run `bin/mocha`, we do a naive parse of the arguments to separate them into "Node.js" and "Mocha" buckets, then recombine these into an array.  No validation occurs.  Config file parsing is disabled via flags (because we've already read them at this point) and either (given this PR) a) the child process is spawned if we have a non-empty Node.js bucket, or b) the main entry function is invoked with the new arguments array. 

While I'm *confident* that this doesn't produce any weirdness in and of itself, I'm also not 100% confident.  Yet, at this point, I don't think running the same integration tests twice against both executables would buy us much.  

I think a test that would run the `bin/mocha` executable, then invoke the command handler, then compare this to the final parsed options when running `bin/_mocha` should be sufficient.  

That's tangential to this PR though, since if this didn't land, we'd still want it--just instead of invoking the command handler, we'd spawn the child `bin/_mocha`  process.

boneskull:
@craigtaub 	Could you be more specific?

juergba:
> This is likely a quick win, but there may be unforeseen consequences.

@boneskull We could add a new flag like `--experimental-no-child`. A default value of `false` would keep the current behaviour, `true` plus no `node` flags would run this PR's behavior.

We should implemet this PR, it's a good idea.

bin/mocha
```diff
@@ -118,32 +117,39 @@ if (nodeArgs.gc) {
...
+    stdio: 'inherit'
   });
-});
 
-// terminate children.
-process.on('SIGINT', () => {
-  proc.kill('SIGINT'); // calls runner.abort()
-  proc.kill('SIGTERM'); // if that didn't work, we're probably in an infinite loop, so make it die.
-});
+  proc.on('exit', (code, signal) => {
```
juergba: The cb of the `exit` event takes only one parameter. `signal` should always be `undefined`.

bin/mocha
```diff
@@ -118,32 +117,39 @@ if (nodeArgs.gc) {
...
+    });
+  });
+
+  // terminate children.
+  process.on('SIGINT', () => {
+    proc.kill('SIGINT'); // calls runner.abort()
+    proc.kill('SIGTERM'); // if that didn't work, we're probably in an infinite loop, so make it die.
+  });
+} else {
+  require('../lib/cli/cli').main(unparse(mochaArgs, {alias: aliases}));
```
juergba: It does not make sense to unparse here and then re-parse again in "lib/cli/cli".
This unparse/re-parse thing could probably be skipped when no child process is spawned.

bin/mocha
```diff
@@ -118,32 +115,41 @@ if (nodeArgs.gc) {
   delete nodeArgs.gc;
 }
 
-debug('final node args', nodeArgs);
+if (Object.keys(nodeArgs).length) {
+  const {spawn} = require('child_process');
+  const mochaPath = require.resolve('../lib/cli/cli.js');
```
juergba: For my understanding:
now we spawn "lib/cli/cli.js" instead of "bin/_mocha". Right?
Both ways do fine and pass our `npm test`.

boneskull: yes. I would prefer to "standardize" on `lib/cli/cli.js` and replace all usages of `bin/_mocha` with this, essentially

plroebuck:
Looked into this problem generally a long while back.
Preferred the direction taken by `npx`, although it used [this](https://github.com/zkat/npx/blob/c803bd71b0ecdab9c653615a338904791be60937/index.js#L268).

bin/mocha
```diff
@@ -151,5 +151,5 @@ if (Object.keys(nodeArgs).length) {
     proc.kill('SIGTERM'); // if that didn't work, we're probably in an infinite loop, so make it die.
   });
 } else {
-  require('../lib/cli/cli').main(unparse(mochaArgs, {alias: aliases}));
+  require('../lib/cli/cli').main(['preParsedOptions', mochaArgs]);
```
boneskull: this feels like a pretty strange way to get this done.  I'd prefer a more "designed" approach to this, and one that can be tested.

at this point, it feels like premature optimization.  the overhead is *forking the child process*, which this PR avoids.  avoiding double-parsing will be much less of a perf gain.  either way, it can be addressed in another PR...  how would you feel about reverting this and creating another PR for it?

juergba: @boneskull  It's ok, I will revert it.

@plroebuck which is `npx`s solution? This code is not self-explanatory for me. I can't imagine an alternative way without spawning. Before launching Node, we don't know wether there are Node options or not.

plroebuck:
Note concerning existing code, not PR change itself...

Existing code doesn't give child process time to terminate before
sending second signal.

```js
  const childProcess = spawn(process.execPath, args, {
    stdio: 'inherit'
  });

  childProcess.once('exit', (code, signal) => {
    process.once('exit', () => {
      if (signal) {
        process.kill(process.pid, signal);
      } else {
        process.exit(code);
      }
    });
    childProcess = null;
  });

  // Terminate child process on keyboard abort
  process.once('SIGINT', () => {
    childProcess.kill('SIGINT'); // calls runner.abort()
    setTimeout(() => {
      if (childProcess) {
        // If _still_ not dead, try again with alternate signal...
        childProcess.kill('SIGTERM');
      }
    }, 5000);
  });
} else {
...
```

PR #3830, opened 2019-03-12
Replace findup-sync with find-up for faster startup

cspotcode:
### Requirements

### Description of the Change

This PR replaces findup-sync with find-up, which reduces mocha's startup time. Related to #3743.

An informal benchmark shows it reduces startup time by about 30%:
https://github.com/cspotcode/repros/tree/mocha-find-up-perf
https://github.com/cspotcode/repros/runs/76724994 (pardon the verbose logging; you can search for the word "real" to jump to the `time` invocations)

### Alternate Designs

I considered using cosmiconfig, but this is a more minimal change that preserves mocha's existing diagnostic logging.  https://github.com/mochajs/mocha/issues/3743#issuecomment-466132812

Startup time for a CLI tool like mocha is important, especially considering that mocha may spawn an extra node process at startup.

### Benefits

Fast mocha startup; better developer experience.

### Possible Drawbacks

`package` flag does not support globs.  I don't think anyone intended or was relying on this behavior, and removing it simplifies mocha's behavior, which should make it easier to debug and triage issues related to `--package` loading.

### DIFF
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -9,7 +9,7 @@
  */
 
 const fs = require('fs');
-const findUp = require('findup-sync');
+const findUp = require('find-up');
 const path = require('path');
 const debug = require('debug')('mocha:cli:config');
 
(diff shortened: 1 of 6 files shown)


cspotcode:
Apparently this also reduces the installed size from 17 to 12MB.  That wasn't my goal, but it's a nice side-effect.

*From the github actions log:*
```
/github/workspace
17M	./mocha/node_modules
16800	./mocha/node_modules
12M	./mocha-proposed/node_modules
12276	./mocha-proposed/node_modules
```

boneskull:
LGTM, thanks!

cspotcode:
Whoops!  Fixed.

boneskull:
looks like there was some netlify flake.  seems to work now.

PR #3831, opened 2019-03-12
fix timeout/slow string values and duplicate arguments

boneskull:
This fixes two problems:

1. ~~Duplicate values are no longer allowed on the command line, and an error will be thrown (e.g., `--timeout 0 --timeout 0`)~~
1. String values of `timeout` and `slow` (e.g., `1s`, `500ms`) are no longer evaluated to `NaN` (because their type was set to `number`).  It appears that a value of `NaN` would cause timeouts to be disabled.

Ref: #3817 

UPDATE: Duplicate arguments are now allowed; the last value is chosen.

### DIFF
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -258,6 +258,13 @@ exports.builder = yargs =>
         }
       });
 
+      types.boolean
+        .concat(types.string, types.number)
+        .filter(opt => Array.isArray(argv[opt]))
+        .forEach(opt => {
+          argv[opt] = argv[opt].pop();
+        });
+
       // yargs.implies() isn't flexible enough to handle this
       if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
         throw createMissingArgumentError(
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -258,13 +258,6 @@ exports.builder = yargs =>
         }
       });
 
-      types.boolean
-        .concat(types.string, types.number)
-        .filter(opt => Array.isArray(argv[opt]))
-        .forEach(opt => {
-          argv[opt] = argv[opt].pop();
-        });
-
       // yargs.implies() isn't flexible enough to handle this
       if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
         throw createMissingArgumentError(
--- a/test/integration/fixtures/options/slow-test.fixture.js
+++ b/test/integration/fixtures/options/slow-test.fixture.js
@@ -1,0 +1,11 @@
+'use strict';
+
+describe('a suite', function() {
+  it('should succeed in 500ms', function(done) {
+    setTimeout(done, 500);
+  });
+
+  it('should succeed in 1.5s', function(done) {
+    setTimeout(done, 1500);
+  });
+});
--- a/test/integration/invalid-arguments.spec.js
+++ b/test/integration/invalid-arguments.spec.js
@@ -3,18 +3,20 @@
 var invokeMocha = require('./helpers').invokeMocha;
 
 describe('invalid arguments', function() {
-  it('should exit with failure if arguments are invalid', function(done) {
-    invokeMocha(
-      ['--ui'],
-      function(err, result) {
-        if (err) {
-          return done(err);
-        }
-        expect(result, 'to have failed');
-        expect(result.output, 'to match', /not enough arguments/i);
-        done();
-      },
-      {stdio: 'pipe'}
-    );
+  describe('when argument is missing required value', function() {
+    it('should exit with failure', function(done) {
+      invokeMocha(
+        ['--ui'],
+        function(err, result) {
+          if (err) {
+            return done(err);
+          }
+          expect(result, 'to have failed');
+          expect(result.output, 'to match', /not enough arguments/i);
+          done();
+        },
+        {stdio: 'pipe'}
+      );
+    });
   });
 });
--- a/test/integration/options/timeout.spec.js
+++ b/test/integration/options/timeout.spec.js
@@ -1,0 +1,52 @@
+'use strict';
+
+var helpers = require('../helpers');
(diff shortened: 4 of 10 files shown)


plroebuck:
> Duplicate values are no longer allowed on the command line, and an error will be thrown (e.g., --timeout 0 --timeout 0)

This is a completely valid cmdline argument usage -- last value wins.

juergba:
> This is a completely valid cmdline argument usage -- last value wins. It's not an error.

Agree. Sometimes it's even necessary to overwrite flags, i.e. ones set by IDE's.

lib/cli/run.js
```diff
@@ -258,6 +258,16 @@ exports.builder = yargs =>
...
+      types.boolean.concat(types.string, types.number).forEach(opt => {
+        if (Array.isArray(argv[opt])) {
+          throw createInvalidArgumentValueError(
+            `Argument "${opt}" can only be provided once`,
+            opt,
+            argv[opt]
+          );
+        }
+      });
+
```
plroebuck: Disagree. If `Array`, convert back to original type using last value of `argv[opt]`
(i.e., pretend `yargs-parser.duplicate-arguments-array = false` for _just_ these types).

plroebuck: Further, convert `slow` and `timeout` back to numeric using `ms`.

test/integration/fixtures/options/slow-test.fixture.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+describe('a suite', function() {
+  it('should succeed, slowly', function(done) {
```
plroebuck:
```js
  it('should succeed within half second', function(done) {
```

test/integration/fixtures/options/slow-test.fixture.js
```diff
@@ -0,0 +1,11 @@
...
+    setTimeout(done, 500);
+  });
+
+  it('should succeed, very slowly', function(done) {
```
plroebuck:
```js
  it('should succeed within 1.5 seconds', function(done) {
```

boneskull:
> This is a completely valid cmdline argument usage -- last value wins. It's not an error.

Hm, ok. Assuming the array yargs-parser gives us is *in order*, then I can just pop it.

lib/cli/run.js
boneskull: This happens within `Suite#timeout`, so I didn't do it here.

boneskull:
@juergba 	@plroebuck 	I've changed this to use the _last_ value instead of throwing an exception.

test/integration/options/timeout.spec.js
```diff
@@ -0,0 +1,51 @@
...
+        return;
+      }
+      expect(res, 'to have failed')
+        .and('to have passed test count', 1)
+        .and('to have failed test count', 1);
+      done();
+    });
+  });
+
+  it('should allow numeric value', function(done) {
```
plroebuck: duplicate test?

lib/cli/run.js
```diff
@@ -258,6 +258,13 @@ exports.builder = yargs =>
         }
       });
```
plroebuck:
```js
      // For any option given multiple times, use value of its array's last element
      types.boolean
```

test/integration/invalid-arguments.spec.js
```diff
@@ -3,18 +3,20 @@
...
-        if (err) {
-          return done(err);
-        }
-        expect(result, 'to have failed');
-        expect(result.output, 'to match', /not enough arguments/i);
-        done();
-      },
-      {stdio: 'pipe'}
-    );
+  describe('when argument is missing required value', function() {
```
plroebuck: Said this before and provided code for it.
This should be run for _all_ arguments expecting a value, not just `--ui`.

juergba:
I haven't understood this "yargs" / "yargs-parser" thing completely, yet. 
- why can't "yargs-parser" use the default values defined in [lib/cli/run.js](https://github.com/mochajs/mocha/blob/4601d000f112e5d455d73c2a16bc4e96a08eaa09/lib/cli/run.js#L46)
- couldn't we use the `duplicate-arguments-array` option to reduce duplicate arguments?
```js
var parse = require('yargs-parser');

var args = parse(
    ['--timeout', '2s', '--timeout', '1s', '--timeout', '800'], {
        string: ['timeout'],
        configuration: {
            'duplicate-arguments-array': false,
            'camel-case-expansion': false
        }
});
console.log(args);         // { _: [], timeout: '800' }
```

lib/cli/run.js
```diff
@@ -258,6 +258,13 @@ exports.builder = yargs =>
...
 
+      types.boolean
+        .concat(types.string, types.number)
+        .filter(opt => Array.isArray(argv[opt]))
+        .forEach(opt => {
+          argv[opt] = argv[opt].pop();
+        });
+
```
juergba: This part does not work correctly:
- input per CLI: `--slow 5 --slow 10`
- result after parsing: `s: ["5"], slow: "10"`

juergba:
There is a bug in "yargs-parser" where the coerce function returns a value of type `number` as `string`. I opened a issue for that.

This affects the parsing of our `--retries` option and I had to adapt some of our tests. Nevertheless there are no implications since `Suite#retries` converts the value to `number`.

Ready for review.

juergba:
I have chosen `yargs-parser#coerce()` in order to reduce duplicate arguments:
- it reduces the flags **incl. all its aliases**
- it is part of `loadOptions()` which is a public function. This is not the case with `yargs.coerce()`. Arguments should be reduced when a third party package depends on Mocha's `loadOptions()`.

plroebuck:
Upstream fix for yargs-parser unnecessary then?

juergba:
For our purposes, yes unnecessary.

juergba:
I will merge these changes tomorrow.

PR #3833, opened 2019-03-14
remove yargs-unparser; run yargs only once

cspotcode:
### Requirements

### Description of the Change

This removes the yargs-unparser dependency and avoids loading yargs within `_mocha` unless necessary.  I still need to double-check these results, but it seems to reduce the installed size from 12M to about 6.6M.

Faster startup and install time; simpler runtime behavior because yargs is only running once.

### Benefits

Faster startup and install time; simpler runtime behavior because yargs is only running once.

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -11,22 +11,20 @@
 
 const {deprecate, warn} = require('../lib/utils');
 const {spawn} = require('child_process');
-const {loadOptions} = require('../lib/cli/options');
+const {parseArgv} = require('../lib/cli/cli');
 const {
   unparseNodeFlags,
   isNodeFlag,
   impliesNoTimeouts
 } = require('../lib/cli/node-flags');
-const unparse = require('yargs-unparser');
 const debug = require('debug')('mocha:cli:mocha');
-const {aliases} = require('../lib/cli/run-option-metadata');
 const nodeEnv = require('node-environment-flags');
 
 const mochaPath = require.resolve('./_mocha');
 const mochaArgs = {};
 const nodeArgs = {};
 
-const opts = loadOptions(process.argv.slice(2));
+const opts = parseArgv(process.argv.slice(2));
 debug('loaded opts', opts);
 
 /**
@@ -123,7 +121,8 @@ debug('final node args', nodeArgs);
 const args = [].concat(
   unparseNodeFlags(nodeArgs),
   mochaPath,
-  unparse(mochaArgs, {alias: aliases})
+  '--preParsedJsonOptions',
+  JSON.stringify(mochaArgs)
 );
 
 debug(`exec ${process.execPath} w/ args:`, args);
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -11,10 +11,7 @@
 
 const debug = require('debug')('mocha:cli:cli');
 const symbols = require('log-symbols');
-const yargs = require('yargs');
 const path = require('path');
-const {loadOptions, YARGS_PARSER_CONFIG} = require('./options');
-const commands = require('./commands');
 const ansi = require('ansi-colors');
 const {repository, homepage, version, gitter} = require('../../package.json');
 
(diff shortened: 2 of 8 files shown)


boneskull:
@cspotcode did you disable the pre-commit hooks?  eslint should have fixed the lint failures automatically (assuming it could)

cspotcode:
I didn't run eslint, no.  I still need to cleanup this PR before it's ready to merge; I'll fix the linter failures and squash all those commits.

cspotcode:
This is ready for review.  I didn't need the base64 encoding on Windows, so I moved that to another branch just in case it's relevant in the future.

Installed size drops by about half, with more modest improvements in startup time.

bin/mocha
```diff
@@ -123,7 +121,8 @@ debug('final node args', nodeArgs);
 const args = [].concat(
   unparseNodeFlags(nodeArgs),
   mochaPath,
-  unparse(mochaArgs, {alias: aliases})
+  '--preParsedJsonOptions',
+  JSON.stringify(mochaArgs)
 );
```
juergba: This way duplicate alias options don't get eliminated. With `unparse()` we get a clean array without any aliases, so I'm not sure wether we can do without  `unparse()` as easily.

It's a very good idea not to parse a second time in "/cli.cli.js".

lib/cli/cli.js
```diff
@@ -32,7 +29,33 @@ exports.main = (argv = process.argv.slice(2)) => {
...
+  args.command = null;
+  handler(args);
+};
+
+exports.yargsParse = function(argv) {
+  const {loadOptions} = require('./options');
+  return exports
+    .createYargs()
+    .parse(argv, Object.assign(loadOptions(argv), {command: {}}));
+};
```
juergba: I haven't understood yet (neither in the current master) why the object returned by `loadOptions()` is passed as `context` parameter to `parse()`.
- the coerce functions are not called that way, see #3826 
- duplicate alias option don't get eliminated

bin/mocha
cspotcode: I don't think it's necessary to remove duplicate alias options using this method.  They won't pass through yargs' parser again; the object is passed straight to mocha.

When mocha v6 re-parses the output of yargs-unparser, it creates those duplicate aliases all over again.  So either way, the output object contains aliases, and mocha is built to handle it.

lib/cli/cli.js
cspotcode: Yeah, sounds like a separate bug.

bin/mocha
juergba: Yes, I think the `mocha` instance looks quite messy with all those aliases, but it seems to be able to handle them correctly.

> [...]; the object is passed straight to mocha.

This could be to early, since the coerce function of `reporter-option` hasn't run yet?

lib/cli/cli.js
juergba: Even `node` flags are allowed in config files. When run with `_mocha` they are useless, I don't think they get filtered out, probably they are just ignored.

juergba:
I tried to fix that separate bug mentioned above: [juergba/issue/3826](https://github.com/mochajs/mocha/tree/juergba/issue/3826).

You will not like it, because I'm using "yargs-unparser". Maybe you can merge my fix into this PR, otherwise I would open a seperate PR.

cspotcode:
The proper solution is to stop using yargs for option processing that does
not apply exclusively to argv positionals.  Using unparse is masking this
deeper code smell.

If yargs' API is only running coerce functions against values from argv,
then the coerce functionality is *not* intended for loading options from a
config file.  We should not use it for that purpose.  Instead, the coerce
functions should be moved outside of yargs and executed as a post-process
step.  If there is coerce logic that applies exclusively to argv values and
not to values from a config file, then it should be kept in yargs coerce
functions.

juergba:
We use "yargs-parser" to read CLI-argv plus all config files. Mocha has its own wrapper function `parse()` for this purpose. The resulting totalised argv is then passed to "yargs" to run the final coercing.

`Yargs#parse([args], [context], [parseCallback])`
IMO the seperate bug results, because the config object is passed as "context" parameter. This is wrong, or at least I haven't understood the reason.

soryy708:
Looks like `yargs-unparser` (1.5.0) depends on `lodash` (4.17.11) and GitHub claims there's a [prototype pollution vulnerability](https://github.com/lodash/lodash/pull/4336) in versions < 4.17.13.
Removing the `yargs-unparser` dependency not only optimizes, but also fixes a security vulnerability.

soryy708:
@cspotcode You've got merge conflicts.

cspotcode:
@soryy708 If the maintainers say they want to merge this change, I'll resolve the merge conflicts.  Until then, it's not worth the potentially wasted effort.

PR #3834, opened 2019-03-14
fix extension and spec handling; closes #3808

### LINKED ISSUE #3808: Extension option does not clear the default .js

boneskull:
Also fixes some issues in the options unit tests and adds coverage around `spec` handling.

This needs some manual testing, as I had to monkey with `watch` a bit.

cc @dedene

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -10,7 +10,8 @@ const fs = require('fs');
 const yargsParser = require('yargs-parser');
 const {types, aliases} = require('./run-option-metadata');
 const {ONE_AND_DONE_ARGS} = require('./one-and-dones');
-const mocharc = require('../mocharc.json');
+// paranoia
+const mocharc = Object.freeze(require('../mocharc.json'));
 const {list} = require('./run-helpers');
 const {loadConfig, findConfig} = require('./config');
 const findUp = require('find-up');
@@ -321,6 +322,30 @@ const loadOptions = (argv = []) => {
     args.opts = false;
     args._ = args._.concat(optsConfig._ || []);
   }
+  // special case: "extension" option should not combine with default value.
+  // normally we want to combine "array"-type options, and we _do_ with "extension", but only
+  // within user-defined configuration (args or anything else).
+  // we must also search through any aliases of "extension" because while the arguments are ///
+  // normalized by this point, the config file values are not.
+  // only the "canonical" option name is used in `mocharc`, so we needn't worry about clearing
+  // multiple options.
+  // NOTE: as of this writing, "extension" is the only default value which is of an "array" type;
+  // it's unknown whether the the below strategy should be generalized to any other future
+  // "array"-type default option.
+  const processedMocharc = Object.assign({}, mocharc);
+  if (
+    args.extension ||
+    ['extension']
+      .concat(aliases.extension)
+      .some(
+        opt =>
+          Object.hasOwnProperty(rcConfig, opt) ||
+          Object.hasOwnProperty(pkgConfig, opt) ||
+          Object.hasOwnProperty(optsConfig, opt)
+      )
+  ) {
+    delete processedMocharc.extension;
+  }
 
   args = parse(
     args._,
@@ -328,7 +353,7 @@ const loadOptions = (argv = []) => {
     rcConfig || {},
     pkgConfig || {},
     optsConfig || {},
-    mocharc
+    processedMocharc
   );
 
   // recombine positional arguments and "spec"
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -10,8 +10,7 @@ const fs = require('fs');
 const yargsParser = require('yargs-parser');
 const {types, aliases} = require('./run-option-metadata');
 const {ONE_AND_DONE_ARGS} = require('./one-and-dones');
-// paranoia
-const mocharc = Object.freeze(require('../mocharc.json'));
+const mocharc = require('../mocharc.json');
 const {list} = require('./run-helpers');
 const {loadConfig, findConfig} = require('./config');
 const findUp = require('find-up');
@@ -81,11 +80,12 @@ const nargOpts = types.array
 /**
  * Wrapper around `yargs-parser` which applies our settings
  * @param {string|string[]} args - Arguments to parse
+ * @param {Object} defaultValues - Default values of mocharc.json
  * @param  {...Object} configObjects - `configObjects` for yargs-parser
  * @private
  * @ignore
  */
-const parse = (args = [], ...configObjects) => {
+const parse = (args = [], defaultValues = {}, ...configObjects) => {
   // save node-specific args for special handling.
   // 1. when these args have a "=" they should be considered to have values
   // 2. if they don't, they just boolean flags
@@ -110,6 +110,7 @@ const parse = (args = [], ...configObjects) => {
   const result = yargsParser.detailed(args, {
     configuration,
     configObjects,
+    default: defaultValues,
     coerce: coerceOpts,
     narg: nargOpts,
     alias: aliases,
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -575,9 +575,15 @@ exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
...
-    } else {
+    // check all extensions
+    if (
+      !extensions.some(function(ext) {
+        if (fs.existsSync(filepath + '.' + ext)) {
+          filepath += '.' + ext;
+          return true;
+        }
+      })
+    ) {
```
plroebuck: The _only_ reason this was here was to allow the historical usage of "test.js" via the default spec "test". IMO, we should **not** propagate this further.

The "Handle directory" code below already handles extensions for "test/*".

lib/cli/options.js
```diff
@@ -10,7 +10,8 @@ const fs = require('fs');
 const yargsParser = require('yargs-parser');
 const {types, aliases} = require('./run-option-metadata');
 const {ONE_AND_DONE_ARGS} = require('./one-and-dones');
-const mocharc = require('../mocharc.json');
+// paranoia
+const mocharc = Object.freeze(require('../mocharc.json'));
```
plroebuck:
```js
const rcDefaults = Object.freeze(require('../mocharc.json'));
```

lib/cli/options.js
```diff
@@ -321,14 +322,38 @@ const loadOptions = (argv = []) => {
...
+  // special case: "extension" option should not combine with default value.
+  // normally we want to combine "array"-type options, and we _do_ with "extension", but only
+  // within user-defined configuration (args or anything else).
+  // we must also search through any aliases of "extension" because while the arguments are ///
+  // normalized by this point, the config file values are not.
+  // only the "canonical" option name is used in `mocharc`, so we needn't worry about clearing
+  // multiple options.
+  // NOTE: as of this writing, "extension" is the only default value which is of an "array" type;
+  // it's unknown whether the the below strategy should be generalized to any other future
+  // "array"-type default option.
```
plroebuck: Use [block comment style](https://github.com/airbnb/javascript#comments--multiline) here.

lib/cli/options.js
```diff
@@ -321,14 +322,38 @@ const loadOptions = (argv = []) => {
...
+  const processedMocharc = Object.assign({}, mocharc);
```
plroebuck: Deep clone, or see if this (e.g., ES7 `_.omit`) is supported...
```js
  let key = 'extension';
  const { [key]: _ignore, ...processedMocharc } = mocharc;
```

`Object.assign` here does shallow clone (after shallow freeze).
Surprised the `delete` didn't throw `TypeError`.

plroebuck:
Rebase

juergba:
I think there is a standard yargs way to achieve this, using the "default" option. I just did some short testing, nevertheless it seems to work.
```js
// line 109
const result = yargsParser.detailed(args, {
    default: {extension: ["js"]},      // set default values with "mocharc"
    configuration,
    configObjects,
    coerce: coerceOpts,
    narg: nargOpts,
    alias: aliases,
    string: types.string,
    array: types.array,
    number: types.number,
    boolean: types.boolean.concat(nodeArgs.map(pair => pair[0]))
  });
```
```js
// line 326
args = parse(
    args._,
    args,
    rcConfig || {},
    pkgConfig || {},
    optsConfig || {},
    //mocharc                           don't set default values here
  );
```

boneskull:
@juergba Hmm, you're right.

I think instead of this strategy, I'll change it so that we can still keep the default config in our `lib/mocharc.json`, but use its values as defaults instead of just another config object which is merged in.

juergba:
Also in "lib/cli/run-helpers.js" the default extension of "js" is set hardcoded several times.
i.e. `watchRun()` or `runMocha()`.

juergba:
@boneskull is it ok for you, if I work on this one?

boneskull:
@juergba yes, go ahead

plroebuck:
@juergba second time now you merged without any review (despite prior request to do otherwise) and this time you introduced bug...

plroebuck:
Corrected code...


**"lib/utils.js"**
```js
/**
 * Lookup file names at the given `path`.
 *
 * @description
 * Filenames are returned in _traversal_ order by the OS/filesystem.
 * **Make no assumption that the names will be sorted in any fashion.**
 *
 * @public
 * @memberof Mocha.utils
 * @param {string} filepath - Base path to start searching from.
 * @param {string[]} [extensions=[]] - File extensions to look for.
 * @param {boolean} [recursive=false] - Whether to recurse into subdirectories.
 * @return {string[]} An array of file pathnames.
 * @throws {TypeError} if `filepath` is directory and `extensions` not provided or empty.
 */
exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
  var files = [];
  var stat;

  if (extensions === undefined) {
    extensions = [];
  }
  if (recursive === undefined) {
    recursive = false;
  }

  if (!fs.existsSync(filepath)) {
    // Check all extensions, using first match found (if any)
    if (
      !(extensions.some(function(ext) {
        var filepathWithExtname = filepath + '.' + ext;
        if (fs.existsSync(filepathWithExtname)) {
          filepath = filepathWithExtname;
          return true;
        }
      })
    ) {
      // Handle glob
      files = glob.sync(filepath);
      if (!files.length) {
        throw createNoFilesMatchPatternError(
          'Cannot find any files matching pattern ' + exports.dQuote(filepath),
          filepath
        );
      }
      return files;
    }
  }

  // ...no further changes to merged code...
}
```

juergba:
@plroebuck 
Yes, without @boneskull contributions to "needs-review"-PRs have been close to null during the last few weeks. There is no one stepping into his gap, that's worrying.

`you introduced bug...`  can you be more ... specific?

`corrected code`: this is the solution @boneskull already proposed, so nothing new. It only uses the first match of `filepath + '.' + ext` and ignores the rest of matches. I wonder why this should be correct.

juergba:
> As I commented originally in his PR [...]

You are right, a directory named "test.js" will fail. I will open a PR and use your proposition. I don't agree completely, though. With two files (eg. "unit.ts" and "unit.js") a call "unit" plus extension: ['ts', 'js'] should return both files, not just some/the first one.
I will also have to check wether the glob always returns files.

> True, but almost no PRs here [...]

I can live with that procedure.

plroebuck:
> You are right, a directory named "test.js" will fail. I will open a PR and use your proposition. I don't agree completely, though. With two files (eg. "unit.ts" and "unit.js") a call "unit" plus extension: ['ts', 'js'] should return both files, not just some/the first one.

The `filepath + '.js'` _hack_ was **never** intended to handle what you're trying to do with it. Its purpose was to allow small packages that only exported a couple functions (typically from "index.js") to use "test.js" to test them.

If a package is exporting functions in more than one language, its specifications _should_ live in the "test" directory -- the simplistic specification scheme above shouldn't apply.

PR #3835, opened 2019-03-14
Fix this.skip() async calls: "self.test" undefined in Browser

juergba:
### Description of the Change

In addition to #3745 

> it also appears we need this around 390
```js
        if (err instanceof Pending) {
          if (name === HOOK_TYPE_BEFORE_EACH || name === HOOK_TYPE_AFTER_EACH) {
            if (self.test) {
              self.test.pending = true;
            }
          } else {
```
> When Pending is thrown async like this, we might no longer have a test, which causes its own exception.


plroebuck:
How did you add an `if` condition _and_ gain a coverage increase from Coveralls?

juergba:
It's the way a true genius works...

PR #3850, opened 2019-03-27
Remove pound icon showing on header hover on docs #3714

### LINKED ISSUE #3714: duplicated header anchor links on mouseover
1. Go to https://next.mochajs.org
1. Mouseover a heading

image

Remove the `#`; keep the other one.

cc @Munter

jd2rogers2:
### Requirements

### Description of the Change

cannot pass nothing to `markdown-it-anchor` so passing span
min-height on span tag or defaults to 0 and svg doesn't show

### Alternate Designs

could remove svg all together and pass link icon but then it looks too far to the left

just a doc patch

### Benefits

cleaner look

### Possible Drawbacks

pretty sure span won't effect anything but i could be wrong

### Applicable issues

bug fix #3714

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -141,6 +141,7 @@ a.direct-link {
   background: url(../images/link-icon.svg) center center no-repeat;
   background-size: auto 60%;
   opacity: 0;
+  overflow: hidden;
   position: absolute;
   text-decoration: none;
   text-indent: -60px;


Munter:
Thanks for the PR.

Having had a closer look at the problem, I believe the solution is not to take any of the patches you applied here, but instead add an `overflow: hidden` to the `a.direct-link` block. If you are up for making this change instead I'd be happy to review the result again

docs/css/style.css
```diff
@@ -146,6 +146,7 @@ a.direct-link {
   text-indent: -60px;
   transform: translateX(-100%);
   width: 30px;
+  min-height: 1.55em;
```
Munter: I think hiding the negatively indented text with an `overflow` setting is probably the correct way to go

```suggestion
  overflow: hidden;
```

jd2rogers2: @Munter that worked, was able to avoid making the change in .eleventy js (changing line 24 from `    permalinkSymbol: '#'` to `    permalinkSymbol: ' '`) as well. lmk if you still wanted that made

docs/css/style.css
```diff
@@ -146,6 +146,7 @@ a.direct-link {
   text-indent: -60px;
   transform: translateX(-100%);
   width: 30px;
+  overflow: hidden;
```
plroebuck: Move to its lexical position (L143.5)

jd2rogers2: done, thanks!

Munter:
This looks good to me.

Thank you very much for the contribution!

jd2rogers2:
@Munter i'm still seeing this bug in production despite this being merged to master. is there another level of deployment before it would be live?

Issue #332, opened 2012-03-19
skip suite and test without failing


ehartford:
I would like some function like skipTest() and skipSuite() that would allow me to skip a test or suite without failing it.

If you are in suiteSetup() then you could skip the suite
if you are in test() then you could skip the rest of the suite or the test
if you are in setup() then you could skip the rest of the suite or the test
if you are in tearDown() then you could skip rest of the suite
if you are in suiteTearDown() then you could not skip either

dasilvacontin:
Closed by #946.

ToshioMagic:
I'm still confused about this. I'm doing an API get, getting a 404, and then calling this.skip(). But when i call this.skip() I get an uncaught reference error to "this."

it looks like:

beforeEach(function(done) {
    // blahblahblah
    this.skip();
    done();
});

cmbuckley:
@ToshioMagic This made it into the 2.2.0 release of mocha; what version are you using?

ToshioMagic:
@cmbuckley 2.2.5

cmbuckley:
@ToshioMagic sounds like you're dealing with a separate problem, As the `this.skip()` functionality is available in your version. You may find more help on [Gitter](https://gitter.im/mochajs/mocha).

ryanshawty:
My test structure is defined inside a require module e.g.

````js
define(function(require){
  before(function(){
    // do stuff
  });

  require('spec1');
  require('spec2');
  //etc
});
````
And in the specs there is something similar to

````js
describe('Suite', function(){
  var enabled = false;
  before(function(){
    if(!loaded) return this.skip();
      // ...
  });

  it('should do something', function(){
    // ...
  });
});
````

From what I can tell if the module is not loaded it should skip the rest of suite but not skip the spec2, in my case it is skipping ALL tests after that suite. Is this by design?

**Comment by code-guru** [comment]
@ToshioMagic could you solved the problem? I have the same problem, when I use skip() inside a async callback I got following error from mocha. 
 
```JavaScript
 Error: the object {
  "message": [undefined]
  "uncaught": true
} was thrown, throw an Error :)
      at Runner.fail (\node_modules\mocha\lib\runner.js:225:11)
      at Runner.uncaught (\node_modules\mocha\lib\runner.js:682:8)
      at process.uncaught (\node_modules\mocha\lib\runner.js:727:10)
```

I'm using mocha v2.3.4 . can anyone confirm this problem? Should I open a new issue?

[...]

PR #3859, opened 2019-04-04
Fix hook pattern of this.skip() in it() tests

juergba:
### Description

The current hook pattern is incorrect. When a test is skipped at runtime, all `beforeEach` hooks have already been executed, nevertheless all `afterEach` hooks are ignored completely.

As a general rule: when a `beforeEach` hook runs, the corresponding `afterEach` should run also for correct cleanup. This same pattern has already been implemented with failing hooks, `this.skip()` in `beforeEach`  and also while using the `--bail` option.

The correct hook pattern should be:
- run all `beforeEach` hooks ("A", "B")
- abort test execution and set test to `pending`
- run all corresponding `afterEach` hooks ("Y", "Z")

By contrast: tests skipped at compile-time with `it.skip()` do not run neither `beforeEach` nor `afterEach` hooks.

```js
describe('outer', function() {
    beforeEach(function() {
      console.log("A");              // yes, does run
    });
      beforeEach(function() {
        console.log("B");            // yes, does run
      });
      it('test case', function() {
        this.skip();                 // pending
        console.log("C");            // aborted, does not print "C"
      });
      afterEach(function() {
        console.log("Y");            // no, does not run  => should run
      });
    });
    afterEach(function() {
      console.log("Z");              // no, does not run  => should run
    });
  });
```

### Applicable issues

#3740 partially

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -673,11 +673,6 @@ Runner.prototype.runTests = function(suite, fn) {
             self.fail(test, err);
           }
           self.emit(constants.EVENT_TEST_END, test);
-
-          if (err instanceof Pending) {
-            return next();
-          }
-
           return self.hookUp(HOOK_TYPE_AFTER_EACH, next);
         }
 
--- a/test/integration/fixtures/pending/skip-async-spec.fixture.js
+++ b/test/integration/fixtures/pending/skip-async-spec.fixture.js
@@ -1,12 +1,30 @@
 'use strict';
+var assert = require('assert');
 
 describe('skip in test', function () {
+  var runOrder = [];
+  beforeEach(function () {
+    runOrder.push('beforeEach');
+  });
+
   it('should skip async', function (done) {
     var self = this;
     setTimeout(function () {
       self.skip();   // done() is not required
     }, 0);
   });
+  it('should run other tests in suite', function () {});
 
-  it('should run other tests in the suite', function () {});
+  afterEach(function() {
+    runOrder.push('afterEach');
+  });
+  after(function() {
+    runOrder.push('after');
+    assert.deepStrictEqual(runOrder, [
+      'beforeEach', 'afterEach',
+      'beforeEach', 'afterEach',
+      'after'
+    ]);
+    throw new Error('should throw this error');
+  });
 });
--- a/test/integration/fixtures/pending/skip-sync-spec.fixture.js
+++ b/test/integration/fixtures/pending/skip-sync-spec.fixture.js
@@ -1,10 +1,28 @@
 'use strict';
+var assert = require('assert');
 
 describe('skip in test', function () {
+  var runOrder = [];
+  beforeEach(function () {
+    runOrder.push('beforeEach');
+  });
+
   it('should skip immediately', function () {
     this.skip();
     throw new Error('never run this test');
   });
+  it('should run other tests in suite', function () {});
 
-  it('should run other tests in the suite', function () {});
+  afterEach(function() {
+    runOrder.push('afterEach');
+  });
+  after(function() {
+    runOrder.push('after');
+    assert.deepStrictEqual(runOrder, [
+      'beforeEach', 'afterEach',
+      'beforeEach', 'afterEach',
+      'after'
+    ]);
+    throw new Error('should throw this error');
+  });
 });
--- a/test/integration/pending.spec.js
+++ b/test/integration/pending.spec.js
@@ -59,13 +59,14 @@ describe('pending', function() {
       it('should immediately skip the spec and run all others', function(done) {
         run('pending/skip-sync-spec.fixture.js', args, function(err, res) {
           if (err) {
-            done(err);
-            return;
+            return done(err);
           }
-          assert.strictEqual(res.stats.pending, 1);
-          assert.strictEqual(res.stats.passes, 1);
-          assert.strictEqual(res.stats.failures, 0);
-          assert.strictEqual(res.code, 0);
+          expect(res, 'to have failed with error', 'should throw this error')
+            .and('to have failed test count', 1)
+            .and('to have pending test count', 1)
+            .and('to have pending test order', 'should skip immediately')
+            .and('to have passed test count', 1)
+            .and('to have passed tests', 'should run other tests in suite');
           done();
         });
       });
@@ -192,13 +193,14 @@ describe('pending', function() {
       it('should immediately skip the spec and run all others', function(done) {
         run('pending/skip-async-spec.fixture.js', args, function(err, res) {
           if (err) {
-            done(err);
-            return;
+            return done(err);
           }
-          assert.strictEqual(res.stats.pending, 1);
-          assert.strictEqual(res.stats.passes, 1);
-          assert.strictEqual(res.stats.failures, 0);
-          assert.strictEqual(res.code, 0);
+          expect(res, 'to have failed with error', 'should throw this error')
+            .and('to have failed test count', 1)
+            .and('to have pending test count', 1)
+            .and('to have pending test order', 'should skip async')
+            .and('to have passed test count', 1)
+            .and('to have passed tests', 'should run other tests in suite');
           done();
         });
       });


plroebuck:
https://github.com/mochajs/mocha/issues/332#issuecomment-22217934

lib/runner.js
```diff
@@ -674,9 +674,9 @@ Runner.prototype.runTests = function(suite, fn) {
           }
           self.emit(constants.EVENT_TEST_END, test);
 
-          if (err instanceof Pending) {
-            return next();
-          }
+          // if (err instanceof Pending) {
+          //   return next();
+          // }
```
plroebuck: You were planning on removing these lines, right? :)

juergba:
> #332 (comment)

@plroebuck what are your conclusions about this message?

plroebuck:
> what are your conclusions about this message?

I was adding some context for how the code you were "removing" got this way.
So it appears `afterEach` has been disabled via PR #946 for more than 4 years now...
The PR added tests, but nothing tested `afterEach`.

plroebuck:
> As a general rule: when a beforeEach hook runs, the corresponding afterEach should run also for correct cleanup.

Was curious how you were planning to determine "corresponding" though.
If I have five `beforeEach`, three `afterEach`, and `this.skip()` in the second `beforeEach`, how will you know what to run?

juergba:
In this particular case the definition of "corresponding" is easy. When the test is skipped conditionally at runtime, **all** `beforeEach` hooks have been executed. Therefore **all** `afterEach` hooks should run as well.

There is another PR of mine, `this.skip()` in `beforeEach` hooks. The `suite` level is defining the corresponding hook-groups.
- suite level-1: `beforeEach` and `afterEach` will run
- suite level-2: first `beforeEach` runs /  `this.skip()` in second `beforeEach` / **all** `afterEach` will run
- suite level-3: `beforeEach` and `afterEach` will not run

So all `afterEach` hooks on suite level-2 run, not just the first or the second.

SiegSpace:
Looking forward to this being merged as the lack of after each closures is making a mess of my tests. Might have to fork in the mean time...

mrstux:
This issue just took down our CI. The exception generated broke out of mocha and into the build system, resulting in false positive builds.

Build engineers are looking into handling the exception to prevent the false positives.

Manually patched in the fix using `patch-package` to unblock many teams.

Looking forward to this getting merged into the next release.

juergba:
I haven't received any reviews for this PR for months by any maintainer.

@mrstux @rfd3344 @SiegSpace I need your help to finally get these changes into next major release.
I kindly ask you to review this PR:
- your opinion about the hook pattern described above.
- optional: applied code change
- the ones who patched their Mocha version: feedback, any bugs?

Their is another, similar PR (3741) open, please keep the two topics strictly apart.

mrstux:
In my opinion, and other developers I've spoken to, the understanding is that

`before` will be run before all tests in a suite. `beforeEach` will be run before each test. if `beforeEach` is run, even if the test is skipped, then the `afterEach` *must* be run. And likewise `after` must be run after all tests in a suite, and their hooks are run.

The applied code change seems to work correctly in all cases we checked. We applied the code change to both mocha@latest and mocha@^3.0

ie by commenting out the following lines, and using patch-package to apply.

```
if (err instanceof Pending) {		
             return next();		
           }
```

The changes worked correctly and seemed to be correct in all tests we performed. 

We did run into compatibility issues with mocha@latest on some parts of the CI, which is why we also applied to mocha@3

Now, the only issue we have is problems caused by `patch-package`

The following tests were used to confirm all proper behaviour, as part of testing before deploying. In our case, the critical functionality is that we need to be able to skip from inside a function that is called by the test at the beginning of the test, in this case `TestHelper.setExpectedDurationWithTimeoutAndSkip`. Because `.skip()` triggers an exception, this works correctly.

The issue is that when `afterEach` is not called, and the `beforeEach` has been called, then all tests can cascade fail.

Our test suite can take over 4 hours to run. Thus it is important to be able to skip the more long-running tests programmatically.

```
import TestHelper from '../../helpers/TestHelper'; // implements setExpectedDurationWithTimeoutAndSkip
import { assert } from 'chai';


describe('Mocha', function () {
	let createdObj = 0;

	let beforeEachAsync = function (done) {
		assert.strictEqual(createdObj, 0);
		setTimeout( function () {
			assert.strictEqual(createdObj, 0);
			createdObj++;
			done();
		}, 100);
	};
	let beforeEachSync = function () {
		assert.strictEqual(createdObj, 0);
		createdObj++;
	};
	let afterEachAsync = function (done) {
		assert.strictEqual(createdObj, 1);
		setTimeout( function () {
			assert.strictEqual(createdObj, 1);
			createdObj--;
			done();
		}, 100);
	};
	let afterEachSync = function () {
		assert.strictEqual(createdObj, 1);
		createdObj--;
	};

	let doTest = function () {
		let timeList = [150, 50, 235, 452];   // in milliseconds
		timeList.forEach( function (time) {
			it(`should call afterEach after done is called with async tests, time=${time}`, function (done) {
// uses command-line options to skip tests which are expected to take too long.
				TestHelper.setExpectedDurationWithSkipAndTimeout(this, 1500);
				assert.strictEqual(createdObj, 1);
				setTimeout( function () {
					assert.strictEqual(createdObj, 1);
					done();
				}, time );
			});

			it(`should skip correctly with async tests, time=${time}`, function (done) { // eslint-disable-line no-unused-vars
				assert.strictEqual(createdObj, 1);
				this.skip();
			});

			it(`should call afterEach with sync tests, time=${time}`, function () {
				assert.strictEqual(createdObj, 1);
			});

			it(`should skip correctly with sync tests, time=${time}`, function () {
				assert.strictEqual(createdObj, 1);
				this.skip();
			});
		});
	};

	describe( 'Mocha', function () {
		it( 'should skip a test which calls this.skip()', function () {
			this.skip();
			throw new Error( `this test should've been skipped`);
		});
	});

	describe('async beforeEach, async afterEach', function () {
		beforeEach('init', beforeEachAsync);
		afterEach('dispose', afterEachAsync);
		doTest();
	});

	describe('sync beforeEach, async afterEach', function () {
		beforeEach('init', beforeEachSync);
		afterEach('dispose', afterEachAsync);
		doTest();
	});

	describe('async beforeEach, sync afterEach', function () {
		beforeEach('init', beforeEachAsync);
		afterEach('dispose', afterEachSync);
		doTest();
	});

	describe('sync beforeEach, sync afterEach', function () {
		beforeEach('init', beforeEachSync);
		afterEach('dispose', afterEachSync);
		doTest();
	});
});
```

danielmcmahon02:
This change made testing much more reliable for me.

juergba:
> When an after each hook fails, the test suite is exited, and the next test suite begun.
This is an improvement from the previous behaviour which did not exit and caused cascade failures.

Changes in the "after each" hook behavior - independently of `this.skip()` -  are not intended. I have to check this.

Without this change the "after each" hook is never executed and therefore does not fail. I hope this is what you have meant.

juergba:
@mochajs/core 

I will merge this PR within a few days based on:
- the identical hook pattern is used in similar szenarios like `--bail`
- feedback I received by some of our users

juergba:
@mrstux I have another question and would appreciate your opinion.

Our docu states:

> [...] It's also important to note that calling this.skip() will effectively abort the test.
[...]
Best practice: To avoid confusion, do not execute further instructions in a test or hook after calling this.skip().

Currently `this.skip()` aborts execution of the test/hook by throwing an exception which is swallowed by the `uncaughtException` handler. This pattern is not really a piece of art and makes it difficult to handle complex situation eg. successfully passed tests which fail retrospectively in asynchronous way.

I'm thinking about removing the aborting part. The user would be responsible to make sure there are no more instructions after `this.skip()`.
Could you live with a change like this?

vkarpov15:
@juergba @craigtaub would you be open to a PR backporting this to 5.x? The MongoDB driver team has been running into this issue, but they still need Node 4 support.

If I were to make a PR backporting this change, would I make it against this branch? https://github.com/mochajs/mocha/tree/release/v5.2.x

craigtaub:
@vkarpov15 believe thats the latest v5 branch. Would that not break semver given its a major change?

vkarpov15:
@craigtaub you're right. I had assumed this was going in to 6.2.x since the PR was merged to master, but that was incorrect. The MongoDB driver team will likely make a major release that drops Node 4 and Node 6 support soon enough - I don't think it's worth it to backport this behind a flag to 5.x.

PR #3862, opened 2019-04-05
Fix `--config` loading: cwd-relative, fallback to require()

### LINKED ISSUE #3822: --config does not resolve paths relative to CWD

cspotcode:
### Requirements

### Description of the Change

Fixes #3822.

`--config` is first interpreted as a cwd-relative path.  We check if it exists and is a file. (not directory)
If found, we try to parse it according to its file extension. (this behavior is unchanged)
If not found, we attempt to `require.resolve()` it relative to cwd.
If `require.resolve()` is able to locate a file, we `require()` it.  We do not check file extension nor use a custom loader.

Error messages are logged for a) failing to locate a file in the first place and b) failing to parse or load the file that was found.  For b) the error message logs the resolved path, so users can see that we found e.g. `../../node_modules/monorepo-shared-config/mocha.js`.

---

Worth noting, `cwd` may not *always* be the correct directory for resolving.  For example, if in the future the `--config` value comes from a package.json file -- `"mocha": "@cspotcode/my-mocha-config"` -- then we should resolve relative to the directory containing package.json, not necessarily cwd.

### Applicable issues

#3822 
#3823 
#3829

### DIFF
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -12,6 +12,7 @@ const fs = require('fs');
 const findUp = require('find-up');
 const path = require('path');
 const debug = require('debug')('mocha:cli:config');
+const utils = require('../utils');
 
 /**
  * These are the valid config files, in order of precedence;
@@ -31,6 +32,8 @@ exports.CONFIG_FILES = [
 /**
  * Parsers for various config filetypes.  Each accepts a filepath and
  * returns an object (but could throw)
+ *
+ * Filepath *must* be absolute; external logic resolves relative paths.
  */
 const parsers = (exports.parsers = {
   yaml: filepath =>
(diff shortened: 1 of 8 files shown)


cspotcode:
Added tests.

juergba:
The `--config` loading should be identical for all RC file extensions, so the status-quo doesn't make sense. You are right.
@cspotcode are you still interested in this PR?

We could try to get this one in somehow. Since _lib/utils.js_ is part of the browser bundle, we should not put your helper function into this module.

cspotcode:
IIRC the discussions for #3822, #3823 and #3829 left a bad taste in my mouth, so I'd rather someone else finish this one.

juergba:
@cspotcode ok. Our team has changed meanwhile, but thanks anyway. I remember times when I was sweating blood because of this Windows CRLF publishing story.
Closing.

PR #3872, opened 2019-04-12
rename "exclude" to "ignore" and create alias; closes #3871

### LINKED ISSUE #3871: rename `exclude` to `ignore` and alias `exclude` to `ignore`
as per #2036, "ignore" functionality is available via the `exclude` option of a `.mocharc.*` config file.  Since `ignore` is more of an obvious (and  precedented) option name, let's prefer `ignore` as the canonical name and alias `exclude` to it.

boneskull:
- Renamed all occurrences of `exclude` to `ignore`, as the latter is now the canonical name
- Updated docs to reflect change

### DIFF
--- a/example/config/.mocharc.yml
+++ b/example/config/.mocharc.yml
@@ -6,8 +6,6 @@ check-leaks: false
 color: true
 delay: false
 diff: true
-exclude:
-  - /path/to/some/excluded/file
 exit: false # could be expressed as "no-exit: true"
 extension:
   - js
@@ -25,6 +23,8 @@ global:
 # fgrep and grep are mutually exclusive
 # grep: something
 growl: false
+ignore:
+  - /path/to/some/ignored/file
 inline-diffs: false
 # needs to be used with grep or fgrep
 # invert: false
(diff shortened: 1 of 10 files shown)


boneskull:
@mochajs/core any objections?

plroebuck:
Have no issue with adding `--ignore` as cmdline alias for `--exclude`, but most of the rest just seems like unnecessary churn.

Related but not, I could have _sworn_ many of these "docs/index.md" changes have been made several times over now (e.g., "plain" vs. "text", ["test/"] vs. ["test"]), yet here they are again?

example/config/.mocharc.yml
```diff
@@ -6,8 +6,8 @@ check-leaks: false
 color: true
 delay: false
 diff: true
-exclude:
-  - /path/to/some/excluded/file
+ignore:
```
plroebuck: If renamed, it should move to L28.

(and move `v8-stack-trace-limit` to L45)

markowsiak: +1

lib/cli/run-option-metadata.js
```diff
@@ -14,7 +14,7 @@
  */
 exports.types = {
   array: [
-    'exclude',
+    'ignore',
```
markowsiak: Relocation for alphabetization?

boneskull:
@plroebuck I get the same feeling about those changes... that's why we need to pull  the auto-generated parts of `index.md` out of VCS.  There's an issue for that, and maybe even a pull request, but haven't checked on its status in awhile

PR #3874, opened 2019-04-13
WIP: Reporter Output via Mixins

plroebuck:
### Description of the Change

Created class mixins to allow a reporter to support multiple output destinations.
Hope would be to _eventually_ make reporters use these mixins to standardize
built-in reporters across the board.

This PR transformed `XUnit` first, since it already had unit tests dealing with file output.
Would like both `JSON` and `TAP` to get the "treatment" soon.

Looking for comments on the approach taken.

### Alternate Designs

We _could_ add the same "XUnit" code to each reporter that desired file output.

**Magnets!!!**

### Benefits

Trivializes adding support for file output to another reporter.

### Possible Drawbacks

None found in XUnit thusfar.

### Applicable issues

semver-minor

### DIFF
--- a/lib/reporters/writers/console.js
+++ b/lib/reporters/writers/console.js
@@ -1,0 +1,73 @@
+'use strict';
+
+/**
+ * Save timer references (mochajs/mocha#237).
+ */
+var setTimeout = global.setTimeout;
+
+/**
+ * Provides methods used for output by console-oriented reporters.
+ *
+ * @description
+ * Not meant to be used directly.
+ *
+ * @package
+ * @mixin ConsoleWriter
+ */
+var ConsoleWriter = function() {
+  var buffer = '';
+
+  /**
+   * Writes string to reporter output.
+   *
+   * @package
+   * @param {string} str - String to write to the reporter output.
+   * @returns {this}
+   * @chainable
+   */
+  this.write = function(str) {
+    if (str) {
+      buffer += str; // Store until `console.log` invoked
+    }
+    return this;
+  };
+
+  /**
+   * Writes newline-terminated string to reporter output.
+   *
+   * @package
+   * @param {string} str - String to write to the reporter output.
+   * @returns {this}
+   * @chainable
+   */
+  this.writeln = function(str) {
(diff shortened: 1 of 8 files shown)


lib/reporters/writers/console.js
```diff
@@ -0,0 +1,73 @@
...
+   * Writes newline-terminated string to reporter output.
+   *
+   * @package
+   * @param {string} str - String to write to the reporter output.
+   * @returns {this}
+   * @chainable
+   */
+  this.writeln = function(str) {
+    this.write(str);
+    console.log(buffer);
```
craigtaub: Worth saving a reference like the timer? (let users stub)

lib/reporters/writers/console.js
```diff
@@ -0,0 +1,73 @@
...
+   *
+   * @package
+   * @param {Function} cb - Callback invoked when cleanup done
+   */
+  this._done = function(cb) {
+    if (buffer) {
+      var delay = 1000; // One second
+
+      this.writeln();
+      setTimeout(cb, delay);
```
craigtaub: Why does only console writer need a timeout? Why 1s?

craigtaub:
Few questions. Approach makes sense to me.

lib/reporters/writers/console.js
plroebuck: As this method is invoked immediately prior to `Mocha.run` finishing up, the delay is there to (hopefully) allow completion of flushing any buffered (i.e., unwritten) output to the screen. The '1s' time is arbirtrary for the moment; it's rather long for CPU-related tasks, but as this is IO-related I'm uncertain just _what_ a good value would be.

Be happy to change the comment to "arbitrary", as I usually would have...

lib/reporters/writers/console.js
plroebuck: I hadn't bothered integrating with your PR at this point, but it would be done in this one file for all reporters implementing `ReportWriter`. It's quite honestly trivial to implement for a reporter; it's writing the unit tests that takes all the time...

PR #3879, opened 2019-04-17
chore: add .npmrc, ensure registry=https://registry.npmjs.org/

markowsiak:
### Description of the Change

Adding a `.npmrc` file that explicitly points to `https://registry.npmjs.org/` allows contributors to more easily and confidently execute npm commands on the appropriate registry.

For the reason mentioned above.

### Benefits

Eases the efforts of contributors.

### Possible Drawbacks

None.

### Applicable issues

Patch level update.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+registry=https://registry.npmjs.org/
+


boneskull:
If you don't want to ignore globally, you could also run `git config core.excludesFile /someplace/some-ignore-file` in your working copy, where `some-ignore-file` is:

```plain
.npmrc
```

https://www.git-scm.com/docs/git-config#Documentation/git-config.txt-coreexcludesFile

markowsiak:
@boneskull thanks for the idea.
I went with the excludes file approach and it worked well 👍 

I think we can close this PR, and I'll raise a different one to document this approach in `MAINTAINERS.md`.  Let me know if you think it would be worthwhile 👍

boneskull:
@markowsiak yes updating MAINTAINERS.md sounds good.  thanks!

PR #3881, opened 2019-04-17
Remove noise output from JSON reporter test

plroebuck:
### Description of the Change
Squelches STDOUT from test specification using sinon stub.

### Alternate Designs
NA

NA

### Benefits
Easier to see test results.

### Possible Drawbacks
None

### Applicable issues
semver-patch

### DIFF
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var sinon = require('sinon');
 var Mocha = require('../../');
 var Suite = Mocha.Suite;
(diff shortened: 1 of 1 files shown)


plroebuck:
> Is there a reason that only the json reporter has this noise and other reporters don't require to be muted during our test suite execution?

Outsider and I cleaned up most of the reporters a while back by introducing the `runReporter` method, but test implementation for some reporters couldn't easily use it.

Believe all reporters but `Base` have now been addressed noise-wise. I'll get to that one soon enough.

Issue #3884, opened 2019-04-19
Drop Node.js 6.x support


outsideris:
Node.js v6 will be the end of life on April 30th.

There isn't an issue to drop supporting Node.js v6 for me. Do you know any problem?

juergba:
Then the `debug` options could be deprecated as well.

Bamieh:
No problems for us in dropping support for Node.js v6 that I know of.

mojosoeun:
@juergba what you said means is `--debug` or `--debug-brk` could be deprecated. right?
https://mochajs.org/#-debug-inspect-debug-brk-inspect-brk-debug-inspect

Issue #3819, opened 2019-03-08
meta: LTS strategy


boneskull:
To support "LTS"--or long-running releases--we will need to make some development workflow changes.

In terms of a branching strategy, I'm thinking the most reasonable approach would be [trunk-based development](https://trunkbaseddevelopment.com), which I had experimented with before without knowing what it was called.  Node.js uses something very similar to this.

In a nutshell:

- all development work is done targeting `master` (*except* in the case that the bug cannot be reproduced in `master`; it then must happen on release branches)
- each LTS release (e.g., each major or even-numbered majors) branches from `master`
- bug fixes, features, etc. are *cherry-picked* from `master` into the release branch
- releases are tagged from the release branches, not `master`
- "feature" branches exist how we've been doing them--no change here. these are any short-lived branches to be rebased onto `master` and merged via a PR. note that this includes bug fixes, "chores"--pretty much anything.
- in the case that work *must* be done against a release branch, a PR should target that branch, and *cherry-picked* from release into `master`, *if* it makes sense.
- at no point is a release branch ever merged *back* into `master`.

In addition to whatever we're going to get from users reporting bugs against multiple release lines, this is the overhead we're looking at for LTS releases.  There *are* some tools available which we can either use outright or adapt for our uses (from the Node.js project), but we may need a purpose-built solution.  For example, it'd maybe make sense to automatically _attempt_ cherry-picking of new changesets in `master` into the active release branches.  

(AFAIK, cherry-picking is used because otherwise each changeset in `master` would create an extra merge commit when pulled into an active release branch.)

A noted headache is cross-referencing changesets with their original PRs.  It becomes critical to link to the PR number somewhere in the commit message.  Of course, you can't add the PR to the commit message when you create a changeset, so the commit message must be amended at time-of-merge.  Node.js also has tooling (and a convention) around this.

* * *

Another unconventional way to tackle this is to literally snapshot the codebase for each release line and stuff it in some subdirectory.  I'm pretty sure I didn't just dream this up.  Anyway, I don't think the tooling's there to support this sort of thing, though at its core, it sounds similar to a monorepo.  The major challenge would be to coordinate patch application across multiple directories, assuming something like Lerna could manage the rest of it.  While the above branching solution is non-trivial, I still don't see a clear advantage; both will require tooling.

cc @mochajs/core

boneskull:
would love input or suggestions from anyone in @nodejs/package-maintenance or @nodejs/release

boneskull:
[...]

PR #3885, opened 2019-04-20
Drop node.js v6 support

mojosoeun:
### Description of the Change

According to this issue(#3884), Node.js v6 will be the end of life on April 30th. 
So I removed v6 and changed md file.

### Alternate Designs

N/A

It's minimum Node.js version.

### Benefits

Mocha doesn't care issues with Node.js v6 anymore.

### Possible Drawbacks

N/A

### Applicable issues

close(#3884)


docs/index.md
```diff
@@ -98,7 +98,7 @@ or as a development dependency for your project:
 $ npm install --save-dev mocha
 ```
 
-> As of v6.0.0, Mocha requires Node.js v6.0.0 or newer.
+> As of v6.0.0, Mocha requires Node.js v8.0.0 or newer.
```
Bamieh: This is going to be a breaking change so `As of mocha v7.0.0`

outsideris: Agreed.

mojosoeun: Okay :) I fixed it!

docs/index.md
```diff
@@ -98,7 +98,7 @@ or as a development dependency for your project:
 $ npm install --save-dev mocha
 ```
 
-> As of v6.0.0, Mocha requires Node.js v8.0.0 or newer.
+> As of v6.0.0, Mocha requires Node.js v7.0.0 or newer.
```
juergba: Is this correct? Should be:
"As of v7.0.0, Mocha requires Node.js v8.0.0 or newer."
Next major Mocha release will be v7. And Node will be supported >=v8.

boneskull:
This is a good opportunity to think more about our branching strategy.  If we merge this, then we cannot just cut a release from `master`.

I think we should discuss this in #3819.

juergba:
@mojosoeun could you rebase, please. Probably there will also be conflicts to solve.
I would like to merge this PR, thanks!

juergba:
merged, but `Smoke` still runs with node v6.

PR #3891, opened 2019-04-24
Add Node v12 in test matrix instead v11

outsideris:
Today Node v12 has been released. So, it's a new current version of Node.
Test mocha with Node v12 instead of v11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ stages:
 
 # defaults
 language: node_js
-node_js: '11'
+node_js: '12'
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,15 @@ node_js: '12'
 addons:
   apt:
     packages:
+      # Growl
       - libnotify-bin
+      # Canvas
+      - pkg-config
+      - libcairo2-dev
       - libpango1.0-dev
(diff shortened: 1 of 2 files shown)


outsideris:
AppVeyor has a problem with Node v12. I will dig it.

juergba:
[boneskull/node-environment-flags](https://github.com/boneskull/node-environment-flags/blob/master/flags.json) should also be revised.
The `Node` flags for the release v12 have to be added.

juergba:
@outsideris I just realised it's not necessary to update "boneskull/node-environment-flags".:

> If the current Node.js version is v10.10.0 or newer, the native implementation will be provided instead.

 => native implementation: `process.allowedNodeEnvironmentFlags`

juergba:
Travis also fails with:
```
1) node-flags
       isNodeFlag()
         special cases
           should return true for flags starting with "preserve-symlinks":
     UnexpectedError: 
expected true to be false
```

plroebuck:
Rebase and try this again.

outsideris:
I already rebased this. However, there is a weired error at `npm ci`. 
Also, AppVeyor looks that still doesn't to support node v12. So, I'm trying to test it with Travis Windows Supporting.

boneskull:
I think we should go with #3919, ty.

PR #3905, opened 2019-05-09
Fix regression/bug in "lookupFiles()"

juergba:
### Description of Change

* fix (my) regession: in some cases directories are returned, but only files are expected.

* fix existing bug: in some cases `glob.sync` returns directories, but only files are expected.
Directories can be excluded with options `{nodir: true}`.
BTW: Mocha's flag `--recursive` has no effect in our `glob.sync` implementation, but the same result can be achieved by using the glob pattern `**`, eg. "test/**/*"

* implement: a filename can be entered without extension. In this case all matches with option `extension[]` are returned. Currently only a constant extension `.js` is used.

### Applicable issues

#3834

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -576,25 +576,27 @@ exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
   var stat;
 
   if (!fs.existsSync(filepath)) {
-    // Check all extensions, using first match found (if any)
-    if (
-      !extensions.some(function(ext) {
-        if (fs.existsSync(filepath + '.' + ext)) {
-          filepath += '.' + ext;
-          return true;
-        }
-      })
-    ) {
-      // Handle glob
-      files = glob.sync(filepath, {nodir: true});
-      if (!files.length) {
-        throw createNoFilesMatchPatternError(
-          'Cannot find any files matching pattern ' + exports.dQuote(filepath),
-          filepath
-        );
-      }
-      return files;
+    var pattern;
+    if (glob.hasMagic(filepath)) {
+      // Handle glob as is without extensions
+      pattern = filepath;
+    } else {
+      // glob pattern e.g. 'filepath+(.js|.ts)'
+      var strExtensions = extensions
+        .map(function(v) {
+          return '.' + v;
+        })
+        .join('|');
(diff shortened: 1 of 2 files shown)


plroebuck:
Please just put in the fix as was given rather than reimagining how it _should_ work.
If they want multiple files, then they should provide an appropriate glob for multiple patterns.

lib/utils.js
```diff
@@ -563,33 +563,38 @@ function isHiddenOnUnix(pathname) {
...
- * @param {boolean} recursive - Whether to recurse into subdirectories.
+ * @param {string[]} [extensions=[]] - File extensions to look for.
+ * @param {boolean} [recursive=false] - Whether to recurse into subdirectories.
  * @return {string[]} An array of paths.
  * @throws {Error} if no files match pattern.
  * @throws {TypeError} if `filepath` is directory and `extensions` not provided.
  */
 exports.lookupFiles = function lookupFiles(filepath, extensions, recursive) {
   extensions = extensions || [];
+  recursive = recursive || false;
```
plroebuck: Don't change the input argument values unless they are not provided (i.e., `undefined`).

juergba: Any falsy input (0, "", null, undefined) or no argument provided (=undefined) will be changed to its default value. What should be wrong about changing those falsy values?

juergba:
https://github.com/mochajs/mocha/pull/3834#issuecomment-491603113

> The filepath + '.js' hack was never intended to handle what you're trying to do with it. Its purpose was to allow small packages that only exported a couple functions (typically from "index.js") to use "test.js" to test them.
If a package is exporting functions in more than one language, its specifications should live in the "test" directory -- the simplistic specification scheme above shouldn't apply.

Sorry, I don't understand a word. Currently if I run `node mocha mytests`, then mocha adds the extension ".js" and runs "mytests.js". I do this every day, they same pattern works in node, typescript, java, kotlin ... So which hack, which small packages, what languages...?
Now the only change I have applied, the extension is not constant ".js" anymore, but dynamically defined with `extension[]`.

juergba:
> Please just put in the fix as was given rather than reimagining how it should work.

Is this a joke? Or serious?

boneskull:
But should have tests. Please try to add something

Issue #3912, opened 2019-05-13
Improve file watching


geigerzaehler:
**Is your feature request related to a problem or a nice-to-have?? Please describe.**
When using the `--watch` option Mocha reruns tests when files change. This is quite helpful but the current implementation suffers from a couple of problems. I’d be more than glad to implement the required changes.

Current problems

* New files are not picked up (#2176)
* What files to watch can only be controlled by the `extensions` configuration. In particular that makes the following uses impossible:
  * Only watch files in a certain directory (#2702)
  * Watch files without extensions
  * Exclude certain files and directories from watching (#2554)
* It always uses the slow [`fs.watchFile`](https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_watchfile_filename_options_listener) even when `fs.watch` is available.
* It is impossible to watch files in `node_modules` and `.git`

**Describe the solution you'd like**

* Use [`chokidar`](https://github.com/paulmillr/chokidar) to get reliable and fast file watching.
* Add a `watchFiles` (name subject to bikeshedding) configuration option that accepts a list of file paths, directories, and globs to watch.
* Rerun tests when a file matching `watchFiles` is added 
* Add `watchIgnore` (name subject to bikeshedding) that accept a list of globs to exclude from watching. Defaults to `['node_modules', '.git/**']`
* Keep backwards compatibility if none of the new options are specified

**Describe alternatives you've considered**
Using `nodemon` would address all the problems. Since this creates a new process on file changes this approach is to slow, especially when using Babel or TypeScript for on-the-fly compilation. There also has been talk (#1780) of deprecating the `--watch` options and use separate tools to accomplish reruns. However people are relying on watch (https://github.com/mochajs/mocha/issues/1780#issuecomment-373974766).

juergba:
@geigerzaehler yesterday I tried to get into this "watch" topic with #3851. I  realised there should be two lists - a watch list and a test list - and the current solution has to be improved or deprecated otherwise.

I would be very happy to get your help with these improvements. So please go ahead with your implementation! In this repo sometimes it can be very painful to push big changes upto the final merge. I will do my best to support you.

What are your reasons for choosing "chokidar"?

geigerzaehler:
Thanks for the support @juergba!


> I realised there should be two lists - a watch list and a test list - and the current solution has to be improved or deprecated otherwise.

[...]

PR #3915, opened 2019-05-14
Extend tests for `--watch` options

geigerzaehler:
### Requirements

Test coverage for the `--watch` option needs to be improved. We also need a solid base to be able to test the changes proposed in #3912.

### Description of the Change

This PR adds two test cases to test the `--watch` option. We check that touching a test file reruns the tests and we test that touching file that has a correct extensions reruns the test.

With the support code introduced we also update the one existing test.

The change uses promises instead of callbacks. For example we change `runMochaJSONRaw` to `runMochaJSONRawAsync`. The use of promises improves the control flow and gives better error messages.

### Benefits

* Better test coverage
* It is now easy to write more watcher tests.
* The use of promises provides better control flow guarantees and error messages.

### Possible Drawbacks

The tests introduce `fs-extra` as a dev dependency. We need this for the following reasons:
* Node v6 does not support `fs.copyFile`
* Nodes vanilla `fs` module has no method to remove a directory and its content.

The tests require quite a lot of non-trivial support code.

With the introduction of `runMochaJSONRawAsync` we mix promise based and callback based APIs. While this is not ideal the benefits of using promises (described above) warrant this inconsistency

### Applicable issues

#3912


juergba:
Yes, this is non-trivial stuff and I will need some time to digest this.

I have doubts about the change from callback to promises. I would have expexted the tests to crash, since our integration tests should also include browser and yes IE11 tests.
If you push your branch directly to our repos mochajs/mocha, then additional tests will run and your changes will probably fail. I'm not sure though...

runMochaJSONRaw: it has been required in just one single spec file?

geigerzaehler:
> I would have expexted the tests to crash, since our integration tests should also include browser and yes IE11 tests.

As far as I can see from `karma.conf.js` only the unit tests (`test/unit/*.spec.js`) are run in the browser. It seems to me that running the integration tests in the browser does not make sense since they actually consist of running the the `mocha` executable.

> runMochaJSONRaw: it has been required in just one single spec file?

Yes, only in the watcher spec. (I would actually prefer to move it to the spec file but did not want to make to many changes.)

juergba:
Thanks @geigerzaehler 
I asked @boneskull to help me out and have a look at this.

juergba:
I pushed your branch to this repo and both additional tests have succeeded.
Note to myself: don't worry, be happy.

boneskull:
@juergba 

> I have doubts about the change from callback to promises. I would have expexted the tests to crash, since our integration tests should also include browser and yes IE11 tests.

This is untrue (unfortunately); only the stuff in `test/unit` and `test/browser-specific` run in a browser.   I have an experimental branch where I'm trying to reconcile these and run some portion of the integration tests in a browser, but it's incomplete.

I was also hoping to switch to `Promise`-based test helpers, so this is a welcome change from my view.

.eslintrc.yml
```diff
@@ -25,6 +25,7 @@ overrides:
       - bin/*
       - lib/cli/**/*.js
       - test/node-unit/**/*.js
+      - test/integration/**/*.js
```
boneskull: would prefer if we can list individual files here; would rather not open the floodgates to ES6 in the integration tests, as there are tentative plans to run some portion of them in a browser

test/integration/helpers.js
```diff
@@ -282,7 +287,7 @@ function resolveFixturePath(fixture) {
   if (path.extname(fixture) !== '.js') {
     fixture += '.fixture.js';
   }
-  return path.join('test', 'integration', 'fixtures', fixture);
+  return path.resolve('test', 'integration', 'fixtures', fixture);
```
boneskull: does this actually change anything?

test/integration/options/watch.spec.js
```diff
@@ -19,38 +21,156 @@ describe('--watch', function() {
...
+      const [mocha, outputPromise] = runMochaJSONWatchAsync(this.testFile, [], {
+        cwd: this.tempDir
+      });
 
-          var sigint = 2;
-          expect(data.code, 'to be', exitStatusBySignal(sigint));
-          done();
-        },
-        spawnOpts
+      return sleep(1000)
```
boneskull: this is more idiomatic when using `unexpected`:

```js
return expect(sleep(1000).then(...), 'to have length', 2);
```

please change here & other places

test/integration/options/watch.spec.js
```diff
@@ -19,38 +21,156 @@ describe('--watch', function() {
...
+    return testResults;
+  });
+  return [mocha, testResults];
+}
+
+/**
+ * Touch a file by appending a space to the end. Returns a promise that resolves
+ * when the file has been touched.
+ */
+function touchFile(file) {
```
boneskull: this function doesn't return anything.

test/integration/options/watch.spec.js
```diff
@@ -19,38 +21,156 @@ describe('--watch', function() {
...
+ * by SIGINT.
+ */
+function runMochaJSONWatchAsync(fixture, args, spawnOpts) {
+  args = ['--watch'].concat(args);
+  const [mocha, mochaDone] = runMochaJSONRawAsync(fixture, args, spawnOpts);
+  const testResults = mochaDone.then(data => {
+    expect(data.code, 'to be', sigintExitCode);
+
+    const testResults = data.output
+      // eslint-disable-next-line no-control-regex
```
boneskull: what's the necessity of this?

test/integration/options/watch.spec.js
```diff
@@ -19,38 +21,156 @@ describe('--watch', function() {
...
+ * are an array of JSON objects generated by the JSON reporter.
+ *
+ * Checks that the exit code of the mocha command is 130, i.e. mocha was killed
+ * by SIGINT.
+ */
+function runMochaJSONWatchAsync(fixture, args, spawnOpts) {
+  args = ['--watch'].concat(args);
+  const [mocha, mochaDone] = runMochaJSONRawAsync(fixture, args, spawnOpts);
+  const testResults = mochaDone.then(data => {
+    expect(data.code, 'to be', sigintExitCode);
```
boneskull: I think we can skip this.  these tests aren't asserting Mocha exits how we expect; they are `--watch` tests.

test/integration/helpers.js
```diff
@@ -115,31 +115,36 @@ module.exports = {
...
     args = args || [];
 
-    return invokeSubMocha(
-      args.concat(['--reporter', 'json', path]),
-      function(err, resRaw) {
-        if (err) return fn(err);
+    let childProcess;
+    const resultPromise = new Promise((resolve, reject) => {
+      childProcess = invokeSubMocha(
+        args.concat(['--reporter', 'json', path]),
```
boneskull: we can use rest/spread syntax here: `[...args, '--reporter', 'json', path]`

package.json
```diff
@@ -550,6 +550,7 @@
     "eslint-plugin-prettier": "^3.0.1",
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
+    "fs-extra": "^8.0.1",
```
boneskull: if `fs-extra` doesn't pollute `fs`, this is fine.  in a perfect world, I would prefer individual modules to handle whatever operations we need, but not willing to die on that hill

geigerzaehler: `fs-extra` does not pollute `fs`.

test/integration/options/watch.spec.js
geigerzaehler: The [`no-control-regex` rule](https://eslint.org/docs/rules/no-control-regex) disallows control characters in regular expressions. In this case here we actually want to split the string when a control character (`\u001b`) occurs. We disable the rule to be able to use this regular expression.

.eslintrc.yml
geigerzaehler: https://github.com/mochajs/mocha/pull/3915/commits/563ac00efdfa39da30ec686fcf7a225caf4d4a1e

test/integration/helpers.js
geigerzaehler: This was necessary before but I changed the signature of `runMochaJSONRawAsync` to not use a special `fixturePath` argument. This allows us to revert this change and make the code more understandable.

https://github.com/mochajs/mocha/pull/3915/commits/3f6372e1b138756edaf89cea2a578ec8cb6a1a5a

test/integration/options/watch.spec.js
geigerzaehler: https://github.com/mochajs/mocha/pull/3915/commits/03049f6906d905c4cce2da2665c50e2299bfa840

test/integration/options/watch.spec.js
geigerzaehler: Updated the documentation in https://github.com/mochajs/mocha/pull/3915/commits/3948056608857728f63732b7f6a11286a5cf0447

test/integration/options/watch.spec.js
geigerzaehler: https://github.com/mochajs/mocha/pull/3915/commits/b9f42a872426cecb1bc72158f49bff09c2c95442

test/integration/helpers.js
geigerzaehler: https://github.com/mochajs/mocha/pull/3915/commits/dc7615ae078baa07fc3dec08e84983bab990ad21

geigerzaehler:
Thanks for the review @boneskull. I hope I have addressed all your comments

test/integration/options/watch.spec.js
```diff
@@ -19,38 +21,155 @@ describe('--watch', function() {
       }
```
juergba: Only the first test fails on Windows, the three remaining ones run successfully. So no need to skip these tests.

juergba:
@boneskull why do we test with "cross-spawn", when in real life Mocha runs with node's "child_process"?

boneskull:
I don’t think there’s a great reason. There are issues with child_process spawn on windows, but don’t know if those affect mocha in any meaningful way

juergba:
@craigtaub could you review this PR, please? I think you have been the author of the first watch test.

test/integration/options/watch.spec.js
geigerzaehler: Fixed in https://github.com/mochajs/mocha/pull/3915/commits/97de23f5f239e08101de766bf9c1fc9ee88fbe2a

test/integration/helpers.js
```diff
@@ -114,32 +114,37 @@ module.exports = {
...
-   * returning the **raw** string output, as well as exit code.
+   * Invokes the mocha binary with the given arguments fixture using
+   * the JSON reporter. Returns the child process and a promise for the
+   * results of running the command. The result includes the **raw**
+   * string output, as well as exit code.
    *
    * By default, `STDERR` is ignored. Pass `{stdio: 'pipe'}` as `opts` if you
-   * want it.
-   * @param {string} fixturePath - Path from __dirname__
+   * want it as part of the ocwresultutput.
```
craigtaub: `ocwresultutput`?

geigerzaehler: Looks like I was briefly possessed by a daemon. Fixed, squashed, and pushed

.eslintrc.yml
```diff
@@ -25,6 +25,8 @@ overrides:
       - bin/*
       - lib/cli/**/*.js
       - test/node-unit/**/*.js
+      - test/integration/options/watch.spec.js
```
boneskull: :+1:

boneskull:
@geigerzaehler thanks a ton!

geigerzaehler:
> @geigerzaehler thanks a ton!

Thanks for the support

PR #3919, opened 2019-05-17
Update CI config files to use Node-12.x

plroebuck:
### Description of the Change

Update CI configuration files to use current version of Node.
* AppVeyor's images still do not have pre-built Node-12 so changed script to use alternative install method.
* Changed Travis browser test to use Node-10 so we can use pre-built "canvas" package.

### Alternate Designs

NA

### Benefits

Code tested against current Node release.

### Possible Drawbacks

Code **not** being tested against current Node release.

### Applicable issues

Use of Node-10 for browser test can be revisited once we determine AG-B dependency updates for Node-12. See further explanation [here](https://github.com/mochajs/mocha/pull/3940#issuecomment-499559492), but as noted, the Node version is essentially meaningless for this test anyway.

semver-patch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,6 @@ addons:
     packages:
       # Growl
       - libnotify-bin
-      # Canvas
-      - libpango1.0-dev
-      - libgif-dev
 # `nvm install` happens before the cache is restored, which means
 # we must install our own npm elsewhere (`~/npm`)
 before_install: |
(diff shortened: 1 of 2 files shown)


outsideris:
So, do you want to close #3891 ?

plroebuck:
Assuming they keep previous schedule, new AppVeyor images should show up mid-June -- Node-12 dropped _just_ after their previous update.

plroebuck:
We _could_ still retain Node-12 for Mocha's Travis browser tests if @Munter could provide what AG-B dependencies versions need be changed in our "package-lock.json" file. I tried various patch updates, but each just lead to another failure deeper in.

PR #3929, opened 2019-05-22
More, improved integration tests for watching

geigerzaehler:
### Requirements

We want to further extend the test coverage of the watch feature and prepare the code so that we can easily write tests for the features described in #3912.

### Description of the Change

To extend the test coverage for mocha in watch mode we add the following two tests:
* Check that test files are reloaded
* Check that watched dependencies are reloaded

To support this change we consolidate `runMochaJSONWatchAsync`, `runMochaJSONRawAsync`, and repeated code in tests into `runMochaWatch`. We introduce `invokeMochaAsync` in `test/integration/helpers` as an async alternative to `invokeMocha`.

We also eliminate the test for the cursor control character in the output. Its usefulness is dubious as it relies on an implementation detail and the other tests cover the intended behavior.

We are also more explicit which test fixtures are used. Instead of setting `this.testFile` in a `beforeEach` hook we do this explicitly for the tests that require it. This prevents interference in tests that do not use the file.

### Alternate Designs

N.A.

### Benefits

* Better test coverage of the watch feature
* Tests for watch features are easily extendible

### DIFF
--- a/test/integration/options/watch.spec.js
+++ b/test/integration/options/watch.spec.js
@@ -4,9 +4,6 @@ const fs = require('fs-extra');
 const os = require('os');
 const path = require('path');
 const helpers = require('../helpers');
-const runMochaJSONRawAsync = helpers.runMochaJSONRawAsync;
-
-const sigintExitCode = 130;
 
 describe('--watch', function() {
   describe('when enabled', function() {
@@ -15,11 +12,6 @@ describe('--watch', function() {
 
     beforeEach(function() {
       this.tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mocha-'));
-
-      const fixtureSource = helpers.DEFAULT_FIXTURE;
-
-      this.testFile = path.join(this.tempDir, 'test.js');
-      fs.copySync(fixtureSource, this.testFile);
     });
 
     afterEach(function() {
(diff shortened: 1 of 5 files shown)


geigerzaehler:
I need to investigate the test failures. Some tests for node 11.13 on windows and 11.15 on linux fail but not for other versions. I’m also not able to reproduce this locally. 🔍

boneskull:
you might want to just pull Node.js 11 out of the matrix, as it's no longer supported.  @mochajs/core thoughts?

juergba:
> you might want to just pull Node.js 11 out of the matrix, as it's no longer supported.

It does not run neither on Node v12.2.0 on my Windows laptop.

boneskull:
> EDIT: just running watch.spec.js on its own does work , but npm test fails.

this is worrisome and likely means that some previous test is poisoning something.  unless it's obvious, the only way I can suggest to track it down is start with running `watch.spec.js` and then keep adding files to the list until it breaks.

boneskull:
also for local reproduction, you may want to cross-check that the tests are running in the same order on CI as they are locally.  if not, you can try to use `--sort` or, worst-case, `--file <file1> --file <file2>` to reproduce the order.

juergba:
If I reduce `package-script.js` to just on single spec file (`watch.spec.js`), it fails.
If I run the same spec file with `node mocha watch.spec.js` or `node _mocha watch.spec.js`, it finishes successfully.

juergba:
`invokeMochaAsync()`: when swapping the binary from `MOCHA_EXECUTABLE` to `_MOCHA_EXECUTABLE`, then it works, locally on Windows, Node v12.2.0.

geigerzaehler:
> `invokeMochaAsync()`: when swapping the binary from `MOCHA_EXECUTABLE` to `_MOCHA_EXECUTABLE`, then it works, locally on Windows, Node v12.2.0.

Thanks a lot for the investigation, @juergba! Unfortunately I’m not able to reproduce this issue locally. So the course of action would be to rename `invokeMochaAsync` to invokeSubMochaAsync` and use `_MOCHA_EXECUTABLE`. Before I do that I would like to investigate this some more to get an understanding of what actually went wrong and why it failed.

juergba:
@geigerzaehler The master has changed in an important way.
Only when `node` flags are passed via Mocha, a child-process is spawned.
The tests of this PR should pass now with both binaries, since no `node` flags are found and therefore no child-processes are spawned.

geigerzaehler:
Indeed, the tests now pass 🕺

test/integration/options/watch.spec.js
```diff
@@ -111,50 +63,98 @@ describe('--watch', function() {
...
-        'when fulfilled',
-        'to have length',
-        1
-      );
+    it('reloads test files when they change', function() {
+      const testFile = path.join(this.tempDir, 'test.js');
+      copyFixture('options/watch/test-file-change', testFile);
+
+      return runMochaWatch(
+        [testFile, '--watch-files', '**/*.js'],
```
juergba: We don't have a `--watch-files` flag, yet. Why does this test pass then?

geigerzaehler: Oops. This is an artifact of extracting it from my wip branch. It does not fail the tests because we use the `--extensions` option to determine which files to watch and the default value is `js`. We also don’t use yarg’s [`strict`option](https://github.com/yargs/yargs/blob/master/docs/api.md#strictenabledtrue), something that I would encourage. This makes yargs silently accept unknown flags.

I’ve removed the artifact.

test/integration/options/watch.spec.js
```diff
@@ -111,50 +63,90 @@ describe('--watch', function() {
...
+      copyFixture('options/watch/test-file-change', testFile);
+
+      return runMochaWatch([testFile], this.tempDir, () => {
+        replaceFileContents(
+          testFile,
+          'testShouldFail = true',
+          'testShouldFail = false'
+        );
+      }).then(results => {
+        expect(results, 'to have length', 2);
```
boneskull: you can do this in a single assertion using `and()` and `to satisfy`:

```js
expect(results, 'to have length', 2)
  .and('to satisfy', [
    {
      passes: expect.it('to be empty'),
      failures: expect.it('to have length', 1)
    },
    {
      passes: expect.it('to have length', 1),
      failures: expect.it('to be empty')
    }
  ]);
```

also, this might work (starting on L82):

```js
return expect(runMochaWatch(/*...*/), 'when fulfilled', 'to have length', 2)
  .and('to satisfy', [
    {
      passes: expect.it('to be empty'),
      failures: expect.it('to have length', 1)
    },
    {
      passes: expect.it('to have length', 1),
      failures: expect.it('to be empty')
    }
  ]);
```

test/integration/options/watch.spec.js
```diff
@@ -111,50 +63,90 @@ describe('--watch', function() {
...
+      const dependency = path.join(this.tempDir, 'lib', 'dependency.js');
+      copyFixture('options/watch/dependency', dependency);
+
+      return runMochaWatch([testFile], this.tempDir, () => {
+        replaceFileContents(
+          dependency,
+          'module.exports.testShouldFail = false',
+          'module.exports.testShouldFail = true'
+        );
+      }).then(results => {
```
boneskull: likewise here

test/integration/options/watch.spec.js
```diff
@@ -111,50 +63,90 @@ describe('--watch', function() {
...
+    [...args, '--watch', '--reporter', 'json'],
+    {cwd}
+  );
+
+  return sleep(1000)
+    .then(() => change())
+    .then(() => sleep(1000))
+    .then(() => {
+      mochaProcess.kill('SIGINT');
+      return resultPromise.then(data => {
```
boneskull: this promise needn't be nested. just return `resultPromise` and continue the chain from L139

test/integration/options/watch.spec.js
geigerzaehler: If it is ok with you, @boneskull, I’d like to stick with the current implementation. I find it easier to read and understand. It does not rely on a DSL so much.

PR #3930, opened 2019-05-22
Extract `watchRun` into separate module

geigerzaehler:
### Requirements

As per @juergba’s suggestion in #3912 we extract the `watchRun` code into its separate module to prepare for #3912.

### Description of the Change

We move `watchRun` from `lib/cli/run-helpers` to `lib/cli/watch-run`. We also don’t use default values for the options anymore. The defaults are handled by `runMocha` and we want to avoid the default values getting out of sync. No other code has changed.

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -12,11 +12,12 @@ const path = require('path');
 const ansi = require('ansi-colors');
 const debug = require('debug')('mocha:cli:run:helpers');
 const minimatch = require('minimatch');
-const Context = require('../context');
-const Mocha = require('../mocha');
 const utils = require('../utils');
+const watchRun = require('./watch-run');
 
 const cwd = (exports.cwd = process.cwd());
+
+exports.watchRun = watchRun;
 
 /**
  * Exits Mocha when tests + code under test has finished execution (default)
@@ -63,32 +64,6 @@ const exitMocha = code => {
   });
 
   done();
-};
-
-/**
- * Hide the cursor.
- * @ignore
- * @private
- */
-const hideCursor = () => {
-  process.stdout.write('\u001b[?25l');
-};
-
-/**
- * Show the cursor.
- * @ignore
- * @private
- */
-const showCursor = () => {
-  process.stdout.write('\u001b[?25h');
-};
-
-/**
- * Stop cursor business
- * @private
- */
-const stop = () => {
-  process.stdout.write('\u001b[2K');
 };
 
 /**
@@ -208,76 +183,6 @@ exports.singleRun = (mocha, {files = [], exit = false} = {}) => {
 };
 
 /**
- * Run Mocha in "watch" mode
- * @param {Mocha} mocha - Mocha instance
- * @param {Object} [opts] - Options
- * @param {string[]} [opts.extension] - List of extensions to watch
- * @param {string|RegExp} [opts.grep] - Grep for test titles
- * @param {string} [opts.ui=bdd] - User interface
- * @param {string[]} [files] - Array of test files
- * @private
- */
-exports.watchRun = (
-  mocha,
-  {extension = [], grep = '', ui = 'bdd', files = []} = {}
-) => {
-  let runner;
-
-  console.log();
-  hideCursor();
-  process.on('SIGINT', () => {
-    showCursor();
-    console.log('\n');
-    process.exit(130);
-  });
-
-  const watchFiles = utils.files(cwd, extension);
-  let runAgain = false;
-
-  const loadAndRun = () => {
-    try {
-      mocha.files = files;
-      runAgain = false;
-      runner = mocha.run(() => {
-        runner = null;
-        if (runAgain) {
-          rerun();
-        }
-      });
-    } catch (e) {
-      console.log(e.stack);
-    }
-  };
-
-  const purge = () => {
-    watchFiles.forEach(Mocha.unloadFile);
-  };
-
-  loadAndRun();
-
-  const rerun = () => {
-    purge();
-    stop();
-    if (!grep) {
-      mocha.grep(null);
-    }
-    mocha.suite = mocha.suite.clone();
-    mocha.suite.ctx = new Context();
-    mocha.ui(ui);
-    loadAndRun();
-  };
-
-  utils.watch(watchFiles, () => {
-    runAgain = true;
-    if (runner) {
-      runner.abort();
-    } else {
-      rerun();
-    }
-  });
-};
-
-/**
  * Actually run tests
  * @param {Mocha} mocha - Mocha instance
  * @param {Object} [opts] - Options
@@ -295,7 +200,7 @@ exports.runMocha = (
   files = []
 ) => {
   if (watch) {
-    exports.watchRun(mocha, {extension, grep, ui, files});
+    watchRun(mocha, {extension, grep, ui, files});
   } else {
     exports.singleRun(mocha, {files, exit});
   }
--- a/lib/cli/watch-run.js
+++ b/lib/cli/watch-run.js
@@ -1,0 +1,107 @@
+'use strict';
+
+const utils = require('../utils');
+const Context = require('../context');
+const Mocha = require('../mocha');
+
+/**
+ * Exports the `watchRun` function that runs mocha in "watch" mode.
+ * @see module:lib/cli/run-helpers
+ * @module
+ * @private
+ */
+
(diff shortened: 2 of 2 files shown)


lib/cli/watch-run.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
```
juergba: Please add a short description and `@private`, `@module` (see CLI modules)

lib/cli/watch-run.js
```diff
@@ -0,0 +1,99 @@
...
+const utils = require('../utils');
+const Context = require('../context');
+const Mocha = require('../mocha');
+
+/**
+ * Run Mocha in "watch" mode
+ * @param {Mocha} mocha - Mocha instance
+ * @param {Object} opts - Options
+ * @param {string[]} opts.extension] - List of extensions to watch
```
juergba: opts.extension] => [opts.extension]

lib/cli/watch-run.js
```diff
@@ -0,0 +1,99 @@
...
+ * @param {string|RegExp} [opts.grep] - Grep for test titles
+ * @param {string} [opts.ui] - User interface
+ * @param {string[]} [opts.files] - Array of test files
+ * @param {string[]} [opts.cwd] - Current working directory
```
juergba: it's not an array?

geigerzaehler:
I addressed all your comments, @juergba.

lib/cli/watch-run.js
```diff
@@ -0,0 +1,106 @@
...
+ * @see module:lib/cli/run-helpers
+ * @module
+ * @private
+ */
+
+/**
+ * Run Mocha in "watch" mode
+ * @param {Mocha} mocha - Mocha instance
+ * @param {Object} opts - Options
+ * @param {string[]} [opts.extension] - List of extensions to watch
```
plroebuck: Why not plural if it's a list?

juergba: Because it's the canonical name of the `extension` option.
We already have a total mess in the Mocha constructor with canonical names, aliases, camelcased version ...

lib/cli/watch-run.js
```diff
@@ -0,0 +1,106 @@
...
+ * Exports the `watchRun` function that runs mocha in "watch" mode.
+ * @see module:lib/cli/run-helpers
+ * @module
+ * @private
+ */
+
+/**
+ * Run Mocha in "watch" mode
+ * @param {Mocha} mocha - Mocha instance
+ * @param {Object} opts - Options
```
plroebuck: Why does this no longer provide a default value ({}) as it did before?

plroebuck: As well as defaults for the object properties?

juergba: see "Description of the change"

lib/cli/watch-run.js
plroebuck: Well, the function argument's name isn't the same thing as the Mocha option (though they may be one and the same in usage). Just because it's frak'd up elsewhere is no reason to propogate further into new code.

lib/cli/watch-run.js
plroebuck: Then all object properties _must_ exist, right? So why are they marked as optional?

lib/cli/watch-run.js
```diff
@@ -0,0 +1,106 @@
...
+  const watchFiles = utils.files(cwd, extension);
+  let runAgain = false;
+
+  const loadAndRun = () => {
+    try {
+      mocha.files = files;
+      runAgain = false;
+      runner = mocha.run(() => {
+        runner = null;
+        if (runAgain) {
```
plroebuck: Would think there should be a `clearScreen` here... see #2312

lib/cli/watch-run.js
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  const purge = () => {
+    watchFiles.forEach(Mocha.unloadFile);
+  };
+
+  loadAndRun();
+
+  const rerun = () => {
+    purge();
```
plroebuck: Would think there should be a `clearScreen` here... see #2312

lib/cli/run-helpers.js
```diff
@@ -295,7 +198,7 @@ exports.runMocha = (
   files = []
 ) => {
   if (watch) {
-    exports.watchRun(mocha, {extension, grep, ui, files});
+    watchRun(mocha, {extension, grep, ui, files, cwd});
```
plroebuck: Shouldn't this still be exported here?

lib/cli/watch-run.js
```diff
@@ -0,0 +1,106 @@
...
+ */
+module.exports = (mocha, {extension, grep, ui, files, cwd}) => {
+  let runner;
+
+  console.log();
+  hideCursor();
+  process.on('SIGINT', () => {
+    showCursor();
+    console.log('\n');
+    process.exit(130);
```
plroebuck: Think this should be rewritten as `128+2`?

lib/cli/watch-run.js
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Run Mocha in "watch" mode
+ * @param {Mocha} mocha - Mocha instance
+ * @param {Object} opts - Options
+ * @param {string[]} [opts.extension] - List of extensions to watch
+ * @param {string|RegExp} [opts.grep] - Grep for test titles
+ * @param {string} [opts.ui] - User interface
+ * @param {string[]} [opts.files] - Array of test files
+ * @param {string} [opts.cwd] - Current working directory
```
plroebuck: This new object property `opts.cwd` was previously just a referenced variable that defaulted to `process.cwd()`; at last count, it isn't a Mocha option either. Why bother passing this? Maybe L36 should be:
```js
  const watchFiles = utils.files(process.cwd(), extension);
```

juergba:
@geigerzaehler feel free to consider anything of latest comments, otherwise your commits stand so far.

IMO: it's a 1:1 extraction of `watchRun` as preparation for further PRs, so it's far to early for nitting games.

lib/cli/watch-run.js
geigerzaehler: I’d like to avoid changing behavior in this PR. It is also not clear to me which behavior we exactly want.

lib/cli/run-helpers.js
geigerzaehler: `lib/cli/run-helpers` is marked as a private module and since the export was never used I removed it. If you feel that we should keep it for backwards compatibility reasons then I’d be ok adding it back.

lib/cli/watch-run.js
geigerzaehler: Addressed in https://github.com/mochajs/mocha/pull/3930/commits/595b275411d944656ffd3f7874c22bcd77483d1c

lib/cli/watch-run.js
geigerzaehler: Addressed in https://github.com/mochajs/mocha/pull/3930/commits/595b275411d944656ffd3f7874c22bcd77483d1c

lib/cli/watch-run.js
geigerzaehler: Good point. Addressed in https://github.com/mochajs/mocha/pull/3930/commits/6f69d01b7ca008f3992b7d1f5537240a4f68b305

lib/cli/run-helpers.js
plroebuck: Prefer to keep the `export`.

geigerzaehler:
I addressed all the comments, rebased, and squashed.

lib/cli/run-helpers.js
geigerzaehler: Done

lib/cli/run-helpers.js
```diff
@@ -295,7 +200,7 @@ exports.runMocha = (
   files = []
 ) => {
   if (watch) {
-    exports.watchRun(mocha, {extension, grep, ui, files});
+    watchRun(mocha, {extension, grep, ui, files});
```
plroebuck: Why not just replace the `exports` here, eliminating "diff" lines?

plroebuck:
@juergba, again with the merge with review comments still ongoing?

juergba:
@plroebuck maybe it's time to talk about reasonable comments, coding style questions, nit picking blocking, editing other people comments and OCD. We can do that in public, or by private communication.

PR #3931, opened 2019-05-23
Bug/3840 --forbid-only doesn't recognize `it.only` when `before` crashes

### LINKED ISSUE #3840: --forbid-only doesn't recognize `it.only` when `before` crashes
Not sure if it's a feature or a bug, but `--forbid-only` crashes only in case of approaching `describe.only` but not `it.only`.

It caught me by a surprise.

Is it supposed to work that way? I assume it should crash same way it crashes for `describe.only`

rpgeeganage:
### Requirements

### Description of the Change

`--forbid-only` doesn't recognize `it.only` when `before` crashes.

This is an attempt to solve the above-mentioned bug.

### Benefits
This is an attempt to solve the above-mentioned bug.

### Possible Drawbacks
None

### Applicable issues
Fixes #3840 

* Is it a bug fix, or does it not impact production code (patch release)? **Yes**

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -621,14 +621,14 @@ Runner.prototype.runTests = function(suite, fn) {
     }
 
     if (test.isPending()) {
-      if (self.forbidPending) {
+      if (self.forbidPending || (self.forbidOnly && suite.hasOnly())) {
+        var error = self.forbidPending
+          ? new Error('Pending test forbidden')
+          : new Error('`.only` forbidden');
+
         test.isPending = alwaysFalse;
-        self.fail(test, new Error('Pending test forbidden'));
-        delete test.isPending;
-      } else if (self.forbidOnly && suite.hasOnly()) {
-        var error = new Error('`.only` forbidden');
         self.fail(test, error);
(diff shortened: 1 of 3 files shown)


juergba:
@rpgeeganage thanks for this PR. 

Some thoughts:
- You are supposing that `.only` has precedence over `skip()`. Have you found any evidence for this in the docu or code base? BTW I don't know the answer.
- Your solution works for `describe.skip()`and `before(... this.skip())`, but fails for `beforeEach(... this.skip())`
- please rename your test and put it in the directory "test\integration\fixtures\options\forbid-only". It's difficult not to loose overview in a sea of tests named with regression numbers.

rpgeeganage:
@boneskull , @juergba ,
sorry for the late reply. I may have screwed up something. I'll look into it.

rpgeeganage:
@boneskull , @juergba 
I have opened a new PR (https://github.com/mochajs/mocha/pull/3948). 
Please check. I have moved the tests and changed the fix to work with `before` and `beforeEach`.
Sorry for unable to open this PR.

PR #3934, opened 2019-05-28
Fix for duplicate entries in mocha.options.globals

pascalpp:
Fixes https://github.com/mochajs/mocha/issues/3933

### Requirements

### Description of the Change

Adds a deduping reducer to Mocha.prototype.globals to ensure that no duplicate items are added.

### Alternate Designs

Decided against using Array.prototype.includes, opted to use IndexOf instead, to ensure the solution has a broad base of support since mocha runs in a lot of places. The method is already using Array.prototype.filter, so I think it's safe to use Array.prototype.reduce, which I believe has equivalent support.

Slight change to a core method of the Mocha constructor API. Not a candidate for its own package.

### Benefits

Mostly just house-cleaning.

### Possible Drawbacks

Possible speed impact, but probably negligible.

### Applicable issues

https://github.com/mochajs/mocha/issues/3933
Originated from a discussion with @juergba here: https://github.com/mochajs/mocha/pull/3914

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -553,7 +553,13 @@ Mocha.prototype._growl = growl.notify;
 Mocha.prototype.globals = function(globals) {
   this.options.globals = (this.options.globals || [])
     .concat(globals)
-    .filter(Boolean);
+    .filter(Boolean)
+    .reduce(function dedupe(memo, item) {
+      if (memo.indexOf(item) < 0) {
+        memo.push(item);
+      }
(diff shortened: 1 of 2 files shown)


pascalpp:
@juergba Here's that PR I promised you to fix the duplicate globals issue. Let me know if you think this is the right approach.

pascalpp:
Will attempt to resolve this in https://github.com/mochajs/mocha/pull/3914

PR #3937, opened 2019-05-31
Fix `utils.clean` to recognise async functions

Alhadis:
### Description of the Change
Currently, the `clean` helper in [`utils/clean.js`](https://github.com/mochajs/mocha/blob/29b7615148b383f4b177ecd21c917bad6f621561/lib/utils.js#L150) isn't stripping the function header from async functions, or generators without whitespace (`function*()`):

Figure 1

This PR fixes that:

Figure 2


outsideris:
What is the report in your screenshots?

Alhadis:
@outsideris Are you referring to the reporter? It's one I wrote myself, intentionally designed to match the look-and-feel of Mocha's `spec` reporter, sans being in HTML.

juergba:
What a terrible regex!
I haven't really tried to understand this thing, but is `'$1$2$3'` correct?
I recognize only two parts `/^ ... }$ | ^( ... )$/`.

Do generator functions work with Mocha?

Alhadis:
> I haven't really tried to understand this thing, but is `'$1$2$3'` correct?

Does this help?

~~~regexp
(?:
    (?# Named functions/function expressions)
    ^ (?:async\s+)?
    function
    (?:
        \*?                    (?# Match `function*foo`)
        (?: \s*
         |  \s+ [^(]*)         (?# Match `function ()`)
         |  \*  [^(]*          (?# Match `function * foo`)
         )
        
        \( [^)]* \)            (?# $1: Argument list)
        
        \s* \{
        ((?:.|\n)*?)           (?# $2: Function body)
        \s* \}
        $
    
    |
    
    (?# Arrow functions)
    ^ (?:async\s+)?
    
    \( [^)]* \)                (?# Argument list)
    \s* => \s*                 (?# Separator)
    
    (?: \{ ((?:.|\n)*?) \s* \} (?#3: Body, bracketed)
     |     ((?:.|\n)*)         (?#4: Body, unbracketed)
     )
     $
)
~~~

The capturing groups which don't match are simply empty strings, due to the quantifiers they use. Though the replacement string should actually be `$1$2$3$4` now that I notice...

> Do generator functions work with Mocha?

Yes.

juergba:
Yes, this helps, thanks.

I guess there is no way to write this regex in a more readable, multiline way?

I have two doubts:
- arguments list: no "()" are allowed within? or it runs up to the last ")"?
- "\n": I'm quite sure this will not work with Windows.

Alhadis:
> I guess there is no way to write this regex in a more readable, multiline way?

Without adding a runtime dependency on something to generate pretty-printed RegExp literals? No.

> arguments list: no "()" are allowed within? or it runs up to the last ")"?

Arrow function arguments can't include brackets, so the pattern is safe.

> "\n": I'm quite sure this will not work with Windows.

Newlines are already normalised by the function before stripping:

https://github.com/mochajs/mocha/blob/36294b082c74af381960a27f5f44527e2d08b3ea/lib/utils.js#L150-L152

juergba:
> Arrow function arguments can't include brackets, so the pattern is safe.

What about non-arrow functions, eg setting default values within its argument list?

Alhadis:
Erh, the part of the pattern you pointed out is *only* matching arrow functions.

juergba:
I don't understand the meaning of `Erh,` but I dare to ask again.

I change the pointer and now point to the first part which processes the non-arrow functions.
What about non-arrow functions, eg setting default values **with brackets** within its argument list?

Alhadis:
An arrow function's arguments *must* be enclosed by an outermost pair of brackets, with the lone exception of a solitary identifier, which must not have a default value either (or anything resembling an expression).

In other words, these are illegal syntax:

~~~js
arg = {method(){ return true; } => { body }
arg1, arg2 = {foo(){ return "bar"; } => { body } 
~~~

Clear enough?

Alhadis:
As for the non-arrow functions... that was there before I got here. :wink: I'm not responsible for that part.

Alhadis:
@juergba, @outsideris Anything I can do to hurry this along?

PR #3940, opened 2019-06-05
add node 12 to build; remove node 11

boneskull:
node 11 is no longer maintained

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ stages:
 
 # defaults
 language: node_js
-node_js: '11'
+node_js: '12'
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


boneskull:
to be clear about the why's of that hack around browser tests:

1. the browser tests are the only place where we allow native modules to be downloaded and/or compiled.  we need these to launch the browser tests.
1. unfortunately, this also means we get `canvas`, which is a transient dependency of `builder`.
1. `canvas` provides a prebuilt binary for node 10
1. `canvas` does *not* provide a prebuilt binary for node 12, which means we must compile it from source
1. ...and the prerequisites to do so are not available in the default travis environment

So we could either download all of those dependencies (which is a handful of stuff grabbed via `apt`) and build `canvas` from source *which we won't even use, because it's for building the docs*.... or we just run the browser tests in Node.js v10.  the Node.js version doesn't *really* matter, because these are browser tests, after all.

boneskull:
I would not consider this a breaking change as long as we support versions of Node.js prior to v11. We can consider this "implicit support".  

in the far future, when we drop Node.js v10 support, we will be essentially dropping Node.js v11 support as well.

boneskull:
closing in lieu of #3919

PR #3948, opened 2019-06-10
Bug/3840 --forbid-only doesn't recognize `it.only` when `before` crashes

### LINKED ISSUE #3840: --forbid-only doesn't recognize `it.only` when `before` crashes

rpgeeganage:
### Requirements

### Description of the Change

`--forbid-only` doesn't recognize `it.only` when `before` crashes.

This is an attempt to solve the above-mentioned bug.

### Benefits
This is an attempt to solve the above-mentioned bug.

### Possible Drawbacks
None

### Applicable issues
Fixes #3840 

* Is it a bug fix, or does it not impact production code (patch release)? **Yes**


juergba:
https://github.com/mochajs/mocha/issues/3840#issuecomment-474921006
Taking above comment into account I have doubts.

We have a parsing/definition phase and a runtime phase. `describe.only()` plus `--forbid-only` throws an error during the parsing phase (lib\interfaces\common.js). This PR with `it.only()` plus `--forbid-only` throws an error during the runtime phase.

IMO it makes sense to throw an error early during the parsing phase, without execution any test.
@boneskull thoughts?

The documentation states:
> --forbid-only causes Mocha to fail when an exclusive ("only'd") test or suite is encountered, and it will abort further test execution.

The docu confirms the runtime solution, since some tests do run before an error is thrown.

PR #3950, opened 2019-06-10
fix: fixed mocha grep regex instantiation, closes #3949

### LINKED ISSUE #3949: fix: mocha grep RegExp instance
On creating an instance of `RegExp` if the given value is `string` the regex-like extraction is not correct:
``` javascript
Mocha.prototype.grep = function(re) {
  if (utils.isString(re)) {
    // extract args if it's regex-like, i.e: [string, pattern, flag]
    var arg = re.match(/^\/(.*)\/(g|i|)$|.*/);
    this.options.grep = new RegExp(arg[1] || arg[0], arg[2]);
  } else {
    this.options.grep = re;
  }
  return this;
};
```
Base on `/^\/(.*)\/(g|i|)$|.*/` if `re` is regex-like the pattern will be stored in `arg[1]` and flags will be stored in `arg[2]` otherwise the whole string is stored in `arg[0]`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp:
``` javascript
new RegExp(pattern[, flags])
```
Regex should be called like this if `re` is regex-like:
``` javascript
new RegExp(arg[1], arg[2])
```
Or like this if `re` is a normal string:
``` javascript
new RegExp(arg[0])
```

mrastiak:
### Description of the Change

There is a problem in the instantiation of `RegExp` in `mocha` `grep`
The problem is explained in https://github.com/mochajs/mocha/issues/3949

Without this passing regex-like string to `grep` will not work correctly

### Benefits

Enhancing `grep` option

### Applicable issues

This is a bug fix

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -428,7 +428,9 @@ Mocha.prototype.grep = function(re) {
   if (utils.isString(re)) {
     // extract args if it's regex-like, i.e: [string, pattern, flag]
     var arg = re.match(/^\/(.*)\/(g|i|)$|.*/);
-    this.options.grep = new RegExp(arg[1] || arg[0], arg[2]);
+    this.options.grep = arg[1]
+      ? new RegExp(arg[1], arg[2])
+      : new RegExp(arg[0]);
   } else {
     this.options.grep = re;
   }


juergba:
Some examples would help to understand this issue. We have several [unit tests](https://github.com/mochajs/mocha/blob/master/test/unit/grep.spec.js) for `Mocha#grep()` which seem to work fine. 
Please advice which test constellations fail with the current version of this function, thanks.

Issue #2167, opened 2016-03-25
Mocha 2.3.4: Errors That Occur Outside of `beforeEach`/`it` in an AMD Module Lose Stack Trace


machineghost:
**Short Version:** If an error occurs inside a Mocha function inside an AMD module, but not inside a Mocha function that expects errors (eg. `it`), the stack trace of the error somehow gets eaten.

**Reproduction Steps:**

Step 1: npm install mocha
Step 2: create `testMain.js` (ie. a file which does nothing but use Require.js to import `testFoo.js`):

```
requirejs = require('node-requirejs').requirejs();
requirejs('testFoo');
```

Step 3: create `testFoo.js` (ie. a simple test file with errors thrown in unexpected places)

```
define([], function() {
    throw new Error('a: no trace');
    describe('Foo', () => {
        throw new Error('b: no trace');

        it('does bar', () => {
            throw new Error('c: has stack trace');
        });
    });
});
```

Step 4: run `mocha testMain.js`; you will get the following output (without a stack trace):

```
[Error: a: no trace]
```

**NOTE 1:** The stack trace is also lost if you comment out error "a" and make it to error "b"
**NOTE 2:** The stack trace is NOT lost if you comment out the first two errors and make it to error "c".
**NOTE 3:** The stack trace is also not lost if you throw a new error before the mocha code starts (ie. if you `throw new Error('d')` before the `define`).

Obviously it's very hard to debug a problem when this happens (you have to resort to "delete debugging", ie. removing random chunks of modified code until the error goes away), so if there's any way Mocha could provide a stack trace under these conditions it would be very helpful.

P.S.  I suppose one could argue that this is a Require.js problem, not a Mocha problem.  However, the fact that everything works as expected outside Mocha in an AMD environment, or inside an AMD environment inside an  `it`/`after`/`beforeEach`/etc., but not inside the other Mocha functions, suggests that Mocha is doing something in those functions that's the root cause of the issue (even though Requirejs/AMD does seem to be a key component).

juergba:
closing, since outdated Mocha version and AMD modules.

JoshuaKGoldberg:
Re-filed in #5106.

JoshuaKGoldberg:
#5107 is merged; we'll release a new version with the fix soon. 🚀

Issue #1979, opened 2015-11-25
🚀 Feature: Return full error with window.onerror handler


roadhump:
Now `window.onerror` handler from `support/browser-entry.js` creates new error with uncaught error message and short location. [Spec](https://developer.mozilla.org/en/docs/Web/API/GlobalEventHandlers/onerror) says that `window.onerror` has full uncaught error as last argument and it seems that all latest browsers support this spec. It would be great for reporting to return full uncaught error for mocha tests in browsers in case it is supported. I am ready to create PR if you find this idea ok.

JoshuaKGoldberg:
I believe this was actually resolved by #4174. 🚀 

If I'm wrong about that, please file a new issue using one of our new templates. They'll ask for a bit more clarifying info that'll help explain why #4174 wasn't a resolution. Cheers!

PR #3952, opened 2019-06-12
Display stack for uncaught errors in the browser

### LINKED ISSUE #5106: 🐛 Bug: Browser doesn't include call stack for global uncaught errors
### Bug Report Checklist

### Expected

When an uncaught error occurs during a test in browsers, you'd expect to see the call stack. That's useful.

```plaintext
Error: Oh no!
    at inner (test.example.js:8:15)
```

### Actual

Mocha's browser error handling code was written before browsers gave nice stacks (https://github.com/mochajs/mocha/commit/eb58761bb3346d2597e8365e60273417127b72c4).

```plaintext
Error: Error: Oh no! (http://127.0.0.1:8081/test.example.js:8)
```

Using a minimal browser setup such as in https://github.com/mochajs/mocha-examples/pull/72:

```js
it("example", (done) => {
    function inner() {
        throw new Error("Oh no!");
    }

    setTimeout(inner);
});
```

### Versions

Mocha: 10.3.0

### Additional Info

This was filed long ago in #2167, then a fix proposed in #3952. Since #2167 is quite old, re-filing to be more current.

prust:
### Description of the Change

This change captures the stack of uncaught exceptions in the browser, so it will be displayed in Mocha. Before this change, uncaught exceptions in the browser are displayed with a URL and line number, but without a stack:

no-stack

This pull request updates the global error handler for the browser so that the stack is captured:

with-stack

The issue was that the browser-specific hook for uncaughtExceptions in `browser-entry.js` was using the pre-2014 three-parameter `window.onerror`, which only provides the error message, the URL and the line number. [In 2014](https://danlimerick.wordpress.com/2014/01/18/how-to-catch-javascript-errors-with-window-onerror-even-on-chrome-and-firefox/), two more parameters were added in both [the spec](http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html#errorevent) and Chrome, with other browsers following shortly thereafter (including IE11) -- the last of these parameters is the Error object itself, which has a populated stack.

### Alternate Designs

The design could be simplified by only supporting the five-parameter `onerror()`, but since mocha is used in all sorts of environments, I figured it was safest to use the `err` parameter if it exists and to fall back to the previous functionality if it doesn't.

The design could be a little more complex if it is important to preserve the message provided to the `onerror` handler, which -- in Chrome, at least -- has an extra "Uncaught Error: " prefix. This would add a couple more lines to the top of `onerror` in `browser-entry.js`:

```javascript
if (err)
  err.message = msg;
```

This would bring consistency to the quality of error reporting in mocha's core (uncaught exceptions in Node and caught exceptions in Node and the Browser all include stacks).

### Benefits

The inclusion of the stack will make it easier for developers to determine the cause of the error.

### Possible Drawbacks

With this particular design choice, if a mocha user was grepping or testing for the browser-supplied "Uncaught Error:" prefix in a browser test, this change would cause it to not be found. If this is important, the two lines mentioned above in "Alternate Designs" could be added.

### Applicable issues

* I think #2167 may be fixed by this pull request (but I'm not familiar with AMD/RequireJS, so I'm not in a good position to test it)

Regarding semver, I think I would call this an enhancement (providing stack traces for uncaught exceptions). I suppose it could be considered a breaking change, since it is hypothetically possible that someone is relying on the browser-supplied "Uncaught Error:" prefix somehow, but this is highly unlikely. If it's a concern, it could be address by the 2-line fix above.

### Tests

I did write a browser-specific test for this (which was orders of magnitude harder than fixing the issue itself), but the uncaught exception is causing the test to fail. I tried passing `--allow-uncaught true/false` in the relevant `mocha.opts`, as well as `mocha.allowUncaught()`, but neither made a difference.

Perhaps I should try putting the test with the `allowUncaught` or `uncaught()` tests in `runner.spec.js`, but it looks like the `allowUncaught` tests always throw an error somewhere that it will be caught, and the `uncaught()` tests don't actually throw errors (they just pass an error object directly to `uncaught()`), the former would probably run into the same issue I'm currently running into and the latter wouldn't exercise the code in question.

If someone with familiarity with the mocha's exception-testing tests could lend a hand or direct me in a direction that is likely to work, I would appreciate it.

Fixes #5106.


prust:
Update: @boneskull outlines the issue & how to fix it beautifully in https://github.com/mochajs/mocha/pull/2983#issuecomment-350428522 and this is already being done in `throw.spec.js`. I'm going to follow suit here (and/or move this test into `throw.spec.js`)

Lindsay-Needs-Sleep:
@juergba 
Can you review this issue?

I have been using this improvement.  It is very useful.  I would like to see it in the next release.

I have also run ‘npm run test’ on this branch, everything passes.

juergba:
@Lindsay-Needs-Sleep I don't have the browser knowledge to review this PR.
Mocha is still supporting IE11, [this: Browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) doesn't help neither.

Lindsay-Needs-Sleep:
@juergba Do you know who might have the knowledge (and power to review this?)

Due to the way javascript works, if a browser does not support the 4th and 5th arguments of `onerror` they will just have a value of `undefined` when `onerror` is triggered.  This is fine because in that situation, this line:
```
fn(err || new Error(msg + ' (' + url + ':' + line + ')'));
```
will evaluate to this:
```
fn(new Error(msg + ' (' + url + ':' + line + ')'));
```
Which is exactly what happened before this change.

Basically, this change just results in an improved error object for browsers that do support the 4th and 5th args.  Browsers that don't, will be unaffected.

(Also, I did try out IE11.  It does support the 4th and 5th args! Yay!)

juergba:
@prust Are you still interested in this PR?
I'm trying to understand and finally get it merged.

If yes, could you please rebase? I would like to push your branch to our repo, to make sure all tests do pass correctly. There are additional tests running (eg IE11).

Does this PR solve #1979 ?

juergba:
I did some testing in Chrome with `fullTrace: true` and `reporter: 'spec'`:
```js
describe('uncaught', function() {
  it('throw delayed error', (done) => {
    setTimeout(() => {
      throw new Error('Whoops!');
    }, 10)
    setTimeout(done, 10);
  });
```
Now there is even less stack information printed than before. I guess skipping the `msg` is not a good idea. An `uncaughtException` does not always mean that the runner throws and aborts. In above example the handler recovers from the error (fails the still open test) and then keeps going. In this case we loose stack information now.
I will do more testing tomorrow.

JoshuaKGoldberg:
Moving over to #5107 with a co-authored-by attribution. Cheers! 🤎

JoshuaKGoldberg:
Ah @prust hello! I'd just made a new PR because I didn't have permissions to push to [`CSNW:master`](https://github.com/CSNW/mocha/tree/master). If you want to make a new PR from another branch I'm happy to switch to that one, to preserve attribution for you as the author. No worries if you don't have the time though. 🙂

JoshuaKGoldberg:
#5107 is merged; we'll release a new version with the fix soon. 🚀

prust:
Thanks @JoshuaKGoldberg!

PR #3953, opened 2019-06-15
Collect test files later

geigerzaehler:
### Requirements

As part of the improvements to file watching (#3912) we want Mocha to pick up test files that are added while mocha is running in watch mode. This PR makes the necessary preparation to be able implement this behavior.

Our goal is to make the `watchRun` function responsible for collecting the tests files so that it can decide to collect test files again in the future.

### Description of the Change

Currently test files are collected with the `handleFiles` function in the `handler` in `lib/cli/run`. The collected files are then passed to `watchRun` and `singleRun`. We change this by having `watchRun` and `singleRun` collecting the test files.

We rename `handleFiles` to `collectFiles` and move it to a separate module. (This is necessary because `watchRun` and `singleRun` live in different modules.) The implementation remains unchanged.

In order for `watchRun` and `singleRun` to be able to call `collectFiles` we also need to pass all command line options required for file collection to these two functions. To accomplish this we now just pass the command line options object down, making the signatures of functions more uniform.

Edit: The implementation slightly changed. See the [comment below](#issuecomment-505516375)

### DIFF
--- a/lib/cli/collect-files.js
+++ b/lib/cli/collect-files.js
@@ -1,0 +1,84 @@
+'use strict';
+
+const path = require('path');
+const ansi = require('ansi-colors');
+const debug = require('debug')('mocha:cli:run:helpers');
+const minimatch = require('minimatch');
+const utils = require('../utils');
+
+/**
+ * Exports a function that collects test files from CLI parameters.
+ * @see module:lib/cli/run-helpers
+ * @see module:lib/cli/watch-run
+ * @module
+ * @private
+ */
+
+/**
+ * Smash together an array of test files in the correct order
+ * @param {Object} opts - Options
+ * @param {string[]} opts.extension - File extensions to use
+ * @param {string[]} opts.spec - Files, dirs, globs to run
+ * @param {string[]} opts.ignore - Files, dirs, globs to ignore
+ * @param {string[]} opts.file - List of additional files to include
+ * @param {boolean} opts.recursive - Find files recursively
+ * @param {boolean} opts.sort - Sort test files
+ * @returns {string[]} List of files to test
+ * @private
+ */
+module.exports = ({ignore, extension, file, recursive, sort, spec} = {}) => {
+  let files = [];
+  const unmatched = [];
+  spec.forEach(arg => {
+    let newFiles;
+    try {
+      newFiles = utils.lookupFiles(arg, extension, recursive);
+    } catch (err) {
+      if (err.code === 'ERR_MOCHA_NO_FILES_MATCH_PATTERN') {
+        unmatched.push({message: err.message, pattern: err.pattern});
+        return;
+      }
+
+      throw err;
+    }
+
+    if (typeof newFiles !== 'undefined') {
+      if (typeof newFiles === 'string') {
+        newFiles = [newFiles];
+      }
+      newFiles = newFiles.filter(fileName =>
+        ignore.every(pattern => !minimatch(fileName, pattern))
+      );
+    }
+
+    files = files.concat(newFiles);
+  });
+
+  if (!files.length) {
+    // give full message details when only 1 file is missing
+    const noneFoundMsg =
+      unmatched.length === 1
+        ? `Error: No test files found: ${JSON.stringify(unmatched[0].pattern)}` // stringify to print escaped characters raw
+        : 'Error: No test files found';
+    console.error(ansi.red(noneFoundMsg));
+    process.exit(1);
+  } else {
+    // print messages as an warning
+    unmatched.forEach(warning => {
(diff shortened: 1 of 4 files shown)


lib/cli/run.js
```diff
@@ -290,8 +289,5 @@ exports.builder = yargs =>
 exports.handler = argv => {
   debug('post-yargs config', argv);
   const mocha = new Mocha(argv);
-  const files = handleFiles(argv);
-
-  debug('running tests with files', files);
-  runMocha(mocha, argv, files);
+  runMocha(mocha, argv);
```
juergba: This is existing code, I know. But `argv` is passed twice to `runMocha()`, once with `mocha` instance and a second time as a function parameter. Some options may have been processed meanwhile within `mocha`.

geigerzaehler:
> > To accomplish this we now just pass the command line options object down, making the signatures of functions more uniform.
> 
> This object is rather messy, it includes canonical flags, its aliases, its camel-cased versions, evt. node flags, unkown/silently ignored flags... I don't really like the idea to pass this object further down. But let's wait for addional feedback first.

We could address this by by extracting all the required options from `argv` in `runMocha` and passing only those on to `singleRun` and `watchRun`.

lib/cli/run.js
geigerzaehler: Do you have a suggestion how to improve this?

juergba: At least `grep` is processed within the `Mocha` constructor, therefore using the unprocessed `grep` of `argv` seems incorrect to me. Why don't we use just `mocha.options`?

I also wonder why I don't see `fgrep` anywhere. It is also processed within the `Mocha` constructor, but not used in `runMocha()`.

juergba:
> We could address this by by extracting all the required options from argv in runMocha and passing only those on to singleRun and watchRun.

I prefer this way, but your opinion does count as well for me.

geigerzaehler:
@juergba: I added a [commit](https://github.com/mochajs/mocha/pull/3953/commits/ad0a13daa0c90a51654c973464efddd7a5af4b3a) that addresses your concern about passing the CLI options as is. I believe the change is a definite improvement.

Instead of passing the whole object with all the CLI options down to `singleRun` and `watchRun` and then to `collecFiles` we create a dedicated object with all the parameters for collecting test files and pass this object.

This approach has the advantage that we only make only the options that are strictly necessary available to the two `run` functions. This also consolidates the normalization of CLI options in one place (`runMocha`) instead of spreading it also to `collectFiles`.

lib/cli/run.js
geigerzaehler: Is it ok if I address this separately? It seems to be an orthogonal concern to me and there is a chance this changes the behavior, something I’d like to avoid. It also looks like this should be tackled more comprehensively by rethinking whether `handler` and `runMocha` should be separate functions.

geigerzaehler: I started to address this in https://github.com/mochajs/mocha/pull/3960 which should also answer the question about `fgrep`.

lib/cli/collect-files.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+const path = require('path');
+const ansi = require('ansi-colors');
+const debug = require('debug')('mocha:cli:run:helpers');
+const minimatch = require('minimatch');
+const utils = require('../utils');
+
```
juergba: Could you add this "@module" and "@private" stuff, please. To be honest I don't know the purpose of it, but let's follow the line of the other modules.

lib/cli/run-helpers.js
```diff
@@ -93,116 +91,56 @@ exports.handleRequires = (requires = []) => {
...
-  }
-
-  // add files given through --file to be ran first
-  files = fileArgs.concat(files);
-  debug('files (in order): ', files);
-  return files;
-};
-
 /**
  * Give Mocha files and tell it to run
```
juergba: This description should be adapted.

lib/cli/watch-run.js
```diff
@@ -15,14 +16,16 @@ const Mocha = require('../mocha');
...
  * @param {string|RegExp} opts.grep - Grep for test titles
  * @param {string} opts.ui - User interface
- * @param {string[]} opts.files - Array of test files
+ * @param {Object} fileCollectParams - Parameters that control test
+ *   file collection. See `lib/cli/collect-files.js`.
+ * @param {string[]} fileCollectParams.extension - List of extensions to watch
  * @private
  */
-module.exports = (mocha, {extension, grep, ui, files}) => {
+module.exports = (mocha, options, fileCollectParams) => {
```
juergba: `options`: should be `opts` or `{grep, ui}` ?

PR #3956, opened 2019-06-24
fix: do not redeclare variable

DanielRuf:
### Description of the Change
There is a variable redeclared in an example which would not work in some cases. In fact we do not have to redeclare this variable.

### Alternate Designs
N/A

N/A

### Benefits
N/A

### Possible Drawbacks
N/A

### Applicable issues
N/A

### DIFF
--- a/docs/example/Array.js
+++ b/docs/example/Array.js
@@ -13,7 +13,7 @@ describe('Array', function(){
       var arr = [];
       var n = arr.push('foo');
       expect(n).to.equal(1);
-      var n = arr.push('bar');
+      n = arr.push('bar');
       expect(n).to.equal(2);
     })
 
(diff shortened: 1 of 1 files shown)


craigtaub:
Thanks for the contribution.

craigtaub:
Could you have a look at the failures? Then we can get this in, thanks

DanielRuf:
https://travis-ci.org/mochajs/mocha/jobs/549688891#L471

> curl: (7) Failed to connect to nodejs.org port 443: No route to host

Not related to my changes as it seems.
Can you trigger the failing jobs to rerun?
Only people with the needed rights can do this.

craigtaub:
Re-triggering seems to work but unfortunately can't do it for deploy preview and need that to be green. If you could rebase and push that would be really helpful please.

DanielRuf:
I'll check later.

docs/example/Array.js
```diff
@@ -70,4 +70,5 @@ describe('Array', function(){
       expect(arr).to.have.length(1);
     })
   })
-})
\ No newline at end of file
+})
+
```
outsideris:
```suggestion

```

PR #3960, opened 2019-06-25
Don't re-initialize grep option on watch re-run

### LINKED ISSUE #2027: --fgrep doesnt work with --watch
### Expected

when using `--fgrep "my search string" --watch` one would think a file change would run tests with the provided fgrep string.
### Actual

all tests re-run when a file is changed as if `--fgrep` was never used.

geigerzaehler:
### Requirements

* Fix #2027
* Remove unexplained and unneeded code
* Clearer separation of responsibilities and dependencies.

### Description of the Change

We remove code that called `mocha.grep(null)` on watch re-runs if the `--grep` option was not supplied. The code seems to serve no purpose and is the cause of #2027.

### Alternate Designs

N.A.

### Benefits

Fixes #2027

We consolidate CLI option handling. The majority of CLI options are handled by the `Mocha` constructor. But this reponsibility is shared with `watchRun` which changes behavior based on a CLI option. With this change we reduce the reliance of `watchRun` on the CLI options, a point that was brought up in https://github.com/mochajs/mocha/pull/3953#discussion_r294672564. To finish this it would make sense to follow up with a PR that makes the use of the `ui` option unnecessary.

### Possible Drawbacks

The code might actually serve some purpose that is not covered by the test suite. Since it was introduced in #3556 maybe @boneskull can shed some light on this.

### Applicable issues

Fixes #2027

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -118,7 +118,6 @@ exports.runMocha = (mocha, options) => {
   const {
     watch = false,
     extension = [],
-    grep = '',
     ui = 'bdd',
     exit = false,
     ignore = [],
@@ -138,7 +137,7 @@ exports.runMocha = (mocha, options) => {
   };
 
   if (watch) {
-    watchRun(mocha, {ui, grep}, fileCollectParams);
+    watchRun(mocha, {ui}, fileCollectParams);
   } else {
     exports.singleRun(mocha, {exit}, fileCollectParams);
   }
--- a/lib/cli/watch-run.js
+++ b/lib/cli/watch-run.js
@@ -16,14 +16,13 @@ const collectFiles = require('./collect-files');
  * Run Mocha in "watch" mode
  * @param {Mocha} mocha - Mocha instance
  * @param {Object} opts - Options
- * @param {string|RegExp} opts.grep - Grep for test titles
  * @param {string} opts.ui - User interface
  * @param {Object} fileCollectParams - Parameters that control test
  *   file collection. See `lib/cli/collect-files.js`.
  * @param {string[]} fileCollectParams.extension - List of extensions to watch
  * @private
  */
-module.exports = (mocha, {grep, ui}, fileCollectParams) => {
+module.exports = (mocha, {ui}, fileCollectParams) => {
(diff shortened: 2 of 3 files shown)


juergba:
I did some testing and indeed both `grep` and `fgrep` seem to work now.
`fgrep` is converted into `grep` within the `Mocha` constructor.

Since `watch` reuses the same `mocha` instance, I don't see any reason why `mocha.grep`should be set to `null`.

juergba:
@geigerzaehler please resolve the conflicts, thanks.

geigerzaehler:
> Could we add tests for this in `test/integration/options/watch.spec.js`?

I’d like to propose an alternative to this: Let’s extract the “reset code”
https://github.com/mochajs/mocha/blob/b2148622460863050ab59f629dba729825d06326/lib/cli/watch-run.js#L66-L68
and test it with unit tests. (I would take care of this in a separate PR)

The watch integration tests are quite expensive (they take two seconds each) and require a lot of code to understand. The more lightweight tests would give us the same test coverage but a lot faster. We would also be able to extend the tests and assert some additional behavior.

juergba:
@outsideris There are unit tests for `Mocha#fgrep()` which matches `fgrep` into `grep`. And `mocha.grep === mocha.grep` since it's not updated anymore.

@geigerzaehler please add one test with a `fgrep` input.
We can pick up your idea with the "reset code" extraction when refactoring the `watch` code. It doesn't cover the specific `fgrep` case.

geigerzaehler:
Could you have another look @juergba?

juergba:
@geigerzaehler I'm glad you are still alive, last few days have been disappointing.

I need this test I asked you for, it's not a question of making sense or not. I'm sorry. Most probably there will be no more input to this PR, so we can merge it afterwards.

geigerzaehler:
Sorry, @juergba. I actually added this test two weeks ago but didn’t notice that I did not push 🤦‍♂. Everything should be there now.

geigerzaehler:
> If you have some time left, you could fix some typing errors in the test file comments (L119, L177, L187).

Done, squashed, and rebased.

PR #3967, opened 2019-07-20
Remove deprecated "getOptions()" and "lib/cli/options.js"

juergba:
### Description of the Change

The function `getOptions()` has its origin in times before `Yargs` and was used to load `mocha.opts` options. It was deprecated with Mocha v6.0.0 and is no longer used internally.
Since the implementation of `Yargs` it has been replaced by the function `loadMochaOpts()`.

Additionally we remove the deprecated "bin/options.js" which source was moved to "lib/cli/options.js".


juergba:
@craigtaub @outsideris @boneskull 
I will merge this PR within a few days.

PR #3971, opened 2019-07-24
Adopt the OpenJSF Code of Conduct

craigtaub:
### Description of the Change

First task for https://github.com/mochajs/admin/issues/7. Just a blanket copy-paste from [here](https://github.com/openjs-foundation/cross-project-council/blob/master/CODE_OF_CONDUCT.md).
Let me know if you think there is a nicer format or way to do this.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -55,7 +55,7 @@ Original text courtesy of the Speak Up! project and Django Project.
 
 ## QUESTIONS?
 
-If you have questions, please see the FAQ. If that doesn’t answer your questions, feel free to email conduct@js.foundation.
+If you have questions, please see the FAQ. If that doesn’t answer your questions, feel free to email report@lists.openjsf.org.
 
 # OpenJS Foundation Code of Conduct
 
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -33,14 +33,14 @@ community should be respectful when dealing with other members as well as with p
 We are a community of professionals, and we conduct ourselves professionally. Be kind to others. Do not insult or put
 down other participants. Harassment and other exclusionary behavior aren’t acceptable. This includes, but is not limited to:
 
-* Violent threats or language directed against another person.
-* Discriminatory jokes and language.
-* Posting sexually explicit or violent material.
-* Posting (or threatening to post) other people’s personally identifying information (“doxing”).
-* Personal insults, especially those using racist or sexist terms.
-* Unwelcome sexual attention.
-* Advocating for, or encouraging, any of the above behavior.
-* Repeated harassment of others. In general, if someone asks you to stop, then stop.
+- Violent threats or language directed against another person.
+- Discriminatory jokes and language.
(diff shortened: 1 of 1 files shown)


boneskull:
@munter @outsideris @juergba Please take a look at this

.github/CODE_OF_CONDUCT.md
```diff
@@ -56,3 +56,39 @@ Original text courtesy of the Speak Up! project and Django Project.
...
+- [english](https://www.contributor-covenant.org/version/1/4/code-of-conduct)
+- [translations](https://www.contributor-covenant.org/translations)
+
+Refer to the section on reporting and escalation in this document for the specific emails that can be used to report and escalate issues.
+
+## Reporting
+
+### Project Spaces
+
+For reporting issues in spaces related to a member project please use the email provided by the project for reporting. Projects handle CoC issues related to the spaces that they maintain. Projects maintainers commit to:
```
outsideris: `please use the email provided by the project for reporting`

Have we the email?

craigtaub: Good catch. 
Sounds like pointing to Gitter is not acceptable. Altho seems overkill to introduce an email just for this, but if its part of the CoC 🤷‍♂ .

@boneskull @Munter @juergba any thoughts?

.github/CODE_OF_CONDUCT.md
```diff
@@ -56,3 +56,39 @@ Original text courtesy of the Speak Up! project and Django Project.
 ## QUESTIONS?
 
 If you have questions, please see the FAQ. If that doesn’t answer your questions, feel free to email conduct@js.foundation.
+
```
juergba: `conduct@js.foundation` probably is outdated, should be something with openjs.

.github/CODE_OF_CONDUCT.md
juergba: If I call a hotline to complain about a service and I never get an answer, that's just a pure provocation. Better no hotline at all, than a hotline that doesn't really care.
So an email without defining a responsible maintainer: IMO no.
Do we have resources for this: IMO no.

.github/CODE_OF_CONDUCT.md
craigtaub: I'll update to `report@lists.openjsf.org`

.github/CODE_OF_CONDUCT.md
outsideris: I agree with no hotline. 
So, we will remove this? And can we remove this because it is OpenJS template?

craigtaub: I didnt think we should remove it for that reason. I feel as a whole we do adhere to the `Projects maintainers commit to` points.

craigtaub:
@Munter think your the last one left to look at this. Pls have a look when u get a mo. Thanks

.github/CODE_OF_CONDUCT.md
boneskull: I will be the point of contact for this for now

boneskull:
It’s been sitting long enough I think it’s ok to merge.

PR #3987, opened 2019-08-16
feat: make forbid-only as default on CI

### LINKED ISSUE #3030: Make --forbid-only (but not --forbid-pending) default
Eventually let's make `--forbid-only` the default and use `--no-forbid-only` to disable it. Lots of CI will get safer for free!

(Pending, on the other hand, isn't so dangerous to begin with and has more long-term use-cases. It'll be inconsistent, but let's leave pending as-is.)

JaeHyeonKim19:
### Description of the Change

* When `process.env.CI` is true, `forbid-only : true` is default value. And I changed one of conditional statements in `package-scripts.js` to keep its function.

### Alternate Designs

* There're may be a better way, but it's my best. 😂

* Due to this change, We can prevent pushing to remote repo with `.only`.

### Benefits

* Due to this change, We can prevent pushing to remote repo with `.only`.

### Possible Drawbacks

* If this change apply to mocha, Existing users can be confused. Because project which use '.only' on purpose gonna make conflict on CI. So We've to announce about this change when this change is applied.

### Applicable issues

* close #3030


outsideris:
Tests which use a fixture that has `.only` are failed. In Ci, we should handle this.

JaeHyeonKim19:
Any review😂?

JoshuaKGoldberg:
👋 coming back to this @JaeHyeonKim19, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

_(I do see that your last message was asking for review - unless directed otherwise try to review it once we're ramped up enough!)_

package-scripts.js
```diff
@@ -15,8 +15,8 @@ function test(testName, mochaParams) {
   if (process.env.CI) {
     // suppress coverage summaries in CI to reduce noise
     coverageCommand += ' --reporter=json';
-    if (!/^only-/.test(testName)) {
-      mochaParams += ' --forbid-only';
+    if (/^only-/.test(testName)) {
+      mochaParams += ' --no-forbid-only';
```
JoshuaKGoldberg, line 19: [Question] 🤔 Is `--no-forbid-only` necessary? My interpretation of https://github.com/mochajs/mocha/issues/3030#issuecomment-333044369 was that the issue can be resolved with just the change to `forbid-only`'s default value.

lib/cli/run.js
JoshuaKGoldberg, line 1: [Docs] https://mochajs.org/#-forbid-only should be updated to mention this, too. Folks will want to know if the default value isn't always falsy.

JoshuaKGoldberg (changes requested):
The functional default value change looks good to me! Requesting changes I'm not sold on the `--no-forbid-only`, and would like to see docs. Thanks! 🙌 

(if you don't have time for this, we'll go ahead and send a new PR after a while)

Issue #3675, opened 2019-01-17
Mocha failing when error message too big


furstenheim:
### Description

Mocha never ends with big error messages

### Steps to Reproduce

```
const {expect} = require('chai')
 it('Test failing', function () {
      const longArray = _.times(10000, function (i) {
        return {a : i}
      })
      const shortArray = []
      expect(longArray).deep.equal(shortArray)
})
```

Then execute `yarn mocha test --exit`, the test will execute it will be marked as failure, but the summary will never appear and the process will not exit.

If on contrary one writes:

```
const {expect} = require('chai')
 it('Test failing', function () {
      const longArray = _.times(10000, function (i) {
        return {a : i}
      })
      const shortArray = []
      try {
        expect(longArray).deep.equal(shortArray)
     } catch (e) {
        console.error(e)
        throw new Error('Small error')
     }
})
```

The test fails and it displays an error in the summary. Error is logged without any problem.

**Expected behavior:** [What you expect to happen]
Mocha should be able to exit the process even if the error is overly long. Maybe the message could be reduced?

**Actual behavior:** [What actually happens]
The process hangs forever

**Reproduces how often:** [What percentage of the time does it reproduce?]
Always
### Versions
Mocha 5.1.1

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`:
- The output of `node --version`:
v8.11.2
- The version and architecture of your operating system:
Ubuntu
- Your shell (bash, zsh, PowerShell, cmd, etc.):
bash
- Your browser and version (if running browser tests):
- Any other third party Mocha related modules (with versions):
- The code transpiler being used:

boneskull:
Please retry this but use the built in `assert` module. same result? if so, the problem may be in our diff-generating code. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

plroebuck:
If it just takes a long time, is that necessarily a _"confirmed bug"_?

outsideris:
Because it doesn't take so long without mocha. 
So, I thought there is a performance issue in mocha even if I didn't find why.

cyjake:
@furstenheim Try running mocha again with `--full-trace` enabled, if it performs well then it sure is related to #3686

furstenheim:
@cyjake Just tried mocha 5.2.0 (latest) with `--full-trace` option and the problem persists

cyjake:
[...]

PR #3989, opened 2019-08-16
WIP: replaced diff to diff-match-patch (#3675)

jalnanco:
### Description of the Change
I checked the #3675  issue and changed the existing diff library to Google's diff-patch-match.

### Benefits
* Diff speed is improved for large data.

Test Code (copy from issue)
```javascript
var _ = require('lodash');
const {expect} = require('chai')
it('Test failing', function () {
  const longArray = _.times(10000, function (i) {
    return {a : i}
  })
  const shortArray = []
  expect(longArray).deep.equal(shortArray)
}
```

Before change: About 2 minutes 1 second
```bash
/bin/mocha ../../pycon/mocha/my/test_issue  121.94s user 0.55s system 100% cpu 2:01.74 total
```
 After change: About 1 second
```bash
./bin/mocha ../../pycon/mocha/my/test_issue  0.91s user 0.09s system 93% cpu 1.071 total
```

### Applicable issues

Diff-patch-match is character-based, unlike diff.
So it is written to be processed line by line using line-mode.

However, the result screen is different because there is a difference from Patch Process.

Test Code
```javascript
it('Test failing2', function () {
  const expected = { a: 123 };
  const actual = { a: 123, b: 456 };
  expect(expected).deep.equal(actual);
})
```

#### Comparison of Patch Results
Before
``` javascript
Index: string
===================================================================
--- string
+++ string
@@ -1,3 +1,4 @@
 {
   "a": 123
+  "b": 456
 }
\ No newline at end of file
```

After
```javascript
@@ -6,9 +6,20 @@
 a": 123

+  "b": 456

 }
```

#### Comparison of Excution Results
Before
``` javascript
       {
         "a": 123
      +  "b": 456
       }
```

After
```javascript
       a": 123
      +  "b": 456
       }
```

#### RESULT
So I don't think it's user friendly.

Would it be better to work on the same output as the existing result? Or would you like to hear other ways to find out?

Please, I want to listen to the comments and advices.

* All test code is run on MacBook Pro 2017 (16GB)
* reference links
[Line mode](https://github.com/google/diff-match-patch/wiki/Line-or-Word-Diffs)
[That library is character-based](https://github.com/google/diff-match-patch/wiki/Unidiff)

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -8,6 +8,7 @@
 
 var tty = require('tty');
 var diff = require('diff');
+var DiffMatchPatch = require('diff-match-patch');
 var milliseconds = require('ms');
 var utils = require('../utils');
 var supportsColor = process.browser ? null : require('supports-color');
@@ -394,6 +395,28 @@ function inlineDiff(actual, expected) {
 }
 
 /**
+ * diff-match-patch
+ *
+ * @private
+ * @param {String} actual
+ * @param {String} expected
+ * @return {string} The diff.
+ */
+
+function createPatch(actual, expected) {
+  var dmp = new DiffMatchPatch();
+  var a = dmp.diff_linesToChars_(actual, expected);
+  var lineText1 = a.chars1;
+  var lineText2 = a.chars2;
+  var lineArray = a.lineArray;
+  var diffs = dmp.diff_main(lineText1, lineText2);
+  dmp.diff_charsToLines_(diffs, lineArray);
+  var patch = dmp.patch_make(diffs);
+  var msg = dmp.patch_toText(patch);
+  return decodeURI(msg);
+}
+
+/**
  * Returns unified diff between two strings with coloured ANSI output.
  *
  * @private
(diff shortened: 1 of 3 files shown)


wnghdcjfe:
I tested it with your code and there was no problem. What part was wrong?
```js

const {expect} = require('chai')
it('Test failing2', function () {
    const expected = { a: 123 };
    const actual = { a: 123, b: 456 };
    expect(expected).deep.equal(actual);
  })
```
```
  1) Test failing2:

      AssertionError: expected { a: 123 } to deeply equal { a: 123, b: 456 }
      + expected - actual

       {
         "a": 123
      +  "b": 456
       }
```

jalnanco:
> I tested it with your code and there was no problem. What part was wrong?
> 
> ```js
> const {expect} = require('chai')
> it('Test failing2', function () {
>     const expected = { a: 123 };
>     const actual = { a: 123, b: 456 };
>     expect(expected).deep.equal(actual);
>   })
> ```
> 
> ```
>   1) Test failing2:
> 
>       AssertionError: expected { a: 123 } to deeply equal { a: 123, b: 456 }
>       + expected - actual
> 
>        {
>          "a": 123
>       +  "b": 456
>        }
> ```

If you use Google's diff-patch-match instead of old diff.
You can get speed but It's difficult to recognize.

So Test is not a problem. Result output is a problem.

Issue #2159, opened 2016-03-17
🚀 Feature: Indent all lines on message


piranna:
As discussed at https://github.com/BadgeLabs/mocha-eslint/issues/38#issuecomment-197735716 `mocha` only indent the first line of the report message, making it ugly since it gets unaligned with the mocha output. I've also seen that the code to indent the message is duplicated in several of the reporters, so if you like the idea I could try to unify them and implement it myself.

/cc @Turbo87

danielstjules:
Can you give an example where formatting is broken? E.g. this seems correct to me:

```
describe('suite', function() {
  it('test', function() {});
  it('test2', function() {
    throw new Error('foo');
  });
});

$ mocha example.js


  suite
    ✓ test
    1) test2


  1 passing (10ms)
  1 failing

  1) suite test2:
     Error: foo
      at Context.<anonymous> (example.js:4:11)
```

Turbo87:
@danielstjules the issue is visible if you have a multi-line error message:

```js
describe('suite', function() {
  it('test', function() {});
  it('test2', function() {
    throw new Error('foo\nbar\nbaz');
  });
});
```

```
$ mocha example.js


  suite
    ✓ test
    1) test2


  1 passing (11ms)
  1 failing

  1) suite test2:
     Error: foo
bar
baz
      at Context.<anonymous> (/private/tmp/example.js:4:11)
```

piranna:
Yes, the problem is the one that @Turbo87 has described, if the error message has multiple lines they get unindented. It would be as simple as split the message on new lines and apply the indention to all of them.

2kindsofcs:
Hi there. I checked out exactly the same issue is still present. I would like to fix this problem.

JoshuaKGoldberg:
This feels to me like something that should be behind an option. It's useful to be able to select & copy the exact characters of an error in some cases.

JoshuaKGoldberg:
Coming back to this: per #5027 we're trying not to make too many big changes, and there are already a lot of options in Mocha. This issue has received, on average, about one positive interaction every 1.5 years since it was filed. I don't think it makes the cut.

If you really want this to happen, please do ask to re-open. We certainly can. But please also provide a concrete proposal for what the option would look like. 🙂

piranna:
> please also provide a concrete proposal for what the option would look like. 🙂

I think @Turbo87 already showed the actual issue...

JoshuaKGoldberg:
[...]

PR #3991, opened 2019-08-16
Add indentation to error message when it has several line breaks(#2159)

2kindsofcs:
At the moment mocha only indent the first line of the report message.
This PR would indent all lines of the message when it has line breaks so that it would look neat. 
I checked out that it worked out with the example below.
```
  it('test', function() {});
  it('test2', function() {
    throw new Error('foo\nbar\nbaz');
  });
});
```
by the way, I ran "npm test" before creating PR and got 185 passing, 4 pending, 6 failing.
but the errors seems regardless of code that I added. 
So I ran "npm test" again after commenting out changes I made and got exactly same result. 
[errlog_original.txt](https://github.com/mochajs/mocha/files/3509598/errlog_original.txt)
I was wondering if there's any other ways to check this changes is safe.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -257,7 +257,14 @@ exports.list = function(failures) {
       }
       testTitle += str;
     });
-
+    if (msg.includes('\n')) {
+      var errStrSplit = msg.split('\n');
+      var strIndented = errStrSplit[0];
+      errStrSplit.slice(1).forEach(function(word) {
+        strIndented = strIndented + '\n     ' + word;
+      });
+      msg = strIndented;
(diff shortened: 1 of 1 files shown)


outsideris:
@2kindsofcs Please sign CLA by clicking the CLA badge above.

2kindsofcs:
@outsideris I signed CLA jsut a minute ago. thanks.

outsideris:
ref: https://github.com/mochajs/mocha/issues/2159

marcbachmann:
This issue is still present. Is there any reason not to merge this?

PR #3996, opened 2019-08-19
[WIP] FIX issue #3990 & Question for progress

DrizzlingCattus:
#### Abstraction 

This PR is WIP, because there is confusing things.

So i make **question** section and if question is resolved, then i'll update commit.

All environment and test dir structure is almost set like #3990 .

#### Progress

First I check original process.argv like below picture.

bin_mocha_argv_debug

'mocha:cli:mocha real raw argv' is what i want and I found one problem.

'test/issue/external*.js' and 'test/issue/**/*.js' is extended, so it cannot distinguish what is ignore files or mocha spec files.

To make problem simple, I use --file option like below.

2_using_file_option

Next, i check yargs-parser result that is output of yargs-parser.detailed() in lib/cli/options.js.

I found problem that issue #3990 refer.

2_1_using_file_option

Ignore only have one file, but it must have external/1.js, ,2.js and 3.js.

So, i search some options and have identified the cause of the bug.

There are yargs-parser options and 'nargs' is a point.

'nargs' option limit the command's argument count, like nargs.ignore = 1, that means ignore option's argument count is only one.

It is happend in this code, so it cannot accept several ignore argument, like --ignore a b c.

3_problem_narg

nargs.ignore is 1, so yargs-parser inner also interpret ignore is only accept one, although it set array option. (array option is for accepting '--ignore a b c')

3_1_log_of_yargs_Inner

3_2_yargs_parser_inner

Last, i delete narg.ignore for accepting more than one arguments, like below code. 

4_2_code_for_test

It works well. ignore has 3 files.

4_1_result_options

##### First

I don't know how to distinguish between the option input and mocha spec input.

```
./bin/mocha --ignore test/external/*.js test/*
```

how to distinguish what is ignore files or mocha spec files before?

##### Second

I found this code that make yargs-parser's narg option.

```
/**
 * We do not have a case when multiple arguments are ever allowed after a flag
 * (e.g., `--foo bar baz quux`), so we fix the number of arguments to 1 across
 * the board of non-boolean options.
 * This is passed as the `narg` option to `yargs-parser`
 * @private
 * @ignore
 */
const nargOpts = types.array
  .concat(types.string, types.number)
  .reduce((acc, arg) => Object.assign(acc, {[arg]: 1}), {});
```

I was confused to see this. The kind of options like ignore requires multiple arguments. But this code comment say that mocha don't want to use '--foo bar baz quux'. Then, how should we take multiple arguments?

This comment stopped me from working on more. (I cannot change code logic, because i don't ensure what is right)

### Applicable issues

* Is this a breaking change (major release)?
* Is it an enhancement (minor release)?

WIP, so I can't judge yet.

Thank you for reading this PR! :)

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -118,6 +118,7 @@ const parse = (args = [], defaultValues = {}, ...configObjects) => {
     []
   );
 
+  delete nargOpts['ignore'];
   const result = yargsParser.detailed(args, {
     configuration,
     configObjects,


juergba:
If I inspect `process.argv` it's obviously your shell - not Mocha - which expands
- `test/issue/external/*.js` to
- `test/issue/external/1.js, test/issue/external/2.js, test/issue/external/3.js`

This is different in Windows, which AFAIK doesn’t support shell wildcard expansion.

Please try to use double or single quotes, see [docu](https://mochajs.org/#the-test-directory).

DrizzlingCattus:
@juergba 
Humm... Taken together, this issue is due to misuse of ignore option.
I could learn new things here. (I was surprised that the Windows shell doesn't support wildcard expansion.)
Thank you for answer!

Issue #3667, opened 2019-01-09
🐛 Bug: mochajs.org needs a CSP (Content Security Policy)


boneskull:
From [this comment](https://github.com/mochajs/mocha/pull/3652#issuecomment-452403364) by @plroebuck w/ comments by @munter.

JoshuaKGoldberg:
Closing per https://github.com/mochajs/mocha/pull/3999#issuecomment-653881172. Someone yell at me if that's wrong. 🙂

PR #3999, opened 2019-08-26
Add a CSP(content security policy)

sconeman:
### Description of the Change

I checked the #3667 issue and add a CSP(content security policy).
The CSP specification is referenced at `https://github.com/`.
Maybe we need to discuss the CSP specification further.

### DIFF
--- a/docs/_includes/default.html
+++ b/docs/_includes/default.html
@@ -3,6 +3,12 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
+    <meta
+      http-equiv="Content-Security-Policy"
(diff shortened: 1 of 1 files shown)


craigtaub:
Would be really helpful if you could explain in description what each rule is doing.

Munter:
I'm closing this because its outdated.

My previous experience leads me to not recommend using CSP without also setting up rigorous monitoring and active triaging and handling of all cases. This requires a high level of activity, and I don't think we have the bandwidth to spare on it right now.

If we ever do want to enable CSP on our website we can do so by switching on `--contentsecuritypolicy` in assetgraph-builder, which will auto generate a policy based on its knowledge of our static assets dependency graph

PR #4003, opened 2019-08-28
Remove superfluous 'files'

arku:
Removes unnecessary 'files' from the text "Explicitly _include_ a test file to be loaded before other test **files files**".


juergba:
This typo will be fixed in another PR. Thank you.

PR #4004, opened 2019-08-28
Mocha constructor: some fixes and cleanup

juergba:
### Description

When Mocha is launched via CLI a whole bunch of options is dumped into the `mocha` instance, like canonical / canonical camel-cased / alias / alias camel-cased versions of option names. This is prone to errors, but not part of this PR ==> see #4011.

When the user manually calls the `Mocha` constructor, there are some bugs to be fixed.
Furthermore we remove two deprecated options.

### Description of the Change

- remove deprecated `enableTimeouts` option
- remove deprecated `useColors` option
- remove `ignoreLeaks` option: this option does not exist and is redundant to `checkLeaks`
- fix `inlineDiffs` option: `useInlineDiffs` does not exist
- fix `diff` option: `hideDiff` does not exist
- fix `reporterOption` option: its alias `reporterOptions` was processed instead
- fix `fullTrace`: `fullStackTrace` does not exist
- cleanup `global` option: no alias processing

This is a breaking change.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -69,18 +69,18 @@ exports.Test = require('./test');
  * @param {boolean} [options.allowUncaught] - Propagate uncaught errors?
  * @param {boolean} [options.asyncOnly] - Force `done` callback or promise?
  * @param {boolean} [options.bail] - Bail after first test failure?
- * @param {boolean} [options.checkLeaks] - If true, check leaks.
+ * @param {boolean} [options.checkLeaks] - Check for global variable leaks?
+ * @param {boolean} [options.color] - Color TTY output from reporter?
  * @param {boolean} [options.delay] - Delay root suite execution?
- * @param {boolean} [options.enableTimeouts] - Enable timeouts?
+ * @param {boolean} [options.diff] - Show diff on failure?
  * @param {string} [options.fgrep] - Test filter given string.
  * @param {boolean} [options.forbidOnly] - Tests marked `only` fail the suite?
  * @param {boolean} [options.forbidPending] - Pending tests fail the suite?
- * @param {boolean} [options.fullStackTrace] - Full stacktrace upon failure?
+ * @param {boolean} [options.fullTrace] - Full stacktrace upon failure?
  * @param {string[]} [options.global] - Variables expected in global scope.
  * @param {RegExp|string} [options.grep] - Test filter given regular expression.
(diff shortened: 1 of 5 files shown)


juergba:
I will merge this PR within the next few days.

lib/mocha.js
```diff
@@ -69,18 +69,18 @@ exports.Test = require('./test');
...
  * @param {string} [options.fgrep] - Test filter given string.
  * @param {boolean} [options.forbidOnly] - Tests marked `only` fail the suite?
  * @param {boolean} [options.forbidPending] - Pending tests fail the suite?
- * @param {boolean} [options.fullStackTrace] - Full stacktrace upon failure?
+ * @param {boolean} [options.fullTrace] - Full stacktrace upon failure?
  * @param {string[]} [options.global] - Variables expected in global scope.
  * @param {RegExp|string} [options.grep] - Test filter given regular expression.
  * @param {boolean} [options.growl] - Enable desktop notifications?
- * @param {boolean} [options.hideDiff] - Suppress diffs from failures?
- * @param {boolean} [options.ignoreLeaks] - Ignore global leaks?
```
craigtaub: Do you need to remove its mention on `https://github.com/mochajs/mocha/blob/master/docs/index.md#browser-configuration` ?

lib/mocha.js
```diff
@@ -99,35 +97,14 @@ function Mocha(options) {
...
-
-  if ('enableTimeouts' in options) {
-    utils.deprecate(
-      'enableTimeouts is DEPRECATED and will be removed from a future version of Mocha. Instead, use "timeout: false" to disable timeouts.'
-    );
-    if (options.enableTimeouts === false) {
-      this.timeout(0);
-    }
-  }
+    .globals(options.global);
```
craigtaub: If we have removed the alias do you need to update it on https://github.com/mochajs/mocha/blob/master/docs/index.md#--global-variable-name ?

lib/mocha.js
juergba: done

lib/mocha.js
juergba: No, the user can still use the alias `globals` **in the CLI**, the parser will understand the alias name and transform it to its canonical name `global`. The Mocha contructor processes only canonical names (or its camel-cased version), in order to avoid having several variables with the identical information which could loose synchronisation.

PR #4019, opened 2019-09-16
uncaughtException after runner's end: improve exit code

### LINKED ISSUE #3917: Crashes that happen after test completion are reported with success exit code
### Description
Sort of follow up to https://github.com/mochajs/mocha/issues/3226 which was [fixed](https://github.com/mochajs/mocha/pull/3257) but it seems not fully.

### Steps to Reproduce

1. Install locall mocha v6.1.4

2. Create`test.js` test file:

```javascript
describe('test', () => {
  it('test', () => {
    setTimeout(() => {
      throw new Error('Unexpected crash');
    }, 10);
  });
});
```

2. Run `npx mocha test.js`

Output is valid:

```sh
$ npx mocha test.js

  test
    ✓ test


  1 passing (5ms)

/Users/medikoo/npm-packages/serverless/test.js:6
      throw new Error('Unexpected crash');
      ^

Error: Unexpected crash
    at Timeout._onTimeout (/Users/medikoo/npm-packages/serverless/test.js:6:13)
    at listOnTimeout (internal/timers.js:531:17)
    at processTimers (internal/timers.js:475:7)
```

But exit code is not:

```sh
$ echo $?
0
```

This makes such crashes not detected as fails in CI

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: **v6.1.4**
- The output of `node --version`: **v12.2.0**
- Your operating system:
  - name and version: **macOS-10.14.4**
  - architecture (32 or 64-bit): **64-bit**

### LINKED ISSUE #3938: Uncaught exceptions are silenced if happen in timeframe of skipped test
Possibly related to https://github.com/mochajs/mocha/issues/3740

### Steps to Reproduce

```javascript
describe('test', () => {
  it('test1', () => {
    setTimeout(() => {
      throw new Error('Uncaught');
    }, 3);
  });
  it('test2', function () {
    this.skip();
  });
  it('test3', function () {
    this.skip();
  });
  it('test4', function () {
    this.skip();
  });
});
```

**Expected behavior:**

Tests should fail unconditionally

**Actual behavior:**

In most cases suite passes (there is a race condition,due to which occasionally error gets exposed)

### Versions

mocha:  6.1.4
Node.js: 12.3.1
macOS: 10.14.5

juergba:
### Description

This PR forked various PR's concerning uncaught exceptions:
- related PR: #4025 #4030 #4033 #4083 #4150
- regression: #4147

### Description of the Change

- refactor and move existing tests
- add one additional test

### Applicable issues

closes #3917


juergba:
@medikoo I would highly appreciate your feedback to this draft PR, thank you!

lib/cli/run-helpers.js
```diff
@@ -25,7 +25,10 @@ exports.watchRun = watchRun;
  */
 const exitMochaLater = code => {
   process.on('exit', () => {
-    process.exitCode = Math.min(code, 255);
+    // when uncaught error, keep exitCode 1
+    if (!process.exitCode) {
```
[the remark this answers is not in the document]
juergba: We still support Node v8, so this should work for all versions.
I added a new `uncaughtException` handler  at runner's end which sets the `exitCode` to 1 and prints the error, similar to the default behavior of Node v10+. This handler can be removed when we drop Node v8.

juergba:
> Concerning #3938 it fixes issue but only when `--allowUncaught` is set.

TBH I don't really want to touch this recovery feature when `--allowUncaught: false`. I did some testing and the behavior is buggy and unpredictable. I will have a deeper look at this topic though.
For your needs is there anything opposing against using `--allowUncaught: true` and not using the recovery feature?

lib/runner.js
```diff
@@ -910,6 +917,11 @@ Runner.prototype.run = function(fn) {
   this.on(constants.EVENT_RUN_END, function() {
     debug(constants.EVENT_RUN_END);
     process.removeListener('uncaughtException', uncaught);
+    process.on('uncaughtException', function(err) {
+      // TODO: remove this listener when Node v8 support is dropped
+      process.exitCode = 1;
+      console.error('\n', err);
```
medikoo: I believe it should be `throw err`, as otherwise process will not be stopped

lib/cli/run-helpers.js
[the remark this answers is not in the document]
medikoo: Ok, looks good, I just added one remark

lib/runner.js
juergba: You are right about the process. With `throw err` the process stops as expected, but with `exitCode: 7`. I haven't succeeded in overwriting this code, even `process.exitCode` doesn't help.
I decided therefore to leave with `process.exit(1)` to copy Node's default behavior as close as possible.

juergba:
I have to investigate further about the dependencies of `--allow-uncaught` and `--exit`. While combining both options the outcome of my tests is suboptimal.

lib/runner.js
juergba: Node calls my exit strategy :

> a misuse of the process.exit() method that could lead to data printed to stdout being truncated and lost:
[...]
The reason this is problematic is because writes to process.stdout in Node.js are sometimes asynchronous and may occur over multiple ticks of the Node.js event loop. Calling process.exit(), however, forces the process to exit before those additional writes to stdout can be performed.

Maybe I should really use `throw err`...

medikoo: Interesting, but I'm not sure if `throw err` makes real difference.

Also it's not clear in what cases it'll be an async (how to reproduce the issue?) I've just tried following on macOS

```javascript
process.stdout.write("foo".repeat(2 ** 27) + "\nEND\n");
process.exit(1);
```

And always I get all output written, maybe it depends on a system.

juergba: It depends on [here](https://nodejs.org/api/process.html#process_process_stdout).
We have several output truncation issues in Mocha.

medikoo: It says that pipes are async on POSIX, not sure how reliable is below example, but i truncates same way with `process.exit` and `throw`

```javascript
const { Duplex } = require("stream");

class MyDuplex extends Duplex {
    _write(chunk, encoding, callback) {
        this.written = true;
        this.push(chunk);
        callback();
    }

    _read(size) {
        if (this.written) this.push(null);
    }
}

const interimStream = new MyDuplex();
interimStream.pipe(process.stdout);
interimStream.write("foo".repeat(2 ** 17) + "\nEND\n");
// Uncomment to prevent any output
// process.exit(1);
```

juergba:
The more I'm testing this, the more weird behavior pops up.
```js
describe('test', () => {
  it('test1', function() { });
  it('test2', function () {
    this.skip();
  });  
  it('test3', function () {
    this.skip();
  });  
  it('test4', function () {
    this.skip();
  });
});
```
Running with `node bin\mocha --allow-uncaught test.js`:
```
test
    √ test1
```
There is no (uncaught) exception, but all `this.skip()` tests plus the test summary are truncated.

medikoo:
> Node's docu states

In my understanding, it's just to indicate that we cannot be sure of side effects/broken states that come with uncaught exceptions, and therefore we should not try to recover from them. Otherwise if we know what we're doing (e.g. recovering from plain `throw new Error('Example')`) it can be safe.

Still, I totally I agree, that designing `this.skip()` to work through `uncaughtException` channel is very controversial.
I assume it was decided back in a past, when async tests were not necessarily written with promises, and were settled via `done` callback.
Then idea was that any `this.skip()` between start of a test, and `done` callback invocation, should be caught and recognized, so `uncaughtException` seemed the only way to cover that.

Maybe it'll be good with a new major, to narrow async support just to promises, and that way errors handling could be simplified.

When sync tests and async tests resolved via promises are concerned, I believe that any `uncaughtException` should be treated as an error that should be reported (even one thrown by `this.skip()`)) as it's a clear signal of orphaned async flow, which is an error on its own.

With how current implemention works, I would definitely not ignore uncaught exceptions that follows ones thrown via `this.skip()`. Fact that they're ignored now is source of #3938

juergba:
@medikoo thank you.

> Then idea was that any this.skip() between start of a test, and done callback invocation, should be caught and recognized, so uncaughtException seemed the only way to cover that.

```js
it('test', function (done) {
      var self = this;
      setTimeout(function () {
        instruction1;
        self.skip();       // includes implicit 'done()' and throws a 'Pending' error
        instruction2;     // does not run
        instruction3;     // does not run
      }, 100);
    });
```
The idea was to abort an asyncly running callback and not to execute `instruction2` and `instruction3`.

The user could easily get the same result by using `if skip/else` and we wouldn't need to throw this `Pending` error. We would just have to label the test as `pending`.

medikoo:
I think fact that `this.skip()` throws is desirable, questionable is how it's caught :)

But in async tests reported via `done`, I believe the `uncaughtException` channel is the only way to go.

juergba:
@medikoo thank you for your support!

PR #4024, opened 2019-09-19
Don't issue deprecation error for -compilers with --ignore-compilers

alexhisen:
### Description of the Change

Don't throw an error if deprecated --compilers option is provided if --ignore-compilers is also provided.

### Benefits

This is useful when a project has multiple test directories, some using older mocha with -compilers and others that don't need a compiler. WebStorm only supports a single global configuration in a project for mocha options, so with this --ignore-compilers option, the older mocha continues to use compilers and the new mocha doesn't throw an error.

### Possible Drawbacks

None

### Applicable issues

Patch release

### DIFF
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -279,7 +279,7 @@ exports.builder = yargs =>
         );
       }
 
-      if (argv.compilers) {
+      if (argv.compilers && !argv['ignore-compilers']) {
         throw createUnsupportedError(
           `--compilers is DEPRECATED and no longer supported.
           See https://git.io/vdcSr for migration information.`


juergba:
I don't like the idea to add additional options for each and every detail to Mocha's API. Furthermore this PR is incomplete, `--ignore-compilers` would have to be defined in our `yargs-parser` and `yargs` CLI implementation and there are no tests.

Can't you replace `--compilers` by `--require`, see [here](https://github.com/mochajs/mocha/wiki/compilers-deprecation)?

alexhisen:
The whole point is that we can't require babel in some directories but with a single run configuration, there is no way to do that in some and not in others. This solves the problem as long as you stick to an older version of mocha in the directory that does need babel. Someday WebStorm will allow for directory-specific configurations and then it won't be needed.

juergba:
I will close this PR within the next few days.

PR #4030, opened 2019-09-26
uncaughtException: fix "--allow-uncaught" with "this.skip()"

juergba:
### Description

```js
describe('test suite', () => {
  it('test1', function () { });
  it('test2', function () {
    this.skip();
  });
  it('test3', function () { });
  it('test4', function () {
    this.skip();
  });
});
```
`mocha --allow-uncaught test.js` silently aborts execution after the first `this.skip()`:
```
test suite
    √ test1
```
Our docu states:

> [...] It's also important to note that calling `this.skip()` will effectively abort the test.

We achieve this abortion of the test/hook execution by throwing a `Pending` exception. In async scenarios this results in an intended `uncaughtException` which has to be swallowed somehow. Obviously `--allow-uncaught` does not distinguish correctly between "bad" `uncaughtException` (user code) and "good" `uncaughtException` (`this.skip()`).

### Description of the Change

- independently of `--allow-uncaught`, we never allow the `Pending` exceptions to propagate.
- directly in `Runnable#skip()` we set the test to pending and immediately catch/swallow the `Pending` exception (sync scenario).
- for async `this.skip()` we register an addional `uncaughtException` handler for exceptions bubbling up after EVENT_RUN_END.
- the only purpose of the `Pending` is to abort the test execution, we don't use it for anything else.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -386,34 +386,30 @@ Runner.prototype.hook = function(name, fn) {
       if (testError) {
         self.fail(self.test, testError);
       }
-      if (err) {
-        if (err instanceof Pending) {
-          if (name === HOOK_TYPE_AFTER_ALL) {
-            utils.deprecate(
-              'Skipping a test within an "after all" hook is DEPRECATED and will throw an exception in a future version of Mocha. ' +
-                'Use a return statement or other means to abort hook execution.'
-            );
-          }
-          if (name === HOOK_TYPE_BEFORE_EACH || name === HOOK_TYPE_AFTER_EACH) {
-            if (self.test) {
-              self.test.pending = true;
-            }
-          } else {
-            suite.tests.forEach(function(test) {
-              test.pending = true;
-            });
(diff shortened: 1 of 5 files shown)


juergba:
I will do some additional testing and then merge this PR within the next few days.

lib/runner.js
```diff
@@ -910,6 +915,12 @@ Runner.prototype.run = function(fn) {
   this.on(constants.EVENT_RUN_END, function() {
     debug(constants.EVENT_RUN_END);
     process.removeListener('uncaughtException', uncaught);
+    process.on('uncaughtException', function(err) {
```
weswigham: This adding an `uncaughtException` handler every time the `end` event of a runner is called - this means for every finished runner, a new handler gets added. This, in turn, causes a `Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit` warning to be fired by node.

:(

Issue #2785, opened 2017-05-03
Errors in Async aren't caught correctly


jonstorer:
This feels very similar to #1128, however, that solution was helped along with promises. I can't find a way to solve this with promises. My specific example is when I'm testing an Express.js controller:

```javascript
// the Controller
let Promise = require('bluebird');

var controller = function (req, res, next) {
  let promise = new Promise(function(resolve) {
    setTimeout(resolve);
  });

  promise.then(function () {
    res.redirect('/otherPath');
  });
};

// the Test
let expect = require('chai').expect;

it('propagates errors correctly', function (done) {
  let res = {
    redirect: function(route) {
      expect(route).to.eql('/somePath');
      done();
    }
  };

  controller({}, res, function () {});
});
```

There is a promise involved which causes the mocha test runner to lose context, however, the promise happens in the controller, then resolves through a callback. 

Here it is in JSFiddle... https://jsfiddle.net/ccwapndu/ (ProTip: open the console)

ScottFreeCode:
If your promise is swallowing the exceptions because it does not use `catch` and does not pass the promise out/back to anything that could add `catch`, it's at risk for hiding errors outside of test cases, so you'll want to change the promise usage to fix that in any case -- other than that ultimately you'll need to report the resulting error/rejection to `done` there's not much about Mocha one way or the other, there...

...Except that Mocha could do a better job of reporting the problem, for which we have issue #2640. (Closing this as a duplicate of that, since that's the solution -- or the closest thing we can get to one anyway -- on Mocha's end.)

For your specific case, however: I'm fairly sure\* the third `next` parameter to Express handlers, besides being useful for middleware, is also useful for asynchronous error reporting... In fact, it functions pretty much identical to Mocha's `done` -- call it with no argument, it continues to the next handler, call it with an argument, it treats the argument as an error. (This is unfortunately, to my knowledge, only documented at the end of a seemingly unrelated section halfway down [the page on writing error handlers](http://expressjs.com/en/guide/error-handling.html), and not on the pages about writing regular handlers that would actually do the `next(<some error>)` calling. ...I meant to submit a PR to fix that months ago; I should double-check whether I had anything drafted for it...) So what you'd want is to change your controller to be like this:
```js
var controller = function (req, res, next) {
[...]

Issue #2640, opened 2016-12-17
unhandledRejection


stevenvachon:
These should be handled automatically, as to reduce boilerplate. A common issue as a result of this is #1128.

```js
process.on('unhandledRejection', (reason, promise) => throw promise);
```

For web browsers, we could use `window.addEventListener`.

ScottFreeCode:
This is definitely something we should add, since the current timeout behavior is quite uninformative and generally promise implementations support this hook; but let's make sure the resulting test failure message indicates that these errors would have been swallowed by the promise in production and should be somehow handled or communicated to the caller for handling (at which point it's easy for the test to communicate the failure back to Mocha without relying on unhandled rejection functions) -- a lot of the questions we get about the current behavior come from lack of awareness of the need for `catch`.

ScottFreeCode:
One clarification to my earlier comment: I was thinking specifically of unhandled promise rejections -- I believe Mocha already catches uncaught non-promise-swallowed exceptions?

stevenvachon:
Yes, I think that's true.

rogierschouten:
+1  We were completely caught by surprise that mocha catches thrown errors but not unhandled rejections. Our tests passed when actually they should have failed.

boneskull:
This is what I think needs to happen:

1.  An unhandled rejection needs to result in the same (or similar) behavior as when an uncaught exception is thrown
2.  A new flag, `--allow-unhandled`, should be added (if we piggybacked on `--allow-uncaught`,  it would be more difficult to pin down either one in particular)
3.  Plenty of tests.

Since Mocha doesn't use `Promise`s internally, this *shouldn't* be a particular severe change to the codebase.  I *think* we can get away with basically copy-and-pasting the majority of the "uncaught exception" behavior, as well as the test suites.

There are likely some edge cases here I'm not considering, so this is best-case scenario.  😇

lrowe:
Not sure if this is a separate issue or not, but I see similar behaviour without promises when deferring an error with setImmediate or process.nextTick.

```javascript
// swallow.js
it("swallows deferred error", function() {
  setImmediate(() => { throw new Error("thrown error"); });
});
```
Running the test I get success with exit code 0.
```
$ node bin/mocha swallow.js


  ✓ swallows deferred error

  1 passing (5ms)

```
Whereas running the same code with node v8.9.0 gives prints the stack trace and exits non-zero:
```
$ node -e 'setImmediate(() => { throw new Error("thrown error"); });'
[...]

PR #4032, opened 2019-09-26
Catch unhandled promise rejections during tests

Zirak:
### Description of the Change

Listen to the unhandledRejection event when running tests, catching promises who
got rejected without being cared for, and failing the test in which said
promises were left unhandled.

### Alternate Designs

AFAIK, this is the cleanest way to gracefully capture unhandled rejections.

This changes how tests are run in mocha, and provides a nice parallel to the
existing `allowUncaught` option.

### Benefits

Async tests which now trigger, by themselves or down their call tree,
promises left unresolved now silently pass, with a scary stack trace printed in
the console. For existing issues, see #2640 and #2785 (among others).

### Possible Drawbacks

This is only part of handling unhandled rejections, some things not yet included
are:

- Handling unhandled rejections outside spec tests (e.g. testhooks)
- Making it work in a browser
- Amend error messages which reference unhandled rejections
- Testing the new signature of the `Runner.prototype.uncaught` method
- A sister flag to `allow-uncaught`

If this commit looks interesting enough, a future commit will amend these shortcomings.

### Applicable issues

This may be a breaking change. Tests passing with unhandled rejections will fail starting with this change.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -783,16 +783,19 @@ Runner.prototype.runSuite = function(suite, fn) {
  * Handle uncaught exceptions.
  *
  * @param {Error} err
+ * @param {string} unwhat It this an 'uncaught' exception, or 'unhandled' rejection
  * @private
  */
-Runner.prototype.uncaught = function(err) {
+Runner.prototype.uncaught = function(err, unwhat) {
+  unwhat = unwhat || 'uncaught';
+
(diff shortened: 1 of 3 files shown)


juergba:
@Zirak thanks.
In Mocha v8.2.0 we added an `unhandledRejection` listener.

PR #4038, opened 2019-10-01
Node esm support

giltayar:
### Description of the Change

This PR adds support for [Node.js's native ESM (ES modules) support](https://nodejs.org/api/esm.html).
Specifically, up till now, Mocha could only load test files that were CommonJS,
and this PR enables it to load test files that are ES modules,
and thus enable the test files to import stuff from other ES modules
(not to mention other goodies coming up the line, like enabling top-level await, which is an ESM only JS feature).

Note that this support is for Node.js only, and has no connection to browser ESM support.
Also, prior to Node v12, there was no or quirky support for ESM, so this support is only for Node v12 and above.

### Alternate Designs

This is not a trivial change. The main problem is that loading an ESM is an _async_ operation, and yet
the call path to `Mocha.prototype.loadFiles` is a sync one. Changing the call path would necessitate a
semver-major change, which I'm not sure I have the ability to push, so I decided on a design
that keeps the signatures of all the functions as they are, _unless there is an ESM test file to load_, by which
time they morph to async functions. So their return type in the signature looks
something like `Runnner|Promise<Runner>`, where `Runner` is for cases where there are no ESM test files,
and `Promise<Runner>` is for the case there are

_This is not a good solution_, but it enables us to push this change without it being semver-major, and once
there is a decision to change to semver-major, the signature change and code change that does this is trivial.

If y'all decide that this PR can change the major version of Mocha, I can fix it so that the signatures
change and the weirdness goes away.

Another not insignificant change was the error path. In some cases, errors found by Mocha were synchronous, i.e.
they occur during the `yargs.parse` phase, which throws the error. But, because the call path to `loadFiles` could
be asynchronous, I changed the command handler (in `lib/cli/run.js`) to be async (i.e to return a `Promise`).
This means that errors (such as bad tap versions) are no more synchronous, and will always arrive at the 
`fail` yargs handler (in `lib/cli.js`). To make them look like the old errors, I changed the code
there. No tests fail now, so it _seems_ OK, but please check this in the review.

I'm assuming this is obvious: it has to change `Mocha.prototype.loadFiles` to be async. There is the `esm` module that simulates ESM in userland, but native support has to be async. And native support is crucial if the developer's code is running using Node's native ESM support.

### Benefits

Developers starting to write Node.js code in ES modules will benefit by being able to write their tests using
ESM too, and not have to dynamically import their ES modules using dynamic import (aka `await import`), which is a hassle and a kludge.

### Possible Drawbacks

This _may_ impact code using Mocha's API, as we are changing the signature (even if not in the short run,
definitely in the long run). Also, error handling has subtly changed, and this may impact error handling reporting
in ways I cannot know.

### Applicable issues

This is currently an enhancement (minor release), that should be turned in the future into a semver-major,
as described in the "Alternate Designs" section.

### Limitations

* "Watch" mode does not currently work with ESM, as there is currently no way to remove an ESM module from Node.js' module cache. We would have to either wait for it to be implemented in Node (slim chance, IMHO), or refactor watch mode to use subprocesses every time a change is detected.
* Module-level mocks via libs like [proxyquire](https://npm.im/proxyquire) or [rewiremock](https://npm.im/rewiremock) or [rewire](https://npm.im/rewire) cannot work.  **If you are using these libs or similar, hold off on using ESM for your test files.**. This is not inherently a limitation in Mocha, but rather a limitation in Node's ESM support, which will probably be rectified sometime in Node v13.
* ESM is not supported for custom reporters, nor is it supported for custom interfaces.


giltayar:
I fixed the bug in Windows (it was actually a bug in the code, but it only manifested itself in Windows). 

Unfortunately, there is a bug in Windows support for ESM. The third ESM test I wrote fails in Windows in Node v12.6.0 (the version Appveyor runs), but passes in Node v12.11.1 (the latest Node version I downloaded today).

I'm not sure how to upgrade the Node version in Appveyor, so I can disable this test for Node versions < 12.11.1?

giltayar:
It seems the bug in Node ESM implementation is not just in Windows, but also in Linux, so I disabled ESM support in Mocha for Node <v12.11.0.

juergba:
> This is not a good solution, but it enables us to push this change without it being semver-major, and once there is a decision to change to semver-major, [...]

We should make this decision at first place. Why should we try to avoid `semver-major`?
The user - while using Mocha programmatically - would have to load the files manually, right?
```js
var mocha = new Mocha();
mocha.addFile();
mocha.loadFiles();
mocha.run();
```
Would the user be affected in any other breaking way by a `semver-major`?
There is even an issue proposing to turn `loadFiles()` into a public function, to ease live of plugin creators.

I strongly tend to vote for `semver-major` in order to avoid the changes `Mocha.prototype.run()`.

PS: Could you please edit your description above and add the `require.cache` / `watch` restrictions?

giltayar:
@juergba today `loadFiles` is internal and part of run. To understand what will change, let's look at your code (modified to remove `loadFiles`):

```
var mocha = new Mocha();
mocha.addFile();
mocha.run();
console.log('Done!)
```

If we broke the interface, it would _have_ to look something like this:
```
var mocha = new Mocha();
mocha.addFile();
mocha.run().then(() => console.log('done'));
```

Because `Mocha.run` will now return a `Promise<Runner>`, and not a `Runner`. That means that if this code is in a function, that function, instead of being sync, and returning whatever, it will have to be async and return `Promise<whatever>`. That's not a small change for everyone that uses us. It's pretty big.

But, because ESM is async, this is a change that has to be done. What I did was this: if you have tests using ESM in Node, you can still have your old sync code. But! If you have tests that are ESM, `Mocha.run` will be async and will return a promise.

I am fully aboard with breaking compatibility, and in the long run it had to be done, but I _did_ want to give y'all the option of having it both ways. I can easily patch it to break compatibility and be "correct".

giltayar:
In addition, if we expose `loadFiles`, which would have to be async if we break compatibility, 
the code would be

```js
mocha.addFile(...)
mocha.loadFiles().then(() => mocha.run()).then(() => console.log('Done')
```

juergba:
@giltayar thank you.
If we exposed `loadFiles`, there would be no need to change `Mocha#run`. It could remain sync, always returning a `Runner`, right?
```js
mocha.addFile(...)
mocha.loadFiles().then(() => mocha.run(() => console.log('Done')))
```

juergba:
```js
exports.singleRun = (mocha, {exit}, fileCollectParams) => {
  const files = collectFiles(fileCollectParams);
  debug('running tests with files', files);
  mocha.files = files;
  return mocha.run(exit ? exitMocha : exitMochaLater);
};
```
Couldn't we implement it here, even as `semver-minor`?
- `experimental-module: true`: we load our files asyncly outside of `mocha.run()`, mark the files somehow as loaded, then call `mocha.run()`. Files are already loaded, so we don't load them again inside `mocha.run()`.
- `experimental-module: false`: files are loaded synchly inside `mocha.run()`, as current state.

`mocha.run` would always return a `Runner`, no `Promise`.
`watch-run`: does not work with ems anyway (`require.cache`), so no changes here.

giltayar:
@juergba yes, if we take out `loadFiles`, then `run` could be sync. But taking out `loadFiles` is anyway a breaking change.

Any in any case, `singleRun` couldn't be sync, as it has to have `loadFiles` in it. And that changes the interface anyway, right? So we still can't avoid it.

As for your idea of `loadFiles` being called outside, and we check whether it was already called or not, and if it was called, OK, and if not, load synchronously: it's still an awkward interface. Not clean and full of ifs and buts.

I say this: the correct solution is to make `loadFiles` and `singleRun` async, and that's a breaking change, whether or not `loadFiles` is inside `run` or not. So either go and break compatibility and be semver-major in a clean way, or we implement a weird solution, which could either be my solution, or yours. And I prefer my solution not because it's better (it's not!), but because people who are not native-node-ESM aware need not care about it at all. And, today, frankly, that's 99.9% of the people.

juergba:
I prefer the clean way, `semver-major` yes/no can't really be a main criteria.
`singleRun` is a private function, turning it into async would not be a breaking change (?).

@demurgos @maxnordlund I need some help with this one and would highly appreciate your input.

giltayar:
@juergba I will submit to the decision of my elders. 😀

Once you decide, I'll do the switch. It'll probably take no more than an hour or two.

Note that there are two decisions to make:

1. Do we enable `singleRun` and others be async and thus breaking compat?
2. If we do, do we separate `Mocha.prototype.loadFiles` from `Mocha.prototype.run` and have only it be async?

boneskull:
I know I'm probably blocking this, but I expect I can get to review it in about 10 days.

David-Else:
THANK YOU @giltayar  !! This is most definitely not just a 'nice to have' feature... can't believe that was added as a tag.

@boneskull Any update? Cheers!

boneskull:
(back when the issue was created, we added the tag because it wasn’t clear when the feature would leave “experimental” status; so nice-to-have before that. now, it’s needed, obviously)

I’ll try to review this asap

boneskull:
@giltayar IMO, zalgo is bad.

A third option would be to leave `Mocha#run` as-is and create a new method (e.g., `runAsync`) to implement the new behavior.  We can call that function from the CLI, and we can continue to support `Mocha#run` until it's clear we don't want to support it any longer.  I'd prefer not to break 3rd-party code relying on it (at least for the immediate future), even if we no longer use the API internally.

`runAsync`, then, would _always_ return a `Promise`.  

So... Promises. Because Mocha doesn't use _any_ Promises internally (prior to this PR), we'll need to ensure Mocha's behavior on unhandled rejection is sane. Mocha has well-defined behavior around how it traps and handles uncaught exceptions.  Rejection handling doesn't need to be _the same_ as this, because it might not make sense, but if it's different, that behavior should be _defined and documented_.

If you're interested in this change, feel free to implement, but I can also run with this PR and get it over the finish line if you don't have time.  Also what does @juergba think of this "new method" idea?

boneskull:
Regarding "major" or "minor", I don't really care either way, but I still would rather not introduce a breaking API change lacking any clear necessity.  If we can side-step the existing API for programmatic users, then IMO that's the best solution.

giltayar:
@boneskull - I like the idea of leaving `#run` alone, for backward compatibility with code that uses it (and that cannot use ESM), and implementing a new (clean) `#runAsync` function that is Promise-driven and supports ESM, and which is called from the CLI. This also resolves the major/minor conundrum: we can leave it minor, as we're not breaking backward compatibily.

Definitely better than my version. I can do it tomorrow (it's midnight now here in Israel), but feel free to implement it yourself if you feel like it. I'm OK.

Let's do it! Let's make Mocha the first testing tool to support native Node ESM!

giltayar:
@boneskull I didn't think about unhandled rejections. Good point.

Handling them like unhandled exceptions (I'm guessing Mocha exits with an error?) is problematic, as I've seen code in the wild that ignores unhandled rejections (bad idea, but it happens because sometimes it's basically impossible to catch unhandled rejections in third party libraries), and which wouldn't want Mocha to fail the test because of it.

This is currently a moot issue anyway *for now*, as I'm guessing Mocha handles unhandled rejections when _running_ the test, but you're talking about unhandled rejections when _loading_ the test files? In this case, and those won't happen till Top-Level Await happens and introduces _real_ asynchrony when loading the test files.

boneskull:
> This is currently a moot issue anyway for now, as I'm guessing Mocha handles unhandled rejections when running the test, but you're talking about unhandled rejections when loading the test files?

That would be an incorrect guess.  😝   There's no global `unhandledRejection` listener added by Mocha.  Yeah, I suppose we can punt on this for now, but will have to deal with it soon with top-level await.  Pretty much not happy with how Node.js deals w/ unhandled rejections, but at least there's [--unhandled-rejections=mode](https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode) now.

boneskull:
> Let's do it! Let's make Mocha the first testing tool to support native Node ESM!

I'm pretty sure tap or tape beat us to it. But thanks for picking it up again, and sorry for the delay--I've been too busy with other stuff the past 6mo or so.

Also:

1. [ ] We need to add v13 to the test matrix.
1. [ ] We need an integration test to run _unflagged_ in v13.2.0

juergba:
> Also what does @juergba think of this "new method" idea?

It's a good idea.
We also have a `--require` option, which should be able to load ESM modules, I guess.

> I can do it tomorrow (it's midnight now here in Israel), but feel free to implement it yourself if you feel like it. I'm OK.

You should find an agreement on this, who is going to do what. I think it would be fair to leave the implementation to @giltayar, if he wants so.

giltayar:
Just started working on it. Shouldn't be too difficult—the code is there, I just need to massage it a little.

giltayar:
I'm assuming I should also have a parallel `Mocha#runAsync` that supports ESM? I'm assuming I do.

giltayar:
@juergba - I have a problem with `--require` because it is called in the `yarg` `.check` callback, which is sync, and cannot be turned to async. It also says this:

`// load requires first, because it can impact "plugin" validation` (`run.js`)

giltayar:
There is another problem with watch mode: watch mode unloads the test files before rerunning. This is impossible to do (_currently_) with ES modules (they don't share the same cache, and the cache is currently private to Node).

Because module keys are _URLS_ and not file paths, we can use cache buster so that each time we load a different file. But that means that watch mode will currently load and load and load the same files, without unloading the previous ones. VERY easy to implement, but I am not sure this is a good idea.

juergba:
> @juergba - I have a problem with `--require` because it is called in the `yarg.check` callback, which is sync, and cannot be turned to async.

Maybe `yarg.check` is the wrong place for our `--require` handling. Yargs has a `middleware()` method which possibly could help. We could also split this topic into a separate PR, if it gets to complex.

re watch mode: do we have a new problem, or the one we have already known for weeks? I thought we had agreed on not supporting `watch` with `esm`?

PS: I will send my feedback bit by bit and don't expect immediate action. I don't feel comfortable with this topic, yet. Anyway IMO we should not publish this PR with `v7.0.0`, but with next `minor` `v7.1.0`.

docs/index.md
```diff
@@ -1499,7 +1500,7 @@ To specify custom report title, use `--reporter-options suiteName="Custom name"`
 ### Third-Party Reporters
 
 Mocha allows you to define custom reporters. For more information see the [wiki][mocha-wiki-more-reporters].
-
+.equal
```
juergba: a typo?

docs/index.md
```diff
@@ -1509,6 +1510,31 @@ Examples:
...
+this means either ending the file with a `.mjs` extension, or, if you want to use the regular `.js` extension, by
+adding `"type": "module"` to your `package.json`.
+More information can be found in the [Node.js documentation](https://nodejs.org/api/esm.html).
+
+> Mocha supports ES modules only from Node.js v12.11.0 and above. Also note that
+> to enable this, you need to add `--experimental-modules` when running
+> Mocha. Last, but not least, the ESM implementation Mocha is not yet
+> stable so given that the specification in Node.js may
+> change, this may necessitate changes in Mocha's integration.
+
```
juergba: In Node >=v13 does it work without that flag?

lib/mocha.js
```diff
@@ -835,22 +875,29 @@ Object.defineProperty(Mocha.prototype, 'version', {
  * already in the `require` cache), make sure to clear them from
  * the cache first!
  *
+ * This method supports both ES Modules and CommonJS ones
```
juergba: no, only CJS?

lib/mocha.js
```diff
@@ -835,22 +875,29 @@ Object.defineProperty(Mocha.prototype, 'version', {
...
+ *
  * @public
  * @see {@link Mocha#unloadFiles}
  * @see {@link Runner#run}
  * @param {DoneCB} [fn] - Callback invoked when test execution completed.
- * @return {Runner} runner instance
+ * @return {Runner|Promise<Runner>} runner instance
```
juergba: only Runner

docs/index.md
giltayar: Yup. Probably me typing in the incorrect place. Will fix.

docs/index.md
giltayar: Node >= v13.2.0. And you're right, I should change this text to reflect the stability.

lib/mocha.js
giltayar: Oops, you're right. Will fix.

lib/mocha.js
giltayar: Yup. Will fix.

lib/esm-utils.js
```diff
@@ -0,0 +1,27 @@
...
+// and Node supports async/await in all its non-dead version.
+
+const url = require('url');
+const path = require('path');
+
+exports.requireOrImport = file => {
+  try {
+    file = path.resolve(file);
+    return require(file);
+  } catch (err) {
```
juergba: Is there no other way to find out, extension `*mjs` or `type: module`.
Or some internal Node property?

package.json
```diff
@@ -554,6 +553,7 @@
     "acorn": "^7.0.0",
     "assetgraph-builder": "^6.10.1",
     "autoprefixer": "^9.6.0",
+    "babel-eslint": "^10.0.3",
```
juergba: Why do we need this dependency?

boneskull:
Because the global `mocha` is essentially monkeypatched in the browser (see `browser-entry.js`) we can remove `runAsync` (maybe it's just `delete mocha.runAsync`) from its interface. We can try to make it work later

package.json
boneskull: I think this can be replaced with something in the `ecmaFeatures` config.  still, that should not be allowed except in our sources that use `import`

boneskull: otoh, it may be that eslint doesn't support `import()`?

lib/esm-utils.js
boneskull: yes, we should be able to short-circuit this a bit:

1. Use `require` with `.cjs` files
1. Use `import()` with `.mjs` files

Other than that, I'm not seeing that the package scope resolution method is exposed anywhere (I'll need to look around in the node source).  So if we wanted to consume `type` in `package.json`, we'd need to use [read-pkg-up](https://github.com/sindresorhus/read-pkg-up) and manage our own cache, which seems gross.

Also, the cache-clearing method that `Mocha` exposes won't work with ESM...

boneskull: OK, `import()` will handle _both_ ESM and CJS, so we don't need to try/catch _at all_ here.

boneskull: Cache-clearing using ESM is an unanswered question atm, and it's best to not worry our pretty heads about it.

boneskull: this means we can't support "watch" mode until that's solved in Node core.  there is no cache available to clear.  right now, using ESM for tests has two major drawbacks due to this missing feature:

1. You can't hot-reload a module for "watch" mode
2. You can't hot-reload a module to do module-level mocking (like with [rewiremock](https://npm.im/rewiremock))

giltayar: @juergba: no, there is currently no way to determine if a file is ESM or CJS, although Node's Module Working Group is deliberating on that (https://github.com/nodejs/node/issues/30514)

@boneskull, while it is true that `import()` should handle CJS, remember that _all_ the importing/requiring in Node will go through this function. I was hesitant in making `import()` the way we import all test files in Mocha, as who knows what hidden bugs and weird behavior lurks there.

So I opted for this weird `require`/`import` style to solve both problems: not knowing what type the file is, and fear of unknown behavior when `import`-ing CJS. I think it's a good way to start this, given that practically 100% of test files out there are CJS. We can revisit this decision in a couple of months once people really start using this feature. Till then, I believe using `require` optimistically and falling back on `import` is the better decision.

giltayar: @boneskull - yup, watch mode cannot yet be implemented for ESM, unless we use a cache-busting mechanism (ESM paths are URLs, so we can add arbirtrary query parameters to them) to it.

I have pinged the modules working group on this problem (https://github.com/nodejs/modules/issues/307#issuecomment-558462605).

package.json
giltayar: @boneskull - yup, eslint doesn't support `import()`. This is what happens when I remove `babel-eslint` from `.eslintrc.yml`:
```
> nps test
...
nps is executing `lint.code` : eslint . "bin/*"

/Users/giltayar/code/mocha/lib/esm-utils.js
  14:14  error  Parsing error: Unexpected token import
```

.eslintrc.yml
```diff
@@ -32,7 +32,12 @@ overrides:
       ecmaVersion: 6
```
juergba: Can't we update this (ES6 === ECMAScript 2015) to ECMAScript 2018 as well?
And use `async`/`await` especially in our "lib/cli/**/*.js" files?

lib/cli/run.js
```diff
@@ -299,8 +299,8 @@ exports.builder = yargs =>
     .number(types.number)
     .alias(aliases);
 
-exports.handler = argv => {
+exports.handler = function(argv) {
   debug('post-yargs config', argv);
   const mocha = new Mocha(argv);
-  runMocha(mocha, argv);
+  return Promise.resolve(runMocha(mocha, argv));
```
juergba: I have never liked to get an Yargs error, when some exception in eg the runner was thrown. IMO Yargs should not handle those type of exceptions. Yargs should parse and handle exceptions thrown during parsing, nothing more.

We could activate ECMA2018, turn this line into `async`/`await` and add `try`/`catch` . The error path would be shorter.

lib/mocha.js
```diff
@@ -311,6 +314,43 @@ Mocha.prototype.loadFiles = function(fn) {
...
+ * @returns {Promise}
+ */
+Mocha.prototype.loadFilesAsync = function() {
+  var self = this;
+  var suite = this.suite;
+
+  if (!esmUtils) {
+    return new Promise(function(resolve) {
+      self.loadFiles(resolve);
+    });
```
juergba: BTW why does our eslinter - set to ES5 - not complain about that Promise?

lib/esm-utils.js
juergba: Ok, I could live with this (ugly) solution to start with, since Node's v13.2.0 docu still labels the esm module with "Stability: 1 - Experimental". But then please add  a short comment in order to remind us there is still some work to do.

.eslintrc.yml
giltayar: I would prefer that, but I didn't want to change anything that major. Which is why we get my shenanigans around that.

If you say yes, I can make the change. It will probably also mean transpiling async/await in the babel config.

lib/cli/run.js
giltayar: I couldn't make it run in any other way. I really tried, but I believe I failed because it is actually `yargs` that is the top-level code that runs everything, so everything goes through it. But I have to admit that the code confused me, and maybe a second look now that I'm more familiar with the code.

Any idea on what I _should_ do?

lib/mocha.js
giltayar: No idea. The ways of eslint are mysterious. :-)

giltayar:
Regarding supporting "watch" mode: the Modules working group are reluctant to add an "unload ESM" function to ESM support in Node. Understandably so, given the complexity of the edge cases: https://github.com/nodejs/modules/issues/307

What they suggested instead, is going the "subprocess" route, whereby in Watch mode, we run the tests in a subprocess, and whenever the thing changes, we rerun the subprocess, with a new cache and no problems. 

This seems like a cleaner solution than unloading, and I believe you're doing something like that in the CLI today, no?

juergba:
re "watch" mode: no, we are not spawning new child-processes, we manually delete `require.cache`. The same with our new implementation with chokidar, going to be published with v7.0.0. Nodemon uses child-processes, I think, but its said to be slow.
We only spawn a child-process when the user passes Node options via Mocha.
I propose to forget about watch mode for now. But you should add this caveat (no watch with esm) to our docu.

giltayar:
@juergba - will add the documentation.

juergba:
We added Node v13 to our CI test matrix, please rebase.

.eslintrc.yml
juergba: Our main problem is IE11 (to be dropped in 2025 😁), that's why we have to use ES5.
Our "lib/cli/**/*.js" files have been overwritten to ES6 for months. This means:
- our browser part does not use those files (seems very plausible)
- since we dropped Node v6, we can update from ES6 to ES2018 (ES5 remains default)
- we should exclude those "lib/cli/**/*.js" files while browserify'ing

So from my side you get a go. Let's wait for some days, if there is no objection, then my will be done.

lib/esm-utils.js
```diff
@@ -0,0 +1,27 @@
+// This file is allowed to use async/await because it is not exposed to browsers (see the `eslintrc`),
```
juergba: This file should be set into our "bin/cli" folder?

package-scripts.js
```diff
@@ -23,7 +23,7 @@ function test(testName, mochaParams) {
 module.exports = {
   scripts: {
     build: {
-      script: `browserify -e browser-entry.js --plugin ./scripts/dedefine --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' --ignore chokidar -o mocha.js`,
+      script: `browserify -e browser-entry.js --plugin ./scripts/dedefine --ignore "./lib/esm-utils.js" --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' --ignore chokidar -o mocha.js`,
```
juergba: I propose to remove all `--ignore` options and add them in our "karma.config.js" file. `chokidar` is missing in latter, but in general these options are redundant.
And IMO some files as "bin/cli/*"  are missing completely.

lib/esm-utils.js
giltayar: If we decide to move our lib/cli folder to ES2017/8, then yes, that's a good idea, although it's a bit weird having it there because it has nothing to do with CLI.

package-scripts.js
juergba: We also have a `browser` field in "package.json", plus the script above and "karma.config.js". I lost orientation, so lets wait with this.

lib/cli/cli.js
```diff
@@ -45,10 +44,13 @@ exports.main = (argv = process.argv.slice(2)) => {
...
-    .fail((msg, err, yargs) => {
+    .fail((msg, err) => {
       debug(err);
-      yargs.showHelp();
-      console.error(`\n${symbols.error} ${ansi.red('ERROR:')} ${msg}`);
+      if (err) {
+        console.error(`\n${err.stack || err}`);
+      } else {
+        console.error(`\nError: ${msg}`);
+      }
```
juergba: By this change the error formatting is lost,
eg. see "test/integration/fixtures/options/forbid-only/only-suite.fixture.js".
IMO it's better to leave this code untouched and to handle errors earlier in the yargs handler, also to avoid `UnhandledPromiseRejectionWarning`.

lib/cli/run.js
juergba: You could catch the error here:
```js
return Promise.resolve(runMocha(mocha, argv)).catch(function(err) {
  console.error(`\n${symbols.error} ${ansi.red('ERROR:')} ${err.message}`);
  process.exit(1);
});
```
There are two tests failing though, which probably could get fixed.
That way we would not have to touch the yargs part.

lib/esm-utils.js
juergba: Thinking about it I vote for:
- short-circuit: require() for .cjs / import() for .mjs
- then your "weird require/import style" since we still have experimental status.

juergba:
Besides loading of test files and the already mentioned `--require` option, we additionally have `--ui` and `--reporter` options which use `require()`. We should take a decision wether we handle these options now or in a separate PR.

boneskull:
that can be a separate PR

.eslintrc.yml
juergba: updated from ES6 to ES2017 in separate PR.

test/integration/esm.spec.js
```diff
@@ -0,0 +1,54 @@
...
+var run = require('./helpers').runMochaJSON;
+var utils = require('../../lib/utils');
+
+if (!utils.supportsEsModules()) return;
+
+describe('esm', function() {
+  it('should pass a passing esm test that uses esm', function(done) {
+    run(
+      'esm/esm-success.fixture.mjs',
+      ['--experimental-modules', '--no-warnings'],
```
juergba: should be an empty array when Node >=13.2.0

lib/cli/cli.js
```diff
@@ -12,8 +12,7 @@
  */
 
 const debug = require('debug')('mocha:cli:cli');
-const symbols = require('log-symbols');
-const yargs = require('yargs/yargs');
+const yargs = require('yargs');
```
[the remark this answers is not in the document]
boneskull: yes, please revert the change; we want `yargs/yargs`

package.json
boneskull: Was this resolved with #4115?

lib/esm-utils.js
boneskull: > So I opted for this weird require/import style to solve both problems: not knowing what type the file is, and fear of unknown behavior when import-ing CJS. I think it's a good way to start this, given that practically 100% of test files out there are CJS. We can revisit this decision in a couple of months once people really start using this feature. Till then, I believe using require optimistically and falling back on import is the better decision.

I think this is fair until we've seen `import()` in wider use.

boneskull:
@juergba and anyone else interested: discussion in https://github.com/nodejs/modules/issues/307

Alhadis:
You should probably add `.mjs` and `.cjs` to the [list of supported config extensions](https://github.com/mochajs/mocha/blob/03b58f25afff19f35c4179a6f7e9adb02671d90a/lib/cli/config.js#L23-L29), too. Should a codebase have [`"type": "module"`](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_code_package_json_code_code_type_code_field) in its `package.json`, Node v13.2+ will interpret `.mocharc.js` as an ESM module instead of CommonJS.

One can always modify their `.mocharc.js` file to use `export default` instead of `module.exports =`, but obviously that'll cause problems when running tests against older versions of Node.

giltayar:
@boneskull - I'll work on adding your notes to the documentation. I believe this necessitates a separate "ESM" section in the documentation? I'll assume yes. 

@Alhadis - good catch! I'll make the change in this PR.

package.json
giltayar: Unfortunately, no. `import()` is ES2018, and #4115 updated to ES2017, due to the need to support Node v8.

lib/cli/run.js
giltayar: I'll try to do this again.

giltayar: Done! It seems the problems stemmed from the weird sync/async trick I did, and once we go full async, there's no problem anymore and all error handling can be done from `run.js`.

lib/esm-utils.js
giltayar: I decided not to move it to `lib/cli` because it has nothing to do with `cli`, and is needed from `mocha.js` (conditionally), so it doesn't make sense to have a dependency _from_ `mocha.js` to `cli`.

Of course, if you disagree, I will promptly move it there.

lib/cli/cli.js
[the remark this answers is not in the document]
giltayar: Wow, didn't know that! Thanks.

Done.

lib/cli/run-helpers.js
```diff
@@ -98,21 +98,29 @@ exports.handleRequires = (requires = []) => {
  * @param {boolean} [opts.exit] - Whether or not to force-exit after tests are complete
  * @param {Object} fileCollectParams - Parameters that control test
  *   file collection. See `lib/cli/collect-files.js`.
- * @returns {Runner}
+ * @returns {Promise<Runner>}
```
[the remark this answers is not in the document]
giltayar: Yes. Note a few lines above it the call to `.then(...)`

lib/cli/cli.js
giltayar: I did it, and reverted it back to what it was! As I said above, the weird running "sync & async into "always async" forced me to do it this way, and moving to "always async" fixed it and allowed me to revert to the correct way of doing this.

test/integration/esm.spec.js
giltayar: True! Done.

giltayar:
I have implemented all the above review comments, except for supporting `mocharc.mjs` (the PR _does_ enable supporting `mocharc.cjs`) - @Alhadis, FYI.

The reason I excluded `mocharc.mjs` support is because that would have made `loadConfig` an async function, which would have made more of the code async. I will add a separate commit for that, and you can choose whether to accept this or decide to move it to a separate PR.

giltayar:
It turns out that adding ESM `.mocharc.mjs` to Mocha was not trivial! I separated it into a separate commit (the last one) for you to go over and decide whether you want to include it or not.

giltayar:
It turns out that adding ESM .mocharc.mjs to Mocha was not trivial! I separated it into a separate commit (the last one) for you to go over and decide whether you want to include it or not.

juergba:
I don't think this is good idea to squash `.mocharc.mjs` into this PR. We already decided to split
`--require`/`--ui`/`--reporter` into a separate PR. So I don't understand why you want to do this.
I guess it would also turn this PR into `major` again. The option parsing is a public API.

giltayar:
@juergba I actually looked and ensured that any of the functions I touched were _not_ in the public API, but I may have missed it...?

And if it is semver-major, better to combine it with the other semver-major changes this PR is going to do?

But I can take them out and put them in another PR once this is merged. But if the changes _are_ semver-major, I would definitely add them to this PR.

juergba:
`loadOptions` is a public function.

What are the reasons you want this in?

> And if it is semver-major, better to combine it with the other semver-major changes this PR is going to do?

I thought this PR is `minor`?

giltayar:
@juergba - actually, you're right. We're _adding_ to the public api, not changing it. (I got confused)

OK, I'll revert the commits and create another PR after this is merged.

Unless we want to semver-major it along with v7...? I'm all for it! Good PR (as in Public Relations) for v7.

Your call—I won't have too much time tomorrow, but I'll try and do it Friday.

Alhadis:
Wouldn't it be easier to add an `--esm` switch to opt-into the problematic behaviours, rather than drop a breaking release? That way, users have time to prepare and migrate their codebases, and the caveats @boneskull [described earlier](https://github.com/mochajs/mocha/pull/4038#pullrequestreview-326419355) can be outlined in the switch's documentation.

I haven't read through this thread in its entirety, so forgive me if a similar stopgap was already suggested before.

giltayar:
What if I left `loadOptions`, and added `loadOptionsAsync`, which will be the one used by the CLI? It'll be just like we did with `run` and `runAsync`. Then we can keep the minor version.

giltayar:
@Alhadis - the problem is not in the CLI: the CLI is totally backward compatible. The problem is the backward compatibility of the API. But as I said above, we can use @boneskull's method of keeping the API, and adding a _new_ async call that _it_ is used by the CLI.

sebamarynissen:
Hello everyone, I've been following this thread now for a while and I just wanted to pop in to say that I'm willing to do some beta testing if desired.

I have a personal project which I was planning to migrate completely to ESM once node unflagged it. The only thing keeping me from doing this now is that mocha needs to support it as well.
So, if desired, I can try to do it with a beta version and provide some feedback.

giltayar:
@juergba - I pushed the code that returns backward compatibility by "splitting" `loadOptions` into `loadOptions` and `loadOptionsAsync` and calling `loadOptionsAsync` from the CLI.

If this passes tests, I will squash all the code for `mjs` rc file into one commit and you can decide whether to split it into its own PR or have it inside this PR (which I prefer).

sebamarynissen:
@giltayar I went ahead and installed your fork as my global mocha command using
```
npm i -g giltayar/mocha#node-esm-support
```
so I could play a bit with it. One thing I noticed though is that the cli doesn't recognize `.mjs` *test* files. For example, if I have a test folder containing 1 file called `esm-test.mjs` and i run `mocha .` from this folder, then mocha tells me `Error: No test files found`. When I run `mocha esm-test.mjs` everything works fine. The same is true for `.cjs` files.

Is this intentional, or did you accidentally forget this? In my opinion mocha should load those files as well when running a directory of test files.

giltayar:
Pushed a squashed version of ESM support for mocharc so we can decide whether we want to keep it (my preference, especially now that it's semver-minor) or split it into its own PR.

giltayar:
@sebamarynissen - not intentional! I'll fix it now...

sebamarynissen:
Ok, so I've started to migrate my project. I'll be working from top-down, so first of all I'll migrate all my test files. I'm using the following migration strategy:

 - Change the extension of each test file one by one from `.js` to `.mjs` and then replace all `require`s with `import`s. After each file, run all tests in the folder by `mocha .` and ensure they all pass.
 - Once all test files have been migrated and renamed to `.mjs`, change back the extension of all files to `.js` and add a `package.json` *in the test folder* that contains
```js
{
  "type": "module"
}
```
Note that strictly speaking I could've just added a `package.json` from the beginning and fix all test files at once. As such I wouldn't need to change the extensions back and forth, but I wanted to verify that everything works with the `.mjs` extensions as well as with `.js` combined with `package.json`.

All tests are passing fine after I did this, but when running the tests with the `package.json` file I receive a warning from node:
```
(node:21944) Warning: require() of ES modules is not supported.
require() of path/to/test-file.js from AppData/Roaming/npm/node_modules/mocha/lib/esm-utils.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename test-file.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from path/to/test/folder/package.json.
```

It seems that somehow node thinks that `esm-utils.js` is an ESM file - which it isn't. I have no clue why. Could this be something in mocha, or is it a bug in node?

giltayar:
@sebamarynissen - you're right, it's because of the way we import/require. You can silence it by adding `--no-warning` to Mocha.

But it's definitely a problem that we'll have to solve, either by figuring out the type of the file, or by always `import`-ing. Something to think about...

Thank you for beta-testing this!

sebamarynissen:
Just a general thought, perhaps we could add a flag that always uses dynamic `import` for `.js` files, something like ```--always-import```. As such people can quickly test if there are still bugs with node `import`ing CJS modules. Even without any migration of test files, you'd be able to check if node can properly import all existing CJS test files.

The could could look something like this:
```js
async function doImport(file) {
  return (await import(url.pathToFileURL(file))).default;
}

exports.requireOrImport = async file => {

  if (path.extname(file) === '.cjs') {
    return require(file);
  } else if (path.extname(file) === '.mjs' || alwaysImport) {
    return doImport(file);
  }

  // This way of figuring out whether a test file is CJS or ESM is currently the only known
  // way of figuring out whether a file is CJS or ESM.
  // If Node.js or the community establish a better procedure for that, we can fix this code.
  // Another option here would be to always use `import()`, as this also supports CJS, but I would be
  // wary of using it for _all_ existing test files, till ESM is fully stable.
  try {
    return require(file);
  } catch (err) {
    if (err.code === 'ERR_REQUIRE_ESM') {
      return doImport(file);
    } else {
      throw err;
    }
  }
};
```

boneskull:
ugh. if a warning is thrown just use import(). everyone will complain.

tommyZZM:
Report, an error about `import` spawned with `mocha` from `giltayar:node-esm-support`

**node version:** 13.5.0

**package.json**
```json
{
  "scripts": {
    "test": "mocha -r esm"
  }
}
```

**test/test.js**
```
import { expect } from 'chai'
import toTest from '../lib/to-test'

describe('xxx', function () {
  it('xxx', async function () {
    // ...
  })
});

```

```bash
> npm run test
TypeError [ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING]: A dynamic import callback was not specified.
```

warpr:
@giltayar I've pulled this branch into a project with:

    npm install --save-dev giltayar/mocha#node-esm-support

And I'm running a test file without problems with in my case:

    node_modules/.bin/mocha --ui tdd test/main.mjs

However, using --watch seems to break it:

```sh
$ node_modules/.bin/mocha --watch --ui tdd test/main.mjs

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/kuno/code/lud-cue/test/main.mjs
    at Module.load (internal/modules/cjs/loader.js:986:11)
    at Function.Module._load (internal/modules/cjs/loader.js:896:14)
    at Module.require (internal/modules/cjs/loader.js:1028:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at /home/kuno/code/lud-cue/node_modules/mocha/lib/mocha.js:311:36
    at Mocha.loadFiles (/home/kuno/code/lud-cue/node_modules/mocha/lib/mocha.js:308:14)
    at Mocha.run (/home/kuno/code/lud-cue/node_modules/mocha/lib/mocha.js:890:10)
    at Object.run (/home/kuno/code/lud-cue/node_modules/mocha/lib/cli/watch-run.js:87:22)
    at FSWatcher.<anonymous> (/home/kuno/code/lud-cue/node_modules/mocha/lib/cli/watch-run.js:49:14)
    at FSWatcher.emit (events.js:305:20)
    at /home/kuno/code/lud-cue/node_modules/chokidar/index.js:364:35
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
```

PS. thanks for your work on this!

giltayar:
@warpr - yup. `--watch` doesn't work with ESM modules. It was a conscious decision given that we have no way of deleting modules from the ESM cache (not implemented and probably will not be implemented in Node). Doing it in the "recommended" way (by creating a subprocess for every file change is a lot of work, and it was decided to release it without watch mode.

We should add an error when trying to load an ESM module in watch mode. I'll look into it.

giltayar:
@tommyZZM - I would have been surprised if it _had_ worked with the `esm` module! `esm` is a userland implementation of ESM that transpiles all imports to require. Not sure exactly _why_ it's not working, but it makes sense that it doesn't. 

You should remove the `-r esm` and try it. The whole point of native ESM is that you should not need such a module anymore. Note, however, that there are incompatibilities in how the `esm` package views modules and how native ESM does.

Alhadis:
> Not sure exactly *why* it's not working, but it makes sense that it doesn't.

Probably because the `esm` module [tinkers](https://github.com/standard-things/esm/blob/511d672ae13e8bee13ba19bd7fdc2a3206c9d7d7/esm.js#L33) with Node internals, things which are volatile and subject to change across releases. Undoubtedly there was furniture moved around with Node v13.2, so-to-speak.

@giltayar Just a note I've been running your fork without problems locally, and haven't had any issues (even with very contrived scenarios like [`atom-mocha`](http://github.com/Alhadis/Atom-Mocha)). 👍 and :shipit: from me.

juergba:
@giltayar this PR hasn't got any more comments by maintainers. So I have taken some decisions and propose:
- I would like to publish this PR with an experimental release as eg. `v7.0.0-esm-01` or whatever. This can be done within days, probably.
- please rebase with `master`
- I prefer not to include `mocharc` at this time, please remove. It has no priority and I would like to grow slowly into `esm`. I'm not going to review this part, but keep it for later.
- we have errors to fix (`esm` package, `watch` mode) and warnings to avoid.

I hope you are still on board. Thank you.

giltayar:
@juergba Definitely still on board, and will rebase tonight! I will also revert the `mocharc` code. It's definitely a lot of code changes (from sync to async), so maybe a separate PR is a good idea (will also do that once ESM is in).

giltayar:
@juergba - rebased (without the mocharc change) and passed tests.

juergba:
@giltayar thank you! I will review until tomorrow.

lib/cli/cli.js
```diff
@@ -12,12 +12,12 @@
  */
 
 const debug = require('debug')('mocha:cli:cli');
-const symbols = require('log-symbols');
 const yargs = require('yargs/yargs');
 const path = require('path');
+const ansi = require('ansi-colors');
+const symbols = require('log-symbols');
```
juergba: Please insert these requires to their original lines.
It's pedantic, I know, but reviewing is easier that way.

lib/cli/cli.js
```diff
@@ -48,7 +48,9 @@ exports.main = (argv = process.argv.slice(2)) => {
     .fail((msg, err, yargs) => {
       debug(err);
       yargs.showHelp();
-      console.error(`\n${symbols.error} ${ansi.red('ERROR:')} ${msg}`);
+      console.error(
+        `\n${symbols.error} ${ansi.red('ERROR:')} ${err ? err.message : msg}`
+      );
```
juergba: What is the reason for this change?

lib/cli/run.js
```diff
@@ -87,7 +87,7 @@ exports.builder = yargs =>
       },
       extension: {
         default: defaults.extension,
-        defaultDescription: 'js',
+        defaultDescription: 'js,cjs,mjs',
```
juergba: Remove this line, it's not necessary (see watch-ignore for example).

lib/cli/run.js
```diff
@@ -299,8 +299,14 @@ exports.builder = yargs =>
...
   debug('post-yargs config', argv);
   const mocha = new Mocha(argv);
-  runMocha(mocha, argv);
+
+  try {
+    await runMocha(mocha, argv);
+  } catch (err) {
+    console.error(err.stack || `Error: ${err.message || err}`);
+    process.exit(1);
+  }
```
juergba: I'm unsure wether this will be sufficient to avoid "unhandled promise rejections". When a promise further down the call stack is failing and not catched.  We will see ....

package.json
```diff
@@ -533,7 +533,7 @@
     "growl": "1.10.5",
     "he": "1.2.0",
     "js-yaml": "3.13.1",
-    "log-symbols": "2.2.0",
+    "log-symbols": "^3.0.0",
```
juergba: Please remove "^", we don't use it in our dependencies.

juergba:
To suppress the warning, you could temporarily override the function `process.emitWarning` before loading the files, and set it back to its original value after loading.

sebamarynissen:
@juergba Won't that be a problem if you require a test file that somehow calls `process.emitWarning`? By overriding it this would cause the warning to not be emitted. Unless of course you only filter out somehow the `(node:21944) Warning: require() of ES modules is not supported.` and let the other warnings come through.

juergba:
IMO no, if we override just for loading the files, eg in `loadFilesAsync()`. Running the test files - which possibly emit warnings - comes afterwards.
The order is: loading the test files, then parsing the test runner tree, then running the tests.

sebamarynissen:
That's true, but I think it's still possible if you have something like this
```js
// my-test.js
require('./test-setup.js');
describe('...', function() {});

// test-setup.js
process.emitWarning('Some important warning');
```

Personally, I prefer just using `import()` for everything - as @boneskull proposed - rather than overriding `process.emitWarning` which feels quite hacky.

juergba:
The goal of this PR is to be merged into `master` after having proven stable. We can't publish code with `experimental` state in our productive release. `import()` still has `experimental stability`.

lib/cli/cli.js
giltayar: Done.

lib/cli/cli.js
giltayar: Not sure anymore. Probably an artifact of my earlier playing around with async/sync dual code. The fact that everything is now async made it simpler. 

Removed the diff. It's now back to to what it was.

lib/cli/run.js
giltayar: Why is it not necessary? I looked at `watch-ignore` but couldn't understand why this was so. Don't we want to say that the default extensions we look at are those?

lib/cli/run.js
giltayar: If we're dealing with all the promises down the call stack correctly, then there shouldn't be a problem.

package.json
giltayar: Oops. Didn't notice that. Fixed.

lib/cli/run.js
juergba: `default` property does this job already, `defaultDescription` just overwrites it. If you have a look at the `--help` output, `--watch-ignore` shows everything as needed.

giltayar:
@juergba @sebamarynissen - I'll try and override `process.emitWarning` locally (only for that specific warning) during the require.

giltayar:
@sebamarynissen are you sure this still happens. I just used Node.js v13.6.0, with a `test.js` file inside a folder with a `package.json` with `type: module`. The only warning I got was `(node:39080) ExperimentalWarning: The ESM module loader is experimental`.

Could it be that they removed the warning you saw? Or maybe I didn't understand the scenario?

giltayar:
@juergba - rebased again. It is now rebased against the latest commit.

sebamarynissen:
> @sebamarynissen are you sure this still happens. I just used Node.js v13.6.0, with a `test.js` file inside a folder with a `package.json` with `type: module`. The only warning I got was `(node:39080) ExperimentalWarning: The ESM module loader is experimental`.
> 
> Could it be that they removed the warning you saw? Or maybe I didn't understand the scenario?

@giltayar I'm using node 13.3 (on Windows 10, but shouldn't matter) and I still get there error:
```
(node:3452) Warning: require() of ES modules is not supported.
require() of C:\test.js from C:\...\npm\node_modules\mocha\lib\esm-utils.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename test.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from C:\package.json.
```

My test setup:
 - a `package.json` that only contains `{"type": "module"}`
 - a an empty test file with `.js` extension: `test.js`.

If I run `mocha test.js` I get the error above.

giltayar:
@sebamarynissen - I just tried it using 13.3 (using NVM) and got your warning. Then retried with 1.3.6, and didn't get the warning! Tried 13.4, and didn't get the warning.

@juergba - if it's OK with you, I believe we shouldn't do anything to deal with this problem, given that it happens only if you use ESM (with `type: module`), and only when you use some specific versions of Node.js.

juergba:
@giltayar Ok, we leave the warning problem as is.

Could you have a look at the changelog in branch `release/v7.0.0.-esm1`, please?

giltayar:
@juergba - changelog LGTM.

juergba:
We published yesterday an experimental release: `npm i mocha@7.0.0-esm1`

@giltayar thank you!

maxwell-k:
I have tried the experimental release using Node v12.14.0 and v13.1.0 and it works as expected, thank you!

DominusVilicus:
Had to reinstall mocha globally with `npm i -g mocha@7.0.0-esm1` I think it was causing some errors for me until I did that, and 13.4 wasn't working either for some reason

giltayar:
@DominusVilicus - 13.3 was a problematic version (see above comments for a reference to a problem we found). Could you try 13.4 and above, without the global installation? It should work also without it being globally installed. If the problem persists, could you perhaps elaborate on what it was?

sebamarynissen:
I had a similar issue when installing `v7.0.0-esm1` globally. I don't quite remember the error but it had something to do with a global variable not being available. I installed it over of your `node-esm-support` branch though so I didn't really pay attention to it. I uninstalled the global mocha, then did a fresh install and everything worked fine.

sebamarynissen:
I had a similar issue when updating my global mocha from your `node-esm-support` branch to `7.0.0-esm1`. I uninstalled mocha globally, then reinstalled and everything was fine.

I can still reproduce the issue though when I install mocha again using `npm i -g mocha@7.0.0-esm1`. The error I get:
```
internal/modules/cjs/loader.js:964
    throw err;
    ^

Error: Cannot find module 'glob'
Require stack:
- C:\...\AppData\Roaming\npm\node_modules\mocha\lib\utils.js
- C:\...\AppData\Roaming\npm\node_modules\mocha\bin\mocha
[90m    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:961:17)[39m
[90m    at Function.Module._load (internal/modules/cjs/loader.js:854:27)[39m
[90m    at Module.require (internal/modules/cjs/loader.js:1023:19)[39m
[90m    at require (internal/modules/cjs/helpers.js:72:18)[39m
    at Object.<anonymous> (C:\Users\sebas\AppData\Roaming\npm\node_modules\[4mmocha[24m\lib\utils.js:15:12)
[90m    at Module._compile (internal/modules/cjs/loader.js:1128:30)[39m
[90m    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)[39m
[90m    at Module.load (internal/modules/cjs/loader.js:983:32)[39m
[90m    at Function.Module._load (internal/modules/cjs/loader.js:891:14)[39m
[90m    at Module.require (internal/modules/cjs/loader.js:1023:19)[39m {
  code: [32m'MODULE_NOT_FOUND'[39m,
```

Is there perhaps something wrong with the dependencies?

giltayar:
@sebamarynissen I just tried it (`npm i -g mocha@7.0.0-esm1` in Node v13.6, and ran a small test with a passing and failing test), and it worked well for me.

I also checked the dependencies, both in the code and in the raw package in the npm registry, and glob is _definitely_ there.

sebamarynissen:
@giltayar Yeah I did some more tests and now I can't reproduce it anymore. Seems like npm messed up the installation somehow because not only glob was not there, there was *only one* module in `mocha/node_modules` - guess it was `es-abstract`, but I don't think that's relevant.

I guess that this is effectively the issue @DominusVilicus was facing as well. Perhaps related to node 13.3.

corvino:
Been making light use of the experimental branch for about a week without a problem.

Amund:
It works great with simple import (`import THREE from 'three'`), but fails with a partial one (`import { assert } from 'chai'`). If I import only chai, no problem again (`import chai from 'chai'`).

sebamarynissen:
@Amund That's because Node [doesn't support named exports from CJS](https://nodejs.org/api/esm.html#esm_import_statements). There are ways around this, but it requires the package to support it.

I have filed two PRs for this: [one for chai](https://github.com/chaijs/chai/pull/1317) and [one for three.js](https://github.com/mrdoob/three.js/pull/18498) which show how it can be done.

giltayar:
@amund this is because Node.js itself doesn't support named imports from commonjs

Amund:
@giltayar @sebamarynissen Thanks for the quick response !

jkrems:
Is there a timeline for when this may be released in a proper (pre-) release of mocha?

juergba:
@jkrems no, no timeline yet.
IMO it's to early to do so. I would also have expected more user feedback, so the demand seems to be moderate at the moment.

Alhadis:
> I would also have expected more user feedback

What sort of feedback? If it's working perfectly fine for everybody, would you expect an influx of *"works great, thanks!"*-type replies...?

This was even featured on the ES.next mailing list some time ago, so I don't believe it's a trivial improvement whatsoever.

juergba:
Just every kind of constructive feedback.
Especially by consumers using Mocha programmatically, there you can always find some weird edge cases.

juergba:
@giltayar we should coordinate our next steps.

- release or pre-release?
- date? - you should have some time left for fire fighting
- some last changes, mainly JSDoc nits, I think
- test coverage was declining, do we need more tests?
- rebase, conflicts, squash
- release note
- update PR description
- update Wiki [Using-Mocha-programmatically](https://github.com/mochajs/mocha/wiki/Using-Mocha-programmatically)
- Mocha browser: what to do with `mocha.runAsync`?

Do you have anything to add to the list?

giltayar:
@juergba Nope. Looks good.

> release or pre-release?
I would go with release... I seriously don't think there is anyway we can otherwise test it.

> date? - you should have some time left for fire fighting
Should be after the 15th, as I am travelling in the US these two weeks, and wouldn't have time to fix issues. But after the 15th should be OK.

> some last changes, mainly JSDoc nits, I think
Whatever you need.

> test coverage was declining, do we need more tests?
It's very strange. Sometimes it went down some tenth of a percent, sometimes not.

> rebase, conflicts, squash
Yup. Not sure what the specific procedure is.

> release note
Yup. Not sure what the specific procedure is.

> update PR description
Yup. Not sure what the specific procedure is.

> update Wiki Using-Mocha-programmatically
Yup. Not sure what the specific procedure is.

> Mocha browser: what to do with mocha.runAsync?
Theoretically, it shouldn't be able to run in the browser.

jkrems:
> I would also have expected more user feedback

FWIW: My testing says "looks like it's working". But I assume there's a chicken-and-egg problem: There's no big test suites written as (native) modules because nobody would rewrite a big test suite just to try out a prerelease branch. And without big test suites that actually use ESM, it's hard to get strong feedback for the new feature. So :+1: to @giltayar's comment above:

> I would go with release... I seriously don't think there is anyway we can otherwise test it.

juergba:
@jkrems Your point of view is incomplete. Your statement "My testing says "looks like it's working" is just one side of the mountain, the plain one.
I have to be sure that there are no regressions with the existent Mocha. If the ESM loading works, but I have unexpected regressions in existent features, this would be a bad bargain. Not one test would have to be rewritten to test this part, to prove that the existent features keep working.

juergba:
@giltayar it's your turn for (after 15th, when you are home):
- rebase and resolve conflicts, please don't squash yet. We updated ESLint, there may wait some surprises for you.
- update PR description: doesn't reflect your implementation anymore. It has low priority.

My turn: last review

.eslintrc.yml
```diff
@@ -32,7 +32,12 @@ overrides:
       ecmaVersion: 2017
     env:
       browser: no
-
+  - files:
+      - esm-utils.js
+    parserOptions:
+      ecmaVersion: 2018
+      sourceType: module
+    parser: babel-eslint
```
juergba: - shouldn't we add: `env: browser: false` here?
- do we still need _babel-eslint_, despite our _esLint_ update?

docs/index.md
```diff
@@ -1539,6 +1540,38 @@ Alias: `HTML`, `html`
 
 **The HTML reporter is not intended for use on the command-line.**
 
+## Node.JS native ESM support
+
+Mocha supports writing your tests as ES modules (without needing to use the `esm` polyfill module), and not just using CommonJS. For example:
```
juergba: pls add one line: `> _New in v7.1.0_`

docs/index.md
```diff
@@ -1539,6 +1540,38 @@ Alias: `HTML`, `html`
...
+});
+```
+
+To enable this you don't need to do anything special. Write your test file as an ES module. In Node.js
+this means either ending the file with a `.mjs` extension, or, if you want to use the regular `.js` extension, by
+adding `"type": "module"` to your `package.json`.
+More information can be found in the [Node.js documentation](https://nodejs.org/api/esm.html).
+
+> Mocha supports ES modules only from Node.js v12.11.0 and above. Also note that
+> to enable this in vesions smaller than 13.2.0, you need to add `--experimental-modules` when running
```
juergba: typo: `vesions` instead `versions`

docs/index.md
```diff
@@ -1539,6 +1540,38 @@ Alias: `HTML`, `html`
...
+> Mocha. From version 13.2.0 of Node.js, you can use ES modules without any flags.
+
+### Limitations
+
```
juergba: this listing isn't complete, eg. `--require` and the mock story are missing.

lib/cli/run-helpers.js
```diff
@@ -98,21 +98,29 @@ exports.handleRequires = (requires = []) => {
...
  * @private
  */
 exports.singleRun = (mocha, {exit}, fileCollectParams) => {
   const files = collectFiles(fileCollectParams);
   debug('running tests with files', files);
   mocha.files = files;
-  return mocha.run(exit ? exitMocha : exitMochaLater);
+
```
juergba: My IDE tells me, `singleRun` could be converted to an async function. I think she is clever, this module is not exposed to our browser part. I leave it upto you.

lib/cli/run-helpers.js
```diff
@@ -98,21 +98,29 @@ exports.handleRequires = (requires = []) => {
...
 
 /**
  * Actually run tests
  * @param {Mocha} mocha - Mocha instance
  * @param {Object} opts - Command line options
  * @private
+ * @returns {Promise}
  */
 exports.runMocha = (mocha, options) => {
   const {
```
juergba: She remains silent, but same here I Think.

lib/mocha.js
```diff
@@ -883,3 +932,33 @@ Mocha.prototype.run = function(fn) {
...
 };
+
+/**
+ * Runs root suite and invokes `fn()` when complete. Supports ES Modules.
+ *
+ * @description
+ * To run tests multiple times (or to run tests in files that are
+ * already in the `require` cache), make sure to clear them from
+ * the cache first!
+ *
```
juergba: - pls add that ES modules can't be cleared from the cache.
- _Mocha.prototype.unloadFiles_: aprox line380, add to description that ES modules can't be deleted from cache.

These description will be printed in our API docs.

lib/cli/run.js
juergba: You have to remove the complete line: `defaultDescription: 'js',` otherwise it overwrites `defaults.extension`.

lib/cli/run.js
```diff
@@ -299,8 +299,14 @@ exports.builder = yargs =>
     .number(types.number)
     .alias(aliases);
```
juergba: pls delete L90: `defaultDescription: 'js',`

lib/mocha.js
```diff
@@ -883,3 +932,33 @@ Mocha.prototype.run = function(fn) {
 
   return runner.run(done);
 };
+
+/**
+ * Runs root suite and invokes `fn()` when complete. Supports ES Modules.
```
juergba: There is no `fn()` in the async case.

juergba:
@giltayar using Mocha programmatically:
```js
mocha.run(failures => process.exitCode = failures ? 1 : 0)
  .on('suite end', suite =>console.dir(suite));
```
There is no way to register event listeners on the runner instance when using `Mocha@runAsync()`, right?

juergba:
@giltayar I did some testing on the [copy-esm](https://github.com/mochajs/mocha/tree/copy-esm) branch.
- I removed `mocha.runAsync()`
- call `loadFilesAsync` in `singleRun` and label files as loaded
- `mocha.run`:  don't load a second time
```js
mocha.addFile('../test.js');

mocha.loadFilesAsync().then(
  mocha.run().on('suite end', function(suite) {
    console.dir(suite);
  })
);
```
`mocha.loadFilesAsync` would have to be public in this case.

emma-borhanian:
I tried out this prerelease and ran into some problems/complexity.

### Story
I'm interested in switching to mocha from jest because it supports native modules and I'm using an electron/React/sqlite stack for the productivity app I'm building. I'm already using [type: "module"](https://nodejs.org/api/esm.html#esm_package_json_type_field) because I wanted a consistent import-style across my code and React convention is ESM.

### Technical background
* I'm using babel (jsx, class properties, *not* for import-to-require) for React electron renderer-code and no transpilation for electron main code
* `"scripts": { "mocha": "NODE_ENV=development node --experimental-specifier-resolution node --experimental-loader ./src/lib/mocha/transpiler.js -- node_modules/mocha/bin/_mocha --config ./config/mocha.config.cjs" }`
* Using the still-unstable [esm_transpiler_loader](https://nodejs.org/api/esm.html#esm_transpiler_loader) node API
  * to run babel on electron-renderer (React, jsx, etc) files
  * to skip electron css/asset imports

### Problem
Node doesn't throw `ERR_REQUIRE_ESM` for `.jsx` files, instead throws `SyntaxError: Cannot use import statement outside a module`
https://github.com/mochajs/mocha/blob/v7.0.0-esm1/lib/esm-utils.js#L21
https://github.com/nodejs/node/blob/v13.8.0/lib/internal/modules/cjs/loader.js#L1161

#### Solution
```
// MAINTAIN(mocha-version): Hack: mocha wants an exception thrown on load to switch to ESM: https://github.com/mochajs/mocha/blob/v7.0.0-esm1/lib/esm-utils.js#L21
require.extensions['.jsx'] = function (module, filename) {
  // Emulate https://github.com/nodejs/node/blob/v13.8.0/lib/internal/modules/cjs/loader.js#L1167
  throw { code: 'ERR_REQUIRE_ESM' }
}
```

### Aside: Many modules don't support named imports properly despite their documentation
Documentation suggests things like `import React, {Component} from 'react'` which are not compatible with [type: "module"](https://nodejs.org/api/esm.html#esm_package_json_type_field).[](url) with the way those modules are written

lib/esm-utils.js
```diff
@@ -0,0 +1,35 @@
...
+  // This way of figuring out whether a test file is CJS or ESM is currently the only known
+  // way of figuring out whether a file is CJS or ESM.
+  // If Node.js or the community establish a better procedure for that, we can fix this code.
+  // Another option here would be to always use `import()`, as this also supports CJS, but I would be
+  // wary of using it for _all_ existing test files, till ESM is fully stable.
+  try {
+    return require(file);
+  } catch (err) {
+    if (err.code === 'ERR_REQUIRE_ESM') {
+      return import(url.pathToFileURL(file));
```
emma-borhanian: Node doesn't throw ERR_REQUIRE_ESM for .jsx files, instead throws SyntaxError: Cannot use import statement outside a module:
https://github.com/nodejs/node/blob/v13.8.0/lib/internal/modules/cjs/loader.js#L1161

I found a hacky [workaround](https://github.com/mochajs/mocha/pull/4038#issuecomment-587004460).

Could you check `package.json` and import by default if the requiring package has set `type: "module"`?

lib/utils.js
```diff
@@ -831,3 +831,46 @@ exports.defineConstants = function(obj) {
...
+ *
+ * @description
+ * Versions prior to 10 did not support ES Modules, and version 10 has an old incompatibile version of ESM.
+ * This function returns whether Node.JS has ES Module supports that is compatible with Mocha's needs,
+ * which is version 12 and older
+ *
+ * @param {Boolean} unflagged whether the support is unflagged (`true`) or only using the `--experimental-modules` flag (`false`)
+ * @returns {Boolean} whether the current version of Node.JS supports ES Modules in a way that is compatbile with Mocha
+ */
+exports.supportsEsModules = function(unflagged) {
```
juergba: This function is called only once (?), without parameter => `unflagged` === false.
So `unflagged` doesn't make any sense.

lib/mocha.js
```diff
@@ -311,6 +314,43 @@ Mocha.prototype.loadFiles = function(fn) {
...
+ * @see {@link Mocha#run}
+ * @see {@link Mocha#unloadFiles}
+ *
+ * @returns {Promise}
+ */
+Mocha.prototype.loadFilesAsync = function() {
+  var self = this;
+  var suite = this.suite;
+
+  if (!esmUtils) {
```
juergba: I'm unsure wether `esmUtils` is working correctly.
Our CI tests do not enter `if (!esmUtils) {`, not one single time. That's weird.

juergba:
@giltayar I rebased and resolved conflicts.

juergba:
@giltayar I removed `Mocha#runAsync` and moved `Mocha#loadFilesAsync` up into `singleRun`.
```
yargs.handler (sync => async)
> runMocha (sync => async)

>> singleRun (sync => async)                      >> watchRun
>>> collectFiles                                  >>> collectFiles
>>> loadFilesAsync (new, async, public)

>>> mocha.run (remains sync, returns runner instance)
>>>> loadFiles (remains sync, evtl. skipped)
>>>> runner.run
```

juergba:
@giltayar I pushed some changes due to code review and make-happy coveralls.

lib/mocha.js
juergba: Coveralls seems to run only with Node v13, so `esmUtils` is never false.

lib/utils.js
giltayar: It does, because it was used in the tests (`esm.spec.js`). But in the test and in the change you did to the function, you assumed that any v13 is unflagged, which today is a close enough approximation to the truth...

giltayar:
@juergba - I saw what you did with removing `Mocha#runAsync` (by separating loading via `Mocha#loadAsync` and running via `Mocha#run`). I like it. Simpler.

All in all, I went over the changes you did, and they make sense. Thanks for the comprehensive review and changes!

lib/utils.js
juergba: In the productive code `unflagged` has never been used. The method did not check wether user set the `--experimental-modules` or not. We should not adapt our productive code only to make pass our tests.

> you assumed that any v13 is unflagged

Yes, this is true for our CI tests.

juergba:
@giltayar re `Mocha#runAsync`: I did this change because this method does not return `Promise<Runner>`. The possibility to register events to the `runner` is a must-have requirement when using Mocha programmatically.

Can I squash my commits into yours? You remain the owner this way, otherwise I will merge two separate commits.

lib/esm-utils.js
juergba: @emma-borhanian this issue is outside of this PR's scope, I would like to get it merged as is.
Afterwards the ESM implementation will grow and we are open to suggestions. But IMO checking for the nearest `package.json` (for each test file) is too expensive and actually Node's job.

giltayar:
@juergba - you can squash, thanks. And, yes, semver-minor makes sense.

PR #4046, opened 2019-10-05
modify alt text of image from Backers to Sponsors inside Sponsors section in Readme

Mia-jeong:
When I looked at Readme.md I found out that inside Sponsors sections, the alternative text of images is written as Backer. 
so I turned it into Sponsor

스크린샷 2019-10-05 오후 4 54 34

### Description of the Change

I changed the text from `MochaJS Backer` to `MochaJS Sponsor`


juergba:
I just have a question for better understanding.
What is the purpose of this field `![MochaJS Sponsor]`?

Mia-jeong:
@juergba Hi :) It's the alternative text of image same as `alt` of ``. If the required image is not loading with some reasons. the text in the field will show up on screen instead.

PR #4051, opened 2019-10-07
Make unhide function work as expected in html reporter

### LINKED ISSUE #4048: Html Reporter Filtering Issue when number of passes/failures is 0
### Prerequisites

### Description

Filtering by passes/failures doesn't work properly when either number of test passes/failures is 0.

### Steps to Reproduce

Case 1. Run a browser test with all passing tests (n passes, 0 failures). Filter by failure, then filter by passes again. 

Case 2. Run a browser test with all failing tests (0 passes, n failures). Filter by passes, then filter by failures again.

**Expected behavior:** 
(Case 1) It should display an empty result when filtered by failure, then show all passing tests when filtered again by passes.

(Case 2) It should display an empty result when filtered by passes, then show all failing tests when filtered again by failures.

**Actual behavior:** It displays only a single passing/failing test when filtered again.

(Case 1)
Original:
스크린샷 2019-10-05 오후 10 14 05

Filtered by Failures:
스크린샷 2019-10-05 오후 10 14 17

Filtered by Passes Again:
스크린샷 2019-10-05 오후 10 14 29

(Case 2)
Original:
스크린샷 2019-10-05 오후 10 22 11

Filtered by Passes:
스크린샷 2019-10-05 오후 10 22 22

Filtered by Failures Again:
스크린샷 2019-10-05 오후 10 22 35

**Reproduces how often:** Every time

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 6.2.0
- The output of `node --version`: 10.15.3
- Your operating system
  - name and version: MacOS Mojave 10.14.6
- Your shell (e.g., bash, zsh, PowerShell, cmd): Terminal
- Your browser and version (if running browser tests): Chrome 77.0.3865.90 (64 bit)

pec9399:
### Description

Fixed the iteration of the HTMLCollection object in the unhide function of the html reporter. Since the length of the HTMLCollection changes as the elements' class names get updated (as HTMLCollection works 'live'), it cannot be updated using an index-based for loop.

### Description of the Change

Changed the index-based for loop iteration to a while loop, updating the first element while the length of the HTMLCollection is larger than 0.

It a bug (although minor) that makes filtering by passes/failures doesn't work properly as expected.

### Benefits

Filtering by passes / filters in html reporter works properly even after unhiding objects.

### Possible Drawbacks

None

### Applicable issues

Closes #4048

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -357,8 +357,8 @@ function hideSuitesWithout(classname) {
  */
 function unhide() {
   var els = document.getElementsByClassName('suite hidden');
-  for (var i = 0; i < els.length; ++i) {
-    els[i].className = els[i].className.replace('suite hidden', 'suite');
+  while (els.length > 0) {
+    els[0].className = els[0].className.replace('suite hidden', 'suite');
   }
 }
 


juergba:
Could someone test, confirm and review this PR, please? Members, contributors, others ... ?
@Mia-jeong @HyunSangHan @Lindsay-Needs-Sleep

Lindsay-Needs-Sleep:
@juergba @pec9399 
Just tried this out.  Works great.

Tried the described situations (all passes, all failures), and a mix of passes and fails.  All 3 situations worked as expected.

Ran `npm run test`, all tests passed.

juergba:
@Lindsay-Needs-Sleep thanks 👍

juergba:
> Since the length of the HTMLCollection changes as the elements' class names get updated (as HTMLCollection works 'live'), [...]

This sounds very weird to me with little understanding of HTML/browsers. But I got so far that:
- `document.getElementsByClassName` returns an `HTMLCollection` which is a **live** list
- `document.querySelector` returns a `NodeList` which is a **static** list

Is `hideSuitesWithout()` working correctly? As `unhide()` it's also using `getElementsByClassName`.

pec9399:
For clarification, as you noted, ```document.getElementsByClassName``` returns an HTMLCollection which is a live list, and therefore changing the class name of an element in the HTMLCollection will remove that element from this collection right away.  

In the code below, the code in line 3 will change the ith element's class name to 'suite' and consequently pop the ith element from the variable 'els' since the ith element is not included in ```document.getElementsByClassName('suite hidden')``` anymore.

```
  1  var els = document.getElementsByClassName('suite hidden');
  2  for (var i = 0; i < els.length; i++) {
  3    els[i].className = els[i].className.replace('suite hidden', 'suite');
  4  }
```

This will inevitably change the indices of the remaining elements still in the list, making index-based loops cause unexpected behaviors when iterating over the HTMLCollection. So, I changed the code to change the class name of each element while the list is not empty rather than iterating over each index.

Hope this helps your understanding.

As of ```hideSuitesWithout()```, it doesn't really have this problem. The code in line 6 will change the ith element in the variable 'suites' to 'suites hidden', but it still belongs to ```document.getElementsByClassName('suite')```, so it will not be popped from the original HTMLCollection.

```
1  function hideSuitesWithout(classname) {
2   var suites = document.getElementsByClassName('suite');
3   for (var i = 0; i < suites.length; i++) {
4     var els = suites[i].getElementsByClassName(classname);
5     if (!els.length) {
6      suites[i].className += ' hidden';
7     }
8   }
9  }
```

Issue #4047, opened 2019-10-05
🚀 Feature: Better error for `--file` option with non-existent file


outsideris:
If specify non-existed file with `--file` option, mocha throw expection.

```
./bin/mocha --file non-existed
/Users/mocha/node_modules/yargs/yargs.js:1163
      else throw err
           ^

Error: Cannot find module '/Users/mocha/non-existed'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.Module._load (internal/modules/cjs/loader.js:508:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at /Users/mocha/lib/mocha.js:309:36
    at Array.forEach (<anonymous>)
    at Mocha.loadFiles (/Users/mocha/lib/mocha.js:306:14)
    at Mocha.run (/Users/mocha/lib/mocha.js:784:10)
    at Object.exports.singleRun (/Users/mocha/lib/cli/run-helpers.js:108:16)
    at exports.runMocha (/Users/mocha/lib/cli/run-helpers.js:143:13)
    at Object.exports.handler.argv [as handler] (/Users/mocha/lib/cli/run.js:305:3)
    at Object.runCommand (/Users/mocha/node_modules/yargs/lib/command.js:242:26)
    at Object.parseArgs [as _parseArgs] (/Users/mocha/node_modules/yargs/yargs.js:1104:24)
    at Object.parse (/Users/mocha/node_modules/yargs/yargs.js:566:25)
    at Object.exports.main (/Users/mocha/lib/cli/cli.js:68:6)
    at Object.<anonymous> (/Users/mocha/bin/mocha:131:29)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:754:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

It should be handled like `--require`.

```
./bin/mocha --require non-existed
mocha inspect [spec..]

Run tests with Mocha

Rules & Behavior
[...]

PR #4056, opened 2019-10-09
Handled exception of`--file` option with non-existed file(#4047)

leehwarang:
### Description of the Change

In **handleRequires**, a function that handles exceptions when invoking a module that does not exist with the `--require` option. However, function **handleRequires** can also be used to load non-existent files with the `--file` option.

So I added this function to be called when I loaded the file using --file option. And I modified the annotation of this function.

>./bin/mocha --require non-exist-module

module

>./bin/mocha --file non-exist-file

file

### Benefits

When user import a non-existent file or module, command line can show the user the same error message and format.

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -75,10 +75,11 @@ exports.list = str =>
   Array.isArray(str) ? exports.list(str.join(',')) : str.split(/ *, */);
 
 /**
- * `require()` the modules as required by `--require <require>`
- * @param {string[]} requires - Modules to require
+ * `require()` the modules as required by `--require <require>` or `--file <filename>`
+ * @param {string[]} requires - Modules or File to require
  * @private
  */
+
 exports.handleRequires = (requires = []) => {
   requires.forEach(mod => {
     let modpath = mod;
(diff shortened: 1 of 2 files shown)


outsideris:
ref: #4047

lib/cli/run-helpers.js
```diff
@@ -91,6 +92,19 @@ exports.handleRequires = (requires = []) => {
...
 
+/**
+ * @param {string[]} files - Modules to require by `--file <file>`
+ * @private
+ */
+
+exports.handleFiles = (files = []) => {
+  files.forEach(file => {
+    if (!fs.existsSync(file, {cwd})) {
+      throw Error(`Cannot find module '${file}'`);
```
craigtaub: Could you add a test for this please.

lib/cli/run-helpers.js
```diff
@@ -79,6 +79,7 @@ exports.list = str =>
  * @param {string[]} requires - Modules to require
  * @private
  */
+
```
craigtaub: Doesnt need the space (same below)

JoshuaKGoldberg:
🤖 Closing out old PRs to keep the queue small. Please yell at me if you'd like to keep working on this. Cheers!

PR #4057, opened 2019-10-09
Fixed 'extends' option to work. (#3916)

### LINKED ISSUE #3916: Does the extending configuration feature work?
Opening this as a question as I'm not yet sure whether it's a bug or a user error 😄 

The [docs on extending configuration](https://mochajs.org/#extending-configuration) state:
> Configurations can inherit from other modules using the extends keyword

However, I've been unable to get this to work and I haven't been able to find any Mocha-specific examples leveraging `extends`. I've tried using `extends` both with a relative path to a json file, as well as with a module as referenced in the [yargs docs](http://yargs.js.org/docs/#api-configobject-extends-keyword) linked from the Mocha docs, and I also tried the various Mocha config file formats (js, yaml, json, etc.) to define the Mocha config with the `extends` keyword, all to no avail. 

I created a really simple repo [here](https://github.com/calebcartwright/mocha-config-sample) that reproduces this behavior:

`ui` should be set to `tdd` via the `extends` reference in the .mocharc.yml file, but running an `npm test` results in a reference error `ReferenceError: suite is not defined` leading me to believe that `extends` is not working since the `ui` is still set to the default `bdd` 

Any help would be greatly appreciated!

liloyek:
Hi, I fixed this issue by applying default values at yargs(), not yargs-parser.detailed().

- To be exact, applying 'extend' option is a feature of yargs().
- yargs-parser just parses literally options of cli.
- Therefore, extended configuration is overrided in yargs-parser.detailed() by default values even if there is a extends option.
- So default values should be concatenated in yargs() so that extended configuration could be applied.

#### Before
Default values are applied by yargs-parser.detailed in options.js.
```
// lib/cli/options.js

...
const result = yargsParser.detailed(args, {
    configuration,
    configObjects,
    default: defaultValues, // here
    coerce: coerceOpts,
    narg: nargOpts,
    alias: aliases,
    string: types.string,
    array: types.array,
    number: types.number,
    boolean: types.boolean.concat(nodeArgs.map(pair => pair[0]))
});
...
```

```
// lib/cli/cli.js

...
yargs()
    ...
    .parserConfiguration(YARGS_PARSER_CONFIG)
    .config(args)
    .parse(args._);
};
...
```

#### After
```
// lib/cli/options.js

...
const result = yargsParser.detailed(args, {
  configuration,
  configObjects,
  coerce: coerceOpts,
  narg: nargOpts,
  alias: aliases,
  string: types.string,
  array: types.array,
  number: types.number,
  boolean: types.boolean.concat(nodeArgs.map(pair => pair[0]))
});
...
```

```
// lib/cli/cli.js

...
yargs()
    ...
    .parserConfiguration(YARGS_PARSER_CONFIG)
    .config(args)
    .default(mocharc) // here
    .parse(args._);
};
...
```

#### Etc
- Added a test code at /test/integration/options/extends.spec.js.
- Test code of 'extension' is deleted because now default values are not all applied in loadOptions() and there is a integration test for testing extension option including deleted test.

### Applicable issues
fixed #3916


liloyek:
@juergba Thank you for review.

I chose to add default builder as above for yargs() because I thought it is the most simplest way to solve this problem. 

So you mean I must not change the structure of parsing cli options and should keep the location of applying default, right?

Then is it ok to add a extra logic for processing '--extends' options in loadOptions() instead?

juergba:
I already commented this issue [here](https://github.com/mochajs/mocha/issues/3916#issuecomment-494267418). It's certainly more complex than a `good-first-issue` and not suited for a quick-and-dirty merge.

> Then is it ok to add a extra logic for processing '--extends' options in loadOptions() instead?

Yes. I don't see any other solution at the moment.

Our configuration files are probably the ones which inherit by using `--extend`. Check [here](https://github.com/yargs/yargs-parser#requireyargs-parserargs-opts) wether `opt.config` can be of any help. Which priority does it has? Before `opts.configObjects` or after? You really have to study and understand our parsing and its priorities. And yargs-parser and its features.

liloyek:
Ok, I should've understood more about features of yargs and yargs-parser.
I'll try to change the code as the way above.
Thank you for comments.

boneskull:
@liloyek Do you intend to continue with this PR?

liloyek:
@boneskull I still want to continue, but I think it's better to close this PR and make new one after I finish it. Is it ok?

PR #4059, opened 2019-10-11
fix: add global leaks detection on watch-run

### LINKED ISSUE #1948: 🐛 Bug: --check-leaks only detects leak the first time (e.g. when running twice with `--watch`)
I have the following bit of code, for which mocha correctly detects a leak:

```
describe('leak detection', function() {

    function testMe(){
        x = 123; // here I add 'var ' to make the test pass
    }

    it('should not leak', function() {
        testMe();
    });
});
```

By substituting 'x' for 'var x' the test passes. Seemingly, the same applies when running mocha with the --watch task, but when I remove the `var` statement the test still passes without any error about leaking. This is a bug.

```
$ mocha --check-leaks --watch

  leak detection
    ✓ should not leak
    1) should not leak

  1 passing (13ms)
  1 failing

  1) leak detection should not leak:
     Error: global leak detected: x

  leak detection
    ✓ should not leak

  1 passing (1ms)

  leak detection
    ✓ should not leak

  1 passing (2ms)
```

Using mocha 2.3.3 on Node 4.2.1

anpaul0615:
### Description of the Change

- apply global variable reset process when rerunning test in watch mode
- add some test (that detect global leaks and rerun that test through touch that test file, and then detect global leak again)

Fixes #1948.

### DIFF
--- a/lib/cli/watch-run.js
+++ b/lib/cli/watch-run.js
@@ -85,10 +85,17 @@ const createRerunner = (mocha, beforeRun) => {
 
   let rerunScheduled = false;
 
+  let defaultGlobals = null;
+
   const run = () => {
     try {
       beforeRun();
       resetMocha(mocha);
+
+      if (!defaultGlobals) {
+        defaultGlobals = Object.keys(global);
+      }
+
       runner = mocha.run(() => {
         runner = null;
         if (rerunScheduled) {
@@ -114,6 +121,15 @@ const createRerunner = (mocha, beforeRun) => {
   };
 
   const rerun = () => {
+    if (defaultGlobals) {
+      Object.keys(global).forEach(k => {
+        if (defaultGlobals.indexOf(k) === -1) {
+          delete global[k];
+        }
+      });
+      defaultGlobals = null;
+    }
+
     rerunScheduled = false;
     eraseLine();
     run();
--- a/lib/cli/watch-run.js
+++ b/lib/cli/watch-run.js
@@ -85,15 +85,15 @@ const createRerunner = (mocha, beforeRun) => {
 
   let rerunScheduled = false;
 
-  let defaultGlobals = null;
+  let initialGlobals = null;
 
   const run = () => {
     try {
       beforeRun();
       resetMocha(mocha);
 
-      if (!defaultGlobals) {
-        defaultGlobals = Object.keys(global);
+      if (!initialGlobals) {
+        initialGlobals = Object.keys(global);
       }
 
       runner = mocha.run(() => {
@@ -121,13 +121,13 @@ const createRerunner = (mocha, beforeRun) => {
   };
 
   const rerun = () => {
-    if (defaultGlobals) {
+    if (initialGlobals) {
       Object.keys(global).forEach(k => {
-        if (defaultGlobals.indexOf(k) === -1) {
+        if (initialGlobals.indexOf(k) === -1) {
           delete global[k];
         }
       });
-      defaultGlobals = null;
+      initialGlobals = null;
     }
 
     rerunScheduled = false;
--- a/test/integration/fixtures/options/watch/test-global-leak.fixture.js
+++ b/test/integration/fixtures/options/watch/test-global-leak.fixture.js
@@ -1,0 +1,14 @@
+describe('leak detection', function() {
+  function testMe() {
+    x = 123; // leak variable
+  }
+
+  it('should not leak', function(done) {
+    testMe();
+    done(); // just pass
+  });
+
+  before(function() {});
+
+  after(function() {});
+});
--- a/test/integration/options/watch.spec.js
+++ b/test/integration/options/watch.spec.js
@@ -267,6 +267,19 @@ describe('--watch', function() {
         expect(results[1].tests, 'to have length', 2);
       });
     });
+
+    it('check global variable leaks when test reruns', function() {
+      const testFile = path.join(this.tempDir, 'test.js');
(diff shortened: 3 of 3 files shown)


outsideris:
ref: #1948

test/integration/fixtures/options/watch/test-global-leak.fixture.js
```diff
@@ -0,0 +1,14 @@
...
+  }
+
+  it('should not leak', function(done) {
+    testMe();
+    done(); // just pass
+  });
+
+  before(function() {});
+
+  after(function() {});
```
outsideris: hooks needed here?

lib/cli/watch-run.js
```diff
@@ -114,6 +121,15 @@ const createRerunner = (mocha, beforeRun) => {
   };
 
   const rerun = () => {
+    if (defaultGlobals) {
+      Object.keys(global).forEach(k => {
+        if (defaultGlobals.indexOf(k) === -1) {
```
outsideris: use `if (~defaultGlobals.indexOf(k)) {` like https://github.com/mochajs/mocha/blob/master/lib/runner.js#L215 .
same code style is better.

lib/cli/watch-run.js
```diff
@@ -85,10 +85,17 @@ const createRerunner = (mocha, beforeRun) => {
 
   let rerunScheduled = false;
 
+  let defaultGlobals = null;
```
outsideris: Can we use another name? Because I feel it is not "default".

test/integration/fixtures/options/watch/test-global-leak.fixture.js
anpaul0615: because `Runner.prototype.checkGlobals()` is invoked every test-end-event and hook-end-event. (https://github.com/mochajs/mocha/blob/master/lib/runner.js#L136-L141)

lib/cli/watch-run.js
anpaul0615: ok. I got it.

lib/cli/watch-run.js
anpaul0615: how about "prevGlobals" or "memoGlobals"?

JoshuaKGoldberg:
👋 coming back to this @anpaul0615, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

_(I do see that your last message was asking for a naming suggestion - unless directed otherwise we'll try to review it once we're ramped up enough!)_

test/integration/fixtures/options/watch/test-global-leak.fixture.js
JoshuaKGoldberg: That makes sense, but isn't intuitive from the test. Request: how about a quick comment explaining why these are here?

JoshuaKGoldberg (approved):
Looks great! 🙌 

Leaving open and approved for an additional look by @mochajs/maintenance-crew. But only a couple of small nitpicks from me.

lib/cli/watch-run.js
JoshuaKGoldberg, line 1: @Uzlopak @voxpelli and I talked about this approach: we think it's promising and gets the job done partially, but there are blocking edge cases around side effects. For example, if the first time a test is run, the test imports cached polyfills that add globals, deleting the global after the test would mean subsequent tests would fail. `delete global[k]` isn't something that Mocha can reasonably do.

Instead, we think a more workable approach would be to have the runner check for global leaks against the initial state of globals _before all tests_. I.e. keep the `initialGlobals` idea, but not the `delete global[k]` idea. Doing so would mean that necessary globals that can only be introduced exactly one won't be deleted on `--watch`. _(does that make sense?)_

One drawback of this proposed approach is that if a leak exists in a test, then is fixed, it'll still be reported on re-runs. Mocha's messaging when `--check-leaks` _and_ `--watch` are enabled will have to note that if the leak is fixed then the user will have to restart Mocha.

So, two requests, please:

* fix the behavior so that globals are never deleted, and instead runner checks leaks against the same initial globals every time
* add a unit test that adding a global variable only in the first run is still available in subsequent runs

See also #4954 / https://github.com/mochajs/mocha/issues/4954#issuecomment-1942016336. Part of why we're looking at deprecating the `--check-leaks` feature long-term is that it isn't comprehensive enough to get many common modern use cases. The only reliable way for a test framework to achieve true encapsulation(-ish) and leak detection is with VMs the way Jest & co do. For Mocha, you'd have to build a plugin and/or use linting and/or / type checking and/or `"use strict";` and/or proper ESM.

JoshuaKGoldberg (changes requested):
Thanks again for working on this - requesting some design changes. Please let us know if we've missed something! 🙂

JoshuaKGoldberg:
👋 ping @anpaul0615, is this still something you have time for?

JoshuaKGoldberg:
It's been over a month since last ping, so closing this out as stale. If you again have time for this & the issue hasn't yet been fixed, feel free to reopen @anpaul0615.

If anybody else sees this PR is closed and wants to tackle it, great! Go ahead and send your own PR. If it includes substantial code from this one, please include a co-author attribution.

Cheers! 🤎

mark-wiemer:
@mostafaNazari702 we can build off this PR (we will give co-author attribution of course!)

mostafaNazari702:
> @mostafaNazari702 we can build off this PR (we will give co-author attribution of course!)

Sure, Absolutely, i will get started ASAP.

PR #4061, opened 2019-10-11
Fix docs.watch for the API docs(#4060)

### LINKED ISSUE #4060: Watch mode for the API docs
While I was trying to contribute to docs, I found that watch mode(`npm start docs.watch`) does not work for the API docs.

There is not enough description for this in `README.md`, so I think it may be necessary to fix watch mode to work for not only [mocha website](https://mochajs.org/) but also [API docs](https://mochajs.org/api/) OR add some description in the README.md explaining that the watch mode does not work for the API docs.

HyunSangHan:
### Description of the Change
As you can see issue #4060 , watch mode(`npm start docs.watch`) only works for [Mocha website](https://mochajs.org/)(made with `eleventy`), NOT for [API docs](https://mochajs.org/api/)(made with `jsdoc`) now, because it has been served through `eleventy --serve`.
​
Therefore, it is also necessary to be rebuilding process for API docs in watch mode. After I edited the script, when I use `npm start docs.watch`, the process is as follows:
1. Build for the first time.
2. Watch if there is some change in /docs directory, and Serve to the browser at the same time.
3. If there is some change in /docs directory, Rebuild and reflect it in the browser.
* NOTE : I used `chokidar-cli` to watch the docs directory and `concurrently` to watch and serve at the same time.
​
### Benefits
It can be convenient for developers who want to work with API docs to see code changes immediately.
​
### Possible Drawbacks
Until now, watch mode has been implemented for Mocha websites only, so if you only want to modify the website, `eleventy --serve` might be lighter.
​
### Applicable issues
Close #4060


package-scripts.js
```diff
@@ -299,8 +299,26 @@ module.exports = {
...
+          description: 'Build docs'
+        },
+        rebuild: {
+          script:
+            'chokidar "docs/**/*" -i "docs/_dist" -i "docs/_site" -c "nps docs.watch.build"',
+          description: 'Watch for changes and rebuild docs'
+        },
+        server: {
+          script:
+            'browser-sync start --server docs/_site --files "docs/_site/**/*"',
```
outsideris: Currently `browser-sync` included in the `eleventy` package. So, `eleventy` don't use `browser-sync` in future, this command will be broken.
We should include  `browser-sync` in our `devDependencies`.

HyunSangHan: Thank you for your advice! I installed it for `devDependencies` now.

HyunSangHan:
I resolved conflict.

HyunSangHan:
I resolved conflict.

juergba:
I don't think we really need this watch mode for our API docs. We would need two/three new dependencies for a little added value.
I'm working on Windows platform and I'm used to getting error messages when creating our documentation. And it still doesn't successfully run till the end. So I don't like the idea to make the docu process even more complex for only a small additional benefit.

HyunSangHan:
> I don't think we really need this watch mode for our API docs. We would need two/three new dependencies for a little added value.

@juergba 
I agree with you. The additional value seems too small for installing dependencies.

How about adding some description to `Development > Notes` of [`docs/README.md`](https://github.com/mochajs/mocha/tree/master/docs) like below? (Instead of adding scripts and installing dependencies)
- `npm start docs.watch` works not for [mochajs.org/api/](https://mochajs.org/api/) but only for [mochajs.org](https://mochajs.org), because Mocha's API document is built using a JSDoc template.

HyunSangHan:
@juergba 
I see. I think your opinion makes sense.
Thank you for your opinion. :)

PR #4063, opened 2019-10-12
special-case parsing of "require" in unparseNodeArgs(); closes #4035

### LINKED ISSUE #4035: 6.2.1 fails for node 8.x if there is a --require
### Prerequisites

### Description

6.2.0 the command below works

`NODE_ENV=test PORT=9999 node_modules/.bin/mocha -t 5000 --require esm --exit -b --recursive server/test/bootstrap.test.js`

with 6.2.1 if fails with this output on Node 8.x (NB: works on Node 10.x for both)

`bad option: --require=esm`

Node 8.x expects the require to be in this format `--require esm` i.e. no equals `=` sign

I can workaround this by running the underlying node more directly but wanted to let you know about this small regression.

### Steps to Reproduce

See description

See description

See description

### Versions

No global install and local is 6.2.1

- The output of `node --version`: v8.16.1

- Your operating system
  - name and version: Mac OS Mojave 10.14.6 also fails on CircleCI node 8 image which I think is Alpine 
- Any third-party Mocha-related modules (and their versions): None
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): none / ESM as a loader

boneskull:
this adds a special case when `unparseNodeArgs()` detects `require`; it removes the `=`, which is apparently required for everything else but prohibited in Node.js v8.

### DIFF
--- a/lib/cli/node-flags.js
+++ b/lib/cli/node-flags.js
@@ -68,6 +68,7 @@ exports.impliesNoTimeouts = flag => debugFlags.has(flag);
 /**
  * All non-strictly-boolean arguments to node--those with values--must specify those values using `=`, e.g., `--inspect=0.0.0.0`.
  * Unparse these arguments using `yargs-unparser` (which would result in `--inspect 0.0.0.0`), then supply `=` where we have values.
+ * Apparently --require in Node.js v8 does NOT want `=`.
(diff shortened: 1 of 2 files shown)


boneskull:
if this passes, I'm probably going to cut a release asap since I'm blocking on it

PR #4065, opened 2019-10-12
Docs: update browser options

juergba:
### Description of the Change

I tested some CLI options for support in browser context and updated the documentation.
This list of options available in browsers is not conclusive, yet.


juergba:
This list should contain only real browser options, which can be used within the "html" reporter.
Eg. option `color` does not have any effect in browser. 

CLI builtin reporters eg. "spec" can also be used in browser context, sending its output to the browser console. In this case `color` has an effect, but does not belong into this list.

Issue #980, opened 2013-09-20
🐛 Bug: Exception in a reporter silently crashes Mocha


rassie:
If a reporter raises an exception, Mocha just stops without any error message. An example reporter demonstrating the problem (needs a failing test in the suite):

```
module.exports = Crash;

var Spec = require('mocha/lib/reporters/spec');

function Crash(runner) {
    Spec.call(this, runner);

    runner.on('fail', function(test, err){
        throw new Error();
    });
}

Crash.prototype.__proto__ = Spec.prototype;
```

boneskull:
this would be helpful.

reesew:
+2
I'm not even sure how to get the stack trace

mark-wiemer:
It's been a very long time since this issue was reviewed, marking as "in triage" to confirm this is still the case before we accept PRs.

mostafaNazari702:
Reproduced on Mocha 12.0.2 / Node 24 (Windows 11).

crash-reporter.js:

```js
module.exports = function Crash(runner) {
  runner.on('fail', function () {
    throw new Error('boom from reporter');
  });
};
```
test.cjs: 

```
it('fails', function () {
throw new Error('test failure'); 
});
```

Run with: 
```
npx mocha --reporter ./crash-reporter.js test.cjs
```

We expect the failing test to be reported with exit 1 (the spec reporter gives `1 passing / 1 failing`, exit 1) but no output at all, exit 0.

https://github.com/mostafanazari702/mocha/actions/runs/35362730739

mark-wiemer:
@mostafaNazari702 I'm not quite sure how to read repros-via-runs, unfortunately. Here's my repro as a commit, these are easier to review for me: https://github.com/mochajs/mocha/commit/4b9847361f6389d6420865cca77c71a9ebc5ec2c

mostafaNazari702:
> [@mostafaNazari702](https://github.com/mostafaNazari702) I'm not quite sure how to read repros-via-runs, unfortunately. Here's my repro as a commit, these are easier to review for me: [4b98473](https://github.com/mochajs/mocha/commit/4b9847361f6389d6420865cca77c71a9ebc5ec2c)

Definitely fair because i was also a bit lost when i did that, here it is anyway:

<img width="2976" height="1221" alt="A screenshot of the job that shows the reproduction showing the result mentioned (exit 0)." src="https://github.com/user-attachments/assets/17be7a2d-5ae7-47d9-9ab3-fd9be655eb28" />

PR #4068, opened 2019-10-14
XUnit reporter should handle exceptions during diff generation

### LINKED ISSUE #4022: reporter xunit output report incomplete
### Prerequisites

- [X] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend that you _not_ install Mocha globally.

### Description

We are using hippie to test our components. During the request assertion, if the test fails, the output does not display the entire report. Not event the failed test nor the next tests.
```
<testsuite name="Mocha Tests" tests="3" failures="0" errors="1" skipped="0" timestamp="Tue, 17 Sep 2019 13:26:17 GMT" time="0.378">
<testcase classname="test" name="test ok" time="0.001"/>
// missing testsuite end tag.
```
### Steps to Reproduce

```js
"use strict";

const {expect} = require("chai");
const hippie = require("hippie");

describe("test", function() {
	it("test ok", function(){
		expect(true).to.be.equal(true);
	})
	it("test ko", function(){
		return hippie()
			.base("https://mochajs.org")
			.get("/not_found")
			.expectStatus(200)
			.end();
	})
	it("test ok", function(){
		expect(true).to.be.equal(true);
	})
})
```

Test report with spec reporter shows succesfully:

`$ npx mocha test.js --reporter spec`

```
    test
    ✓ test ok
    1) test ko
    ✓ test ok


  2 passing (204ms)
  1 failing

  1) test
       test ko:
     AssertionError: Status code

    Actual:   404
    Expected: 200

      at assert (node_modules/hippie/lib/hippie/assert.js:36:10)
      at statusCode (node_modules/hippie/lib/hippie/expect.js:23:10)
      at verify (node_modules/hippie/lib/hippie/client.js:476:5)
      at Client.verify (node_modules/hippie/lib/hippie/client.js:477:5)
      at /home/sbesson/GIT/TMP/node_modules/hippie/lib/hippie/client.js:437:12
      at Client.exports.raw [as parse] (node_modules/hippie/lib/hippie/parsers.js:33:3)
      at Request._callback (node_modules/hippie/lib/hippie/client.js:435:10)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.<anonymous> (node_modules/request/request.js:1157:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1079:12)
      at endReadableNT (_stream_readable.js:1129:12)
      at process._tickCallback (internal/process/next_tick.js:63:19)
```

But with xunit reporter:
`npx mocha test.js --reporter xunit`

```
<testsuite name="Mocha Tests" tests="3" failures="0" errors="1" skipped="0" timestamp="Tue, 17 Sep 2019 13:33:12 GMT" time="0.195">
<testcase classname="test" name="test ok" time="0.001"/>
```

We expect an output looking like this:
```
<testsuite name="Mocha Tests" tests="3" failures="0" errors="1" skipped="0" timestamp="Tue, 17 Sep 2019 13:37:42 GMT" time="0.251">
<testcase classname="test" name="test ok" time="0.001"/>
<testcase classname="test" name="test ko" time="0"><failure>expected 404 to equal 200

      + expected - actual

      -404
      +200
      
AssertionError: expected 404 to equal 200
    at /home/sbesson/GIT/TMP/test.js:16:31
    at verify (node_modules/hippie/lib/hippie/client.js:475:32)
    at Client.verify (node_modules/hippie/lib/hippie/client.js:477:5)
    at /home/sbesson/GIT/TMP/node_modules/hippie/lib/hippie/client.js:437:12
    at Client.exports.raw [as parse] (node_modules/hippie/lib/hippie/parsers.js:33:3)
    at Request._callback (node_modules/hippie/lib/hippie/client.js:435:10)
    at Request.self.callback (node_modules/request/request.js:185:22)
    at Request.&#x3C;anonymous&#x3E; (node_modules/request/request.js:1157:10)
    at IncomingMessage.&#x3C;anonymous&#x3E; (node_modules/request/request.js:1079:12)
    at endReadableNT (_stream_readable.js:1129:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)</failure></testcase>
<testcase classname="test" name="test ok" time="0"/>
</testsuite>
```

```
<testsuite name="Mocha Tests" tests="3" failures="0" errors="1" skipped="0" timestamp="Tue, 17 Sep 2019 13:33:12 GMT" time="0.195">
<testcase classname="test" name="test ok" time="0.001"/>
```

100%

### Versions

* node : 10.16.0
* mocha : 6.2.0
* hippie : 0.5.2

* OS : Linux Mint 19.1 Tessa
* architecture : 64-bit
* shell : bash

### Additional Information

It works just fine with mocha@5.2.0.

Best regards.

rgroothuijsen:
### Description of the Change

This bugfix addresses a crash in the XUnit reporter, which produced no output because the assertion library used in the ticket provided numerical diff values rather than string values, causing an error during diff generation. To safeguard against unexpected crashes such as these, the same input-checks that the Base reporter uses have been implemented in the XUnit reporter, and exceptions during XUnit diff generation are returned by the reporter.

### Applicable issues

Fixes #4022.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -154,14 +154,14 @@ exports.cursor = {
   }
 };
 
-function showDiff(err) {
+var showDiff = (exports.showDiff = function(err) {
   return (
     err &&
     err.showDiff !== false &&
     sameType(err.actual, err.expected) &&
     err.expected !== undefined
   );
-}
+});
 
 function stringifyDiffObjs(err) {
   if (!utils.isString(err.actual) || !utils.isString(err.expected)) {
@@ -182,9 +182,19 @@ function stringifyDiffObjs(err) {
  * @return {string} Diff
  */
 var generateDiff = (exports.generateDiff = function(actual, expected) {
-  return exports.inlineDiffs
-    ? inlineDiff(actual, expected)
-    : unifiedDiff(actual, expected);
+  try {
+    return exports.inlineDiffs
+      ? inlineDiff(actual, expected)
+      : unifiedDiff(actual, expected);
+  } catch (err) {
+    var msg =
+      '\n      ' +
+      color('diff added', '+ expected') +
+      ' ' +
+      color('diff removed', '- actual:  failed to generate Mocha diff') +
+      '\n';
+    return msg;
+  }
 });
 
 /**
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -163,9 +163,9 @@ XUnit.prototype.test = function(test) {
   if (test.state === STATE_FAILED) {
     var err = test.err;
     var diff =
-      Base.hideDiff || !err.actual || !err.expected
-        ? ''
-        : '\n' + Base.generateDiff(err.actual, err.expected);
+      !Base.hideDiff && Base.showDiff(err)
+        ? '\n' + Base.generateDiff(err.actual, err.expected)
+        : '';
     this.write(
       tag(
         'testcase',
--- a/test/reporters/xunit.spec.js
+++ b/test/reporters/xunit.spec.js
@@ -350,6 +350,42 @@ describe('XUnit reporter', function() {
           '</failure></testcase>';
         expect(expectedWrite, 'to be', expectedTag);
       });
+
+      it('should handle non-string diff values', function() {
+        var runner = new EventEmitter();
+        createStatsCollector(runner);
+        var xunit = new XUnit(runner);
+
+        var expectedTest = {
+          state: STATE_FAILED,
+          title: expectedTitle,
+          parent: {
+            fullTitle: function() {
+              return expectedClassName;
+            }
+          },
+          duration: 1000,
+          err: {
+            actual: 1,
+            expected: 2,
+            message: expectedMessage,
+            stack: expectedStack
+          }
+        };
+
+        sandbox.stub(xunit, 'write').callsFake(function(str) {
+          expectedWrite += str;
+        });
+
+        runner.emit(EVENT_TEST_FAIL, expectedTest, expectedTest.err);
+        runner.emit(EVENT_RUN_END);
+        sandbox.restore();
+
+        var expectedDiff =
+          '\n      + expected - actual\n\n      -1\n      +2\n      ';
+
+        expect(expectedWrite, 'to contain', expectedDiff);
+      });
     });
 
     describe('on test pending', function() {


craigtaub:
This LGTM. @mochajs/core any other thoughts?

juergba:
When this exception is thrown in XUnit reporter, the output is truncated and aborted without any error message. That exception shouldn't be swallowed. 

Could be related to #980.

rgroothuijsen:
@juergba I'm not sure if I understood entirely. Is this particular fix incorrect? Is there something I should add/modify?

juergba:
This issue has two parts. The diff exception and then second the reporter aborting without any message.
It would be helpful to know the cause of the second part in order to decide wether this PR is complete.

rgroothuijsen:
@juergba Uncaught exceptions are intercepted by the runner at [this point](https://github.com/mochajs/mocha/blob/master/lib/runner.js#L938). Looking at [the method in question](https://github.com/mochajs/mocha/blob/master/lib/runner.js#L793), I see that the exception is logged, but only at the debug log level.

lib/reporters/base.js
```diff
@@ -182,9 +182,11 @@ function stringifyDiffObjs(err) {
  * @return {string} Diff
  */
 var generateDiff = (exports.generateDiff = function(actual, expected) {
+  var err = {actual: actual, expected: expected};
```
juergba: Is this line IE11 compatible?

rgroothuijsen: It should be, as according to [this chart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Browser_compatibility) IE supports object initializers.

juergba: Please revert your changes in this function and insert `try/catch`.
An exception in generating diff is not a fatal error, so I propose to catch/continue and just print a message. For example:
```js
try {
    return exports.inlineDiffs
      ? inlineDiff(actual, expected)
      : unifiedDiff(actual, expected);
  } catch (err) {
    var msg =
      '\n      ' +
      colorLines('diff added', '+ expected') +
      ' ' +
      colorLines('diff removed', '- actual:  failed to generate Mocha diff') +
      '\n';
    return msg;
  }
```

lib/reporters/xunit.js
```diff
@@ -163,7 +163,7 @@ XUnit.prototype.test = function(test) {
   if (test.state === STATE_FAILED) {
     var err = test.err;
     var diff =
-      Base.hideDiff || !err.actual || !err.expected
+      !Base.showDiff(err) || Base.hideDiff || !err.actual || !err.expected
         ? ''
         : '\n' + Base.generateDiff(err.actual, err.expected);
```
juergba: Please use the identical condition of `Base`.
```js
  !Base.hideDiff && Base.showDiff(err)
        ? '\n' + Base.generateDiff(err.actual, err.expected)
        : '';
```

test/reporters/xunit.spec.js
```diff
@@ -350,6 +350,47 @@ describe('XUnit reporter', function() {
...
+            stack: expectedStack
+          }
+        };
+
+        xunit.test.call(fakeThis, expectedTest);
+        sandbox.restore();
+
+        var expectedNumberDiff = expectedDiff
+          .replace('foo', 1)
+          .replace('bar', 2);
```
juergba: Please use `var expectedDiff = '\n      + expected - actual\n\n      -1\n      +2\n      ';`

lib/reporters/base.js
```diff
@@ -154,21 +154,21 @@ exports.cursor = {
...
+});
 
-function stringifyDiffObjs(err) {
+var stringifyDiffObjs = (exports.stringifyDiffObjs = function(err) {
   if (!utils.isString(err.actual) || !utils.isString(err.expected)) {
     err.actual = utils.stringify(err.actual);
     err.expected = utils.stringify(err.expected);
   }
-}
+});
```
juergba: Why do you export this function?

lib/reporters/base.js
```diff
@@ -182,9 +182,19 @@ function stringifyDiffObjs(err) {
...
-    ? inlineDiff(actual, expected)
-    : unifiedDiff(actual, expected);
+  try {
+    return exports.inlineDiffs
+      ? inlineDiff(actual, expected)
+      : unifiedDiff(actual, expected);
+  } catch (err) {
+    var msg =
+      '\n      ' +
+      colorLines('diff added', '+ expected') +
```
juergba: replace colorLines with color.

lib/reporters/base.js
```diff
@@ -182,9 +182,19 @@ function stringifyDiffObjs(err) {
...
+      ' ' +
+      colorLines('diff removed', '- actual:  failed to generate Mocha diff') +
```
juergba: dito

lib/reporters/xunit.js
```diff
@@ -162,10 +162,15 @@ XUnit.prototype.test = function(test) {
 
   if (test.state === STATE_FAILED) {
     var err = test.err;
+
+    if (Base.showDiff(err)) {
+      Base.stringifyDiffObjs(err);
+    }
```
juergba: `err` has already been stringified in the `EVENT_TEST_FAIL` handler. Is their a reason you want to call it a second time?
I guess you want to make pass your test, but this is a bad way to achieve this.

test/reporters/xunit.spec.js
```diff
@@ -350,6 +350,46 @@ describe('XUnit reporter', function() {
           '</failure></testcase>';
         expect(expectedWrite, 'to be', expectedTag);
       });
+
+      it('should handle non-string diff values', function() {
```
juergba: You have to emit an EVENT_TEST_FAIL to run this test successfully.
var runner = new EventEmitter();

test/reporters/xunit.spec.js
```diff
@@ -350,6 +350,46 @@ describe('XUnit reporter', function() {
...
+          },
+          duration: 1000,
+          err: {
+            actual: 1,
+            expected: 2,
+            message: expectedMessage,
+            stack: expectedStack
+          }
+        };
+
```
juergba: `runner.emit(EVENT_TEST_FAIL, expectedTest, expectedTest.err);`

lib/reporters/xunit.js
rgroothuijsen: @juergba You're absolutely right, turns out I've been barking up the wrong tree because my test was wrong, thus misunderstanding what was happening. Very silly of me to not notice. :sweat_smile:  I'll revert my previous export of stringifyDiffObjs.

rgroothuijsen:
@juergba Will do. In retrospect, do you think the title still works for this PR?

juergba:
@rgroothuijsen No, not really. Maybe "XUnit reporter: fix crash during diff generation" or similar, but the title is upto you.

PR #4073, opened 2019-10-18
add --overspec option to disable promise checking on tests that take a callback

sixcircuit:
Great library. Thanks a ton.

### Description of the Change

Currently if you take a `done` callback in a test and also return a promise, even if the test passes you get an error "Overspecified test, return a callback or a promise, not both."

This change adds an option (and relevant test coverage) to turn off the error. The option is off by default (leaving the status quo) and gives more sophisticated users an option to avoid what's basically protection for people who may make a mistake.

This is particularly a hassle because if you mark a function as `async` you can't avoid returning a promise (see below). If you're migrating a legacy codebase, this mix is inevitable, and the error stops you from doing perfectly reasonable, safe and legal things, in order to protect people from themselves. Which is a good idea, but I'd like the option to turn it off. It feels like a situation where an option or a warning would probably have been preferable in the first place.

Built it for myself, figured it may help others. Happy to make changes if you want them. Also cool if this doesn't fit the philosophy for some reason.

``` 
test('some test', async function(done){
  const foo = await asyncFunction();
  
  someOtherAsync(foo, function(err, bar){
    if(err){ return done(err); }
    assert(bar);
    done();
  });
});
```

### Alternate Designs

I thought about arguing that anyone taking a `done` callback knows what they're doing, and throwing an error to try to protect them from themselves hurts more sophisticated users. I was going to pull out the error altogether. Breaking existing functionality seemed like a bad idea, so I figured an option that was off by default leaves the status quo, and gives experts an option to override the "warning" and probably doesn't ruffle too many feathers.

I could have turned it on by default (bad idea breaking existing functionality). I could have not built it at all (not really feasible for my use case).

Can't avoid it, it's integral to how the tests are run. You throw an error on tests that would otherwise pass, any function that is marked async will need to be rewritten. You're hurting expert users who know what they're doing.

I'm giving people who know what they're doing the option to avoid an error (that probably should have been a warning or an option in the first place).

### Benefits

Allows tests that take a done callback to be marked `async` and still run. This switch is off by default so it doesn't change existing functionality. Allows people who know what they're doing to make use of more elegant test patterns. Allows for projects to be migrated slowly to async/await without having to rewrite all the tests at once.

Very helpful in large codebase legacy scenarios (like mine).

### Possible Drawbacks

Slight increase in maintenance going forward. May not get a ton of use.

### Applicable issues

Patch release. Totally backwards compatible and off by default.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -835,6 +835,8 @@ Rules & Behavior
   --allow-uncaught           Allow uncaught errors to propagate        [boolean]
   --async-only, -A           Require all tests to use a callback (async) or
                              return a Promise                          [boolean]
+  --overspec                 Allow overspec'd tests. Disable promise checking on
+                             callback style tests.                     [boolean]
   --bail, -b                 Abort ("bail") after first test failure   [boolean]
   --check-leaks              Check for global variable leaks           [boolean]
   --delay                    Delay initial execution of root suite     [boolean]
@@ -916,6 +918,10 @@ This flag is useful when debugging particularly difficult-to-track exceptions.
 ### `--async-only, -A`
 
 Enforce a rule that tests must be written in "async" style, meaning each test provides a `done` callback or returns a `Promise`. Non-compliant tests will be marked as failures.
+
+### `--overspec`
+
+Allow overspec'd tests. Disable checking for promises returned from tests that have an arity of one (that take a `done` callback function). Allows use of `done` callback in tests that are defined as `async function`. Useful for mixing async await and callback style tests.
 
 ### `--bail, -b`
 
--- a/lib/cli/run-option-metadata.js
+++ b/lib/cli/run-option-metadata.js
@@ -27,6 +27,7 @@ exports.types = {
   boolean: [
     'allow-uncaught',
     'async-only',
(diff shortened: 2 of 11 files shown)


boneskull:
Thanks for the PR!

The language used in the message is... not the whole story.  The issue is that Mocha _has no idea_ when your test is finished.

That said, your particular example can be addressed by consuming [util.promisify](https://nodejs.org/api/util.html#util_util_promisify_original) or just doing it manually:

```js
test('some test', async function() {
  const foo = await asyncFunction();
  
  return new Promise((resolve, reject) => {
    someOtherAsync(foo, (err, bar) => {
      if(err){ return reject(err); }
      assert(bar); // also rejects upon failure
      resolve();
  });
});
```

This is inconvenient, but disabling the error message is not the "fix"--the fix would be for Mocha to understand that it needs to finish the test when _both_ a) `done` is called and b) the returned `Promise` is fulfilled--and considering `done(err)` OR a rejected `Promise` to be a failure.

That said, I have yet to see an example of a case that has no workaround.  `async` event emitter listeners seem to be the most awkward to deal with...

PR #4085, opened 2019-10-26
Support util.inspect.custom with fixing #3985

KRAhn:
### Requirements

### Description of the Change

There are two sub issues in the issue https://github.com/mochajs/mocha/issues/3985
1. Fixing a sequence of error generator of `lib/reporters/base.js`
- As Is
1 - message
2 - inspect()
3 - empty string
- To be
1 - inspect()
2 - message
3 - empty string

2. Supporting `util.inspect.custom()`, which seems standard of nodejs for representing an object for debugging purpose, and use it first when generating an error message. (https://nodejs.org/api/util.html#util_util_inspect_custom)

This PR resolves those issues.

### Benefits

Making developers have more autonomy when they make error message. 

### Possible Drawbacks
Fragmentation of a way to generate error message.
I think we can take it because `util.inspect.custom` is general way to make representative string of object.
But hope no fragmentation anymore.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -9,6 +9,7 @@
 var tty = require('tty');
 var diff = require('diff');
 var milliseconds = require('ms');
+var util = require('util');
 var utils = require('../utils');
 var supportsColor = process.browser ? null : require('supports-color');
 var constants = require('../runner').constants;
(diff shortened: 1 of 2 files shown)


KRAhn:
I’m gonna reopen it after fixing build issue.

PR #4089, opened 2019-11-07
Add funding information to package.json

Munter:
See https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md


boneskull:
@Munter please use the branch format `munter/<branchname>` in the future to avoid (more) confusion

PR #4115, opened 2019-11-30
Update eslint parsing from ES6 to ES2017

juergba:
### Description of the Change

Some modules of our code base are not used by Mocha's browser environment.
See also `env: browser: no` in `eslintrc.yml`
Currently these modules are eslint-parsed by ES6 rules. 
- we update the eslint parsingOption from ES6 to ES2017 in order to make available `async`/`await`.
The default remains ES5 for IE11 reasons.
As soon as we drop Node v8 support (end of life: 2019-12-31) we should update to ES2018.

Browserify: regarding ignored files
We have three different configuration settings defining files to be ignored. I guess we have some redundant information here, but I'm unsure about that. To stay on the safe side we update all of them:
- `browser` field in package.json
- `ignore()` in `karma.config.js`
- the browserify script with its `--ignore` option

It's unclear to me why we ignore eg. chokidar while `browserify`ing, but not packages like yargs, yargs-parser etc. `browser-entry.js` could be an explanation since it only requires "./lib/mocha" and none of our "CLI" dependencies. Anyway this topic is not part of this PR.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -29,7 +29,7 @@ overrides:
       - test/integration/helpers.js
       - lib/growl.js
     parserOptions:
-      ecmaVersion: 6
+      ecmaVersion: 2017
     env:
       browser: no
 
@@ -77,7 +77,7 @@ overrides:
   - files:
       - test/**/*.mjs
     parserOptions:
-      ecmaVersion: 6
+      ecmaVersion: 2017
       sourceType: module
 
   - files:
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -31,8 +31,10 @@ module.exports = config => {
     browserify: {
       debug: true,
       configure: function configure(b) {
-        b.ignore('glob')
+        b.ignore('./lib/cli/*.js')
+          .ignore('chokidar')
           .ignore('fs')
+          .ignore('glob')
           .ignore('path')
           .ignore('supports-color')
           .on('bundled', (err, content) => {
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -23,7 +23,7 @@ function test(testName, mochaParams) {
 module.exports = {
   scripts: {
     build: {
-      script: `browserify -e browser-entry.js --plugin ./scripts/dedefine --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' --ignore chokidar -o mocha.js`,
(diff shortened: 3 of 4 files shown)


juergba:
@giltayar I decided for ES2017 instead of ES2018. Can you live with this.

juergba:
What about yargs, yargs-parser, yargs-unparser ... Why aren't they excluded?

giltayar:
@juergba - of course I can! So `async/await` will be available in _all_ our files? That's great. I will be able to remove that weird `esm-utils.js` file in https://github.com/mochajs/mocha/pull/4038.

juergba:
No, not in all! In "lib/cli/*.js".

giltayar:
@juergba - hmm... wait, I'm not sure. I use `import(...)` to dynamically load the modules, and that is `es2018`. Again, I can continue with the `esm-utils` file to specifically use `es2018` on it, because that is used only when ESM support is needed, so we _know_ that we're using Node>=10 and so support es2018 when running.

juergba:
Does ES2017 help you in any way then?

giltayar:
Actually, no. I can survive with Promise-only code. There's not a _lot_ of async there.

But it would be nice to have, given that that path of code is now an async path.

juergba:
End of life of Node v8 will be in one month.
We should plan to publish your esm stuff after Mocha v7.0.0 and after dropping Node v8. At that time it would be ES2018. This would help, right?

juergba:
Or you could silence the eslinter locally for your dynamic `import(...)`. It's just one line?

giltayar:
Yup. Makes everything simpler: now that I _do_ have async/await, and the only reason to keep 2018 is that one line, then I can disable linting on that line.

Good idea. I can go for it.

giltayar:
@juergba - I'm implementing the change in https://github.com/mochajs/mocha/pull/4038, and I just noticed that `mocha.js` did not enter the list of files that are allowed to use `async/await`. Understandably so, because it's used by both node code and browser code.

This means that I won't really be able to get rid of `esm-utils.js`, which means I can still keep the exception for that file to use `es2018` (as it uses `import()`).

There's another reason, an important _runtime_ one, to keep `esm-utils`: `import()`, even if it looks like it, is not a _function_, but rather a keyword, so versions of Node.js that do not support ESM fail when loading that module in CJS (or ESM in Node v8/10). This is why I conditionally `require` it in `mocha.js`. So yet another reason to keep it in a separate file.

PR #4118, opened 2019-12-04
Update node-environment-flags to 1.0.6

kylef:
### Description of the Change

node-environment-flags 1.0.5 and lower do not contain a LICENSE file which can make compliance difficult. This pull request updates the patch release of node-environment-flags to incorporate https://github.com/boneskull/node-environment-flags/pull/5.

### Alternate Designs

Removing the dependency on node-environment-flags.

To ensure the entire dependency chain of Mocha has dependencies contains valid copyright and licenses.

### Possible Drawbacks

I don't think there are any.

### Applicable issues

> * Mocha follows semantic versioning: http://semver.org
> 
> * Is this a breaking change (major release)?
> * Is it an enhancement (minor release)?
> * Is it a bug fix, or does it not impact production code (patch release)?

It's a bug fix, fixing licensing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -537,7 +537,7 @@
     "minimatch": "3.0.4",
     "mkdirp": "0.5.1",
     "ms": "2.1.1",
-    "node-environment-flags": "1.0.5",
+    "node-environment-flags": "1.0.6",
     "object.assign": "4.1.0",
     "strip-json-comments": "2.0.1",
     "supports-color": "6.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -11374,18 +11374,18 @@
       }
     },
     "node-environment-flags": {
-      "version": "1.0.5",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -537,7 +537,7 @@
     "minimatch": "3.0.4",
     "mkdirp": "0.5.1",
     "ms": "2.1.1",
-    "node-environment-flags": "1.0.5",
+    "node-environment-flags": "^1.0.6",
```
juergba: Please remove the "^", thanks.

kylef: Perhaps worth configing the `save-prefix` in a project level npmrc so that by default `npm install --save` behaves as intended with project rules?

See https://docs.npm.red/misc/config#save-prefix

kylef: Updated in 5faba2c3

PR #4122, opened 2019-12-06
Export yargs options and check function for consumption by wrapper modules

Jack-Barry:
# Description of the Change

Exports the Yargs `options` and `check` function that `mocha` uses to allow wrapper modules to consume them. This is accomplished by assigning the `options` to an exportable object and the exporting the `check` function before they are used by `mocha` internally.

# Alternate Designs

Without exporting the `options` and `check`, wrapper modules are stuck re-implementing the `options` and `check` themselves, which stinks 💩

# Why should this be in core?

This would provide external modules what they need to avoid reinventing the wheel, and doesn't break or disrupt anything here.

# Benefits

Allows us to implement a fix for [this](https://github.com/sysgears/mochapack/issues/34)

# Possible Drawbacks

I can't think of any.

# Applicable issues

[Aforementioned issue](https://github.com/sysgears/mochapack/issues/34)

* [ ] Is this a breaking change (major release)?
* [x] Is it an enhancement (minor release)?
* [ ] Is it a bug fix, or does it not impact production code (patch release)?

### DIFF
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -245,6 +245,47 @@ exports.options = {
   }
 };
 
+exports.check = yargs =>
+  function(argv) {
+    // "one-and-dones"; let yargs handle help and version
+    Object.keys(ONE_AND_DONES).forEach(opt => {
+      if (argv[opt]) {
+        ONE_AND_DONES[opt].call(null, yargs);
+        process.exit();
+      }
+    });
+
+    // yargs.implies() isn't flexible enough to handle this
+    if (argv.invert && !('fgrep' in argv || 'grep' in argv)) {
+      throw createMissingArgumentError(
+        '"--invert" requires one of "--fgrep <str>" or "--grep <regexp>"',
+        '--fgrep|--grep',
+        'string|regexp'
+      );
+    }
+
+    if (argv.compilers) {
+      throw createUnsupportedError(
+        `--compilers is DEPRECATED and no longer supported.
+          See https://git.io/vdcSr for migration information.`
+      );
+    }
+
+    if (argv.opts) {
+      throw createUnsupportedError(
+        `--opts: configuring Mocha via 'mocha.opts' is DEPRECATED and no longer supported.
+          Please use a configuration file instead.`
+      );
+    }
+
+    // load requires first, because it can impact "plugin" validation
+    handleRequires(argv.require);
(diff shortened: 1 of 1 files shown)


larixer:
@Jack-Barry Looks good to me

Jack-Barry:
👋🏻 Just wanted to check in on this to see if it can be merged, would be a huge help to some changes needed on our end. Thanks!

JamesMcMahon:
Would love to see this merged as it would be a great boon to https://github.com/sysgears/mochapack/.

rfftrifork:
Wouldn't it make sense to get this feature into mocha?

Jack-Barry:
Incorporated latest changes from Mocha's master branch before resolving the conflict in current version of `run.js`.

juergba:
@Jack-Barry I don't know wether this PR makes sense or not.
But this branch is a complete mess now after "Incorporated latest changes" with 44 commits and 99 files changed. Must be the corona virus, I would never merge this PR in that shape.
I recommend a clean rebase to `master`, then a forced push.

Jack-Barry:
@juergba Apologies for the messiness, I am still new to contributing to other peoples' projects and using Git features such as rebase. I have tried to clean up this PR as you suggested. I also added another change we'd need to make this a little better on our end to be able to use Mocha's `check` functionality.

If there is another way that Mocha can be initiated from a tool that wraps it in their own CLI aside from having to re-implement all of the Yargs options, I'm open to that alternative path. Otherwise, this is the best way we could come up with to allow our project to minimize re-implementing features already written in the Mocha codebase.

It allows us to parse Mocha-targeted args with a simpler implementation like this:

```ts
import { options, check } from 'mocha/lib/cli/run'
import { aliases, types } from 'mocha/lib/cli/run-option-metadata'

export const parseMochaArgs = (argv) =>
  yargs
    .options(options)
    .check(check(yargs))
    .alias(aliases)
    .array(types.array)
    .boolean(types.boolean)
    .string(types.string)
    .number(types.number)
    .parse(argv)
```

As opposed to re-implementing all of the `options` and the `check` functionality. In an ideal world, we would just have a `parseMochaArgs` function from Mocha that spits out the parsed arguments completely, but this seemed like a good start at least as it doesn't really require any further maintenance overhead for the Mocha team.

PR #4123, opened 2019-12-08
Add option to forbid empty test suites

### LINKED ISSUE #4062: Add option to fail if no tests are run

At the moment, if `mocha.run()` is called without any tests having been defined, the result is considered a success as the number of failures is 0. In messageformat/messageformat#261 this is causing a problem when Selenium browser tests are being [run in a CI environment](https://travis-ci.org/messageformat/messageformat/jobs/553023787) using [mocha-selenium-bridge](https://github.com/eemeli/mocha-selenium-bridge), and the tests "pass" (note the IE 11 results):

```
▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬
Testing Chrome 49.0...
  83 passing (179ms)

    ✓ Chrome 49.0 (8602ms)

▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬
Testing Firefox 66.0...
  83 passing (358ms)

    ✓ Firefox 66.0 (10933ms)

▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬
Testing IE 11.0...
  0 passing (0ms)

    ✓ IE 11.0 (6062ms)
```

In our case, this went unnoticed for a number of months as the result was marked as a success. The bug itself was caused by some unsupported-by-IE11 syntax in a test file.

I would like to have an option to consider "0 passing" tests to be considered an error.

I could probably write a custom reporter that would catch this, but that seems like a rather clumsy solution. I'm pretty sure the proper place to fix this would be somewhere [around here](https://github.com/mochajs/mocha/blob/3633a9062458d30d425d743bd0df65c1815b2908/lib/runner.js#L903), but the details of course depend on the implementation.

**Additional context**
I'm willing to submit a PR for this, but I'd rather get feedback first on how e.g. the error code should be handled, and/or whether a separate event would be appropriate for reporters to customise.

eemeli:
### Description of the Change

Adds an option `forbidEmptySuite`, available in the CLI as `--forbid-empty-suite`, that will fail a suite with an error `Empty suite forbidden` if it contains no tests.

### Alternate Designs

A similar result could be achieved by an individual customised reporter, but that seems like the wrong layer at which to address this.

Because the other options are also defined in the core.

### Benefits

At the moment, if the test definitions fail in a context that does not fail a subsequent `mocha.run()` (e.g. a browser processing multiple `<script>` tags), this failure may not be caught by Mocha, and instead it will erroneously display a positive result with "0 passing" tests. This is problematic in particular in CI tests, where the results are often left unchecked if they "pass".

### Possible Drawbacks

This adds yet another option.

### Applicable issues

Fixes #4062

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -732,6 +732,9 @@ Runner.prototype.runSuite = function(suite, fn) {
 
   debug('run suite %s', suite.fullTitle());
 
+  if (!total && this.forbidEmptySuite) {
+    this.fail(suite, new Error('Empty suite forbidden'));
+  }
   if (!total || (self.failures && suite._bail)) {
     return fn();
   }
--- a/docs/index.md
+++ b/docs/index.md
@@ -840,6 +840,7 @@ Rules & Behavior
   --check-leaks              Check for global variable leaks           [boolean]
   --delay                    Delay initial execution of root suite     [boolean]
   --exit                     Force Mocha to quit after tests complete  [boolean]
+  --forbid-empty-suite       Fail if a test suite contains no tests    [boolean]
   --forbid-only              Fail if exclusive test(s) encountered     [boolean]
   --forbid-pending           Fail if pending test(s) encountered       [boolean]
   --global, --globals        List of allowed global variables            [array]
@@ -953,6 +954,12 @@ To ensure your tests aren't leaving messes around, here are some ideas to get st
 - Use the new [`async_hooks`][node-async-hooks] API ([example][gist-async-hooks])
 - Try something like [wtfnode][npm-wtfnode]
 - Use [`.only`](#exclusive-tests) until you find the test that causes Mocha to hang
+
+### `--forbid-empty-suite`
+
+Enforce a rule that test suites must define at least one test, either directly or in an inner suite.
+
+`--forbid-empty-suite` causes Mocha to fail when an empty suite is encountered.
 
 ### `--forbid-only`
 
--- a/lib/cli/run-option-metadata.js
+++ b/lib/cli/run-option-metadata.js
@@ -33,6 +33,7 @@ exports.types = {
     'delay',
     'diff',
     'exit',
+    'forbid-empty-suite',
     'forbid-only',
     'forbid-pending',
     'full-trace',
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -106,6 +106,10 @@ exports.builder = yargs =>
         normalize: true,
         requiresArg: true
       },
+      'forbid-empty-suite': {
+        description: 'Fail if a test suite contains no tests',
+        group: GROUPS.RULES
+      },
       'forbid-only': {
         description: 'Fail if exclusive test(s) encountered',
         group: GROUPS.RULES
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -75,6 +75,7 @@ exports.Test = require('./test');
  * @param {boolean} [options.delay] - Delay root suite execution?
  * @param {boolean} [options.diff] - Show diff on failure?
  * @param {string} [options.fgrep] - Test filter given string.
+ * @param {boolean} [options.forbidEmptySuite] - Require at least one test.
  * @param {boolean} [options.forbidOnly] - Tests marked `only` fail the suite?
  * @param {boolean} [options.forbidPending] - Pending tests fail the suite?
  * @param {boolean} [options.fullTrace] - Full stacktrace upon failure?
@@ -125,6 +126,7 @@ function Mocha(options) {
     'color',
     'delay',
     'diff',
+    'forbidEmptySuite',
     'forbidOnly',
     'forbidPending',
     'fullTrace',
@@ -829,6 +831,21 @@ Mocha.prototype.delay = function delay() {
 };
 
 /**
+ * Causes running a suite with no tests to fail it.
+ *
+ * @public
+ * @see [CLI option](../#-forbid-empty-suite)
+ * @param {boolean} [forbidEmptySuite=true] - Whether each suite is required to
+ *   define at least one test.
+ * @returns {Mocha} this
+ * @chainable
+ */
+Mocha.prototype.forbidEmptySuite = function(forbidEmptySuite) {
+  this.options.forbidEmptySuite = forbidEmptySuite !== false;
+  return this;
+};
+
+/**
  * Causes tests marked `only` to fail the suite.
  *
  * @public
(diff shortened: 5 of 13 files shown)


juergba:
@eemeli I apalogize for not responding to your PR. Are you still interested?
I would like to review and push it through now.

eemeli:
Yes, I'm still interested in this. Happy to have this reviewed and hopefully merged!

juergba:
What about `pending` tests?
- static: `it.skip()`
- conditionally at runtime: `before(function() {   this.skip();  });` 

We could end up with zero tests run, but the script would still pass.
If we want to cover above scenario, we would have to check the `stats-collector` somehow, at `runner`'s end for the total of tests run.

Some users also create tests dynamically at runtime, when the parsing/creating of the test runner has already been completed.

eemeli:
I would argue that a suite with a skipped test is not "empty", and that the right thing in that case is not to complain. Sure, skipping all tests would therefore result in a suite passing even if none of the tests run, but that'd be intentional in this case.

This logic also allows for using `forbidEmptySuite: true` together with `it.only`, which would otherwise fail if there are multiple suites.

eemeli:
Regarding dynamically generated tests, that's an entirely valid point. I don't think the proposed behaviour would be very surprising in that case, but that should probably still be accounted for. Are there other places than `before()` that might dynamically add tests in an otherwise empty suite? And would the right thing then be to fail before or after the suite-end event is emitted?

juergba:
I don't understand this one:
> This logic also allows for using forbidEmptySuite: true together with it.only [...]

With dynamically generated tests I was refering to scenarios like #4126 etc. You can do this in `it()` and any hook type, but hooks need at least one test in order to run. So the suite wouldn't be empty anyway. 

> And would the right thing then be to fail before or after the suite-end event is emitted?

At suite-end event, I guess.

In your issue you wrote:

> I would like to have an option to consider "0 passing" tests to be considered an error.

In this PR you switched to checking each and every suite, instead of a final test at `runner`'s end. What is your reason?

I'm not sure yet, wether we should go your way, or check once at `runner`'s end.

eemeli:
Perhaps I should rephrase the option to consider it an error if a suite "defines" no tests, rather than "containing" no tests? That might make it altogether clearer.

Regarding `it.only`, I mean the case of having two parallel suites (say, A and B) each of which defines some tests. If one test in B is marked as `it.only`, all of the tests in suite A are skipped. With the current formulation, running those test suites with `--forbid-empty-suite` will pass, but if it's changed to require each suite to contain at least one test that's run, it'll fail.

> With dynamically generated tests I was refering to scenarios like #4126 etc. You can do this in `it()` and any hook type, but hooks need at least one test in order to run. So the suite wouldn't be empty anyway.

Expect for the `before()` hook, right? As far as I can tell, that's normally run even for an empty suite. So it provides a way for an initially empty suite to end up with some tests. And there's probably all sorts of other places, if you're adding tests to other suites than the current one.

> In this PR you switched to checking each and every suite, instead of a final test at `runner`'s end. What is your reason?

Three reasons:
1. Prefer to fail earlier rather than later
2. If suite A contains only suite B, which in turn contains only suite C, make sure that we only throw one error.
3. I presumed that the emptiness of a suite could be determined before it's even run

I'll amend and rebase this PR on top of the latest master, to account for dynamically added tests.

eemeli:
On an entirely different note, I rather appreciate the timing of pointing at an issue filed on the last Friday the 13th, on a Friday the 13th.

juergba:
Yes, and now the corona virus. Will there be any survivors left for using your new feature?

> Expect for the before() hook, right? As far as I can tell, that's normally run even for an empty suite.

No, a `before()` shouldn't run without at least one test.

eemeli:
Well, haven't we all heard recently that testing in particular is rather important just now?

> No, a `before()` shouldn't run without at least one test.

You're right, that was a misunderstanding on my part.

I added a test for a dynamically-defined test in an otherwise empty suite, and it turns out that the current approach already handles [that case](https://github.com/mochajs/mocha/pull/4123/files#diff-c66a17cc5e6589170f15c55ee5b24117) just like it should.

And then I realised that I can use the already-calculated `total` from `suite.grepTotal` to determine if a suite has no tests, and dropped the `suite.isEmpty` method. This means that the complete change required to `runner.runSuite` is [this if statement](https://github.com/mochajs/mocha/pull/4123/files#diff-b3b53682a18f203ac8d29b0e277cad26R735-R737).

And therefore I didn't move the test to the suite-end, because the existing behaviour after those lines prevents the suite-begin and suite-end events from being emitted if the suite is empty, and there's no reason to change the existing behaviour.

I also clarified the docs a little bit, and rebased the branch on the current `master`.

juergba:
@eemeli I will review this weekend.

lib/runner.js
```diff
@@ -732,6 +732,9 @@ Runner.prototype.runSuite = function(suite, fn) {
 
   debug('run suite %s', suite.fullTitle());
 
+  if (!total && this.forbidEmptySuite) {
+    this.fail(suite, new Error('Empty suite forbidden'));
```
juergba: `Suite` does not extend `Runnable` (as does `Test` and `Hook`), so you can't `fail` a `Suite`.
The reporter output is weird, if the test file contains only an empty suite.
Maybe a dummy test?

boneskull:
@eemeli Are you still interested in working on this?

boneskull:
FWIW I think this is a reasonable addition.  It's not something that can be discovered via static analysis, and people _do_ want to be able to run Mocha w/o this sort of restriction.  

At this point though I wonder if we should add something like a "CI mode" which enables various "strict" behaviors...

eemeli:
Sorry, this fell off my radar. Yes, I'm still interested in working on this. I'll need to review @juergba's comments and get back to you on this.

eemeli:
Closing, as I clearly haven't had the time to work on this.

PR #4127, opened 2019-12-14
Add reporter names enable and Unite names into one

### LINKED ISSUE #4054: (documentation) Inconsistent formatting of reporter names
The report section of the documentation has inconsistent formatting.

eg:

* The `xunit` reporter..
* The Landing Strip (`landing`) reporter...
* The dot matrix (or "dot") reporter...
* The "min" reporter...
* This is the default reporter. The "spec" reporter...
* The HTML reporter...

I believe the main purpose of the quoted or code-blocked text is to tell you what exact string names you should use when indicating your desired reporter.  I would say none of these formats do a good job at explicitly stating that.  It is more a hint that you then have to use as a guess.

Additionally, it appears that the reporters have multiple acceptable names. (Possibly dependent on whether you are running in node or on the browser?)

In addition to the existing formats, I would like to suggest a format something like this for clarity:

> ### JSON Stream
> Names:  `"JSONStream"` or `"json-stream"`
> The JSON stream reporter outputs newline-delimited JSON "events" as they occur, beginning with a "start" event, followed by test passes or failures, and then the final "end" event.

(Or maybe no alternate names at all? [Their existence seems to be unnecessary, separate issue in the possible future.  Though, if people agree that the alternate names should not be encouraged, then only the preferred name should be in the documentation.])

I created this issue because I didn't realize just making the the reporters all follow the most common format would be a big deal.  I had originally tacked this onto a different PR.  It has since been removed.  Another issue I created also attracted similar attention.  

Here are the previously mentioned issues / pull requests:
* PR #4026
* Issue #4027 & PR #4037 

I will be stepping back from this conversation as I don't have a very strong opinion about this either way, and have already spent way to much time redoing commits for this cosmetic change.

### LINKED ISSUE #4134: Mocha docs has wrong third-party example link
### Description

In third-party reporter part of docs, there is a `our working directory` link to mocha's official example.
[reference](https://mochajs.org/#third-party-reporters)

스크린샷 2019-12-29 오후 2 57 57

this link refers to `https://github.com/mochajs/mocha-examples/tree/master/third-party-reporter`. but when you try to access to that address, github returns 404 page.

the right address is `https://github.com/mochajs/mocha-examples/tree/master/packages/third-party-reporter`

I think there are changes to example repository but It doesn't reflect to docs.

khg0712:
### Description
- Reporter part of Mocha docs was so unkind about enable reporter name.
  - there are two enable names but current docs makes users confused.

### Description of the Change
- Add **enable names** to docs clearly
- Unite reporter name to call a reporter type

### Summary
*before*
The "progress" reporter implements a simple progress-bar:

*after*
**Enable Names:** `Progress`, `progress`
The Progress reporter implements a simple progress-bar:

fix #4054 
fix #4134

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1917,7 +1917,7 @@ or the [source](https://github.com/mochajs/mocha/blob/master/lib/mocha.js).
 [example-mocha-test]: https://github.com/mochajs/mocha/tree/master/test
 [example-mocha-config]: https://github.com/mochajs/mocha/tree/master/example/config
 [example-superagent-test]: https://github.com/visionmedia/superagent/tree/master/test/node
-[example-third-party-reporter]: https://github.com/mochajs/mocha-examples/tree/master/third-party-reporter
+[example-third-party-reporter]: https://github.com/mochajs/mocha-examples/tree/master/packages/third-party-reporter
 [example-typescript]: https://github.com/mochajs/mocha-examples/tree/master/packages/typescript
 [example-websocket.io-test]: https://github.com/LearnBoost/websocket.io/tree/master/test
 [expect.js]: https://github.com/LearnBoost/expect.js
--- a/docs/index.md
+++ b/docs/index.md
@@ -1535,6 +1535,8 @@ Examples:
 
 ### HTML Reporter
 
+Alias: None
+
 **The HTML reporter is not intended for use on the command-line.**
 
 ## Running Mocha in the Browser
--- a/docs/index.md
+++ b/docs/index.md
@@ -1376,24 +1376,32 @@ Mocha reporters adjust to the terminal window, and always disable ANSI-escape co
 
 ### Spec
 
-This is the default reporter. The "spec" reporter outputs a hierarchical view nested just as the test cases are.
+**Enable Names:** `Spec`, `spec`
+
+The Spec reporter outputs a hierarchical view nested just as the test cases are.
 
 ![spec reporter](images/reporter-spec.png?withoutEnlargement&resize=920,9999){:class="screenshot" lazyload="on"}
 ![spec reporter with failure](images/reporter-spec-fail.png?withoutEnlargement&resize=920,9999){:class="screenshot" lazyload="on"}
 
 ### Dot Matrix
 
-The dot matrix (or "dot") reporter is simply a series of characters which represent test cases. Failures highlight in red exclamation marks (`!`), pending tests with a blue comma (`,`), and slow tests as yellow. Good if you prefer minimal output.
+**Enable Names:** `Dot`, `dot`
+
+The Dot Matrix reporter is simply a series of characters which represent test cases. Failures highlight in red exclamation marks (`!`), pending tests with a blue comma (`,`), and slow tests as yellow. Good if you prefer minimal output.
 
 ![dot matrix reporter](images/reporter-dot.png?withoutEnlargement&resize=920,9999){:class="screenshot" lazyload="on"}
 
 ### Nyan
 
-The "nyan" reporter is exactly what you might expect:
+**Enable Names:** `Nyan`, `nyan`
+
+The Nyan reporter is exactly what you might expect:
 
 ![js nyan cat reporter](images/reporter-nyan.png?withoutEnlargement&resize=920,9999){:class="screenshot" lazyload="on"}
 
 ### TAP
+
+**Enable Names:** `TAP`, `tap`
 
 The TAP reporter emits lines for a [Test-Anything-Protocol][] consumer.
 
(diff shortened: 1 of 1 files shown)


outsideris:
@khg0712 Your email for git commit is not connected your GitHub account.

khg0712:
@outsideris I changed it!

docs/index.md
```diff
@@ -1376,69 +1376,91 @@ Mocha reporters adjust to the terminal window, and always disable ANSI-escape co
 
 ### Spec
 
-This is the default reporter. The "spec" reporter outputs a hierarchical view nested just as the test cases are.
+**Enable Names:** `Spec`, `spec`
```
juergba: I'm unsure about "Enable Names", is this an official expression or korean english (I'm not english native neither)? [Here](https://github.com/mochajs/mocha/blob/master/lib/reporters/index.js#L3) it's called `alias`.
I would also prefer a format with less bold letters.

docs/index.md
```diff
@@ -1376,69 +1376,91 @@ Mocha reporters adjust to the terminal window, and always disable ANSI-escape co
...
+
+The Spec reporter outputs a hierarchical view nested just as the test cases are.
```
juergba: Please don't remove "This is the default reporter.".

docs/index.md
```diff
@@ -1485,14 +1507,18 @@ View SuperAgent's [Makefile][superagent-makefile] for reference.
...
 example, here is the Connect [test output][connect-test-output].
 
 ### XUnit
 
-The `xunit` reporter is also available. It outputs an XUnit-compatible XML document, often applicable in CI servers.
+**Enable Names:** `XUnit`, `xunit`
+
+The XUnit reporter is also available. It outputs an XUnit-compatible XML document, often applicable in CI servers.
 
 By default, it will output to the console. To write directly to a file, use `--reporter-options output=filename.xml`.
```
juergba: Further down you find the HTML reporter, please fix this one as well.

khg0712: @juergba 
in docs, HTML reporter is explained that `The HTML reporter is not intended for use on the command-line.`

so, I think you mean that I should add there is no alias with HTML reporter.
Is that correct?

juergba: No, this is not correct.
Yes, HTML is not intended for CLI, but Mocha browser also has a setup option:
`mocha.setup({ reporter: 'html' });`
So I suppose both aliases `HTML` and `html` are valid names.

Please also fix the broken link you mentioned in #4134.

khg0712: @juergba 
then, which Is better specifying alias is  for `cli` or `browser` or writing just alias?

docs/index.md
```diff
@@ -1889,8 +1917,8 @@ or the [source](https://github.com/mochajs/mocha/blob/master/lib/mocha.js).
 [example-mocha-test]: https://github.com/mochajs/mocha/tree/master/test
 [example-mocha-config]: https://github.com/mochajs/mocha/tree/master/example/config
 [example-superagent-test]: https://github.com/visionmedia/superagent/tree/master/test/node
-[example-third-party-reporter]: https://github.com/mochajs/mocha-examples/tree/master/third-party-reporter
-[example-typescript]: https://github.com/mochajs/mocha-examples/tree/master/packages/typescript
+[example-third-party-reporter]: https://github.com/mochajs/mocha-examples/tree/master/packages/third-party-reporter
+[example-typescript]: https://github.com/mochajs/mocha-examples/tree/master/typescript
```
juergba: This change is wrong.

khg0712:
@juergba 
I want to add alias to `mocha --reporters`

스크린샷 2019-12-29 오후 6 03 00

I dig down to solve this problem. I encountered this [code](https://github.com/mochajs/mocha/blob/master/lib/cli/one-and-dones.js#L21).  `showKeys` function prevent showing capitalized word. how about modify reporter's showKeys logic?

docs/index.md
khg0712: I'll change it

juergba:
> I dig down to solve this problem.

I don't see any problem, IMO this list is perfect as it is.

khg0712:
> I don't see any problem, IMO this list is perfect as it is.

doesn't it needs to specify all alias like `HTML`, `html`?

docs/index.md
juergba: I don't understand your question, but just write alias.

juergba:
No, HTML is the default reporter for Mocha browser and should not show up in this CLI listing.

khg0712:
sry, I cited wrong example. 

> @juergba
> I want to add alias to `mocha --reporters`
> 
> 스크린샷 2019-12-29 오후 6 03 00
> 

I meant that it seems to need to be added capitalized alias to cli options

juergba:
I have understood your concerns, but this listing is good as is, please leave it unchanged.

Issue #792, opened 2013-04-03
🚀 Feature: Highlight browser failures in red


tj:
to make it more obvious without wasting eye-movement effort to look at the count haha. I prefer github.com/visionmedia/mocha-matrix for this reason but it's not a good fit as a default

danielstjules:
I agree that'd be helpful :) So far, failures aren't generally obvious unless they're early in the suite:

![screen shot 2015-03-08 at 9 53 52 pm](https://cloud.githubusercontent.com/assets/817212/6550121/d9a6b07c-c5dd-11e4-8f19-33285a5bbb87.png)

I'll a submit a PR that adds a horizontal bar in the header. It'll be light grey while running, green if it passed, and red if it failed.

mayankshah1607:
Hi. I am new here and would like to contribute.
From what I understand, when a test suite fails, messages must be displayed in red and if success, messages must be displayed in green. From what I know, I think mocha already does this. I would like to know what must be done exactly, and would like a deeper insight to this issue.

Thank you! :)

dhuang612:
Hi,

is this available to work on?

outsideris:
Sure

dhuang612:
created pr #4082

dhuang612:
updated pr with new code and included screenshots.

Please let me know if anything else needs to be changed

dhuang612:
I see a ref to the new pr but just wanted to make sure you guys were aware.

created a new pr for this issue:
#4148

pedro-qwkin:
Why isn't this merged?

dhuang612:
@outsideris Hi can I get an update on my PR?
https://github.com/mochajs/mocha/pull/4148

outsideris:
@dhuang612 We missed it. I will check your PR.

mark-wiemer:
#4148 looks like it can be resurrected, I'll see what I can do tomorrow :)

mark-wiemer:
FYI #5218 is ready for review :)

JoshuaKGoldberg:
Thanks @mark-wiemer! Coming over from https://github.com/mochajs/mocha/pull/5218#discussion_r1791954219: my personal take is that the best implementation for this issue would be one that doesn't change page layout. From that comment: the taller the summary section, the less fits on the page in smaller viewports. A lot of people are accustomed to the placement as-is. 

[...]

PR #4148, opened 2020-01-09
Make browsertest failures standout

dhuang612:
### Description of the Change

resolves issue #792 created a bar that starts out as green when the tests start.
Then if one of the tests fails bar will turn red to indicate test failure.
From the previous attempt learned that code coverage is decreased.
Will work on creating a test for this to get this resolved.

### Benefits

Allows for less eye movement to notice tests that don't pass when run from google chrome. 

### Possible Drawbacks

there shouldn't be any side effects because the changes are gui level

changes made are an enhancement.


dhuang612:
reduced the size of the bar and gave them curved edges 

success:
Screen Shot 2020-01-20 at 1 46 38 PM

failure:
Screen Shot 2020-01-20 at 1 50 11 PM

Please let me know if anything else needs to be changed

pedro-qwkin:
Why isn't this merged?

lib/reporters/html.js
```diff
@@ -35,7 +35,7 @@ exports = module.exports = HTML;
 /**
  * Stats template.
  */
-
+var statsBar = '<h1 id="progress-bar"></h1>';
```
outsideris, line 38: How about we use `<div>` or `<progress>` tag since it is not heading.

outsideris (commented):
It works with the latest master even we're using rollup instead browserify.

lib/reporters/html.js
dhuang612: @outsideris changed from h1 to a div

dhuang612:
I tried to write some tests under base.spec they complete successfully but don't add to the coverage. Since there are no tests right now for the html browser. 
Let me know what you want done next.

juergba:
I recently pinned #3071 in the issue tracker and our exApplitools browser testing crack @giltayar commented: https://github.com/mochajs/mocha/issues/3071#issuecomment-863771039.

I'm aware of that we don't have any html reporter tests, yet. Wouldn't it be much easier to take the path Gil mentioned? A snapshot test with Selenium WebDriver? I'm attracted by Playwright though, I don't know exactly how it works, but it's nice 😎. We could forget about IE11 as we most probably are going to drop its support with the next major release v10.
We wouldn't need any mocking, bundling or Saucelab, just a HTML file with a failing test and compare the output against an existing snapshot.

juergba:
@dhuang612 I got a negative reply by the person I contacted, so the browser tests remains unsolved. He recommended to use playwright. I don't know yet how to proceed.

Your screenshots look nice, the colors look evtl. a bit strong. I haven't tested, yet.
`#mocha-stats.fail`: is there no other way like `extend`, more elegant than repeating the entire `#mocha-stats`? Is CSS so ugly?

dhuang612:
I'll take a look at using extend.

dhuang612:
@juergba 
I added in Sass as a dev package.
I then update the css file name to sass.
Finally, I switched the class I added to extend the original, and only updated the status bar colors themselves.

I tested the feature out using the example files to make sure they worked.

package.json
```diff
@@ -144,6 +144,7 @@
     "rollup-plugin-node-globals": "^1.4.0",
     "rollup-plugin-node-polyfills": "^0.2.1",
     "rollup-plugin-visualizer": "^4.1.0",
+    "sass": "^1.39.2",
```
JoshuaKGoldberg, line 147: [Structure] `sass` is a pretty big dependency to use for exactly one `@extend`. I don't think it's justified here. Let's revert the change from `.css` to `.sass`, unless there's a strong reason for it?

mocha.sass
```diff
@@ -263,8 +263,18 @@ body {
...
-  color: #888;
+  color: #1d1e1f;
   z-index: 1;
+  background-color: #00e600;
+  border: solid  #00e600;
+  border-radius: 10px;
+}
+
+#mocha-stats.fail {
+ @extend #mocha-stats;
```
JoshuaKGoldberg, line 274: [Style] I don't think this `@extend` is really needed. We can achieve the same result in raw CSS, such as with CSS variables or different class names. 

Thinking in terms of semantic class names, I think it'd make sense to have a class like `.fail`, `.pass`, etc. that each define their own colors.

Answering https://github.com/mochajs/mocha/pull/4148#issuecomment-915374638: no, CSS can overcome these issues. 🙂

lib/reporters/html.js
```diff
@@ -111,7 +111,6 @@ function HTML(runner, options) {
       hideSuitesWithout('test fail');
     }
   });
-
```
JoshuaKGoldberg, line 114: [Style] Nit: unnecessary change?

mocha.sass
JoshuaKGoldberg, line 1: Looking at the styles:

<table>
<thead>
<tr>
<th>Pass</th>
<th>Fail</th>
</tr>
</thead>
<tbody>
<td>
image
</td>
<td>
image
</td>
</tr>
</tbody>
</table>

A few things that make me not a fan yet:
* The new colors are hard to read on top of, to the point of not having [accessible color contrast](https://webaim.org/resources/contrastchecker/?fcolor=000000&bcolor=CC0000).
* The colors also don't match the existing aesthetic of the page. The page right now has very minimalist colors and a subtle design. This new green and red is pretty glaring. We should go with something that doesn't clash so much.
* It's a little inconsistent in how its laid out: more padding on the top vs. the bottom, and less horizontal padding than other UI elements.

I'm not sure what would be the "right" approach. A few things come to mind, but I haven't thought very deeply:

* Using much thinner border lines and/or much less saturated colors
    * Even switching to an [nprogress](https://github.com/rstacruz/nprogress)-style bar on the stats header or the page
* Instead of a background or border, using bold & colored text

(I'm not convinced any of that is better, just ideating)

JoshuaKGoldberg (changes requested):
I think we'll want to have some more discussion into what this UI should look like. Proposal: can we have that in the backing issue, #792? That way we can figure out what it should look like a bit more before jumping into the implementation.

JoshuaKGoldberg:
👋 ping @dhuang612, do you still have time for this?

JoshuaKGoldberg:

mark-wiemer:
@JoshuaKGoldberg I'll be coming to this one as a software engineer at Microsoft specializing in a11y :) please let me know if you have further comments before I start resolving the existing ones. I'll probably be asking for some help in the discord as well :) thanks again for picking up this project!

package.json
mark-wiemer: Removed in #5218 ;)

PR #4150, opened 2020-01-13
uncaughtException: fix recovery when current test is still running

juergba:
### Description

```js
describe('uncaught', function() {
  it('throw delayed error', (done) => {
    setTimeout(() => {
      throw new Error('Whoops!');
    }, 10)
    setTimeout(done, 10);
  });
  it('should wait 15ms', (done) => {      
    setTimeout(done, 15);
  });
  it('test 3', () => { });

  afterEach(function() {
    console.log('inside afterEach:', this.currentTest.title);
  });
});
```
This bug produces very variable results as throwing `TypeError` and failing empty tests. In above example the output is complete nonsense.
```
uncaught
    1) throw delayed error                  // first test fails
    √ throw delayed error                   // then same test passes
inside afterEach: throw delayed error      // same hook twice 
inside afterEach: throw delayed error
    √ test 3                                // second test is missing
    √ test 3                                // third test runs twice
inside afterEach: test 3                   // same hook twice 
inside afterEach: test 3

  3 passing (41ms)
  1 failing

  1) uncaught
        throw delayed error:
      Uncaught Error: Whoops!
      [...]
```
Fixed output:
```
uncaught
    1) throw delayed error
    2) throw delayed error
inside afterEach: throw delayed error
    √ should wait 15ms
inside afterEach: should wait 15ms
    √ test 3
inside afterEach: test 3

  2 passing (62ms)
  2 failing

  1) uncaught
       throw delayed error:
      Uncaught Error: Whoops!
      [...]
  2) uncaught
       throw delayed error:
      Error: done() called multiple times
      [...]
```

### Description of the Change

We remove the recovery of a failing open test out of `Runner#uncaught` and pass the error to `runnable`'s callback. `done` finishes the `runnable` and triggers the standard error handling incl. hook pattern.

### Applicable issues

closes #2906, see [comment](https://github.com/mochajs/mocha/issues/2906#issuecomment-574564419)

PR #4152, opened 2020-01-14
multiple async done() calls result in failure; closes #4151

### LINKED ISSUE #4151: multiple calls to done() doesn't always fail if second done is called async
Example:

```js
it('should fail', function (done) {
    done();
    setTimeout(done);
  });
});
```

The above won't fail with a nonzero exit code. `EVENT_TEST_FAIL` is emitted, but nothing is listening at this point.

To address this:

1. The `Runner` must have a notion of a "stopped" state
1. If a test fails _while the `Runner` is in the stopped state_, then it must throw an exception.
1. How this exception is handled depends on whether the standard uncaught exception handler is still bound.  If it is, we let it be handled, and we get a proper stack trace back to the test.
1. If the standard uncaught exception handled has been disabled (which happens when `EVENT_RUN_END` is emitted)--and the process will fail with a bog-standard uncaught exception--Mocha will need to retain extra information in the error so the user can find which test is responsible.  I'm not sure why we lose the stack trace back to the test; maybe somebody can help.

Anyway, I have a fix in-progress for this.

Ref: https://twitter.com/jsumners79/status/1216391159699988481

boneskull:
This fixes bug #4151.

- added a method in `errors` module to create a "multiple done" err
- modernize `multiple-done.spec.js`
- refactor errors into constants in `errors` module
- remove `Runner#started` prop; replace with `Runner#state` prop + constants
- add a catchall `createFatalError()` function to `errors` module; this is called
  when a test fails twice by other means (unsure what those means are yet)

Please note that because of how the fixture for this is written (using `setTimeout`), the behavior of the associated test is non-deterministic.  We _could_ increase the delay, but we're just hacking around and will likely have to increase or decrease it again (and write another test!).  One of two things will happen:

- The fixture will fail before the `Runner` completes, or
- The fixture will fail after the `Runner` completes

In the first case, it's handled by the `Runner`'s uncaught exception listener, (assuming `--allow-uncaught` is not set) ~~and will exit with code 1~~.  In the second case, it's caught by the, uh, _other_ uncaught exception listener, `uncaughtEnd()`, (@juergba will know more about this) ~~and exits with code 7~~. 

I've tried to give the user enough context to track down the problem.  The test covers whichever of these two happens to occur via a dynamic call to `skip()`.

Update: Can't rely on the exit codes for anything


juergba:
This PR is a good one, it helps me out with a problem I have in #4150. 
I will need some days for reviewing.

Following test case shows more reliable results, maybe you can fix the failing CI tests that way.
```js
describe('uncaught test', function() {
  this.timeout(1500);
    it('throw delayed error', (done) => {      
      setTimeout(() => {done(/*new Error()*/);}, 1000);
      setTimeout(() => {done();}, 2000);
    });
});
```
Btw, I have not understood yet why (most parts of) `Runner#uncaught()` is registered on `process.on('uncaughtException', ...)` instead of `runner.on('error', ...)`.

lib/runner.js
```diff
@@ -834,7 +857,7 @@ Runner.prototype.uncaught = function(err) {
     runnable = new Runnable('Uncaught error outside test suite');
     runnable.parent = this.suite;
 
-    if (this.started) {
+    if (this.state === constants.STATE_RUNNING) {
       this.fail(runnable, err);
     } else {
       // Can't recover from this failure
```
juergba: This part is probably wrong as well. Why don't we just abort here, or throw? When the runner has stopped, we throw anyway. When it hasn't even started, then there is nothing to recover or fail or to report, so we should just throw.

juergba:
When a reporter errors (except `generateDiff` now), Mocha aborts without any message and parts of the reporter's output / epilogue are missing.
I don't know wether that bug is related to this PR as well.

lib/runner.js
boneskull: I'm not sure if it's wrong or not.  If we know which `Runnable` is throwing the uncaught error, then I think it makes sense to report it as a failure instead of just bailing.  But I could be convinced otherwise.  Either way, I think we can address it in another PR?

boneskull:
@juergba If a reporter errors, that's an internal error, and shouldn't we just bail? Don't think that's related to this, though.

- I fixed the test and behavior to be "more consistent".
- Looks like I need to add some unit tests, as I suspected.

boneskull:
> When you remove the first setTimeout which results in a timeout error, then the second done is swallowed. Otherwise when the first done reports an error, the second done throws. That's weird, not consistent.

I'll add a case for this and see if I can get to the bottom of it and maybe make it more consistent, but no guarantees right now.

when an async test times out, Mocha _calls `done(err)` for the test_ where `err` is a timeout error.  So, when the `done()` is eventually called by the test--e.g., `setTimeout(done, 5000)`--it's already considered a "multiple call to done" and is ignored.  at least, I think that's how it works.  I think we want to keep ignoring stuff because otherwise anything timing out which also calls `done(err)` may be noisy, but, again, I could be convinced otherwise

lib/runner.js
juergba: Agree, separate PR.

juergba:
> I'll add a case for this and see if I can get to the bottom

I know that in `Runner#uncaught` errors are ignored when `runnable`'s is already failed. Is there another place where errors get ignored intentionally?

lib/runner.js
```diff
@@ -928,7 +951,7 @@ Runner.prototype.run = function(fn) {
     if (rootSuite.hasOnly()) {
       rootSuite.filterOnly();
     }
-    self.started = true;
+    self.state = constants.STATE_RUNNING;
```
juergba: Is setting runner's state not too early, in case of `delay`?

boneskull: not sure.  will see.

boneskull:
> Since we have a Runner#state property now, we don't need to unregister Runner#uncaught and register Runner#uncaughtEnd when runner ends. We just need one uncaught handler, where we rethrow when runner.state === constants.STATE_STOPPED;

Ooh, that's good.  Yes.

> > Btw, I have not understood yet why (most parts of) Runner#uncaught() is registered on process.on('uncaughtException', ...) instead of runner.on('error', ...).
> Do you know why?

If we catch an error in `Runner` code, then we can emit it as an `error` event, but there's no guarantee we can always do that.

For example, a simple assertion error would be catchable, and could be emitted as `error`:

```js
it('should fail', function() {
  throw new Error()
});
```

likewise

```js
it('should fail', function(done) {
  setTimeout(() => {
    done(new Error())
  });
});
```

or

```js
it('should fail', async function() {
  return Promise.reject()
});
```

as far as I recall, this very common pattern is handled by `uncaughtException`; the Runner cannot catch this otherwise:

```js
it('should fail', function(done) {
  setTimeout(() => {
    throw new Error();
  });
});
```

last time I checked, Mocha doesn't even do anything with this:

```js
it('should fail', function() {
  Promise.reject();
});
```

The only way to handle this--like the previous case--is a listener on `process` (`unhandledRejection`, in this case).

We should probably add an `unhandledRejection` handler 😜

boneskull:
@nicojs Some of this touches stuff you changed recently.  I found an issue, but I think it has existed for a long time.

In particular, in `Runner#run`:

```js
  var self = this;
  function uncaught(err) {
    self.uncaught(err);
  }

  self._removeEventListener(process, 'uncaughtException', uncaught);
  self._addEventListener(process, 'uncaughtException', uncaught);
```

Do you see the bug? 😄

We use this `uncaught` function because when a listener is invoked by an `EventEmitter`, the listener is called with the context of the `EventEmitter`.  But we need it to use the context of the `Runner` instance, so this is a workaround.

`EventEmitter#removeListener(eventName, listener)` needs the exact same `listener` as was registered in the first place in order to remove it (and only it).  But every time we call `Runner#run`, a new `uncaught` function is created.  So, in essence, `self._removeEventListener(process, 'uncaughtException', uncaught);` does nothing, because our newly-created `uncaught` will not have been registered.  Thus, if ten calls are made to `Runner#run`, ten different listeners will be registered.

This is unlikely to be a problem during the normal course of operation, I think, which is why it wouldn't have been noticed.

The fix isn't awesome.  I renamed `Runner#uncaught` to `Runner#_uncaught`, and in the `Runner` constructor:

```js
this.uncaught = this._uncaught.bind(this);
```

Now, we change the aforementioned code to:

```js

  var self = this;
  self._removeEventListener(process, 'uncaughtException', self.uncaught);
  self._addEventListener(process, 'uncaughtException', self.uncaught);
```

Even having access to lambda functions, we'd still need the listener to call with the correct context.  So even then, the `Runner` constructor would have to do this:

```js
this.uncaught = err => self.uncaught(err);
```

This could also probably be solved with a `Proxy`, but that's even worse, IMO.

PR #4153, opened 2020-01-14
Fix backwards compatibility break for `reporterOptions`

### LINKED ISSUE #4142: Support `require` and `reporter-options` in `mocharc`

I am not able to get `reporter-options` (or `reporterOptions`) working with either `_mocha` or `mocha` when the config is within `mocha` on `package.json` or in `.mochrarc.js`

`require` is also not working either as an array or set of strings.

But other properties are working.

I'd like it to be possible to add `reporterOptions` and `require` within `mocharc.js`.

It seems unnecessarily clunky to add as flags to the CLI, especially when I've already added an RC file.

**Additional context**

-

holm:
### Description of the Change

In https://github.com/mochajs/mocha/pull/4004 there was a change to use the documented `reporterOption` in favour of the setting that had always been used in practice called `reporterOptions`. This broke a lot of configurations that used the `reporterOptions`, which was the only way it every worked AFAIK. This is also clear in that all the built-in reporters use this name.

This changes the documentation to specify `reporterOptions` instead and ensure that any one that has switched to `reporterOption` after upgrade, still works.

### Alternate Designs

`reporterOption` could have been chosen as the new preferred way, but objects with options are generally referred to in their plural form in JS. I think the root cause of the confusion comes from the command line option `reporter-option`. I think it makes sense that this is singular, since it is a single option, but it is compiled into a `reporterOptions` object.

To fix related issues like https://github.com/michaelleeallen/mocha-junit-reporter/issues/106#issuecomment-574015463

### Benefits

Make it easier to upgrade to Mocha 7

### Possible Drawbacks

Slight overhead of supporting two aliases

### DIFF
--- a/test/unit/mocha.spec.js
+++ b/test/unit/mocha.spec.js
@@ -446,6 +446,30 @@ describe('Mocha', function() {
       var mocha = new Mocha(opts);
       expect(mocha.reporter(), 'to be', mocha);
     });
+
+    it('should keep reporterOption on options', function() {
+      var mocha = new Mocha({
+        reporter: 'spec',
+        reporterOption: {
+          foo: 'bar'
+        }
+      });
+      expect(mocha.options.reporterOption, 'to have property', 'foo', 'bar');
+      // To support the legacy property name that can be used by reporters
+      expect(mocha.options.reporterOptions, 'to have property', 'foo', 'bar');
+    });
+
+    it('should support legacy reporterOptions', function() {
+      var mocha = new Mocha({
+        reporter: 'spec',
+        reporterOptions: {
+          foo: 'bar'
+        }
+      });
+      expect(mocha.options.reporterOption, 'to have property', 'foo', 'bar');
+      // To support the legacy property name that can be used by reporters
+      expect(mocha.options.reporterOptions, 'to have property', 'foo', 'bar');
+    });
   });
 
   describe('#run(fn)', function() {
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -100,7 +100,10 @@ function Mocha(options) {
   this.grep(options.grep)
     .fgrep(options.fgrep)
     .ui(options.ui)
-    .reporter(options.reporter, options.reporterOption)
+    .reporter(
+      options.reporter,
+      options.reporterOption || options.reporterOptions // reporterOptions was previously the only way to specify options to reporter
+    )
     .slow(options.slow)
     .global(options.global);
 


holm:
I'm unsure what change you are requesting here. I'm not sure why the change I am proposing is inconsistent with anything.

Surely you do not want to require everyone to update their settings and all reporters to release new versions of their packages, when then i turn have to manage backwards compatibility with older versions of Mocha?

juergba:
`reporterOptions` is not a canonical name, but an alias.

I'm planning to reduce the options passed into `Mocha` constructor (in our case: `reporter-option`, `reporter-options`, `reporterOption`, `reporterOptions`, `O`). So I don't agree to the change from `reporterOption` to `reporterOptions` as expected option name.

For backwards compability (L105) we could add `reporterOptions` as fallback, but `reporterOption` should remain the option to be.

holm:
The `Mocha` constructor is never called with `reporter-option` or `reporter-options` AFAIK. That is only done via the CLI.

Searching the code base, up until your change, `reporterOption` was never used outside the documentation mistake. If anyone had passed it in, it would simply be ignored. From what I can tell you want to change it to use `reporter-option` on the command line and `reporterOption` in the rest of the code. I don't really see why that is better than my proposal of `reporter-option` in the CLI and `reporterOptions` in the rest of the code. They are not the same, one is a single option, the other is a set of options.

Since `reporterOptions` have always been used by everyone, rather than try to force this change on everyone, wouldn't it be better to just use my suggestion?

juergba:
No, I don't want to change anything for use in CLI, everything remains as is.
In CLI you can use comma separators, so `--reporter-option` is not only a single option.

For the constructor, I need this rule for all options: "canonical or its camel-cased version".
"[...] have always been used [...]" is not a good argument, especially in a major release. Furthermore if we keep backwards compability with the mentioned fallback.

test/unit/mocha.spec.js
```diff
@@ -446,6 +446,36 @@ describe('Mocha', function() {
...
+
+    it('should keep reporterOption on options', function() {
+      var mocha = new Mocha({
+        reporter: 'spec',
+        reporterOption: {
+          foo: 'bar'
+        }
+      });
+      expect(mocha.options.reporterOption, 'to have property', 'foo', 'bar');
+    });
```
juergba: pls merge the first two tests into one and add:
`expect(mocha.options.reporterOptions, 'to have property', 'foo', 'bar');`

test/unit/mocha.spec.js
```diff
@@ -446,6 +446,36 @@ describe('Mocha', function() {
...
+
+    it('should support legacy reporterOptions', function() {
+      var mocha = new Mocha({
+        reporter: 'spec',
+        reporterOptions: {
+          foo: 'bar'
+        }
+      });
+      expect(mocha.options.reporterOption, 'to have property', 'foo', 'bar');
+    });
```
juergba: also add here: `expect(mocha.options.reporterOptions, 'to have property', 'foo', 'bar');`

PR #4155, opened 2020-01-15
change minimum Node version requirement to 8.10.0

natesilva:
### Description of the Change

Mocha’s `package.json` shows that it runs on Node 8.0.0 or higher. But when using Node 8.0.0 it throws an error:

```shell
~/Desktop/node-8.0.0-test.tmp
❯ ./node_modules/.bin/mocha --version
/Users/nate/Desktop/node-8.0.0-test.tmp/node_modules/chokidar/index.js:904
  const options = {type: EV_ALL, alwaysStat: true, lstat: true, ...opts};
                                                                ^^^

SyntaxError: Unexpected token ...
    at createScript (vm.js:74:10)
    at Object.runInThisContext (vm.js:116:10)
    at Module._compile (module.js:533:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/nate/Desktop/node-8.0.0-test.tmp/node_modules/mocha/lib/cli/watch-run.js:4:18)
```

Of Mocha’s direct dependencies, `chokidar` has the highest requirement, at 8.10.0. With 8.10.0, Mocha passes all tests, and I can run it successfully from the CLI.

Therefore this PR changes the `engines` value to `>= 8.10.0`.

### Benefits

Users (both human and automated systems) will know which versions of Node they can successfully use with Mocha.

### Applicable issues

Updating the minimum Node version can be a breaking change, but in this case I don’t think it matters since users of earlier versions were not able to use Mocha in the first place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -526,7 +526,7 @@
     "test": "./test"
   },
   "engines": {
-    "node": ">= 8.0.0"
+    "node": ">= 8.10.0"
   },
   "scripts": {
     "prepublishOnly": "nps test clean build",


juergba:
@natesilva thank you for this PR.

> Of Mocha’s direct dependencies, chokidar has the highest requirement

How have you checked that? Do I have to check each dependency manually?

NodeJS reached end-of-life on 2019/12/31 and LTS release was `v8.17.0`.
We could also change `engines` to =`v8.17.0`? And yes, it's close to being a breaking change.

natesilva:
I checked the dependency versions with ripgrep in a directory that had nothing but `mocha` installed:

```
rg -A1 "engines" | rg \"node\"\:
```

I’m not sure how practical it is to track down the engine versions of all dependencies and sub-dependencies. Some projects may not test this, and others don’t set it.

For example, `yargs` is a direct dependency of `mocha` that does not have an `engines` value. `lodash` is an indirect dependency that does not have an `engines` value.

There’s also a sub-dependency of `chokidar` that has a higher requirement (`fsevents` with `8.16.0`).

What I do know is that at Node v8.0.0 `mocha` does not pass its tests. At v8.10.0, it does.

juergba:
@natesilva Thank you.
eg. Yargs/-parser incorrectly use `engine` instead of `engines` in their package.json.

Issue #4154, opened 2020-01-15
ship TS type declarations


boneskull:
AFAIK, it's recommended by both the TypeScript team and DefinitelyTyped (the people that publish everything under the `@types` scope on npm) that packages ship their own type declarations (`.d.ts` files).  I think this is essentially for two reasons:

1. It's tough to maintain types for a package you don't control
1. The package owners are the best people to understand what the types are

I was on the fence about this for awhile, because I wasn't interested in trying to manually maintain declarations.  

Recently, TypeScript introduced support for generation of type declarations from JavaScript sources.

I've used this feature pretty extensively (see [report-toolkit](/IBM/report-toolkit)) and have had good success with it.  The way it works is this:

1. Use JSDoc-style comment tags, where appropriate, which TS can understand.  

    Mocha is already most of the way there, I think.  Depending on the extent of the type inference, we may be able to restrict it to a subset the source code (initially, the user-facing, _non-programmatic_ APIs), instead of needing to meticulously work through the entire codebase under `lib/`, but it's hard to say until we try it.

    Custom types (describing the shape of a plain object or the signature of a callback) are accomplished via `@typedef`.  These are equivalent to TS' `type` keyword and subject to the same [limitations](https://stackoverflow.com/a/52682220/1146762). 

    One place where we may get tripped up is that TS is unable to understand prototypal inheritance when not using the `class` keyword.  So, though `Test` extends `Runnable`, we won't be able to illustrate that via the declaration--but if we focus on non-programmatic usage, this is moot.

    We need to make sure we're not breaking our API doc generation, though.  I don't think this will be much of an issue, as that's mainly focused on the programmatic API.  We can revisit the programmatic API at a later point.
1. Use `tsc` to generate a `.d.ts` artifact from the sources we care about.  Make it part of the build process; it doubles as a "type linter", because if something is incorrectly typed, the build will fail. 
 Publish the resulting file(s) and keep them out of VCS.

Comments?  Hate the idea?

DanielRosenwasser:
I'm super happy to see the enthusiasm here! I'd be a bit cautious though because depending on how your types differ, users may see some differences (at least initially) or have dependencies on how the types are declared. I think it's still worth trying to compare/contrast though!

DanielRosenwasser:
(also we're looking for feedback)

weswigham:
[...]

PR #4156, opened 2020-01-17
WIP Playing around with .d.ts generation

cspotcode:
Related to #4154

EDIT: got rid of the stuff about compiling; switched to `emitDeclarationsOnly: true`.

To play around with it locally, `yarn nps build.types` and `yarn nps test.types`

### DIFF
--- a/lib/cli/options_t.ts
+++ b/lib/cli/options_t.ts
@@ -1,0 +1,7 @@
+export interface MochaArgs {
+  /** Path to `package.json` or `false` to skip */
+  config?: string | false;
+  package?: boolean;
+  /** Filepath to mocha.opts; defaults to whatever's in `mocharc.opts`, or `false` to skip */
+  opts?: string | false;
+}
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -708,7 +708,7 @@ exports.isPromise = function isPromise(value) {
  * Clamps a numeric value to an inclusive range.
  *
  * @param {number} value - Value to be clamped.
- * @param {numer[]} range - Two element array specifying [min, max] range.
+ * @param {number[]} range - Two element array specifying [min, max] range.
  * @returns {number} clamped value
  */
 exports.clamp = function clamp(value, range) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -29,7 +29,10 @@ var assign = (exports.assign = require('object.assign').getPolyfill());
  * @throws {TypeError} if either constructor is null, or if super constructor
  *     lacks a prototype.
  */
-exports.inherits = util.inherits;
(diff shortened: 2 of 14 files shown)


package-scripts.js
```diff
@@ -23,8 +24,25 @@ function test(testName, mochaParams) {
...
   scripts: {
     build: {
-      script: `browserify -e browser-entry.js --plugin ./scripts/dedefine --ignore './lib/cli/*.js' --ignore 'chokidar' --ignore 'fs' --ignore 'glob' --ignore 'path' --ignore 'supports-color' -o mocha.js`,
-      description: 'Build browser bundle'
+      default: {
+        script: 'nps build.tsc build.copy build.browserify',
+        description: 'Build'
+      },
+      tsc: {
+        script: 'tsc -p . || true',
```
weswigham: FYI, you don't need to compile with TS. You can set `noEmit: true` or `emitDeclarationsOnly: true` and avoid transpiling the JS. I imagine that's better for keeping the project simpler?

cspotcode: Yeah, I wanted to give it a try just to see if everything still works, but you're right.  I'll make it clearer that we don't need to do this.

lib/utils.js
```diff
@@ -29,7 +29,10 @@ var assign = (exports.assign = require('object.assign').getPolyfill());
  * @throws {TypeError} if either constructor is null, or if super constructor
  *     lacks a prototype.
  */
-exports.inherits = util.inherits;
+exports.inherits = function(ctor, superCtor) {
+  // Wrapper prevents type declarations from referring to @types/node
```
weswigham: You could also just add a `/** @type {whatever} */` annotation onto this declaration to override the inferred type, instead of changing the implementation.

lib/cli/options_t.ts
```diff
@@ -0,0 +1,7 @@
+export interface MochaArgs {
+  /** Path to `package.json` or `false` to skip */
+  config?: string | false;
+  package?: boolean;
+  /** Filepath to mocha.opts; defaults to whatever's in `mocharc.opts`, or `false` to skip */
+  opts?: string | false;
+}
```
boneskull: please do this with a `@typedef`; I really don't want to add any typescript sources.

boneskull:
needs conflicts to be resolved

voxpelli:
I like this idea, I think this is something to move forward with

JoshuaKGoldberg:
👋 coming back to this @cspotcode, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

PR #4173, opened 2020-02-01
docs: document Node.js' --enable-source-maps flag

bcoe:
## Description of the Change

Newer versions of Node.js introduce the `--enable-source-maps` flag which, when enabled, caches source-maps and provides accurate stack traces to users running transpiled code, e.g., TypeScript.

## Benefits

Many users of Node.js are writing TypeScript, this allows them to receive better stack traces in mocha without installing any dependencies.

## Drawbacks

There's still work to be done to polish this feature in Node.js (would very much appreciate getting feedback).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -37,7 +37,7 @@ Mocha is a feature-rich JavaScript test framework running on [Node.js][] and in 
 - [easily meta-generate suites](#markdown) & [test-cases](#list)
 - [config file support](#-config-path)
 - [mocha.opts file support](#-opts-path)
-- [source-map support](#-enable-source-maps)
+- [source-map support](#--enable-source-maps)
 - clickable suite titles to filter test execution
 - [node debugger support](#-inspect-inspect-brk-inspect)
 - [node native ES modules support](#nodejs-native-esm-support)
--- a/docs/index.md
+++ b/docs/index.md
@@ -36,6 +36,8 @@ Mocha is a feature-rich JavaScript test framework running on [Node.js][] and in 
 - [auto-exit to prevent "hanging" with an active loop](#-exit)
 - [easily meta-generate suites](#markdown) & [test-cases](#list)
 - [config file support](#-config-path)
+- [mocha.opts file support](#-opts-path)
+- [source-map support](#-enable-source-maps)
 - clickable suite titles to filter test execution
 - [node debugger support](#-inspect-inspect-brk-inspect)
 - [node native ES modules support](#nodejs-native-esm-support)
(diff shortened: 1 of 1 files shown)


juergba:
@bcoe thank you for this PR.

We do not whitelist or document NodeJs options in our documentation. There are just a few exceptions like `--inspect`/`--inspect-brk` or `--experimental-modules`.
IMO we should not start doing this, it's abottomless task.
In our repo [mocha-examples](https://github.com/mochajs/mocha-examples) we offer some working samples incl. TS setups, maybe that's the place to recommend the  `--enable-source-maps` flag.

bcoe:
> IMO we should not start doing this, it's a bottomless task.

I understand 100% where you're coming from. 

Worth considering though, something like 60% of folks writing JavaScript these days spend at least some time writing TypeScript. Rather than just documenting this flag, it might be worth having a section of docs that speaks to making mocha work well with TypeScript:

1. speaking to which reporters work well with TypeScript (I believe some have source-map-support as a dependency?).
2. documenting the `-enable-source-maps` flag.
3. speaking to the `source-map-support` library.

boneskull:
I think this is worth adding.  I realize that documenting node flags is a slippery slope (as @juergba wrote, there's no end to it), but I think we should take it on a case-by-case basis.  It's plainly useful information, and I'd like to see more of this on the docs site.  That said, we _should_ add to the examples repo.

@bcoe can you please resolve the conflict?

docs/index.md
```diff
@@ -1082,6 +1083,19 @@ Specify an explicit path to a [`package.json` file](#configuring-mocha-nodejs) (
 
 By default, Mocha looks for a `package.json` in the current working directory or nearest ancestor, and will use the first file found (regardless of whether it contains a `mocha` property); to suppress `package.json` lookup, use `--no-package`.
 
+### `--enable-source-maps`
+
+> _New in Node.js v12.12.0_
+
+If the [`--enable-source-maps`](https://nodejs.org/dist/latest-v12.x/docs/api/cli.html#cli_enable_source_maps) flag
+is passed to Node.js, source-maps will be collected and used to provide accurate stack traces for transpiled code:
```
boneskull: this should be "source maps will be collected...", I think.

boneskull: maybe "is passed to `mocha`" as well, to avoid confusion about how to use.

bcoe:
@boneskull 👍 will try to get this done ASAP.

boneskull:
@bcoe The build is [failing](https://app.netlify.com/sites/mocha/deploys/5e79929d15d38f000b767897) because there's an incorrect anchor link.  Also, @outsideris had a request

bcoe:
@boneskull done 👍

PR #4175, opened 2020-02-04
Remove configuration via 'mocha.opts'

juergba:
### Description

The configuration via `mocha.opts` has been deprecated in Mocha v7.0.0.
In Mocha v8.0.0 we remove it.

### Description of the Change

- we remove corresponding code, unit and integration tests
- we adapt docs
- Mocha does not complain, when an unknown/mistyped option is used. Many user do not read our documentation/release notes, therefore we throw an error when user sets `--opts` option.


juergba:
coverage: the report shows a decrease of -0.06% in _lib/options.js_. When studying the details there is not one uncovered line of code, and the three newly added lines are covered by tests.
So IMO no need for action.

PR #4184, opened 2020-02-16
Support .mocharc.cjs

emma-borhanian:
### Description of the Change

This allows for `"type": "module"` in package.json while also using a
mocha config file, which was previously impossible.
See: https://nodejs.org/api/esm.html

Compare to https://github.com/facebook/jest/issues/9086

### Alternate Designs

N/A

N/A

### Benefits

`.mocharc.cjs` works

### Possible Drawbacks

N/A

### Applicable issues

https://github.com/mochajs/mocha/pull/4038

### DIFF
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -22,6 +22,7 @@ const findUp = require('find-up');
  */
 exports.CONFIG_FILES = [
   '.mocharc.js',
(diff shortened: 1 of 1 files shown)


emma-borhanian:
Nvm somehow I missed it's in here https://github.com/mochajs/mocha/pull/4038

Issue #2456, opened 2016-08-28
🐛 Bug: before() and after() hooks run in a test file with only skipped tests


fabiosantoscode:
I recently stumbled upon this. A test file containing the following:

```
describe('something', () => {
  before(() => { console.log('I should really not run') })
  it.skip('skipped test', () => {})
  after(() => { console.log('I should really not run either') })
})
```

Will produce the following output:

```
  something
I should really not run
    - skipped test
I should really not run either


  0 passing (8ms)
  1 pending
```

The workaround is to skip the entire `describe()` instead of the individual tests, so I wonder if there's interest in fixing this. Should a `before` or `after` function be preparing or cleaning up a test which didn't run?

Thanks for this amazing piece of software :heart:

boneskull:
This looks like an edge case... maybe not an easy fix either.  We'd have to ensure that *nowhere* within a suite, at *any* nested level, there's a runnable test.  

I'd accept a PR for this, but the solution would have to be pretty clean.

lucasfcosta:
Hi @boneskull, I've been reading mocha's code since I've also been dealing with Chai and I think I can make this pretty clear and straightforward.
Would you mind if I send an experimental PR in the next few days so we can talk about it?
Thanks in advance and keep on rocking, mocha is awesome :smile:

mark-wiemer:
It's been a very long time since this issue was reviewed, marking as "in triage" to confirm this is still the case before we accept PRs.

fabiosantoscode:
I can no longer reproduce with Mocha v11.7.5

Much love <3

PR #4189, opened 2020-02-22
Do not call before() and after() hooks in a completely skipped test file 

majecty:
# Description of the Change

I fixed the issue https://github.com/mochajs/mocha/issues/2456.
Currently, although all tests in a file are skipped, the before and after hooks are called.
This PR makes mocha does not call before and after hooks if there is no test in a file.

Changes:
* I modified the `runSuite` function not to call hooks.
* I changed previous tests that depend on the previous behavior. They were expecting Mocha to call hooks even though there is no runnable test.
* I added a test to show the effect of this PR.

### Alternate Designs

1. Calling `runTests` and `runSuite`

I made the `runSuite` function not to call HOOK_TYPE_BEFORE_ALL, and HOOK_TYPE_AFTER_ALL hooks. The `runSuite` function calls `runTests` and `runSuite` function for its tests and child suites. Since there is no runnable test, it is okay not to run the `runTests` and `runSuite` functions.

2. Handling grep

When a user uses grep option and there is no matching test in a file, the before() and after() hooks are called. Should the mocha call the before() and after() hooks in that case?

Calling hooks are managed by the core.

### Applicable issues

If we merge this PR, the mocha's behavior will be change when the mocha run a test file that does not contain runnable tests.


boneskull:
> This PR makes mocha does not call before and after hooks if there is no test in a file.

It's maybe more accurate to say "if there are only skipped tests in a file".  Just because a file has no tests doesn't mean that the hooks in that file should not run.  A file containing only top-level hooks is completely reasonable.

boneskull:
Actually...

I can envision a scenario in which this would _not_ be desirable.

Say you have `a.spec.js`, and it contains top-level hooks:

```js
// a.spec.js

beforeEach(function() {
  // setup
});

describe('stuff', function() {
  it('should work', function() {
    // assertions
  });
});

afterEach(function() {
  // teardown
});
```

You also have `b.spec.js`:

```js
// b.spec.js

describe('something', function() {
  it('should work', function() {
    // assert
  });
});

describe('something else', function() {
  it('should work', function() {
    // assert
  });
});
```

You need your top-level hooks in `a.spec.js`, but you only want to run the "something else" tests:

```bash
$ mocha --fgrep "something else" --file a.spec.js b.spec.js
```

I would expect the top-level hooks in `a.spec.js` to be run, even though the "stuff" suite is not.  Does this work?  If not, this PR should be changed to ensure top-level hooks are always run, even if all tests in a file containing top-level hooks are skipped.

majecty:
@boneskull 
Thanks for your review. I'm using Mocha in my projects. I always feel thanks to the contributors.

> It's maybe more accurate to say "if there are only skipped tests in a file".

I agree with you. I'll fix the phrase "if there is no test in a file" to "if there are only skipped tests in a file".

> If not, this PR should be changed to ensure top-level hooks are always run, even if all tests in a file containing top-level hooks are skipped.

I understand what you mean. You mean that Mocha should run top-level hooks even though all tests are skipped.

Actually, my PR make Mocha not to run top-level tests if there are only skipped tests.

I'll change my PR like below:
* Run top-level hooks always.
* Other hooks will not run if there are only skipped tests the suite.
* A test can be skipped for two reasons. One is that a user explicitly called `.skip` method for the test. The other one is that the test couldn't be passed the "grep" filter.

I'd appreciate it if you would check my direction is right or not.

Thanks!

boneskull:
@majecty 

> Run top-level hooks always.

Yes.  Even in files where we skip _everything else_.

> Other hooks will not run if there are only skipped tests the suite.

Yes.

> A test can be skipped for two reasons. One is that a user explicitly called .skip method for the test. The other one is that the test couldn't be passed the "grep" filter.

`--grep`, `--fgrep` (which is just "grep" under the hood), and `.skip` on a test (`it.skip()`) _or_ on a suite (`describe.skip()`).  tests can also be skipped at runtime via `this.skip()`, but that may not be applicable here.

greg-hoarau:
@boneskull, @majecty Do you have any news about this feature?

Thank you for your work :pray:

PR #4194, opened 2020-02-28
fix: check if module.paths really exists

### LINKED ISSUE #2505: We shouldn't rely on the existence of "module.paths" (Meteor)

ematipico:
### Requirements

### Description of the Change
`module.paths` was not defined when Mocha was run in an serverless environment. Probably due to the absence of `node_modules` folder. I had bundled the test using a bundler.

I added a check to make sure it doesn't throw an error.

### Alternate Designs
Not sure if there are any other alternatives but open to suggestions.

I don't think userland should mock `module.paths`.

### Benefits
It would not throw an error during the execution of tests in a serverless environment.

### Possible Drawbacks
I don't see any.

### Applicable issues
Fixes #2505 
This should be a bug fix

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -30,7 +30,9 @@ const {repository, homepage, version, gitter} = require('../../package.json');
 exports.main = (argv = process.argv.slice(2)) => {
   debug('entered main with raw args', argv);
   // ensure we can require() from current working directory
-  module.paths.push(process.cwd(), path.resolve('node_modules'));
+  if (typeof module.paths !== 'undefined') {
+    module.paths.push(process.cwd(), path.resolve('node_modules'));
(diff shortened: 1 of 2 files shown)


outsideris:
Serverless means AWS Lambda or Google Cloud Functions?

ematipico:
In my particular case, AWS Lambda

Issue #2839, opened 2017-05-27
run Mocha's tests concurrently


boneskull:
I don't see why we couldn't leverage something like [concurrently](https://npm.im/concurrently) to run our Node.js-based tests concurrently, since we have to invoke Mocha many different times for each type of test.

osalama7:
We run integration tests, unit tests with Mocha and it takes up to 36 mins to finish all the tests, Which is primarily an issue with how the tests were written in the first place, discrete, and repeating the same operations in different tests
Solutions include a more outlined test sets, with possible utilisation of global variables available for similar tests, and finally how to run tests in parallel, Also would like to follow up on this issue, and appreciate any input to help reduce runtime of tests

ORESoftware:
This is harder to implement than it sounds.

I have this shell script utility that runs shell commands in parallel, waits for all to exit, and exits with code 0 only if all subshells exit with 0

https://github.com/ORESoftware/generic-subshell

bajtos:
We are interested in this feature for LoopBack too. Our main development is happening in https://github.com/strongloop/loopback-next, a medium-sized monorepo.

Few comments from the experiments I did so far using https://www.npmjs.com/package/mocha-parallel-tests:

- The module uses worker threads on Node.js 12+ and child processes on older versions. Are failing when run via worker threads, because our CLI tests are calling `process.chdir`, which is not supported in worker threads. `process.chdir` is changing the current working directory, which is needed because of the way how Yeoman works. (Our CLI is implemented via yeoman generators, unfortunately.)

- LoopBack is very expensive to load. At the moment, our Todo example application takes `1.4s` to `2s` to load. If you spawn a new child process of every test file, then the tests will spend most of the time inside require. I have verified this via mocha-parallel-tests, where our tests took several times longer to complete when run in parallel.

- Regarding node-tap: That framework is running each test file in a new process from the beginning, so the users must have already accommodated for the fact that every test file is going to load their module from fresh. Switching from serial to parallel execution is less disruptive.

For Mocha and loopback-next, it would be ideal if the same child process could be re-used to run multiple test files. For example, the "master" process can spawn multiple child processes and then fan out test files to run to children one-by-one. When a child finishes running a test file, it can signal the results to master and master can send it another file to run.

By re-using child processes this way, I think there will be less (if any) need to switch from child processes to worker threads, because there will be a fixed & small number of child processes to spawn at the beginning.

[...]

PR #4198, opened 2020-03-03
[DO NOT MERGE] WIP: run tests concurrently; see #2839

boneskull:
~~This is **experimental**, **untested**, and **broken**.  I'm opening it to track progress.~~

~~**This is no longer experimental, broken, nor untested, but it _does_ need to be split into separate PRs.  This PR will not get merged, but I will reference the new PRs here.**~~

**NEW PRS:**

This PR has been split into many separate ones:

- [ ] #4235 - enable emoji in docs
- [ ] #4236 - better debug output (_optional_)
- [ ] #4237 - better errors when validating plugins
- [ ] #4238 - fix nyc ignorelist (_optional_)
- [ ] #4239 - signal handling fixes
- [ ] #4240 - various test fixes
- [ ] #4241 - test helper improvements
- [ ] #4242 - Travis-CI changes; add Node.js v14 to matrix (_optional_)
- [ ] #4243 - wallaby config updates (_optional_)
- [ ] #4244 - Root Hook Plugin support
- [ ] #4245 - Parallel test support
- [ ] #4246 - Docs for Root Hook Plugins and parallel tests

Everything _not_ marked _optional_ is required to land this one way or another.  Please follow the individual PRs for further discussion!

* * *

It does, however, achieve a 50% speedup over our unit tests.

Things that need to happen:

- [x] Determine the bounds for what we can expect to work. I do not expect we can run our entire test suite this way.
- [x] If it's clear something should work but does not, we might need to swap out the worker threads for child processes.  It shouldn't be too terribly difficult to create an adapter to allow operation either way.  Worker threads will be more performant, if they can prove themselves widely suitable.  However, there's stuff [you can't do](https://nodejs.org/api/worker_threads.html#worker_threads_class_worker) with worker threads--and [nyc](https://npm.im/nyc) might not work with them (need to test this).  
- [x] Decide whether to ship both worker and child-process implementations.  If we can pull it off with a slick adapter, I think we should.  Probably default to "worker thread mode" unless it's just too hinky.
- [x] Once we feel confident in the implementation, run LoopBack's test suites in parallel, and maybe TypeScript's.  See if we can uncover any other problems.
- [x] Tests -- our own test runs should prove this out pretty well, but will still need some basic integration & unit tests. Check it works with ESM. Windows.
- [x] Documentation.  Probably necessitates its own section in the docs, and will need a clear explanation of the limitations.  Need to check generated API docs as well.
- [x] Verify browser build doesn't ship this stuff.
- [x] Use with "watch" mode.   "Only rerunning changed code" and other enhancements to watch mode are out-of-scope.

Future, once this ships:

- [ ] Party!! 🎉 🎈 
- [ ] The abstraction we're using, [workerpool](https://npm.im/workerpool), purports to work in a browser. Can this work in a browser for Mocha?


boneskull:
status:

Ensuring the event data (`Test`, `Suite` and `Hook` instances) can be communicated to the main process via the structured clone algorithm. This needs to be done as to not break our contract with reporters, since they expect these objects; these objects must have the correct prototype and references.

boneskull:
update: (de-)serialization is too slow; when doing this, the parallel runs are slower against our unit test suite than serial.

the most performant solution would be to restrict what reporters can expect; e.g., not emit an entire `Test` object, but rather one that's a duck-typed subset.  This would need to be documented, and should be considered a breaking change to the API.  We'd still need implementations of certain functions known to be used, like `fullTitle()`, to avoid _completely_ breaking third-party reporters.

boneskull:
update: I've decided to abandon workers for now.

The main problem with them is that worker threads _are not currently debuggable_.  This makes it both difficult to implement, and will be hostile to any user trying to debug tests. See nodejs/node#26609.

boneskull:
A potential "gotcha" with these tests is that the duration of an individual test seems to have more variability if that test performs I/O (depending on the number of concurrent jobs).  If you are running, say, 7 jobs in parallel, and you have 8 cores, there's not necessarily any "extra" cores for libuv's thread pool, so the I/O sits for longer waiting to get processed.

but, good news:  using child processes (instead of workers), our Node.js integration tests are passing.  When run with four (4) concurrent jobs, the execution time is cut **from 2m23s to 46s**.
 
There are some rough edges, and I suspect certain reporters will not work very well with this (instead of seeing the result of each test as it finishes, the results of each file are buffered and dumped at once).  Also, I haven't yet written tests specifically intended to run in parallel, which should surface some issues with certain options (`--allowUncaught`, `--delay` & `--retries` are dubious).

boneskull:
unbuffered output would make sense for certain reporters (dot, min, tap) but not the default (spec) because it'd be impossible to tell what was going on due to the indenting.  unbuffered output seems like a good candidate for a future PR

boneskull:
from what I can tell, there isn't a noticeable speedup when using child processes vs worker threads.  

I'm not sure if running `mocha` vs `_mocha` with concurrent jobs makes a significant different in performance.

maybe we should add some benchmark tests (but not in this PR)

boneskull:
we need to increase the global default timeout from 200ms.  running tasks concurrently in CI means that there's much more variability around how long any given test will take (see AppVeyor builds)

boneskull:
thoughts on how `--file` should work:

`--file` causes test files to be loaded in a specific order.  The difference between `mocha --file a.spec.js "*.spec.js"` and `mocha "*.spec.js"` is that we are guaranteed that `a.spec.js` will always be executed first; otherwise the order in which files are executed is dependent on either `glob` or your shell's wildcard expansion.

When running tests with `--parallel`, we have even fewer guarantees of the order in which files are executed; the execution time of an individual file takes _also_ becomes a variable.

Why would we care about running files in order in the first place?  Why is `--file` a thing?

This is because in `a.spec.js`, above, we might have _top-level hooks_; hooks that are attached to the _root suite_.  A hook attached to the root suite will execute across _all subsequently-loaded files_.  This means you can do some "global" setup in a "before each" or "before all" hook.  If your suites themselves do not depend on async behavior to programmatically generate--but all your _tests_ need async setup--this is a more appropriate solution than using `--delay`.

The current implementation of `--parallel` _does not have a single root suite_.  Each _file_ has its own root suite.  This breaks the above use case of `--file`.

To solve this:

- Any test file specified by `--file` will _not_ run in parallel.  These must run in serial.
- I'll look into sharing (or cloning, or otherwise reproducing) the root suite in its state after the above serial run.

This surfaces a couple holes in our integration tests.  Note that they all currently pass on Travis CI!

boneskull:
after looking at this a bit, I don't think there's a way to make `--file` work in a way that's akin to what we have now (in a way that I'd want to implement).

instead of `--file`, if someone needs this sort of behavior, you're going to need some boilerplate:

```js
// root-hooks.js

beforeEach(async function() {
  await somethingAsync();
});
```

```js
// test.spec.js
require('./root-hooks');

describe('my thing', function() {

});
```

boneskull:
it's truly annoying that it's not easy (I don't know how to do it!) to re-use a `Mocha` instance by removing files and adding new ones after a call to `mocha.run()` has completed.  I'm not entirely sure what the problem is, but something certainly breaks.

giltayar:
@boneskull a thought: when running in parallel, and with `--watch`, you will probably remove the need for clearing the module cache. If this becomes the way to go, even for watch with non-parallel execution, then that solves the ES module cache problem (the problem is that there is no mechanism to clear the ES module cache).

juergba:
The root-suite before/after hooks problematic is not only limited to the `--file` option. Root hooks can be declared in any of the test files, independently of its index. Since you don't know the content of those hooks, you can't decide wether to execute them once per test run or once for each sub-process.
@bajtos already mentioned this problem.

boneskull:
OK.  Since:

1. The behavior of root-level hooks are dependent upon the order in which test files are run, and
1. `--parallel` will run test files in a nondeterministic order,
1. We cannot support `--file`

We _can_ support root-level hooks still, but as this PR is written, they will only be applicable to the tests within the file in which they are defined.

boneskull:
@giltayar Good insight.  That'd make getting `--watch` running with this more valuable.  I'll investigate further.

OK, so this is looking pretty solid.  Next steps:

- [ ] look at watch mode
- [ ] write docs

This is going to be a pretty significant thing to review...

boneskull:
Note that we're likely to see this message in CI:

```
(node:6291) Warning: (Mocha) not enough CPU cores available (2) to run multiple jobs; avoid --parallel on this machine
```

This is true; we only have one core for the main process, and another for a single worker process.  This makes things slightly slower, but I think it's important to run our test suite in parallel mode.  That said, we should also not ignore serial mode, so I think a serial run of the unit tests as a new build job is appropriate.

boneskull:
I've updated the description with a list of new PRs supporting this feature.  Please refer to those; I'm going to close this.

PR #4164, opened 2020-01-22
Drop node.js v8 support

UlisesGascon:
### Description of the Change

I removed support to v8 in CI and documentation as explained in #4163

### Alternate Designs

N/A 

### Why should this be in core?

Bump minimum Nodejs version

### Benefits

No need to worry about support for Node v8

### Possible Drawbacks

N/A

### Applicable issues

 Close #4163

PR #4200, opened 2020-03-14
Drop mkdirp and replace it with fs.mkdirSync

### LINKED ISSUE #4199: Replace vulnerable mkdirp with fs.mkdir

### Description

Mocha does depend on `mkdirp`, which depends on an old version of `minimist`, which has a known Prototype pollution vulnerability: https://snyk.io/test/github/mochajs/mocha

It does not seem that `minimist` is getting updated soon, and also, `fs.mkdir(path[, options], callback)` can be run recursively, effectively obsoleting `mkdirp`. Furthermore, packages depending on any version of `mocha` currently exhibit this vulnerability (cf. any package on snyk.io), and are shown to the user as vulnerable, too.

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 7.1.0
- The output of `node --version`: v13.6.0
- Your operating system
  - name and version: Windows 10
- Your shell (e.g., bash, zsh, PowerShell, cmd): PS

HyunSangHan:
### Description of the Change
- Replace `mkdirp` with `fs.mkdirSync` using `{recursive: true}`
- Drop the dependency `mkdirp` from Mocha

### Alternate Designs
There is also a way to update to the latest `mkdirp` version, but Node.js version 10.12.0 has already added a native support for `mkdirSync` to create a directory recursively with `{recursive: true}` option as the following:
```js
fs.mkdirSync(pathname, { recursive: true });
```

So, there is no longer necessary to depend on third-party packages. `fs` module is enough to us.

`mkdirp` depends on an old version of `minimist`, and it has a prototype pollution vulnerability.

### Benefits
Can create a new directory and any necessary subdirectories at the directory **without mkdirp**.

### Possible Drawbacks
Can't think of any.

### Applicable issues
I think it have to be released with **`semver-major`(maybe v8.0.0)**, because the `{recursive: true}` option of `fs.mkdir` is supported by Node.js v10.12.0 or later.

Closes #4199

### DIFF
--- a/test/reporters/xunit.spec.js
+++ b/test/reporters/xunit.spec.js
@@ -4,7 +4,6 @@ var EventEmitter = require('events').EventEmitter;
 var fs = require('fs');
 var os = require('os');
 var path = require('path');
-var mkdirp = require('mkdirp');
 var rimraf = require('rimraf');
 var sinon = require('sinon');
 var createStatsCollector = require('../../lib/stats-collector');
@@ -51,12 +50,12 @@ describe('XUnit reporter', function() {
     };
 
     describe('when fileStream can be created', function() {
-      var mkdirpSync;
+      var fsMkdirSync;
       var fsCreateWriteStream;
 
       beforeEach(function() {
         sandbox = sinon.createSandbox();
-        mkdirpSync = sandbox.stub(mkdirp, 'sync');
+        fsMkdirSync = sandbox.stub(fs, 'mkdirSync');
         fsCreateWriteStream = sandbox.stub(fs, 'createWriteStream');
       });
 
@@ -67,7 +66,13 @@ describe('XUnit reporter', function() {
         XUnit.call(fakeThis, runner, options);
 
         var expectedDirectory = path.dirname(expectedOutput);
-        expect(mkdirpSync.calledWith(expectedDirectory), 'to be true');
+        expect(
+          fsMkdirSync.calledWith(expectedDirectory, {
+            recursive: true
+          }),
+          'to be true'
+        );
+
         expect(fsCreateWriteStream.calledWith(expectedOutput), 'to be true');
       });
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -42,7 +42,7 @@ Follow these steps to get going. If you are having trouble, don't be afraid to [
 
 > PRO TIP: After `npm install`, run `npm start` to see a list of commands which can be run with `npm start <command>` (powered by [nps](https://npm.im/nps)).
 
-1. [Install Node.js 10.x or newer](https://nodejs.org/en/download/).
+1. [Install Node.js 10.12.0 or newer](https://nodejs.org/en/download/).
    - If you're new to installing Node, a tool like [nvm](https://github.com/creationix/nvm#install-script) can help you manage multiple version installations.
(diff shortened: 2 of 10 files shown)


fabiosantoscode:
An alternative implementation would be to update to the latest `mkdirp` version, which is not vulnerable.

HyunSangHan:
@fabiosantoscode 
Thank you for your comment.
That's also one of the valid methods, but Mocha is ready to drop NodeJS version v8 support (ref: #4164). After v8.0.0 release, I think, Mocha can use `fs.mkdirSync` so no longer depend on `mkdirp`.

fabiosantoscode:
Uh oh :) I still support node 8 in my library.

bigman73:
Updating mkdirp seems like an easy fix without breaking backward compatibility

JimmyBjorklund:
Adding this patch or upgrading to the latest mkdirp ( do not use minimst) is relevant to:
https://github.com/advisories/GHSA-7fhm-mqm4-2wp7

juergba:
The latest version of _mkdirp_ supports Node >=10. It seems we have to drop Node v8 anyway.

bigman73:
nodejs 8 is not supported as of 2020, so indeed why bother.

test/reporters/xunit.spec.js
```diff
@@ -43,20 +42,20 @@ describe('XUnit reporter', function() {
   });
 
   describe("when 'reporterOptions.output' is provided", function() {
-    var expectedOutput = path.join(path.sep, 'path', 'to', 'some-output');
```
juergba: Why have you had to remove `path.sep`?

SebastianSpeitel:
> An alternative implementation would be to update to the latest `mkdirp` version, which is not vulnerable.

> Updating mkdirp seems like an easy fix without breaking backward compatibility

There is no update of mkdirp that fixes this.
The [PR](https://github.com/isaacs/node-mkdirp/pull/8) isn't even merged yet.

test/reporters/xunit.spec.js
HyunSangHan: I thought it is not necessary and it causes the test to fail.
But now I know it was my misunderstanding. I put it back now.
Thank you for your feedback.

pezzullig:
Great work on getting this PR done! 👍 Any chance of getting this reviewed and tagged any time soon?

Hypnosphi:
> An alternative implementation would be to update to the latest `mkdirp` version, which is not vulnerable.

Or even just to `0.5.3`, to make the change minimal:

```
% npm info mkdirp@0.5.3 dependencies 
{ minimist: '^1.2.5' }
```

juergba:
I will update today to _mkdirp@0.5.3_ and publish as Mocha v7.1.1. This version is new and deprecated, but seems to fix this security issue.
_yargs-parser_ is also reported to have security issues, so I will update to _yargs@13.3.2_ and _yargs-parser@13.1.2_ .

@HyunSangHan this PR remains valid for Mocha v8.0.0.

test/reporters/xunit.spec.js
juergba: I got it finally. It works both ways, with or without `path.sep`. The tests don't check for the value of `expectedOutput`.

HyunSangHan:
@juergba 
I did rebase and squash just now. Please let me know if there is something wrong. :-)

PR #4218, opened 2020-03-25
fix for issue 4047

LeonardoSalvucci:
CLI will verify if args passed with --file exists and if not just ignore them and show a warning.


outsideris:
#4047

PR #4219, opened 2020-03-25
Exposing filename in JSON, doc, and json-stream reporters

Daniel0113:
### Description of the Change

This change makes the `json`, `doc`, and `json-stream` reporters output the test filenames in addition to their other attributes 

### Alternate Designs

No other designs were considered in the making of this pull request

There currently is not a way to access filenames through reporters when running tests even though mocha has full access to it. 
### Benefits

People who want to see what file is associated with their test step will easily be able to see it in the `json`, `doc`, and `json-stream` reporters. This is especially useful for people with large code/test base where it may be difficult to find the test. 

### Possible Drawbacks
If we intend to encapsulate the test's filename, then this code change would be an issue

### Applicable issues
This would be considered a minor release.


Daniel0113:
@outsideris I have added filename exposure to the `doc` and `json-stream` reporters (and updated tests for all the reporters I touched). Unless there are other reporters we have in mind, this is ready for a second look 😄

Daniel0113:
Given that I also can't think of a test that _isn't_ located in a file, I'm good with merging as-is then 😄 . If anyone has strong opinions on the matter though I am open to that.

PR #4223, opened 2020-04-06
Improve hook pattern of 'this.skip()' in beforeAll hooks

juergba:
### Description

```js
describe('outer suite', function() {
  before(function() {
    console.log('** before-1');
    this.skip();
  });
  before(function() { console.log('** before-2'); });  // second before hook should not run

  it('outer pending test', function() { });

  describe('inner suite', function() {
    before(function(done) {              // inner async before hook should not run
      console.log('**** before: done');
      done();
    });

    it('inner pending test', function() { });

    after(function(done) {               // inner async after hook should not run
      console.log('**** after: done');
      done();
    });    
  });

  after(function() { console.log('** after'); });
});
```
```
  outer suite
** before-1
** before-2                  // second before hook should not run
    - outer pending test
    inner suite
**** before: done            // inner async before hook should not run
      - inner pending test
**** after: done             // inner async after hook should not run
** after

  0 passing (16ms)
  2 pending
```
### Description of the Change

- since `this.skip()` aborts hook execution, subsequent `before` hooks on the same suite level are not executed anymore.
- inner `before`/`after` hooks: explicit async hooks with `done` argument are not executed anymore.
Note: sync or promise-returning hooks work correctly, no changes are required.
- `test.state`: is set to `pending`.
We have three `state`'s now: `failed`, `passed`, `pending`

### Applicable issues

related #3225

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -295,6 +295,8 @@ Runnable.prototype.run = function(fn) {
   var finished;
   var emitted;
 
+  if (this.isPending()) return fn();
+
   // Sometimes the ctx exists, but it is not runnable
   if (ctx && ctx.runnable) {
     ctx.runnable(this);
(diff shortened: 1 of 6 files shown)


teddibaer:
It seems to me, that the current behavior of mocha regarding calling this.skip inside of before() hooks is not correctly reflected in the documentation, where it states that:

"before/after hooks are skipped unless they are defined at the same level as the hook containing this.skip()"

PR #4225, opened 2020-04-07
Add reporter options example

pkuczynski:
### Description of the Change

I tried to add reporter options as `reporterOptions` or `reporterOption` and failed miserably for quite a while until I accidentally found out that it needs to be specified like in the example below.

### Benefits

I think it's worth adding an example to the documentation to save other people's time.

### DIFF
--- a/example/config/.mocharc.js
+++ b/example/config/.mocharc.js
@@ -1,17 +1,47 @@
 'use strict';
 
-// Here's a JavaScript-based config file.
+// This is a JavaScript-based config file containing every Mocha option plus others.
 // If you need conditional logic, you might want to use this type of config.
 // Otherwise, JSON or YAML is recommended.
 
 module.exports = {
+  'allow-uncaught': false,
+  'async-only': false,
+  bail: false,
+  'check-leaks': false,
+  color: true,
+  delay: false,
   diff: true,
+  exit: false, // could be expressed as "'no-exit': true"
   extension: ['js'],
+  // fgrep: something, // fgrep and grep are mutually exclusive
+  file: ['/path/to/some/file', '/path/to/some/other/file'],
+  'forbid-only': false,
+  'forbid-pending': false,
+  'full-trace': false,
+  global: ['jQuery', '$'],
+  // grep: something, // fgrep and grep are mutually exclusive
+  growl: false,
+  ignore: ['/path/to/some/ignored/file'],
+  'inline-diffs': false,
+  // invert: false, // needs to be used with grep or fgrep
+  jobs: 1,
   package: './package.json',
+  parallel: false,
+  recursive: false,
   reporter: 'spec',
-  slow: 75,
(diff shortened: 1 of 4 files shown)


outsideris:
We know you spent your time to figure this out, but this example configs aren't for explaining every option. 
`--reporter-option` for XUNIT already described in our [docs](https://mochajs.org/#xunit).

pkuczynski:
I know that, but what was surprising for me was the format. I was expecting to pass params like:

```js
reporterOptions: {
     suiteName: "My Tests"',
     output: "./build/tests.xml"
}
```

like its displayed across all the documentation, while instead, mocha expected this:

```js
'reporter-options': [
     'suiteName="My Tests"',
     'output=./build/tests.xml'
   ],
```

This was very confusing for me and I guess it would be good to have it in those examples. Especially your other examples are much larger: https://github.com/mochajs/mocha/blob/master/example/config/.mocharc.yml

I can change the example to be more generic like in `.mocharc.yml` with `foo`, `bar`..

What you think?

pkuczynski:
@juergba @tj @boneskull what do you think?

juergba:
IMO it is rather confusing, see eg. #4142.
In our configuration files `reporter-option` has to be an `array`. When passing it directly to the `Mocha`  constructor, then it is an `object`.

pkuczynski:
My point exactly! There is no clear documentation about it, or at least I could not find it. However there is plenty documentation for passing an object to the constructor. 

That’s why I strongly believe that adding example from this PR would help people...

Validation of mocharc structure would help too, but that’s a separate topic...

outsideris:
I understood what @pkuczynski pointed out. And I realized our [.mocharc.yml](https://github.com/mochajs/mocha/blob/master/example/config/.mocharc.yml) is much verbose than others. 
`reporter-option` is `Array` but each value is not same type with other Array type options. Therefore, IMO it is helpful for mocha newbies.

pkuczynski:
I can do that tonight if you want?

pkuczynski:
I just updated my PR to align with `.mocharc.yml`. The only thing I noticed is that `.mocharc.js` had `package` property which `.mocharc.yml` does not. Not sure if this is a bug?

pkuczynski:
I have not modified anything else than `example/config/.mocharc.js`. Possibly I misunderstood @outsideris comment and added all options from `yaml` to `json`. I personally feel that having all options would make it easier for newbies to understand. Please confirm if I should revert and stick to the simple version was here before?

I suggest updating other files in a separate PR once we align on `.mocharc.js`

pkuczynski:
Build is finally green now :)

boneskull:
thanks!!

boneskull:
@juergba can you pls confirm your comments have been addressed; if so please merge

boneskull:
and by merge i mean "squash"

juergba (approved):

PR #4229, opened 2020-04-15
fix: check process.browser, before access process.stdout

rob2468:
### Description of the Change

In browser context, `process.stdout` is `undefined`.

I implement a custom reporter, when run macha in browser, as the following code show. Error occur in the first line of code, because of the above reason.

```
const Base = require('mocha/lib/reporters/base');

function MochaReporter(runner, options) {
  ...
  // Call the Base mocha reporter
  Base.call(this, runner);
  ...

  // Process the full suite
  runner.on('end', () => {
    try {
        const rootSuite = mapSuites(this.runner.suite, totalTestsRegistered, this.config);
        const obj = {
          stats: this.stats,
          results: [ rootSuite ],
        };
        // Save the final output to be used in the future
        this.output = obj;
      }
    } catch (e) {
    }
  });
}
```

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -25,7 +25,7 @@ exports = module.exports = Base;
  * Check if both stdio streams are associated with a tty.
  */
 
-var isatty = process.stdout.isTTY && process.stderr.isTTY;
+var isatty = !process.browser && process.stdout.isTTY && process.stderr.isTTY;
 
 /**
  * Save log references to avoid tests interfering (see GH-3604).


juergba:
We are shimming `process.stdout` in our bundled _mocha.js_:
`process.stdout = require('browser-stdout')({label: false});`

Our HTML reporter is extending Base as well. So I don't understand why `process.stdout` should be `undefined`.

rob2468:
I got it. The problem lies in the importing order of `mocha` and `reporter`. Need import `mocha` firstly.

Before
```
import 'customer-reporter';
import 'mocha';
```

After
```
import 'mocha';
import 'customer-reporter';
```

PR #4221, opened 2020-04-01
Report skipped tests upon beforeAll hook failure

juergba:
### Description

related: #4233

```js
describe("parent suite", function() {
  before(function() {
    throw new Error("this is a failure in a before hook");
  });

  it("test 1", function() { });
  it("test 2", function() { });

  describe("inner suite", function() {
    it("test 3", function() { });
  });
})
```
**Before**: a failing `before` hook ignores the three tests completely. They disappear and are not even summarised in the epilogue.
```
parent suite
  1) "before all" hook for "test 1"

0 passing (15ms)
1 failing

1) parent suite
     "before all" hook for "test 1":
   Error: this is a failure in a before hook
   [...]
```
**After**:
```
parent suite
  1) "before all" hook for "test 1"
  - test 1
  - test 2
  inner suite
    - test 3

0 / 3 passing (15ms)
3 skipped
1 failing

1) parent suite
     "before all" hook for "test 1":
  Error: this is a failure in a before hook
  [...]
```
 
### Description of the Change

- add a new event `EVENT_TEST_SKIPPED`
- add a new `test.state`: `skipped`
- beforeAll hook: each skipped test is reported and summerised in the epilogue. The output is very similar to `pending` test cases, but in color red.
- there are no changes to the hook pattern, skipped tests will not be executed.
- adapt reporter SPEC and JSON

We now have four test states:
- `passes`
- `pending`: set by the user
  - test without callback `it('some test')`
  - static: `it.skip()` or `describe.skip()`
  - conditional: `this.skip()`
- `skipped`: test not run because of a failing hook
- `failed`: test failed or hook failed

We are aware of `pending` not being an optimal label. Nevertheless we will stick to it for historical reasons, it has been established for many years. 

### Applicable issues

ToDo: 
- `afterEach` hooks
- reporter HTML
 
#1955
#1815

Issue #1955, opened 2015-11-05
🐛 Bug: Hook failures change the shape of the test suite


cowboyd:
If a `beforeEach` hook fails, then all subsequent tests in a suite and all sub-suites are not run. For example:

``` javascript
// hook-test.js
describe("outer context", function() {
  beforeEach(function() {
    throw new Error("this is a failure in a before each hook");
  });
  it("reports the first assertion", function() {

  });
  it("does not report the existence of this test case", function() {

  });
  describe("inner context", function() {
    it("does not report report its existence in the output", function() {

    });
  });

});
```

reports only a single testcase, even though there are three defined:

```
$ mocha --reporter min hook-test.js
0 passing (6ms)
  1 failing

  1) outer context "before each" hook for "reports the first assertion":
     Error: this is a failure in a before each hook
      at Context.<anonymous> (test/mocha-test.js:3:11)
```

As outlined in #1043, this is the intended behavior for both `beforeEach` as well as other hooks. This makes sense from an efficiency prespective; after all, there is no point in actually running the testcases when it is assured that they are going to fail.

The problem with this is that when you're refactoring a large codebase, or doing something like upgrading an underlying library, or rehabilitating a codebase that has allowed its test suite to get out o sync.... work that might take days or weeks, this behavior alters _how many total testcases are reported_. So as you make changes, the reporting varies widely. We've seen the pass/total numbers of test cases jump from `95/220`, to `4/8`, then down to `0/1` in the case of a global hook failure, then back up to `35/160` in a matter of minutes.

This can be very disorienting, and it obscures your overall progress towards your goal which is a completely green suite where **all** tests are passing. The fact that a test is not run, does not mean that it doesn't exist, and that it is not important.

Rather than exclude a test completely from the output, it makes more sense to not run it, but still report it as a failure. That way, the shape of the test suite remains constant. If I have 225 testcases, then that's what I have, even if only 26 of them are passing. I at least know that I'm 12% there, which test suites are related by the same failure, and I can track total progress.

If this makes sense, it would be something I'd be happy to help implement.

cowboyd:
Any thoughts on this? I'm cc/ing @travisjeffery @boneskull and @danielstjules since you all seem to be most active in the issues.

As mentioned, I'd be willing to take a stab at implementing this if I thought that the PR were likely to be accepted. Cheers!

travisjeffery:
if the code to implement the fix isn't too crazy i'd likely accept it

[...]

Issue #1815, opened 2015-07-26
Summary lists skipped tests as "pending" which isn't entirely correct


binarykitchen:
Summary example

```
  27 passing (1m)
  2 pending
  1 failing
```

You know, you can skip tests with `.skip()` and these appear in the summary at the end under the `pending` category which is IMO not correct. Pending != Skipped. Pending can be interpreted as "not completed yet" whereas "skipped" can mean, code is not complete yet or is breaking for whatever reason.

Any chance skipped tests can be summarised under a new group, i.E.

```
  27 passing (1m)
  2 skipped
  1 failing
```

danielstjules:
The final output could be:
```
  27 passing (1m)
  1 skipped
  1 pending
  1 failing
```

Introducing a new state for specs would be a major breaking change, so it'd have to target 3.0.0. I don't think this would be likely to get implemented in the near future. However, I do like the idea of distinguishing between the two. :)

bananu7:
I'd like to voice that I think it's a very important feature. Could anyone with relevant knowledge elaborate on the "breaking changes" mentioned by @danielstjules? Is this something that could be reasonably contributed by an outsider?

mekdev:
+1 I think we need a skipped status, currently we have to triage all "pending" tests and see when one is actually pending because of a failed hook or actually skipped.

ScottFreeCode:
In principle I'm generally for being able to differentiate pending tests from tests that are skipped due to hook failures. (Something breaking in the test itself, on the other hand, should typically be a test failure rather than skipping.) That being said...

- It isn't commonly reported in the JS testing ecosystem in general -- see [here](https://github.com/js-reporters/js-reporters/issues/4#issuecomment-54575585) for a survey at least as of a couple years ago and the resulting spec to this day does not have a separate "pending" status, only a "skipped" status. (If it matters, Mocha doesn't currently implement this spec to my knowledge, but there was interest in updating it to do so.)

- 
> Could anyone with relevant knowledge elaborate on the "breaking changes"...

There are definitely custom (ie third-party) Mocha reporters relying on the current set of statuses. Additionally, there could (although it's both a hairier situation and a less likely one) be users or third-party tools parsing the reporter output that are likewise relying on the current set of statuses.

- 
> Additionally, I could imagine specifying a reason for the test being skipped to be quite useful as well.

[...]

PR #4233, opened 2020-04-21
Report skipped tests upon beforeEach hook failure

juergba:
### Description

required: #4221 

```js
describe("parent suite", function() {
    beforeEach(function() {
      throw new Error("this is a failure in a beforeEach hook");
    });
  
    it("test 1", function() { });
    it("test 2", function() { });
  
    describe("inner suite", function() {
      it("test 3", function() { });
    });
})
```
**Before:** a failing `beforeEach` hook ignores the three tests completely.
```
parent suite
    1) "before each" hook for "test 1"

0 passing (14ms)
1 failing

1) parent suite
     "before each" hook for "test 1":
   Error: this is a failure in a beforeEach hook
   [...]
```
**After:** 
```
parent suite
    1) "before each" hook for "test 1"
    - test 1
    - test 2
    inner suite
      - test 3

0 / 3 passing (20ms)
3 skipped
1 failing

1) parent suite
     "before each" hook for "test 1":
   Error: this is a failure in a beforeEach hook
   [...]
```

### Description of the Change

- add a new event `EVENT_TEST_SKIPPED`
- add a new `test.state`: `skipped`
- beforeEach hook: each skipped test is reported and summerised in the epilogue. The output is very similar to `pending` test cases, but in color red.
- there are no changes to the hook pattern, skipped tests will not be executed.
- adapt reporter JSON-STREAM, LIST, TAP and XUNIT

### Applicable issues

#1955
#1815


docs/index.md
```diff
@@ -449,9 +447,73 @@ setTimeout(function() {
...
+      // will be skipped and reported as 'skipped' test
+    });
+  });
+});
+```
+
 ## Pending Tests
 
-"Pending"--as in "someone should write these test cases eventually"--test-cases are simply those _without_ a callback:
+"Pending" - as in "someone should write these test cases eventually" - test-cases are simply those _without_ a callback:
```
boneskull, line 516: Is this an "EM DASH"?  Ideally we should have something like smartypants that translates `--` to `—`.  Maybe revert this?

docs/index.md
```diff
@@ -462,7 +524,9 @@ describe('Array', function() {
 });
 ```
 
-Pending tests will be included in the test results, and marked as pending. A pending test is not considered a failed test.
+By appending `.skip()`, you may also tell Mocha to ignore a test: [inclusive tests](#inclusive-tests).
```
boneskull, line 527: suggestion:

```md
By appending `.skip()`, you may also tell Mocha to [ignore a test](#inclusive-tests).
```

docs/index.md
```diff
@@ -1641,7 +1705,7 @@ mocha.setup({
 ### Browser-specific Option(s)
 
 Browser Mocha supports many, but not all [cli options](#command-line-usage).
-To use a [cli option](#command-line-usage) that contains a "-", please convert the option to camel-case, (eg. `check-leaks` to `checkLeaks`).
+To use a [cli option](#command-line-usage) that contains a "-", please convert the option to camel-case, (e.g. `check-leaks` to `checkLeaks`).
```
boneskull, line 1708: :+1:

docs/index.md
boneskull: wait... I'm confused now.  `this.skip()` means "test is pending", not "test is skipped"?

I think we could be more clear here.  For your consideration:

- Maybe we can, instead, make `this.skip()` _report a test as skipped_, and `this.pending()` _report a test as pending_.  The behavior would be roughly similar (the test wouldn't get run), but the test would not be displayed as "pending" _unless_ the user called `this.pending()` instead of `this.skip()` (likewise for `describe.skip()`/`describe.pending()`/`it.skip()`/`it.pending()`).  This would be _less of_ a breaking change than just trying to rename `skip()` to `pending()`.
- Maybe instead of calling them "skipped tests", as you introduce above, we can just use a different word.  "Omitted" tests, maybe?  
- This may be a good place to update the terminology we use elsewhere re: "inclusive" (`it.skip()`) and "exclusive" (`it.only()`) tests.  Those terms always felt vague to me.  Instead of "inclusive", we could say "omitted", and instead of "exclusive", we could say "selected" or "picked" or something.

lib/runnable.js
```diff
@@ -40,6 +40,7 @@ function Runnable(title, fn) {
   this._retries = -1;
   this._currentRetry = 0;
   this.pending = false;
+  this.skipped = false;
```
boneskull, line 43: You know, if we have `Test#state`, maybe we should have `Runnable#state` instead, and do away with these booleans...?

lib/runnable.js
```diff
@@ -166,6 +167,17 @@ Runnable.prototype.isPassed = function() {
...
 };
 
+/**
+ * Return `true` if this Runnable has been skipped.
+ * @return {boolean}
+ * @private
+ */
+Runnable.prototype.isSkipped = function() {
+  return (
+    !this.pending && (this.skipped || (this.parent && this.parent.isSkipped()))
```
boneskull, line 177: There should always be a `parent` afaik

lib/runner.js
```diff
@@ -358,11 +363,10 @@ Runner.prototype.hook = function(name, fn) {
   var hooks = suite.getHooks(name);
   var self = this;
 
-  function next(i) {
+  function nextHook(i) {
```
boneskull, line 366: :+1:

lib/runner.js
```diff
@@ -358,11 +363,10 @@ Runner.prototype.hook = function(name, fn) {
...
     var hook = hooks[i];
-    if (!hook) {
-      return fn();
-    }
+    if (!hook) return fn();
```
boneskull, line 368: I much prefer the style as-is, but as long as prettier or eslint doesn't complain, I can live with it.

lib/runner.js
```diff
@@ -410,6 +406,13 @@ Runner.prototype.hook = function(name, fn) {
             suite.pending = true;
           });
           hooks = [];
+        } else if (name === HOOK_TYPE_BEFORE_EACH) {
+          if (self.test) {
+            self.test.pending = true;
+          }
+          self.emit(constants.EVENT_HOOK_END, hook);
+          hook.pending = false;
+          return fn(new Error('abort hookDown'));
```
boneskull, line 415: what's this?

lib/runner.js
```diff
@@ -418,17 +421,40 @@ Runner.prototype.hook = function(name, fn) {
         }
       } else if (err) {
         self.failHook(hook, err);
-        // stop executing hooks, notify callee of hook err
-        return fn(err);
+
+        if (name === HOOK_TYPE_BEFORE_ALL) {
+          suite.tests.forEach(function(test) {
```
boneskull, line 426: would prefer a method on `Suite.prototype` which marked all of its tests (and child suites) as `skipped`, rather than digging into the properties on a `Suite` instance

lib/runner.js
```diff
@@ -418,17 +421,40 @@ Runner.prototype.hook = function(name, fn) {
...
+        } else if (name === HOOK_TYPE_BEFORE_EACH) {
+          (function skipTests(sui) {
+            sui.tests.forEach(function(t) {
+              if (!t.state) t.skipped = true;
+            });
+            sui.suites.forEach(function(s) {
+              if (!s.started) s.skipped = true;
+              else skipTests(s);
+            });
+          })(suite);
```
boneskull, line 442: what's this IIFE for?

lib/runner.js
```diff
@@ -635,32 +648,42 @@ Runner.prototype.runTests = function(suite, fn) {
...
-    self.hookDown(HOOK_TYPE_BEFORE_EACH, function(err, errSuite) {
-      // conditional skip within beforeEach
-      if (test.isPending()) {
-        if (self.forbidPending) {
-          self.fail(test, new Error('Pending test forbidden'), true);
+    self.hookDown(HOOK_TYPE_BEFORE_EACH, function cbHookDown(err, errSuite) {
+      if (err || test.isPending()) {
+        if (test.isPending()) {
+          if (self.forbidPending) {
+            self.fail(test, new Error('Pending test forbidden'), true);
```
boneskull, line 668: this should probably be an error with a `code` as in `lib/errors.js`

lib/runner.js
```diff
@@ -726,8 +747,9 @@ Runner.prototype.runSuite = function(suite, fn) {
   }
 
   this.emit(constants.EVENT_SUITE_BEGIN, (this.suite = suite));
+  suite.started = true;
```
boneskull, line 750: if we have any more booleans on `Suite` like this, we should consider a `state` prop instead

lib/runner.js
```diff
@@ -726,8 +747,9 @@ Runner.prototype.runSuite = function(suite, fn) {
...
 
-  function next(errSuite) {
+  function nextSuite(errSuite) {
```
boneskull, line 752: :+1:

lib/runner.js
```diff
@@ -840,7 +852,7 @@ Runner.prototype.uncaught = function(err) {
 
   runnable.clearTimeout();
 
-  if (runnable.isFailed()) {
+  if (runnable.isFailed() || runnable.isSkipped()) {
```
boneskull, line 855: pls update the `debug` statement here

lib/suite.js
```diff
@@ -62,6 +62,7 @@ function Suite(title, parentContext, isRoot) {
   this.suites = [];
   this.tests = [];
   this.pending = false;
+  this.skipped = false;
```
boneskull, line 65: yeah, so, if it's not possible to have `this.pending && this.skipped === true`, then we should use a `state`

docs/index.md
juergba: The current situation (incl. this PR) is:
- `pending`: set by the user
  - test without callback it('some test')
  - static: it.skip() or describe.skip()
  - conditional: this.skip()
- `skipped`: test not run because of a failing hook

I haven't changed anything on the first item `pending`. This mismatch between our terminology and function is an old problem. We could replace label `pending` by label `skipped`, but this would break a few third-party tools, reporters, ...
I certainly don't want to rename/add a new `this.pending()`.
And the `pending` part is out of this PR's scope. 

The second label `skipped` is new and is about failing hooks. We could rename it to "omitted" or "dropped" or whatever. Eg. Cypress and Mochawesome are using `skipped` for tests omitted by failing hooks. That's why I have chosen `skipped`. But yes, it is confusing.

We need two labels: pending (or we rename it to skipped) and a new one for hook-failed-omitted tests. So I need to know which name to chose.

dasilvacontin: A bit of brainstorm:

My thoughts are, when I mark a test not to run, I want Mocha to ignore it, to disable it, to skip it, to omit it.

If it's skipped because of hook-failed, Mocha is skipping it, leaving out, or the test was unrun, unused, benched, called off, unprepared, dropped, withdrawn, canceled, halted. And some of this could apply to the previous case as well.

I think my favorites so far for hook-failed-omitted are `called off`, `canceled`, `halted`. I think called off even best.

It could be interesting to amass a large number of these potential labels, and run a survey through users of Mocha to see how well they feel each of those words describes (from 1 to 10) those two cases. Not to base our choice on our opinion alone.

mark-wiemer:
I think https://github.com/mochajs/mocha/pull/5519 supercedes this one and provides more clarity, should be merged in a v12 prerelease as well. Closing this one for now.

Issue #2314, opened 2016-06-16
Run Mocha tests from within existing Node.js code


lastmjs:
It would be nice to be able to run Mocha tests from within normal Node.js code. Here's what I'm talking about:

```
const Mocha = require('Mocha');
const mocha = new Mocha();
const assert = require('chai').assert;

//do normal Node.js things

mocha.run(() => {
     describe('an amazing suite of tests', () => {
         it('should be true', () => {
             assert.equal(true, true);
         });
     });
});

// do some more Node.js things
```

Or maybe a template string could work if we had to:

```
const Mocha = require('Mocha');
const mocha = new Mocha();
const assert = require('chai').assert;

//do normal Node.js things

mocha.run(`
     describe('an amazing suite of tests', () => {
         it('should be true', () => {
             assert.equal(true, true);
         });
     });
`);

// do some more Node.js things
```

I'm running into problems with the limitations of only being able to run Mocha tests by loading stand-alone files. I'm trying to get some interesting test environments working inside of Electron and with web components.

boneskull:
"node-able" test files are a goal for the future; see #1969 and work done in [mocha-core](/mochajs/mocha-core).  as such, closing this as a dupe

pps83:
That #1969 is quite useless if one wants to figure out how to run mocha tests as a regular nodejs app. Any clue?
There was something suggested there: `const {describe, it, timeout} = require('mocha');` but this doesn't work.

fbonds:
I'm experiencing this as well in vs code with node, selenium & mocha.  Looking for a solution. Thanks.

Gribbs:
+1 TypeError: describe is not a function even when using:

let mocha = require('mocha')
let describe = mocha.describe

fasatrix:
I'd love the feature explained  by @lastmjs too. I am using AWS SDK to query infrastructure resources that I'd like to assert afterwords 
 The SDK allows to run node js code as a node app. At the moment I am using a workaround by writing out to a file response from AWS API and then read that file with mocha and do the mocha things, but it would be extremely helpful to be able to run mocha from within a node app. Any updates on this?

audioscavenger:
this is how i got it working within electron:
main.js:
```
var Mocha = require('mocha')
var mochaOptions = {
  timeout: 10000,
  ui: 'tdd',
  bail: true,
  cleanReferencesAfterRun: true,     // https://github.com/mochajs/mocha/pull/4234
}
function testMochaRenderer() {
[...]

PR #4234, opened 2020-04-21
Add ability to run tests in a mocha instance multiple times

### LINKED ISSUE #2783: Mocha can't run tests twice programmatically
Programmatically, I cannot run a Mocha test twice.

I.e. I can't do `mocha.run()` twice.

I looked at a similar issue (#[995](https://github.com/mochajs/mocha/issues/995)), however, the solutions are not optimal and/or not work.

For example, deleting the require cache is not a solution for me, as I need to use the require cache.

Is there a way to `mocha.run()` twice?

Thanks in advance!

nicojs:
### Description of the Change

These changes allow mocha to run multiple times with the same test suite programmatically (without reload). It does this with minimal code changes. 

It also adds to validations, for when people call `mocha.run` at the wrong time.

For more details and reasoning, please see https://github.com/mochajs/mocha/issues/2783#issuecomment-616446938

Fixes #2783

### Alternate Designs

Alternatively, we could have totally removed all state from tests, suites and runnable, but we don't want to change the entire way of keeping state. Please see #2783 for the entire history. 

A lot of people seem interested in running a test suite multiple times without reloading files from disk. See #2783. I personally will use it to [implement mutation switching in Stryker](https://github.com/stryker-mutator/stryker/issues/1514)

### Benefits

Huge performance gain for my mutation testing use case. Other people have valid use cases as well.

### Possible Drawbacks

It should be backward compatible, but since I had to rewrite parts in the core, I might have introduced undesired side effects 🤷‍♂️. Although the integration test suite is awesome!

### Applicable issues

#2783 

### Todo:

- [x] Test: implement `reset`
- [x] Suite: implement `reset`
- [x] Runnable: implement `reset`
- [x] Mocha: implement `dispose` and `cleanReferencesAfterRun`
- [x] Mocha: throw errors when `run` call is invalid.
- [x] Runner: make sure to not leak event listeners.
- [x] Runner: implement `dispose` and `cleanReferencesAfterRun`
- [x] Unit tests
- [x] Integration tests
- [x] Docs


nicojs:
>  Coverage increased (+0.2%) to 93.213% 

You're welcome 💁‍♂️

nicojs:
@boneskull the documentation website build failed because of an http error, not my doing. Please rerun that job if you can, thanks!

afonsojramos:
I will be paying close attention to this PR, for now, it looks good!

boneskull:
Huh, I am not sure how to re-deploy this on netlify.  I don't see the option.

browser-entry.js
```diff
@@ -52,6 +52,17 @@ process.removeListener = function(e, fn) {
   }
 };
 
+/**
+ * Implements listenerCount for 'uncaughtException'.
+ */
+
+process.listenerCount = function(name) {
```
boneskull, line 59: NOTE:  this is an observation and not a request for changes.

I don't love how `process` is shimmed here.  we might be able to get away with changing the prototype to the `EventEmitter` shim (loaded via browserify via [events](https://npm.im/events), because `process` should be an `EventEmitter`.

lib/errors.js
```diff
@@ -129,6 +129,35 @@ function createInvalidExceptionError(message, value) {
   return err;
 }
 
+/**
+ * Creates an error object to be thrown when a mocha object's `run` method is executed while it is already disposed.
+ * @param {string} message The error message to be displayed.
+ * @param {boolean} cleanReferencesAfterRun the value of `autoDispose`
```
boneskull, line 135: this references `autoDispose` but I don't see this name elsewhere in the file.  should it be `cleanReferencesAfterRun`?

lib/mocha.js
```diff
@@ -97,6 +101,7 @@ function Mocha(options) {
   this.options = options;
   // root suite
   this.suite = new exports.Suite('', new exports.Context(), true);
+  this._cleanReferencesAfterRun = true;
```
boneskull, line 104: do you think it'd be better to use a finite state machine here?  when we start dealing with more than a couple boolean flags that get mutated, it starts to get hairy...

lib/mocha.js
```diff
@@ -488,6 +494,39 @@ Mocha.prototype.checkLeaks = function(checkLeaks) {
   return this;
 };
 
+/**
+ * Enables or disables whether or not to dispose after each test run.
+ * Disable this to ensure you can run the test suite multiple times.
+ * If disabled, be sure to dispose mocha when you're done to prevent memory leaks.
+ * @public
+ * @see {@link Mocha#dispose}
+ * @default true
```
boneskull, line 503: remove `@default true`

L504 should be something like: 

```
* @param {boolean} [cleanReferencesAfterRun=true] - If `false`, disables dispose
```

lib/mocha.js
```diff
@@ -488,6 +494,39 @@ Mocha.prototype.checkLeaks = function(checkLeaks) {
...
+Mocha.prototype.cleanReferencesAfterRun = function(cleanReferencesAfterRun) {
+  this._cleanReferencesAfterRun = cleanReferencesAfterRun !== false;
+  return this;
+};
+
+/**
+ * Manually dispose this mocha instance. Mark this instance as `disposed` and unable to run more tests.
+ * It also removes function references to tests functions and hooks, so variables trapped in closures can be cleaned by the garbage collector.
+ * @public
+ * @returns {void}
```
boneskull, line 517: you can remove this line, it's implied

lib/runner.js
```diff
@@ -123,16 +123,22 @@ module.exports = Runner;
  * @class
  * @param {Suite} suite Root suite
  * @param {boolean} [delay] Whether to delay execution of root suite until ready.
+ * @param {boolean} [cleanReferencesAfterRun] Whether to clean references to test fns and hooks when a suite is done.
  */
-function Runner(suite, delay) {
+function Runner(suite, delay, cleanReferencesAfterRun) {
```
boneskull, line 128: we're going to be collecting options here (#4245 does it as well).  Let's change the function signature, but in a non-breaking way:

```js
/**
 * Initialize a `Runner` at the Root {@link Suite}, which represents a hierarchy of {@link Suite|Suites} and {@link Test|Tests}.
 *
 * @extends external:EventEmitter
 * @public
 * @class
 * @param {Suite} suite - Root suite
 * @param {Object|boolean} [opts] - Options. If `boolean`, whether or not to delay execution of root suite until ready (for backwards compatibility).
 * @param {boolean} [opts.delay] - Whether to delay execution of root suite until ready.
 * @param {boolean} [opts.cleanReferencesAfterRun] - Whether to clean references to test fns and hooks when a suite is done.
 */
function Runner(suite, opts) {
  var delay;
  if (typeof opts === 'boolean') {
    delay = opts;
    opts = {};
  }
  // etc
```

lib/runner.js
```diff
@@ -123,16 +123,22 @@ module.exports = Runner;
...
   this._abort = false;
   this._delay = delay;
   this.suite = suite;
   this.started = false;
+  this._cleanReferencesAfterRun = cleanReferencesAfterRun;
   this.total = suite.total();
   this.failures = 0;
+  /**
+   * @types {[[EventEmitter, string, function]]}
+   */
```
boneskull, line 140: `@types` isn't understood by JSDoc.  afaik there's no way to define a tuple either. the best you can get is `@type {(EventEmitter|string|function)[]}`, I think.

lib/runner.js
```diff
@@ -162,6 +168,53 @@ Runner.immediately = global.setImmediate || process.nextTick;
...
+};
+
+/**
+ * Replacement for `target.removeListener(eventName, listener)` that also updates the bookkeeping.
+ * @param target {EventEmitter}
+ * @param eventName {string}
+ * @param fn {function}
+ */
+Runner.prototype._removeEventListener = function(target, eventName, listener) {
+  var eventListenerIndex = this._eventListeners.findIndex(function(
```
boneskull, line 189: This looks to only be used for `process`, but it iterates over everything in `this._eventListeners`.  That doesn't look necessary.  Maybe instead we keep an array of _process_ event listeners (containing only `eventName` and `listener`, or an object with `eventName` key and an array of listener functions as value)--say, `this._processEventListeners`--and instead of naming this function `_removeEventListener`, call it `_removeProcessEventListener`.  In `dispose()`, iterate over this data structure in addition to `this._eventListeners`... assuming removing process listeners in `dispose()` is intended.

lib/suite.js
```diff
@@ -92,6 +92,19 @@ function Suite(title, parentContext, isRoot) {
...
+ * Resets the state initially or for a next run.
+ */
+Suite.prototype.reset = function() {
+  this.delayed = false;
+  this.suites.forEach(function(suite) {
+    suite.reset();
+  });
+  this.tests.forEach(function(test) {
+    test.reset();
+  });
```
boneskull, line 105: what of hooks?

test/unit/runnable.spec.js
```diff
@@ -179,6 +179,24 @@ describe('Runnable(title, fn)', function() {
...
+    beforeEach(function() {
+      run = new Runnable();
+    });
+
+    it('should reset current run state', function() {
+      run.timedOut = true;
+      this._currentRetry = 5;
+      this.pending = true;
+      run.reset();
+      expect(run.timedOut, 'to be', false);
```
boneskull, line 194: fwiw this could be `'to be false'`

test/integration/fixtures/multiple-runs/multiple-runs-different-output.fixture.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+const Mocha = require('../../../../lib/mocha');
+
+const mocha = new Mocha({ reporter: 'json' });
+if (process.argv.indexOf('--no-clean-references') >= 0) {
```
boneskull, line 5: Is there precedent for sending these custom arguments elsewhere?  It may have unintended consequences.  a more foolproof could be to modify `env` in the options object (third parameter to `invokeMocha`)

is the reason for these just that you can reuse the fixture in several different ways?  if that's the case, I'd probably prefer just copying the fixtures into separate files, each with their own behavior.  if you need to reuse some code, put that in a common module and `require` it from the fixtures

browser-entry.js
nicojs: Yeah I was surprised that that wasn't yet the case.

lib/mocha.js
nicojs: Yeah, although I think `_cleanReferencesAfterRun` should still be a separate  boolean, since it is independent. The other state bits can be converted I think. Let me see what I can come up with.

lib/runner.js
nicojs: It is also used for `'error'` listeners on tests and hooks. (L440 and L592)

lib/suite.js
nicojs: I'm not 100% sure I know what you mean with hooks, but I think it is `beforeEach` and friends.

I super duper don't want them to be cleaned. They should stay, just like the test fn's themselves. We're going to need them for the next run.

test/integration/fixtures/multiple-runs/multiple-runs-different-output.fixture.js
nicojs: > is the reason for these just that you can reuse the fixture in several different ways

Yes, they are. Will make 3 fixtures for them 👩‍👧‍👧

lib/runner.js
nicojs: Good point. Will fix.

lib/mocha.js
nicojs: I've implemented a simple state machine. Will push as soon as local build is green 💚

```js
/**
 * A Mocha instance is a finite state machine.
 * These are the states it can be in.
 */
var mochaStates = utils.defineConstants({
  /**
   * Initial state of the mocha instance
   */
  INIT: 'init',
  /**
   * Mocha instance is running tests
   */
  RUNNING: 'running',
  /**
   * Mocha instance is done running tests and references to test functions and hooks are cleaned.
   * You can reset this state by unloading the test files.
   */
  REFERENCES_CLEANED: 'referencesCleaned',
  /**
   * Mocha instance is disposed and can no longer be used.
   */
  DISPOSED: 'disposed'
});
```

nicojs:
I've implemented (most of) the requested changes. 

I've also added one last bit. When disposed of, a Mocha instancet will now unload files internally. That makes a lot of sense to me. If someone wants to create a new mocha instance in the same process and add the same files again, he should be able to.

lib/runner.js
boneskull: ah, didn't see that.  I am concerned that this is a CPU-bound bottleneck in very large test files / suites.  if it turns out to be, we can refactor it later to be more difficult to understand but more performant. :smile:

lib/suite.js
boneskull: yes, hooks are `beforeEach` etc.  I think I am confused about why suites and tests are getting `reset()` called, but not hooks.

boneskull:
ty for the update.  looks like there's a conflict.

nicojs:
Conflict gone 🤸‍♂️

lib/suite.js
boneskull: to be clear:  I'm still confused.

boneskull:
@nicojs there's another conflict 😜

nicojs:
You're a busy guy. Fixed once more 🤸‍♂️

lib/suite.js
nicojs: I think I now understand your criticism. Hooks have state as well. And they inherit from `Runnable`. Didn't realize. Will add it when I have some time. Thanks for pointing it out!

nicojs:
You're quite good at keeping me busy 🙊

I've implemented the reset on hooks. I saw that `state` and `err` could also be set on a test/hook during a run, so I've added those fields to `Runnable.reset()`. I've also solved the merge conflict.

As far as I know, I've fixed all issues/recommendations.

lib/errors.js
```diff
@@ -158,6 +158,25 @@ function createMochaInstanceAlreadyRunningError(message, instance) {
...
+function createInvalidPluginError(message, pluginType, pluginId) {
+  switch (pluginType) {
+    case 'reporter':
+      return createInvalidReporterError(message, pluginId);
+    case 'interface':
+      return createInvalidInterfaceError(message, pluginId);
+    default:
+      throw new Error('unknown pluginType "' + pluginType + '"');
+  }
+
```
afonsojramos, line 179: You seem to have missed a closing bracket.

nicojs: Yes, did a merge in the browser 🤷‍♂️

Fixed now. I'm hoping you guys do squash commit? If not, I could do it as well.

afonsojramos:
@boneskull can you take a look if it is good to go? 🚀

lib/errors.js
boneskull: yes, will squash.

boneskull:
Seems to be yet another conflict.

boneskull:
I would like to call this a "bugfix", since it really always should have worked properly.  But I could see how others may consider this a feature.  Thoughts, @mochajs/core?

boneskull:
Something's gone wonky with the changesets here.

I tried to rebase back onto `master`, but that didn't work, and I had more conflicts.  _then_ I did a merge via the GitHub UI to resolve those conflicts, and now I just have...more conflicts.

@nicojs Sorry, I tried to help.  You'll need to pull your branch again from your fork, and pull `master` as well, then retry a rebase back onto `master`.

boneskull:
It looks as if there are changesets in here that shouldn't be in here; 
66d9b29 and later.  A rebase will need to ensure those changesets don't end up in this PR.

@nicojs I'll give you a chance, but please let me know if you want me to give it another go.  I may have screwed it up myself!

Might even be best to _not_ pull your branch back down, and instead just retry a rebase onto `master` from your working copy's branch.

nicojs:
> Might even be best to not pull your branch back down, and instead just retry a rebase onto master from your working copy's branch.

This is basically what I did.

```shell
git checkout feat/hot-reload
git fetch origin master:master
git rebase master
# Then I fixed the rebase issues that came up and continued rebasing (18 commits). 
git push --force
```

boneskull:
thanks, looks like you fixed it.

boneskull:
IMO this can land, but want to know if @juergba has any concerns.

afonsojramos:
Any updates on this PR @boneskull @juergba ?

lib/runner.js
```diff
@@ -121,18 +121,30 @@ module.exports = Runner;
...
+ * @param {Object|boolean} [opts] - Options. If `boolean`, whether or not to delay execution of root suite until ready (for backwards compatibility).
+ * @param {boolean} [opts.delay] - Whether to delay execution of root suite until ready.
+ * @param {boolean} [opts.cleanReferencesAfterRun] - Whether to clean references to test fns and hooks when a suite is done.
  */
-function Runner(suite, delay) {
+function Runner(suite, opts) {
+  if (opts === undefined) {
+    opts = {};
+  }
+  if (typeof opts === 'boolean') {
```
Munter, line 133: It's good that this change is backwards compatible. But if the updated pattern is something we'd like to go ahead with in the future, maybe we should consider a deprecation notice here and queue a removal in a future major release

boneskull: maybe, but we could also just keep it in, since it doesn't look too hairy (at least for now!).  IMO, if/when it becomes a maintenance burden I'd advocate for this approach.

boneskull:
@mochajs/core any thoughts on whether this should be a patch or minor?  If our next release will be v8.0.0, I guess it's moot.

nicojs:
It should be a minor version IMHO, since it adds a new feature, right?

Munter:
Agree on it being a minor release (or rolled into the next major)

boneskull:
Will give @juergba 48h to look at this; if he declines to review I'll merge it

boneskull:
This is one of those "missing features" that is only a bug because it was missing.  😄

nicojs:
True. 

To be fair, I think we same problem in Stryker itself. Creating an instance of `Stryker` and running mutation testing twice is probably not working 🙄 Just don't tell anyone 🤐

nicojs:
So... I guess this will be merged soon 🤗

boneskull:
yes

boneskull:
So it looks like I can't actually use this for the parallel runner.  What I had hoped to do is re-use the same Mocha instance... except with different files (specifically, a single file) on every run.  

@nicojs Wondering if you have an idea about what could be causing this behavior.  What I'm trying to do is:

1. Initialize a `Mocha` instance.  Set `cleanReferencesAfterRun` to `false`.
1. Repeat for _n_ files:
    1. Add a file to `Mocha` instance
    1. call `Mocha#run()`
    1. Upon completion, "reset" Mocha and remove all files from the `Mocha#files` array. 

I attempted to "reset" by calling `dispose()` upon completion (step 2.iii.).  But this fails at the next run, because:

```md
Mocha instance is already disposed, cannot start a new test run. Please create a new mocha instance. Be sure to set disable `cleanReferencesAfterRun` when you want to reuse the same mocha instance for multiple test runs
```

Of course, if I _don't_ call `dispose()`, things get really confused due to extra hooks hanging around.

Why can't I reuse the `Mocha` instance after `dispose()`?

boneskull:
Should I just not call `cleanReferencesAfterRun`?  Would it work then?

boneskull:
(I suppose I'm rubbber-ducking now...)

boneskull:
I think, maybe, that it wasn't the problem you were trying to solve.  but it seems to get most of the way there.

boneskull:
OK, so, I am thinking it's really not worth the bother to try to reuse `Mocha` instances in the manner I'm trying (in #4245), even with your code now merged.  Too many things to worry about.  Like dumping files out of the module cache--but we're still running root-level hooks from those files, root-level hooks being applied nondeterministically, etc.  I mean, it's possible, just difficult to do correctly (for some definition of correctly).

anyway, thanks for the PR 😄

nicojs:
Yeah, it wasn't the problem I was trying to solve. The reason I added the behavior of throwing an error on run after dispose was just because thought that made the most sense. Feel free to change that behavior, no harm done since it wasn't released yet, right? As long as I can run the exact same suite without reloading files, I'm happy ;)

birtles:
This appears to cause breakage for me:
https://github.com/birtles/rikaichamp/runs/707671695#step:5:71
I am getting `createMochaInstanceAlreadyRunningError` when trying to update mocha.

I haven't had a chance to debug why yet. Any tips would be appreciated.

craigtaub:
@birtles Please could you submit this bug as a new issue? Please include details of versions as well as how you are calling mocha. Thanks

birtles:
> @birtles Please could you submit this bug as a new issue? Please include details of versions as well as how you are calling mocha. Thanks

Hi. I went to submit a bug but the bug form was requiring a little bit more time than I can currently afford, unfortunately. I discovered that dropping a call to `mocha.run()` resolved the issue for me. I have no idea why the call to `mocha.run()` was needed in the first place.

For what it's worth, I'm running with `karma` and `karma-firefox-launcher`.

audioscavenger:
i updated the thread https://github.com/mochajs/mocha/issues/2314 and I found out that there must be a bug in the options parser.
by forcing _cleanReferencesAfterRun = false after you instantiated a new mocha, it actually let you re-run your tests! lost 2 hours on this one. You're welcome.
```
var mocha = new Mocha(mochaOptions)
mocha._cleanReferencesAfterRun = false
```

PR #4235, opened 2020-04-21
enable emoji on website; enable normal ul elements

boneskull:
grouped these together:

- Site can now use emoji :party:
- Using the `.single-column` class will render a `ul` element as a "normal" list
- Suppress the warning about `text` being an unknown language coming out of 11ty/Prism/markdown-it

Ref: #4198


boneskull:
updated with lockfile changes to grab the missing dep

boneskull:
this is pretty low impact and can be yanked out if needed, so I'm merging this

PR #4236, opened 2020-04-21
better debug output

boneskull:
- make sure that template strings aren't used in debug statements, since they are eager

Ref: #4198

### Requirements

### DIFF
--- a/bin/mocha
+++ b/bin/mocha
@@ -34,7 +34,7 @@ debug('loaded opts', opts);
  */
 const disableTimeouts = value => {
   if (impliesNoTimeouts(value)) {
-    debug(`option "${value}" disabled timeouts`);
+    debug('option %s disabled timeouts', value);
     mochaArgs.timeout = 0;
     delete mochaArgs.timeouts;
     delete mochaArgs.t;
@@ -108,7 +108,11 @@ if (Object.keys(nodeArgs).length) {
     unparse(mochaArgs, {alias: aliases})
   );
 
-  debug(`exec ${process.execPath} w/ args:`, args);
+  debug(
+    'forking child process via command: %s %s',
+    process.execPath,
+    args.join(' ')
+  );
 
   const proc = spawn(process.execPath, args, {
     stdio: 'inherit'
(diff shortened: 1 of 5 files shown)


juergba:
> make sure that template strings aren't used in debug statements, since they are eager

I don't understand this statement.

lib/cli/config.js
```diff
@@ -69,7 +69,7 @@ const parsers = (exports.parsers = {
  */
 exports.loadConfig = filepath => {
   let config = {};
-  debug(`loadConfig: "${filepath}"`);
+  debug('loadConfig: "%s"', filepath);
```
arvidOtt, line 72: I would suggest to change `line 98` as well to `debug('findConfig: found "%s"', filepath);`

boneskull:
> > make sure that template strings aren't used in debug statements, since they are eager
> 
> I don't understand this statement.

@juergba I mean that a template literal is always evaluated regardless.  If debug is not enabled, the string containing placeholders _won't_ be evaluated.  The impact is probably minimal in our case, but IMO it's good practice.

lib/cli/config.js
boneskull: done

bin/mocha
```diff
@@ -108,7 +108,11 @@ if (Object.keys(nodeArgs).length) {
     unparse(mochaArgs, {alias: aliases})
   );
```
juergba, line 110: more line 37

lib/cli/config.js
```diff
@@ -69,7 +69,7 @@ const parsers = (exports.parsers = {
  */
 exports.loadConfig = filepath => {
   let config = {};
-  debug(`loadConfig: "${filepath}"`);
+  debug('loadConfig: found "%s"', filepath);
 
   const ext = path.extname(filepath);
   try {
```
juergba, line 75: more line 98

lib/cli/options.js
```diff
@@ -178,16 +178,16 @@ const loadPkgRc = (args = {}) => {
     try {
```
juergba, line 178: more line 85, 88

boneskull:
this is low-impact, so merging

PR #4237, opened 2020-04-21
refactor validatePlugins to throw coded errors

boneskull:
- add `createInvalidPluginError` for reporters, UIs, and future plugins
- ensures the original error is output if the module exists, but it throws (see `test/node-unit/cli/fixtures/bad-module.fixture.js`)
- remove unneeded `process.cwd()` from call to `path.resolve()`

Ref: #4198

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -12,8 +12,10 @@ const path = require('path');
 const debug = require('debug')('mocha:cli:run:helpers');
 const watchRun = require('./watch-run');
 const collectFiles = require('./collect-files');
+const {format} = require('util');
 
 const cwd = (exports.cwd = process.cwd());
+const {createInvalidPluginError} = require('../errors');
 
 /**
  * Exits Mocha when tests + code under test has finished execution (default)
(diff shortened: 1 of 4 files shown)


lib/cli/run-helpers.js
```diff
@@ -146,22 +147,31 @@ exports.runMocha = async (mocha, options) => {
...
+    throw createInvalidPluginError(
+      `"--${key} <${key}>" can only be specified once`,
+      key
+    );
   }
 
-  const unknownError = () => new Error(`Unknown "${key}": ${opts[key]}`);
+  const unknownError = err =>
+    createInvalidPluginError(
+      `Could not load ${key} "${opts[key]}": ${err}`,
```
raymondfeng, line 170: The stack trace of `err` will be gone here. Are we good with it?

boneskull: no, good catch.

boneskull:
@raymondfeng updated

boneskull:
I think this is pretty non-controversial, but would still like another maintainer to look.

lib/cli/run-helpers.js
```diff
@@ -146,35 +148,44 @@ exports.runMocha = async (mocha, options) => {
...
+ * of sorts; `Mocha.reporters` is one, where each key corresponds to a reporter
+ * name
  * @private
  */
 exports.validatePlugin = (opts, key, map = {}) => {
   if (Array.isArray(opts[key])) {
-    throw new TypeError(`"--${key} <${key}>" can only be specified once`);
+    throw createInvalidPluginError(
+      `"--${key} <${key}>" can only be specified once`,
+      key
```
craigtaub, line 165: Both calls aren't handing `pluginName` from what I can see? (`, opts[key]`)

lib/cli/run-helpers.js
```diff
@@ -146,35 +148,44 @@ exports.runMocha = async (mocha, options) => {
...
 /**
- * Used for `--reporter` and `--ui`.  Ensures there's only one, and asserts
- * that it actually exists.
- * @todo XXX This must get run after requires are processed, as it'll prevent
- * interfaces from loading.
+ * Used for `--reporter` and `--ui`.  Ensures there's only one, and asserts that
+ * it actually exists. This must be run _after_ requires are processed (see
+ * {@link handleRequires}), as it'll prevent interfaces from loading otherwise.
  * @param {Object} opts - Options object
  * @param {string} key - Resolvable module name or path
```
juergba, line 155: Is this parameter description correct?

test/node-unit/cli/fixtures/bad-module.fixture.js
```diff
@@ -0,0 +1 @@
+throw new Error('this module is wonky');
```
juergba, line 1: CR missing

test/node-unit/cli/run-helpers.spec.js
```diff
@@ -14,36 +14,94 @@ describe('cli "run" command', function() {
     sandbox.restore();
```
juergba, line 14: `sandbox` isn't used in any test.

boneskull:
> Can we exclude the HTML reporter for CLI usage?
> I'm not sure wether the plugin validation is the correct place for this.

The `html` reporter can conceivably work if a user requires [jsdom](https://npm.im/jsdom) and initializes it.  I am not sure of the use case, but just because I don't know what it is doesn't mean it isn't valid.

Right now, w/o `jsdom`, `--reporter html` will fail with `ReferenceError: document is not defined`.  We could maybe do a check before usage of `global.document` to throw more friendly error.  But I don't think that should be in this PR.

lib/cli/run-helpers.js
boneskull: good catch.  it doesn't make sense for the first call, but it does make sense for the second.

lib/cli/run-helpers.js
boneskull: not quite; I've updated it and the variable names to be more descriptive.

boneskull:
Updated to resolve peer review issues

test/node-unit/cli/fixtures/bad-module.fixture.js
boneskull, line 1: this has a final newline locally, but not sure if GitHub shows that.

juergba: GH only shows when the final newline is missing.

lib/cli/run-helpers.js
```diff
@@ -146,35 +148,52 @@ exports.runMocha = async (mocha, options) => {
...
+ * name
  * @private
  */
-exports.validatePlugin = (opts, key, map = {}) => {
-  if (Array.isArray(opts[key])) {
-    throw new TypeError(`"--${key} <${key}>" can only be specified once`);
+exports.validatePlugin = (opts, pluginType, map = {}) => {
+  /**
+   * This should be a unique identifier; either a string (present in `map`),
+   * or a resolvable (via `require.resolve`) module ID/path.
```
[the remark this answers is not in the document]
boneskull: true, might as well make this async too.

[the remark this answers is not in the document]
boneskull: I think that's another PR though.

PR #4238, opened 2020-04-21
fix nyc ignorelist

boneskull:
should keep more garbage out of the coverage


boneskull:
> Does it make sense to exclude the `"scripts/**"` as well?

yeah, good catch

boneskull:
this is low-impact as well; merging

PR #4239, opened 2020-04-21
fixes some signal handling

boneskull:
- `landing` reporter befouls the terminal up if `SIGINT` is issued
- `--exit` test used an incorrect description, and also did not reliably handle `SIGINT` when running

Ref: #4198

### DIFF
--- a/lib/reporters/landing.js
+++ b/lib/reporters/landing.js
@@ -98,6 +98,14 @@ function Landing(runner, options) {
     process.stdout.write('\n');
     self.epilogue();
   });
+
+  // if cursor is hidden when we ctrl-C, then it will remain hidden unless...
+  process.once('SIGINT', function() {
+    cursor.show();
+    process.nextTick(function() {
+      process.kill(process.pid, 'SIGINT');
+    });
+  });
 }
 
 /**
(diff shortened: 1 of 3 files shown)


boneskull:
this needs changes I made to `exit.spec.js` from #4240, so I'll move the changes from there into this

boneskull:
I'm going to close this one and add it to #4245.

boneskull:
ah, nm, I see the issue.

boneskull:
I don't have any idea how we could test the newly-uncovered lines (in the `landing` reporter).  we'll have to deal with the 0.09% decrease in coverage.

boneskull:
cc @arvidott and/or @raymondfeng 

this is short and low-impact, but would appreciate anybody taking a look.

boneskull:
> So it would register "SIGINT" but the cursor would remain hidden? Not good. Good fix 👍

Not quite.  What was happening was if someone was using the `landing` reporter and issued a `SIGINT` (via ctrl-C or whatever) while the cursor was hidden (which it is for much of the reporter), it would not restore the cursor to the user's terminal.  annoying

PR #4240, opened 2020-04-21
assorted test fixes & refactors

boneskull:
- increase default timeout for wiggle room
- specify `watch-ignore` in case we run our own tests in watch mode
- reformat `.mocharc.yml`
- `integration/fixtures/uncaught/listeners.fixture.js`: reduce number of listeners created to avoid max listener warning
- `integration/fixtures/options/watch.spec.js`: do not use context object; be more specific about spawn options
- `integration/fixtures/diffs.spec.js`: do not pass `-C`; the helper already does this
- `node-unit/mocha.spec.js`: do not actually create files; it's a unit test after all.
- `node-unit/cli/config.spec.js`: rewiremock fixes
- `node-unit/cli/options.spec.js`: rewiremock fixes; increase timeout
- `unit/hook-timeout.spec.js`: do not override default (configured) timeout
- `unit/runner.spec.js`: leverage [unexpected-eventemitter](https://npm.im/unexpected-eventemitter)
- `unit/throw.spec.js`: proper teardown: remove uncaught exception listeners
- `unit/timeout.spec.js`: increase timeout to _greater than_ default (configured) value

Ref: #4198

### DIFF
--- a/.mocharc.yml
+++ b/.mocharc.yml
@@ -1,7 +1,14 @@
-require: test/setup
-ui: bdd
+require: 'test/setup'
+ui: 'bdd'
 global:
-  - okGlobalA,okGlobalB
-  - okGlobalC
-  - callback*
-timeout: 300
+  - 'okGlobalA,okGlobalB'
+  - 'okGlobalC'
+  - 'callback*'
+timeout: 1000
+watch-ignore:
+  - '.*'
+  - 'docs/_dist/**'
+  - 'docs/_site/**'
+  - 'node_modules'
+  - 'coverage'
+  - 'cache'
--- a/test/integration/fixtures/uncaught/listeners.fixture.js
+++ b/test/integration/fixtures/uncaught/listeners.fixture.js
@@ -3,7 +3,8 @@
 const assert = require('assert');
 const mocha = require("../../../../lib/mocha");
 
-for (let i = 0; i < 15; i++) {
+// keep this low to avoid warning
+for (let i = 0; i < 5; i++) {
   const r = new mocha.Runner(new mocha.Suite("" + i, undefined));
   r.run();
 }
--- a/test/integration/options/watch.spec.js
+++ b/test/integration/options/watch.spec.js
@@ -7,24 +7,24 @@ const helpers = require('../helpers');
 
 describe('--watch', function() {
   describe('when enabled', function() {
-    this.timeout(10 * 1000);
-    this.slow(3000);
+    let tempDir;
+    this.slow(5000);
 
     beforeEach(function() {
-      this.tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mocha-'));
+      tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mocha-'));
     });
 
     afterEach(function() {
-      if (this.tempDir) {
-        return fs.remove(this.tempDir);
+      if (tempDir) {
+        return fs.remove(tempDir);
       }
     });
 
     it('reruns test when watched test file is touched', function() {
-      const testFile = path.join(this.tempDir, 'test.js');
+      const testFile = path.join(tempDir, 'test.js');
       copyFixture('__default__', testFile);
 
-      return runMochaWatch([testFile], this.tempDir, () => {
+      return runMochaWatch([testFile], tempDir, () => {
         touchFile(testFile);
       }).then(results => {
         expect(results, 'to have length', 2);
@@ -32,15 +32,15 @@ describe('--watch', function() {
     });
 
     it('reruns test when file matching --watch-files changes', function() {
-      const testFile = path.join(this.tempDir, 'test.js');
+      const testFile = path.join(tempDir, 'test.js');
       copyFixture('__default__', testFile);
 
-      const watchedFile = path.join(this.tempDir, 'dir/file.xyz');
+      const watchedFile = path.join(tempDir, 'dir/file.xyz');
       touchFile(watchedFile);
 
       return runMochaWatch(
         [testFile, '--watch-files', 'dir/*.xyz'],
-        this.tempDir,
+        tempDir,
         () => {
           touchFile(watchedFile);
         }
@@ -50,13 +50,13 @@ describe('--watch', function() {
     });
 
     it('reruns test when file matching --watch-files is added', function() {
-      const testFile = path.join(this.tempDir, 'test.js');
+      const testFile = path.join(tempDir, 'test.js');
       copyFixture('__default__', testFile);
 
-      const watchedFile = path.join(this.tempDir, 'lib/file.xyz');
+      const watchedFile = path.join(tempDir, 'lib/file.xyz');
       return runMochaWatch(
         [testFile, '--watch-files', '**/*.xyz'],
-        this.tempDir,
+        tempDir,
         () => {
           touchFile(watchedFile);
         }
@@ -66,15 +66,15 @@ describe('--watch', function() {
     });
 
     it('reruns test when file matching --watch-files is removed', function() {
-      const testFile = path.join(this.tempDir, 'test.js');
+      const testFile = path.join(tempDir, 'test.js');
       copyFixture('__default__', testFile);
 
-      const watchedFile = path.join(this.tempDir, 'lib/file.xyz');
+      const watchedFile = path.join(tempDir, 'lib/file.xyz');
       touchFile(watchedFile);
 
       return runMochaWatch(
         [testFile, '--watch-files', 'lib/**/*.xyz'],
-        this.tempDir,
+        tempDir,
         () => {
           fs.removeSync(watchedFile);
         }
@@ -84,15 +84,15 @@ describe('--watch', function() {
     });
 
     it('does not rerun test when file not matching --watch-files is changed', function() {
-      const testFile = path.join(this.tempDir, 'test.js');
+      const testFile = path.join(tempDir, 'test.js');
       copyFixture('__default__', testFile);
 
-      const watchedFile = path.join(this.tempDir, 'dir/file.js');
+      const watchedFile = path.join(tempDir, 'dir/file.js');
       touchFile(watchedFile);
 
       return runMochaWatch(
         [testFile, '--watch-files', 'dir/*.xyz'],
-        this.tempDir,
+        tempDir,
         () => {
           touchFile(watchedFile);
         }
@@ -102,14 +102,14 @@ describe('--watch', function() {
     });
 
     it('picks up new test files when they are added', function() {
-      const testFile = path.join(this.tempDir, 'test/a.js');
+      const testFile = path.join(tempDir, 'test/a.js');
       copyFixture('__default__', testFile);
 
       return runMochaWatch(
         ['test/**/*.js', '--watch-files', 'test/**/*.js'],
-        this.tempDir,
+        tempDir,
         () => {
-          const addedTestFile = path.join(this.tempDir, 'test/b.js');
+          const addedTestFile = path.join(tempDir, 'test/b.js');
           copyFixture('passing', addedTestFile);
         }
       ).then(results => {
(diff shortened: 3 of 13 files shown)


.mocharc.yml
```diff
@@ -1,7 +1,14 @@
-require: test/setup
-ui: bdd
+require: 'test/setup'
```
boneskull, line 1: the quotes are because YAML tries to be too smart. I like YAML, but for simple things only, and it must be used with caution.  quoting strings goes a long way.

test/integration/fixtures/uncaught/listeners.fixture.js
```diff
@@ -3,7 +3,8 @@
 const assert = require('assert');
 const mocha = require("../../../../lib/mocha");
 
-for (let i = 0; i < 15; i++) {
+// keep this low to avoid warning
+for (let i = 0; i < 5; i++) {
```
boneskull, line 7: @juergba 

the max listener limit is 10, so node was dumping 'max listeners exceeded' warnings.  this went unnoticed because--by default--the integration tests would pipe `STDERR` into `/dev/null`.  with the changes in this PR, `STDERR` is output to terminal, _unless_ the test is configured (via `{stdio: 'pipe'}`) to trap the output so it can be used in an assertion.

test/integration/options/watch.spec.js
```diff
@@ -7,40 +7,40 @@ const helpers = require('../helpers');
 
 describe('--watch', function() {
   describe('when enabled', function() {
-    this.timeout(10 * 1000);
-    this.slow(3000);
+    let tempDir;
```
boneskull, line 10: this is just me being fussy.  but most all of our tests use function scope instead of the context to store data.

test/node-unit/cli/config.spec.js
```diff
@@ -17,6 +16,17 @@ describe('cli/config', function() {
   });
 
   describe('loadConfig()', function() {
+    let parsers;
```
boneskull, line 19: this "unit" test was reading files, and it shouldn't need to read files to assert what's trying to assert.

test/unit/throw.spec.js
```diff
@@ -25,6 +25,7 @@ describe('a test that throws', function() {
   });
 
   afterEach(function() {
+    process.removeAllListeners('uncaughtException');
```
boneskull, line 28: this starts to bite when running in watch mode

boneskull:
timeout-related changes were variously because of test flake, and because I implemented this while implementing parallel mode, the test execution times have a larger variance.

I'm starting to wonder if test timeouts aren't really that helpful, because I spend most of my time increasing the timeouts due to slow CI runs instead of catching significant performance issues.

test/integration/fixtures/uncaught/listeners.fixture.js
boneskull: sorry, the changes in #4241 surfaced this.

test/node-unit/cli/config.spec.js
```diff
@@ -17,6 +16,17 @@ describe('cli/config', function() {
...
+    let loadConfig;
+
+    beforeEach(function() {
+      const config = rewiremock.proxy(
```
juergba, line 23: This is confusing, this `config` shadowing the one of line 8?

test/node-unit/mocha.spec.js
```diff
@@ -1,38 +1,37 @@
...
 
     it('should be chainable', function() {
       const mocha = new Mocha(opts);
-      expect(mocha.addFile(__filename), 'to be', mocha);
+      expect(mocha.addFile('some-file.js'), 'to be', mocha);
     });
   });
 
   describe('#loadFiles', function() {
     it('should load all files from the files array', function() {
```
juergba, line 26: Now you load only one file here.

test/unit/throw.spec.js
juergba: I don't understand this comment.

boneskull: `Runner` will continually add `uncaughtException` listeners to `process` when running tests tests under mocha's "watch" mode.  we need to remove the listeners that we've created by calling `runner.run()`

.mocharc.yml
```diff
@@ -1,7 +1,14 @@
...
+  - 'okGlobalC'
+  - 'callback*'
+timeout: 1000
+watch-ignore:
+  - '.*'
+  - 'docs/_dist/**'
+  - 'docs/_site/**'
+  - 'node_modules'
+  - 'coverage'
+  - 'cache'
```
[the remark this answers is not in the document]
boneskull: `.*` should do that

test/integration/options/watch.spec.js
```diff
@@ -7,40 +7,40 @@ const helpers = require('../helpers');
 
 describe('--watch', function() {
   describe('when enabled', function() {
-    this.timeout(10 * 1000);
```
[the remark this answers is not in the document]
boneskull: the timeout is `10000` for integration tests per `package-scripts.js`

test/node-unit/cli/config.spec.js
boneskull: addressed this

test/node-unit/mocha.spec.js
boneskull: updated the test to add a second file

.mocharc.yml
boneskull: updated

test/node-unit/cli/config.spec.js
```diff
@@ -52,7 +62,7 @@ describe('cli/config', function() {
         it('should use the JS parser', function() {
```
juergba, line 62: A test for the `.cjs` extension is missing.

test/node-unit/mocha.spec.js
```diff
@@ -1,38 +1,39 @@
...
+      this.timeout(1000);
       const mocha = new Mocha(opts);
 
-      testFiles.forEach(mocha.addFile, mocha);
+      mocha.addFile(require.resolve('../../package.json'));
```
juergba, line 30: Hasn't _package.json_ been loaded by the main thread already?

test/unit/throw.spec.js
juergba: In `start()` the runner clears the `uncaughtException` handlers. So I don't see how these listeners could be accumulated. Even in watch mode (and even when aborting the execution before re-run), EVENT_RUN_END should be emitted and those handlers removed.
Our uncaught exception handling is an important part of our async testing. To remove this handler for testing seems strange to me. These tests are weird, but anyway ...

boneskull: > In run() the runner clears the uncaughtException handlers. So I don't see how these listeners could be accumulated. Even in watch mode (and even when aborting the execution before re-run), EVENT_RUN_END should be emitted and those handlers removed.

After `EVENT_RUN_END` is emitted, the `Runner#uncaught` listener is removed from the `Process.uncaughtException` event, _but_ the `Runner#uncaughtEnd` is then _added_ as a listener.  There is nothing that removes this listener _except_ another call to `Runner#run`.  

You'll see in these hooks that the listeners that the _Mocha process_ creates are restored after each test.  We have to remove them up-front because they conflict.  Any event listeners the _test subject_ adds (`Runner`) removed by the call to `removeAllListeners('uncaughtException')`, which--in the happy path--is just the `Runner#uncaughtEnd` listener.  I felt it was probably a better idea to remove any listeners the subject might add here, in case something weird happens.

Another way to approach this might be just stubbing out `process` at the module level via rewiremock.  I have no idea if that would work on a global, or if we'd need to `var process = require('process')`, or what.  Using sinon only to do this--stubbing `Process#on` and `Process#removeListener` would be even worse, because it's not sandboxed, and would break the mocha process which was running the test.

This brings up a good point though: we should not rely on `process#*` as globals.  It should be treated like `setTimeout` and its ilk.

I think it's best to probably revert the changes here and try to address these concerns in a separate PR, because it's more involved than I wanted to get with this one.

test/node-unit/mocha.spec.js
boneskull: yeah, good point.  I've rewritten some of these tests in 798155b3. it's not a great idea to rely on test fixtures like this anyway, so I created a couple.  these still aren't _unit_ tests--because of the reliance on `require()` and I/O--but they are closer now.

test/node-unit/cli/config.spec.js
boneskull: ah, didn't notice!  just added one.

test/unit/throw.spec.js
boneskull: nevermind about reverting--it's trivial to show the problem.  before my change, you see this (wherein in called `touch test/unit/throw.spec.js` to trigger the run):

```
$ bin/mocha --watch test/unit/throw.spec.js



  a test that throws
    non-extensible
      ✓ should not pass if throwing sync and test is sync
      ✓ should not pass if throwing sync and test is async
      ✓ should not pass if throwing async and test is async
    undefined
      ✓ should not pass if throwing sync and test is sync
      ✓ should not pass if throwing sync and test is async
      ✓ should not pass if throwing async and test is async
    null
      ✓ should not pass if throwing sync and test is sync
      ✓ should not pass if throwing sync and test is async
      ✓ should not pass if throwing async and test is async


  9 passing (27ms)



  a test that throws
    non-extensible
      ✓ should not pass if throwing sync and test is sync
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
      ✓ should not pass if throwing sync and test is async
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
      ✓ should not pass if throwing async and test is async
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    undefined
      ✓ should not pass if throwing sync and test is sync
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
      ✓ should not pass if throwing sync and test is async
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
      ✓ should not pass if throwing async and test is async
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    null
      ✓ should not pass if throwing sync and test is sync
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
      ✓ should not pass if throwing sync and test is async
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
      ✓ should not pass if throwing async and test is async
(node:96463) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit


  9 passing (22ms)
```

boneskull: _each_ test of `Runner#run` adds a listener (`Runner#uncaughEnd`).  when we run this the second time, we quickly go over the max listeners limit of 10.

boneskull: this is not noticeable in watch mode otherwise, because the `Runner` instance is reused, and it cleans up after itself upon a second call.

boneskull:
@juergba I think this is ready for your approval again

test/node-unit/cli/config.spec.js
juergba: still missing

boneskull:
forgot to push, sorry.  if this passes CI, I'll merge

PR #4241, opened 2020-04-21
test helper improvements

boneskull:
### Assertions

- enables `RawResult` (the result of `invokeMocha()` or `invokeMochaAsync()`) to check the exit code via `to have code` assertion
- add passed/failing/pending assertions for `RawRunResult` (the result of `runMocha()` or `runMochaAsync()`)

### Integration test helpers

- expose `getSummary()`, which can be used with a `RawResult` (when not failing)
- reorganize the module a bit
- create `runMochaAsync()` and `runMochaJSONAsync()` which are like `runMocha()` and `runMochaJSON()` except return `Promise`s
- better trapping of JSON parse errors
- better default handling of `STDERR` output in subprocesses (print it instead of suppress it!)
- do not let the `DEBUG` env variable reach subprocesses _unless it was explicitly supplied_
- add an easily copy-paste-able `command` prop to summary
- add some missing docstrings

Ref: #4198

### DIFF
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -67,120 +67,120 @@ module.exports = {
   runMochaJSONAsync: runMochaJSONAsync
 };
 
-  /**
-   * Invokes the mocha binary for the given fixture with color output disabled.
-   * Accepts an array of additional command line args to pass. The callback is
-   * invoked with a summary of the run, in addition to its output. The summary
-   * includes the number of passing, pending, and failing tests, as well as the
-   * exit code. Useful for testing different reporters.
-   *
-   * By default, `STDERR` is ignored. Pass `{stdio: 'pipe'}` as `opts` if you
-   * want it.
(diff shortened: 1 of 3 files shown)


test/integration/helpers.js
```diff
@@ -170,9 +226,13 @@ module.exports = {
...
   var code = result.code;
-  result = JSON.parse(result.output);
-  result.code = code;
-  return result;
+  try {
+    result = JSON.parse(result.output);
+    result.code = code;
+    return result;
+  } catch (err) {
+    throw new Error(err.message);
```
boneskull, line 234: this is weird, I'm going to revert this.

boneskull: ok, not quite, but it's more useful now.

boneskull:
resolved conflict

boneskull:
@mocha/core any comments?

PR #4243, opened 2020-04-21
remove cruft from wallaby config

boneskull:
also increase default timeout

Ref: #4198


boneskull:
the test failure is timing flake and I've addressed it in #4240.

boneskull:
since I think I'm the only wallaby-user, I'm going to merge

PR #4244, opened 2020-04-21
add Root Hook Plugins

boneskull:
> (documentation will be in another PR; this depends on #4237)

Adds "root hook plugins", a system to define root hooks via files loaded with `--require`.

This enables root hooks to work in parallel mode.  Because parallel mode runs files in a non-deterministic order, and files do not share a `Mocha` instance, it is not possible to share these hooks with other test files.  This change also works well with third-party libraries for Mocha which need the behavior; these can now be trivially consumed by adding `--require` or `require: 'some-library'` in Mocha's config file.

The way it works is:

1. When a file is loaded via `--require`, we check to see if that file exports a property named `mochaHooks` (can be multiple files).
1. If it does, we save a reference to the property.
1. After Yargs' validation phase, we use async middleware to execute root hook plugin functions--or if they are objects, just collect them--and we flatten all hooks found into four buckets corresponding to the four hook types.
1. Once `Mocha` is instantiated, if it is given a `rootHooks` option, those hooks are applied to the root suite.

This works with parallel tests because we can save a reference to the flattened hooks in each worker process, and a new `Mocha` instance is created with them for each test file.

* * *

Tangential:

- Because a root hook plugin can be defined as an `async` function, I noticed that `utils.type()` does not return `function` for async functions; it returns `asyncfunction`.  I've added a (Node-specific, for now) test for this.
- `handleRequires` is now `async`, since it will need to be anyway to support ESM and calls to `import()`.
- fixed incorrect call to `fs.existsSync()`

Ref: #4198

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -12,10 +12,9 @@ const path = require('path');
 const debug = require('debug')('mocha:cli:run:helpers');
 const watchRun = require('./watch-run');
 const collectFiles = require('./collect-files');
+const {type} = require('../utils');
 const {format} = require('util');
-
-const cwd = (exports.cwd = process.cwd());
-const {createInvalidPluginError} = require('../errors');
+const {createInvalidPluginError, createUnsupportedError} = require('../errors');
 
 /**
  * Exits Mocha when tests + code under test has finished execution (default)
@@ -75,20 +74,60 @@ exports.list = str =>
   Array.isArray(str) ? exports.list(str.join(',')) : str.split(/ *, */);
 
 /**
- * `require()` the modules as required by `--require <require>`
+ * `require()` the modules as required by `--require <require>`.
+ *
+ * Returns array of `mochaHooks` exports, if any.
  * @param {string[]} requires - Modules to require
+ * @returns {Promise<MochaRootHookObject|MochaRootHookFunction>} Any root hooks
  * @private
  */
-exports.handleRequires = (requires = []) => {
-  requires.forEach(mod => {
+exports.handleRequires = async (requires = []) =>
+  requires.reduce((acc, mod) => {
     let modpath = mod;
-    if (fs.existsSync(mod, {cwd}) || fs.existsSync(`${mod}.js`, {cwd})) {
+    // this is relative to cwd
+    if (fs.existsSync(mod) || fs.existsSync(`${mod}.js`)) {
       modpath = path.resolve(mod);
-      debug('resolved %s to %s', mod, modpath);
+      debug('resolved required file %s to %s', mod, modpath);
     }
-    require(modpath);
+    const requiredModule = require(modpath);
+    if (type(requiredModule) === 'object' && requiredModule.mochaHooks) {
+      const mochaHooksType = type(requiredModule.mochaHooks);
+      if (/function$/.test(mochaHooksType) || mochaHooksType === 'object') {
+        debug('found root hooks in required file %s', mod);
+        acc.push(requiredModule.mochaHooks);
+      } else {
+        throw createUnsupportedError(
+          'mochaHooks must be an object or a function returning (or fulfilling with) an object'
+        );
+      }
+    }
     debug('loaded required module "%s"', mod);
-  });
+    return acc;
+  }, []);
+
+/**
+ * Loads root hooks as exported via `mochaHooks` from required files.
+ * These can be sync/async functions returning objects, or just objects.
+ * Flattens to a single object.
+ * @param {Array<MochaRootHookObject|MochaRootHookFunction>} rootHooks - Array of root hooks
+ * @private
+ * @returns {MochaRootHookObject}
+ */
+exports.loadRootHooks = async rootHooks => {
+  const rootHookObjects = await Promise.all(
+    rootHooks.map(async hook => (/function$/.test(type(hook)) ? hook() : hook))
+  );
+
+  return rootHookObjects.reduce(
+    (acc, hook) => {
+      acc.beforeAll = acc.beforeAll.concat(hook.beforeAll || []);
+      acc.beforeEach = acc.beforeEach.concat(hook.beforeEach || []);
+      acc.afterAll = acc.afterAll.concat(hook.afterAll || []);
+      acc.afterEach = acc.afterEach.concat(hook.afterEach || []);
+      return acc;
+    },
+    {beforeAll: [], beforeEach: [], afterAll: [], afterEach: []}
+  );
 };
 
 /**
@@ -106,6 +145,7 @@ const singleRun = async (mocha, {exit}, fileCollectParams) => {
   debug('single run with %d file(s)', files.length);
   mocha.files = files;
 
+  // handles ESM modules
   await mocha.loadFilesAsync();
   return mocha.run(exit ? exitMocha : exitMochaLater);
 };
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -18,6 +18,7 @@ const {
   list,
   handleRequires,
   validatePlugin,
+  loadRootHooks,
   runMocha
 } = require('./run-helpers');
 const {ONE_AND_DONES, ONE_AND_DONE_ARGS} = require('./one-and-dones');
@@ -285,12 +286,17 @@ exports.builder = yargs =>
         );
       }
 
+      return true;
+    })
+    .middleware(async argv => {
       // load requires first, because it can impact "plugin" validation
-      handleRequires(argv.require);
+      const rawRootHooks = await handleRequires(argv.require);
       validatePlugin(argv, 'reporter', Mocha.reporters);
       validatePlugin(argv, 'ui', Mocha.interfaces);
 
-      return true;
+      if (rawRootHooks.length) {
+        argv.rootHooks = await loadRootHooks(rawRootHooks);
+      }
     })
     .array(types.array)
     .boolean(types.boolean)
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -118,6 +118,8 @@ exports.Test = require('./test');
  * @param {number} [options.slow] - Slow threshold value.
  * @param {number|string} [options.timeout] - Timeout threshold value.
  * @param {string} [options.ui] - Interface name.
+ * @param {MochaRootHookObject} [options.rootHooks] - Hooks to bootstrap the root
+ * suite with
  */
 function Mocha(options) {
   options = utils.assign({}, mocharc, options || {});
@@ -165,6 +167,10 @@ function Mocha(options) {
       this[opt]();
     }
   }, this);
+
+  if (options.rootHooks) {
+    this.rootHooks(options.rootHooks);
+  }
 }
 
 /**
@@ -959,3 +965,46 @@ Mocha.prototype.run = function(fn) {
 
   return runner.run(done);
 };
+
+/**
+ * Assigns hooks to the root suite
+ * @param {MochaRootHookObject} [hooks] - Hooks to assign to root suite
+ * @chainable
+ */
+Mocha.prototype.rootHooks = function rootHooks(hooks) {
+  if (utils.type(hooks) === 'object') {
+    var beforeAll = [].concat(hooks.beforeAll || []);
+    var beforeEach = [].concat(hooks.beforeEach || []);
+    var afterAll = [].concat(hooks.afterAll || []);
+    var afterEach = [].concat(hooks.afterEach || []);
+    var rootSuite = this.suite;
+    beforeAll.forEach(function(hook) {
+      rootSuite.beforeAll(hook);
+    });
+    beforeEach.forEach(function(hook) {
+      rootSuite.beforeEach(hook);
+    });
+    afterAll.forEach(function(hook) {
+      rootSuite.afterAll(hook);
+    });
+    afterEach.forEach(function(hook) {
+      rootSuite.afterEach(hook);
+    });
+  }
+  return this;
+};
+
+/**
+ * An alternative way to define root hooks that works with parallel runs.
+ * @typedef {Object} MochaRootHookObject
+ * @property {Function|Function[]} [beforeAll] - "Before all" hook(s)
+ * @property {Function|Function[]} [beforeEach] - "Before each" hook(s)
+ * @property {Function|Function[]} [afterAll] - "After all" hook(s)
+ * @property {Function|Function[]} [afterEach] - "After each" hook(s)
+ */
+
+/**
+ * An function that returns a {@link MochaRootHookObject}, either sync or async.
+ * @callback MochaRootHookFunction
+ * @returns {MochaRootHookObject|Promise<MochaRootHookObject>}
+ */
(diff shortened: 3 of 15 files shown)


lib/cli/run.js
```diff
@@ -286,12 +287,18 @@ exports.builder = yargs =>
       }
 
       // load requires first, because it can impact "plugin" validation
-      handleRequires(argv.require);
+      argv.rawRootHooks = handleRequires(argv.require);
```
juergba, line 290: This should be async as well, in order to remove one of our ESM limitations. Which can't be done in `check()`. Why not use `middleware()` for this job as well?

lib/cli/run-helpers.js
```diff
@@ -73,20 +75,59 @@ exports.list = str =>
...
  * @param {string[]} requires - Modules to require
+ * @returns {MochaRootHookObject|MochaRootHookFunction} Any root hooks
  * @private
  */
-exports.handleRequires = (requires = []) => {
-  requires.forEach(mod => {
+exports.handleRequires = (requires = []) =>
+  requires.reduce((acc, mod) => {
     let modpath = mod;
     if (fs.existsSync(mod, {cwd}) || fs.existsSync(`${mod}.js`, {cwd})) {
```
juergba, line 88: This call is wrong, the signature is: `fs.existsSync(path)`

boneskull: wow, huh.  okay.

lib/cli/run.js
boneskull: ideally, we'd perform validation during the `check` phase--and it needs `handleRequires` to have been called, first, but yeah, I guess you're right.  `handleRequires` needs to be async anyway.

boneskull: addressed this

lib/cli/run-helpers.js
boneskull: fixed this

juergba:
> 4. Once Mocha is instantiated, if it is given a rootHooks option, those hooks are applied to the root suite.

We have one Mocha instance and one root suite per each child-process, right?
Isn't it possible to use one `mocha` over all child-processes?

lib/mocha.js
```diff
@@ -866,3 +872,46 @@ Mocha.prototype.run = function(fn) {
...
+    afterEach.forEach(function(hook) {
+      rootSuite.afterEach(hook);
+    });
+  }
+  return this;
+};
+
+/**
+ * An alternative way to define root hooks that works with parallel runs.
+ * @typedef {Object} MochaRootHookObject
```
juergba, line 906: Do we need this typedef for typescript? Because this file is part of our bundled browser mocha.js.

lib/mocha.js
```diff
@@ -866,3 +872,46 @@ Mocha.prototype.run = function(fn) {
...
+ * Assigns hooks to the root suite
+ * @param {MochaRootHookObject} [hooks] - Hooks to assign to root suite
+ * @chainable
+ */
+Mocha.prototype.rootHooks = function rootHooks(hooks) {
+  if (utils.type(hooks) === 'object') {
+    var beforeAll = [].concat(hooks.beforeAll || []);
+    var beforeEach = [].concat(hooks.beforeEach || []);
+    var afterAll = [].concat(hooks.afterAll || []);
+    var afterEach = [].concat(hooks.afterEach || []);
```
juergba, line 886: I haven't understood this part. Why you pass an object, not a function.
`Suite.prototype.beforeAll = function(title, fn)` expects a function, but here it gets an object.
If you run those hooks with `Runnable#run`, how can this work? The hook has already been executed and contains the return value, not the function anymore?

lib/mocha.js
```diff
@@ -866,3 +872,46 @@ Mocha.prototype.run = function(fn) {
 
   return runner.run(done);
 };
+
+/**
+ * Assigns hooks to the root suite
+ * @param {MochaRootHookObject} [hooks] - Hooks to assign to root suite
```
juergba, line 878: public or private?

lib/mocha.js
boneskull: `hooks` is a collection of all four types of hook, and has properties `beforeAll`, `beforeEach`, `afterAll`, `afterEach`.  At this point in the code, if any of these property is truthy, they are expected to be arrays of functions.  I think I could be more specific about the shape of the object; before we reach this point, `loadHooks()` (in `run-helpers.js`, iirc) does some pre-processing.

Below, we loop thru the properties (again, which are arrays of functions) and assign them to the root suite via the appropriate method in `Suite`.

lib/mocha.js
boneskull: I can move it elsewhere.  It's not intended to be in the public API at this point anyway.  VSCode's TS language server understands it (somewhat).  I just did this instead of doing it inline elsewhere, as describing more than the most trivial values in a `@param` is awkward.

Given that this functionality is intended to be specific to Node.js, maybe I should move it out?  That would be poor encapsulation, though.  I really dislike the points in the codebase where object _A_ is mucking with the internals of object _B_, and that's exactly what would have to happen if `Mocha#rootHooks()` were moved elsewhere; we'd have a `Mocha` instance, then some other method would need to reach into `Mocha#suite` and call methods there.

While this is intended to be used with Node.js, it _could_ be used in the browser.  There's no reason why a browser _can't_ call `Mocha#rootHooks()`, but the consumer would need to pass an object adhering to the  `MochaRootHookObject` shape, which is awkward.

Another way may be to do away with `MochaRootHookObject` entirely, and instead expose four new chainable methods: `Mocha#rootBeforeAll`, `Mocha#rootBeforeEach`, `Mocha#rootAfterAll`, and `Mocha#rootAfterEach`. Each would accept a single hook function.  This would probably be a more ergonomic API for a consumer (which I had not originally considered).

What do you think?

lib/mocha.js
boneskull: As it stands, it should be private, but please see [this comment](https://github.com/mochajs/mocha/pull/4244#discussion_r414029767)

boneskull:
@juergba 

> > Once Mocha is instantiated, if it is given a rootHooks option, those hooks are applied to the root suite.

> We have one Mocha instance and one root suite per each child-process, right?

This is correct.

> Isn't it possible to use one mocha over all child-processes?

It isn't possible to share a single `mocha` instance across multiple child processes, no.  There's no way to share an object reference via IPC (afaik).

It's _not impossible_ for a single child process to _reuse_ a `Mocha` instance.  In fact, PR #4245 could take advantage of @nicojs' PR #4234 to do so.  I haven't looked closely at that PR, so I'm unsure if the root suite would be reset or not.  If we were to use that, root `beforeAll` and `afterAll` hooks would only be called _once_ per child process.  In the current state of #4245, these are called for _each file_.

boneskull:
somehow accidentally closed this

boneskull:
@juergba waiting for response from you in the comments, and also pls @mochajs/core take a look.

I want to make sure

1. If we add an API like this, we are comfortable living with it for awhile, and
1. We understand why it should be a thing
1. We understand the limitations (there's currently no way, when running tests in parallel, to run a root "before all"/"after all" hook _once and only once_; they happen once per child process)

test/integration/fixtures/options/require/root-hook-defs-b.fixture.js
```diff
@@ -0,0 +1,36 @@
...
+    function() {
+      console.log('afterAll array 2');
+    }
+  ],
+  afterEach: [
+    function() {
+      console.log('afterEach array 1');
+    },
+    function() {
+      console.log('afterEach array 2');
```
raymondfeng, line 33: Do you have tests for async functions as hooks?

boneskull: not in the integration tests, but in the [unit tests](https://github.com/mochajs/mocha/pull/4244/files#diff-a8ee42e10c11a0f7455ec42d9c22503aR28)

test/node-unit/cli/run-helpers.spec.js
```diff
@@ -1,8 +1,71 @@
...
+        function f() {}
+        function c() {
+          return {
+            beforeAll: d,
+            beforeEach: g
+          };
+        }
+        async function e() {
+          return {
+            afterEach: f
```
raymondfeng, line 43: Do you have a test that f` is async?

boneskull: heh, I'll just add the `async` keyword to that one I guess.

nicojs:
I guess I'm pretty late to the party 💃, but just wanted to ask anyway. Be sure to throw me out if I'm drunk talking to random women at the bar.

Why do we need a new 'rootHooks' pattern? I think most people are defining their root hooks in setup files which they load with `--file` . Can't we just make sure to load those files in each child runner? 

Since `--concurrent` will be a new feature, it's not breaking to enforce the best practice of marking root hooks with `--file` I think.

boneskull:
> Why do we need a new 'rootHooks' pattern? I think most people are defining their root hooks in setup files which they load with --file . Can't we just make sure to load those files in each child runner?

Unfortunately, we're not guaranteed this is the case.  `--file` could be used for anything and can contain tests.  

As it stands, `--file` is not compatible with parallel mode, as it runs tests in a non-deterministic order.  Not only that, but a new `Mocha` instance is created _for each file_, which means something loaded via `--file` would run not just for every child process, but for every file of every child process.  That's fine if it just contains root hooks, but again, it's not guaranteed.

Another approach would be to load files specified with `--file` in the main process, and _check_ to ensure it doesn't contain tests.  If it doesn't, then we could use it for root hooks. So, if we loaded all of the test files in the main process (but did not _run_ them) we could enable some behavior that doesn't currently work in parallel mode (exclusive tests and a few reporters are incompatible).  I had planned on looking at that later on, if it was needed.

Philosophically, though, I don't want to make `--file` mean one thing in serial mode and another thing in parallel mode.  Adding yet another file-loading option specific to parallel mode sounded confusing for users to me.

What do you think?

boneskull:
FWIW, you could also define root hooks this way:

```js
// test/index.js

beforeEach(function() {
  // I'm a root hook!
});

require('./test-a.spec.js');
require('./test-b.spec.js');
require('./test-c.spec.js');

// run with `mocha`, no options
```

😄

nicojs:
> Philosophically, though, I don't want to make `--file` mean one thing in serial mode and another thing in parallel mode. Adding yet another file-loading option specific to parallel mode sounded confusing for users to me.

This is a great argument, thanks. 

What about adding a new argument `--setup` or `--setup-file` or `--global-setup`? I guess it's comparable to `mochaHooks` export, but then more in line with how mocha is currently working. This is also in line with how [jest's `globalSetup` file](https://jestjs.io/docs/en/configuration#globalsetup-string) works.  It would save me some discussions with colleagues that don't put global hooks in `--file` setup files. Frankly, I've searched for something like `--setup` for a couple of times on the main mocha site and always ended up concluding that `--file` was the place to put global hooks.

We could disable `describe` and `it` in setup files, to make sure nobody is misusing them. In time, we could also go one step further and log a warning if users declare global hooks in other (regular) files.

craigtaub:
> What about adding a new argument --setup or --setup-file or --global-setup?

Sounds sensible to me. Could see this adding to the user confusion around whether to use `file` or `require`. Hopefully no issues running once per child process + still use `rootHooks`.

Assuming waiting for more feedback before adding docs, which is fine.

boneskull:
@craigtaub docs are in #4246

boneskull:
re: `--setup`.  I'm not entirely opposed, but we run the risk of just adding more and more flags if we want to support some other kind of plugin.

this is why I wanted to use `--require` + a convention.  this would position mocha to define more such conventions in the future to expose more kinds of plugins.

if we use `--setup`, it can only really be used for this one thing, and if we want to do something else, then the precedent is to create yet another option.

things that could be supported via `--require` + convention include, but are not limited to:

1. easier interface definition
1. easier reporter definition
1. custom `Runner`s (e.g., a worker-thread based runner)
1. custom stats collecting
1. snapshot test hooks
1. tighter integration with assertion & mocking libraries

> Could see this adding to the user confusion around whether to use file or require

If we merged this, I think we would want to be explicit: do not use `--file` for root-level hooks.  `--file` is intended to describe a specific order in which test files run, not for root hook definition.

boneskull:
I note that you could do this, with minimal changes:

```js
// .mocharc.js

module.exports = {
  require: [
    'source-map-support',
    'something-else',
    {
      mochaHooks: {
        beforeEach() {
          this.foo()
        }
      }
    }
  ]
};
```

boneskull:
@nicojs @craigtaub What do you think of my argument [above](https://github.com/mochajs/mocha/pull/4244#issuecomment-628159188)?  Convinced or no?

nicojs:
Sure, that makes sense. Didn't know about other plugins yet. If you need a plugin system, this makes sense.

```
module​.​exports​ ​=​ ​{​
  ​require​: ​[​
    ​'source-map-support'​,​
    ​'something-else'​,​
    ​{​
      ​mochaHooks​: ​{​
        ​beforeEach​(​)​ ​{​
          ​this​.​foo​(​)​
        ​}​
      ​}​
    ​}​
  ​]​
​}​;
```

This would work, except not for `--concurrent` mode, right?

boneskull:
@nicojs 

```js
module​.​exports​ ​=​ ​{​
  ​require​: ​[​
    ​'source-map-support'​,​
    ​'something-else'​,​
    ​{​
      ​mochaHooks​: ​{​
        ​beforeEach​(​)​ ​{​
          ​this​.​foo​(​)​
        ​}​
      ​}​
    ​}​
  ​]​
​}​;
```

I'm not sure I understand.  The above would theoretically work as a `.mocharc.js` with `parallel: true` or not.  I just need to modify the codebase to know what to do with a `require` value that isn't a `string`.

boneskull:
(rather, you're right, I'd need to re-load the config file for each child process, because even if I did serialize and `eval` the damn thing it'd be missing the enclosing scope)

boneskull:
anyway, not gonna do that right now.  still: a `--require foo.js` where `foo.js` exports a `mochaHooks` prop will work with `--parallel` or without.

nicojs:
I've created a small serialization library you might want to look at for communicating between child process and main process: https://www.npmjs.com/package/surrial
We're using it in Stryker for the same reason.

Anyway, it won't solve the enclosing scope issue. You cannot simply pick up a function and serialize / deseriaze it and expect every scenario to work. 

You could, however, load the config files in child processes for this reason.

That's all totally out of scope for this PR. And off topic.

craigtaub:
> What do you think of my argument above? Convinced or no?

Yep good point, plugin system sounds useful.

lib/mocha.js
boneskull: I mean, there's no reason why the browser _couldn't_ consume `rootHooks()`.  I think it's probably the right place for it for now.

boneskull:
@nicojs Thanks.  There are a lot of serialization libs out there--I just picked one.

You're welcome to try to replace the adhoc serialization stuff with `surrial`, but we have to be careful about a) performance, b) transmitting exceptions, and b) self-referential objects--e.g., I'm not serializing the entirety of a `Suite` for performance reasons.

boneskull:
something is weird with github.  the build (on travis anyway) has passed

boneskull:
still waiting on appveyor.

boneskull:
travis passed.

PR #4246, opened 2020-04-21
add docs for parallel and root hooks

boneskull:
> (this PR needs the changes from the `boneskull/issue/2839-11ty-changes` branch)

Build might not work until that one gets merged

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,11 +18,11 @@ Mocha is a feature-rich JavaScript test framework running on [Node.js][] and in 
 
 - [browser support](#running-mocha-in-the-browser)
 - [simple async support, including promises](#asynchronous-code)
+- [run Node.js tests in parallel](#parallel-tests)
 - [test coverage reporting](#wallabyjs)
 - [string diff support](#diffs)
-- [javascript API for running tests](#more-information)
-- proper exit status for CI support etc
-- [auto-detects and disables coloring for non-ttys](#reporters)
+- [JavaScript API for running tests](#more-information)
+- [auto-detects and disables coloring for non-TTYs](#reporters)
 - [async test timeout support](#delayed-root-suite)
 - [test retry support](#retry-tests)
 - [test-specific timeouts](#test-level)
@@ -35,7 +35,6 @@ Mocha is a feature-rich JavaScript test framework running on [Node.js][] and in 
 - [auto-exit to prevent "hanging" with an active loop](#-exit)
 - [easily meta-generate suites](#markdown) & [test-cases](#list)
 - [config file support](#-config-path)
-- clickable suite titles to filter test execution
 - [node debugger support](#-inspect-inspect-brk-inspect)
 - [node native ES modules support](#nodejs-native-esm-support)
 - [detects multiple calls to `done()`](#detects-multiple-calls-to-done)
@@ -119,7 +118,7 @@ $ npm test
 A brief outline on the order Mocha's components are executed.
 Worth noting that all hooks, `describe` and `it` callbacks are run in the order they are defined (i.e. found in the file).
 
-```js
+```
 run 'mocha spec.js'
 |
 spawn child process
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -1233,6 +1262,288 @@ Prepend `--v8-` to any flag listed in the output of `node --v8-options` (excludi
...
+
+It's unlikely (but not impossible) to see a performance gain from a [job count](#-jobs-count-j-count) _greater than_ the number of available CPU cores. That said, _play around with the job count_--there's no one-size-fits all, and the unique characteristics of your tests will determine the optimal number of jobs; it may even be that fewer is faster!
+
+## Root Hook Plugins
+
+> _New in v8.0.0._
+
+In some cases, you may want to execute a [hook](#hooks) before (or after) every test in every file. Previous to v8.0.0, the way to accomplish this was to use `--file` combined with root-level hooks (see [example above](#root-hooks-are-not-global)). This still works in v8.0.0, but _not_ when running tests in parallel mode!
+
+A Root-Level Hook Plugin is a JavaScript file loaded via [`--require`](#-require-module-r-module) which "registers" one or more root-level hooks to be used across all test files.
```
craigtaub, line 1366: Some users might go straight to `(#-require-module-r-module)` section, worth making a small update there to point here? Thinking the `"Note you cannot use --require to set a global beforeEach() hook"` bit could do with an update.

boneskull: yes, good catch.  I'll update it... want to see how #4244 actually lands first though

docs/index.md
```diff
@@ -1233,6 +1262,288 @@ Prepend `--v8-` to any flag listed in the output of `node --v8-options` (excludi
...
+
+Some types of tests are _not_ so well-suited to run in parallel. For example, extremely timing-sensitive tests, or tests which make I/O requests to a limited pool of resources (such as opening ports, or automating browser windows, hitting a test DB, or remote server, etc.).
+
+Free-tier cloud CI services may not provide a suitable multi-core container or VM for their build agents. Regarding expected performance gains in CI: your mileage may vary. It may help to use a conditional in a `.mocharc.js` to check for `process.env.CI`, and adjust the job count as appropriate.
+
+It's unlikely (but not impossible) to see a performance gain from a [job count](#-jobs-count-j-count) _greater than_ the number of available CPU cores. That said, _play around with the job count_--there's no one-size-fits all, and the unique characteristics of your tests will determine the optimal number of jobs; it may even be that fewer is faster!
+
+## Root Hook Plugins
+
+> _New in v8.0.0._
```
nicojs, line 1362:
```suggestion
> _New in v7.2.0._
```

docs/index.md
```diff
@@ -1233,6 +1262,288 @@ Prepend `--v8-` to any flag listed in the output of `node --v8-options` (excludi
...
+```
+
+When run (in the default "serial" mode) via `mocha --file "./test/setup.js" "./test/**/*.spec.js"`, `setup.js` will be executed _first_, and install the two hooks shown above for every test found in `./test/**/*.spec.js`.
+
+**When Mocha runs in parallel mode, test files do not share the same process.** Consequently, a root-level hook defined in test file _A_ won't be present in test file _B_.
+
+There are a (minimum of) two workarounds for this:
+
+1. `require('./setup.js')` or `import './setup.js'` at the top of every test file. Best avoided for those averse to boilerplate.
+1. _Recommended_: Define root-level hooks in a required file, using the new (also as of v8.0.0) [Root Hook Plugin](#root-hook-plugins) system.
```
nicojs, line 1333:
```suggestion
1. _Recommended_: Define root-level hooks in a required file, using the new (as of v7.2.0) [Root Hook Plugin](#root-hook-plugins) system.
```

docs/index.md
boneskull: I think we need to discuss whether we want to do this in v7.2.0 or v8.0.0.  I would prefer the latter, for marketing reasons. 😜

boneskull:
kept these references as v8.0.0 above, and changed a reference to v7.3.0 to v8.0.0

PR #4247, opened 2020-04-22
Fix the usage of expect(a, 'to be', b)

majecty:
The current `unexpected` library does not support expect(a, 'to be', b, 'some other text') format. Since these `expect` calls are not executed in 'npm run test', it was not found before.

### Description of the Change

You can test the `expect` function in this repl: https://runkit.com/majecty/5ea039f8dd507b001a4eec95
When you run the repl, you will see this error message:

```
UnexpectedError: 
expected 1 to be 1, 'hi'
  The assertion does not have a matching signature for:
    <number> to be <number> <string>
  did you mean:
    <any> [not] to be <any>
    <string> [not] to be <string>
```


juergba:
E.g. _test/only/bdd.spec.js_ does run, see `test.node.only.bdd` in _package-scripts.js_. But the code in question is not executed because of the `.only()` feature. Just two tests at the end are run.
But it's a stupid test, since it does not compare an expected output with the actual output. Just two tests are run, I don't know wether this is correct.
Tbh this PR doesn't really help.

boneskull:
thanks for the PR.   `test/only/bdd.spec.js` looks broken to me.  I would expect the test on line 7 to run, but it doesn't. that suggests either `it.only()` _itself_ is broken, or we changed the behavior before migrating to unexpected and just never noticed.

@juergba @Munter 

Should `it.only()` run the test if a _sibling of its parent suite_ is also marked `only`?  Because right now, it dosen't.

boneskull:
Actually, know what?  We can safely remove these, because the same thing they are testing is covered in `test/integration/only.spec.js` and its associated fixtures.

majecty:
@boneskull I removed those tests!
Could you check this PR again?

boneskull:
looks like those two files weren't removed, but just removed from `package-scripts.js`? 

anyway, I'll fix this up myself if you don't do so

majecty:
@boneskull Sorry, I missed it. I removed them in a new commit.

boneskull:
build failure unrelated.  may have to do with needing to rebase onto `master`, but not necessary

boneskull (approved):
LGTM, thanks!

PR #4249, opened 2020-04-23
Issue 3689: Refactoring improving encapsulation

arvidOtt:
## Requirements

### Description of the Change

#3689  recommended to enhance encapsulation by introducing an instance method appendOnlyTest to the Suite class. This was already done and merged in #3707 .

Going a step further, the Test class should expose an instance method markOnly() which would do the call on its parent so this is not accessed directly by an other class.

### Benefits

The Test class handles calls on his own data resulting in a higher degree of abstraction and encapsulation. 

### Applicable issues

#3689

### DIFF
--- a/test/unit/test.spec.js
+++ b/test/unit/test.spec.js
@@ -102,10 +102,11 @@ describe('Test', function() {
       test.parent = {
         appendOnlyTest: spy
       };
-
       test.markOnly();
 
-      expect(spy, 'to have a call satisfying', [test]).and('was called once');
+      expect(spy, 'to have a call exhaustively satisfying', [test]).and(
+        'was called once'
+      );
     });
   });
 });
--- a/test/unit/test.spec.js
+++ b/test/unit/test.spec.js
@@ -105,19 +105,7 @@ describe('Test', function() {
 
       test.markOnly();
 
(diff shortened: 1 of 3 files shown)


test/unit/test.spec.js
```diff
@@ -83,4 +84,40 @@ describe('Test', function() {
...
+
+    beforeEach(function() {
+      sandbox = sinon.createSandbox();
+    });
+
+    afterEach(function() {
+      sandbox.restore();
+    });
+
+    it('should call appendOnlyTest on parent', function() {
```
boneskull, line 99: you can combine this and the next into a single test via [unexpected-sinon](http://unexpected.js.org/unexpected-sinon/assertions/spy/to-have-a-call-satisfying/):

```js
it('should call appendOnlyTest on parent', function() {
  var test = new Test('foo');
  var spy = sandbox.spy();
  test.parent = {
    appendOnlyTest: spy
  };
  test.markOnly();
  // you might need to remove 'exhaustively' from the string
  expect(spy, 'to have a call exhaustively satisfying', [test]).and('was called once');
});

arvidOtt: Done!

PR #4256, opened 2020-04-28
Bug 3840: --forbid-only doesn't recognize `it.only` when `before` crashes

arvidOtt:
### Requirements

### Description of the Change

Option `--forbid-only` crashes only in case of approaching describe.only but not it.only. Following the discussion in #3840 and the closed PR #3948 it was decided that this should be prevented at definition phase and not during runtime.

This change checks if the `--forbid-only` option is set prior to marking a test to be only. This is the same point at which the check is done for the suite. The error message reported will be the same as for the suite. This change also adds integration test cases to cover the situation.

### Alternate Designs

Do the check at runtime like in PR #3948. This was explicitly discussed and not wanted.

It fixes a bug.

### Benefits

It fixes a bug.

### Applicable issues

#3840

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -3,6 +3,7 @@
 var Suite = require('../suite');
 var errors = require('../errors');
 var createMissingArgumentError = errors.createMissingArgumentError;
+var createUnsupportedError = errors.createUnsupportedError;
 
 /**
  * Functions common to more than one interface.
@@ -126,14 +127,14 @@ module.exports = function(suites, context, mocha) {
         suites.unshift(suite);
         if (opts.isOnly) {
           if (mocha.options.forbidOnly && shouldBeTested(suite)) {
-            throw new Error('`.only` forbidden');
+            throw createUnsupportedError('`.only` forbidden');
           }
 
           suite.parent.appendOnlySuite(suite);
         }
         if (suite.pending) {
           if (mocha.options.forbidPending && shouldBeTested(suite)) {
-            throw new Error('Pending test forbidden');
+            throw createUnsupportedError('Pending test forbidden');
           }
         }
         if (typeof opts.fn === 'function') {
@@ -165,7 +166,8 @@ module.exports = function(suites, context, mocha) {
        * @returns {*}
        */
       only: function(mocha, test) {
-        if (mocha.options.forbidOnly) throw new Error('`.only` forbidden');
+        if (mocha.options.forbidOnly)
+          throw createUnsupportedError('`.only` forbidden');
         test.markOnly();
         return test;
       },
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -93,9 +93,6 @@ module.exports = function(suites, context, mocha) {
        * @returns {Suite}
        */
       only: function only(opts) {
-        if (mocha.options.forbidOnly) {
-          throw createUnsupportedError('`.only` forbidden');
-        }
         opts.isOnly = true;
         return this.create(opts);
       },
@@ -129,7 +126,11 @@ module.exports = function(suites, context, mocha) {
         suite.file = opts.file;
         suites.unshift(suite);
         if (opts.isOnly) {
-          suite.markOnly();
+          if (mocha.options.forbidOnly && shouldBeTested(suite)) {
+            throw createUnsupportedError('`.only` forbidden');
+          }
+
+          suite.parent.appendOnlySuite(suite);
         }
         if (suite.pending) {
           if (mocha.options.forbidPending && shouldBeTested(suite)) {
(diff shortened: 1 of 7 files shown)


lib/interfaces/common.js
```diff
@@ -165,6 +165,7 @@ module.exports = function(suites, context, mocha) {
        * @returns {*}
        */
       only: function(mocha, test) {
+        if (mocha.options.forbidOnly) throw new Error('`.only` forbidden');
```
boneskull, line 168: we're throwing a similar error here and [here](https://github.com/mochajs/mocha/blob/master/lib/runner.js#L527).  there's precedence for using [`createUnsupportedError`](https://github.com/mochajs/mocha/blob/master/lib/errors.js#L61) with checks for `this.skip()`, so instead of `throw new Error()`, please use `throw createUnsupportedError()`

test/integration/options/forbidOnly.spec.js
```diff
@@ -26,13 +26,22 @@ describe('--forbid-only', function() {
...
-    });
+    var spawnOpts = {stdio: 'pipe'};
+    runMocha(
+      fixture,
+      args,
+      function(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res, 'to satisfy', {
```
boneskull, line 37: I _believe_ you can use one of our custom assertions here, but I could be wrong

```js
expect(res, 'to have failed with output', new RegExp(onlyErrorMessage));
```

this should check for exit code > 1 and match on `STDOUT` of the child process.

arvidOtt: Works!

lib/interfaces/common.js
arvidOtt: I changed the usage to `createUnsupportedError()` in `common.js` and `runner.js`

boneskull:
This LGTM, but think @juergba should take a look

arvidOtt:
Thank you @juergba, I agree that this is handled inconsistently. 

In the [CLI Usage](https://mochajs.org/#-forbid-only) it is described: 

> --forbid-only
>
>Enforce a rule that tests may not be exclusive (use of e.g., describe.only() or it.only() is disallowed).
>
>--forbid-only causes Mocha to fail when an exclusive ("only'd") test or suite is encountered, and >it will abort further test execution.

as well as

>--forbid-only              Fail if exclusive test(s) encountered     [boolean]

I found nothing in the docs about priorities between only, grep or skip.

arvidOtt:
Considering that the usage of `.only()` and therefore `--forbid-only` (prevent people from accidentally forgetting `.only()` statements after debugging a test suite or case) as well as your [comment](https://github.com/mochajs/mocha/pull/3948#issuecomment-508073323) I would suggest that no matter if there is a grep or a skip if the `--forbid-only` option is set an error will be thrown during parsing phase. So the rule would be simple, consistent and suites as well as test would be handled the same way. 

What is your opinion on that @juergba @boneskull?

juergba:
@outsideris you implemented `shouldBeTested()`, your opinion please?

> [...] no matter if there is a grep or a skip if the --forbid-only option is set an error will be thrown during parsing phase [...]

I could live with your proposition.
But then the same logic could be applied to `pending` as well. Which here only fails on suites, but not on tests ... Pending tests fail at runtime within the runner, which makes some sense, since we have static and conditional skipping.

We have to decide:
- do we need `shouldBeTested()` for `--fobid-only`? do we need it for `--forbid-pending`?
Can we remove `shouldBeTested()` ==> @outsideris ?
- pending tests: have to be checked at runtime. Do we add additional check here at parse time?

Easy solution: we leave the pending stuff as is, and don't think or ask too much. And we remove the grep check for only-suites.

boneskull:
I agree that `--forbid-only` should be triggered even if `--grep` or `--fgrep` does causes the suite/test to be omitted.

arvidOtt:
@juergba @outsideris @boneskull I applied the following changes:
- Changing `--forbid-only` logic to be consistent for tests and suites throwing error even if omitted by `grep` or `f-grep`
- Refactoring to suite class improving encapsulation by adding markOnly() method. (Implemented similar for tests in #3689 )

Concerning the handling of `pending` I would suggest to open a new PR as this one is primarily aiming to fix the bug #3840 , however, I would be happy to support on that one as well!

lib/suite.js
```diff
@@ -493,6 +493,15 @@ Suite.prototype.appendOnlySuite = function(suite) {
   this._onlySuites.push(suite);
 };
 
+/**
+ * Marks a suite to be `only`.
+ *
+ * @private
+ */
+Suite.prototype.markOnly = function() {
+  this.parent.appendOnlySuite(this);
```
boneskull, line 502: I note that `parent` can be `null`.  I'm not sure how someone could cause this to happen, but might as well do a `this.parent && this.parent.appendOnlySuite(this)`

arvidOtt: Done in #4282

arvidOtt:
@boneskull I reverted the commits concerning the refactoring of the suites handling of `--grep`, `--fgrep` and pushed them to a new PR #4282.

PR #4260, opened 2020-04-30
Remove enableTimeout api + allow overriding a disabled timeout

craigtaub:
# Description of the Change 

Removed `enableTimeout` stuff. Use the latest `timeout` value to determine if enabled/disabled + its value. (disabled is value of 0, else is enabled).

Also fixes issue of tests not timing out when timeout is re-enabled. See associating issue https://github.com/mochajs/mocha/issues/4254.

Fixes https://github.com/mochajs/mocha/issues/4254

### Alternate Designs 

Could remove `enableTimeouts` from public-facing and use it internally, however it is easiest to have a **single source of truth** (the value) and mirror what the user uses. Having 2 different states which are very similar gets confusing.

Simplify the codebase. 

There are use-cases where it is useful to be able to re-enable and make use of a previously disabled timeout. Also seems pretty important that the feature behaves as a user would expect.

### Possible Drawbacks

You can no longer restore a previously set timeout if disabled (`enableTimeout(true)` did) and must always explicitly set it again.

The default timeout value will now never be used as the user has to set a value to re-enable it anyway. That could be a good thing.

Anyone relying on `enableTimeout` to disable or re-enable timeouts will have to swap to using the explicit `timeout`. However if it is the correct behaviour and goes into the next major release, then should be fine.

### Applicable issues

Fixes https://github.com/mochajs/mocha/issues/4254

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -208,7 +208,7 @@ Runnable.prototype.clearTimeout = function() {
  */
 Runnable.prototype.resetTimeout = function() {
   var self = this;
-  var ms = this.timeout() >= 0 ? this.timeout() : 1e9;
+  var ms = this.timeout();
 
   if (ms === 0) {
     return;
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -35,7 +35,6 @@ function Runnable(title, fn) {
   this.sync = !this.async;
   this._timeout = 2000;
   this._slow = 75;
-  this._enableTimeouts = true;
   this.timedOut = false;
   this._retries = -1;
   this._currentRetry = 0;
@@ -81,13 +80,6 @@ Runnable.prototype.timeout = function(ms) {
   var range = [0, INT_MAX];
   ms = utils.clamp(ms, range);
 
-  // see #1652 for reasoning
-  if (ms === range[0] || ms === range[1]) {
-    this.enableTimeouts(false);
-  } else {
-    // Reset now that a new timeout limit has been set
-    this.enableTimeouts(true);
-  }
   debug('timeout %d', ms);
   this._timeout = ms;
   if (this.timer) {
@@ -112,22 +104,6 @@ Runnable.prototype.slow = function(ms) {
   }
   debug('slow %d', ms);
   this._slow = ms;
-  return this;
-};
-
-/**
- * Set and get whether timeout is `enabled`.
- *
- * @private
- * @param {boolean} enabled
- * @return {Runnable|boolean} enabled or Runnable instance.
- */
-Runnable.prototype.enableTimeouts = function(enabled) {
-  if (!arguments.length) {
-    return this._enableTimeouts;
-  }
-  debug('enableTimeouts %s', enabled);
-  this._enableTimeouts = enabled;
   return this;
 };
 
@@ -234,12 +210,16 @@ Runnable.prototype.resetTimeout = function() {
   var self = this;
   var ms = this.timeout() || 1e9;
 
-  if (!this._enableTimeouts) {
+  var timeoutIsDisabled = function() {
+    return ms === 0;
+  };
+
+  if (timeoutIsDisabled()) {
     return;
   }
   this.clearTimeout();
   this.timer = setTimeout(function() {
-    if (!self._enableTimeouts) {
+    if (timeoutIsDisabled()) {
       return;
     }
     self.callback(self._timeoutError(ms));
@@ -309,7 +289,7 @@ Runnable.prototype.run = function(fn) {
     self.clearTimeout();
     self.duration = new Date() - start;
     finished = true;
-    if (!err && self.duration > ms && self._enableTimeouts) {
+    if (!err && self.duration > ms && ms > 0) {
       err = self._timeoutError(ms);
     }
     fn(err);
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -80,8 +80,14 @@ Runnable.prototype.timeout = function(ms) {
   var range = [0, INT_MAX];
   ms = utils.clamp(ms, range);
 
-  debug('timeout %d', ms);
-  this._timeout = ms;
+  // see #1652 for reasoning
+  if (ms === range[0] || ms === range[1]) {
+    this._timeout = 0;
+  } else {
+    this._timeout = ms;
+  }
+  debug('timeout %d', this._timeout);
+
   if (this.timer) {
     this.resetTimeout();
   }
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -83,9 +83,10 @@ Runnable.prototype.timeout = function(ms) {
 
   // see #1652 for reasoning
   if (ms === range[0] || ms === range[1]) {
-    this._enableTimeouts = false;
+    this.enableTimeouts(false);
   } else {
-    this._enableTimeouts = true;
+    // Reset now that a new timeout limit has been set
+    this.enableTimeouts(true);
   }
   debug('timeout %d', ms);
   this._timeout = ms;
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -208,7 +208,7 @@ Runnable.prototype.clearTimeout = function() {
  */
 Runnable.prototype.resetTimeout = function() {
   var self = this;
-  var ms = this.timeout() || 1e9;
+  var ms = this.timeout() >= 0 ? this.timeout() : 1e9;
 
   var timeoutIsDisabled = function() {
     return ms === 0;
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -210,16 +210,12 @@ Runnable.prototype.resetTimeout = function() {
   var self = this;
   var ms = this.timeout() >= 0 ? this.timeout() : 1e9;
 
-  var timeoutIsDisabled = function() {
-    return ms === 0;
-  };
-
-  if (timeoutIsDisabled()) {
+  if (ms === 0) {
     return;
   }
   this.clearTimeout();
   this.timer = setTimeout(function() {
-    if (timeoutIsDisabled()) {
+    if (self.timeout() === 0) {
       return;
     }
     self.callback(self._timeoutError(ms));
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -84,6 +84,8 @@ Runnable.prototype.timeout = function(ms) {
   // see #1652 for reasoning
   if (ms === range[0] || ms === range[1]) {
     this._enableTimeouts = false;
+  } else {
+    this._enableTimeouts = true;
   }
   debug('timeout %d', ms);
   this._timeout = ms;
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -68,7 +68,6 @@ function Suite(title, parentContext, isRoot) {
   this._afterAll = [];
   this.root = isRoot === true;
   this._timeout = 2000;
-  this._enableTimeouts = true;
   this._slow = 75;
   this._bail = false;
   this._retries = -1;
@@ -105,7 +104,6 @@ Suite.prototype.clone = function() {
   suite.root = this.root;
   suite.timeout(this.timeout());
   suite.retries(this.retries());
-  suite.enableTimeouts(this.enableTimeouts());
   suite.slow(this.slow());
   suite.bail(this.bail());
   return suite;
@@ -122,9 +120,6 @@ Suite.prototype.clone = function() {
 Suite.prototype.timeout = function(ms) {
   if (!arguments.length) {
     return this._timeout;
-  }
-  if (ms.toString() === '0') {
-    this._enableTimeouts = false;
   }
   if (typeof ms === 'string') {
     ms = milliseconds(ms);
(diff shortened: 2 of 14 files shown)


lib/runnable.js
```diff
@@ -83,7 +83,10 @@ Runnable.prototype.timeout = function(ms) {
 
   // see #1652 for reasoning
   if (ms === range[0] || ms === range[1]) {
-    this._enableTimeouts = false;
+    this.enableTimeouts(false);
+  } else {
+    // Reset now that a new timeout limit has been set
+    this.enableTimeouts(true);
```
boneskull, line 89: This is how I think it should work:

A test calls `this.timeout(100)`.  The timeout should be set to 100ms.
1. A test subsequently calls `this.timeout(0)`, the timeout should _effectively_ be disabled.  Thereafter:
    - Calling `this.timeout(100)` again would re-enable the timeout
    - Calling `this.enableTimeouts(true)` would _either_ a) be a no-op, or b) should use the default timeout
1. A test subsequently calls `this.enableTimeouts(false)`, the timeout should be disabled.  Thereafter:
    - If a test calls `this.enableTimeouts(true)`, the timeout should be re-enabled and set to 100ms
    - If a test calls `this.timeout(200)`, this will set the internal timeout value to 200ms, _but timeouts will still be disabled; in order to re-enable timeouts, `this.enableTimeouts(true)` must _explicitly be called_.

I realize, currently, the values are conflated.  They should be separate, but `Runnable` should know that a `!this._timeout` _or_ `!_enableTimeouts` will cause timeouts to be ignored.

What do you think?

cc @juergba

boneskull: Fundamentally, I think calling `this.enableTimeouts(false)` then calling `this.timeout(500)` would violate the principal of least surprise.  But I could be wrong.

craigtaub: Think I see what you mean, so value and state are 2 different things. 
- State: takes its priority from the last `enableTimeout()` only falling back to last `timeout` if none exist.
- Value: always comes from last `timeout()`.

juergba: We should deprecate/remove all this `enableTimeouts()` and `_enableTimeouts` stuff.
And use the functionality as this user does. `this.skip(0)` for disabling and `this.skip(10)` for enabling. 
It's just a simple feature, there shouldn't be any potential surprises.

craigtaub: (skip? assuming you mean `this.timeout()`). 

> It's just a simple feature, there shouldn't be any potential surprises. 

I'll have a go

craigtaub: > We should deprecate/remove all this enableTimeouts() and _enableTimeouts stuff.

As we still want to keep the notion of "disabling" it seems easiest if we keep the private `_enableTimeouts` stuff but remove the public-facing side.

craigtaub: Ok i've had a go, let me know what you guys think.

boneskull: if we're going to remove it, IMO it should be soft-deprecated for a major.

but, I think it's important to understand why it exists (see also: [Chesterton's Fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence)).  Mocha is full of these things.

from what I can tell, though, it doesn't look like it was a problem that couldn't have been solved via `--timeout 0` (see #1262).  IIRC, at the time, that didn't actually work properly, and you'd need to pass `--timeout <big-number>` instead.

an alternative which would avoid `semver-major` is:

1. leave `enableTimeouts(false)` in but make it an alias for `timeout(0)`
1. when `enableTimeouts(true)` is called, look at the parent `_timeout` until it finds a nonzero one; if it doesn't, choose the default (2000)
1. remove `_enableTimeouts`

boneskull: I can also see a valid use-case in disabling timeouts temporarily and then restoring to their last value.  you could have `--timeout 1000` on the command line, then someone wants to disable with `this.timeout(0)`, then later restore it, but do you also need to put `timeout(1000)` in your test file?  IMO the nicest thing to do here for the user would be to store the old value and restore it when calling `enableTimeouts(true)`.  so, I guess I kind of _like_ `enableTimeouts()`, but I don't like not having a single source of truth.

boneskull: I don't feel very strongly about this, though would prefer not breaking stuff, or at least providing an intend-to-break notice 😄

juergba: I like clear cuts. And to keep simple things simple. I admit it's a sharp cut, because there hasn't been any deprecation message before removing, but it's ok. As @craigtaub mentioned `enableTimeouts()` is not a documented feature in our user docs.
We don't use `enableTimeouts()` internally neither, eg. we use `timeout(0)` when disabling timeout with Node option `inspect`.
I will review later.

craigtaub: Thanks for having a look guys. To add to @juergba points:
- For the codebase it will be simpler (and easier) to maintain a single value for this throughout.
- For the users I can’t find a single mention of the “restore previous timeout value” feature anywhere in our docs (that includes API docs), just the overriding functionality. Due to that it seems unlikely to be well used (unless the user has gone through our own tests to learn more about it).

Given the above it appears as though a sharp cut should be fine, as it is good for us (roll out the change in 1 go) and good for our users (fixes the bug with overrides).
Thoughts?

juergba: You could add a note in our [docs](https://mochajs.org/#timeouts):
As of v8.0.0 `enableTimeouts()` has been removed, ...

juergba: The coverage decreased by your docs update. Does that mean our docs files are included in the coverage calculation? Which would be nonsense.

lib/runnable.js
```diff
@@ -81,10 +80,6 @@ Runnable.prototype.timeout = function(ms) {
   var range = [0, INT_MAX];
   ms = utils.clamp(ms, range);
 
-  // see #1652 for reasoning
-  if (ms === range[0] || ms === range[1]) {
-    this._enableTimeouts = false;
```
juergba, line 86: Here we check wether the value is inside of a certain range?
If not, shouldn't we set `this._timeout = 0` ?

lib/runnable.js
```diff
@@ -229,14 +208,14 @@ Runnable.prototype.clearTimeout = function() {
  */
 Runnable.prototype.resetTimeout = function() {
   var self = this;
-  var ms = this.timeout() || 1e9;
```
juergba, line 232: What was the purpose of `1e9`?

lib/runnable.js
```diff
@@ -229,14 +208,14 @@ Runnable.prototype.clearTimeout = function() {
...
+  var ms = this.timeout();
 
-  if (!this._enableTimeouts) {
+  if (ms === 0) {
     return;
   }
   this.clearTimeout();
   this.timer = setTimeout(function() {
-    if (!self._enableTimeouts) {
```
juergba, line 239: why don't you use `ms === 0` here as above?

lib/suite.js
```diff
@@ -123,9 +121,6 @@ Suite.prototype.timeout = function(ms) {
   if (!arguments.length) {
```
juergba, line 121: `Suite.prototype.timeout` is different than `Runnable.prototype.timeout`, eg. there is no range check.
I guess that's out of scope of this PR.

test/integration/timeout.spec.js
```diff
@@ -18,4 +18,15 @@ describe('this.timeout()', function() {
...
     });
   });
+
+  it('should allow overriding if disabled per-test', function(done) {
+    run('timeout-override.fixture.js', args, function(err, res) {
+      if (err) {
+        done(err);
+        return;
+      }
+      assert.strictEqual(res.stats.failures, 1);
```
juergba, line 28: Some time ago we agreed to use the `expect` style in our spec files.

lib/runnable.js
craigtaub: Docs:
> a parameter passed to `this.timeout()` greater than the maximum delay value will cause the timeout to be disabled.

Yes you are right.

lib/runnable.js
craigtaub: `"so someone could not create an async test with a timeout of 0 and have Node exit silently"` (details https://github.com/mochajs/mocha/commit/7c1139e8b47f817dd05508129d8134a9eafe027b).

lib/runnable.js
craigtaub: Flagged by [resetTimeout](https://github.com/mochajs/mocha/blob/master/test/unit/runnable.spec.js#L719) test. Sets 10, then reset, then sets 0.

Inside of the above `setTimeout`, `ms` is the initially cached 10 whereas `self.timeout()` is the correct 0.

craigtaub:
Coverage 😕

lib/runnable.js
juergba:
```js
  it('should work', function (done) {
    this.timeout(0);
  })
});
```
Above test does not work, neither with current master nor with your branch. Node exactly does exit silently.
I don't know what's wrong, neither why you just haved removed `1e9`.

craigtaub: @juergba 

> neither why you just haved removed 1e9.

So I understood the previous code to mean, if `enableTimeout(true)` and `timeout(0)` then use `1e9` default value for timeout. However now 0 == disabled so this is default value is no longer required (I thought).

> I don't know what's wrong

We have disabled the timeout errors (via early returns) for `timeout(0)` (see `resetTimeout`) so in this scenario would not get the timeout exceeded error (which would say `'ensure "done()" is called'`). 

So as I understand the issue is that if you disable timeouts but run a test which would never complete, we fail silently. 

Couple of attempts:
- If I removed all the early exits inside `resetTimeout` and put the `1e9` default back then your example test will fail correctly, but will take a hell of a long time to fail, so not a good idea.
- If I have the early return inside the `timer = setTimeout` return a timeout error, then it will throw all the time.

Im not quite sure how to resolve this. Any ideas are welcome. If we think its a bigger issue we could move into a new Issue as this one was never meant to re-define the timeout behaviour to this extent.

lib/runnable.js
boneskull: 1. `enableTimeouts()` [is documented](https://mochajs.org/api/mocha#enableTimeouts), but it is not mentioned in the "main" docs, fwiw.
1. @craigtaub pls rebase onto `master` and add `docs/**` to the `exclude` prop of `.nycrc`

lib/runnable.js
boneskull: IMO, the following test _should_ not exit silently:  

```js
  it('should work', function (done) {
    this.timeout(0);
  })
});
```

But there's not much to be done here.  Waiting `1e9` ms is just kind of dumb.  A more elegant solution would be to leverage `async_hooks` or `AsyncLocalStorage` to _check_, when the test exits, if the user has put a task into the event loop (and `done()` hasn't yet been called).  If no, then we could throw an exception.  But we can't support that behavior in a browser without some kind of an adapter (e.g., Angular's Zones).

I guess I don't think you should do anything about it unless there are open issues about silent exits.

There's a contract here with `done`: either you call it right away (or throw an exception), or you put a task in the event loop which will either a) call it later, or b) throw an exception.  Maybe the correct solution for now is to just make sure it's in writing.

boneskull: to be clear, I think the removal of `1e9` is fair, because it wasn't working as intended _at least_ since the addition of `_enableTimeouts`.

lib/suite.js
boneskull: it should be clamped tho

lib/runnable.js
juergba: I found issue #2537 (from 2016) where Node aborts execution with exitCode==0.

> Waiting 1e9 ms is just kind of dumb.

Everything is better than abort with exitCode==0, and actually it's the behaviour I would expect.

IMO you should first fix/clarify #2537 (in a different PR) making sure we really don't need `_enableTimeouts` anymore. Then fix/merge this one.
I leave it upto you. I backout of this PR.

craigtaub: A different approach is to go back to my original solution ([commit](https://github.com/mochajs/mocha/pull/4260/commits/d091cdc3f1a4b25783689302e198b2482cd8f30b)) which (1) fixed the intended bug, (2) did not remove `this.enableTimeout` api.

craigtaub:
@boneskull thanks for approval, happy to add docs, but seems we have couple of options for this:

1. Add docs and then merge - potentially removing code we will need later
2. Leave it and go fix silent exit issue in alternative PR, then come back in future date - not a fan of leaving this PR open indefinitely and that fix will not be quick
3. Revert to previous bug fix, leaving silent timeout issue for another time - I am more in favour of this tbh

boneskull:
@craigtaub 

> Add docs and then merge - potentially removing code we will need later (fixing silent exit issue)

A little confused on the wording here.  You mean that you want to address the silent exit issue in this PR (by exiting after 1e9ms)?

> Revert to previous bug fix (commit), leaving silent timeout issue for another time - I am more in favour of this tbh. If turns out we can remove enableTimeout should be relatively easy to re-apply. Involves leaving most of the enableTimeout behaviour alone for now aswell, as we hopefully can remove it in future.

Does this mean closing this PR, or just taking the documentation?

If the situation is no "worse" than it already is--meaning we still have a potential silent exit problem--I don't see any reason _not_ to merge this PR, unless I'm missing something.

craigtaub:
@boneskull 

> A little confused on the wording here. You mean that you want to address the silent exit issue in this PR (by exiting after 1e9ms)?

Ah sorry badly written, i meant keep PR mostly as is. Removing code which might be useful for fixing the silent exit issue.

> Does this mean closing this PR, or just taking the documentation?

Meant reverting all changes relating to removing `enableTimeout`, so back to [this](https://github.com/mochajs/mocha/pull/4260/commits/d091cdc3f1a4b25783689302e198b2482cd8f30b).

> If the situation is no "worse" than it already is

The worry is we remove `enableTimeout` and it turns out to be necessary to fix the silent exit issue, I just dont know what that will entail. 

Im actually now thinking if the fix for silent exits is a while away, it is fine to merge this PR as it improves the area for right now, and if we need a new state mechanism it can be added as part of the silent exit fix. Thoughts?

boneskull:
I think we could fix it with an interval which ran infinitely until the test completes or we exit via other means. essentially put a dummy timer into the event loop. this wouldn’t need another state variable. would have to play with it though

boneskull:
do whatever you think is best.

PR #4272, opened 2020-05-07
fetch sponsors at build time, show ALL non-skeevy sponsors; closes #4271

### LINKED ISSUE #4271: display all backers/sponsors on site
Currently we're limited to ~15 of each.  Another drawback is the URL behind the logo is not visible in the browser.

I frequently get complaints/queries about why a logo is not showing.  If we show all of them, then I won't.

I have a proof-of-concept which uses 11ty's data layer to pull the sponsorship information from OC's GraphQL API (the code is based on webpack's [grabber](https://github.com/webpack/webpack.js.org/blob/fba6c1f525917151411760cbd73ff21aaf1e162d/src/utilities/fetch-supporters.js)).  

Since this information won't be live--it will only fetch upon site build--we will need to build the site on a schedule to update the list.

The current status here is that I'm having trouble determining the difference between a "backer" and a "sponsor" (our two donation tiers) in the data coming out of OC's API.  I've asked about this in their support channel.  Worst-case is that the data would need manual cleaning.

So, that's what's holding up moving forward here at the moment.

(There's also some open questions, I think, around the tier structure, but that's a separate issue)

boneskull:
~~This is WIP while we sort out the data issues mentioned in #4271.~~

The data issues have been addressed.  

It appears we want to blacklist some sites from appearing on mochajs.org, because they are skeevy.  See mochajs/admin#10

This PR:

- adds a blacklist of supporters (`mochajs` appears on this list, because we somehow donated to ourself) whose links & logos will _not_ appear on our site
- fetch supporters via OpenCollective's GraphQL API (adds [needle](https://npm.im/needle) to help with making requests, but open to other ideas)
- reorders the sponsors (organizations) to show _before_ the backers (individuals).  generally the sponsors give us more $
- renames the default template (`default.html`) to `default.liquid` because it's a HTML-based liquid template, which should help with syntax highlighting.  does not affect build
- shows ALL supporters otherwise, not just top _n_
- links directly to provided URLs, if they exist.
- removes the gray background from the image container (`.faded-images`)


boneskull:
@outsideris @craigtaub @Munter This is ready

boneskull:
(click thru to deploy preview to see the result)

Munter:
The experience the first time is pretty jarring. When I don't have the images in cache, opencollective is quite slow at responding, so I just get a bunch of empty white space while they are loading. I think we should create a better fallback experience where the images at least have a background color, so I can see that something is supposed to be there. The backer images are easy, because they already have hard coded dimensions of 32x32. lets just put a grey background on there. The sponsors are more difficult, since we don't have the image dimensions as far as I can tell. Can we get those from the opencollective api? I've had a hard time finding their graphql api documentation, or a playground to explore it myself

docs/_includes/backers.md
```diff
@@ -1,7 +1,11 @@
...
 
-<!-- markdownlint-disable MD034 -->
-{% for i in (0..29) %}[![](https://opencollective.com/mochajs/backer/{{ i }}/avatar.jpg)](https://opencollective.com/mochajs/backer/{{ i }}/website){: target="_blank" rel="noopener"}{% endfor %}
-{: .image-list id="_backers" }
+{% comment %}
+Do not remove whitespace below!
+{% endcomment %}
+
+<ul class="image-list faded-images" id="backers">
+{% for supporter in supporters.backers %}<li>{% if supporter.website %}<a href="{{ supporter.website }}" target="_blank" rel="noopener" title="{{ supporter.name }}">{% endif %}<img src="{{ supporter.avatar }}" alt="{{ supporter.name }}" />{% if supporter.website %}</a>{% endif %}</li>{% endfor %}
```
Munter, line 10: If an image fails to load, we'll see a broken image icon and the alt text, which messes up the layout. Lets make the image invisible to screen readers `alt=""` and put an `aria-label` on the anchor instead

```suggestion
{% for supporter in supporters.backers %}<li>{% if supporter.website %}<a href="{{ supporter.website }}" target="_blank" rel="noopener" title="{{ supporter.name }}" aria-label="{{ supporter.name }}">{% endif %}<img src="{{ supporter.avatar }}" alt="" />{% if supporter.website %}</a>{% endif %}</li>{% endfor %}
```

docs/css/style.css
```diff
@@ -74,28 +74,39 @@ nav.badges a + a {
...
-  border: 1px solid;
-  border-color: #ddd #ddd #ccc;
-  border-radius: 3px;
-  padding: 1em;
-  box-shadow: inset 0 0 10px #ccc;
+ul#backers.image-list li img {
+  width: 32px;
+  height: 32px;
+  /* clip images to 32x32 */
+  clip-path: inset(0, 0, 32px, 32px);
```
Munter, line 109: Backers have fixed dimensions, so the images take up space. That means we can put a background color on them and let them display early as grey squares

```suggestion
  clip-path: inset(0, 0, 32px, 32px);
  background-color: #969ba2
```

`#969ba2` is the background color opencollective uses on their missing avatar images, so I'm attempting to optimize for the least change when images load

Munter:
Found the graphql api (postwoman is awesome), and I can see that we won't be able to get the dimensions that way. I think we should merge the two additions I propose and then I'll see if I can do another optimization pass now that we can post-process data at build time. We could potentially load the images and read their dimensions, so the layout doesn't jump

Munter:
I opened https://github.com/mochajs/mocha/pull/4295 against this branch, which should contain a lot of further improvements

docs/_includes/supporters.md
```diff
@@ -0,0 +1,68 @@
...
+      <img src="{{ supporter.avatar }}" alt="{{ supporter.name }}" />
+    {%- if supporter.website -%}
+    </a>
+    {%- endif -%}
+  </li>
+{%- endfor -%}
+</ul>
+
+<script>
+(function() {
```
craigtaub, line 38: Can we not keep this in external script (syntax highlighting etc)?

Munter: I'll have a look at that.

Munter: Done

boneskull:
@Munter are you intending to add any more to this?

Munter:
@boneskull let's get this in as-is. I have another idea I'm tinkering on to improve performance, but it shouldn't hold up this pr

craigtaub:
Once this is merged I will release 7.2.0.

Munter:
Sounds like we're all calling go on this. I'll merge

PR #4275, opened 2020-05-08
fix improper warnings for invalid reporters

boneskull:
also: reorganize, add, and refactor a bunch of problematic unit tests for `lib/mocha.js`.  better isolation except where we can't really do that (calling `require`)

there's still missing tests in here, but this is an improvement.

@craigtaub this touches some stuff regarding warnings for invalid reporters.  it looks like the conditionals were backwards?

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -231,24 +231,24 @@ Mocha.prototype.reporter = function(reporter, reporterOptions) {
         _reporter = require(reporter);
       } catch (err) {
         if (
-          err.code !== 'MODULE_NOT_FOUND' ||
-          err.message.indexOf('Cannot find module') !== -1
+          err.code === 'MODULE_NOT_FOUND' ||
+          err.message.indexOf('Cannot find module') >= 0
         ) {
           // Try to load reporters from a path (absolute or relative)
           try {
-            _reporter = require(path.resolve(process.cwd(), reporter));
+            _reporter = require(path.resolve(reporter));
           } catch (_err) {
-            _err.code !== 'MODULE_NOT_FOUND' ||
-            _err.message.indexOf('Cannot find module') !== -1
-              ? console.warn(sQuote(reporter) + ' reporter not found')
-              : console.warn(
+            _err.code === 'MODULE_NOT_FOUND' ||
+            _err.message.indexOf('Cannot find module') >= 0
+              ? utils.warn(sQuote(reporter) + ' reporter not found')
+              : utils.warn(
                   sQuote(reporter) +
                     ' reporter blew up with error:\n' +
                     err.stack
                 );
           }
         } else {
-          console.warn(
+          utils.warn(
             sQuote(reporter) + ' reporter blew up with error:\n' + err.stack
           );
         }
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -58,7 +58,7 @@ var mochaStates = utils.defineConstants({
  */
 
 if (!process.browser && typeof module.paths !== 'undefined') {
-  var cwd = process.cwd();
+  var cwd = utils.cwd();
   module.paths.push(cwd, path.join(cwd, 'node_modules'));
 }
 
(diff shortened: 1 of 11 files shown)


boneskull:
I should note that I started out adding more coverage and refactoring, but I found a bug while I was doing that...

craigtaub:
Odd the logic been broken for so long for such an important part. Would have been good to see an issue or details in the description to help replicate locally.

- Would this logic only really apply for programmatic use of mocha as seems `cli` will catch bad reporter when I try locally?
- Is this a semvar major if something which used to pass now fails?

lib/mocha.js
```diff
@@ -202,24 +202,24 @@ Mocha.prototype.reporter = function(reporter, reporterOptions) {
...
-            _reporter = require(path.resolve(process.cwd(), reporter));
+            _reporter = require(path.resolve(reporter));
           } catch (_err) {
-            _err.code !== 'MODULE_NOT_FOUND' ||
-            _err.message.indexOf('Cannot find module') !== -1
-              ? console.warn(sQuote(reporter) + ' reporter not found')
-              : console.warn(
+            _err.code === 'MODULE_NOT_FOUND' ||
+            _err.message.indexOf('Cannot find module') >= 0
+              ? utils.warn(sQuote(reporter) + ' reporter not found')
```
craigtaub, line 214: Possibly worth a test case for this to prove the logic works, altho i know coverage would flag if not, up to you.

boneskull: I thought I added a test for this.  hmm.

boneskull: https://github.com/mochajs/mocha/pull/4275/files#diff-0a944d7334aa10653542ebf2744ea850R525

and

https://github.com/mochajs/mocha/pull/4275/files#diff-0a944d7334aa10653542ebf2744ea850R525

boneskull:
can't use `require.resolve` in unit tests because browser.  will need to move some of these into `test/node-unit/mocha.spec.js`

boneskull:
OK, we can't use rewiremock in `test/unit/*.spec.js` because it's apparently incompatible with browserify as written.  so I've ripped it out of there, and moved some tests for `Mocha#reporter()` into `test/node-unit/mocha.spec.js` where we can safely test reporter resolution.

Ref: https://github.com/theKashey/rewiremock/issues/113

cc @Munter -- how's the transpilation coming? 😄

boneskull:
tests broke because you can't use `Object.assign()` in IE11 (I forgot), but I noticed we _also_ didn't have a lint rule preventing it, so I added one.

boneskull:
@nicojs I had heavily refactored `test/unit/mocha.spec.js`, where most of your new tests live--so they, too, are assimilated in this PR.  May want to take a look.

nicojs:
@boneskull  You're not kidding, file looks like a Christmas tree in my browser 🎄. Will checkout this branch locally to take a better look. Will have to wait till tomorrow.

boneskull:
your changes highlight that `Mocha.prototype.run` has a very strange API.

It accepts an _optional_ callback, and that callback is called with a single parameter, `failures`, a `number`.  This doesn't conform to a typical convention.  _If it did_ use an error-first style, that convention expects the function to _not_ throw, and rather call the callback w/ an `Error` instance.

It would likely be highly disruptive to break this, unfortunately.  Off the top of my head, two ways we could improve the situation:

1. The way forward may instead be to create a new `runAsync` (or whatever) in the future and return a `Promise`.  If the `Runner` instance is needed (the current return value) then this would not be appropriate.
1. In `Mocha.prototype.run`, inspect the callback.  If it exists and has an arity _greater than 1_, use an error-first style; otherwise use the current behavior.

Consumers of `Mocha.prototype.run` don't currently expect it to throw (for any reason), and the current situation may be awkward to work with.

test/node-unit/mocha.spec.js
```diff
@@ -54,28 +88,111 @@ describe('Mocha', function() {
...
 
       it('should be chainable', function() {
         expect(mocha.unloadFiles(), 'to be', mocha);
       });
     });
+
+    describe('reporter()', function() {
+      describe('when a reporter exists relative to the cwd', function() {
+        beforeEach(function() {
+          // DANGER!
```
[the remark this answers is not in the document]
boneskull: yeah, not a bad idea.

PR #4282, opened 2020-05-12
Consistent handling of --forbid-only for suites and tests

arvidOtt:
### Description of the Change

The handling of the `--forbid-only` option is currently inconsistent for the`Test` and `Suite` classes. Suites that are marked as only and excluded by `--grep`  or `--fgrep` don't throw an error. Test that are marked as only but excluded by `grep` or `f-grep` do throw an error. 

Following the discussion in #4256 it was suggested that due to the nature of the `--forbid-only` option it should not matter if a suite or test is excluded by a grep. An error should be thrown either way to protect the user from forgetting to remove `.only()` statements from his test code. I moved the changes concerning this inconsistency, including a refactoring of the Suite class to use a `markOnly()` instance method, into this PR to separate them from the fix #4256 for bug #3840.

### Alternate Designs

An alternative to make it consistent would be to adapt the logic so it will be checked if test cases are excluded by `grep` or `fgrep` as well. (see discussion in #4256)

It improves the consistency of the `--forbid-only` option and improves encapsulation between classes.

### Benefits

It improves the consistency of the `--forbid-only` option and improves encapsulation between classes.

### Possible Drawbacks

If the `--forbid-only` option is set and a test or suite is marked only but excluded by grep it will now throw an error.

### Applicable issues

#4256 
#3840

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -94,6 +94,9 @@ module.exports = function(suites, context, mocha) {
        * @returns {Suite}
        */
       only: function only(opts) {
+        if (mocha.options.forbidOnly) {
+          throw createForbiddenExclusivityError(mocha);
+        }
         opts.isOnly = true;
         return this.create(opts);
       },
@@ -130,12 +133,14 @@ module.exports = function(suites, context, mocha) {
           if (mocha.options.forbidOnly && shouldBeTested(suite)) {
             throw createForbiddenExclusivityError(mocha);
           }
-          suite.parent.appendOnlySuite(suite);
+          suite.markOnly();
         }
-        if (suite.pending) {
-          if (mocha.options.forbidPending && shouldBeTested(suite)) {
-            throw createUnsupportedError('Pending test forbidden');
-          }
+        if (
+          suite.pending &&
+          mocha.options.forbidPending &&
+          shouldBeTested(suite)
+        ) {
+          throw createUnsupportedError('Pending test forbidden');
         }
         if (typeof opts.fn === 'function') {
           opts.fn.call(suite);
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -130,9 +130,6 @@ module.exports = function(suites, context, mocha) {
         suite.file = opts.file;
         suites.unshift(suite);
         if (opts.isOnly) {
-          if (mocha.options.forbidOnly && shouldBeTested(suite)) {
-            throw createForbiddenExclusivityError(mocha);
-          }
           suite.markOnly();
         }
         if (
--- a/test/unit/suite.spec.js
+++ b/test/unit/suite.spec.js
@@ -682,8 +682,6 @@ describe('Suite', function() {
   });
 
   describe('.markOnly()', function() {
-    var sandbox;
-
     beforeEach(function() {
       sandbox = sinon.createSandbox();
     });
--- a/test/unit/suite.spec.js
+++ b/test/unit/suite.spec.js
@@ -680,6 +680,31 @@ describe('Suite', function() {
       });
     });
   });
+
+  describe('.markOnly()', function() {
+    var sandbox;
+
+    beforeEach(function() {
+      sandbox = sinon.createSandbox();
+    });
+
+    afterEach(function() {
+      sandbox.restore();
+    });
+
+    it('should call appendOnlySuite on parent', function() {
+      var suite = new Suite('foo');
+      var spy = sandbox.spy();
+      suite.parent = {
+        appendOnlySuite: spy
+      };
+      suite.markOnly();
+
+      expect(spy, 'to have a call exhaustively satisfying', [suite]).and(
+        'was called once'
+      );
+    });
+  });
 });
 
 describe('Test', function() {
--- a/test/unit/suite.spec.js
+++ b/test/unit/suite.spec.js
@@ -682,14 +682,6 @@ describe('Suite', function() {
   });
 
   describe('.markOnly()', function() {
-    beforeEach(function() {
-      sandbox = sinon.createSandbox();
-    });
-
-    afterEach(function() {
-      sandbox.restore();
-    });
-
     it('should call appendOnlySuite on parent', function() {
       var suite = new Suite('foo');
       var spy = sandbox.spy();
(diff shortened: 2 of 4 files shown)


boneskull:
@arvidOtt looks like this now conflicts with `master`

boneskull:
@mochajs/core This is one of those things that, I think, makes sense.  However, it's a) a breaking change, and b) _maybe_ not worth breaking.  Opinions?

boneskull:
Before we merge this--or any breaking change, really--we should answer these questions:

1. Who does this potentially help?
1. Who does this potentially harm?
1. For those who it harms, are there any workarounds, and what are they?

I'll try to answer these:

This change ensures that, when used with `--grep`/`--fgrep`, `--forbid-only` will cause a failure, even if the test in question is to be skipped due to not matching the grep clause.  Users who are using `--forbid-only` with `--grep` (of which there are an unknown number) may find that `--forbid-only` now catches a usage of `only` that it did not previously catch, causing tests to exit with an error.

1. This potentially _helps_ those who expect _any usage of `only` in the entire list of files_ to exit with an error.  The current behavior results in a sort of "false positive", since it won't catch the usage if the tests are skipped using `--grep`.  Upon upgrade, these users will need to remove `only`.
1. This potentially _harms_ those who expect _a usage of `only` in the list of tests to be run_ to exit with an error.  The current behavior, then, is the _expected_ behavior for this group of users.
1. For those who this negatively impacts, there is no complete workaround.  Restricting the list of files to a subset _may_ help.  Using a ESLint plugin to detect these statements can also be an effective solution.  Reserving `--forbid-only` if `process.env.CI` is set may also be a better way to consume the option.

Generally, I don't see a lot of usage of `--grep` outside of ad-hoc Mocha runs, but these usages _do_ exist. However--and this is the reason for `--forbid-only`--it's _highly unlikely_ that using `only` in e.g., a CI build is intended; it is usually a mistake.  

Since the _intent_ of `--forbid-only` is to avoid false positives in CI, I believe this PR gets us closer to that goal, regardless of any potential difficulty this change may introduce to a (likely small) subset of users.

boneskull:
@arvidOtt FWIW this needs conflict resolution again, but you may want to just wait until someone else reviews it.  Or, if it's approved by others, I can resolve the conflicts myself and merge.

arvidOtt:
Maybe one thing to highlight here is that the inconsistency in the behaviour of suites and tests also might "harm" users by unexpected behaviour.

arvidOtt:
@juergba as you brought up this point in your [comment]( https://github.com/mochajs/mocha/pull/4256#pullrequestreview-405111695) in #3840 what is your opinion on this PR aiming to solve that problem?

test/unit/suite.spec.js
```diff
@@ -680,6 +680,31 @@ describe('Suite', function() {
       });
     });
   });
+
+  describe('.markOnly()', function() {
+    var sandbox;
```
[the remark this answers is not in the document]
arvidOtt: Done. Thnx!

boneskull:
> If users are using --forbid-only in conjunction with --grep, they should be able to work around it by making multiple config's for example. We should clearly state so in the changelog, so users know what to do.

@nicojs Can you explain further?

boneskull:
@arvidOtt I've rebased to fix the conflict; please look at the result.  In another PR, I had created a `createForbiddenExclusivityError` function, which should be called instead of `createUnsupportedError`.

lib/interfaces/common.js
```diff
@@ -130,12 +133,14 @@ module.exports = function(suites, context, mocha) {
           if (mocha.options.forbidOnly && shouldBeTested(suite)) {
```
juergba, line 133: Is this part really necessary? When `mocha.options.forbidOnly` you already have thrown above.

arvidOtt: You are right, thanks! I removed the condition.

arvidOtt:
@boneskull that makes sense, thanks! 

@juergba @nicojs I integrated your feedback 👷

boneskull:
Wanted to understand what @nicojs is recommending here insofar as documentation

nicojs:
Sure, I'll right down my thoughts tomorrow.

test/unit/suite.spec.js
```diff
@@ -680,6 +680,29 @@ describe('Suite', function() {
       });
     });
   });
+
+  describe('.markOnly()', function() {
+    beforeEach(function() {
+      sandbox = sinon.createSandbox();
```
nicojs, line 686: This beforeEach and the `afterEach` beneath it should not be there. The suite already creates a `sandbox`, you're creating another one here and assigns it to the same variable. This means the `sandbox` that the suite creates will become unreachable and any stubs that were registered on it would never be `restore`d.

nicojs:
Maybe something like this.

## BREAKING CHANGE

Focussed suites (i.e. `describe.only`) are _never_ allowed in conjunction with `--forbid-only`, _even if_ it's filtered out with `--grep`. This behavior was already the case with focussed tests (i.e. `it.only`), but not for suites.

This is no longer OK to do:

```js
// foobar.spec.js
describe.only('foo', () => {
});

describe('bar', () => {
});
```

```
npx mocha --forbid-only --grep bar *.spec.js
```

**Work around**

If you want to keep using this setup, you can work around this issue by moving focussed suites to different files and configure mocha to exclude them in your CI.

```js
// only-foo.spec.js
describe.only('foo', () => {
});

// bar.spec.js
describe('bar', () => {
});
```

```
npx mocha --forbid-only --grep bar --exclude only-foo.spec.js *.spec.js
```

The same configuration could also be achieved with 2 configuration files.

```js
// .mocharc.json
{
  "spec": "*.spec.js",
  "exclude": "only-foo.spec.js",
  "grep": "bar",
  "forbid-only": true
}

// .mocharc.only.js
const config = require('./.mocharc.json');
config['forbid-only'] = false;
config.spec = ['**.spec.js'];
module.exports = config;
```

```bash
npx mocha # normal run, without foo
npx mocha --config .mocharc.only.js # run with focussed foo
```

test/unit/suite.spec.js
arvidOtt: Ah no sorry my bad. I removed them.

PR #4287, opened 2020-05-13
Use background color to show colored spaces in an inline diff

michael-brade:
### Description of the Change

The `inlineDiffs` option is really nice, showing us in color the added/deleted parts. However, if only whitespace has changed, the diffs are useless because nothing is colored. So this little change replaces all whitespace with the unicode Open Box char.

### Alternate Designs

n.a.

The functionality already is in core.

### Benefits

see above.

### Possible Drawbacks

none.

### Applicable issues

This is an enhancement.


michael-brade:
Darn. Too early. This function is also used to color the summary. I will have to fix this pull request. Don't use it yet.

Munter:
From the description I like this change very much. Maybe you could add a screenshot to highlight the difference in output.

I do have some reservations though.

We are using https://www.npmjs.com/package/diff to create the diffing output. Might such an really belong in the diffing library instead?

What happens when I copy/paste the output from the diff? Will I get spaces or will I get `\u2423` in my paste buffer? If I get `\u2423` I think that would be a problem, as we are not correctly showing in the output what was actually diffed. Is there precedence for this in other libraries?

michael-brade:
Good points. Regarding your first: no, I don't think so. `mocha` actually takes the diff and colors it after the external library has processed it. So the external lib should stay independent of this feature, it is solely on `mocha`'s side.

Regarding copy&paste: yes, obviously you will get `\u2423`. I don't know any other library that does this, I was just annoyed because I work so much with spaces in LaTeX.js, so I need this and came up with this idea. However, the colored diff is unsuitable for copy&paste anyway because you get the old and new text mixed together. Above the colored diff you get `AssertionError: expected '...' to equal '...'` and that does include the original spaces. So I would say we are fine. Screenshot coming up soon.

michael-brade:
before: 
image

after: 
image

Munter:
I like the diffs from https://unexpected.js.org/assertions/string/to-be/ better, where they are using color inversion to turn the background colored. This highlights spaces very well:

image

michael-brade:
Ah, interesting! I have wondered if I should do that instead but did not because it seems more like a break with the current way. However, if no one objects, I'll update the pull request.

Munter:
It's certainly a change that people will notice. If it's for the better or worse I guess is up to interpretation. Should probably have @mochajs/core weigh in on this

michael-brade:
This is what it would look like instead: 
image

Let me know which version to use for the final PR.

michael-brade:
Hm. However, this is indeed much more intrusive: when not using `inlineDiffs`, it looks like this: 
image

michael-brade:
It's a little bit better with black on green: 
image

Anyway, now it's up to you :grin:

boneskull:
Hmm.  Can you show black-on-red as well?

I think I like this idea better than using some random unicode character.

michael-brade:
Sure. Here: 
image

I have now updated the PR to use this last version.

craigtaub:
I think the inversion produces a clearer separation for inline diffs in addition to the fix for whitespace. My only issue is slightly noisier terminal for normal diffs as doesn't invert only specific change, unlike "unexpected", but overall I think an improvement.

boneskull:
I'm OK with it changing the colors for all diffs, but we should make sure that's the intent here (it does not seem to be, at least from the title).  @michael-brade ?

michael-brade:
Well, I don't mind either way. But it makes sense to change only the inline diff, so now I updated the code.

michael-brade:
Fixed.

lib/reporters/base.js
```diff
@@ -189,9 +191,15 @@ var generateDiff = (exports.generateDiff = function(actual, expected) {
   } catch (err) {
     var msg =
       '\n      ' +
-      color('diff added', '+ expected') +
+      color(
```
juergba, line 194: This change in the `catch` block does not make sense. We don't have any diff here, neither a normal (unifiedDiff) one nor an inlineDiff.

michael-brade: Well, I disagree. This is similar to the regular case: it uses the current color style (inline vs normal) to display the error, just like in the normal case it uses that style to explain the color above the diff with "actual - expected". So you have to give me a reason why in this case we should always use the non-inline style to display the error.

juergba: Yes, I see.
In either case the message is `+ expected - actual:  failed to generate Mocha diff`, and we don't have any diff. So I just prefer to keep it simple, there is no need to add any complexity.

michael-brade: So what does that mean, exactly? Do you want me to fix the existing code as well then by basically removing the whole thing and just returning "failed to generate Mocha diff" without any color?

juergba: No, just to revert your changes to the `catch` block.

michael-brade:
Sorry, I was in a hurry before, that's why I left no comment. Thanks for merging! :-)

PR #4288, opened 2020-05-13
fix test/unit/hook.spec.js

boneskull:
`Function.prototype.name` is not available in IE11, and will cause the browser tests to fail.

Ideally, we could catch this with ESLint, but here's hoping we don't need to do that.

cc @nicojs


boneskull:
If you're wondering why this wasn't caught, it's because SauceLabs tests only run for branch commits; not PRs.  Browser tests _do_ get run for PRs--but only in headless Chrome.

boneskull:
also: `Array.prototype.findIndex` isn't a thing either

PR #4289, opened 2020-05-15
do not commit generated content to VCS; closes #3713

### LINKED ISSUE #3713: keep auto-generated content in docs out of VCS
Our documentation has automatically generated sections.  The content in these sections is part of the doc build and should not be in VCS.

```html
<!-- AUTO-GENERATED-CONTENT:START (toc:maxdepth=2&bullets=-) -->
(stuff)
<!-- AUTO-GENERATED-CONTENT:END -->
```

Should *not* be modified when running `npm start docs`.

Instead, the build should be run against a transient copy of the file.  That means:

1. Copy `docs/index.md` to a temp location
1. Run [the preprocess script](https://github.com/mochajs/mocha/blob/e01a54eb2e5d28104625c15df641801747533b63/package-scripts.js#L284) against the temp copy
1. Ensure Eleventy generates documentation from the temp copy
1. Remove the temp copy

Take care not to deploy the temp copy itself; Eleventy should deploy HTML only, and no markdown.

(This also applies to tutorials found in the API docs, which are in incoming pull request #3655; don't worry about these for now.)

boneskull:
This removes `markdown-magic` and replaces its functionality with the built-in data capabilities of 11ty.

The TOC, `--help` output, and source files are now all done via global data scripts in `docs/_data`.

When building documentation, we will no longer get changes to e.g., `docs/index.md` because of the automatically generated content.


boneskull:
I added an official eleventy plugin which detects usages of words like "simply", "obviously", etc., which don't only make the reader feel stupid, but are often straight-up bad writing 😄
 
It emits warnings at build time and will not break the build.

boneskull:
I've confirmed the build has passed for real with my eyeballs, so merging

PR #4291, opened 2020-05-15
smart quotes and such on website; closes #3716

### LINKED ISSUE #3716: smart quotes and smart hyphens in docs
It’d be cool if we could get “smart quotes”—and hyphens—in our documentation. 

that’s probably going to be a markdown processor or eleventy plugin. ideally this would apply to API docs as well but that can be a separate issue since the tools are different (there’s another issue about consolidating the markdown tools somewhere)

boneskull:
I do not know if this actually works, but it might.  I just happened upon this option as I was fiddling with other stuff.  Let's see what the netlify build looks like.

Ref: #3716


boneskull:
This works.

PR #4293, opened 2020-05-17
Rollup and babel in build pipeline

Munter:
We want to modernize the code base, and in order to do so we need to add a transpilation step for building the browser artifacts we distribute, so we can retain our current conservative browser version targets.

The strategy I suggest is applying Babel in our bundling tool to keep the output browser compatible. I suggest switching from browserify to Babel to give us an ability to move to ES modules when we want to, but alsy get the immediate option of creating multiple distributions with different configurations of transpilation applied.

When these changes are all done we should be able to convert the code base to more modern js, allowing things like destructuring, arrow functions etc.

### Description of the Change

- [x] Clean up code paths, so node-only code isn't part of a browser bundles dependency graph
- [x] Replace Browserify with Rollup
- [x] ~~Set up eslint to keep the code node compatible at a version we decide~~
- [x] Add Babel transpilation to rollup bundles
  - [x] Configure browser output that is IE11 compatible --> `mocha.js`
  - [x] ~~Configure a more modern output for evergreen browsers --> `mocha.modern.js`?~~

### Alternate Designs

We could have retained browserify as our bundling tool, but rollup was chosen for its ability to deal with ES modules, which is a very likely future for Mocha, given its browser nature.

### Benefits

No more restrictions on authors to write ES5 and remember what browser globals were available 7 years ago

### Possible Drawbacks

Changing bundling strategy can carry a risk, especially if the bundler changes the way it resolves node global for a browser.

Applying transpilation can carry a risk of not knowing the exact output that will be generated.

We need to test these changes very well.


boneskull:
needs https://npm.im/process

boneskull:
rebased onto `master`

Munter:
This branch is ready to be merged IMO. All the transpilation config has been added, which makes the code base ready to be converted to modern JS.

I made an example branch where I ran `lebab -t arrow` on our lib files: https://github.com/mochajs/mocha/compare/munter/transpilation...munter/lebab?expand=1
The travis build for that is green, meaning that it runs in IE11, where arrow functions would normally throw: https://travis-ci.org/github/mochajs/mocha

I think we should merge this branch separately from doing the actual code modernisation, as an attempt to keep the change-set as small as possible here and make it easier to coordinate when we pull in modernisations that will likely require other branches to resolve conflicts

Munter:
The failing Netlify deploy is something about spriting opencollective images, which also happens on other branches. I'll look into that somewhere else

juergba:
> The strategy I suggest is applying Babel in our bundling tool to keep the output browser compatible.

This means our code plus the code of our dependencies will be transpiled, right?

Munter:
> This means our code plus the code of our dependencies will be transpiled, right?

I'm pretty sure the current config will also transpile `node_modules`, but I haven't actually tested it. Maybe we should make a new branch based on this one and upgrade a dependency we know will break IE11. The Travis build should tell us if it works

.gitignore
```diff
@@ -3,9 +3,13 @@ docs/_site
 docs/_dist
 docs/images/supporters
 mocha.js
+mocha.js.map
```
boneskull, line 6: why not ship this?

lib/reporters/base.js
```diff
@@ -6,15 +6,24 @@
...
 var diff = require('diff');
 var milliseconds = require('ms');
 var utils = require('../utils');
-var supportsColor = utils.isBrowser() ? null : require('supports-color');
+var supportsColor = require('supports-color');
 var constants = require('../runner').constants;
 var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
 var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
 
+var isBrowser = !!process.browser;
```
boneskull, line 17: use `utils.isBrowser()`.  this makes testing easier

lib/reporters/base.js
```diff
@@ -118,9 +127,11 @@ exports.window = {
 };
 
 if (isatty) {
-  exports.window.width = process.stdout.getWindowSize
-    ? process.stdout.getWindowSize(1)[0]
-    : tty.getWindowSize()[1];
+  if (isBrowser) {
```
boneskull, line 130: would prefer stylistic stuff like this doesn't get changed, but NBD

package-scripts.js
```diff
@@ -34,21 +34,7 @@ function test(testName, mochaParams) {
...
-      --ignore 'supports-color' \
-      --ignore './lib/cli/*.js' \
-      --ignore './lib/esm-utils.js' \
-      --ignore './lib/nodejs/serializer.js' \
-      --ignore './lib/nodejs/parallel-buffered-runner.js' \
-      --ignore './lib/nodejs/reporters/parallel-buffered.js' \
-      --ignore './lib/nodejs/worker.js' \
-      --ignore './lib/nodejs/buffered-worker-pool.js' \
-      -o mocha.js`,
+      script: `rollup -c`,
```
boneskull, line 37: 💯

scripts/karma-rollup-plugin.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
```
boneskull, line 1: - is this a fork of something? 
- if so, do we need to retain a copyright notice? license?
- is it suitable to extract into its own module at a later time?

scripts/rollup-manifest-filter.js
```diff
@@ -0,0 +1,23 @@
+/**
```
boneskull, line 1: please add some comment to the top of this file explaining what it does; I have no idea

scripts/karma-rollup-plugin.js
boneskull: also, please add a comment that explains why we need this and why the existing modules don't work for us

.eslintrc.yml
```diff
@@ -5,8 +5,9 @@ extends:
 env:
   node: true
   browser: true
+  es6: true
```
boneskull, line 8: is this needed with `ecmaVersion` below?

karma.conf.js
```diff
@@ -18,43 +19,23 @@ const browserPlatformPairs = {
...
+      'karma-mocha',
+      'karma-mocha-reporter',
+      'karma-sauce-launcher',
+      'karma-chrome-launcher',
+      rollupPlugin
+    ],
+    rollup: {
+      configFile: 'rollup.config.js',
+      include: ['test/**'],
+      bundlePath: bundleDirpath
```
boneskull, line 38: nice

lib/browser/highlightTags.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
```
boneskull, line 1: please rename to `highlight-tags.js` or `highlight_tags.js`

case-insensitive filesystems suck

lib/browser/parseQuery.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
```
boneskull, line 1: pls rename to `parse-query.js`

lib/cli/lookupFiles.js
```diff
@@ -0,0 +1,145 @@
+'use strict';
```
boneskull, line 1: pls rename `lookup-files.js`

package.json
```diff
@@ -156,11 +166,8 @@
   "browser": {
     "./index.js": "./browser-entry.js",
     "./lib/nodejs/growl.js": "./lib/browser/growl.js",
-    "tty": "./lib/browser/tty.js",
-    "./lib/cli/*.js": false,
-    "chokidar": false,
+    "./lib/esm-utils.js": false,
```
boneskull, line 169: so this browser field... I'm thinking it should just be `{"browser": "mocha.js"}`.  all of these mappings are for browserify.

rollup.config.js
```diff
@@ -0,0 +1,43 @@
+// import {resolve} from 'path';
```
boneskull, line 1: remove

rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+  ],
+
+  plugins: [
+    json(),
+    manifestFilter({
+      keys: ['name', 'version', 'homepage', 'notifyLogo']
+    }),
+    commonjs(),
+    globals({
+      // global: false
```
boneskull, line 33: remove

test/unit/parseQuery.spec.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
```
boneskull, line 1: rename `parse-query.spec.js`

.browserslistrc
```diff
@@ -0,0 +1,12 @@
+node >= 10
+last 2 Chrome versions
+last 2 Edge versions
+last 2 Firefox versions
+last 2 Safari versions
+last 2 Opera versions
```
juergba, line 6: We don't run Opera in our CI tests, right? Do you want to list Opera anyway?
This list could be added instead to our _package.json_, easier accessable for our users.

Do we need `node >= 10`?

juergba, line 6: We don't run Opera in our CI tests, right? Do you want to list Opera anyway?
This list could be added instead to our _package.json,_ easier accessable for our users.

Do we need `node >= 10`?

scripts/karma-rollup-plugin.js
```diff
@@ -0,0 +1,122 @@
...
+      includePatterns.some(includePattern =>
+        minimatch(filePath, includePattern)
+      )
+    );
+
+  const bundleFiles = [
+    ...new Set(bundlePatterns.map(pattern => glob.sync(pattern)).flat())
+  ];
+
+  // console.log({ bundleFiles });
```
juergba, line 36: remove

scripts/karma-rollup-plugin.js
```diff
@@ -0,0 +1,122 @@
...
+          served: true,
+          included: true,
+          watched: true
+        });
+        bundleInjected = true;
+      }
+    }
+  }
+
+  // console.log(fileConfigs);
```
juergba, line 69: remove

juergba:
> Maybe we should make a new branch based on this one and upgrade a dependency we know will break IE11.

@Munter I tried to make this test. But I can't get your branch `munter/transpilation` (unchanged) running on my local Windows machine. `npm test` fails at script `test.browser.unit` with code `ELIFECYCLE`. The bundling of `./mocha.js` succeeds with a warning of circular dependencies, but above script fails.

package.json
```diff
@@ -80,7 +80,13 @@
...
+    "@babel/preset-env": "^7.10.2",
     "@mocha/docdash": "^2.1.3",
+    "@rollup/plugin-babel": "^5.0.3",
+    "@rollup/plugin-commonjs": "^11.0.2",
+    "@rollup/plugin-json": "^4.0.2",
+    "@rollup/plugin-multi-entry": "^3.0.1",
+    "@rollup/plugin-node-resolve": "^7.1.1",
     "assetgraph-builder": "^8.0.1",
     "autoprefixer": "^9.7.4",
     "babel-eslint": "^10.1.0",
```
juergba, line 92: Do we still need _browserify_ and _browserify-package-json_ devDependencies?
And below the `browserify` property?

karma.conf.js
```diff
@@ -18,43 +19,23 @@ const browserPlatformPairs = {
...
     },
     reporters: ['mocha'],
     colors: true,
```
juergba, line 41: `client.mocha.opts`: the `opts` option has been removed in Mocha v8.0.0.

.gitignore
Munter: We should totally ship the source map. I'll remove this line

Munter: It's ignored in git like the other build artifacts. I did forget to add `mocha.js.map` to package.json `files` though. I'll do that

Munter: Fixed in 6993cac8

lib/reporters/base.js
Munter: Fixed in 4879b0c0

lib/reporters/base.js
Munter: I ran into trouble with accessing `process.stdout.getWindowSize` when resolving browser/non-browser stuff in rollup with the previous code structure. Otherwise I would definitely have kept this block unchanged

scripts/karma-rollup-plugin.js
Munter: Addressed in af057da9

scripts/karma-rollup-plugin.js
Munter: Fixed in af057da9

scripts/karma-rollup-plugin.js
Munter: Fixed in af057da9

lib/cli/lookupFiles.js
Munter: Fixed in 62bed67e

rollup.config.js
Munter: Fixed in f3ab8f73

rollup.config.js
Munter: Fixed in f3ab8f73

test/unit/parseQuery.spec.js
Munter: Fixed in 59aa3cb4

package.json
Munter: I forgot to clean up properly. Fixed in c052f96a

.browserslistrc
Munter: For browsers, no. But in case we're some day in a situation where our oldest node version has less capabilities than our oldest browser support, it can be useful for linting plugins.

I think I'll need to give the browserslist support range another do-over later anyway, based on a discussion we need to have about transpilation and polyfills

Munter:
> adding failOnEmptyTestSuite: false to karma.conf.js, npm test passes, but no browser tests are run. The summary always states: SUMMARY: √ 0 tests completed.
So the root problem is, the karma-runner does not execute any tests on Windows.

I'd say this is a blocker. You should be able to develop locally on windows. 

I'm a bit confused why our tests from saucelabs are not representative of your experience here.

```js
const browserPlatformPairs = {
  'chrome@latest': 'Windows 10',
  'MicrosoftEdge@latest': 'Windows 10',
  'internet explorer@latest': 'Windows 10',
  'firefox@latest': 'Windows 10',
  'safari@latest': 'macOS 10.13'
};
```

I guess it could be something with the bundling mechanism, since the saucelabs tests are run on linux bundled assets I think.

Maybe we need appveyor to run an additional test of karma and a local headless chrome

juergba:
My local output shows:
```
nps is executing `test.browser.unit` : cross-env NODE_PATH=. karma start --single-run
CI mode disabled

START:
babelHelpers: 'bundled' option was used by default. It is recommended to configure this option explicitly, read more here: 
https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers
Generated an empty chunk: "_rollup:plugin-multi-entry:entry-point"
Generated an empty chunk: "_rollup:plugin-multi-entry:entry-point"
20 06 2020 16:00:19.035:INFO [karma-server]: Karma v4.4.1 server started at http://0.0.0.0:9876/
20 06 2020 16:00:19.046:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
20 06 2020 16:00:19.070:INFO [launcher]: Starting browser ChromeHeadless
20 06 2020 16:00:22.356:INFO [HeadlessChrome 83.0.4103 (Windows 10.0.0)]: Connected on socket NylQbcaiMkoAHu2QAAAA with id 
56280182

Finished in 0.012 secs / 0 secs @ 16:00:22 GMT+0200 (GMT+02:00)

SUMMARY:
√ 0 tests completed
```
These two empty chunks look odd. In Travis' logfile you don't find such a warning.

juergba:
What about this [rollup-plugin-glob-import](https://github.com/jlmakes/karma-rollup-preprocessor/issues/49#issuecomment-607974954)?
Doesn't it bundle multi test files (also glob's) into one single test file?

Munter:
> What about this rollup-plugin-glob-import?
> Doesn't it bundle multi test files (also glob's) into one single test file?

@juergba That plugin works on globs in import statements in code. Useful if you want to bundle thing you can't statically analyse the need for. Like icons or dynamic content etc. We could possibly use it, but only if we created a new entry file for each of our test branches from Karma and import those globs. I'm unsure how that would work with a new wrapper layer i import/exports. It didn't seem very appealing

Munter:
> These two empty chunks look odd. In Travis' logfile you don't find such a warning

@juergba I think I might have messed something up that caused the different windows paths with backslashes not to match in the plugin. I'll be at the office tomorrow and will bring home a windows laptop so I have a setup I can recreate and fix it in.

I still think it would be a good idea to add an appveyor test for running headless chrome tests on windows

package.json
Munter: The rollup plugins use this field the same way browserify did. Removing these confiurations increases the bundle size by 200-300kb. So these remappings for a browser build are neede. But I'll see about getting them moved into the relevant rollup pllugins configuration instead so it becomes more obvious

boneskull:
I'll give this a try on windows too

boneskull:
I'm running this side-by-side on mac and windows, and the result of `nps build` is _different_.  windows builds one thing, and mac builds something else.

boneskull:
alright, the difference is just some object prop ordering.  

but I can definitely reproduce the "can't run browser tests on windows" problem :(

boneskull:
it creates an empty bundle.  we see warnings

```
Generated an empty chunk: "_rollup:plugin-multi-entry:entry-point"
Generated an empty chunk: "_rollup:plugin-multi-entry:entry-point"
```

the contents of the bundle:

```js
(function () {
	'use strict';



}());
//# sourceMappingURL=f16fa37b-ce80-4765-8342-10dd7125ff93.rollup.js.map
```

boneskull:
OK, 008da88 should fix it.  @juergba can you give it a shot?

boneskull:
I'm unsure what the root cause is here.  perhaps there's something in karma that demands POSIX paths?

juergba:
Yes, `npm test` is now running successfully on my Windows laptop. 👍 

I have a good feeling about the first bundling step (building _mocha.js)_. The file is bigger though than before. But I'm not sure about the second one (squeezing our test files into _karma)._ This approach does seem too complicated to me.

juergba:
> Maybe we should make a new branch based on this one and upgrade a dependency we know will break IE11.

This test also passes successfully, see branch `juergba/transpilation`.

boneskull:
(someone needs to rebase this manually before merging)

boneskull:
I marked this `patch` but I'm not entirely confident it won't break somebody.

@mantoni I wanted to let you know that this exists and will be merged soon.  I am not sure if this is going to be good or bad for you, or neither.  😄

mantoni:
@boneskull Thank you for thinking of me. My current experiments with a rewrite of Mochify depend on the pre-bundled version of Mocha. So this won't be an issue.

karma.conf.js
Munter: Fixed in 6be6606c

package.json
Munter: Fixed in 207ef63b

Munter:
I'd like to see if I can silence a few of the rollup warnings before we merge

PR #4295, opened 2020-05-19
Improve image loading behaviors

Munter:
I looked through what we are doing with loading opencollective avatars and found some improvements.

- Sponsor logos now have dimensions assigned to avoid page reflows as they load
- Removed the behavior where we'd wait for all logos to load before showing any of them
- Logos now fade in individually, starting out with a gray background
- Sponsor logos and backer avatars now also show with javascript disabled
- Moved backers and sponsors into the same file and put the javascript into it as well, so it's easier to see what the script is working on.
- Inlined the avatar javascript so the loading behavior is optimized
- Set lazyload on all other content images


docs/_data/supporters.js
```diff
@@ -112,6 +113,16 @@ module.exports = async () => {
       {sponsors: [], backers: []}
     );
 
+  // Fetch images for sponsors and save their image dimensins
+  await Promise.all(
+    supporters.sponsors.map(async sponsor => {
+      for await (const chunk of needle.get(sponsor.avatar)) {
+        sponsor.dimensions = imageSize(chunk);
```
boneskull, line 120: nice.  yeah, they come down with a fixed height, but we don't know the width.

docs/_includes/supporters.md
```diff
@@ -0,0 +1,68 @@
...
+      <img src="{{ supporter.avatar }}" alt="{{ supporter.name }}" />
+    {%- if supporter.website -%}
+    </a>
+    {%- endif -%}
+  </li>
+{%- endfor -%}
+</ul>
+
+<script>
+(function() {
```
boneskull, line 38: I assume this still gets squashed in production.

Munter: Actually I'll just need to check up on the bundling there. I'd want this one to actually stay in place and inlined, which gives us the least amount of lag time between rendering the things it works on and initiating the effect it creates

docs/_data/supporters.js
Munter: Yup. HAving the width makes all the difference for being able to have the browser finalize the height of the section because it knows what space to reserve

boneskull:
feel free to push again to my branch

boneskull:
rather, directly, no need to PR

PR #4298, opened 2020-05-22
esm entrypoint with named exports

giltayar:
### Description of the Change

Now that Mocha supports ESM test files, I found that _importing_ Mocha in ESM files is awkward, if you're like me and don't like to rely on the `it`/`before`/... globals. So, if you used to do this in CJS:

```js
const {it, before} = require('mocha')
```

You'd like to do this in ESM:

```js
import {it, before}  from 'mocha'
```

But you can't, because Mocha doesn't have an ESM entry point, and relies on the fact that CJS files can be imported, but only using default import. So you have to do this:

```js
import mocha  from 'mocha'
const {it, before} = mocha
```

This PR exports an ESM entry point so that you can use named exports in a natural manner when importing Mocha itself.

The way it does this is to add an `exports` section to the `package.json`, as described in the [Node.js documentation for exports](https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_package_entry_points). This exports has a "conditional export" that defines separate entry points for ESM and CJS.

Note that `exports` affects both ESM _and_ CJS, so we should be careful (see ramifications below).

### Alternate Designs

There is no alternate design, as this is the only way to do what are called dual-mode packages in ESM.

Well, where else can it be? ☺️

### Benefits

Importing mocha in a manner that is natural for ESM.

### Possible Drawbacks

Based on the case of `is-promise`, theoretically this can break anybody using Mocha, because adding `exports: ...` to `package.json` affects CJS too! It means that the _only_ entrypoints allowed (in CJS and ESM) are the ones defined in the `exports` .

Which is why I added `"./": "./"` to the `exports`, which says to Node.js: allow any deep importing such as `require("mocha/foo/bar")` in the case that somebody, somewhere, is using that. This was also the solution used by `is-promise`.

But this has not been tested, as Mocha doesn't have an E2E test that checks that the package is OK. The way to write this test is to do an `npm pack` that creates a `tar.gz`, and then create a package that has as a dependency that `tar.gz`, and then test whatever we want in there. This _ensures_ that the package _as a package_ (without the original source code) is OK. I did not add this test, as it feels way out of scope of this little addition.


boneskull:
ty @giltayar.

I started fiddling with [smoke-test](/boneskull/smoke-test) even before I saw this PR... should probably get it done.  there are other efforts in the works too by various individuals in the community

boneskull:
We _do_ have `test/smoke`, which we should be able to use in the meantime.  I'll give that a try.

boneskull:
@giltayar in the future, you can create branches in the `mochajs/mocha` repo; just prefix them with `giltayar/`, e.g., `giltayar/node-esm-support`.  this will give us better test coverage, because the full test suite (on saucelabs) cannot run for PRs alone

boneskull:
@giltayar I've added the missing `index.mjs` to the `files` prop of `package.json` ;)  

rewrote the smoke tests to install mocha from the output of `npm pack`

boneskull:
and rebased onto `master`

boneskull:
@giltayar does `"./"` allow `package.json` as well?  I want to be very sure about this.

giltayar:
> ty @giltayar.
> 
> I started fiddling with [smoke-test](/boneskull/smoke-test) even before I saw this PR... should probably get it done. there are other efforts in the works too by various individuals in the community

What you call "smoke test", I call E2E tests. Oh, well: testing terminology has always been a bit vague.

In any case, my personal preference is for _all_ tests to be part of the `npm test` test suite, and to never have tests that run in CI (unless absolutely necessary, like Windows-only tests).

But anyway, thanks for adding the ESM smoke test and the missing `index.mjs`!

giltayar:
> @giltayar does `"./"` allow `package.json` as well? I want to be very sure about this.

Yes. I just checked it locally on a dummy module, and it works (it allows `require`-ing the `package.json` of a package). I knew it would, but fear of another `is-promise` made me double-check 😀.

package.json
```diff
@@ -2,6 +2,14 @@
...
   "version": "7.2.0",
   "description": "simple, flexible, fun test framework",
+  "main": "./index.js",
+  "exports": {
+    ".": {
+      "require": "./index.js",
+      "import": "./index.mjs"
+    },
+    "./": "./"
+  },
```
ljharb, line 12: For maximum compat, including node v13.0 and v13.1:
```suggestion
  "exports": {
    ".": [
      {
        "require": "./index.js",
        "import": "./index.mjs",
        "default": "./index.js"
      },
      "./index.js"
    ],
    "./": "./"
  },
```

giltayar:
I tried @ljharb's utility and got some strange results. So I played around with `require`-ing mocha and seeing what it really requires, and found out something: `require`-ing `mocha` when running regular code is _not_ the same as `require`-ing `mocha` when running _under_ Mocha.

When running _under_ Mocha, you get additional properties (`before`, `after`...) which you get _only_ when running under Mocha. Not sure why they aren't exported in both cases.

What should be done is to add both of them to the `index.mjs`: both the additional "dynamic" properties (which I have already in the `index.mjs`), and the "static" properties (which I'm going to add right now).

package.json
giltayar: Done

boneskull:
yeah running npm-pack-style tests locally would be good. was easier to hack it out in a shell script for now

ljharb:
@giltayar if you find anything broken, please do file an issue :-)

package.json
```diff
@@ -2,6 +2,18 @@
   "name": "mocha",
   "version": "7.2.0",
   "description": "simple, flexible, fun test framework",
+  "main": "./index.js",
+  "exports": {
+    ".": [
+      {
+        "require": "./index.js",
```
coreyfarrell, line 9:
```suggestion
```

Providing the `require` condition will cause an experimental warning for `Conditional exports` to be emitted from node.js 13.7.0, 13.8.0 and 13.9.0 (when conditional exports were unflagged but still warned).  This warning would be emitted not just for `import mocha from 'mocha'` but also for `require('mocha')`.  Dropping the `require` condition will cause `require('mocha')` to use the `default` condition which never had a warning.

Sorry I didn't get a chance to post here earlier, I wasn't on a PC.

boneskull:
@giltayar errrr... hmm.  upon further consideration, there are problems here:

- Mocha tests are not "node-able".  You cannot run `node test.js` where `test.js` imports/requires `mocha` and expect it to work like it would w/ `mocha`.  you _must_ use `bin/mocha` or must otherwise manually bootstrap it.
- `describe`, `it`, etc., are put into the `global` object dynamically.  furthermore, they are dynamically injected into `module.exports` of the main package entry point!  this is because different interfaces (`bdd` being the default, but there's `tdd`, `exports`, `qunit`, etc.) use different APIs.
- which interface is exported is not determined until Mocha is fully bootstrapped. meaning `require('mocha')` in a test loaded with the `qunit` interface _will have a different API_ than one loaded with the `bdd` interface!
- this works similarly, but even more weirdly, in a browser (`module.exports = global`, anyone?)

**I think we should carefully consider how we want to export APIs for ESM instead of trying to mimic what happens in CJS.**  

Keep in mind, this is new code, so we're not going to break anyone.

But, IMO, `describe`, `it`, `beforeEach`, etc., _should not be statically exported_ as we have here.

## Alternative A

```js
// mytest.spec.mjs
import {describe, beforeEach, it} from 'mocha/bdd';
import {suite, suiteSetup, test} from 'mocha/tdd';
// etc
```

with the documented caveat that you _must_ run your code via `bin/mocha` for this to work.  

## Alternative B

Make tests _work_ via `node test.spec.mjs`.  Again, we have the opportunity to do this because it's new and won't break existing users.  I am not sure what that implementation will look like, but it will necessarily not be as feature-rich as running with `bin/mocha`.  _Personally_, I don't really know if this is worth the bother.  People _do_ want this (we've had an open issue for many years), but it feels more like tossing a bone to the "don't use globals, ever" crowd instead of providing something that's widely useful.

## Alternative C

Don't export anything other than what `lib/mocha.js` stuffs into `module.exports` _before_ bootstrapping (no interface APIs).  If someone wants to consume mocha via ESM, they can use the provided API to do their own bootstrapping.  This will break the "ESM-style smoke test" I've added.

Thoughts?  @nicojs @Munter @craigtaub @outsideris

boneskull:
Thanks @coreyfarrell and @ljharb for looking at this

boneskull:
Mocha is a great example of both the _advanced flexibility_ and _wanton abuse_ of Node.js' CJS module system.

giltayar:
There is **Option D**:

export _all_ interfaces, and let the user choose whichever one they want, which is basically what they do today. I'm not sure how Mocha decides which ones to really export (I always used `describe/it/...`), but maybe we can just export all of them? While not the optimal solution, I'm not sure there _is_ an optimal solution, and given that the interface to the module is dynamic, and we have to "staticize" it, I believe this to be the simplest.

Moreover, it'll probably be the least surprising to users, who got used to doing `const {...} = require('mocha')`. They would be surprised that they need to do something other than `import {...} from 'mocha'`.

package.json
giltayar: Oh, wow. This is really bad. Thanks for the heads up! And I thought I was pretty up to date on ESM and conditional exports, and here I am missing stuff that would have _broken Mocha_ (and I consider such a warning as breaking Mocha). But why would removing the `require` section and keeping the `default` one _not_ show a warning? A fluke that somehow benefits us, or was it intentional?

coreyfarrell: "Conditional exports" is a feature which expands on "exports".  On it's own "exports" gives the ability to remap any specifier (not just main).  So like main can say what `require('mocha')` resolves to, `"exports": {"./tdd": "./lib/tdd.js", "./": "./"}` would say that `require('mocha/tdd')` would actually load `node_modules/mocha/lib/tdd.js`.  On it's own "exports" didn't differentiate between import and require.  "default" was actually part of the original "exports" feature, thus no warning.

Support for "require" and "import" conditions were added as part of conditional exports (possibly "node" as well but I don't remember for sure on that one).  Declaring the "import" condition in package.json isn't enough to trigger the warning but `import('mocha')` on node.js 13.7.0 - 13.9.0 will warn about conditional exports being an experimental feature.  With this latest change `require('mocha')` will will fallback to "default" thus no warning.

craigtaub:
I agree **D** will be the least surprising to users. But my preference is **A** as it introduces more clarity between the interfaces (both require the mocha executable).

Also agree regarding **B**, having interface as point-of-entry bootstrapping the parsing + runner execution phases sounds like introducing more complexity, I feel we should only do this route if we see real value.

giltayar:
I feel like there's either *A* or *D*. I personally would prefer *D*, because I like simple and unsurprising, but of course, whatever will be chosen, I will implement.

My preference for *D* is also thinking about the future, where we'll be getting questions (StackOverflow and such) about why isn't `describe` exported from ESM Mocha, whereas it _was_ exported in _CJS_. Now that I understand, I can definitely explain, but (i) I'll probably have to do the explaining again and again, and (ii) not everybody will have (or take) the time to understand.

craigtaub:
> why isn't describe exported from ESM Mocha, whereas it was exported in CJS

Think thats a fair point. To counter my own argument, there aren't that many interfaces to muddle up, so do not think any added clarity is worth future confusion. Im happy with either A or D.

boneskull:
To be very clear, if you want D:

- Not all interfaces are exported when using `require('mocha')`; only the interface you have chosen (by default that's `bdd`, with `describe`, etc.). This change would not retain "parity", but rather go _beyond_ what commonjs does. 
- If D is implemented, all methods of all interfaces will be exported, but the only ones that will _work_ are those from the chosen interface.  What the others will do is unknown to me, but certainly one of a) throw an exception, b) silently fail, c) something else.

I thought of another way.   It would not require us to export things that are broken, however.

## Alternative E

```js
// or some other name
import {currentInterface} from 'mocha';

// in bdd mode
const {describe, it} = currentInterface();

// in tdd mode
const {suite, test} = currentInterface();
```

Functions can be dynamic.  ESM exports can't.

boneskull:
> My preference for D is also thinking about the future, where we'll be getting questions (StackOverflow and such) about why isn't describe exported from ESM Mocha, whereas it was exported in CJS. Now that I understand, I can definitely explain, but (i) I'll probably have to do the explaining again and again, and (ii) not everybody will have (or take) the time to understand.

Perhaps instead of explaining this over and over, we can direct them to a `faq`-labeled issue, add it to the docs, add it to the wiki, etc.  The gist of which would be:

1. **You do not need to import these variables; they are global**
1. If you must do it anyway, do _x_
1. This is because `require()` is a function, run at runtime; `import` is static.  Mocha exports `describe` and `it` _dynamically_, depending on the chosen interface.

While I expect some questions about this too, I'd rather answer them then ship an API containing a large % of broken functions.  

AFAIK the reason people do this in the first place is for one of two reasons:

1. Their tools don't know that `describe` is a global function, and this helps their dumb tools work.
1. Anti-global-function dogma

I'm not sure it serves the project to bend over backwards for either of these use cases--at least until we can offer an API that works without the `mocha` executable.

boneskull:
Really wish we could detect usages of `require('mocha')` in test files and print a warning message that says "come tell us why you are doing this", so I could be more confident in my dismissal 😄

boneskull:
(Is this a webpack problem?)

boneskull:
@giltayar Why don't you like "relying on globals", as you write?

boneskull:
Note that providing a different API here might break TypeScript users.  I'm unsure if people are doing `import {describe, it} from 'mocha'` in TypeScript, and what would happen if they changed their compile target to use `import` instead of `require`... sigh

giltayar:
> @giltayar Why don't you like "relying on globals", as you write?

1. Linters need to know about these globals. Yes, I know I can config the linter to "know" about Mocha, but I am trying to do _less_ configuration, and not more.
2. Importing/requiring also gives me Intellisense "for free", without any need to configure anything.
3. Because it makes sense intuitively, and conforms to how we do things in JavaScript today. Things "just work" when I require/import.

So my take on this is actually the inverse: I don't understand why developers choose _not_ to require/import. 😀

giltayar:
> Note that providing a different API here might break TypeScript users. I'm unsure if people are doing import {describe, it} from 'mocha' in TypeScript, and what would happen if they changed their compile target to use import instead of require... sigh

I did a bit of Typescripting, and used `import {describe, it} from 'mocha'` in TypeScript. No problems, given that it transpiles it to `require`. Also, typechecking worked.

giltayar:
> Perhaps instead of explaining this over and over, we can direct them to a faq-labeled issue, add it to the docs, add it to the wiki, etc.

True, but people rarely read documentation. The best documentation is no documentation.

My rule of thumb here is "the principle of least surprise": choose the option that will least surprise people.

giltayar:
> Not all interfaces are exported when using require('mocha'); only the interface you have chosen (by default that's bdd, with describe, etc.). This change would not retain "parity", but rather go beyond what commonjs does.

Yup, and if you `const {test} = require('mocha')` you get `undefined`. Just like in option D.

> If D is implemented, all methods of all interfaces will be exported, but the only ones that will work are those from the chosen interface. What the others will do is unknown to me, but certainly one of a) throw an exception, b) silently fail, c) something else.

I tried. You get `undefined`. Just like in option D.


Please note: if Mocha had _from the start_ had by default mandated `mocha/bdd` and `mocha/tdd`, then that is what I would use for ESM. But, currently, Mocha goes for the "export everything and some of it will be undefined dependending on how you choose to use Mocha". And I think this should be the same for ESM.


How about Option F?

## Option F

importing (and requireing!) from `mocha/bdd` and `mocha/tdd` are the new recommended way of getting the api, but importing (and requireing) `mocha` will give you a backward compatible way, where some of the api is `undefined`.

I believe this gives us the best of both worlds:

* Parity between ESM and CJS.
* Principle of least surprise is maintained.
* The "better way" (import `mocha/bdd`) is clearly documented and gives you _exactly_ what you wanted: perfect conformance of API to the interface.
* If Mocha is run with one interface, and the developer imports another, we can theoretically throw an exception.

craigtaub:
Ok im fairly new to ESM, but as I understand

Currently with CJS:
You can configure the mocha interface to BDD and therefore have methods "describe" and "it" available globally. However you can also `require` TDD methods "suite" and "test" from the entrypoint and use them separately. 

Future with ESM: 
A user will get `undefined` if they try and import TDD methods "suite" and "test" while on a mocha instance defaulting to BDD, via the entrypoint. However no problems importing and using BDD here. If they were to ESM import TDD methods via `mocha/tdd` it would work.

If this is right, I can definitely see the value in **F**, specially if we could add some exception handling for incorrect interface importing from entrypoint.

boneskull:
I can live with F, I suppose.  I still don't love it, but there's nothing that says you can't have the value of a named export be  `undefined` and then set it later.

@craigtaub You OK with F? @juergba @Munter @outsideris any other input?

boneskull:
Also: should we target this for v8.0.0?

giltayar:
I'll start working on F. 

BTW, if this is a semver major version, do we want to remove the `"./": "./"` fallback in `exports` that allows deep-linking into Mocha? I say yes, and will implement it thus, but want your decision.

## Summary of F

(For context, here's the link to the documentation of "exports", which allows these shenanigans: https://nodejs.org/api/esm.html#esm_package_entry_points)

1. New entrypoints: `mocha/tdd`, `mocha/bdd`, `mocha/qunit`... Each will export only the functions that are used by these interfaces. This will be implemented for CJS and ESM. **I made need help in understanding which functions belong to which interfaces**.
2. The `mocha` entrypoint will export _all_ functions (a union of all the above exports), but some will be undefined based on the interface the user chose. This is already implemented in CJS and ESM in the current PR.
3. The documentation will have a new section about importing/requiring Mocha as opposed to using the globals, and will describe all the above entrypoints.
4. Those will be the _only_ entrypoints allowed into Mocha. Anyone trying to deep link to any other place will fail (fail only from Node v13, where the "exports" feature is implemented).
5. The exception to point 4 is `package.json`, which _will_ be exported. There's a whole discussion about this in the Node modules Working group (https://github.com/nodejs/node/issues/33460), as it seems that a lot of tooling needs access to all package's `package.json` and packages using "exports" feature is killing this. They're thinking of making `package.json` an exception and allowing it to always be `require`-ed, but in the meantime we can allow it so as not to break tooling.

giltayar:
Just to be sure I was understood: my starting to work on F in no way is intended to force a decision. If y'all decide otherwise, I'll happily `git reset --hard`. ☺️

giltayar:
This is what will be exported in the various entrypoints, according to my understanding:

```js
import Mocha from './index.js';

// API stuff (to only be exported by import/require("mocha")
export const utils = Mocha.utils;
export const interfaces = Mocha.interfaces;
export const reporters = Mocha.reporters;
export const Runnable = Mocha.Runnable;
export const Context = Mocha.Context;
export const Runner = Mocha.Runner;
export const Suite = Mocha.Suite;
export const Hook = Mocha.Hook;
export const Test = Mocha.Test;
export const unloadFile = Mocha.unloadFile;

// functions used by BDD tests (to be exported via "mocha/bdd")
export const describe = Mocha.describe;
export const context = Mocha.describe;
export const it = Mocha.it;
export const specify = Mocha.it;
export const xit = Mocha.xit;
export const before = Mocha.before; // also used by QUNIT tests ("mocha/qunit")
export const after = Mocha.after; // also used by QUNIT tests  ("mocha/qunit")
export const beforeEach = Mocha.beforeEach; // also used by QUNIT tests  ("mocha/qunit")
export const afterEach = Mocha.afterEach; // also used by QUNIT tests  ("mocha/qunit")

// functions used by TDD tests (to be exported via "mocha/tdd")
export const suite = Mocha.suite; // also used by QUNIT tests  ("mocha/qunit")
export const test = Mocha.test; // also used by QUNIT tests  ("mocha/qunit")
export const suiteSetup = Mocha.suiteSetup;
export const suiteTeardown = Mocha.suiteTeardown;
export const setup = Mocha.setup;
export const teardown = Mocha.teardown;
```

Please help me by verifying that the above is correct (or not).

giltayar:
I implemented option F, including documentation.

The build passes, except for netlify, and I don't have access to the Netlify build, so I can't figure out what the problem is (again).

giltayar:
Not esure why the Travis build is "pending", because it's not. It passed.

boneskull:
@giltayar we've had issues with CI talking to github lately.  

also dunno what is going on with netlify

boneskull:
We need to revisit this, but it won't make it until v8.3.0 at the earliest.

ljharb:
Adding the "exports" field - which is required for ergonomic ESM importing - is very difficult to do without it being a breaking change, so if it didn't make v8, it probably has to wait until v9 (just adding the entry point which is only accessible via a deep import with an extension could ofc be a minor)

boneskull:
so what exactly breaks with the addition?

ljharb:
You can try adding it locally, and then running `npx ls-exports path .` to see the differences (add `--json` for a detailed report). Basically, the default CJS resolution exposes a *lot* of specifiers, every single one of which is part of your observable public API, and "exports" excludes everything except explicitly listed entry points.

giltayar:
@ljharb I explicitly did _not_ want to break backward compatibility, which is why I allowed "deep linking" by adding `"./": "./"` to the end of the `exports`. This _should_ make it backward compatible and allow anybody to `require` anything they want in the mocha dir. Obviously, if this will be in a semver-major version, we _could_ discuss limiting deep linking and allowing only "formal" entry points, which would be the best.

Sorry for the radio silence about this. I would really like to press forward with this, and will be resolving all the conflicts here in the coming week. I would love for this to be merged.

ljharb:
Do note that in the next node 15, slash patterns will be runtime-deprecated, without which it is impossible to add “exports” in a non-breaking way.

giltayar:
Rebased, and tests passing!

giltayar:
@ljharb I just noticed what you said about deprecating slash patterns. That is really bad. Is there no pattern I can use that guarantees backward compatibility? Some kind of `**/**` or something?

I can remove the `"./": "./"` mapping in the `exports` section, but that would mean breaking backward compatibility and making this a semver-major change. But maybe it's for the best, if Node.js believes this is not a good pattern.

What do you think? I would go with removing the backwards compatibility mapping and making this a semver-major change.

ljharb:
@giltayar without slash patterns, there is not - full stop.

I'd suggest making this a major change, and thus exposing as *few* things as possible in the process - mocha can always _expand_ the API without a major bump :-)

giltayar:
@ljharb so I'll remove the back compatibility pattern. The rest is definitely exposing _as little_ as possible, but I'll go over it again to see if I missed anything.

juergba:
@giltayar 
The docs of Node v15 states:

> When importing CommonJS modules, the module.exports object is provided as the default export. Named exports may be available, provided by static analysis as a convenience for better ecosystem compatibility.

What is the meaning of "named exports **may be** available"?
Do we still need this PR?

giltayar:
@juergba yes, we still need this PR. Node tries (very bravely, I might add) to infer the exports by static analysis of the source. In the simple cases, it works. Mocha is _definitely_ not a simple case. I know, I tried, just in case.

juergba:
> so I'll remove the back compatibility pattern.

The Node v15 docs still mentions subpath patterns (experimental):

> As a last resort, package encapsulation can be disabled entirely by creating an export for the root of the package "./\*": "./\*". This exposes every file in the package at the cost of disabling the encapsulation and potential tooling benefits this provides.

I haven't understood why you want to take the risk of encapsulating our CJS part. Also there are third party packages using some of our private functions/modules(?), there have been requests in the past making private functions public. But there is no clear strategy about our public API.

giltayar:
@juergba 

> The Node v15 docs still mentions subpath patterns (experimental): ...

I never saw that. @ljharb is this correct? If it is, I can definitely revert the last commit and not have it be a semver major change. I don't have any opinion, here or there.

ljharb:
@giltayar yes, you can use the deprecated slash patterns to make it not be semver-major - i was assuming the preference would be to avoid deprecated features.

juergba:
[subpath_folder_mappings](https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_subpath_folder_mappings) is the one being deprecated.
[subpath_patterns](https://nodejs.org/dist/latest-v15.x/docs/api/packages.html#packages_subpath_patterns) is the experimental one to be used instead.

also available in V14: [subpath_patterns](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html#packages_subpath_patterns)

giltayar:
I finally had the time to look at this. This is a real bug, but not necessarily in this PR. It uncovered a bug in ESM combined with `--parallel`.

On a machine with more than one CPU, run `node bin/mocha test/integration/fixtures/esm/*.fixture.mjs --no-color --no-bail --parallel` (in this branch!). It will run three tests in two fails, as expected.

Now, run the same, but with an additional `--jobs 1`.  You will get a `Error [ERR_REQUIRE_ESM]: Must use import to load ES Module`, because it's not going through the `async` branch of Mocha.

It's something around the connection between parallel and "lazyLoadFiles", but I haven't tracked it yet.

My guess is that fixing this bug will also fix this branch's bug. Looking into it...

giltayar:
The tests here fail due to a real bug: https://github.com/mochajs/mocha/issues/4559

giltayar:
Finally! All tests have passed, and this PR also fixes #4559.

package.json
```diff
@@ -2,6 +2,41 @@
...
+      "./lib/entrypoints/bdd.js"
+    ],
+    "./qunit": [
+      {
+        "require": "./lib/entrypoints/qunit.js",
+        "import": "./lib/entrypoints/qunit.mjs",
+        "default": "./lib/entrypoints/qunit.js"
+      },
+      "./lib/entrypoints/qunit.js"
+    ]
```
ljharb, line 38:
```suggestion
    ],
    "./package.json": "./package.json"
```
so that tools can still locate package.json reliably

ljharb (approved):

juergba (commented):
@giltayar I'm trying to understand this PR, but haven't got it yet.
Why do you insert _package-lock.json_ into this PR?

giltayar:
@juergba From what I see in the commit, I just modified the file that was already there, which I'm assuming my `npm install` did after I changed `package.json` (I only added an `exports` section without touching the dependencies in `package.json`).  Did I do anything wrong here?

juergba:
I just added `main` and `exports` fields to my _package.json_. There are no changes to _package-lock.json_ after `npm install`.

We have had quite a few dependencies updates in recent weeks. It helps to delete your `node_modules` folder periodically before executing `npm install`.

ljharb:
When no dep changes are intended, any lockfile changes that appear should be discarded.

giltayar:
@juergba It's because I used npm v7 (we switched to it at work about a month ago). 

Sp I reverted to the upstream `package-lock.json`, then `npm install`-ed, and pushed.

This made `package-lock.json` move out of this PR.

Nice catch! And sorry for the confusion.

juergba (commented):
@giltayar I haven't digested the complete matter, but my first impressions is, I don't really like it.
The starting point was, that below code is awkward:
```
import mocha  from 'mocha'
const {it, before} = mocha
```
and we want (?) this: `import {it, before}  from 'mocha'`

This PR goes too far and at the same time stops on half way.
Too far:
- IMO big effort/risk for small added value. I know you are an ESM maniac, but ... 😎
- do we really need parity between CJS and ESM, and then - even more complex - backwards compability for CJS?
- the fallback `"./": "./"` has gone
- I haven't understood the proxy part. Does it belong into this PR?

stop on half way:
- we have more interfaces
- what are the consequences for third party interfaces?
- you export `unloadFile` (which makes no sense for EMS)
  - what about the rest of our public functions?
  - what about private functions which are used by other packages (eg. collectFiles (?))

giltayar:
@juergba thanks for the input. Here  are my responses:

> IMO big effort/risk for small added value. I know you are an ESM maniac, but ... 😎

I am! Hopefully, it's a compliment 😉. What I'm seeing in the ecosystem is more and more packages adding ESM wrappers (or, like SIndre Sorhus' packages, migrating all in into ESM). So is it a small value? Last year, maybe. This year, I don't think so. It's slowly becoming a baseline requirement.

> do we really need parity between CJS and ESM, and then - even more complex - backwards compability for CJS?

Not sure I understand what you mean.

> the fallback "./": "./" has gone

That's easy. I can easily put it back. There was input here that maybe we should go all in, and do it as a semver major that abolishes all "illegal" entrypoints into mocha. But if the sentiment is not to do it, I can very easily restore it.

> I haven't understood the proxy part. Does it belong into this PR?

Unfortunately, yes. When using parallel execution, and the number of workers is less than the number of test files, a worker can be reused for two test files. This exposes the fact that the `describe/it/...` that are required/imported are actually tied to the first `Mocha` instance that uses them, and when a new `Mocha` instance is created, they ignore it and use the previous one. To deal with that, I had to proxy all those functions so that they always find the "current" `Mocha` instance.

> we have more interfaces

Ouch. I must have missed those. If you tell me what those are, I can easily add them.

> what are the consequences for third party interfaces?

I have no idea. I didn't know Mocha has this capability. Where can I find documentation (or an example of such an interface) so that I can try and figure out the implications?
> you export unloadFile (which makes no sense for ESM)

Good catch. I'll remove it from the ESM exports.

> what about the rest of our public functions?

I researched carefully, and this is what I found. I may have missed a spot. Which ones did I miss?
what about private functions which are used by other packages (eg. collectFiles (?))

juergba:
refering third party interfaces: see here in our [wiki](https://github.com/mochajs/mocha/wiki/Third-party-UIs)

giltayar:
@juergba Thanks for the pointer. As far as I can tell, the third party ui interface functions will hang the interface functions on `global`. This will continue to work with no problem. This PR didn't change anything around that.

If they wish to have an entry point similar to what Mocha exports via this PR, then we can't do that for them, and they will need to do supply one on their own (outside of Mocha). This is because if we'd wanted to supply one, we'd need to:

1. Understand what interface functions they're supplying. This is currently not supported by the method that the Wiki supplies.
2. Even if we added a way for the third party ui to supply the interface functions, we can't dynamically generate an entry point with the appropriate exports, given that in ESM `export` MUST be static. We _could_ do that for CJS, but I'm not sure it's worth the effort, although I may be wrong and third party UIs are widely used.

BTW, the proxy part of this PR has been superseded by a much cleaner method to do the proxying in  https://github.com/mochajs/mocha/pull/4574 (thanks @nicojs!). Once that PR is merged, I will merge my PR and remove all the proxy code that won't be needed anymore.

PR #4304, opened 2020-05-27
Support --require of ESM; closes #4281

### LINKED ISSUE #4281: Parameter `--require` cannot load an ES module.
### Prerequisites

### Description

Parameter `--require` cannot load an ES module.

### Steps to Reproduce

- download [_testcase.zip_](https://github.com/mochajs/mocha/files/4612189/testcase.zip) which contains 3 files:

> _package.json_:

```json
{
    "name": "testcase",
    "version": "1.0.0",
    "type": "module",
    "scripts": {
        "test": "mocha test.js"
    },
    "dependencies": {
        "mocha": "^7.1.2"
    },
    "mocha": {
        "require": "require.js"
    }
}
```

> _require.js_:

```javascript
// An ES module because { "type": "module" } in package.json.
```

> _test.js_:

```javascript
import assert from "assert";

describe("foo", function () {
    it("bar", function () {
        assert.ok("baz");
    });
});
```

- unzip _testcase.zip_;
- `npm install`
- `npm test`

```
> testcase@1.0.0 test /home/regseb/testcase
> mocha test.js



  foo
    ✓ bar


  1 passing (6ms)
```

```
> testcase@1.0.0 test /home/regseb/testcase
> mocha test.js

mocha inspect [spec..]

Run tests with Mocha

Rules & Behavior
  --allow-uncaught           Allow uncaught errors to propagate        [boolean]
[...]
  --list-reporters   List built-in reporters & exit                    [boolean]

✖ ERROR: Must use import to load ES Module: /home/regseb/testcase/require.js
require() of ES modules is not supported.
require() of /home/regseb/testcase/require.js from /home/regseb/testcase/node_modules/mocha/lib/cli/run-helpers.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename require.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/regseb/testcase/package.json.

```

**Reproduces how often:** [What percentage of the time does it reproduce?] 100%

### Versions

- The output of `mocha --version`: _zsh: command not found: mocha_
- The output of `node node_modules/.bin/mocha --version`: _7.1.2_
- The output of `node --version`: _v13.14.0_
- Your operating system
  - name and version: _Ubuntu 19.10_
  - architecture (32 or 64-bit): _64-bit_
- Your shell (e.g., bash, zsh, PowerShell, cmd): _zsh 5.7.1 (x86_64-ubuntu-linux-gnu)_
- Your browser and version (if running browser tests): N.A.

### Additional Information

Fix proposal: replace [`require`](https://github.com/mochajs/mocha/blob/master/lib/cli/run-helpers.js#L89) by [`import`](https://nodejs.org/api/esm.html#esm_import_expressions). But _dynamic import_ has been added to [Node _13.2.0_](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Browser_compatibility).

JacobLey:
Allow files/modules specified in `--require` to be ESM.

CommonJS loading is still supported and the default.

closes https://github.com/mochajs/mocha/issues/4281

### Description of the Change

Support `--require`ing ESM files in config. Current behavior is to fail on `ERR_REQUIRE_ESM`.

Re-use existing "require or import" logic for loading test files to load required files.

### Alternate Designs

Pipe requiring the ESM file through something like [esm](https://www.npmjs.com/package/esm) or [ts-node](https://www.npmjs.com/package/ts-node). Although that seems much more complicated and overkill.

Requiring JS files before the test suite is a core part of Mocha's functionality. This extends that support to ESM-first codebases, which is natively supported via `"type":"module"`, or `--exerimental-module`.

### Benefits

Allows developers more freedom in choosing to write ESM or CommonJS.

### Possible Drawbacks

Native ESM in NodeJS is still labeled as experimental (although an `--experimental` flag is no longer needed). It is possible the usage will change, but that is justification for re-using existing "require or import" functionality.

### Applicable issues

This should be an "enhancement" (minor release). `--require`ing CommonJS is still the default and fully supported, this will just allow additional support for ESM.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1007,6 +1007,8 @@ Modules required in this manner are expected to do work synchronously; Mocha won
 
 Note you cannot use `--require` to set a global `beforeEach()` hook, for example &mdash; use `--file` instead, which allows you to specify an explicit order in which test files are loaded.
 
+> As of v7.3.0, Mocha supports `--require` for [NodeJS native ESM](#nodejs-native-esm-support). There is no separate `--import` flag.
+
 ### `--sort, -S`
 
 Sort test files (by absolute path) using [Array.prototype.sort][mdn-array-sort].
--- a/docs/index.md
+++ b/docs/index.md
@@ -1450,7 +1450,6 @@ Node.JS native ESM support still has status: **Stability: 1 - Experimental**
 - [Watch mode](#-watch-w) does not support ES Module test files
 - [Custom reporters](#third-party-reporters) and [custom interfaces](#interfaces)
   can only be CommonJS files
-- [Required modules](#-require-module-r-module) can only be CommonJS files
 - [Configuration file](#configuring-mocha-nodejs) can only be a CommonJS file (`.mocharc.js` or `.mocharc.cjs`)
 - When using module-level mocks via libs like `proxyquire`, `rewiremock` or `rewire`, hold off on using ES modules for your test files
 - Node.JS native ESM support does not work with [esm][npm-esm] module
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -92,7 +92,10 @@ exports.handleRequires = async (requires = []) => {
       debug('resolved required file %s to %s', mod, modpath);
     }
     const requiredModule = await requireOrImport(modpath);
-    if (type(requiredModule) === 'object' && requiredModule.mochaHooks) {
+    if (
+      ['object', 'module'].includes(type(requiredModule)) &&
+      requiredModule.mochaHooks
+    ) {
       const mochaHooksType = type(requiredModule.mochaHooks);
       if (/function$/.test(mochaHooksType) || mochaHooksType === 'object') {
         debug('found root hooks in required file %s', mod);
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -15,6 +15,7 @@ const collectFiles = require('./collect-files');
 const {type} = require('../utils');
 const {format} = require('util');
 const {createInvalidPluginError, createUnsupportedError} = require('../errors');
+const {requireOrImport} = require('../esm-utils');
 
 /**
  * Exits Mocha when tests + code under test has finished execution (default)
@@ -81,15 +82,16 @@ exports.list = str =>
  * @returns {Promise<MochaRootHookObject|MochaRootHookFunction>} Any root hooks
  * @private
  */
-exports.handleRequires = async (requires = []) =>
-  requires.reduce((acc, mod) => {
+exports.handleRequires = async (requires = []) => {
+  const acc = [];
+  for (const mod of requires) {
     let modpath = mod;
     // this is relative to cwd
     if (fs.existsSync(mod) || fs.existsSync(`${mod}.js`)) {
       modpath = path.resolve(mod);
       debug('resolved required file %s to %s', mod, modpath);
     }
-    const requiredModule = require(modpath);
+    const requiredModule = await requireOrImport(modpath);
     if (type(requiredModule) === 'object' && requiredModule.mochaHooks) {
       const mochaHooksType = type(requiredModule.mochaHooks);
       if (/function$/.test(mochaHooksType) || mochaHooksType === 'object') {
@@ -102,8 +104,9 @@ exports.handleRequires = async (requires = []) =>
       }
     }
     debug('loaded required module "%s"', mod);
-    return acc;
-  }, []);
+  }
+  return acc;
+};
 
 /**
  * Loads root hooks as exported via `mochaHooks` from required files.
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -93,7 +93,8 @@ exports.handleRequires = async (requires = []) => {
     }
     const requiredModule = await requireOrImport(modpath);
     if (
-      ['object', 'module'].includes(type(requiredModule)) &&
+      requiredModule &&
+      typeof requiredModule === 'object' &&
       requiredModule.mochaHooks
     ) {
       const mochaHooksType = type(requiredModule.mochaHooks);
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -1,8 +1,16 @@
 const path = require('path');
+const url = require('url');
+
+const formattedImport = async file => {
+  if (path.isAbsolute(file)) {
+    return import(url.pathToFileURL(file));
+  }
+  return import(file);
+};
 
 exports.requireOrImport = async file => {
   if (path.extname(file) === '.mjs') {
-    return import(file);
+    return formattedImport(file);
   }
   // This is currently the only known way of figuring out whether a file is CJS or ESM.
   // If Node.js or the community establish a better procedure for that, we can fix this code.
@@ -12,7 +20,7 @@ exports.requireOrImport = async file => {
     return require(file);
   } catch (err) {
     if (err.code === 'ERR_REQUIRE_ESM') {
-      return import(file);
+      return formattedImport(file);
     } else {
       throw err;
     }
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -1,11 +1,8 @@
-const url = require('url');
 const path = require('path');
 
-const requireOrImport = async file => {
-  file = path.resolve(file);
-
+exports.requireOrImport = async file => {
   if (path.extname(file) === '.mjs') {
-    return import(url.pathToFileURL(file));
+    return import(file);
   }
   // This is currently the only known way of figuring out whether a file is CJS or ESM.
   // If Node.js or the community establish a better procedure for that, we can fix this code.
@@ -15,7 +12,7 @@ const requireOrImport = async file => {
     return require(file);
   } catch (err) {
     if (err.code === 'ERR_REQUIRE_ESM') {
-      return import(url.pathToFileURL(file));
+      return import(file);
     } else {
       throw err;
     }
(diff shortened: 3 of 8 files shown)


JacobLey:
No tests included in this first pass.

I don't see any existing tests for this... so not exactly sure where to put them.

lib/esm-utils.js
```diff
@@ -15,7 +12,7 @@ const requireOrImport = async file => {
     return require(file);
   } catch (err) {
     if (err.code === 'ERR_REQUIRE_ESM') {
-      return import(url.pathToFileURL(file));
```
JacobLey, line 18: Not sure what `url.pathToFileURL` was achieving?
`import()` allows a plain string, and this way the required file can still be resolvable via node's logic (e.g. loading from `node_modules`)

lib/esm-utils.js
```diff
@@ -1,11 +1,8 @@
-const url = require('url');
 const path = require('path');
 
-const requireOrImport = async file => {
-  file = path.resolve(file);
```
JacobLey, line 5: Moved `resolve()` to `loadFilesAsync` so the file doesn't have to be a fully-resolved name, but can be a package from node_modules as well.

lib/esm-utils.js
JacobLey: I guess it was required for Windows.
Gosh I hate windows

boneskull:
apologies, the organization of `test` is often not obvious.  probably need to add more info to `test/README.md`

test/integration/options/require.spec.js
```diff
@@ -45,6 +46,51 @@ describe('--require', function() {
...
+                  // mixed with commonjs
+                  '../fixtures/options/require/root-hook-defs-a.fixture.js'
+                ),
+              require.resolve(
+                '../fixtures/options/require/root-hook-test.fixture.js'
+              )
+            ],
+            {
+              env: {
+                ...process.env,
```
JacobLey, line 80: This is what I updated the parser options (parserOptions: 2018) in eslintrc for.

Considering mocha is documented with:
> As of v7.0.0, Mocha requires Node.js v8.0.0 or newer.
This seems safe enough for tests at least.

If there is an easier way to pass `--experimental-modules` to the mocha (instead of calling node directly) happy to use that instead...

boneskull: I apologize for the state of things (we're working on it; see #4293), but...

I wanted to get a subset of the integration tests (`test/integration/`) running in a browser, but haven't completed this work.  Stuff in `test/node-unit/` (and a few other places) is allowed to use ES2018 syntax, but literally everything _else_ under `test/` is limited to ES5.  It's probably OK to allow ES2018 in `test/integration/`, if we expect #4293 to get merged in the near future, since we'll be able to transpile everything down to where it needs to be.  I don't envision browser-based integration tests landing before then, so _IMO_ it's fine.  Others may disagree (@mochajs/core ?)

However, stuff in `test/unit/` and `test/browser` _cannot_, at this time, be ES2018, because these tests run in IE11.  So, please revert the change to `.eslintrc.yml`, then add `'test/integration/**/*.js'` to the list of overrides which already use `parserOptions: 2018` (L39 or so).

test/integration/fixtures/options/require/esm/root-hook-defs-esm.fixture.js
```diff
@@ -0,0 +1,8 @@
+export const mochaHooks = () => ({
```
boneskull, line 1: awesome

lib/cli/run-helpers.js
```diff
@@ -92,7 +92,10 @@ exports.handleRequires = async (requires = []) => {
       debug('resolved required file %s to %s', mod, modpath);
     }
     const requiredModule = await requireOrImport(modpath);
-    if (type(requiredModule) === 'object' && requiredModule.mochaHooks) {
+    if (
+      ['object', 'module'].includes(type(requiredModule)) &&
```
JacobLey, line 96: ```node -e "import('./test/integration/fixtures/options/require/root-hook-defs-esm.fixture.mjs').then(mod => console.log(Object.prototype.toString.call(mod)))"```
-> `[object Module]`

Piped through the `type` function: https://github.com/mochajs/mocha/blob/master/lib/utils.js#L216
Returns `module`.

So added handling for that "module" case (which seems reasonable considering that is what we would expect from ESM... but for the sake of discussing alternatives, could add some handling to `type()` to map module -> object

test/integration/options/require.spec.js
boneskull: Here, you're running `bin/mocha` (which is what `invokeMocha` does), but using the `NODE_OPTIONS` environment variable to set `--experimental-modules`. You don't need to do this; just pass `--experimental-modules` in your array of args.  `bin/mocha` knows what to do with it.

e.g.

```js
invokeMocha(['path/to/fixture', '--some-other-flag'].concat(+process.versions.node.split('.')[0] >= 13 ? [] : ['--experimental-modules'], function() {})
```

I realize this means you can just revert the change to `.eslintrc.yml` entirely, which is maybe what should happen.

lib/cli/run-helpers.js
boneskull: no, this seems correct.  but, we should also add a test in `test/node-unit/utils.spec.js` to cover the behavior.  you don't need to do this unless you want to.

boneskull: we can change it to return `object` if there are unforeseen consequences.  what does `typeof requiredModule` return?

JacobLey: ```node -e "import('./test/integration/fixtures/options/require/root-hook-defs-esm.fixture.mjs').then(mod => console.log(typeof mod))"``` -> `object`

I'll let you interpret that... since there must be _some_ reason not to use `typeof` directly.

> add a test to cover the behavior
I can add that as well, so long as we are fine with Modules returning "module"

test/integration/options/require.spec.js
JacobLey: Reverted change to eslintrc

Didn't realize I could pass the flag directly though mocha! Since I don't need any lint changes, I don't think this is the place to reorganize what is allowed where.

lib/cli/run-helpers.js
boneskull: there's no real great reason to use `type()` here other than trying to maintain some consistency. we should use `typeof` and not worry about it.

boneskull: Ref: #4306

lib/esm-utils.js
```diff
@@ -25,7 +30,7 @@ const requireOrImport = async file => {
 exports.loadFilesAsync = async (files, preLoadFunc, postLoadFunc) => {
   for (const file of files) {
     preLoadFunc(file);
-    const result = await requireOrImport(file);
+    const result = await exports.requireOrImport(path.resolve(file));
```
giltayar, line 33: If the file in question is a package name, `path.resolve` will fail (AFAIK with an exception). Why did you add a `path.resolve` here?

giltayar: Oh, I see. `--require`-d files can be module names, and so you didn't want `requireOrImport`, which is now used also in `handleRequire`, to resolve them to a path. So you moved it to `loadFileAsync`, where it _is_ needed (for Windows, it seems...)

giltayar:
Because "require" is not typically a word used with ESM ("import" is more generally used), I would add specific messaging in the `--require` documentation that says that it also accepts ES modules.

test/integration/options/require.spec.js
```diff
@@ -45,6 +46,46 @@ describe('--require', function() {
...
+      before(function() {
+        if (!utils.supportsEsModules()) this.skip();
+      });
+
+      it('should run root hooks when provided via mochaHooks', function() {
+        return expect(
+          invokeMochaAsync(
+            [
+              '--require=' +
+                require.resolve(
```
juergba, line 60: `--require`works with absolute path and module names. Does it work also with relative paths?

lib/cli/run-helpers.js
```diff
@@ -81,16 +82,21 @@ exports.list = str =>
...
  */
-exports.handleRequires = async (requires = []) =>
-  requires.reduce((acc, mod) => {
+exports.handleRequires = async (requires = []) => {
+  const acc = [];
+  for (const mod of requires) {
     let modpath = mod;
     // this is relative to cwd
     if (fs.existsSync(mod) || fs.existsSync(`${mod}.js`)) {
       modpath = path.resolve(mod);
```
juergba, line 91: > If the file in question is a package name, path.resolve will fail [...]

Will this work for packages?

JacobLey: It should, `fs.existsSync` will return false for a package name, so it is not resolved at all (passed raw) to requireOrImport (e.g. `requireOrImport('@babel/register')`

This behavior is unchanged in this PR

test/integration/options/require.spec.js
JacobLey: Yes, see where you commented on `handleRequires`, relative paths are transformed into absolute paths.

None of the tests use it though because the actual CWD isn't very clear in-test. But maybe it is possible to calculate a relative path in test and pass that through as well.

Again, that particular functionality is unchanged in this PR

docs/index.md
```diff
@@ -1007,6 +1007,8 @@ Modules required in this manner are expected to do work synchronously; Mocha won
 
 Note you cannot use `--require` to set a global `beforeEach()` hook, for example &mdash; use `--file` instead, which allows you to specify an explicit order in which test files are loaded.
 
+> As of v7.3.0, Mocha supports `--require` for [NodeJS native ESM](#nodejs-native-esm-support). There is no separate `--import` flag.
```
boneskull, line 1010: note that the version here may need changing

boneskull:
LGTM.  Thanks much @JacobLey !

PR #4314, opened 2020-06-05
remove .nvmrc; use NODE_VERSION env for netlify

boneskull:
This PR removes `.nvmrc` and uses an environment variable to set the Node.js version when deploying via Netlify.

My reasoning is that `.nvmrc` is consumed by more than just Netlify.  My shell, for instance, will automatically switch to a version specified in `.nvmrc` when I navigate to my working copy.  This may also have other unintended consequences.

So--if this works--I think it's best to not let the Node.js v12 requirement when deploying "leak" out of this context.


boneskull:
It seems to work.

PR #4315, opened 2020-06-06
Fix glob pattern support: no comma splitting for 'spec' and 'ignore' options

### LINKED ISSUE #4307: [BUG] Migrating from mocha.opts to mocharc broke glob pattern support
### Prerequisites

<details>
  <summary>Click for details</summary>

  - [X] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend that you _not_ install Mocha globally.
</details>

---

### Description

- Using `opts configuration`, the [glob pattern](https://globster.xyz/) `{controllers,test}/**/*.test.js` used to expand into:
  + `controllers/**/*.test.js`
  + `test/**/*.test.js`

- Using RC configration files, this glob partten gets tokenized into invalid ones:

```
[mocha/lib/cli/options.js] > parse() > yargsParser.detailed() > result.argv.spec [ '{controllers', 'test}/**/*.test.js' ]

Warning: Cannot find any files matching pattern "{controllers"
Warning: Cannot find any files matching pattern "test}/**/*.test.js"
```

- Commenting the [**yargs** parsing option `coerce: coerceOpts`](https://github.com/mochajs/mocha/blob/master/lib/cli/options.js#L124) fixes the glob tokenization, even though it isn't a long-term fix.
- The `coerceOtps` function can be found here: https://github.com/mochajs/mocha/blob/master/lib/cli/options.js#L63

As a side I should add that these two following `spec` values give the results highlighted above
- `spec: '{controllers,test}/**/*.test.js',`
- `spec: [ '{controllers,test}/**/*.test.js' ],`

---

### Steps to Reproduce

I created a barebone repository to highlight the issue: https://github.com/TheOptimisticFactory/mocha-glob-issue

1. Open a terminal
2. Clone the demo project using `git clone git@github.com:TheOptimisticFactory/mocha-glob-issue.git`
2. Install packages using `cd mocha-glob-issue && npm i`
4. Run any of the following scripts:

-  `npm run test-legacy-working`: Check tests passes when using lecacy opts `test/mocha.opts`
- `npm run test-bugged-baseline`: tests WONT BE FOUND when using `.mocharc.js`
- `npm run test-bugged-baseline`: Dumps the BROKEN file patterns when using `.mocha.multi-paths.js`

---

**LEGACY behavior:** [What used to happen]

<details>
  <summary>LEGACY configuration file: /test/mocha.opts</summary>

  ```javascript
  --require test/setup.js
  {controllers,test}/**/*.test.js
  --exit
  ```
</details>

`npm run test-legacy-working`

image

---

<details>
  <summary>Using .mocharc.js</summary>

  ```javascript
  'use strict';

  module.exports = {
    exit: true,
    require: 'test/setup.js',
    spec: '{controllers,test}/**/*.test.js',
  };
  ```
</details>

`npm run test-bugged-baseline`

image

<details>
  <summary>Using .mocharc.multi-paths.js</summary>

  ```javascript
  'use strict';

  module.exports = {
    exit: true,
    require: 'test/setup.js',
    spec: [ '{controllers,test}/**/*.test.js', 'test/**/*.test.js' ],
  };
  ```
</details>

`npm run test-bugged-showcase`

image

---

About 100% of the time :)

---

### Versions

image

juergba:
### Description

Currently we can pass test files via CLI:
- `node mocha test1 test2`
- `node mocha -- test1 test2`
- `node mocha --spec test1 --spec test2`
- `node mocha --spec test1,test2`

via configuration files:
- `spec: ['test1', 'test2']`
- `spec: ['test1,test2']`

The last option (comma-delimited list) is neither documented nor provided in our _Yargs_ configuration. But it does break Mocha's glob pattern support in certain cases, e.g.
`{controllers,test}/**/*.test.js` ==> `['{controllers', 'test}/**/*.test.js']`

### Description of the Change

For options `spec` and `ignore`:
- _yars-parser:_ we remove splitting by comma
- _yargs:_ no changes necessary

### Applicable issues

closes #4307

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -54,16 +54,19 @@ const configuration = Object.assign({}, YARGS_PARSER_CONFIG, {
 
 /**
  * This is a really fancy way to:
- * - ensure unique values for `array`-type options
- * - use its array's last element for `boolean`/`number`/`string`- options given multiple times
+ * - `array`-type options: ensure unique values and evtl. split comma-delimited lists
+ * - `boolean`/`number`/`string`- options: use last element when given multiple times
  * This is passed as the `coerce` option to `yargs-parser`
  * @private
  * @ignore
  */
+const globOptions = ['spec', 'ignore'];
 const coerceOpts = Object.assign(
   types.array.reduce(
     (acc, arg) =>
-      Object.assign(acc, {[arg]: v => Array.from(new Set(list(v)))}),
+      Object.assign(acc, {
+        [arg]: v => Array.from(new Set(globOptions.includes(arg) ? v : list(v)))
+      }),
     {}
   ),
   types.boolean
--- a/test/node-unit/cli/options.spec.js
+++ b/test/node-unit/cli/options.spec.js
@@ -562,7 +562,9 @@ describe('options', function() {
           readFileSync = sandbox.stub();
           readFileSync.onFirstCall().throws();
           findConfig = sandbox.stub().returns('/some/.mocharc.json');
-          loadConfig = sandbox.stub().returns({spec: '*.spec.js'});
+          loadConfig = sandbox
+            .stub()
+            .returns({spec: '{dirA,dirB}/**/*.spec.js'});
           findupSync = sandbox.stub();
           loadOptions = proxyLoadOptions({
             readFileSync,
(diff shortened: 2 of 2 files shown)


lib/cli/options.js
```diff
@@ -54,16 +54,19 @@ const configuration = Object.assign({}, YARGS_PARSER_CONFIG, {
...
  * This is a really fancy way to:
- * - ensure unique values for `array`-type options
- * - use its array's last element for `boolean`/`number`/`string`- options given multiple times
+ * - `array`-type options: ensure unique values and evtl. split comma-delimited lists
+ * - `boolean`/`number`/`string`- options: use last element when given multiple times
  * This is passed as the `coerce` option to `yargs-parser`
  * @private
  * @ignore
  */
+const globOptions = ['spec', 'ignore'];
```
craigtaub, line 63: Just checking, is `exclude` option alias'd into `ignore`?

craigtaub:
So as I understand `spec: "test/one/index.spec.js,test/two/index.spec.js"`  would work in master but not on the fix (tried locally to confirm).
In which case should this be a breaking change? Regardless of not documenting the feature if we make a fix which breaks a certain setup, that sounds like a major.

lib/cli/options.js
juergba: Yes, but vice versa. `ignore` is the canonical and `exclude` the alias option name. _Yargs-parser_ applies the coerce function automatically to both names.

juergba:
> So as I understand comma-delimited spec: ["test/one/index.spec.js,test/two/index.spec.js"] would work in master but not on the fix  [...]

Yes, that's correct. I'm fine with `major`.

craigtaub:
Fix looks but, but im slightly concerned we have no docs on this option and its usage, can we add some to add clarity? Introducing a breaking change, be useful to be able to point people to an explanation. Thoughts?

juergba:
There is another docs [section](https://mochajs.org/#configuration-format).
It's somehow confusing to have a positional argument `spec` and an option `--spec`. There is no need to use `--spec` in the CLI, actually it's only needed for the configuration files.
I'm not sure how to document that, I will think about.

juergba:
Yes, this is correct  for `spec` and `ignore`:
- CLI: `node mocha --spec foo.js,bar.js`
- configuration file: `spec: ['foo.js,bar.js']`

The following has never worked and remains unchanged:
- CLI: `node mocha foo.js,bar.js`

PR #4318, opened 2020-06-08
Switch opencollective images to self-hosted spritesheet

Munter:
After we added all our opencollective sponsors and backers to the documentation, the load time and the total download size of the documentation increased significantly. It can easily take several seconds to finish loading the opencollective images, despite them being on a CDN served on http2.

This network graph for our current situation looks like this:

1_waterfall_thumb

There is no way for a user to out out of downloading these 256 images at about 700kb in total.

- We should increase the speed of download, since these images are literally the first thing people see when visiting the documentation
- We should reduce the amount of bytes downloaded

### Description of the Change

The opencollective fetcher script now downloads all sponsor logos and avatar images in order to self-host. A sprite sheet post process is added in order to reduce the amount of requests and further lossy compression is applied.

- Fetch full images and store them. Previously we fetched the first chunk only to figure out domensions
- Change supporters template to use self-hosted images
- Add a spritesheet definition to make assetgraph-builder sprite the self-hosted images
- Apply lossy compression to the sprite sheet: pngquant 256 colors. Reduces the download size by approximately 8x
- Remove avatar.js, since sprite sheets are background images that don't trigger onload events
- Filtered out personal backers that specifically have asked to be anonymous

**Improvements:**
- Reduced download size by about 500kb
- Faster time to final render with backer images
- Less CPU time usage (due to removal of JS to image loaded animation)

Here's a before and after comparison: https://www.webpagetest.org/video/compare.php?tests=200607_Z2_58f219880cacb17ae211d74b8e670c1c,200608_DX_cd846fd4961c94ad247dab95eaac76d8-r%3A1-c%3A0&thumbSize=200&ival=500&end=full

image

image

image

image

### Alternate Designs

We could self host the individual images, but speed measurements showed less improvement than a sprite sheet.

### Benefits

Faster time to final rendering of the top part of the documentation. Significantly less downloaded bytes

### Possible Drawbacks

The backer image loading animation is replaced by a slightly less appealing raw image loading progress. On the other hand it's much faster.


.gitignore
```diff
@@ -141,3 +142,6 @@ Temporary Items
 
 # SauceConnect
 *.sock
+
+# Local Netlify folder
+.netlify
```
boneskull, line 147: needs newline

docs/_includes/supporters.md
```diff
@@ -2,13 +2,36 @@
...
+  }
+  .backer {
+    -sprite-selector-for-group: backers;
+    -sprite-location: url(/images/sprite-backers.png?pngquant);
+    -sprite-image-format: png;
+    width: 32px;
+    height: 32px;
+  }
+
+</style>
```
boneskull, line 20: does this want to be extracted into `style.css`?

Munter: Yeah. I think I'll do that. That should also remove an artifact where the page will jump on reload when you're scolled down a bit. Currently there's a render flush when the browser hits this style tag because it becomes external. That makes the page scrollheight briefly be equal to the offset of this tag, which makes the page jump up to the top, then immediately down again once the async css for the sprites has loaded

PR #4319, opened 2020-06-08
disable parallel mode by default

boneskull:
Having run benchmarks against our unit tests and integration tests, I've found that the unit tests do not benefit from parallel mode--and in fact, they run more slowly.  Integration tests still see a 50% performance boost, so we will use parallel tests for those only.


ORESoftware:
duh

PR #4321, opened 2020-06-09
trigger netlify deploy nightly

boneskull:
This sets up a job to run at 00:00 UTC every day, which rebuilds the site from the `mochajs.org` branch.  This keeps the list of supporters up-to-date.  It relies on a secret env var to do its thing.

Will try to figure out how to test this...

### DIFF
--- a/.github/workflows/nightly-site-deploy.yml
+++ b/.github/workflows/nightly-site-deploy.yml
@@ -12,6 +12,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Webhook Action
-        uses: joelwmale/webhook-action@1
+        uses: joelwmale/webhook-action@1.0.0
         env:
           WEBHOOK_URL: ${{ secrets.NETLIFY_NIGHTLY_DEPLOY_URL }}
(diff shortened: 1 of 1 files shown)


boneskull:
Yeah, I don't know how to test this.  Maybe we just merge it and see if it does anything.

boneskull:
The reason is because we (I?) tend to get a fair amount of email from people wondering why their logo isn't on the site.  Since we're only fetching the logos and links at build time now, we need a nightly job to keep it up-to-date.

nicojs:
> Yeah, I don't know how to test this. Maybe we just merge it and see if it does anything.

Why not configure it to run on a `push` action once. When you see it works, you can remove the push action.

boneskull:
I want to merge this, but given that all deploys seem to be failing atm, that needs fixing first

boneskull:
hmm, maybe it's fixed now.

boneskull:
I think `mochajs.org` needs fast-forwarding.

boneskull:
OK, I see why it's breaking.  See https://github.com/mochajs/mocha/pull/4339

The svg's on OC's side are broken

PR #4147, opened 2020-01-09
Fix leaking global 'uncaughtException' handler

juergba:
### Description

```js
const mocha = require("mocha");
for (let i = 0; i < 15; i++) {
    const r = new mocha.Runner(new mocha.Suite(""+i, undefined));
    r.run();
}
```
When using Mocha programmatically with multiple runners, a `MaxListenersExceededWarning` is emitted by Node. For every runner instance an `uncaughtException` handler is registered, but never removed.
This is a regression.

### Description of the Change

Remove eventual `uncaughtException` handler before starting a new runner instance.

### Applicable issues

closes #4144

Issue #4144, opened 2020-01-07
Mocha 7: A global `uncaughtException` handle is leaked for every completed test runner


weswigham:
### Description
A global `uncaughtException` handle is leaked for every completed test runner, in turn causing a `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit` warning to be thrown by `node`.

Since every handler attached is identical in content [here](https://github.com/mochajs/mocha/pull/4030/files#r363540897), attaching _one_ for the first completed runner should be just as correct (although, functionality-wise, it may need to be detcahced once a new runner has started - it's unclear how the sequencing between multiple runners is intended to work out for this handler).

### Steps to Reproduce

```js
const mocha = require("mocha");
for (let i = 0; i < 15; i++) {
    const r = new mocha.Runner(new mocha.Suite(""+i, undefined));
    r.run();
}
```

**Expected behavior:**
No warnings from node.

**Actual behavior:**

A `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit` warning is printed.

**Reproduces how often:**
Every time. You can paste that example code into a [repl](https://runkit.com/weswigham/5e13d0b85c4aff001b6c768f).

juergba:
@weswigham thanks for reporting and testing!

PR #4329, opened 2020-06-10
Move removal of `uncaughtException` handler into `end` callback

weswigham:
And only attach it initially within `start`. Since `self.uncaught` is a bound function, "removing" it immediately before adding it only prevents the same instance from leaking multiple handles - it still allows multiple runners to leak handles forever, even when those runners runs have ended (resulting in an issue akin to #4144).

### Description of the Change

Essentially, it _seems like_ mocha 8 regressed #4144, but the test added in #4147 didn't seem to catch it, I'm guessing because now the process of of attaching the handlers during `.run` was deferred in some fashion by another change.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -989,6 +989,7 @@ Runner.prototype.run = function(fn, opts) {
   fn = fn || function() {};
 
   function start() {
+    self._addEventListener(process, 'uncaughtException', self.uncaught);
     debug('run(): starting');
     // If there is an `only` filter
     if (rootSuite.hasOnly()) {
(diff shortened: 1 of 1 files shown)


lib/runner.js
```diff
@@ -1027,11 +1028,9 @@ Runner.prototype.run = function(fn, opts) {
     debug(constants.EVENT_RUN_END);
     debug('run(): emitted %s', constants.EVENT_RUN_END);
     fn(self.failures);
+    self._removeEventListener(process, 'uncaughtException', self.uncaught);
```
juergba, line 1031: Removing the uncaught exception handler here is incorrect, too early, IMO. There can be exceptions bubbling up after runner's end. We even have a check for this (`STATE_STOPPED`) in `Runner#_uncaught`.

weswigham: If a runner `'end'`s, then results are likely to have been reported based on that runner. If a test could error post-end, it's too late to include it in results... Simultaneously, if a test still has outstanding actions like that, the runner probably shouldn't be `'end'`ing!

boneskull:
Sounds like @weswigham won't have time to work on this, so I can pick it up unless someone beats me to it.

boneskull:
@weswigham I think v8.2.1 corrected a similar problem... I'm going to give it a shot against typescript.

boneskull:
ref: microsoft/typescript#41403 microsoft/typescript#41404

PR #4334, opened 2020-06-13
Fix wrong sponsors images from OC

outsideris:
After merging #4318 , building our docs are failed with `Error: Unsupported image type`. See [here](https://app.netlify.com/sites/mocha/deploys/5ee47f8ed28ad60008aa8f19) or [here](https://app.netlify.com/sites/mocha/deploys/5ee3de46550b6a0008874fc1)

One of our backers(id is `7mywxoz3-409rl6ka-anpvenbd-j7gk85ag`)'s avatar URL is <https://images.opencollective.com/andréleifeld/97285d0/avatar/32.png>. Because `é` character in path, cloudflare return blocked HTML page instead avatar image.

스크린샷 2020-06-13 20 26 53

This HTML page is `Sorry, you have been blocked. You are unable to access opencollective.com`. So, the image has `png` extension but it is not an image. 

I'm not sure whether it is [needle](https://www.npmjs.com/package/needle)'s problem or cloudflare's problem.

But, we should encode avatar URL before request it.

### DIFF
--- a/docs/_data/supporters.js
+++ b/docs/_data/supporters.js
@@ -128,7 +128,7 @@ module.exports = async () => {
   await Promise.all(
     supporters.sponsors.map(async sponsor => {
       const filePath = resolve(supporterImagePath, sponsor.id + '.png');
-      const {body} = await needle('get', sponsor.avatar);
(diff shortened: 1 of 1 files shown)


Munter:
Great find!

PR #4336, opened 2020-06-16
fix(vulnerabilty): switch tiny dependency and use same code

martinoppitz:
### Requirements

See https://github.com/mochajs/mocha/issues/4332

The problem is the maintainment of two dependencies.
Main impact comes from `lodash`.
Second impact is the missing update from `yargs-unparser`.

### Description of the Change

The current master code of `yargs-unparser` is okay. So I forked it an publish this code as npm packge named `@care-for/yargs-unparser`.

An replace `yargs-unparser` with `@care-for/yargs-unparser` in the `package.json` of `mocha`.

### Benefits

Zero vulnerabilities

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The obligatory patch after a major.
 
 ## :bug: Fixes
 
+- [#4332](https://github.com/mochajs/mocha/issues/4332): Replace `yargs-unparser` with `@care-for/yargs-unparser` to fix lodash vulnarability ([**@martinoppitz**](https://github.com/martinoppitz))
+- [#4331](https://github.com/mochajs/mocha/issues/4331): Update `serialize-javascript` to 3.1.0 ([**@martinoppitz**](https://github.com/martinoppitz))
 - [#4328](https://github.com/mochajs/mocha/issues/4328): Fix `--parallel` when combined with `--watch` ([**@boneskull**](https://github.com/boneskull))
 
 # 8.0.0 / 2020-06-10
--- a/bin/mocha
+++ b/bin/mocha
@@ -17,7 +17,7 @@ const {
   isNodeFlag,
   impliesNoTimeouts
 } = require('../lib/cli/node-flags');
-const unparse = require('yargs-unparser');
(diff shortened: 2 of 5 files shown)


martinoppitz:
Hello, I dont know, whats going wrong. 

-- netlify/mocha/deploy-preview

martinoppitz:
@juergba 

Yes, this solution is not the best.

But how we will solve the lodash blocker?!

Munter:
It's not a blocker. The problem is not in mocha but in your local tooling. Don't block on non-curated false positive security audits

martinoppitz:
@Munter 

Sorry, thats not right. See here https://snyk.io/test/npm/mocha?tab=dependencies.

The problem is to fix the lodash vulnerability, because lodash is not maintained.

So I tried to fix yargs-unparser ... I hope anybody publish a new version.

Munter:
@martinoppitz You did obviously not read @boneskull's reponse at https://github.com/mochajs/mocha/issues/4332#issuecomment-644314512 properly.

There is a security vulnerability in lodash, but yargs-unparser is not using the vulnerable function. This means the vulnerability does actually not affect yargs-unparser and mocha.

Now, if you want to talk about security, asking us to switch our yargs-unparser dependency to a random fork that could very easily be compromised in other ways. Which does not have a history of active maintenance or a record of keeping up to date... That's a hard no.

This is a case of automated security checks causing false positives and resulting in generating more work for open source maintainers. You have now had 3 maintainers spend their time, not only on reviewing the security audit and follow the code path to conclude there's nothing we can do, but also on having to explain the same thing to you multiple times.

The problem that is blocking you is **your** CI configuration. Configure your CI differently.

I'm going to close these PR's

martinoppitz:
Hi @Munter ,

I have read the description from @boneskull - read for yourself: https://github.com/mochajs/mocha/issues/4332#issuecomment-644610177

And I would not discuss about security. I will say, that mocha depends on:

- yargs-unparser@1.6.0 and this published package depends on
- lodash@4.7.15

and **both probably not maintained**.

image

image

*I tried my best to explain the vulnerability problem.*

PR #4360, opened 2020-07-01
implementation of global setup/teardown; closes #4308

### LINKED ISSUE #4308: global setup / teardown hooks
With the addition of [root hook plugins](#4244) and [parallel mode](#4245), an important use case may have been omitted.  

In serial mode, a root _before all_ (`before()`) hook can perform async setup (e.g., starting a test server) and a root _after all_ (`after()`) can perform async teardown (closing the server).  This is not perfect, given Mocha could crash and the server would not get cleaned up, but I think it's behavior people expect.

In parallel mode, this behavior not available.  Root hooks "before all" and "after all" run _once per file_.  If you want to run something once and only once, you're kind of stuck.

One could abuse `mochaHooks` by defining a `Promise`-returning function, and doing setup there...

```js
exports.mochaHooks = async () => {
  // do something once only
});
```

... but there's no associated "teardown".

The only real workaround that I'm aware of is via script, e.g., `{"scripts": {"test": "startServer && mocha; stopServer"}}`.  Top-level `await` suffers from the same problem as the above example--no associated teardown.

An idea for this would be to implement two new "hooks" in `mochaHooks`, e.g., `begin` and `end`.

- These would be aggregated before `Mocha` begins running tests, similar to the behavior of the other hooks
- The hooks are guaranteed (within reason) to run _once and only once_.
- These hooks can share a context _with each other_, but nothing else.  You cannot access the current test context from `begin`, for instance.  You _could_ define `this.foo` in `begin` and access it in `end`, I suppose.
- Standard Node.js-style callbacks and `Promise`-returning functions allowed
- They will always run in the main process

Alternatives:

- Instead of using `mochaHooks`, use a different property.
- Instead of having a single property, use `mochaBegin` and `mochaEnd` properties.
- Name it something else, but be careful not to conflate it with existing hooks or interfaces (`setup` is going to be very confusing for those using the `tdd` interface, for example).

cc @nicojs, who may have opinions about how this should work.

boneskull:
This PR adds `mochaGlobalSetup` and `mochaGlobalTeardown` fixtures, loaded like `mochaHooks` via `--require`.  the function(s) defined here are _guaranteed_ to run once and only once, in serial, parallel, serial-watch or parallel-watch modes. this should be the _preferred_ way of running global setup/teardown code, and I'll need to update the docs accordingly.  the setup and teardown functions will share a single context object, and will only run in the main process.

more:

- adds a generic "plugin loader" which handles root hooks, global setup/teardown, future.  these "plugin definitions" (of which there are three: root hooks, global setup, and global teardown) adhere to a rough "interface" at this point, but are not entirely self-contained--functionality is still tightly coupled to the `Mocha` class, which needs instance methods to support each.  I'm choosing to defer any decisions on this design for now
- renamed `validatePlugin` to `validateLegacyPlugin`. reporters/interfaces could/should eventually be refactored to use the same system.  this would make it less awkward to define a custom reporter, for instance.
- soft-deprecates `errors.createPluginError`; directs consumers to use `errors.createLegacyPluginError` instead


boneskull:
this will get rebased onto `master` once #4293 lands.  just can't be bothered to write ES5

boneskull:
the design here is up for discussion btw.  we have a new file, `lib/plugin.js` which knows about "the special stuff you can export from a `--require`'d file" (e.g., `mochaHooks`, `mochaGlobalSetup`, etc.) and has validation/processing functionality thrown in there.  maybe each "special thing" should have its own module.  I dunno.

was thinking of taking it a step further and moving _all_ functionality into some sort of system that, say, uses event listeners to define functionality.  so there'd be some implementation of `mochaGlobalSetup`, `mochaGlobalTeardown`, and `mochaHooks` which would be completely decoupled from `lib/mocha.js`, `lib/runner.js`, etc.  so the implementation for global setup & teardown, which currently lives in `Runner#run`, could just listen for the `EVENT_RUN_BEGIN` event and then handle its business.

lib/cli/run-helpers.js
```diff
@@ -83,33 +84,21 @@ exports.list = str =>
  * @private
  */
 exports.handleRequires = async (requires = []) => {
-  const acc = [];
-  for (const mod of requires) {
+  const pluginLoader = PluginLoader.create();
+  for await (const mod of requires) {
```
boneskull, line 88: this may not be strictly necessary, but I think it's going to help reason about the order in which requires are loaded.

lib/plugin.js
```diff
@@ -0,0 +1,142 @@
...
+}));
+const {
+  PLUGIN_ROOT_HOOKS,
+  PLUGIN_GLOBAL_SETUP,
+  PLUGIN_GLOBAL_TEARDOWN
+} = constants;
+const PLUGIN_NAMES = new Set(Object.values(constants));
+
+exports.PluginLoader = class PluginLoader {
+  constructor({
```
boneskull, line 19: overriding the defaults here is just for testability at this point.

lib/plugin.js
```diff
@@ -0,0 +1,142 @@
...
+          return true;
+        }
+        return isFound;
+      }, false);
+    }
+    return false;
+  }
+
+  async finalize() {
+    const mochaHooks = this.pluginMap.get(PLUGIN_ROOT_HOOKS);
```
boneskull, line 53: I should add this in a code comment, but the idea here is that we call `load()` on all required modules, which fishes the special named exports out of those modules, and then calling `finalize()` aggregates/flattens everything into a single data structure, which is suitable for passing into `Mocha` constructor

lib/plugin.js
```diff
@@ -0,0 +1,142 @@
...
+    isValid = false;
+  }
+  if (!isValid) {
+    throw createUnsupportedError(
+      `${pluginType} must be a function or an array of functions`
+    );
+  }
+};
+
+const PluginValidators = {
```
boneskull, line 93: I dunno, these are some validator functions to check that the named exports e.g. `mochaHooks` are of the correct shape.  the `createFunctionArrayValidator` one above looks for a function or array thereof.  if we're going to be in the business of validating a bunch of code this way, maybe we should use a suitable library (e.g., `ajv` or `joi` or something) instead of this.

lib/plugin.js
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Loads root hooks as exported via `mochaHooks` from required files.
+ * These can be sync/async functions returning objects, or just objects.
+ * Flattens to a single object.
+ * @param {MochaRootHookObject[]|MochaRootHookFunction[]} rootHooks - Array of root hooks
+ * @private
+ * @returns {MochaRootHookObject}
+ */
+const aggregateRootHooks = (exports.aggregateRootHooks = async (
```
boneskull, line 116: moved from `run-helpers` module

lib/runner.js
```diff
@@ -974,6 +982,54 @@ Runner.prototype._uncaught = function(err) {
...
+  debug('run(): global teardown complete');
+  this.emit(constants.EVENT_POST_GLOBAL_TEARDOWN);
+};
+
+Runner.prototype.runGlobalFixtures = function runGlobalFixtures(
+  fixtureFns,
+  events
+) {
+  const context = (this._globalContext = this._globalContext || {});
+  return fixtureFns.reduce(async (promise, fixtureFn) => {
```
boneskull, line 1018: maybe I want `for await ..of` here

boneskull:
What I mean by "reporters and interfaces should use this plugin system" above is that it's currently awkward to implement a third-party reporter or interface.  You have to essentially set properties on `Mocha.interfaces` or `Mocha.reporters`.  This new system could detect the presence of a reporter/interface in a `--require`'d file via named exports and do the plumbing for you.  We'd also have a chance, if e.g., `--require my-reporter` was used, to _enable_ the reporter automatically if no other `--reporter` was given (and this would need to be limited to a _single_ reporter, since we don't have multiple reporter support).

boneskull:
rebasing onto branch `boneskull/rollup-umd`

boneskull:
To support watch mode effectively, I need to rip this impl out of `Runner` and put it into `Mocha`.

boneskull:
hey, green build finally.

looks like it needs some more tests though.

lib/cli/run-helpers.js
```diff
@@ -79,62 +79,31 @@ exports.list = str =>
...
-    ) {
-      const mochaHooksType = type(requiredModule.mochaHooks);
-      if (/function$/.test(mochaHooksType) || mochaHooksType === 'object') {
-        debug('found root hooks in required file %s', mod);
-        acc.push(requiredModule.mochaHooks);
-      } else {
-        throw createUnsupportedError(
-          'mochaHooks must be an object or a function returning (or fulfilling with) an object'
-        );
+    if (requiredModule && typeof requiredModule === 'object') {
```
boneskull, line 95: the logic here is "look in the module we've loaded via `--require` for any known named exports and figure out what to do with them".  since these could be multiple and spread out over several files, on L102 below, we call `pluginLoader.finalize()` which allows aggregation of these into a single object, suitable for passing into the `Mocha` constructor.  

for example, we might have `mochaHooks` in `a.js` and `mochaHooks` in `b.js`; the `pluginLoader.load()` call will _collect_ these, and we mash them all together into an object (containing props for each hook type, and arrays of hook callback functions) in `pluginLoader.finalize()`.

lib/cli/run.js
```diff
@@ -339,13 +338,10 @@ exports.builder = yargs =>
...
-        validatePlugin(argv, 'reporter', Mocha.reporters);
-        validatePlugin(argv, 'ui', Mocha.interfaces);
-
-        if (rawRootHooks && rawRootHooks.length) {
-          argv.rootHooks = await loadRootHooks(rawRootHooks);
-        }
+        const plugins = await handleRequires(argv.require);
+        validateLegacyPlugin(argv, 'reporter', Mocha.reporters);
+        validateLegacyPlugin(argv, 'ui', Mocha.interfaces);
+        Object.assign(argv, plugins);
```
boneskull, line 344: instead of just a `rootHooks` prop on the options object for `Mocha`, the `plugins` object will have several properties which we want to merge in.

lib/cli/watch-run.js
```diff
@@ -177,10 +178,39 @@ const createWatcher = (
...
+      }
+    });
+  }
+
+  let exiting = false;
+  process.on('SIGINT', async () => {
     showCursor();
-    console.log('\n');
-    process.exit(128 + 2);
+    console.error(`${logSymbols.warning} [mocha] cleaning up, please wait...`);
```
boneskull, line 201: I can rip this out--it's useful, and I needed it for debugging--but it adds some feedback in watch mode, so tangential to this PR.

lib/mocha.js
```diff
@@ -205,6 +207,11 @@ function Mocha(options) {
    */
   this.isWorker = Boolean(options.isWorker);
 
+  this.globalSetup(options.globalSetup)
```
boneskull, line 210: the functions are in `globalSetup`/`globalTeardown`, and `enableGlobalSetup`/`enableGlobalTeardown` is consumed by the watcher.

lib/mocha.js
```diff
@@ -1097,8 +1123,145 @@ Mocha.prototype.lazyLoadFiles = function lazyLoadFiles(enable) {
...
+};
+
+/**
+ * Run global fixtures sequentially with context `context`
+ * @private
+ * @param {MochaGlobalFixture[]} [fixtureFns] - Fixtures to run
+ * @param {object} [context] - context object
+ * @returns {Promise<object>} context object
+ */
+Mocha.prototype._runGlobalFixtures = async function _runGlobalFixtures(
```
boneskull, line 1207: this can either be static or just a helper function outside of the `Mocha` class.  thoughts?

lib/nodejs/parallel-buffered-runner.js
```diff
@@ -14,6 +14,18 @@ const {BufferedWorkerPool} = require('./buffered-worker-pool');
 const {setInterval, clearInterval} = global;
 const {createMap} = require('../utils');
 
+/**
```
boneskull, line 17: certain options are not supported by worker processes, listed below.  first and foremost, a worker cannot run mocha in parallel 😄 

these options are stripped out before `Mocha` is instantiated.

`globalSetup`/`globalTeardown` are the responsibility of the _main process only_, by design

lib/mocha.js
```diff
@@ -1108,7 +1271,40 @@ Mocha.prototype.lazyLoadFiles = function lazyLoadFiles(enable) {
...
  */
+
+/**
+ * A function that's invoked _once_ which is either sync or async.
+ * Can be a "teardown" or "setup".  These will all share the same context.
+ * @callback MochaGlobalFixture
+ * @returns {void|Promise<void>}
+ */
+
+/**
```
boneskull, line 1285: This describes the shape of a plugin definition. the builtin plugins are defined in `lib/plugin-loader.js` around L24 (`MochaPlugins`).  A plugin definition defines everything about how a user would provide their plugin implementation, and how that implementation ends up in a `Mocha` instance.  once the implementation reaches the `Mocha` instance, it's no longer managed by this definition.  It's _not_ atomic in this way--it's tightly coupled to the `Mocha` class, which must do whatever it needs to do to "use" the implementation.  e.g., the plugin definition says "use the `mochaHooks` export, it should pass this validation function, and call the resulting option `rootHooks` when instantiating `Mocha`".  after that, it's `Mocha`'s problem.

what this looks like exactly is _not_ something we need to make public atm. I tried to keep the abstraction minimal, but it may not be enough or it may be too much. in other words, we don't support 3rd-party plugin definitions

boneskull: I think it would be _nice_ to make it completely self-contained, but we just don't have any place to "hook in" to Mocha to handle the implementation (e.g., "run any global setup fixtures as async functions--in the main process only--before the `run` event is emitted").  adding those hooks would be _very_ involved, and this PR is big enough already.

lib/plugin-loader.js
```diff
@@ -0,0 +1,286 @@
...
+      if (
+        Array.isArray(value) ||
+        (typeof value !== 'function' && typeof value !== 'object')
+      ) {
+        throw createInvalidPluginImplementationError(
+          `mochaHooks must be an object or a function returning (or fulfilling with) an object`
+        );
+      }
+    },
+    async finalize(rootHooks) {
```
boneskull, line 42: note the other two plugin definitions do _not_ have a "finalize" implementation. it's needed here (this code is not new; it came from elsewhere) because of the deeply-nested data structure and the ability to define hooks asynchronously (the user supplies an async function which returns an object containing root hooks).

global setup/teardown fixtures can be async, of course, but I don't yet see a legit use-case for allowing a function _which returns_ a fixture.

lib/plugin-loader.js
```diff
@@ -0,0 +1,286 @@
...
+    }
+  },
+  /**
+   * Global setup fixture plugin definition
+   * @type {PluginDefinition}
+   */
+  {
+    exportName: 'mochaGlobalSetup',
+    optionName: 'globalSetup',
+    validate(value) {
```
boneskull, line 78: there's duplication here, but I thought abstracting it was kind of premature.

lib/runner.js
```diff
@@ -984,66 +985,71 @@ Runner.prototype._uncaught = function(err) {
...
+    debug('run(): root suite completed; emitting %s', constants.EVENT_RUN_END);
+    this.emit(constants.EVENT_RUN_END);
+  };
+
+  const begin = () => {
+    debug('run(): emitting %s', constants.EVENT_RUN_BEGIN);
+    this.emit(constants.EVENT_RUN_BEGIN);
+    debug('run(): emitted %s', constants.EVENT_RUN_BEGIN);
+
+    this.runSuite(rootSuite, async () => {
```
boneskull, line 1005: I don't think this needs to be async...

lib/runner.js
```diff
@@ -984,66 +985,71 @@ Runner.prototype._uncaught = function(err) {
...
-      debug('run(): emitted %s', constants.EVENT_RUN_END);
-    });
-  }
+    return begin();
+  };
 
   // references cleanup to avoid memory leaks
   if (this._opts.cleanReferencesAfterRun) {
-    this.on(constants.EVENT_SUITE_END, function(suite) {
+    this.on(constants.EVENT_SUITE_END, suite => {
```
boneskull, line 1028: feels good to clean this up

lib/runner.js
```diff
@@ -984,66 +985,71 @@ Runner.prototype._uncaught = function(err) {
...
+    this.state = constants.STATE_STOPPED;
     debug('run(): emitted %s', constants.EVENT_RUN_END);
-    fn(self.failures);
+    fn(this.failures);
   });
 
-  self._removeEventListener(process, 'uncaughtException', self.uncaught);
-  self._addEventListener(process, 'uncaughtException', self.uncaught);
+  this._removeEventListener(process, 'uncaughtException', this.uncaught);
+  this._removeEventListener(process, 'unhandledRejection', this.uncaught);
```
boneskull, line 1041: this may have unintended consequences.  but we now have a global `unhandledRejection` listener.

I think it probably wants some integration tests.

craigtaub:
Started having a read, plugins look good. 

So we are following the same approach as Jest + not allowing state or resources to be available for tests? So is intended for things like start/stop of a server or db, but will need root suite hook to create connection instance to use in tests. It is fine I just wanted to be clear.

I was thrown by the "fixture" term, as we use it elsewhere in the repo and then we follow the common JS ecosystem meaning. 
I understand the difficulty though as:
1. Plugins could/might be for more than just hooks so they arent synonymous
2. "globalSetup" is not a hook 
3. Its useful to have "globalSetup plugin" + "globalTeardown plugin" coupled semantically as they are in code

So im not sure what to suggest, seems useful to have this new category, but could get confusing.

boneskull:
Thanks @craigtaub. Yeah there's really no reasonable way to share a file, db handle, socket etc across spawned worker processes. Each worker process would have its own handle, and this would be suitable for root hook plugins instead of these globals. 

You can share state between global setup and teardown, but the suites, hooks, and tests themselves can not access that state. 

"Fixture", I think, is not inappropriate (see [wiki](https://en.m.wikipedia.org/wiki/Test_fixture)). we're using it to mean roughly the same thing in our test "fixture" files--it's some environment or a context we test in.

boneskull:
@craigtaub Can you take a look, specifically, at the changes to the "run cycle overview" in `docs/index.md`? It was out-of-date with respect to parallel mode, root hooks, and this PR (and I felt like it needed further expansion)

boneskull:
I'm referencing version 9 in the docs here, but now I can't remember why.  I thought I was breaking something, but maybe I'm not.  Need to double-check.

docs/index.md
```diff
@@ -116,51 +116,60 @@ $ npm test
...
+
+The following is a mid-level outline of Mocha's "flow of execution" when run in Node.js; the "less important" details have been omitted.
+
+In a browser, test files are loaded by `<script>` tags, and calling `mocha.run()` begins at step 9 [below](#serial-mode).
+
+### Serial Mode
+
+1. User (that's you) executes `mocha`
+2. Loads options from config files, if present
+3. Mocha processes any command-line options provided
```
craigtaub, line 129: Probably worth linking [option priority](https://mochajs.org/#merging).

docs/index.md
```diff
@@ -116,51 +116,60 @@ $ npm test
...
-  |<-------------- per spec file end
-|<-------------- inside child process end
-```
+> Updated for v9.0.0.
+
+The following is a mid-level outline of Mocha's "flow of execution" when run in Node.js; the "less important" details have been omitted.
+
+In a browser, test files are loaded by `<script>` tags, and calling `mocha.run()` begins at step 9 [below](#serial-mode).
+
+### Serial Mode
```
craigtaub, line 125: Worth splitting into parsing and execution phases?

docs/index.md
```diff
@@ -116,51 +116,60 @@ $ npm test
...
+1. User (that's you) executes `mocha`
+2. Loads options from config files, if present
+3. Mocha processes any command-line options provided
+4. If known flags for the `node` executable are found:
+   1. Mocha will spawn `node` in a child process, executing itself with these flags
+   2. Otherwise, Mocha does not spawn a child process
+5. Mocha loads modules specified by `--require`
+   1. If a file loaded this way contains known Mocha-specific exports (e.g., [root hook plugins]), Mocha "registers" these
+   2. If not, Mocha ignores any exports of a `--require`'d module
+6. Mocha _discovers_ test files; when given no files or directories, it finds files with extensions `.js`, `.mjs` or `.cjs` in the `test` directory (but not its children), relative to the current working directory
```
craigtaub, line 136: Do we validate `--reporter` and `--ui` a step before this?

docs/index.md
```diff
@@ -116,51 +116,60 @@ $ npm test
...
+7. The (default) [bdd interface](#bdd) loads the test files _in no particular order_, which are given an interface-specific `global` context (this is how, e.g., `describe()` ends up as a global in a test file)
+   1. When a test file is loaded, Mocha executes all of its suites and finds--\_but does not execute\_\_--any hooks and tests therein.
```
craigtaub, line 138: `finds--\_but does not execute\_\_--any` is that formatting correct?

docs/index.md
```diff
@@ -116,51 +116,60 @@ $ npm test
...
+2. All test files found are put into a queue (they are _not_ loaded by the main process)
+3. Mocha runs [global setup fixtures], if any
+4. Mocha creates a pool of subprocesses ("workers")
+5. Each worker loads `--require`'d modules (but _not_ global fixtures)
+6. When a worker receives a test file to run, the worker creates a new Mocha instance _for the single test file_, and:
+7. Repeats step 7 from [above](#serial-mode)
+8. Repeats step 9 from [above](#serial-mode), with the caveat that the worker _does not_ report test results directly; it holds them in a memory buffer
+9. When the worker completes the test file, buffered results are returned to the main process, which then gives them to the user-specified reporter (`spec` by default)
+10. The worker makes itself available to the pool; the pool gives the worker another test file to run, if any remain
+11. Mocha prints a summary, if applicable
```
craigtaub, line 164: Possible worth `Repeats step 10 (and 11) from above`

craigtaub:
Nice, "run cycle overview" is much improved

test/unit/plugin-loader.spec.js
```diff
@@ -0,0 +1,514 @@
...
+            return expect(pluginLoader.finalize(), 'to be fulfilled with', {
+              baz: ['polar bears']
+            });
+          });
+        });
+      });
+    });
+  });
+
+  describe('root hoots plugin', function() {
```
craigtaub, line 355: `hooks` 🦉

lib/mocha.js
craigtaub: Unless there is a particular use-case for it elsewhere, or a consequence of it being here, im fine with it here.

docs/index.md
boneskull: I'm not convinced I can make it any clearer by drawing a line there. do you have any specific suggestions on how to explain it clearly?

docs/index.md
boneskull: yes

docs/index.md
boneskull: nope 😄

docs/index.md
boneskull: the summary is the epilogue, not the usual reporter output. i'll make it more clear

docs/index.md
craigtaub: Not for clarity, just to codify our definition of where these phases start and what's included. Was just imagining a simple divide I.e.
```
--- Parsing phase
1. ....
...
4. ...
---
--- Execution phase
5. ....
----
`
But they are mainly useful for internal purposes anyway so happy to drop. Leave with you.

Comment (boneskull) [reply docs/index.md:125]: hmmm......


I'm not sure _I_ know what the definition is.  certainly, we run our suite callbacks (I think this is what you're referring to with "parsing phase") in their entirety before our hook and test callbacks.  In the changes I've made, step 8. refers to this:

> 8. The (default) [bdd interface](#bdd) loads the test files _in no particular order_, which are given an interface-specific `global` context (this is how, e.g., `describe()` ends up as a global in a test file)
>     1. When a test file is loaded, Mocha executes all of its suites and finds--_but does not execute_--any hooks and tests therein.
>     2. Top-level hooks, tests and suites are all made members of an "invisible" _root suite_; there is only _one_ root suite for the entire process

after that, we run global setup fixtures, which really has nothing to do with the suites.  after _that_, we run the hook/test callbacks.  so to me anyway, there's no clear "logical grouping" of tasks, like you've written above.  it's just a sequence of steps.

now... we could make an argument that global setup fixtures should actually run _before_ test files are loaded.  that _might_ make more sense, but it would mean that we'd have to expose APIs for programmatic users to trigger setup _and_ teardown execution manually.  maybe someone wants this, but a) you can toggle the behavior on and off anyway via the public `Mocha#enableGlobalSetup()` and `Mocha#enableGlobalTeardown()` methods, and b) it's not very friendly to make programmatic users go thru extra hoops to get standard behavior.  

however, this brings up the point that `Mocha#runGlobalSetup()` and `Mocha#runGlobalTeardown()` should be public, if a programmatic users wants to fiddle with when these run:

```js
const mocha = new Mocha();
mocha.addFile('/some/file.js');
mocha.enableGlobalSetup(false);
// run the global setup early, before mocha.run() is called
await mocha.runGlobalSetup();
// do something else here
mocha.run();
```

boneskull:
this has gotten too unwieldy and I'm going to try to break it up

lib/runner.js
```diff
@@ -976,73 +1006,92 @@ Runner.prototype._uncaught = function(err) {
...
 
+/**
+ * Like {@link Runner#run}, but does not accept a callback and returns a `Promise` instead of a `Runner`.
+ * This function cannot reject; an `unhandledRejection` event will bubble up to the `process` object instead.
+ * @public
+ * @memberof Runner
+ * @param {Object} [opts] - Options for {@link Runner#run}
+ * @returns {Promise<number>} Failure count
+ */
+Runner.prototype.runAsync = async function runAsync(opts = {}) {
```
nicojs, line 1089: I like this! 😍

nicojs (approved):
This cleans up the code nicely and is an excellent feature.

docs/index.md
craigtaub: Hmm interesting.
Sorry I should have said, my understanding is the parsing phase is just options and validation. So it's very short. I think all the loading of files and running comes under execution phase. But that's just my own mental model.

craigtaub (approved):

boneskull:
I am not going to attempt to break this up, because that's a pain in the ass.  suffice to say there are things in here that could technically be in separate PRs.  I am sorry.

boneskull:
I'm going to assume we're OK with moving `deprecate()` and `warn()` into `lib/errors` from `lib/utils`.  It is not ideal, but it is problematic to attempt to bundle a cyclic dependency, as `utils` needed `errors` and vice-versa.

boneskull:
when AppVeyor passes I'm going to merge this, finally.

boneskull:
I have some ideas around documenting this further, but I can do it in a subsequent PR.

wbt:
FYI as feedback to help you calibrate semver choices for the future, this was a breaking change for reasons summarized [here](https://github.com/winstonjs/winston/issues/2018), and should've gone out in 9.0.0. 
Thanks for all your work on this project!

Issue #4356, opened 2020-06-30
reduce netlify build times for PRs


boneskull:
The only branch that needs the "full production build" (the `docs.postbuild` script in `package-scripts.js`) is the `mochajs.org` branch, which is what mochajs.org is built from.  This script takes up _most_ of the build time, and we want to reduce the build minutes on netlify, because we're using too many.

I _think_ it's relatively safe to disable.  It's been a source of flake, and the problems are usually (but not always) corrected by a cache clear.

In its place, we'd need something to essentially just _rename_ `docs/_site` to `docs/_dist`, as `docs/_dist` is ultimately what gets deployed.

What do we think? @mochajs/core @Munter

Munter:
Some of the production build creates significant performance improvements for our visitors, so I wouldn't want to disable them for the master branch. We could probably skip them for branches that don't work on the build system

outsideris:
I've tried to optimize build time at #4264 , but I couldn't reduce time effectively.

I think we can run Netlify builds when the only PR has changes under `docs`.

boneskull:
@outsideris can #4264 be closed?

this should have been closed via #4358

PR #4361, opened 2020-07-02
Exclude jsdoc pages from bundling and optimization

Munter:
By excluding the jsdoc output from being handled by assetgraph-builder we gain a significant speed improvement on postprocessing the docs.

Downside: API docs aren't as optimized. But I think the speed increase is worth the trade-off

Relates to #4356

**Latest build on `master`:**
> Build time: 5m 39s. Total deploy time: 5m 41s

**PR preview build from this branch:**
> Build time: 2m 36s. Total deploy time: 2m 38s

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -302,7 +302,7 @@ module.exports = {
       },
       postbuild: {
         script:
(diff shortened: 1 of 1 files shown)


boneskull:
Closing this in lieu of #4358 which just disables optimization entirely for pull requests.

PR #4362, opened 2020-07-02
Handle non-png responses for opencollective avatars with static grey replacement images

Munter:
Some time OpenCollective respond to our avatar downloads during site build with something other than a PNG.
Our current version of the avatar download just ignores the reponse mime type and writes the response to disk. If the response is not a PNG the spriting automation fails in the docs post processing.

This change ensures that the response is png before writing it to disk.
If the response is not png, a replacement avatar is written to disk in the right dimensions

### DIFF
--- a/docs/_data/supporters.js
+++ b/docs/_data/supporters.js
@@ -7,6 +7,7 @@ const {resolve} = require('path');
 const debug = require('debug')('mocha:docs:data:supporters');
 const needle = require('needle');
 const imageSize = require('image-size');
+const fileType = require('file-type');
 const blocklist = new Set(require('./blocklist.json'));
 
 const API_ENDPOINT = 'https://api.opencollective.com/graphql/v2';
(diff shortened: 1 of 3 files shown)


boneskull:
@Munter should we combine this with #4357

afaik, we _always_ get PNG files from OC.  when I look at my downloaded cache, they are all PNGs.

it's just that certain PNGs are unreadable by canvas, apparently?  or maybe corrupted?  either way, `image-size` did not seem to be of help.  our build will break when `canvas` encounters an invalid image, so I think it's best to let `canvas` decide whether its valid or not.  

my solution in #4357 was to just _drop_ the image entirely, but replacing it with a blank graphic as you've done here is a better idea.

boneskull:
@munter should we resolve the conflicts or drop this one?  up to you

docs/_data/supporters.js
```diff
@@ -139,7 +165,15 @@ module.exports = async () => {
     supporters.backers.map(async backer => {
       const filePath = resolve(supporterImagePath, backer.id + '.png');
       const {body} = await needle('get', encodeURI(backer.avatar));
-      await writeFile(filePath, body);
+
+      let imageBuffer = body;
+
+      // Some times OpenCollective responds with a non-image
+      if ((await fileType.fromBuffer(body)).mime !== 'image/png') {
```
boneskull, line 172: so yeah, with my changes, _`canvas`_ is doing the file checking.  currently, if it fails, it just discards the image, but you could replace that with this that just uses a blank one... or not.

Munter:
Let's just drop it. The current solution seems fine

PR #4363, opened 2020-07-04
Reduce rollup noise

Munter:
These changes reduce rollup noise

- Silenced circular dependency warning
- Moved sinon out as an external when running bundled browser tests. This speeds browser test builds and removes warning about deoptimization in babel
- Skip evaluating browser test file globs and let rollup do that


scripts/karma-rollup-plugin.js
```diff
@@ -125,18 +120,31 @@ function bundlePreprocessor(config) {
 
     const bundle = await rollup.rollup({
       input: fileMap.get(file.path),
-      plugins: [...plugins, multiEntry({exports: false})]
+      plugins: [...plugins, multiEntry({exports: false})],
+      external: ['sinon'],
+      onwarn: (warning, warn) => {
```
boneskull, line 125: looks like we could just copy the `onwarn` from `rollup.config.js` since we already require it, right?

Munter: We could indeed. I'll make that change

Munter: Fixed in e5fd667d

PR #4365, opened 2020-07-05
Replace utils constants and map functions with native versions

Munter:
This replaces `utils.defineConstants` with `Object.freeze` and `utils. createMap` with actual map usage.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -37,7 +37,7 @@ exports = module.exports = Mocha;
  * These are the states it can be in.
  * @private
  */
-var mochaStates = utils.defineConstants({
+var mochaStates = Object.freeze({
   /**
    * Initial state of the mocha instance
    * @private
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -582,25 +582,6 @@ exports.createMap = function(obj) {
 };
 
 /**
- * Creates a read-only map-like object.
- *
- * @description
- * This differs from {@link module:utils.createMap createMap} only in that
- * the argument must be non-empty, because the result is frozen.
- *
- * @see {@link module:utils.createMap createMap}
- * @param {...*} [obj] - Arguments to `Object.assign()`.
- * @returns {Object} A frozen object with no prototype, having `...obj` properties
- * @throws {TypeError} if argument is not a non-empty object.
- */
-exports.defineConstants = function(obj) {
-  if (type(obj) !== 'object' || !Object.keys(obj).length) {
-    throw new TypeError('Invalid argument; expected a non-empty object');
-  }
-  return Object.freeze(exports.createMap(obj));
-};
-
-/**
  * Whether current version of Node support ES modules
  *
  * @description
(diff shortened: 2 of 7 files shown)


lib/mocha.js
```diff
@@ -37,7 +37,7 @@ exports = module.exports = Mocha;
  * These are the states it can be in.
  * @private
  */
-var mochaStates = utils.defineConstants({
+var mochaStates = Object.freeze({
```
juergba, line 40: Can we use `const` instead of `var`, does it get transpiled?

lib/utils.js
```diff
@@ -559,47 +559,6 @@ exports.dQuote = function(str) {
...
 
-/**
- * Creates a map-like object.
- *
- * @description
- * A "map" is an object with no prototype, for our purposes. In some cases
- * this would be more appropriate than a `Map`, especially if your environment
- * doesn't support it. Recommended for use in Mocha's public APIs.
- *
- * @public
```
juergba, line 570: This is a public function. So theoretically we should soft deprecate it first ...
I don't think that many people are using it.

lib/mocha.js
Munter: We can, it does get transpiled. But I think I'd prefer to do everything at once with lebab instead of trickling those in in unrelated changes

lib/utils.js
Munter: Good point. That need to be corrected

lib/mocha.js
boneskull: I'm unsure how I feel about this... 

On the one hand, we've done huge sweeping changes (e.g., prettier, eslint) before. But this makes `git blame` less useful.

On the other, this will look inconsistent and weird if we change things piecemeal.

I'd like to get the opinion of @mochajs/core on how we should proceed with "modernizing" our codebase

boneskull:
_renaming_ `createMap` to avoid confusion is maybe a good idea. `createObjectMap`?

lib/mocha.js
juergba: I guess modernizing our code base won't be a big bang. We should include smaller ES6 changes in our daily PR's. There should be dedicated PR's for bigger changes as e.g. turning runner's callback hell into something more pleasant.
So above not changing `var` to `const` looks odd and IMO should be changed.

lib/utils.js
juergba: I don't see why this function has to be public. We should remove it from our API.

juergba:
An `Object` is probably the better choice than `Map` for these constants. `Map`'s are not freezable as mentioned before. The resulting object is similar to `Enum` (eg. in Java), we could rename the function also to `createEnum`. Btw IE11 does support `Map`'s (as per mozilla).

I don't understand:
- why `createMap` has to be public.
- why `createMap` is working on IE11, as per mozilla IE11 does not support `Object.assign`.

boneskull:
Yeah, I'm not sure why `createMap` is public.  It probably shouldn't be.

boneskull:
@munter please uh... do something with this 😄

Issue #2426, opened 2016-08-08
add integration tests for AMD


boneskull:
We should have something that ensures Mocha can be used with require.js & its ilk.

Relevant: #2424, #880 sort of

Bamieh:
Is this still valid? @boneskull

juergba:
closed by #4366.

PR #4366, opened 2020-07-06
build UMD bundle and polyfill language features

boneskull:
- remove old ESlint rule to disallow `Object.assign`
- consume Rollup config's `output` prop in `karma-rollup-plugin`
- simplify `rollup.config.js` `output` prop
- fix invalid `script` tag `src` attrib in `lib/browser/template.html` (& reformat)


boneskull:
@Munter I found that when using `umd` output while `window.mocha.setup` did not exist, `window.mocha.mocha.setup` _did_ exist.  

using object spread syntax `module.exports = {...global, ...mocha}` causes Rollup (?) to generate a ES module then complains about having default and named exports.  I have no idea why.

I think even the `global.mocha = mocha` and `global.Mocha = Mocha` lines are probably unnecessary, since the UMD bundle will stuff `module.exports` into `global`.

boneskull:
update: the `global.mocha` and `global.Mocha` stuff _is_ necessary.  I guess I really don't get it.  but whatever.

boneskull:
looks like I jumped the gun a little there.  it worked for a test file anyway (as created by `mocha init`)

boneskull:
alright, tweaked the exports. it's now

```js
mocha.Mocha = Mocha;
mocha.mocha = mocha;
module.exports = Object.assign(mocha, global);
```

boneskull:
AMD bundle seems to "work" per manual test, but the failing test isn't of much value.

boneskull:
added a proper test for AMD; closes #2426

karma.conf.js
```diff
@@ -134,19 +135,38 @@ module.exports = config => {
         {pattern: 'test/browser-specific/esm.spec.mjs', type: 'module'}
       ];
       break;
+    case 'requirejs':
+      cfg.plugins.push('karma-requirejs');
+      cfg.frameworks.unshift('requirejs');
```
boneskull, line 140: why does order matter? _I have no idea_

karma.conf.js
```diff
@@ -134,19 +135,38 @@ module.exports = config => {
...
 
-  cfg.files.unshift(
-    require.resolve('unexpected/unexpected'),
-    {pattern: require.resolve('unexpected/unexpected.js.map'), included: false},
-    require.resolve('unexpected-sinon'),
-    require.resolve('unexpected-eventemitter/dist/unexpected-eventemitter.js'),
-    require.resolve('./test/browser-specific/setup')
-  );
+  if (MOCHA_TEST !== 'requirejs') {
+    cfg.files.unshift(
```
boneskull, line 157: unexpected is broken when used with AMD.  something about anonymous defines

boneskull:
@munter something is not pulling in the target browser list here

boneskull:
wow, it works!

test/unit/utils.spec.js
```diff
@@ -496,11 +496,11 @@ describe('lib/utils', function() {
     if (typeof global.Symbol === 'function') {
       it('should handle Symbol', function() {
         var symbol = Symbol('value');
-        expect(stringify(symbol), 'to be', 'Symbol(value)');
+        expect(stringify(symbol), 'to match', /^Symbol\(value\)/);
```
boneskull, line 499: `core-js` adds weird characters to the `Symbol` string representation in IE11.  🤷‍♂️

karma.conf.js
```diff
@@ -17,7 +17,7 @@ const browserPlatformPairs = {
 };
 
 module.exports = config => {
-  let bundleDirpath;
+  let bundleDirpath = path.join(baseBundleDirpath, hostname);
```
boneskull, line 20: previously, this would have been added to the `cfg` object before it was defined.  now, we only add it if creating the directory succeeds.

maybe we should ignore a `mkdir` failure and just use the tmp dir...  I don't want to run the build again 😜

lib/browser/template.html
```diff
@@ -1,16 +1,18 @@
...
   </head>
   <body>
     <div id="mocha"></div>
     <script src="mocha.js"></script>
-    <script>mocha.setup('bdd');</script>
-    <script src="tests.js"></script>
+    <script>
+      mocha.setup('bdd');
+    </script>
+    <script src="tests.spec.js"></script>
```
boneskull, line 15: this was wrong.

we should run linkcheck on this thing.

.travis.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       node_js: '10'
       name: 'Node.js v10'
 
-    - script: npm start test.bundle test.browser
+    - script: npm start test.browser
```
boneskull, line 52: `test.bundle` becomes `test.browser.requirejs`

juergba:
@boneskull I can't review this PR explained by my complete incompetence in this matter, sorry.
This rollup thing is a wild fusion of bundling, transpiling and karma/testing without clear distinction. I don't understand neither why we need an additional module definition UMD. What definition have we used before with browserify?

boneskull:
our browserify bundle is not _quite_ UMD, but it 1. defines globals and 2. enables bundling of the bundle.  it does not supply a AMD-compatible bundle, but it also does not _inhibit_ AMD bundles.

The rationale for this PR, then, is:

1. the `iife` output of Rollup does _not_ seem to allow number 2, which is going to break consumers; we need the `umd` output style
2. after #4293 landed, we're able to use _some_ modern syntax everywhere, but not _all_ modern syntax everywhere.  the changes to our babel config in this PR make it possible to use, e.g., `async`/`await` anywhere in our sources and tests (because Node.js v10, our minimum-supported version of Node.js, allows this).  We could not use the native `Promise` object, either--but now we can, because of the polyfilling core-js does.
3. the `umd` output style of Rollup _does_ provide an AMD bundle, which made the previous test case (`test.bundle`) fail.  I had to write some new tests to assert compatibility (and learn some RequireJS...ugh)

with this PR--for whatever reason--the resulting bundle size is about 20Kb less.

boneskull:
that said, this has been sitting for like 3 weeks, and I need to move forward with it to get global fixtures in.

Issue #4368, opened 2020-07-08
[Proposal] Safer, cleaner, and simpler teardowns through generator fixtures


SalmonMode:
Tl;dr: Added ability to use yield in before and beforeEach to separate setup and teardown logic. Teardowns will then only get executed if the setup doesn't have any issues.

## Example scenario

An e2e test needs to instantiate a client to interact with the admin API, use that client to create a user, launch a web browser session, use that session to go to a login page, and then login using the created user. In order to properly tear down after this test, the browser session must be quit, and the user must be deleted.

### The challenge introduced

While some parts of this setup process are independent of each other, there must be some serializable order to them; certain fixtures will have to execute/resolve before others. There is a chance that an error can occur in an earlier fixture that would prevent further fixtures from setting up.

Mocha currently handles this part just fine, as it will just stop executing `before` and `beforeEach` functions for that test.

The challenge is introduced with how teardowns can get handled. When mocha reaches a given `describe` layer for a test without any errors popping up before then, if a `before`/`beforeEach` throws an error, mocha will still try to execute all the `after`/`afterEach` for that layer, no matter how many `before`/`beforeEach` succeeded in that later. This presents a risk in that mocha could still attempt to tear down a fixture even if its setup wasn't completed.

Here's a quick mockup demonstrating the risks of this if the engineer assumes the setups won't have issues:

```javascript
describe('loggin in', () => {
  var driver;
  var loginPage;
  var landingPage;
  var adminClient;
  var user;

  before('create user', () => {
    user = User({ name = 'Jimmy', email = `test+${uuidv4()}@example.com`, password = 'P4$$w0rd!' });
    adminClient = AdminClient();
    adminClient.createUser(user);
  });

  before('launch driver', () => {
    driver = Chrome();
  });

  before('go to login page', () => {
    driver.get("https://mysite.com/login");
    loginPage = LoginPage(driver);
  });

  before('login', () => {
    loginPage.login(user);
    landingPage = LandingPage(driver);
  });

  it('should show user name in header', () => {
    expect(landingPage.header.text, 'to equal', user.name);
  });

  after('quit driver', () => {
    driver.quit();
  });

  after('delete user', () => {
    adminClient.deleteUser(user);
  });
});
```

[...]

PR #4369, opened 2020-07-08
[WIP] Yield fixtures

SalmonMode:
### Requirements

### Description of the Change

Tl;dr: Added ability to use yield in `before` and `beforeEach` to separate setup and teardown logic. Teardowns will then only get executed if the setup doesn't have any issues.

(copied from [issue](https://github.com/mochajs/mocha/issues/4368))

## Example scenario

An e2e test needs to instantiate a client to interact with the admin API, use that client to create a user, launch a web browser session, use that session to go to a login page, and then login using the created user. In order to properly tear down after this test, the browser session must be quit, and the user must be deleted.

### The challenge introduced

While some parts of this setup process are independent of each other, there must be some serializable order to them; certain fixtures will have to execute/resolve before others. There is a chance that an error can occur in an earlier fixture that would prevent further fixtures from setting up.

Mocha currently handles this part just fine, as it will just stop executing `before` and `beforeEach` functions for that test.

The challenge is introduced with how teardowns can get handled. When mocha reaches a given `describe` layer for a test without any errors popping up before then, if a `before`/`beforeEach` throws an error, mocha will still try to execute all the `after`/`afterEach` for that layer, no matter how many `before`/`beforeEach` succeeded in that later. This presents a risk in that mocha could still attempt to tear down a fixture even if its setup wasn't completed.

Here's a quick mockup demonstrating the risks of this if the engineer assumes the setups won't have issues:

```javascript
describe('loggin in', () => {
  var driver;
  var loginPage;
  var landingPage;
  var adminClient;
  var user;

  before('create user', () => {
    user = User({ name = 'Jimmy', email = `test+${uuidv4()}@example.com`, password = 'P4$$w0rd!' });
    adminClient = AdminClient();
    adminClient.createUser(user);
  });

  before('launch driver', () => {
    driver = Chrome();
  });

  before('go to login page', () => {
    driver.get("https://mysite.com/login");
    loginPage = LoginPage(driver);
  });

  before('login', () => {
    loginPage.login(user);
    landingPage = LandingPage(driver);
  });

  it('should show user name in header', () => {
    expect(landingPage.header.text, 'to equal', user.name);
  });

  after('quit driver', () => {
    driver.quit();
  });

  after('delete user', () => {
    adminClient.deleteUser(user);
  });
});
```

If creating the user throws an error, the driver will never be launched, but mocha would still attempt to quit it in the first `after`. What's worse, is that if the driver `before` was the one to have the error, the `driver.quit()` call would also throw an error, as there's no driver to actually quit, and, as a result, mocha would not attempt to finish the remaining teardowns, which means the user wouldn't get deleted.

There is no _guaranteed_ solution for the latter risk, as even if the driver had launched successfully, it's still possible (however unlikely) that quitting that driver session could throw an error.

The prior risk, though, is at least manageable. By doing a preliminary check in the `after`/`afterEach` to see if there is something that needs to be torn down would mitigate this risk. The `before` for user creation would have to be modified to this:

```javascript
  before('create user', () => {
    let _user = User({ name = 'Jimmy', email = `test+${uuidv4()}@example.com`, password = 'P4$$w0rd!' });
    adminClient = AdminClient();
    adminClient.createUser(_user);
    // necessary to delay assignment until user creation completion for simpler teardown logic
    user = _user;
  });
```

and the `after` for the driver and user creation would look like this:

```javascript
  after('quit driver', () => {
    if (driver) {
      driver.quit();
    }
  });

  after('delete user', () => {
    if (user) {
      adminClient.deleteUser(user);
    }
  });
```

This, however, increases the [complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of the tests, which would be ideal to avoid. Depending on how complex it is to either check for a need to teardown a fixture, or to do the teardown itself, this can drastically increase the complexity and the surface area for bugs in the test code (although, ideally for maximum safety/simplicity, each fixture would be atomic so the resulting complexity would be minimal).

## Proposed solution

Rather than having increased complexity in the test code to handle conditionally tearing down, it would be beneficial to not have the conditional at all, thus reducing complexity. Going one step further, it would also be beneficial to explicitly tie the teardown logic for a fixture to its setup logic in such a way that the teardown logic _can't_ be executed if the associated setup logic had a problem, thus reducing complexity further and increasing teardown safety.

By making fixtures into generators this can be easily achieved. I'm told generators aren't widely used for JavaScript development, but in this use case, they become more like context managers than generators.

Here's what the above example would look like using them:

```javascript
describe('loggin in', () => {
  var driver;
  var loginPage;
  var landingPage;
  var adminClient;
  var user;

  before('create user', function*() {
    user = User({ name = 'Jimmy', email = `test+${uuidv4()}@example.com`, password = 'P4$$w0rd!' });
    adminClient = AdminClient();
    adminClient.createUser(u);
    yield;
    adminClient.deleteUser(user);
  });

  before('launch driver', function*() {
    driver = Chrome();
    yield;
    driver.quit();
  });

  before('go to login page', () => {
    driver.get("https://mysite.com/login");
    loginPage = LoginPage(driver);
  });

  before('login', () => {
    loginPage.login(user);
    landingPage = LandingPage(driver);
  });

  it('should show user name in header', () => {
    expect(landingPage.header.text, 'to equal', user.name);
  });
});
```

In my implementation, mocha evaluates a hook's result after executing it initially and calling `next()` on it once. If the hook made it through to the `yield` successfully, then it is appended to the relevant `after`/`afterEach` array for that `describe` layer. Because they are executed sequentially (assuming the fixtures are executed in a serializable fashion), then the teardowns for them will be executed in the reverse order, effectively creating a fixture stack. So because the user was created _before_ the driver was launched, the user would be deleted _after_ the driver was quit.

If a hook throws an error during its setup logic, it will not be added the the relevant `after`/`afterEach` array, and as a result, mocha won't try to execute its teardown code, but any hooks that had already been setup successfully will still be torn down.

There's also the bonus of not having to worry about the order of the `after`/`afterEach` definitions. for a given layer, which increases test code maintainability/extensibility/reusability.

### Alternate Designs

Callbacks were considered, but they introduced more complexity than the `yield` option, and, depending on the implementation, could mean that setup error still wouldn't prevent teardown execution. 

It would apply to any use case where a teardown is needed, provides increased safety and simplicity for mocha users in a standard way (there is precedent in pytest), and requires the changes be made to the core logic of mocha's runner. Their use is also completely optional.

### Benefits

Increased teardown simplicity and safety, and test maintainability/extensibility/reusability.

### Possible drawbacks

None that I can identify.

It works interchangeably with other hook types. The only quirk is that the yield hook is always appended to the _end_ of the relevant teardown array, so it would always be executed after the relevant `after`/`afterEach` definitions that already exist in that describe layer.

### Applicable issues

This would be purely an enhancement with no backwards compatibility issues. One thing I haven't added yet is documentation, but I wanted to ask where it would be best to put it.

[Here's the related issue.](https://github.com/mochajs/mocha/issues/4368)

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -32,6 +32,11 @@ module.exports = Runnable;
 function Runnable(title, fn) {
   this.title = title;
   this.fn = fn;
+  if (this.isGenerator()) {
+    if (!this.yieldsOnce()) {
+      throw new Error('Yield fixture must yield exactly once.');
+    }
+  }
   this.body = (fn || '').toString();
   this.async = fn && fn.length;
   this.sync = !this.async;
@@ -55,6 +60,9 @@ Runnable.prototype.reset = function() {
   this.pending = false;
   delete this.state;
   delete this.err;
+  delete this.iterable;
+  delete this.teardownAdded;
+  delete this.readyForReset;
 };
 
 /**
@@ -133,6 +141,46 @@ Runnable.prototype.slow = function(ms) {
 Runnable.prototype.skip = function() {
   this.pending = true;
   throw new Pending('sync skip; aborting execution');
+};
+
+/**
+ * Check if this runnable is a generator function.
+ * @return {boolean}
+ * @private
+ */
+Runnable.prototype.isGenerator = function() {
+  if (!this.fn) {
+    return false;
+  }
+  return ['GeneratorFunction', 'AsyncGeneratorFunction'].includes(
+    this.fn.constructor.name
+  );
+};
+
+/**
+ * Checks if the generator function only yields once. A fixture yielding more
+ * than once will result in code that never runs.
+ * @return {boolean}
+ * @private
+ */
+Runnable.prototype.yieldsOnce = function() {
+  if (!this.isGenerator) {
+    throw new TypeError("Non-generator functions can't yield.");
+  }
+  var count = 0;
+  var pos = 0;
+  var token = 'yield';
+  var stringifiedFn = this.fn.toString();
+  while (true) {
+    pos = stringifiedFn.indexOf(token, pos);
+    if (pos >= 0) {
+      count++;
+      pos += token.length;
+    } else {
+      break;
+    }
+  }
+  return count === 1;
 };
 
 /**
@@ -327,7 +375,7 @@ Runnable.prototype.run = function(fn) {
     };
 
     try {
-      callFnAsync(this.fn);
+      callFnAsync(this);
     } catch (err) {
       // handles async runnables which actually run synchronously
       errorWasHandled = true;
@@ -343,7 +391,7 @@ Runnable.prototype.run = function(fn) {
 
   // sync or promise-returning
   try {
-    callFn(this.fn);
+    callFn(this);
   } catch (err) {
     errorWasHandled = true;
     if (err instanceof Pending) {
@@ -354,8 +402,21 @@ Runnable.prototype.run = function(fn) {
     done(Runnable.toValueOrError(err));
   }
 
-  function callFn(fn) {
-    var result = fn.call(ctx);
+  function callFn(runnable) {
+    var result;
+    if (!runnable.iterable) {
+      // The runnable should only be called if the runnable isn't a
+      // generator, or it's a generator that hasn't been called yet.
+      result = runnable.fn.call(ctx);
+    }
+    if (runnable.isGenerator()) {
+      // The result of calling the generator would be the iterable object, but
+      // the result of the runnable is still needed. This will provide the
+      // result of both the setup and teardown parts of a generator fixture.
+      runnable.iterable = runnable.iterable || result;
+      result = runnable.iterable.next();
+      runnable.readyForReset = result.done;
+    }
     if (result && typeof result.then === 'function') {
       self.resetTimeout();
       result.then(
@@ -382,8 +443,8 @@ Runnable.prototype.run = function(fn) {
     }
   }
 
-  function callFnAsync(fn) {
-    var result = fn.call(ctx, function(err) {
+  function callFnAsync(runnable) {
+    var callback = function(err) {
       if (err instanceof Error || toString.call(err) === '[object Error]') {
         return done(err);
       }
@@ -404,7 +465,22 @@ Runnable.prototype.run = function(fn) {
       }
 
       done();
-    });
+    };
+    var result;
+    if (!runnable.iterable) {
+      // The runnable should only be called if the runnable isn't a
+      // generator, or it's a generator that hasn't been called yet.
+      result = runnable.fn.call(ctx, callback);
+    }
+    if (runnable.isGenerator()) {
+      // The result of calling the generator would be the iterable object, but
+      // the result of the runnable is still needed. This will provide the
+      // result of both the setup and teardown parts of a generator fixture.
+      runnable.iterable = runnable.iterable || result;
+      result = result.next();
+      runnable.readyForReset = result.done;
+    }
+    // result = runnable.fn.call(ctx, callback);
   }
 };
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -478,6 +478,13 @@ Runner.prototype.hook = function(name, fn) {
       var testError = hook.error();
       if (testError) {
         self.fail(self.test, testError);
+      } else if (hook.isGenerator() && !hook.teardownAdded) {
+        if (hook.parent._beforeAll.includes(hook)) {
+          hook.parent._afterAll.push(hook);
+        } else if (hook.parent._beforeEach.includes(hook)) {
+          hook.parent._afterEach.push(hook);
+        }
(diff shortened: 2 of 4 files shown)


boneskull:
Thanks for this.  Given that generators _aren’t_ widely used, the likelihood that anyone will use this feature is pretty low.  That alone makes it, IMO, a poor candidate for inclusion. 

I would like to consider other implementations. For example, hooks accept an optional `title`. What if we were able to associate two hooks with the same title?  So:

```js
before(‘setup user’, async function() {
  return setupUser();
});

// only run if a sibling “before” hook with the same name does not fail 
after(‘setup user’, async function() {
  return teardownUser();
});
```

This might be a breaking change, since the title currently has no function. It’s magical, as well, so maybe a more explicit API would be preferable. Would something like the above solve your problem?

It’s also possible to write a module that would “wrap” hook functions to do something similar—essentially your `if (something) { teardown() }` example, but generalized using closures. And such a module could be implemented as a custom interface, if desired.

SalmonMode:
Thanks for replying!

Perhaps you can provide a little more background on that for me. Is there a reason generators aren't widely used in JS? That might give me some more insight into why they might not be a solid approach, and possibly what some better alternatives may be. I tried doing some digging but I get the sense that their lack of use may be because they're a fairly new addition to JS.

boneskull:
there are a few reasons, but from what I can tell

1. generators are low-level
1. they were introduced around when native Promise support was added
1. Promises are often more ergonomic and cover many of the same use cases
1. use cases didn’t develop for generators because people mostly just reached for Promises
1. async/await is like a high-level implementation of Promises built on top of generators (I think `await` actually calls `yield` somewhere in there) and is much easier to reason about

while the use-cases are not 1:1, for many applications, consuming Promises or async/await can fulfill the same needs. in your example above, you could use async await to achieve something similar (sorry on phone)

```
let promise

before(function() {
  promise = setupThing()
  return promise
})

// test here

after(async function() {
  await promise
  return teardownThing()
})
```

SalmonMode:
I see. But is the implication for Promises that they will in fact be completed and it's just up to the event loop that determines when they get control back, and with pure yields there isn't such an implication?

I'm still trying to get a grasp on the "meta" so to speak regarding JS development.

boneskull:
> I see. But is the implication for Promises that they will in fact be completed and it's just up to the event loop that determines when they get control back, and with pure yields there isn't such an implication?
I _think_ I understand what you're asking.  That sounds right.  Generators are not _necessarily_ asynchronous.  `Promise`s are always asynchronous.  AFAICT there's very little you could do with an async generator that you couldn't do with `async`/`await` and/or `Promise`s.  The event loop determines when a `Promise` _resolves_.  A task is inserted into the queue, and when the currently-running (blocking) code finishes, then the event loop pops the task from the queue and runs it.

```js
// prints bar, foo
Promise.resolve().then(() => {
  console.log('foo');
});
console.log('bar')

// prints foo, bar, baz
new Promise(resolve => {
  console.log('foo');
  resolve();
}).then(() => {
  console.log('baz');
});
console.log('bar');
```

// prints foo, bar
await console.log('foo');
console.log('bar')
```

A `Promise` is not guaranteed to ever fulfill, inasmuch as you could do this:

```js
new Promise(resolve => {
  setTimeout(resolve, Infinity);
});
```

or this

```js
Promise.resolve().then(() => {
  console.log('done'); // this never prints because we're blocking on the while loop
});
while (true) {
  1;
}
```

boneskull:
In Mocha, those "infinite" examples will only run for as long as the test can before it times out (e.g. `--timeout 2s`).  They will still be in the task queue, however, and will keep the Mocha process running indefinitely unless you use the `--exit` option.

SalmonMode:
@boneskull I've had a thought. What about about something like this?

```javascript
before('some description', () => {
  // do stuff
}).after('possible description', () => {
  // undo the stuff
});

beforeEach('another description', () => {
  // do other stuff
}).afterEach('another possible description', () => {
  // undo the other stuff
});

// test here
```

or even this (a `then`, instead of a variant for each hook type)?

```javascript
before('some description', () => {
  // do stuff
}).then('possible description', () => {
  // undo the stuff
});

beforeEach('another description', () => {
  // do other stuff
}).then('another possible description', () => {
  // undo the other stuff
});

// test here
```

boneskull:
`then` implies a `Promise` return value, so that specific API is off the table. 

tbh with any of this I'd want to be convinced it's a feature that would be used often/widely. I do see its use, but the way it works now hasn't been so onerous that people have been asking for a specific feature around it. there are several ways to DIY in a way that works for a given use-case, and the option to implement a custom interface is always there (I'd encourage you to use the `bdd` interface as a starting point and customize it). it's good to get a different perspective nonetheless.

boneskull:
going to close this for now as I think a custom interface is more appropriate, and this would not be used widely (even though I think it is somewhat useful)

PR #4381, opened 2020-07-22
fix broken template processing in jsdoc output

boneskull:
because the files generated by jsdoc were output to `docs/_site`, they were not processed as liquid templates, which meant the custom reporter tutorial was broken (the included "simple reporter" fixture file was not loaded, and the raw template tag was output instead).

to fix this, we needed to:

1. change the output directory of jsdoc to `docs/api` (just somewhere else)
2. add this to `.gitignore`, obviously
3. tell eleventy to _ignore_ `.gitignore` (just rely on its `.eleventyignore` file)
4. tell eleventy to pass-through all assets underneath `docs/api` (css, images, etc)
5. update the file-loading 11ty data script to output a raw file
6. update the tutorial to use a fenced code block (jsdoc markdown plugin seems to strip html?) and tell Prettier to ignore the liquid template directive

also:

- removed link to google group
- prettier-related reformats
- tweak data script to use `fs.promises`

_(I lost like 30m on the `.gitignore` thing. that's surprising behavior on 11ty's part, IMO)_

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -16,6 +16,9 @@ module.exports = function(eleventyConfig) {
   eleventyConfig.addPassthroughCopy('docs/_headers');
   eleventyConfig.addPassthroughCopy('docs/favicon.ico');
   eleventyConfig.addPassthroughCopy('docs/example');
+  eleventyConfig.addPassthroughCopy('docs/api/images');
+  eleventyConfig.addPassthroughCopy('docs/api/scripts');
+  eleventyConfig.addPassthroughCopy('docs/api/styles');
 
   /* Markdown Plugins */
   const markdown = require('markdown-it')({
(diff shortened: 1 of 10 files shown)


boneskull:
Ref: needs https://github.com/mochajs/mocha-docdash/pull/6

boneskull:
That update to docdash fixes the URLs.  11ty wants to rename all of the files from `filename.html` to `filename/index.html`, but jsdoc's links all use `filename.html`, so all the links were broken (as rightfully caught by the link checker).

The change in docdash should be a major bump since it throws front matter into all of the `.html` files.

boneskull:
I have some "experimental" changes to the workflow here which:

1. Avoids re-downloading a bunch of images if the supporters images are already on disk
1. Does _not_ fail if we cannot download a supporter image
1. Using `npm start docs.dev` (renamed from `docs.watch`--an alias for `eleventy --serve`), 11ty watches and executes `jsdoc` when source files change (and avoids this if they do not change, but e.g., `docs/index.md` _does_)

I'll open a PR for that stuff after this one lands

boneskull:
This isn't urgent, so I can wait a bit longer to merge.  cc @mochajs/core

PR #4383, opened 2020-07-26
Fix #4348

### LINKED ISSUE #4348: Hook only renamed when it fails, not when it begins and ends successfully
### Prerequisites

### Description

A hook's title is only rewritten to include the test's name when it fails.  This means the name is inconsistent between hook start and hook failure, which is messing up the logging of a custom reporter.  Additionally, the hook name is incomplete when it starts, because it does not include the name of the associated test.

Rewriting is done here: https://github.com/mochajs/mocha/blob/master/lib/runner.js#L419-L432

'hook' event: title = `"before each" hook`
'fail' event: title = `"before each" hook for "name of test case"`, originalTitle = `"before each" hook`

### Steps to Reproduce

Run this reproduction: https://github.com/cspotcode/repros/tree/mocha-hook-title-discrepancy

Hook's `title` includes the `from "name of test case"` for *all* events.

Hook title does not include `from "name of test case"` suffix in 'hook' event and never includes if it succeeds.

all the time

### Versions

- `mocha --version`: N/A, not installed
- `node node_modules/.bin/mocha --version`: 8.0.1
- The output of `node --version`: v14.2.0
- Your operating system
  - name and version: Windows 10, WSL2, running Ubuntu 16.04.6 LTS
  - architecture (32 or 64-bit): 64bit
- Any third-party Mocha-related modules (and their versions): custom reporter which listens to `hook`, `hook end`, and `fail` events

cspotcode:
### Requirements

### Description of the Change

Rewrites a hook's title always, instead of only when the hook fails.  Fixes issues described in #4348.

### Alternate Designs

None

It's a bug as described in #4348

### Benefits

Bug is fixed

### Possible Drawbacks

Any change is potentially breaking, even a bugfix.

### Applicable issues

#4348

* Is this a breaking change (major release)? no
* Is it an enhancement (minor release)? no
* Is it a bug fix, or does it not impact production code (patch release)? yes

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -419,20 +419,6 @@ Runner.prototype.fail = function(test, err, force) {
  * @param {Error} err
  */
 Runner.prototype.failHook = function(hook, err) {
-  hook.originalTitle = hook.originalTitle || hook.title;
-  if (hook.ctx && hook.ctx.currentTest) {
-    hook.title =
-      hook.originalTitle + ' for ' + dQuote(hook.ctx.currentTest.title);
-  } else {
-    var parentTitle;
-    if (hook.parent.title) {
-      parentTitle = hook.parent.title;
-    } else {
-      parentTitle = hook.parent.root ? '{root}' : '';
-    }
-    hook.title = hook.originalTitle + ' in ' + dQuote(parentTitle);
-  }
-
   this.fail(hook, err);
 };
 
@@ -455,6 +441,20 @@ Runner.prototype.hook = function(name, fn) {
       return fn();
     }
     self.currentRunnable = hook;
+
+    hook.originalTitle = hook.originalTitle || hook.title;
+    if (hook.ctx && hook.ctx.currentTest) {
+      hook.title =
+        hook.originalTitle + ' for ' + dQuote(hook.ctx.currentTest.title);
+    } else {
+      var parentTitle;
+      if (hook.parent.title) {
+        parentTitle = hook.parent.title;
+      } else {
+        parentTitle = hook.parent.root ? '{root}' : '';
+      }
+      hook.title = hook.originalTitle + ' in ' + dQuote(parentTitle);
+    }
 
     if (name === HOOK_TYPE_BEFORE_ALL) {
       hook.ctx.currentTest = hook.parent.tests[0];
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -464,8 +464,6 @@ Runner.prototype.hook = function(name, fn) {
       hook.title = hook.originalTitle + ' in ' + dQuote(parentTitle);
     }
 
-    console.log([hook.title, hook.originalTitle]);
-
     hook.allowUncaught = self.allowUncaught;
 
     self.emit(constants.EVENT_HOOK_BEGIN, hook);
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -442,6 +442,14 @@ Runner.prototype.hook = function(name, fn) {
     }
     self.currentRunnable = hook;
 
+    if (name === HOOK_TYPE_BEFORE_ALL) {
+      hook.ctx.currentTest = hook.parent.tests[0];
+    } else if (name === HOOK_TYPE_AFTER_ALL) {
+      hook.ctx.currentTest = hook.parent.tests[hook.parent.tests.length - 1];
+    } else {
+      hook.ctx.currentTest = self.test;
+    }
+
     hook.originalTitle = hook.originalTitle || hook.title;
     if (hook.ctx && hook.ctx.currentTest) {
       hook.title =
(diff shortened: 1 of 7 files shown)


cspotcode:
@boneskull this is ready for review.

lib/runner.js
```diff
@@ -514,8 +502,25 @@ Runner.prototype.hook = function(name, fn) {
       }
       self.emit(constants.EVENT_HOOK_END, hook);
       delete hook.ctx.currentTest;
+      setHookTitle(hook);
```
juergba, line 505: What is this for?

cspotcode: When hook.ctx.currentTest is deleted, this changes the hook's title so that it no longer mentions the test case and only mentions the test suite.  It affects the "multiple done calls" errors.

cspotcode:
@juergba it sounds like you're saying that the title rewriting, which mocha is already doing, is bad.  Do you want to submit an alternative PR which fixes that?  Do we know why the current rewriting behavior was implemented?

juergba:
No, not bad. When a hook fails, normally no more related tests are executed. E.g. when a `beforeAll` hook fails, all corresponding tests "disappear".  Renaming such a failing hook is a different thing than renaming all hooks. It's easier to identify the failing hook in the reporter output like this.

Actually I don't like how failing `beforeAll` and `afterAll` hooks are renamed, based on `hook.ctx.currentTest`.
Setting `if (name === HOOK_TYPE_BEFORE_ALL) { hook.ctx.currentTest = hook.parent.tests[0]; } ...` is not really correct. The context should be the suite not the first/last test. But this topic is out of scope and should be checked carefully before changing anything.

cspotcode:
Right, so the current title rewriting is bad and should be changed.  What's
the right approach for teams affected by #4383?

test/integration/multiple-done.spec.js
```diff
@@ -97,7 +97,7 @@ describe('multiple calls to done()', function() {
       expect(res.failures[0], 'to satisfy', {
         fullTitle: 'suite "before all" hook in "suite"',
         err: {
-          message: /done\(\) called multiple times in hook <suite "before all" hook> of file.+multiple-done-before\.fixture\.js/
+          message: /done\(\) called multiple times in hook <suite "before all" hook in "suite"> of file.+multiple-done-before\.fixture\.js/
```
boneskull, line 100: this doesn't read quite right... `suite "before all" hook in "suite"`

test/integration/multiple-done.spec.js
```diff
@@ -119,20 +119,18 @@ describe('multiple calls to done()', function() {
...
-          fullTitle: 'suite "before each" hook in "suite"',
-          err: {
-            message: /done\(\) called multiple times in hook <suite "before each" hook> of file.+multiple-done-before-each\.fixture\.js/
-          }
+      expect(res.failures[0], 'to equal', res.failures[1]);
+      expect(res.failures[0], 'to satisfy', {
+        fullTitle: 'suite "before each" hook in "suite"',
+        err: {
+          message: /done\(\) called multiple times in hook <suite "before each" hook in "suite"> of file.+multiple-done-before-each\.fixture\.js/,
+          multiple: [
```
boneskull, line 127: this error object is weird, but I guess that's what it does.

you can combine this into a single assertion:

```js
expect(res.failures[0], 'to equal', res.failures[1])
  .and('to satisfy', {
    // etc.
  });
```

lib/runner.js
```diff
@@ -419,20 +419,6 @@ Runner.prototype.fail = function(test, err, force) {
  * @param {Error} err
  */
 Runner.prototype.failHook = function(hook, err) {
```
boneskull, line 421: could probably just remove `failHook` and replace its usage with calls directly to `fail`

test/integration/multiple-done.spec.js
boneskull: actually, it appears `"suite"` is the _name_ of the suite, which is confusing, so maybe it's not a blocker at all.

test/integration/multiple-done.spec.js
boneskull: (likewise the assertion(s) in the new test may be able to be combined this way)

test/integration/multiple-done.spec.js
cspotcode: Yeah, tripped me up, too.

test/integration/multiple-done.spec.js
cspotcode: Yeah, you probably already know this, but I rewrote the assertion to point out that both items of the array actually refer to the same object.  I can't do a `===` check in the tests, because it's been converted to/from JSON, so the assertions are running against 2 different parsed copies of the same object.  Does `'to equal'` handle that?

This happens because when a single test/hook has multiple failures, the first error is reused, and the `multiple` array is created to store the others.  Ideally I think mocha should be creating a new "hookinvocation" instance for each time the hook runs, where each "hookinvocation" can store a reference to the test case it's paired to.  But I don't think I'll have time for that size of code change.

lib/runner.js
cspotcode: Will do.

cspotcode:
@boneskull thanks, I think I addressed all your feedback and tests are passing.

PR #4389, opened 2020-07-30
Refactoring: Consuming log-symbols alternate to code for win32 in reporters/base.

MoonSupport:
### Description of the Change

I used the `log-symbols` instead of the code for win32.

### Alternate Designs

No alternate designs

### Benefits

Mocha already uses `log-symbols`, so we can remove duplicated code and maintain consistency.

### Possible Drawbacks

I also deleted the `dot` case, But I don't know if the size difference between dot for win32(`.`) and dot for the others(`․`) is a problem.

So I think this might be a problem.

### Applicable issues

https://github.com/mochajs/mocha/issues/4353

The above issue is not directly related to the PR.

But I think you may want to use alphabets rather than `log-symbols` to resolve the issue.

like this

```
success : v or o
fail : x
```

Please leave a good comment or question at any time :)

* Is this a breaking change (major release)? no
* Is it an enhancement (minor release)?  maybe, yes
* Is it a bug fix, or does it not impact production code (patch release)? no

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -10,6 +10,7 @@ var diff = require('diff');
 var milliseconds = require('ms');
 var utils = require('../utils');
 var supportsColor = require('supports-color');
+var symbols = require('log-symbols');
 var constants = require('../runner').constants;
 var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
 var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
@@ -88,19 +89,12 @@ exports.colors = {
  */
 
 exports.symbols = {
-  ok: '✓',
-  err: '✖',
-  dot: '․',
+  ok: symbols.success,
+  err: symbols.err,
+  dot: '.',
   comma: ',',
   bang: '!'
 };
-
-// With node.js on Windows: use symbols available in terminal default fonts
-if (process.platform === 'win32') {
-  exports.symbols.ok = '\u221A';
-  exports.symbols.err = '\u00D7';
-  exports.symbols.dot = '.';
-}
 
 /**
  * Color `str` with the given `type`,


MoonSupport:
I did it

outsideris:
I've re-run AppVeyor and it passed.

juergba:
@MoonSupport could you rebase to `master`, please? Thanks.
If all tests have passed, I will merge this PR.

MoonSupport:
@juergba Done :)

juergba (approved):
@MoonSupport thank you for this PR.

jimmywarting:
Currently looking into all the dependencies... and one of your dependencies are [log-symbol](https://npmgraph.js.org/?q=log-symbols@4.1.0)

log-symbols seems like a perfect package if it had just only precomputed the values instead of [using chalk](https://github.com/sindresorhus/log-symbols/blob/main/index.js) as an dependency 
the size could have been a lot smaller if it was just written with:

```js
import isUnicodeSupported from 'is-unicode-supported';

const main = {
  info: '\x1B[34mℹ\x1B[39m',
  success: '\x1B[32m✔\x1B[39m',
  warning: '\x1B[33m⚠\x1B[39m',
  error: '\x1B[31m✖\x1B[39m'
};

const fallback = {
  info: '\x1B[34mi\x1B[39m',
  success: '\x1B[32m√\x1B[39m',
  warning: '\x1B[33m‼\x1B[39m',
  error: '\x1B[31m×\x1B[39m'
};

const logSymbols = isUnicodeSupported() ? main : fallback;

export default logSymbols;
```

Here is one tough: how about inlining this symbols into mocha itself?

juergba:
@jimmywarting thank you. The size of the code is not the only criteria.

All those codes/numbers (unicode?) isn't the detail level I want to deal with. We need a package which works for different OS/browsers and is well maintained by someone else, not by Mocha.
Time is the main criteria when maintaining a package like Mocha. So inlining those symbols isn't a good idea.

PR #4396, opened 2020-08-06
Update dependencies & dev dependencies

GChuf:
### Requirements

### Description of the Change

Updates dependencies and dev dependencies, fixes some vulnerabilities.

### Benefits

300+ less vulnerabilities & less dependencies (node_modules folder decreases in size from 415MB to 394MB and has 2000 less files)


boneskull:
going to need to run a branch check on this one, lemme see

boneskull:
build is here: https://travis-ci.org/github/mochajs/mocha/builds/718013127, branch `boneskull/pr/GChuf-dep2`

GChuf:
Travis looks okay, what's up with appveyor not finding package.lock file? re-run tests?

boneskull:
appveyor's kind of hinky. looking to move to GH actions for that stuff, see #4402. should be fine to merge this otherwise

boneskull:
I'm seeing this during Rollup build:

```
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
node_modules\escape-string-regexp\index.js: (7:12)
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
node_modules\escape-string-regexp\index.js
5: import { newArrowCheck as _newArrowCheck } from "\0rollupPluginBabelHelpers.js";
6: 
7: var _this = this;
               ^
8: 
9: var escapeStringRegexp = function escapeStringRegexp(string) {
...and 1 other occurrence
created ./mocha.js in 10.1s
```

which is concerning, especially since no tests fail.

When you ran these upgrades, did you use `npm audit fix --force` or just upgrade everything?  Did `escape-string-regexp` need upgrading?

GChuf:
I just upgraded escape-string-regexp to `@latest`. I didn't do it out of any need, so you can try downgrading to v3 or v2, see what happens. Worst case scenario it goes back to v1.0.5, I don't think it makes much difference anyway ...
I ran `npm audit fix` without `--force`, so that shouldn't be a problem.

boneskull:
I think that error might be due to the `spec: true` in the rollup config.  I'll see if removing it addresses the problem

boneskull:
okay, that seemed to work.

boneskull:
thanks!

PR #4397, opened 2020-08-06
Update esm-utils.js

davidluanrsantos:
SyntaxError: Unexpected token import

### Requirements

### DIFF
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -3,9 +3,9 @@ const url = require('url');
 
 const formattedImport = async file => {
   if (path.isAbsolute(file)) {
-    return import(url.pathToFileURL(file));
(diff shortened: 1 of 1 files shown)


davidluanrsantos:
On my machine the command mocha return "Unexpected token error" because the "import" is unsupported, to resolve this problem i need change "import" signature for "require"

boneskull:
@davidluanrsantos Hi, thanks for contributing.  It sounds like the problem is likely a misconfiguration on your end, but since it's not clear how you're trying to use `esm-utils.js`, we can't really help you further.  There is no bare `import` statement in `esm-utils.js`, which would be the source of such a syntax error.  Also, please be sure you're using Node.js v10 or newer.

PR #4399, opened 2020-08-08
fix(base): if message and stacktrace overlap remove duplicates

thiporia:
### Requirements

### Description of the Change

The problem was that the message of the issue contained a stacktrace.
There is a logic to remove the existing duplicates.
The stacktrace received as a message and the internally created stacktrace are not the same.
We can consider how to make the shape of the internally generated stack trace and the trace passed to the message the same.
So I used utils.stackTraceFilter.
Through stackTraceFilter, we have worked to secure the efficiency of the existing deduplication logic.

### Alternate Designs

If possible, it would be nice to consider ignoring the stacktrace if there is a stacktrace in the error message.

if message and stackTrace overlap, remove duplicate message or stacktrace.

### Benefits

You can get a cleaner error report.

### Possible Drawbacks

I don't think it will be a problem because all of the existing tests have passed.

### Applicable issues

#3992

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -15,7 +15,7 @@ var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
 var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
 
 const isBrowser = utils.isBrowser();
-var stackFilter = utils.stackTraceFilter();
+const stackFilter = utils.stackTraceFilter();
 
 function getBrowserWindowSize() {
   if ('innerHeight' in global) {
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -15,6 +15,7 @@ var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
 var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
 
 const isBrowser = utils.isBrowser();
+var stackFilter = utils.stackTraceFilter();
 
 function getBrowserWindowSize() {
   if ('innerHeight' in global) {
@@ -249,6 +250,8 @@ exports.list = function(failures) {
     } else {
       message = '';
     }
+
+    message = test.fullStackTrace ? message : stackFilter(message);
     var stack = err.stack || message;
     var index = message ? stack.indexOf(message) : -1;
 
--- a/test/reporters/base.spec.js
+++ b/test/reporters/base.spec.js
@@ -444,7 +444,7 @@ describe('Base reporter', function() {
     });
   });
 
-  describe.only('When there is StackTrace in the error message', function() {
+  describe('When there is StackTrace in the error message', function() {
     var baseConsoleLog;
     var err;
     var test;
--- a/test/reporters/base.spec.js
+++ b/test/reporters/base.spec.js
@@ -443,4 +443,40 @@ describe('Base reporter', function() {
       sinon.restore();
     });
   });
+
+  describe.only('When there is StackTrace in the error message', function() {
+    var baseConsoleLog;
+    var err;
+    var test;
+
+    beforeEach(function() {
+      baseConsoleLog = sinon.spy(Base, 'consoleLog');
+      err = {
+        message:
+          'AssertionError: foo bar\n' +
+          'at EventEmitter.<anonymous> (/usr/local/dev/test.js:16:12)',
+        stack:
+          'AssertionError: foo bar\n' +
+          'at EventEmitter.<anonymous> (/usr/local/dev/test.js:16:12)\n' +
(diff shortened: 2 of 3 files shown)


test/reporters/base.spec.js
```diff
@@ -443,4 +443,40 @@ describe('Base reporter', function() {
       sinon.restore();
     });
   });
+
+  describe('When there is StackTrace in the error message', function() {
+    var baseConsoleLog;
```
boneskull, line 448: You can use `let` instead of `var` here now

test/reporters/base.spec.js
```diff
@@ -443,4 +443,40 @@ describe('Base reporter', function() {
...
+    it('You should be able to remove the overlap between the message and the stacktrace.', function() {
+      var expected = [
+        '  %s) %s:\n     %s\n%s\n',
+        1,
+        'test title',
+        'AssertionError: foo bar\nat EventEmitter.<anonymous> (/usr/local/dev/test.js:16:12)',
+        '  at Context.<anonymous> (/usr/local/dev/test.js:19:5)'
+      ];
+      Base.list([test]);
+      assert.deepEqual(baseConsoleLog.lastCall.args, expected);
```
boneskull, line 475: we want to use [unexpected](https://unexpected.js.org) for assertions, so you'd rewrite this as

```js
expect(Base.consoleLog, 'to have a call satisfying', expected);
```

You don't need to actually use the `baseConsoleLog` variable; you can reference `Base.consoleLog` instead, because Sinon replaces the function with its spy.

test/reporters/base.spec.js
```diff
@@ -443,4 +443,40 @@ describe('Base reporter', function() {
...
+    });
+
+    afterEach(function() {
+      baseConsoleLog.restore();
```
boneskull, line 479: This shouldn't be necessary, as `sinon.restore()` is called after each test on L48

lib/reporters/base.js
```diff
@@ -15,6 +15,7 @@ var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
 var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
 
 const isBrowser = utils.isBrowser();
+var stackFilter = utils.stackTraceFilter();
```
boneskull, line 18: can use `const` here

test/reporters/base.spec.js
thiporia: i got it!

test/reporters/base.spec.js
thiporia: thanks!
i'll change it.

test/reporters/base.spec.js
thiporia: i got it :)

lib/reporters/base.js
thiporia: thanks!

thiporia:
First of all, thank 'boneskull' for taking the time to review.

If you test the code below without applying the `--full-trace` option,
The stacktrace is duplicated as in the image below.
```
function oupsStackStack() {
  const err = new Error();
  throw new Error(`An error occured with following trace:\n\n${err.stack}`);
}
describe("describe", () => {
  it("it oupsStackStack", () => {
    oupsStackStack();
  });
});
```
스크린샷 2020-08-19 21 01 24

The logic of mocha to remove duplicates already exists.
However, it only worked when `--full-trace` was set.
So, I added logic to remove duplicates even if the `--full-trace` option does not exist.

스크린샷 2020-08-19 22 05 56

boneskull:
thanks.  I'm going to need to take a closer look at the current behavior and try to determine what was originally intended.

boneskull:
OK, thanks anyway... but I don't think Mocha should modify `err.message`.  duplicating the stack in this manner is unusual, and I would expect that there's another problem you're trying to solve (source maps?).  please let me know what you're trying to do, and perhaps I can offer an alternative approach.

dubzzz:
@boneskull I believe the PR was related to an issue I opened a while ago https://github.com/mochajs/mocha/issues/3992 in Mocha.

_Here is the context of the issue I opened:_

I am the author of a pbt framework - _property based testing_ - called fast-check [[npm](https://www.npmjs.com/package/fast-check)][[github](https://github.com/dubzzz/fast-check)]. This framework is test runner agnostic, it can be used within Mocha but also Tape, Jest or even JavaScript without any test runner.

In pbt, users want to make sure that a given property (eg.: a + b === b + a for any integers) holds on the whole scope of inputs. To do that fast-check executes the property of the user against various inputs and whenever one of the combinations fail (either throw or just invalid output), it reports it. In case the property throws an eror for this combination the behaviour of the library is to `throw new Error('property failed with input ' + stringify(input) + '\n\n' + err.stack)`.

Given how we wrap the original error, we get a duplicated stack-trace in Mocha.

PR #4401, opened 2020-08-08
fix: add Mocha to global variable

irrationnelle:
### Requirements

### Description of the Change

To apply UMD pattern, Mocha module was located in `mocha` as module name. But browsers are be able to access Mocha as global variable with this change.

I added *rollup config* `footer` to assign Mocha module to global variable. I hope that this would be the alias to acutal Mocha module,  not deep cloning another Mocha.

### Alternate Designs

in `browser-entry.js` from line `213`

```javascript
mocha.Mocha = Mocha;
mocha.mocha = mocha;
```

I would like to suggest like next

```javascript
global.Mocha = Mocha;
global.mocha = mocha;
```

OR,

```javascript
mocha.Mocha = Mocha;
mocha.mocha = mocha;

(function(currentGlobal) {
  currentGlobal.Mocha = mocha.Mocha;
})(global);
```

anyway, I was afraid of any side-effects and then I selected this version as PR.

### Benefits

- Users who used mocha for browsers could update the latest mocha
- Other third party modules with global variable Mocha could update the latest mocha

### Possible Drawbacks

There would be duplicated Mocha modules on browser.
IMHO, an alias for this module would be better than assign object but I would like to avoid to modify core code.

### Applicable issues

#4395
 
### Versioning Info

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -17,7 +17,8 @@ const config = {
     file: './mocha.js',
     format: 'umd',
     sourcemap: true,
-    name: 'mocha'
+    name: 'mocha',
+    footer: 'window.Mocha = window.mocha.Mocha;'
   },
   plugins: [
     json(),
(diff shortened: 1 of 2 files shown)


boneskull:
LGTM, thanks

boneskull:
maybe putting this in `browser-entry.js` would be better, but we can change it later.

Issue #4117, opened 2019-12-03
🛠️ Repo: Switch automations to GitHub Actions (GHA)


craigtaub:
Can we make some of the manual processes which are time consuming (and limited to those with an OS which can run the required tasks) automated via [Githubs actions](https://github.com/features/actions)?

Look at:
1. NPM release
2. Documentation website release (See https://github.com/netlify/actions)

Note only mocha repository collaborators will have permission create workflows.

boneskull:
release might be problematic. we should be using npm's 2FA, which--last time I checked--means a machine cannot publish.  I'll poke around and see what that looks like.

tbh, I'm interested in moving the entire build to GH actions.  I've been using it in my other projects, and it's nice & fast.  not sure what the SauceLabs integration looks like though.

craigtaub:
> I'll poke around and see what that looks like.

Thanks man. Let's break it down and see where the issues are.

outsideris:
We can use Actions to trigger Netlify( #3644 ) and replace AppVeyor for windows testing ( #4013 )

juergba:
Regarding publishing and 2FA: could [Wombat](https://github.com/GoogleCloudPlatform/wombat-dressing-room) evtl. be a solution?

cspotcode:
Moving to Github Actions means running everything in docker, which conveniently means a consistent environment for running stuff locally, too.  So e.g. folks on Windows can publish as long as they have docker installed.

For publishing, you can tell npm to publish a pre-built tarball
https://docs.npmjs.com/cli/publish
This means Github Actions could do the entire build, version bump, prepare the release notes, push git tags, then post the tarball for a maintainer to publish.  The maintainer would run a (dockerized) publication script on their machine that downloads the tarball from Github and runs `npm publish next-version.tar.gz`

juergba:
@cspotcode Interesting.
On my Windows I have WSL2.0 / Ubuntu / VSCode WSL-extension installed. As an alternative to your docker solution, I just need to install npm on Ubuntu side, then I could run `npm publish downloaded-tarball.tar.gz`. Right?

cspotcode:
Even on Windows without WSL, that should work correctly.

ssbarnea:
How active is mocha development? I see an interesting https://github.com/mochajs/mocha/pull/4822 which is still waiting for the approval to run the tests for 3 days.

msimerson:
> we should be using npm's 2FA, which--last time I checked--means a machine cannot publish.

[...]

PR #4391, opened 2020-07-30
cache Growl installer on AppVeyor

refack:
# Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions.

### Applicable issues

Fixes https://github.com/mochajs/mocha/issues/4390

PR #4402, opened 2020-08-09
ci(win): setup GH actions for windows CI

### LINKED ISSUE #4390: cache growl on appveyor
To run growl-related tests on windows, we install growl:

```yml
  ## Manual Growl install
  - ps: Add-AppveyorMessage "Installing Growl..."
  - ps: $seaURL = "https://github.com/briandunnington/growl-for-windows/releases/download/final/GrowlInstaller.exe"
  - ps: $seaPath = "$($env:USERPROFILE)\GrowlInstaller.exe"
  - ps: $webclient = New-Object Net.WebClient
  - ps: $webclient.DownloadFile($seaURL, $seaPath)
```

I think github's rate-limiter may not appreciate this sort of thing, and occasionally the download will fail.  We shouldn't need to download this installer on every build; we can cache the installer, and only download if it is not present in the cache.

I'd do this myself, but I don't know enough (read: any) PowerShell to skip the download step if the file exists, _and_ I don't know if the cached files are loaded before this build step (but it seems like they should be).

So, hopefully someone who knows a bit of PS can help us out.  The `$seaPath` value is what we want to cache, and [here's the relevant AppVeyor documentation](https://www.appveyor.com/docs/build-cache/#configuring-cache-items).

outsideris:
### Description of the Change
Replace AppVeyor with GitHub Actions for testing on Windows. 
Because windows builds were failed on AppVeyor, I believe it is the platform issue, not ours, and AppVeyor introduced new Node versions sometimes, I set GitHub Actions for Windows CI. 

You can see [build results](https://github.com/outsideris/mocha/actions?query=workflow%3A%22Windows+CI%22) on my repository.

I tested [actions/cache](https://github.com/actions/cache), but it spent more time than without cache since downloading cache files.

### Benefits
We can have robust windows build results than before.

### Possible Drawbacks
* I set it up on [Windows Server 2019](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md). Do we need to set it up on [Windows Server 2016](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2016-Readme.md) as well?
* I didn't install [Growl for Windows](https://github.com/mochajs/mocha/wiki/Growl-Notifications#windows) because all test already passed.
* I added smoke tests and linting on Windows. I thought we need to do it even we run it on Travis. If you think it doesn't need on Windows, let me know.
* GitHub provides 2,000 build minutes. [Our Travis used 2,200 build mins last month](https://travis-ci.org/github/mochajs?tab=insights) and Travis build time is about 2 times than GitHub Actions Windows build. So, I think we can use GitHub Actions freely, but we will pay for Github Actions in the future.

### Applicable issues
#4117
Close #4390

I want to hear feedback from @mochajs/core .


outsideris:
> Whats the value in adding these on Windows? Just thinking that it adds time and expense. Wondering if we have Windows users who get passing CI builds only to fail on release.

The purpose of smoke tests is to confirm that mocha will work when we release it. So, I thought it is better if we run smoke tests on Windows as well. 

Linting is less useful than smoke tests. I'm not sure that failure to lint on windows while lint passed on Linux is possible. Just I want to check it for our windows contributors.

I can remove them from it if we agree it is useless.

> Any idea if there is a performance difference against AppVeyor?

AppVeyor took about 22 mins and GitHub Actions took about 12 mins even though GitHub Actions have more tasks.
IMO, AppVeyor is free for us.

appveyor.yml
```diff
@@ -1,76 +0,0 @@
-###
-### appveyor.yml
-###
-
-## General configuration
-version: '{build}'
-skip_commits:
-  message: /\[ci\s+skip\]/
```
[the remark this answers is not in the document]
outsideris: I will check it.

craigtaub:
> AppVeyor took about 22 mins and GitHub Actions took about 12 mins even though GitHub Actions have more tasks.

🤩 🚀 

> So, I thought it is better if we run smoke tests on Windows as well.

I suppose with perf high and cost low its worth it. Linting I agree probably not necessary.

boneskull:
My only requirement is that it runs on Windows 10 and PowerShell. It should not be WSL or "Git Bash" or whatever. 

If that's the case then I am 100000000% interested in merging this. 🥳

outsideris:
I'm trying to add a feature that skip CI with `[ci skip]`. Skipping CI based on branches is easy, but based on PRs is complex because GitHub doesn't provide the latest commit message on their context when triggered by PRs. 

I will back after some test for that.

outsideris:
@boneskull I'm not familiar with Windows as development environment. 
GitHub Actions only supports [`Windows Server 2016` and `Windows Server 2019`](https://github.com/actions/virtual-environments/tree/win19/20200802.1) and Windos Server 2019 looks similar with Windows 10. 
And [Powershell is the default shell](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell) for Windows in GitHub Actions.

> powershell: This is the default shell used on Windows. The Desktop PowerShell.

outsideris:
Done. 

Now, it will be skipped with `[ci skip]` commit message.  You can check it in my repository for [push](https://github.com/outsideris/mocha/actions/runs/211818656) and [pull_request](https://github.com/outsideris/mocha/actions/runs/211820145). 

`Retrive head commit message` job always runs to get a head commit message when action triggered by pull request.

스크린샷 2020-08-17 오후 6 33 41

And I removed lint job from Windows CI.

Munter:
I really like this addition. I just double checked out settings. Looks like we have 3000 minutes per month, apparently from a free open source team plan which must have been set up at some point.

We can also limit spending to $0 if we are afraid of running up a bill at some point. That would require someone to put down a credit card though, and I assume we don't have a cards associated with our opencollective profile, so it's probably an individual that needs to do this

boneskull:
I can put my feelers out and see if we can get more minutes if we start running into the limit.

boneskull:
sounds good @outsideris. travis has windows builds but they aren't powershell envs afaik. would be interested in moving everything to GHA eventually

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Cache Growl Installer
+        id: cache-growl
+        uses: actions/cache@v2
+        with:
+          path: GrowlInstaller
+          key: ${{ runner.os }}-growl-installer
+          restore-keys: |
+            ${{ runner.os }}-growl-installer
+      - name: Download Growl Installer
+        if: steps.cache-growl.outputs.cache-hit != 'true'
```
outsideris, line 52: dowlnload Growl Installer only when the cache is not hit,

.github/workflows/windows-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+      - run: npm ci --ignore-scripts
+      - name: npm start test.node
+        run: |
+          echo "Starting Growl service..."
+          Start-Process -NoNewWindow Growl
+          ## Growl requires some time before it's ready to handle notifications
+          echo "Verifying Growl responding"
+          Start-Sleep -s 5
+          growlnotify test
+          echo "Staring test"
```
outsideris, line 94: Start and verify growl before running the test because GH actions don't keep growl process previous steps.

outsideris: 스크린샷 2020-09-05 오후 7 03 48

Sometimes, there are `DllNotFoundException`, not always. But we can verify Growl responding regardless the exception.

outsideris:
I added Growl installer and cache it which mentioned #4391

outsideris:
It is ready to merge.

outsideris:
I will merge this in a few days.

Munter:
Great work on this! Do good to have both speed increases and better access for everyone to see the results

PR #4407, opened 2020-08-13
fix: fully support 'extends' for configurations

mf-pherlihy:
### Requirements

### Description of the Change

There have been multiple issues reported where configurations do not actually support the `extends` property:

https://github.com/mochajs/mocha/issues/3916
https://github.com/mochajs/mocha/pull/4057

I did some digging and found the root cause to be the `options.parse()` method. This method relies on `yargs-parser` and NOT `yargs` which does the actual configuration extending.

It was a pretty straightforward fix. I imported the `yargs` function that extends the configuration objects and then applied it to all the arguments passed to `parse()`.

I added a test to support this case and verified the test fails when this fix is not applied.

### Alternate Designs

There was already an existing PR open that modified external methods to do this, so I wanted to avoid that.

It fixes a long standing bug.

### Benefits

Configuration files can now be extended properly.

### Possible Drawbacks

I dont see any?


boneskull:
Yeah, I think that test should probably live in `test/node-unit/cli/options.spec.js` instead.

bcoe:
>I'm a little concerned that yargs/lib/apply-extends isn't a public API. If it isn't, maybe we can vendor it (e.g., just copy/paste it into a function in lib/cli/options.js). @bcoe can you comment?

I would be happy to make this a public API, but as this PR stands today, it will be broken by the next major version of yargs,  which moves to using an export map.

I'd be happy  _to_ export this helper though.

boneskull:
@bcoe If you want to add it to the export map, that'd work for us--we can change our code to use it when we upgrade.

boneskull:
@mf-pherlihy Are you interested in doing further work on this? I'd like to merge the PR, but the test should probably move.  If I don't hear back in a couple days, I'll merge this as-is and fix it after.

mf-pherlihy:
@boneskull Yeah I can finish this up. Just to confirm:
1. Move to `test/node-unit/cli/options.spec.js`
2a. Copy `applyExtends` from yargs to here with a TODO to swap out once yargs is updated?
2b. Use `applyExtends` as-is with a TODO to swap out once yargs is updated?

boneskull:
1. Move the test
2. AFAICT you are free to use `applyExtends` as it [has been made part of yargs' public API](https://github.com/yargs/yargs/pull/1733).  But it sounds like when we upgrade yargs we will need to change it?  Unclear, since we're pulling it out of `yargs-parser` and not `yargs` itself, right?  @bcoe?

boneskull:
ping @bcoe @mf-pherlihy

bcoe:
> AFAICT you are free to use applyExtends

Yes, my intention was that it's a blessed public helper now. The caveat is that [we need to fix this](https://github.com/yargs/yargs/issues/1789).

mf-pherlihy:
Just tried upgrading yargs and using the defined helpers. Unfortunately however, the latest version breaks a bunch of other tests.

bcoe:
@mf-pherlihy yargs@16.x.x introduces ESM, and was a somewhat major shuffling around of code. I'd be interested in making sure we address any issues you're bumping into with tests.

juergba:
@bcoe I did upgrade yargs and yargs-parser with #4543. There have been changes in yargs' array handling with `yargs.config()` and combinig array values. I had to remove aliased options from our pre-parsed result (by yargs-parser) in order to prevent double entries.
I do not consider the new behavior as a bug, it seems more consistent now than before.

bcoe:
> I do not consider the new behavior as a bug, it seems more consistent now than before.

@juergba that's good to hear 👍 

Sound like we perhaps just need to rebase this and give the helpers another try?

juergba:
> Sound like we perhaps just need to rebase this and give the helpers another try?

I guess so, but I haven't taken a look at this PR though. But please go ahead @mf-pherlihy.
The browser tests will fail anyway, but that's off topic. Github actions does not have access to GH secrets (login data for SauceLabs) on PR's from forked repos. We haven't solved that, yet.

juergba:
@mf-pherlihy I rebased and tests are passing using the public helpers.

juergba:
```
Priority list:
 * 1. Command-line args
 * 2. RC file (`.mocharc.c?js`, `.mocharc.ya?ml`, `mocharc.json`)
 * 3. `mocha` prop of `package.json`
 * 4. default configuration (`lib/mocharc.json`)
```
I have to check to which `configObects` you are applying `extend`. To RC files only or to _package.json_ `mocha` field as well.
And when the `extend` values are applied, wether the priority is respected.
I'm unsure wether this is correct.

juergba (changes requested):
@mf-pherlihy I did some testing and have found quite a few issues:
- `extends` does not work for relative paths
- array typed options: values are not combined, the configuration of _yargs-parser_ seems to be ignored by `applyExtends`.
- boolean typed option: is buggy, eg `no-diff: true` gets overwritten by default value: `diff: true`
- _package.json_  with field `mocha`: haven't tested yet, but ?

As an alternative without _yargs_ public helper:
we just let `options.parse()` do the job by passing additional configObjects. The order would be crucial.
`configObjects: [{base RC-options}, {extends RC-options}, {base pkg options}, {extends pkg options}]`

What do you think?

mf-pherlihy:
> @mf-pherlihy I did some testing and have found quite a few issues:
> 
> * `extends` does not work for relative paths
> * array typed options: values are not combined, the configuration of _yargs-parser_ seems to be ignored by `applyExtends`.
> * boolean typed option: is buggy, eg `no-diff: true` gets overwritten by default value: `diff: true`
> * _package.json_  with field `mocha`: haven't tested yet, but ?
> 
> As an alternative without _yargs_ public helper:
> we just let `options.parse()` do the job by passing additional configObjects. The order would be crucial.
> `configObjects: [{base RC-options}, {extends RC-options}, {base pkg options}, {extends pkg options}]`
> 
> What do you think?

If we can cover all the issues with yargs-parser and still fix the root issue with extends then im all for it.

bcoe:
@mf-pherlihy if any of this behavior would make sense in yargs/yargs-parser, happy to take patches.

SmashingQuasar:
:wave: Where are we at regarding this PR?
Extending configuration is essential to properly manage mono-repositories.
If you need any help to finish it just say the word and I'll happily look into whatever issue remains. :+1:

JoshuaKGoldberg:
Note: #4980 also touches this area. We should review them together in case one is a dup.

lib/cli/options.js
```diff
@@ -158,7 +173,12 @@ const parse = (args = [], defaultValues = {}, ...configObjects) => {
 const loadRc = (args = {}) => {
   if (args.config !== false) {
     const config = args.config || findConfig();
-    return config ? loadConfig(config) : {};
+    const configObject = config ? loadConfig(config) : {};
+
+    // Set _configPath for use by parse()
+    configObject._configPath = config;
```
JoshuaKGoldberg, line 179: [Refactor] I'm not a big fan of setting "magic" properties that are only used by specific other places. They get hard to keep track of and can accidentally end up being exposed to users. Which is what's happening here: `loadRc` is a `@public` function and thus part of the public API. https://mochajs.org/api/module-lib_cli#.loadRc

I think the root need for this code change is that the `config` value also needs to be available in `parse()`, right? If so: how about extracting the contents of this `if` into a private shared function, then utilizing that function in two places?

```ts
  if (args.config !== false) {
    return myNewFancyFunction(args).configObject;
  }
```

_(roughly that, but maybe with more clear naming?)_

lib/cli/options.js
JoshuaKGoldberg, line 1: [Question] #4980 is also meant to fix `extends`, and has less code. It looks like #4980 uses the higher-level `yargs/yargs` -> `yargs()` instead of this PR's `yargs/helpers` -> `applyExtends`. I think it's normally cleaner to use higher-level functions when possible. They tend to do more of the work for consumers - as seen in #4980. Is there an advantage to `applyExtends`?

test/integration/options.spec.js
JoshuaKGoldberg, line 1: [Praise] Really clean test, nice 🙂

test/integration/options.spec.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+var path = require('path');
+var loadOptions = require('../../lib/cli/options').loadOptions;
+
+describe('options', function() {
+  it('Should support extended options', function() {
```
JoshuaKGoldberg, line 7: [Style] Nit: other tests in this repo generally lower-case the 's'. For consistency...

```suggestion
  it('should support extended options', function() {
```

JoshuaKGoldberg (changes requested):
Seems like a reasonable start - and I like the test case!

I'm interested in what you think of this approach compared to #4980's. If possible, it'd be nice to avoid having to reach into lower-level helpers. But I'm not super familiar with yargs and would love to know if that's misguided here. 🙂

JoshuaKGoldberg:
Marking as `semver-major` because some projects might accidentally be relying on the current, arguably broken behavior.

JoshuaKGoldberg:
👋 ping @mf-pherlihy, do you still have time for this?

lib/cli/options.js
JoshuaKGoldberg: 👋 @mf-pherlihy was hoping to get your thoughts here. What do you think?

JoshuaKGoldberg:

PR #4409, opened 2020-08-18
parallel mode: enable custom worker reporters and object references

boneskull:
@segrey please see this one.  It will allow a reporter to call `runner.linkPartialObjects(true)` (see example in docstring) to replace duplicated deserialized objects with references.  The objects themselves are still shallow, but if there's a particular property you need that is missing, we can likely add it.

I didn't test this too hard, but please let me know if this would help.  I'm not sure about the missing `root` prop issue...

Ref: #4403 

- fixed an issue in rollup config causing weird problems
- changed how we define `global.Mocha` in `browser-entry.js` (tangential)
- removed `object.assign` polyfill (tangential)
- renamed misnamed `buffered-runner.spec.js` to `parallel-buffered-runner.spec.js`

### DIFF
--- a/lib/nodejs/parallel-buffered-runner.js
+++ b/lib/nodejs/parallel-buffered-runner.js
@@ -12,7 +12,12 @@ const {EVENT_RUN_BEGIN, EVENT_RUN_END} = Runner.constants;
 const debug = require('debug')('mocha:parallel:parallel-buffered-runner');
 const {BufferedWorkerPool} = require('./buffered-worker-pool');
 const {setInterval, clearInterval} = global;
-const {createMap} = require('../utils');
+const {createMap, constants} = require('../utils');
+const {MOCHA_ID_PROP_NAME} = constants;
+
+const DEFAULT_WORKER_REPORTER = require.resolve(
+  './reporters/parallel-buffered'
+);
 
 /**
  * List of options to _not_ serialize for transmission to workers
@@ -68,7 +73,7 @@ const states = createMap({
 /**
  * This `Runner` delegates tests runs to worker threads.  Does not execute any
  * {@link Runnable}s by itself!
- * @private
+ * @public
  */
 class ParallelBufferedRunner extends Runner {
   constructor(...args) {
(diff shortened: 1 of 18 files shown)


segrey:
@boneskull Great! I've just tried it with the example from #4403, and now `test` objects are the same across different events. This fixes "`started: false` instead of `started: true`" sub-problem from #4403 description.

boneskull:
@segrey So this looks good as-is?  I'll wrap it up then.

boneskull:
@segrey if you can share it, I'd love to see the implementation you've built with this

boneskull:
@mochajs/core please review, otherwise will merge in next couple days

lib/nodejs/parallel-buffered-runner.js
```diff
@@ -98,12 +107,63 @@ class ParallelBufferedRunner extends Runner {
...
+        let newObj;
+        if (obj && typeof obj === 'object' && obj[MOCHA_ID_PROP_NAME]) {
+          const id = obj[MOCHA_ID_PROP_NAME];
+          newObj = this._linkedObjectMap.has(id)
+            ? Object.assign(this._linkedObjectMap.get(id), obj)
+            : obj;
+          this._linkedObjectMap.set(id, newObj);
+          parent[prop] = newObj;
+        } else {
+          newObj = obj;
```
boneskull, line 151: this is defensive and afaict will never be hit.  it should probably throw.

boneskull: updated to throw

segrey:
@boneskull Please see [mocha-intellij.zip](https://github.com/mochajs/mocha/files/5308495/mocha-intellij.zip). Main reporter is `mocha-intellij/lib/mochaIntellijReporter.js`. There is some code duplication in main reporter and worker reporter, but it's just a proof-of-concept :)

There are some minor questions left in https://github.com/mochajs/mocha/issues/4403#issuecomment-684091528. Would be great if you could take a look.

boneskull:
@segrey This has landed... will be released in next version which will be v8.2.0

Issue #4415, opened 2020-08-25
Print wrong count concurrent jobs in debug log


outsideris:
# Description
In `mocha:cli:run:helpers` debug logs, it will print current concurrent jobs like below:

```
$ DEBUG=mocha:cli:run:helpers ./bin/mocha --parallel -jobs=3 sample.test.js
  mocha:cli:run:helpers files (in order):  [ '/Users/jeonghoon/Dropbox/projects/github/mocha/tt.js' ] +0ms
  mocha:cli:run:helpers executing 1 test file(s) across 3 concurrent jobs +0ms
```

But without `--jobs` option, it prints `NaN`.

```
$ DEBUG=mocha:cli:run:helpers ./bin/mocha --parallel sample.test.js
  mocha:cli:run:helpers files (in order):  [ '/Users/jeonghoon/Dropbox/projects/github/mocha/tt.js' ] +0ms
  mocha:cli:run:helpers executing 1 test file(s) across NaN concurrent jobs +0ms
```

### Steps to Reproduce
Run mocha with `--parallel` option but, without `--jobs` option.

```
$ DEBUG=mocha:cli:run:helpers ./bin/mocha --parallel sample.test.js
```

**Expected behavior:** [What you expect to happen]
Print current concurrent job count instead `NaN`.

### Versions
- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 8.1.1
- The output of `node --version`: v12.18.1
- Your operating system
  - name and version: macOS
  - architecture (32 or 64-bit): 64bit
- Your shell (e.g., bash, zsh, PowerShell, cmd): zsh

boneskull:
fixed in #4478

PR #4416, opened 2020-08-25
fix: the number of concurrent job with --parallel option in mocha:cli:run:helpers

sujin-park:
### Requirements

### Description of the Change

When users run mocha with ```--parallel``` option wihtout ```--jobs``` option, ```undefined``` was passed to the parallelRun method.

I changed options.jobs to the number of CPU cores less 1 when options.jobs was false.

**AS-IS**
스크린샷 2020-08-25 오후 6 09 07

**TO-BE**
스크린샷 2020-08-25 오후 6 08 37

### Alternate Designs

in ```lib/cli/run-helpers.js``` from line ```179```

**AS-iS**
```javascript
options.jobs
```

**TO-BE**
```javascript
options.jobs ? options.jobs : workerpool.cpus - 1
```

It's a confirmed bug as described in #4415. 

### Benefits

Users can know exact current concurrent job count instead NaN.

### Possible Drawbacks

N/A

### Applicable issues

#4415

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -173,11 +173,7 @@ const singleRun = async (mocha, {exit}, fileCollectParams) => {
  */
 const parallelRun = async (mocha, options, fileCollectParams) => {
   const files = collectFiles(fileCollectParams);
-  debug(
-    'executing %d test file(s) across %d concurrent jobs',
-    files.length,
-    options.jobs ? options.jobs : workerpool.cpus - 1
-  );
+  debug('executing %d test file(s) in parallel mode', files.length);
   mocha.files = files;
 
   // note that we DO NOT load any files here; this is handled by the worker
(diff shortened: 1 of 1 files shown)


lib/cli/run-helpers.js
```diff
@@ -175,7 +176,7 @@ const parallelRun = async (mocha, options, fileCollectParams) => {
   debug(
     'executing %d test file(s) across %d concurrent jobs',
     files.length,
-    options.jobs
+    options.jobs ? options.jobs : workerpool.cpus - 1
```
boneskull, line 179: I would just remove the job count. It's printed in the debug logs elsewhere:

https://github.com/mochajs/mocha/blob/e677219a78d6deb4bd95ab6a484dff930df7717d/lib/nodejs/buffered-worker-pool.js#L72

maybe just say `executing %d test file(s) in parallel mode`

sujin-park: I fixed your feedback. I removed options.jobs params and added text ```in parallel mode```.
Thanks for your comment.

outsideris (approved):

boneskull:
thanks!

PR #4419, opened 2020-08-25
alias lookupFiles, loadRc, loadPkgRc and loadOptions to lib/cli module; fixes #4398

### LINKED ISSUE #4398: Mocha.utils.lookupFiles is no longer available
The commit linked below migrated `lookupFiles` from utils to cli. This caused it to be removed from `Mocha.utils.lookupFiles`, and it appears that it is no longer accessible anywhere.

This was still exposed as of 8.0.1 but is no longer available as of 8.1.1.

Additionally, `Mocha.utils.lookupFiles` does still appear in the current version of `@types/mocha`.

We are using this when calling `mocha.run` programatically.

This is the call we are making to `lookupFiles`

```javascript
const testFiles = Mocha.utils
    .lookupFiles('**/*spec.[tj]s')
    .filter((file) => !file.includes('node_modules'));
```

https://github.com/mochajs/mocha/commit/40f951baa5cf8aab40dedec9bcb0f60ee9581333#diff-50e3aa130a4f97a42ee2cf111c7b1d9d

boneskull:
- fixes API documentation for all of these (and `module:lib/cli.main`) via JSDoc aliasing
- aliases `lib/cli/cli.js` to `module:lib/cli`; `module:lib/cli/cli` is no longer a thing
- the `lib/cli/options.js` and `lib/cli/lookup-files.js` _modules_ (not necessarily their _contents_) are now private

example usage:

```js
const {loadRc, loadPkgRc, loadOptions, lookupFiles} = require('mocha/lib/cli');
```

**UPDATE**

- restores `lookupFiles` to `lib/utils` to un-break the breaking change introduced in v8.1.0.
- use of `lookupFiles` in `lib/utils` will print a deprecation notice directing the consumer to the new path
- use of `lookupFiles` in `lib/utils` _in a browser_ (since it's accessible) will throw an exception.

**ANOTHER UPDATE**

- moves `require.cache`-touching code out of `lib/mocha.js` and into `lib/nodejs/file-unloader.js`; excludes this file from the bundle
- add test to assert a webpack build which imports the mocha bundle emits no errors or warnings

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -3,19 +3,24 @@
 'use strict';
 
 /**
- * This is where we finally parse and handle arguments passed to the `mocha` executable.
- * Option parsing is handled by {@link https://npm.im/yargs yargs}.
- * If executed via `node`, this module will run {@linkcode module:lib/cli/cli.main main()}.
- *
- * @private
- * @module
+ * Contains CLI entry point and public API for programmatic usage in Node.js.
+ * - Option parsing is handled by {@link https://npm.im/yargs yargs}.
+ * - If executed via `node`, this module will run {@linkcode module:lib/cli.main main()}.
+ * @public
+ * @module lib/cli
  */
 
 const debug = require('debug')('mocha:cli:cli');
 const symbols = require('log-symbols');
 const yargs = require('yargs/yargs');
 const path = require('path');
-const {loadOptions, YARGS_PARSER_CONFIG} = require('./options');
+const {
+  loadRc,
+  loadPkgRc,
+  loadOptions,
+  YARGS_PARSER_CONFIG
+} = require('./options');
+const lookupFiles = require('./lookup-files');
 const commands = require('./commands');
 const ansi = require('ansi-colors');
 const {repository, homepage, version, gitter} = require('../../package.json');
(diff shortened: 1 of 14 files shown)


boneskull:
@mochajs/core I know y'all had some thoughts about #4398, so would appreciate feedback here.  I think it's important to fix the breaking change and issue a deprecation notice, but I won't protest if you'd like me to revert my other API-surface changes here.

Munter:
I moved `lookupFiles` out of `utils` because `utils` is required in the browser build and having a dependency graph in there that starts requiring node libraries was problematic. If we add node specific code back to the `utils` dependency graph we need to make sure the browser bundle doesn't break, and given #4422 that it also doesn't cause bundlers to break

nicojs (approved):
Good to add this API back. 👍

boneskull:
yes, it does not appear in the bundle.

boneskull:
Will fix #4422 here as well.

boneskull:
@munter to be clear, I don't disagree that file-touching stuff should _not_ be in `lib/utils.js` or `lib/mocha.js`.  

@mochajs/core we can move `utils.lookupFiles()` and `Mocha.unloadFiles()` (`Mocha.unloadFile()` is not a public API) for good in v9, if we want.  I did not soft-deprecate `Mocha.unloadFiles()` here, but perhaps I should, and export it from `lib/cli` like the others?

boneskull:
please note that this now adds _webpack_ to our development dependencies 😦 .  probably better to do declare it explicitly than to just `npx` it.

Munter:
A future major to move these out of the browser graph would be great. I have a few ideas for attempting a cleaner cut that I can start a draft of soon

Munter (approved):

craigtaub (approved):
LGTM

auroq:
As per request for comment here: https://github.com/mochajs/mocha/issues/4398#issuecomment-681135481

In terms of whether we need globbing to be part of Mocha, the answer is no. It's not difficult to do globbing ourselves. However, what we are trying to do when calling `mocha.run` programmatically is emulate the CLI in an automated fashion.

Thus, my argument is one for consistency. As boneskull stated, something like globbing being part of mocha is "of dubious necessity" as third party globbing tools could be used. If that's the case, though, I would expect to also need to do globbing beforehand if I was calling mocha from CLI or in a bash script.

Basically, I advocate that one shouldn't have to relearn a new approach to calling a tool like mocha for every method of interaction it supports. It should be basically the same overall approach.

Issue #4322, opened 2020-06-09
make parallel mode smarter


boneskull:
I have a few ideas for improvements here now that we're past the MVP:

1. Detect if we're only running a single file, and disable parallel mode automatically.  Running a single file in parallel mode will _always_ (AFAIK) be slower than running in serial.  Add a command-line option to disable this behavior (a general-purpose, contextual `--force` might be helpful)
1. Automatic optimization via duration caching
    - Cache _per file_ durations (cache could live in `node_modules/.cache/mocha`, which is an unofficial convention).  
        - This would be timing the `run()` call in `lib/nodejs/worker.js`, from beginning to end.  
        - Maybe calculate the mean over _n_ most recent runs?
    - On subsequent runs, execute slowest test files _first_.  This will help avoid the case at the end of the run where there's only a single worker process, munching on a meaty test file, and the other workers are idle.  TypeScript uses a strategy like this in their custom tooling around Mocha (look at their implementation for ideas).

boneskull:
(Note new `parallel` label)

stoicchild8910:
@boneskull  
hello, I read this issue and working on it and came up with two questions.

first, where to change.
I found mocha has two spots to turn on parallel function, 
1) parallelMode function (in lib/mocha.js)
2) parallelrun (in lib/run-helpers)

Because parallelMode function comes first, I think it would be better to check and switch in parallelMode function if the file is better with running in single mode. I want to know your opinion.

second, the testing size
would it be enough only with a single file?
maybe two files, for three.. I also want to know your opinion on this!

thank you for your time.

JoshuaKGoldberg:
This would be really nice for performance - and since parallel runs aren't very deterministic right now, I don't think it'd likely be received as a big breaking change for users. But just to be safe, slapping the `semver-major` label on.

JoshuaKGoldberg:
This would be really nice for performance - other test runners such as Jest take similar strategies. Since parallel runs aren't very deterministic right now, I don't think it'd likely be received as a big breaking change for users. But slapping the `semver-major` label on just to be safe.

PR #4428, opened 2020-08-29
Fix parallel feature: disable parallel automatically/ add command line option

stoicchild8910:
### Requirements

### Description of the Change

This PR comes from the idea #4322
AS-IS
[single run]
parallelrun
[parallel run]
singlerun

(1) fix parallel feature : When running a single file, disable parallel mode automatically.

If we run only a single file in parallel mode, it is much slower.
So I added a feature to automate swap from parallel mode to single mode when we're running a single file.

(2) Added Command Line Option: added an option to FORCE PARALLEL MODE in (1) situation

### Benefits

When running a single file, Mocha automatically run in a faster way.

### Possible Drawbacks

There must be a way to force parallel mode when running a single file if needed.

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -217,13 +217,12 @@ exports.runMocha = async (mocha, options) => {
   };
 
   const files = collectFiles(fileCollectParams);
-  if(files.length===1) {
-    if(options.parallel && !options.parallelForce) {
+  if (files.length === 1) {
+    if (parallel && !parallelForce) {
       mocha.parallelMode(false);
-      parallel=false;
+      parallel = false;
     }
   }
-  
   let run;
   if (watch) {
     run = parallel ? watchParallelRun : watchRun;
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -203,7 +203,8 @@ exports.runMocha = async (mocha, options) => {
     parallel = false,
     recursive = false,
     sort = false,
-    spec = []
+    spec = [],
+    parallelForce = false
   } = options;
 
   const fileCollectParams = {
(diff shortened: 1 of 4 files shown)


stoicchild8910:
I'm working on failed test cases!

stoicchild8910:
I am still working on test cases and PRs.

boneskull (commented):
Thanks for the PR!

In general, I agree with this--Mocha should not run in parallel if there is only a single file, _and_ we need a way to force it to run in parallel.  This will affect many of our tests, as you have probably discovered.  A way to address this is to detect `--parallel` in the arguments (see `test/integration/helpers.js`) and append `--force-parallel` to it.  I would prefer the option is called `--force-parallel`, not `--parallel-force`.

boneskull:
I am unsure if this would be a breaking change.  Ideally, it would not have any impact on consumers, but... you never know. 
 @mochajs/core thoughts?

boneskull:
Please mark this PR as "ready" when the tests are passing

stoicchild8910:
> Please mark this PR as "ready" when the tests are passing
Thank you for your opinion!

I also don't think many customers would run a single file in parallel option.
And that's the point where I thought to predict running time and automatize running mode.
For example we can test on many different files and how much time it needs to run -> if we have our standard we can automize running mode.
This is just my opinion!

And I will do mark PR when test passed.

boneskull:
@stoicchild8910 

> I also don't think many customers would run a single file in parallel option.

right, consumers may not, but Mocha needs to do so in order to efficiently test parallel mode.  so this is why I'd want `--force-parallel`

JoshuaKGoldberg:
👋 coming back to this @stoicchild8910, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

Issue #4286, opened 2020-05-12
Mocha.prototype.run API is weird and inconsistent


boneskull:
Currently, `Mocha.prototype.run` looks like this:

```js
Mocha.prototype.run = function(fn) {
  this._guardRunningStateTransition(); // can throw

  // ... setup runner, etc ...

  function done(failures) {
    // some stuff
    fn = fn || utils.noop;
    if (reporter.done) {
      reporter.done(failures, fn);
    } else {
      fn(failures);
    }
  }

  return runner.run(done);
}
```

Note:

1. This function can throw if the `Mocha` instance is not in the correct state.
1. The callback parameter, `fn`, is _optional_.
1. `Runner#run` returns a `Runner` instance.
1. `Runner#run` will call its callback, `done`, asynchronously.
1. `fn` will then be called (asynchronously) with a single numeric argument, `failures`.

This is... not like any API I'm familiar with.  It is **awkward and unfriendly**.

If we can adopt an conventional style for this API, I think consumers of `Mocha.prototype.run` (consumers running Mocha programmatically; third-party libs built on Mocha, etc.) will have an easier time of it--especially now that we're potentially throwing an exception synchronously.

Now, we can't just go and break this function w/o befouling presumably a lot of tooling.

We have, as I see it, two options:

1.  Make a best-effort guess at whether the consumer expects _error-first_ ("nodeback") style.  This can naively be done by inspecting the `length` prop of `fn`.  If the value is `2`, we can call it with an `Error` (if any), and `failures` as the second argument.  This will not work if `fn` doesn't care about `failures` (e.g., `fn` is `(err) => {}` instead of `(err, failures) => {}`).
1. Create a new, error-first-style function with the desired API, and soft-deprecate this one.  This _cannot_ be a `Promise`-returning function; it needs to return the `Runner` instance (we use it in the "watch" mode implementation). It _could_ return an object of type `{{Runner, Promise<number>}}`.

If we do go and make a new function, we should re-evaluate whether calling the callback (or resolving) with a _number_ (`failures`) is the best we can do.  We could return a big-ass object full of information about the run instead.  If we _also_ made a reporter _optional_, we could enable some cool stuff to be more easily built, e.g.:

```js
const Mocha = require('mocha');

async function main() {
  const mocha = new Mocha();

  someFiles.forEach(file => {
    mocha.addFile(file);
  });

  const result = await (mocha.run().result);

  return result.tests.filter(test => test.status === 'failed');
}
```

[...]

PR #4429, opened 2020-08-30
Fix: mocha return interface #4286

MoonSupport:
### Requirements

### Description of the Change

`mocha.run` return `{Runner: runner, Promise<Test[]>}`

To be precise, `runner.run` return this.

### Alternate Designs

Before
mocha.run() : runner

After
mocha.run() : {Runner: runner, Promise<Test[]>}

https://github.com/mochajs/mocha/issues/4286

### Benefits

It can provide an easier interface to consumers running Mocha programmatically.

### Possible Drawbacks

I was careful because it was a modification of the interface of the core function.

the callback is deprecated, but it couldn't be deleted for existing users using callback.
And I think it could be a bit slow in the process of putting the tests into an array.
Finally, since I don't have enough background on Mocha, I desperately need your help to prevent side-effect.

### Applicable issues

https://github.com/mochajs/mocha/issues/4286

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -999,16 +999,16 @@ Mocha.prototype.run = function(fn) {
   exports.reporters.Base.inlineDiffs = options.inlineDiffs;
   exports.reporters.Base.hideDiff = !options.diff;
 
-  const done = failures => {
+  const done = tests => {
     this._previousRunner = runner;
     this._state = this._cleanReferencesAfterRun
       ? mochaStates.REFERENCES_CLEANED
       : mochaStates.INIT;
     fn = fn || utils.noop;
     if (typeof reporter.done === 'function') {
-      reporter.done(failures, fn);
+      reporter.done(tests, fn);
     } else {
-      fn(failures);
+      fn(tests);
     }
   };
 
@@ -1017,14 +1017,15 @@ Mocha.prototype.run = function(fn) {
       this.options.enableGlobalSetup && this.hasGlobalSetupFixtures()
         ? await this.runGlobalSetup(runner)
         : {};
-    const failureCount = await runner.runAsync({
+    const tests = await runner.runAsync({
       files: this.files,
       options
     });
+
     if (this.options.enableGlobalTeardown && this.hasGlobalTeardownFixtures()) {
       await this.runGlobalTeardown(runner, {context});
     }
-    return failureCount;
+    return tests;
   };
 
   // no "catch" here is intentional. errors coming out of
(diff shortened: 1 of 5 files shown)


boneskull (changes requested):
Thanks for this!

So, while the function itself is weird, I think maybe it's unwise to change the API of `Mocha.prototype.run`, as I suggested in #4286.  Maybe better to expose a new method, e.g., `Mocha.prototype.runAsync`.  PR #4360 adds a `Runner.prototype.runAsync` method, which would work nicely in tandem with a `Mocha.prototype.runAsync`.

My advice is to wait until #4360 lands (likely very soon), then:

1. Create method `Mocha.prototype.runAsync`
1. Either this method will _wrap_ `Mocha.prototype.run` _or_ extract common behavior into a third private method, e.g., `Mocha.prototype._run()` and call `_run()` from both places.
1. Update usages of `Mocha.prototype.run` to use `Mocha.prototype.runAsync` instead
1. Copy unit tests for `Mocha.prototype.run` and assert they work against the new `Promise`-returning API
1. Also: don't forget `mocha.run()` in `browser-entry.js`.  Run browser tests in headless Chrome via `npm start test.browser`

At that point, we can decide if `Mocha.prototype.run()` should be deprecated.  I'm thinking the impact of removing it would outweigh the burden of keeping it around, even if we don't use it internally.

boneskull:
Once this is done, please mark this PR as "ready"; I've converted it to a draft

MoonSupport:
Good!
I'll wait while checking it!

MoonSupport:
Hello!
I adjusted your update.
However, in the parallel mode, there is a problem with the test, so in this case, it operates as before.
The stack seems to be full, but I don't know the exact cause.

Sorry for not being able to help 😫

boneskull (commented):
thanks.  I'm not sure I understand what the problem is that you're describing... but the changes to `Runner.prototype.run` aren't making sense to me.  

if this helps: in parallel mode, Mocha uses `ParallelBufferedRunner` which is in `lib/nodejs/parallel-buffered-runner.js`.  this is a subclass of `Runner`.  _only_ the _main process_  uses `ParallelBufferedRunner`.  each worker process in the pool will invoke its own Mocha and use `Runner`.  the `Mocha` instance in the worker process does not necessarily know or care if it is in parallel mode.

are you interested in working on this further?

MoonSupport:
I don't have enough time to watch this problem right now. but if you leave this pr, I'll be able to see it later :) 
I still want to do it, but it's okay to close it.

boneskull:
Feel free to reopen later

Issue #4430, opened 2020-08-30
Broken wallaby logo on mochajs.org


outsideris:
<img width="709" alt="스크린샷 2020-08-30 오후 2 06 36" src="https://user-images.githubusercontent.com/390146/91651647-28fa3100-eaca-11ea-845f-744ad96bdd63.png">

In our mochajs.org site, [wallaby-logo.png](https://github.com/mochajs/mocha/blob/master/docs/images/wallaby-logo.png) is broken. 

It is only happened in `docs.production` build, so I think it is optimization issue.

Donghoon759:
Let me work on it :)

boneskull:
fixed by #4431

boneskull:
This should be in the process of being deployed to https://mochajs.org

PR #4431, opened 2020-08-30
Fix: Broken wallaby logo on mochajs.org #4430

Donghoon759:
### Requirements

### Description of the Change

After executing ```docs.production``` script without ```trim``` on ```wallaby-logo```, the logo  takes 126 bytes.
However, with ```trim```, the logo takes 271bytes and it is not shown properly.
So I think it is better to remove trim operation option on ```wallaby-logo.png``` in ```default.liquid```

The reason why the logo is not shown properly after ```trim``` is a bug with ```impro``` used by ```assetgraph-builder``` that execute ```trim``` operation with ```gm```, not ```sharp```. So, it is actually a bug with ```gm```. I created another pr on ```impro``` to fix this as well.
https://github.com/papandreou/impro/pull/4

### Alternate Designs

X

#4430 

### Benefits

The size of wallaby-logo after production build will decrease and it will appear properly.

### Possible Drawbacks

X

### Applicable issues

#4430

### DIFF
--- a/docs/_includes/default.liquid
+++ b/docs/_includes/default.liquid
@@ -52,7 +52,7 @@
         title="Mocha is sponsored by Wallaby"
       >
         <figure id="wallaby-logo">
-          <img src="images/wallaby-logo.png?trim" loading="lazy" alt="Wallaby logo" />
+          <img src="images/wallaby-logo.png" loading="lazy" alt="Wallaby logo" />
           <figcaption>Wallaby</figcaption>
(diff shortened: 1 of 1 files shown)


outsideris (approved):
스크린샷 2020-08-30 오후 4 39 52

Thank you. I confirmed it on my local because Netlify doesn't use production build.

boneskull:
LGTM, thanks!

PR #4434, opened 2020-09-01
Add speed in -R json option (#4226)

### LINKED ISSUE #4226: Expose `test.speed` (or `slow` value) in reports (at least JSON)

Relative to my purposes to allow precise badge-making from Mocha reporters that takes into account test `speed`/`slow` , it is a problem.

I'd like for `test.speed` to be recorded in JSON results (and/or the `slow` value associated with the file as a whole and individual tests).
 
Currently, when I build a badge (here based on a template showing all of the various possible meta-data) from the `mocha-badge-generator` reporter, I can utilize information from a JSON file saved from a Mocha JSON reporter:

badge-template

However, the number of "fast", "medium", and "slow" tests do not take into account whether individual tests allowed for slower speeds (via calls to `this.slow()`) nor does it take into account if the default `slow` value (of 75ms) was changed for the original tests (I have instead needed to base the notion of `slow` on config repeated at runtime for this after-the-fact badge maker about what constitutes "slow" where I followed your algorithm for considering "medium" as half of "slow").

But if I had access to the `slow` value for each test (or whether a test was considered "slow", "medium", or "fast" when the test report was built), I could have my totals take into account expectations by the test makers.

In addition to `speed` (or `slow`) information, `timedOut` (and maybe `sync` too) would be useful to report also, so that reporters can aggregate information on such test types, counting for example how many tests failed because of timing out.

Where tests have to run separate times or otherwise have to act on reports previously generated, there are not really options here that I can think of.

Thanks!

wwhurin:
### Requirements

### Description of the Change

I modified the ` -R json` option to show speed.

### Alternate Designs

N/A

#4226 

### Benefits

You can see speed result in `json reporter `

### Possible Drawbacks

N/A

### Applicable issues
Fixes #4226

### DIFF
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -11,7 +11,6 @@ describe('JSON reporter', function() {
   var runner;
   var testTitle = 'json test 1';
   var testFile = 'someTest.spec.js';
-  var testSpeed = 'fast';
   var noop = function() {};
 
   beforeEach(function() {
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -98,7 +98,7 @@ describe('JSON reporter', function() {
             {
               title: testTitle,
               file: testFile,
-              speed: testSpeed
+              speed: /(slow|medium|fast)/
             }
           ]
         }
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -92,7 +92,6 @@ describe('JSON reporter', function() {
     suite.addTest(test);
 
     runner.run(function(failureCount) {
-      sinon.restore();
       expect(runner, 'to satisfy', {
         testResults: {
           passes: [
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -86,9 +86,7 @@ describe('JSON reporter', function() {
   });
 
   it('should have 1 test pass', function(done) {
-    var test = new Test(testTitle, function(done) {
-      done();
-    });
+    const test = new Test(testTitle, () => {});
 
     test.file = testFile;
     suite.addTest(test);
--- a/test/reporters/json.spec.js
+++ b/test/reporters/json.spec.js
@@ -11,6 +11,7 @@ describe('JSON reporter', function() {
   var runner;
   var testTitle = 'json test 1';
   var testFile = 'someTest.spec.js';
+  var testSpeed = 'fast';
   var noop = function() {};
 
   beforeEach(function() {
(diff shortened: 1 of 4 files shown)


boneskull (changes requested):
Thanks!  Please update `test/reporters.json.spec.js` to reflect this change.  You would likely modify an existing assertion, e.g.,

```js
      expect(runner, 'to satisfy', {
        testResults: {
          failures: [
            {
              title: testTitle,
              file: testFile,
              speed: expect.it('to be greater than', 0),
              err: {
                message: error.message
              }
            }
          ]
        }
      });
```

outsideris (changes requested):
Could add this in `lib/reporters/json-stream.js` as well? 
And it needs to change `test/reporters/json-stream.spec.js` like boneskull said.

wwhurin:
> Thanks! Please update `test/reporters.json.spec.js` to reflect this change. You would likely modify an existing assertion, e.g.,
> 
> ```js
>       expect(runner, 'to satisfy', {
>         testResults: {
>           failures: [
>             {
>               title: testTitle,
>               file: testFile,
>               speed: expect.it('to be greater than', 0),
>               err: {
>                 message: error.message
>               }
>             }
>           ]
>         }
>       });
> ```

I have changed json.spec.js! However, speed is not a number, but things like 'fast' and'slow'. Also, speed does not pass unless it passes. But json.spec.js does not have a pass case, so I added a pass case to check the speed option. Would it be okay? 🤔

wwhurin:
> Could add this in `lib/reporters/json-stream.js` as well?
> And it needs to change `test/reporters/json-stream.spec.js` like boneskull said.

OK. I changed `json-stream.js` and `json-stream.spec.js`! You can see the speed in json-stream option.

outsideris (approved):
Thank  you.

test/reporters/json.spec.js
```diff
@@ -84,6 +85,32 @@ describe('JSON reporter', function() {
     });
   });
 
+  it('should have 1 test pass', function(done) {
+    var test = new Test(testTitle, function(done) {
+      done();
+    });
```
boneskull, line 91:
```suggestion
    const test = new Test(testTitle, () => {});
```

test/reporters/json.spec.js
```diff
@@ -84,6 +85,32 @@ describe('JSON reporter', function() {
...
+
+    runner.run(function(failureCount) {
+      sinon.restore();
+      expect(runner, 'to satisfy', {
+        testResults: {
+          passes: [
+            {
+              title: testTitle,
+              file: testFile,
+              speed: testSpeed
```
boneskull, line 104: It's highly unlikely that `speed` would not be `fast` here, but since we haven't mocked the runner (as in `json-stream.spec.js`), it would be safer to do this:

```suggestion
              speed: /(slow|medium|fast)/
```

Also, you can remove the addition on [L14](https://github.com/mochajs/mocha/pull/4434/files#diff-1fa46f0ad13d99b90fbf14520f5b565eR14).

boneskull (approved):
Thanks for updating this.  I have a few suggestions.  If you don't apply them, I will do so after merge.

test/reporters/json.spec.js
```diff
@@ -84,6 +85,32 @@ describe('JSON reporter', function() {
...
+  it('should have 1 test pass', function(done) {
+    var test = new Test(testTitle, function(done) {
+      done();
+    });
+
+    test.file = testFile;
+    suite.addTest(test);
+
+    runner.run(function(failureCount) {
+      sinon.restore();
```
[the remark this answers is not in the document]
wwhurin: That's my fault. I didn't know exactly what role it is. 😂 Thank you!

PR #4435, opened 2020-09-02
Change assert module's deprecated methods in testing files

sujin-park:
### Requirements

### Description of the Change

I removed deprecated methods and tested it. ```equal, deepEqual``` methods are deprecated since v.9.9.0. [Assert Reference](https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message)

I used ```strictEqual, deepStrictEqual``` instead.

### Alternate Designs

**deprecated**
```javascript
assert.equal(getTitle(this), undefined);
assert.deepEqual(sock.address().family, 'IPv4');
```

**changed**

```javascript
assert.strictEqual(getTitle(this), undefined);
assert.deepStrictEqual(sock.address().family, 'IPv4');
```

For now, PR's just changed methods that was deprecated. I wonder if i can change code using unexpected library as shown below. I found [this issue](https://github.com/mochajs/mocha/pull/3343) about migrating assert to Unexpected assertion library.  I wonder if there are still ```assert``` left.

**assert**
```javascript
assert.strictEqual(getTitle(this), undefined);
assert.deepStrictEqual(sock.address().family, 'IPv4');
```

**unexpected assertion library**
```javascript
expect(getTitle(this), 'to be', undefined);
expect(sock.address().family, 'to equal', 'IPv4');
```

N/A

### Benefits

Users can view the modified version of the function within code.

### Possible Drawbacks

N/A

### Applicable issues

patch release

### DIFF
--- a/docs/example/debug-hanging-mocha.js
+++ b/docs/example/debug-hanging-mocha.js
@@ -15,7 +15,7 @@ describe('how to debug Mocha when it hangs', function () {
 
   it('should complete, but Mocha should not exit', function(done) {
     const sock = net.createConnection(10101, () => {
-      assert.deepEqual(sock.address().family, 'IPv4');
+      assert.deepStrictEqual(sock.address().family, 'IPv4');
       done();
     });
   });
--- a/test/integration/diffs.spec.js
+++ b/test/integration/diffs.spec.js
@@ -72,6 +72,12 @@ describe('diffs', function() {
   var diffs, expected;
 
   before(function(done) {
+    // @TODO: It should be removed when Node.js 10 LTS is not supported.
+    const nodeVersion = parseInt(process.version.match(/^v(\d+)\./)[1], 10);
+    if (nodeVersion === 10) {
+      this.skip();
+    }
+
     run('diffs/diffs.fixture.js', [], function(err, res) {
       if (err) {
         done(err);
--- a/test/integration/diffs.spec.js
+++ b/test/integration/diffs.spec.js
@@ -13,7 +13,8 @@ var path = require('path');
  * returns {string[]}
  */
 function getDiffs(output) {
-  var diffs, i, inDiff, inStackTrace;
+  var diffs, i, inDiff, isEncountered, inStackTrace;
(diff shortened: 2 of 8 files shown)


sujin-park:
I'm working on test cases 😢

boneskull:
Please mark this PR as ready when ready.  I know there's at least _one_ place in our tests that must use `assert`.

sujin-park:
Thank you for your feedback.
I modified 2 test files including diff tests. When using strictEqual, deepStrictEqual function instead of strictEqual, deepEqual with node version of 10, tests failed because inline diff outputs twice from AssertionError.

스크린샷 2020-09-10 오후 6 50 15

I modified diffs.spec.js so that only one of the two outputs can be printed. also i changed ```should not display a diff``` 
test to skip when node version is 10. It should be removed when Node.js 10 LTS is not supported.

Please let me know what needs to be modified. Thanks! I marked this PR.

boneskull (changes requested):
Thanks.  I think maybe better to just detect Node.js v10.x and skip the tests.  So revert the changes to `test/integration/diffs.spec.js` and use the same strategy as in `test/integration/no-diff.spec.js`. do it in a `before()` hook within the top-level `describe()`.

it's worth mentioning (in the docs, maybe?) that Mocha's "no diff" behavior does not work if the diff is embedded in the `message` prop of an `Error`.

sujin-park:
> Thanks. I think maybe better to just detect Node.js v10.x and skip the tests. So revert the changes to `test/integration/diffs.spec.js` and use the same strategy as in `test/integration/no-diff.spec.js`. do it in a `before()` hook within the top-level `describe()`.

I reverted the changes to test/integration/diffs.spec.js and use skip tests in the file as in test/integration/no-diff.spec.js.

> it's worth mentioning (in the docs, maybe?) that Mocha's "no diff" behavior does not work if the diff is embedded in the `message` prop of an `Error`.

I'm considering mentioning [this part of the docs](https://mochajs.org/#-diff) that you told, so can I proceed with it in another PR? please let me know if it's okay to proceed like that. Thank you for your comments.

boneskull (approved):

boneskull:
lgtm, thanks

Issue #4433, opened 2020-09-01
Mocha does not pass "--prof" parameter to Node and thus does not allow for profiling


FlourishS:
### Expected

When mocha is run with the --prof argument it will pass this argument to Node thus enabling Node to be profiled.

Example package.json:

```
...
  "scripts": {
    "start": "nodemon ./bin/www",
    "test": "mocha --prof --exit",
...
```

### Actual

The --prof argument is not passed to Node. It is therefore not possible to run node in profiling mode when using Mocha.

A dirty workaround is to patch bin/mocha as follows:

```
// sort options into "node" and "mocha" buckets
Object.keys(opts).forEach(opt => {
  if (isNodeFlag(opt) || opt == 'prof') { // dirty workaround
    console.log("node flag: " + opt);
    nodeArgs[trimV8Option(opt)] = opts[opt];
  } else {
    console.log("mocha flag: " + opt);
    mochaArgs[opt] = opts[opt];
  }
});
```

Tested with Mocha 8.1.2 and Node v12.15.0

Thank you very much for considering this bug.

outsideris:
I confirmed `--prof` doesn't work. I think it is because of `--prof` is not in `process.allowedNodeEnvironmentFlags`.

boneskull:
See #4439 for reasoning

boneskull:
For anyone who wants to use `--prof`:

1. All V8 options (of which `--prof` is one) can be specified by prepending `--v8-` to the flag name, so use `--v8-prof`
1. Alternatively, use `NODE_OPTIONS=--prof` env var

dgolant:
Heads up: second option above will throw "--prof not allowed in NODE_OPTIONS"

frank-dspeed:
@boneskull we do not allow setting --prof in node options you can try it with a current nodejs version

PR #4439, opened 2020-09-06
Add support for node v8 option 'prof' parameter (#4433)

Donghoon759:
### Requirements

### Description of the Change

I modified `--prof` to be considered as a 'Node' flag so that Node can handle it.

### Alternate Designs

N/A

#4433 

### Benefits

Users can pass `--prof` argument to mocha and Node can handle it by profiling.

### Possible Drawbacks

N/A

### Applicable issues

#4433

### DIFF
--- a/lib/cli/node-flags.js
+++ b/lib/cli/node-flags.js
@@ -27,6 +27,7 @@ const debugFlags = new Set(['inspect', 'inspect-brk']);
  *   - `--trace-*`
  *   - `--es-staging`
  *   - `--use-strict`
+ *   - '--prof'
  *   - `--v8-*` (but *not* `--v8-options`)
  * @summary Whether or not to pass a flag along to the `node` executable.
  * @param {string} flag - Flag to test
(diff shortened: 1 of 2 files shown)


boneskull:
Ahh, yes, the v8 options... they refuse to expose them in any sort of public API.

boneskull (commented):
Thanks!

So, I don't think this is an actual _bug_.  It is not a `node` option, e.g., is not exposed in `process.allowedNodeEnvironmentFlags`.  The ones whitelisted here are _historical_ and remain for backwards compatibility only--they are V8 flags which Mocha previously hard-coded (before `process.allowedNodeEnvrionmentFlags` landed).  All V8 options should be accessible via `--v8-<option-name>`, e.g., `--v8-prof`.

I'm inclined to _not_ accept this PR, because it's really a slippery slope--taken to its logical conclusion, we end up hard-coding the entire list of V8 options.

Thoughts?  @mochajs/core @outsideris

juergba:
We shouldn't add this option to our special CLI supported list. regex expressions like this one are a nightmare to maintain anyway.
I don't know why in Node's docs `--prof` is listed as option, but isn't exposed in `process.allowedNodeEnvironmentFlags`. There must be a reason which I don't understand.

We should deprecate the complete list of historically supported options and instruct users to use `NODE_OPTIONS` instead. I haven't checked our historically supported list, but even some V8 options could be passed to Node that way. `--perf-prof`yes, but `--prof` no.

boneskull:
@juergba it's because `--prof` is not a Node option _per se_--it's a V8 option.  Node supports a subset of these "natively", and a subset of those in `NODE_OPTIONS`.

Further, `NODE_OPTIONS` and the allowed command-line flags are not 1:1.  For example, `--preserve-symlinks` is a command-line option for node, but it is _not_ allowed in `NODE_OPTIONS`.

For this reason, I don't think we can just do away with it entirely and demand users use `NODE_OPTIONS`.

boneskull:
Either way, sounds like we have some agreement that we don't want to merge this.  

The reasoning is:

1. There are multiple workarounds; `--v8-prof` and `NODE_OPTIONS=--prof`
1. The regex is "historical" and should not be updated (I didn't check to see if this is actually in the code comments there, but it should be).

Thank you!

PR #4443, opened 2020-09-12
Update eslint version

Donghoon759:
### Requirements

### Description of the Change

Updating ESLint version from v6.8.0 to v7.8.1

### Alternate Designs

N/A

Because it is updating ESLint version of mocha core.

### Benefits

ESLint v7.0.0 drops support for Node.js v8 as well. After updating, it will be possible to update other plugins.

### Possible Drawbacks

Lint rules are changed.

### Applicable issues

N/A

### DIFF
--- a/test/browser-specific/fixtures/esm.fixture.mjs
+++ b/test/browser-specific/fixtures/esm.fixture.mjs
@@ -1,2 +1,3 @@
+/* eslint-disable-next-line import/no-absolute-path */
 import '/base/mocha.js';
 window.MOCHA_IS_OK = true;
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -342,7 +342,7 @@ exports.canonicalize = function canonicalize(value, stack, typeHint) {
       });
       break;
     case 'function':
-      /* eslint-disable guard-for-in */
+      /* eslint-disable-next-line no-unused-vars */
       for (prop in value) {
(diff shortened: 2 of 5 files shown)


test/browser-specific/fixtures/esm.fixture.mjs
```diff
@@ -1,2 +1,3 @@
+/* eslint-disable-next-line import/no-absolute-path */
 import '/base/mocha.js';
```
Donghoon759, line 2: @boneskull I wonder the purpose of this line. When I tested, I have deleted the line, changed to refer to another file or missing file. But the test always passed. So I just left the linting issue by disabling it. Thank you for reading.

boneskull: `/base/mocha.js` is the path to the Mocha source as loaded by `karma-mocha` and served by karma's webserver.  we need to assert that we can import mocha via ESM in a browser, and this is the path where it's located when running under karma.

the eslint directive sounds appropriate, then.

boneskull (approved):

test/browser-specific/fixtures/esm.fixture.mjs
Donghoon759: @boneskull Thanks a lot!

PR #4446, opened 2020-09-14
do not leak polyfills into global context; closes #4423

### LINKED ISSUE #4423: Mocha 8.1.0 adds forEach property to CSSStyleDeclaration, at least inside a Proxy
### Prerequisites

### Description

Since mocha 8.1.0, CSSStyleDeclaration has a forEach property, at least inside a Proxy.

This is not the case in mocha 8.0.1.

### Steps to Reproduce

Checkout https://github.com/edwinm/carbonium/commit/0ea6c45b42e6fea28720b459dbe29820ad66b5ef :

```
git clone https://github.com/edwinm/carbonium.git
cd carbonium
git checkout 0ea6c45b42e6fea28720b459dbe29820ad66b5ef
```

Run `npm i`

Run `npm test`

All test pass

Install mocha 8.0.1: `npm i @types/mocha@8.0.1 mocha@8.0.1`

In `src/carbonium.ts` on line 146, change

-- if ("forEach" in target && !(target instanceof CSSStyleDeclaration)) {

to 

++ if ("forEach" in target) {

Now my bugfix is removed.

Run `npm test`. All tests pass.

Install mocha 8.1.0: `npm i mocha@8.1.0`

Run `npm test`. 

**Actual behavior:** 

One test fails.

**Expected behavior:**

All tests pass.

**Reproduces how often:**

Always.

**Context**

Cambrium is like jQuery. It uses Proxy to merge a list of elements and element properties together.

It uses `"forEach" in target` to see whether we're dealing with a list of elements or element properties.

This test fails in mocha when it encouters `element.style`, because `CSSStyleDeclaration` now also has a `forEach` property.

I fixed this by also checking for `target instanceof CSSStyleDeclaration`.

It is still strange that in mocha 8.1.0, CSSStyleDeclaration suddenly has a `forEach` property.

This is a fringe issue and fixed on my end. The bug might still trigger problems in other (not fringe) code
and it seems like a good idea to figure out why this happens.

### Versions

- The output of `npx mocha --version` and `node node_modules/.bin/mocha --version`:
both 8.1.0 (same bug in 8.1.2)
v14.5.0
- Your operating system
  - name and version:
macOS Catalina 10.15.6
64
zsh
Chrome, Firefox
TypeScript 3.9.7

boneskull:
when we transitioned to Rollup/Babel, we misconfigured Babel such that all language APIs polyfilled by `core-js` would be accessible not just by Mocha, but by tests and code under test.

This change attempts to "restrict" the polyfills to Mocha itself.

I'm not entirely sure the best way to test this, since our tests also get bundled and would thus have access to our polyfills.  We need to basically add a separate script to a Karma config which would get loaded via `<script>` tag to make any assertions about the global environment.  _I am too lazy to do this rn._

Will it work with IE?  Let's find out.

Bonus: 50k bundle size reduction!


boneskull:
still misconfigured... i hate javascript

boneskull:
yeah, I am not sure how to do this.  using the "runtime" babel helper, doesn't make sense, because _everything_ must be bundled (no externals).  using "bundled" pollutes the globals.

Munter:
Ugh. Babel is clearly not built for this. It's possible we should try buble instead. That only handles syntax though. Then we'd have to curate our polyfills ourselves. But it seems preferable to the magic that's keeping us from getting the output we want right now

boneskull:
I think I could curate the polyfills w/o needing to swap out babel. it just seems incredibly weird that babel won't handle this for us. 

afaict there are some movements in this direction (babel-polyfill-corejs3 for instance) but I haven't been able to get it working. seems experimental at this point, but I _think_ that's the problem trying to be solved. 

configuring babel is so much not fun. it is very low-level, and I think our use-case is unique enough to make something higher-level out of reach as wel

JoshuaKGoldberg:
Closing to keep our PR queue clean as there are a lot of merge conflicts. But we'll take another look at the underlying issue when we have time - reducing the size & clobbering-ness of polyfills is going to be a nice thing for us to do.

PR #4447, opened 2020-09-15
feat: support running with TrustedTypes enforced

rictic:
### Description of the Change

Trusted Types is a new Content Security Policy specification,
currently implemented in browsers based on Chromium 83 or higher, which
requires that data passed to APIs which may result in arbitrary code
execution must go through an explicit policy. This helps to catch
unintended use of dangerous APIs, and reduces the surface area for
some security reviews.

I'm not sure if test infrastructure like mocha is a likely target
for attack – seems like in most cases an attacker could only access test
data, and it is rare for tests to handle untrusted data. However,
there's value for infrastructure to be compatible with running with
Trusted Types enabled, as it will allow users to write tests to ensure
that the code under test can run with Trusted Types.

This change creates and applies policies for the two places in mocha
that call innerHTML, and adds a temporary patch to the rollup build.
With those changes in place, we can run mocha's karma tests with
Trusted Types enabled (save for the one test that runs with requirejs,
which relies on eval).

### Alternate Designs

It would be possible to integrate dompurify into the HTML reporter, which would make it more secure. Alternatively, it may be possible to do a larger refactoring to construct the HTML output in a more structured way, rather than using string concatenation of HTML.

This must be in core in order for these parts of core to be trusted types compatible.

### Benefits

Users will be able to test their code with mocha in a browser with native Trusted Types enforcement enabled.

### Possible Drawbacks

This change adds additional code, which adds weight to the mocha package for users, and maintenance burden on the mocha core devs.

Our own dependencies may inadvertently break us by doing otherwise safe operations (e.g. eval(str), Function(str), setting innerHTML, etc).

This is an enhancement (minor release).

### More info and related work

* Spec: https://w3c.github.io/webappsec-trusted-types/dist/spec/#introduction
* Related PR adding support to karma: https://github.com/karma-runner/karma/pull/3360

### DIFF
--- a/lib/browser/highlight-tags.js
+++ b/lib/browser/highlight-tags.js
@@ -25,7 +25,7 @@ function highlight(js) {
     );
 }
 
-var highlightPolicy = {
+let highlightPolicy = {
   createHTML: function(html) {
     // The highlight function escapes its input.
     return highlight(html);
--- a/lib/browser/highlight-tags.js
+++ b/lib/browser/highlight-tags.js
@@ -25,6 +25,22 @@ function highlight(js) {
     );
 }
 
+var highlightPolicy = {
+  createHTML: function(html) {
+    // The highlight function escapes its input.
+    return highlight(html);
+  }
+};
+if (
+  typeof window !== 'undefined' &&
+  typeof window.trustedTypes !== 'undefined'
+) {
+  highlightPolicy = window.trustedTypes.createPolicy(
+    'mocha-highlight-tags',
+    highlightPolicy
+  );
+}
+
 /**
  * Highlight the contents of tag `name`.
  *
@@ -34,6 +50,6 @@ function highlight(js) {
 module.exports = function highlightTags(name) {
   var code = document.getElementById('mocha').getElementsByTagName(name);
   for (var i = 0, len = code.length; i < len; ++i) {
-    code[i].innerHTML = highlight(code[i].innerHTML);
+    code[i].innerHTML = highlightPolicy.createHTML(code[i].innerHTML);
   }
 };
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -313,6 +313,27 @@ function error(msg) {
   document.body.appendChild(fragment('<div id="mocha-error">%s</div>', msg));
 }
 
+var policy = {
+  createHTML: function(html) {
+    /**
+     * Note that this policy lets html through unchanged. This is potentially
+     * a security vulnerability if untrusted data is set to innerHTML, as it
+     * allows arbitrary code execution.
+     *
+     * Ideally this code would be refactored to not use .innerHTML, and this
+     * policy deleted, or this policy could return the html after it has been
+     * processed by a secure sanitization system like dompurify
+     */
+    return html;
+  }
+};
+if (
+  typeof window !== 'undefined' &&
+  typeof window.trustedTypes !== 'undefined'
+) {
+  policy = window.trustedTypes.createPolicy('mocha-html-reporter', policy);
+}
+
 /**
  * Return a DOM fragment from `html`.
  *
@@ -323,15 +344,17 @@ function fragment(html) {
   var div = document.createElement('div');
   var i = 1;
 
-  div.innerHTML = html.replace(/%([se])/g, function(_, type) {
-    switch (type) {
-      case 's':
-        return String(args[i++]);
-      case 'e':
-        return escape(args[i++]);
-      // no default
-    }
-  });
+  div.innerHTML = policy.createHTML(
+    html.replace(/%([se])/g, function(_, type) {
+      switch (type) {
+        case 's':
+          return String(args[i++]);
+        case 'e':
+          return escape(args[i++]);
+        // no default
+      }
+    })
+  );
 
   return div.firstChild;
 }
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -327,7 +327,7 @@ let policy = {
     return html;
   }
 };
-if (window.trustedTypes != null) {
+if (typeof window !== 'undefined' && window.trustedTypes != null) {
   policy = window.trustedTypes.createPolicy('mocha-html-reporter', policy);
 }
 
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -313,7 +313,7 @@ function error(msg) {
   document.body.appendChild(fragment('<div id="mocha-error">%s</div>', msg));
 }
 
-var policy = {
+let policy = {
   createHTML: function(html) {
     /**
      * Note that this policy lets html through unchanged. This is potentially
@@ -327,10 +327,7 @@ var policy = {
     return html;
   }
 };
-if (
-  typeof window !== 'undefined' &&
-  typeof window.trustedTypes !== 'undefined'
-) {
+if (window.trustedTypes != null) {
   policy = window.trustedTypes.createPolicy('mocha-html-reporter', policy);
 }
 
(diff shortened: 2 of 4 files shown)


rollup.config.js
```diff
@@ -11,6 +12,36 @@ import visualizer from 'rollup-plugin-visualizer';
...
+ * whenever we would run `Function("return this")()` we can instead first look
+ * whether `globalThis` is defined, and if so, just use that.
+ *
+ * The latest version of regenerator-runtime does rely on calling Function
+ * to get globalThis, so we only need this plugin until the updated version
+ * has percolated through our dependency tree. We can try to remove it on
+ * 2021-01-01. This behavior is tested, so we can just remove the plugin
+ * from our array and try `npm test`. If the tests pass, this can be removed.
+ */
+const applyTemporaryCspPatchPlugin = {
```
boneskull, line 34: Seems like others would have this problem.  Is there not an "official" fix anywhere?

lib/reporters/html.js
```diff
@@ -313,6 +313,27 @@ function error(msg) {
...
+     * allows arbitrary code execution.
+     *
+     * Ideally this code would be refactored to not use .innerHTML, and this
+     * policy deleted, or this policy could return the html after it has been
+     * processed by a secure sanitization system like dompurify
+     */
+    return html;
+  }
+};
+if (
```
boneskull, line 330: fwiw, `window` will not be `undefined`; the `html` reporter will throw mad exceptions without `window`.

also, you can use `const` or `let` instead of `var` across the codebase now

boneskull (commented):
Thanks!  Seems like a good addition.  Unfortunately, I don't know anything about Trusted Types, so I think I need to educate myself or defer the decision to another maintainer who does (e.g., @Munter?).

My main question is: if a test is written using `eval()`, and we ship this change, will the test now fail when run in a Trusted-Types-supporting browser?  If so, this behavior must be _opt-in_.  And _if_ it needs to be opt-in, does this defeat the purpose?

boneskull:
Further, I think it'd be good to see a test which asserts the behavior works as intended.

lib/reporters/html.js
rictic: +1, switched these changes to use `let`

Dropped the `window` check here. Should I drop it from highlight-tags.js as well? If someone is importing highlight-tags but not calling it from node that could fail, because we're eagerly initializing the policy.

rictic:
> My main question is: if a test is written using eval(), and we ship this change, will the test now fail when run in a Trusted-Types-supporting browser? If so, this behavior must be opt-in. And if it needs to be opt-in, does this defeat the purpose?

Trusted type enforcement is opt in. Like other CSP features, you enable it with either an HTTP header or with a `<meta>` tag in the page's `<head>`.

The change should have no effect on existing tests, which won't have opted into enforcing Trusted Types.

Shipping this change will make mocha compatible with trusted types, so it will enable testing with mocha with TT enforced. See e.g. https://github.com/Polymer/lit-html/pull/1323

> Further, I think it'd be good to see a test which asserts the behavior works as intended.

See the changes to karma.conf.js, which adds headers to enable Trusted Type enforcement for mocha's existing browser tests (in supporting browsers, which is currently browsers based on a recent Chromium like Brave, Chrome, and Edge).

lib/reporters/html.js
rictic: Looks like the window check is needed, tests fail without it: https://github.com/mochajs/mocha/pull/4447/checks?check_run_id=1171116538

I think the html reporter is sometimes imported from a node environment even if it isn't used.

boneskull: yeah, that's a bug.  ok. ty

boneskull:
>> Further, I think it'd be good to see a test which asserts the behavior works as intended.
> See the changes to karma.conf.js, which adds headers to enable Trusted Type enforcement for mocha's existing browser tests (in supporting browsers, which is currently browsers based on a recent Chromium like Brave, Chrome, and Edge).

sorry, what I meant was that the existing tests assert compatibility, but we do not have:

- a test asserting some sort of exception is thrown when unsafe code is run
- a test asserting no such exception is thrown when unsafe code is run

boneskull:
(AppVeyor is no longer building our PRs; I've removed it from the list of checks)

rictic:
I'm not 100% sure I follow. You'd like to see a test of native Trusted Types functionality? A test that only runs in Chromium browsers and would fail if the trusted types headers in karma.conf.js didn't result in TT being enforced?

boneskull:
> I'm not 100% sure I follow. You'd like to see a test of native Trusted Types functionality? A test that only runs in Chromium browsers and would fail if the trusted types headers in karma.conf.js didn't result in TT being enforced?

Yes, essentially:

- a test that runs in Chromium that asserts "unsafe" code will break
- a test that runs either in Chromium w/ the headers disabled, or somewhere else that lacks support, and asserts "unsafe" code does not break

The point is not to test the browser feature--I'm sure Chromium has plenty of tests--but rather to establish a baseline of behavior which Mocha will exhibit when encountering these configurations.  We need to make sure that future changes in Mocha do not violate the assumptions of its behavior established here.

outsideris:
I need to run TrustedTypes.

boneskull:
@rictic are you able to continue on this PR?

rollup.config.js
JoshuaKGoldberg: Speaking from the perspective of a reviewer in September 2020: I'd want a comment in the code pointing to the relevant tracking issue. That way it's easier to get context on it and check whether it's been resolved.

Speaking from the perspective of a reviewer now, in 2024: I'm betting this has since been resolved. 😄

lib/browser/highlight-tags.js
```diff
@@ -25,6 +25,22 @@ function highlight(js) {
     );
 }
 
+let highlightPolicy = {
+  createHTML: function(html) {
+    // The highlight function escapes its input.
+    return highlight(html);
+  }
```
JoshuaKGoldberg, line 32: [Style] Nit: no need for this wrapper, I think?

```suggestion
  // The highlight function escapes its input.
  createHTML: highlight,
```

lib/reporters/html.js
```diff
@@ -313,6 +313,24 @@ function error(msg) {
...
+     * Note that this policy lets html through unchanged. This is potentially
+     * a security vulnerability if untrusted data is set to innerHTML, as it
+     * allows arbitrary code execution.
+     *
+     * Ideally this code would be refactored to not use .innerHTML, and this
+     * policy deleted, or this policy could return the html after it has been
+     * processed by a secure sanitization system like dompurify
+     */
+    return html;
+  }
```
JoshuaKGoldberg, line 328: [Bug] Indeed, this looks like it _partially_ (not completely) bypasses the point of Trusted Types. Which worries me. If Mocha were to ship with this change as-is and claim to supported Trusted Types, that claim would be a little misleading.

I'd think making this truly "safe" should be a blocker to the PR. But I'm very open to being convinced otherwise if I'm off.

JoshuaKGoldberg (changes requested):
Looks like a good start & most of the way there! Requesting changes on the HTML reporter, with the note that I'm not confident. 🙂

JoshuaKGoldberg:
👋 coming back to this @rictic, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

justinfagnani:
I would still like to see this resolved, especially as Webkit has decided to support the standard as of two weeks ago: https://github.com/WebKit/standards-positions/issues/186#issuecomment-2172422968 and Mozilla as of Dec 2023: https://github.com/mozilla/standards-positions/issues/20#issuecomment-1853427823

PR #4450, opened 2020-09-16
Implement Open Collective categories for extended filtering

znarf:
At Open Collective, we're starting to implement a few changes to help Collectives like yours filter undesirable contributions.

The goal is double:
- help you filter contributions and protect your image
- cut the incentive for these people to contribute in the first place

I'm aware of your previous efforts in this direction and I'm just extending it here, based on a categorization system that we started implementing.

Happy to discuss and iterate.


boneskull (commented):
Thanks!  I'll test this out, but otherwise lgtm

boneskull:
works, thanks

Issue #4306, opened 2020-05-28
refactor: utils.type()


boneskull:
`utils.type(value)` is a function which returns a human-readable "type" of whatever value you pass it.

About `type()`:

- It works by first comparing against a set of tricky values (`undefined`, `null`, etc); if it does not find these, it then looks at the result of `Object.prototype.toString.call(value)` (e.g., `[Object object]`) and extracting the second token from within the brackets
- It's pretty useful, as we get things like `date` for `type(new Date())` and `error` for `type(new Error())`; more granularity than `typeof` offers
- Originally, this was used for object comparison--specifically, displaying diffs
- Over time, we (I?) started using it elsewhere, because of the convenience it offers: a single function to determine the type of a given value, instead of several different methods, which invariably turn into sprawling if/else-if/else statements:
  - `Array.isArray(value)` for arrays (where available)
  - `value === null` for `null`
  - `typeof value` for most other stuff

With additions of async functions and ESM, the `Object.prototype.toString` strategy adds new potential return values of `type()`:
  - `type(async function() {})` returns `asyncfunction`
  - given module `n` imported via `let n = await import(name)`, `type(n)` returns `module`

While the new values may be good for displaying diffs, they are _not helpful in the general case_.  For example, a return value of `asyncfunction` will tell us that `value` is a function which returns a `Promise`, but a function needn't be `async` to return a `Promise`, so it adds little value. We have no known use case for a return value of `module` outside of diffing (and maybe not even then--we should probably play with that one in our test suite).

This is what I'd like to propose:

- `type()` should be renamed to e.g., `canonicalType()`
- its use should be limited to `utils.canoncalize()`, which is part of the comparison/diff code
  - we may want to actually pull the diffing-related functions into its own module, just for sake of organization.  `utils` is just a code dump, as you'd expect.
- a new function (`type()`?) should be created to be used elsewhere in the codebase
- the new function should either be:
    - a custom implementation tailored the various use-cases around the codebase, or
    - a thin wrapper around a third-party type-checking library
- usages of `typeof`, `Array.isArray()`, around the codebase should be replaced with this new function, for consistency

Happy to hear alternatives, as long as those alternatives do not involve "rewrite Mocha in TypeScript" :trollface:

peanutenthusiast:
Hi @boneskull I can look into this.

PR #4457, opened 2020-10-01
refactor: utils.type()

ValeriaVG:
## Description of the Change
- Existing `utils.type()` is renamed to `utils.canonicalType()`
- Call from `utils.canonise()` is changed to `canonicalType()`
- New function `type()`added to the `utils` implementing generic type retrieval based on [MDN:DataStructures](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures) and several additional cases: Error => 'error' and String = > 'string', aiming to replace `Array.isArray()`, `typeof` and `Buffer.isBuffer`.

### Alternate Designs
Diff functionality, as suggested in the related issue, should be separated in its own module. 
The only difference is that this PR implies incremental changes.

An alternative could be to instead implement `isType(value, type)` due to the use intentions.

- It is part of the core functionality

### Benefits
- Lays foundation for future `utils` refactoring
- Adds ability to perform strict and generic type comparisons
- 

### Possible Drawbacks
- The only file that uses new `utils.type()` is `lib/node/serializer.js`

### Applicable issues
https://github.com/mochajs/mocha/issues/4306

### DIFF
--- a/lib/nodejs/serializer.js
+++ b/lib/nodejs/serializer.js
@@ -190,7 +190,8 @@ class SerializableEvent {
       parent[key] = Object.create(null);
       return;
     }
-    if (type(value) === 'error') {
+    let _type = type(value);
+    if (_type === 'error') {
       // we need to reference the stack prop b/c it's lazily-loaded.
       // `__type` is necessary for deserialization to create an `Error` later.
       // `message` is apparently not enumerable, so we must handle it specifically.
@@ -200,10 +201,11 @@ class SerializableEvent {
         __type: 'Error'
       });
       parent[key] = value;
-      // after this, the result of type(value) will be `object`, and we'll throw
+      // after this, set the result of type(value) to be `object`, and we'll throw
       // whatever other junk is in the original error into the new `value`.
+      _type = 'object';
     }
-    switch (type(value)) {
+    switch (_type) {
       case 'object':
         if (type(value.serialize) === 'function') {
           parent[key] = value.serialize();
--- a/lib/nodejs/serializer.js
+++ b/lib/nodejs/serializer.js
@@ -6,7 +6,7 @@
 
 'use strict';
 
-const {canonicalType: type} = require('../utils');
+const {type} = require('../utils');
 const {createInvalidArgumentTypeError} = require('../errors');
 // this is not named `mocha:parallel:serializer` because it's noisy and it's
 // helpful to be able to write `DEBUG=mocha:parallel*` and get everything else.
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -152,6 +154,49 @@ var type = (exports.type = function type(value) {
...
+ * type(new Date()) // 'object'
+ * type(/foo/) // 'object'
+ * type('type') // 'string'
+ * type(global) // 'object'
+ * type(new String('foo') // 'string'
+ */
+exports.type = function type(value) {
+  // Null is special
+  if (value === null) return 'null';
+  var primitives = [
```
boneskull, line 180: you can use a `Set` here.

lib/runner.js
```diff
@@ -1171,7 +1171,11 @@ function isError(err) {
  */
 function thrown2Error(err) {
   return new Error(
-    'the ' + type(err) + ' ' + stringify(err) + ' was thrown, throw an Error :)'
+    'the ' +
```
boneskull, line 1174: you can use a template literal here

lib/nodejs/serializer.js
```diff
@@ -190,7 +190,7 @@ class SerializableEvent {
       parent[key] = Object.create(null);
       return;
     }
-    if (type(value) === 'error' || value instanceof Error) {
+    if (type(value) === 'error') {
```
boneskull, line 193: good

lib/utils.js
```diff
@@ -152,6 +154,49 @@ var type = (exports.type = function type(value) {
...
+ * type(false) // 'boolean'
+ * type(Infinity) // 'number'
+ * type(null) // 'null'
+ * type(new Date()) // 'object'
+ * type(/foo/) // 'object'
+ * type('type') // 'string'
+ * type(global) // 'object'
+ * type(new String('foo') // 'string'
+ */
+exports.type = function type(value) {
```
boneskull, line 177: This should be marked `@private` unless there's reason to make it public

boneskull (changes requested):
Thanks!  Looks really good.  Just had a couple comments.  "Modern" ES syntax is now usable throughout the codebase, so that should simplify/improve some of the changes here.

ValeriaVG:
Thank you, makes total sense.
Changes made:
- `utils.type()` now uses `Set` instead of an array and `const` instead of `var`
- `runner.thrown2Error()` now uses template literal
- `serializer._serialize()` now stores computed type value and explicitly sets it to `object`, when Error is being transformed to an object

boneskull (approved):
LGTM! tyvm

PR #4459, opened 2020-10-02
run all node.js tests on GHA

boneskull:
In #4402, @outsideris moved the build off of AppVeyor and onto GHA.

This PR moves _Linux Node.js_ builds _off of Travis-CI_ and on to GHA.  They are run in the same matrix as the existing Windows builds.  Included are the lint checks and coverage reporting to Coveralls.

I've consolidated a few of the existing jobs to reduce waiting / increase parallelization.  

Remaining on Travis-CI are the browser tests, as I'm currently struggling to get SauceLabs tests running in GHA's env.


boneskull:
See #4460 for browser tests on GHA

.github/workflows/mocha.yml
```diff
@@ -0,0 +1,158 @@
...
+        run: |
+          echo "::set-output name=dir::$(npm config get cache)"
+      - uses: actions/checkout@v2
+      - name: 'Cache node_modules'
+        uses: actions/cache@v2
+        with:
+          path: '${{ steps.npm-cache.outputs.dir }}'
+          key: "${{ matrix.os }}-node-v${{ matrix.node }}-${{ hashFiles('**/package-lock.json') }}"
+          restore-keys: |
+            ${{ matrix.os }}-node-v${{ matrix.node }}-
```
outsideris, line 141: When I test cache on windows before, it took about 80s to download over 600MB cache files.

In this GHA, "Install Dependencies" took like below without cache.  So, I worried about using cache takes more time.

### Ubuntu
```
Run npm ci --ignore-scripts
added 2827 packages in 24.523s
```
### Windows
``` 
Run npm ci --ignore-scripts
added 2827 packages in 65.224s
```

outsideris (approved):
LGTM except Node.js 10 and 12 disabled.

.github/workflows/mocha.yml
boneskull: I'll profile this and see if we should be caching or no

.github/workflows/mocha.yml
```diff
@@ -0,0 +1,158 @@
...
+    env:
+      NODE_OPTIONS: '--trace-warnings'
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-2019
+        node:
+          # - 10
+          # - 12
```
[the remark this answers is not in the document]
boneskull: good catch ty

outsideris:
Interesting. 
[Restore cache](https://github.com/mochajs/mocha/runs/1211398664?check_suite_focus=true) took under 5s with 53MB on Ubuntu, but it took over 1m with 654MB on windows.

.github/workflows/mocha.yml
```diff
@@ -0,0 +1,157 @@
...
+        if: "${{ matrix.os == 'windows-2019' }}"
+        uses: actions/cache@v2
+        with:
+          path: GrowlInstaller
+          key: '${{ runner.os }}-growl-installer'
+          restore-keys: |
+            ${{ runner.os }}-growl-installer
+      - name: Add Growl Installer to Path (Windows)
+        if: "${{ matrix.os == 'windows-2019' }}"
+        run: echo "C:\Program Files (x86)\Growl for Windows" >> $GITHUB_PATH
```
outsideris, line 108: Nice. [I just found the deprecating announcement](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

.github/workflows/mocha.yml
boneskull: turns out windows has a cache of like 600MiB, and linux is 50MiB, so that's why.

I changed this to use the cache for linux but not windows.

.github/workflows/mocha.yml
boneskull: I was unable to get this working.  [Going to try this suggestion now](https://github.community/t/migration-from-deprecated-add-path-on-windows/136265)

boneskull:
Still unable to get the new "environment file" working on Windows, so merging as-is.

This PR adds inline annotations for Mocha tests and ESLint checks.  This means that you'll see test and lint failures pinned to the relevant line numbers _inline, in the diff_ on GitHub.

Of note, the Mocha annotations arrive via https://github.com/daniellockyer/mocha-github-actions-reporter.  This reporter is _not_ added to our `devDependencies`, and is only installed on GHA as a separate step

Issue #4456, opened 2020-10-01
parallel workers should have an ID that can be used to prevent conflicts among workers


catdad:
**Is your feature request related to a problem or a nice-to-have?? Please describe.**
I have a problem. I need to start a server on a pre-know port (due to limitations of a third-party library). I usually hard-code this port, but when running parallel tests, I need to start multiple unique servers -- one for each worker. I can't just check for a free port due to race conditions. Usually, parallel systems such as this have a way to identify each workers -- for example, and unique integer ID is given to each worker.

**Describe the solution you'd like**
An unique ID integer would be great. This can be passed to the worker with an environment variable (`MOCHA_WORKER_ID`) or just as an extra value in the worker arguments (accessible with `process.argv[2]`). Really, any acceptable solution to identify unique workers.

**Describe alternatives you've considered**
I've considered using just a random port every time, but that gets problematic because I don't know how long any given port will be used for due to different lengths of tests, or when a port could conflict with something else running on the operating system. I also considered using a file on the file system to identify used/free resources, but there are similarly many race conditions to consider when multiple processes write to a single file.

**Additional context**
Here is the same request in `jest`: https://github.com/facebook/jest/issues/2284

rgroothuijsen:
If registering workers with a unique ID is all that's required, one option would be to use dedicated cluster orchestration tools such as [ZooKeeper](https://zookeeper.apache.org/). They're designed to handle the race conditions you describe.

catdad:
I'm not really describing cluster orchestration. In fact, my software has nothing to with clustering. And I feel like integrating a complicated orchestration framework in order to write trivial tests is not feasible. Providing an ID for test workers is a very common for parallel testing libraries. Am I missing something?

outsideris:
Couldn't you use a kind of [get-port](https://www.npmjs.com/package/get-port) module to address your issue?
Otherwise, how will you use `MOCHA_WORKER_ID` for your available port?

catdad:
I actually did use that module, and there's a problem with doing so. This module and others like it work in a deterministic way every time. They start at one port, and just increment up until they find a free port. That means that when you first start several workers, they will all find the same free port. Then when you try to start the server, only one of the workers will actually succeed starting on that port.

[...]

PR #4463, opened 2020-10-03
adding a MOCHA_WORKER_ID environment variable, unique for each worker

catdad:
### Requirements

### Description of the Change

This change addresses #4456. It adds a `MOCHA_WORKER_ID` environment variable that is unique for each worker, so that it may be used for by any test resources that need to use global resources (network ports, filesystem paths, etc.) but ensure that these are unique in order to avoid conflicts during parallel tests.

### Alternate Designs

I considered using `forkArgs` instead of an environment variable, so that an ID may be accessed using `process.argv[2]`. I did not want to do this in case adding arguments was necessary for internal uses in the future.

It is common for tests to use global resources like network ports, which can conflict during parallel testing. Without a feature like this in core, users would be left on their own to create a home-brewed solution (using filesystem files as signals, sockets that elect a leader in the cluster so that one process hands out all global data, a standalone cluster orchestration system), all of which add a lot of complication to writing tests. This would also be a difference when writing tests that run in series vs. parallel, as this concern would not exist (and potentially needs to be turned off) for test in series, and would add pain points when taking existing serial tests and having them run in parallel. This makes using mocha significantly less fun.

This feature takes that pain point away. For example, allocating ports would once again be trivial, using `const port = 1234 + Number(process.env.MOCHA_WORKER_ID || 0)`.

Other parallel testing frameworks, like jest, already use this approach: https://github.com/facebook/jest/issues/2284

### Benefits

This solves common pain points for developers writing tests.

### Possible Drawbacks

Other than more complexity, I can't think of a drawback for the feature itself.

Since `workerpool` does not itself provide an option for worker-specific options or to execute code deterministically in a specific worker (only an option to add to the queue to execute on any available worker), I had to use a workaround in the implementation. This workaround could make internal updates more difficult and adds extra requirements to replacing `workerpool` if a need arises to use a different implementation internally. I personally believe that this tradeoff is worth it, though I am probably biased here.

### Applicable issues

#4456

* Is it an enhancement (minor release)? **Yes**
* Is it a bug fix, or does it not impact production code (patch release)? **No**

### DIFF
--- a/lib/nodejs/buffered-worker-pool.js
+++ b/lib/nodejs/buffered-worker-pool.js
@@ -79,7 +79,7 @@ class BufferedWorkerPool {
 
     this.options = Object.assign({}, WORKER_POOL_DEFAULT_OPTS, opts, {
       maxWorkers,
-      forkOpts: Object.assign({}, WORKER_POOL_DEFAULT_OPTS || {}, {
+      forkOpts: Object.assign({}, WORKER_POOL_DEFAULT_OPTS.forkOpts || {}, {
         env: Object.defineProperties({ ...process.env }, {
           MOCHA_WORKER_ID: {
             enumerable: true,
--- a/lib/nodejs/buffered-worker-pool.js
+++ b/lib/nodejs/buffered-worker-pool.js
@@ -75,9 +75,20 @@ class BufferedWorkerPool {
       process.execArgv.join(' ')
     );
 
+    let count = 0;
+
     this.options = Object.assign({}, WORKER_POOL_DEFAULT_OPTS, opts, {
-      maxWorkers
+      maxWorkers,
+      forkOpts: Object.assign({}, WORKER_POOL_DEFAULT_OPTS || {}, {
+        env: Object.defineProperties({ ...process.env }, {
+          MOCHA_WORKER_ID: {
+            enumerable: true,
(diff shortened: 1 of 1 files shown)


catdad:
I submitted this as a draft for now because I have not gotten a chance to write tests yet.

I am also interested in getting some feedback as early as possible in the PR process. Since `workerpool` does not have the option to add per-worker options, or even directly execute code on a specific worker, I had to get a little bit creative in the implementation.

boneskull:
Have you tried using the PID, e.g.:

```js
const port = 1234 + Number(process.pid || 0)
```

catdad:
The randomness of PIDs makes this undesirable to me. A PID can cause a conflict with an existing port being in use on the system, and this will not be consistent across test runs. I can run the tests successfully once, then get a conflict the second time, then run them successfully on the third, etc. A solution for this should be deterministic and repeatable (i.e. a parallel test run with 3 jobs will always get the numbers `[1, 2, 3]` every time, albeit not necessarily the same number for the same exact test each time).

forty:
For reference another use case where deterministic IDs would be desirable https://github.com/mochajs/mocha/issues/4456#issuecomment-844924184

juergba:
superseded by #4813

PR #4464, opened 2020-10-05
Change serializer errors to use error codes

evaline-ju:
### Description of the Change

After the implementations of https://github.com/mochajs/mocha/pull/3467 and https://github.com/mochajs/mocha/pull/3666, there are a few instances in the code that do not use error codes yet. This PR addresses the instances in the serializer code and does not cover all remaining instances of lack of error codes (I mainly wanted to make sure I was on the right track here before trying to change as many remaining errors as possible). 

### Alternate Designs

N/A

Using codes is a best practice like in Node.js, codes help identify errors as mocha errors at a glance

### Benefits

Using codes is a best practice like in Node.js, codes help identify errors as mocha errors at a glance

### Possible Drawbacks

None I'm aware of

### Applicable issues

Contributes to https://github.com/mochajs/mocha/issues/3656, semver-patch

### DIFF
--- a/lib/nodejs/serializer.js
+++ b/lib/nodejs/serializer.js
@@ -7,7 +7,10 @@
 'use strict';
 
 const {type} = require('../utils');
-const {createInvalidArgumentTypeError} = require('../errors');
+const {
+  createInvalidArgumentTypeError,
+  createInvalidArgumentValueError
+} = require('../errors');
 // this is not named `mocha:parallel:serializer` because it's noisy and it's
 // helpful to be able to write `DEBUG=mocha:parallel*` and get everything else.
 const debug = require('debug')('mocha:serializer');
@@ -131,7 +134,12 @@ class SerializableEvent {
    */
   constructor(eventName, originalValue, originalError) {
     if (!eventName) {
-      throw new Error('expected a non-empty `eventName` string argument');
+      throw createInvalidArgumentValueError(
+        'Empty `eventName` string argument',
+        'eventName',
+        eventName,
+        'is empty'
+      );
     }
     /**
      * The event name.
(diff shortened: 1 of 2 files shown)


lib/nodejs/serializer.js
```diff
@@ -131,7 +134,12 @@ class SerializableEvent {
    */
   constructor(eventName, originalValue, originalError) {
     if (!eventName) {
-      throw new Error('expected a non-empty `eventName` string argument');
+      throw createInvalidArgumentValueError(
```
boneskull, line 137: if `eventName` is `undefined` (e.g., someone wrote `new SerializableEvent()`) then one could argue this is  an invalid _type_ (it should be a `string`; not `undefined`) instead of an invalid _value_.  contrast that with passing `''` for `eventName`--the type is correct but the value is invalid.  in this case, we're checking for _falsy_, which is both a type and a value (because JavaScript)--and is intended--so I don't feel strongly that we need two different errors here.  just wanted to mention it

further, these two types of errors are unlikely to be caused by an end-user--it's an error only Mocha contributors should ever encounter.  we don't usually do this sort of exhaustive type-checking unless it's user input

boneskull (approved):
Thanks! LGTM, but had a comment

boneskull:
calling this `semver-minor` as it does not fix any known bug, but is rather additive.

lib/nodejs/serializer.js
evaline-ju: Makes sense to me. I wasn't tied to the two different error types, so I've gone ahead and changed this one to throw a `type` error too.

outsideris (approved):

boneskull:
Thanks!

PR #4465, opened 2020-10-05
make guarantees about orphaned processes

boneskull:
- updates workerpool to v6.0.2, which guarantees child processes exit before `Pool#terminate()` resolves
- cleanup `test/integration/options/parallel.spec.js`

@segrey this will make the guarantee about the worker processes you were asking about in #4403.

### DIFF
--- a/lib/nodejs/buffered-worker-pool.js
+++ b/lib/nodejs/buffered-worker-pool.js
@@ -75,9 +75,7 @@ class BufferedWorkerPool {
       process.execArgv.join(' ')
     );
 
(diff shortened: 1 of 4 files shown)


boneskull:
Ref: josdejong/workerpool#176

boneskull:
@mochajs/core any review will do, this one is not big. [here](https://github.com/mochajs/mocha/pull/4465/files#diff-c6a958cc571c8bf9b758da48555899b1R490) is what you should look at

outsideris (approved):

PR #4478, opened 2020-10-15
chore(deps): chokidar@3.4.3, pin @ungap/promise-all-settled

AviVahl:
### Description of the Change

- Upgrade to `chokidar@3.4.3`.
- Pinned `@ungap/promise-all-settled` for consistency.

### Alternate Designs

Perhaps use `^` in dependencies as well.

New chokidar version fixes infinite loops on circular symlinks: https://github.com/paulmillr/chokidar/blob/master/.github/full_changelog.md

### Benefits

Better deduping in consumer projects, as most of the ecosystem uses ^ and picks up the latest of that major.

### Possible Drawbacks

New versions always have the possibility of a regression.

### Applicable issues

None.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4959,9 +4959,9 @@
       "dev": true
     },
     "chokidar": {
-      "version": "3.4.2",
-      "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-3.4.2.tgz",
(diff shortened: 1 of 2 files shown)


boneskull:
LGTM, thanks.

PR #4486, opened 2020-10-23
fix for issue 4047

krishnq:
CLI will verify if args passed with --file exists and if not, throws error createNoFilesMatchPatternError("File ${FileName} not found")


boneskull:
please read the PR template when sending a PR

PR #4494, opened 2020-11-02
docs: Add example of generating tests with a HOF

maxwellgerber:
### Requirements

### Description of the Change

Quick docs update showing a workaround with some common issues run into w/data-driven tests.

Namely, it's very hard to run a single testcase from an IDE because tests inside a `.forEach` callback aren't detected, Compare
image
with
image

### Alternate Designs

N/A (docs update)

N/A (docs update)

### Benefits
This approach has been useful for me & my team - hope it can be useful for someone else too

### Possible Drawbacks
Slightly more verbose than pure data-driven approach.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -720,7 +720,7 @@ $ mocha
 ```
 
 Tests added inside a `.forEach` handler often don't play well with editor plugins, especially with "right-click run" features.
-Another way to paramaterize tests is by generating them with a higher order function. This example is equivalent to the one above.
+Another way to parameterize tests is to generate them with a closure. This following example is equivalent to the one above:
 
 ```js
 describe('add()', function() {
--- a/docs/index.md
+++ b/docs/index.md
@@ -684,25 +684,23 @@ Given Mocha's use of `Function.prototype.call` and function expressions to defin
 Take the following example:
 
 ```js
-var assert = require('chai').assert;
+const assert = require('chai').assert;
 
 function add() {
-  return Array.prototype.slice.call(arguments).reduce(function(prev, curr) {
-    return prev + curr;
-  }, 0);
+  return Array.prototype.slice.call(arguments).reduce((prev, curr) => prev + curr, 0);
 }
 
-describe('add()', function() {
-  var tests = [
+describe('add()', () => {
+  const tests = [
     {args: [1, 2], expected: 3},
     {args: [1, 2, 3], expected: 6},
     {args: [1, 2, 3, 4], expected: 10}
   ];
 
-  tests.forEach(function(test) {
-    it('correctly adds ' + test.args.length + ' args', function() {
-      var res = add.apply(null, test.args);
-      assert.equal(res, test.expected);
+  tests.forEach(({args, expected}) => {
+    it('correctly adds ' + args.length + ' args', () => {
+      const res = add(...args);
+      assert.equal(res, expected);
     });
   });
 });
@@ -723,13 +721,11 @@ Tests added inside a `.forEach` handler often don't play well with editor plugin
 Another way to parameterize tests is to generate them with a closure. This following example is equivalent to the one above:
 
 ```js
-describe('add()', function() {
-  function testAdd(test) {
-    return function () {
-      var res = add.apply(null, test.args);
-      assert.equal(res, test.expected);
-    };
-  }
+describe('add()', () => {
+  const testAdd = ({args, expected}) => () => {
+    const res = add(...args);
+    assert.equal(res, expected);
+  };
 
   it('correctly adds 2 args', testAdd({args: [1, 2], expected: 3}));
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -719,6 +719,26 @@ $ mocha
     ✓ correctly adds 4 args
 ```
 
+Tests added inside a `.forEach` handler often don't play well with editor plugins, especially with "right-click run" features.
+Another way to paramaterize tests is by generating them with a higher order function. This example is equivalent to the one above.
+
+```js
+describe('add()', function() {
+  function testAdd(test) {
```
boneskull, line 727: Can we use more modern syntax here?  Looks better IMO.  :wink:

```suggestion
  function testAdd({args, expected}) {
    const result = add(...args);
    assert.equal(result, expected);
  }
```

docs/index.md
```diff
@@ -719,6 +719,26 @@ $ mocha
     ✓ correctly adds 4 args
 ```
 
+Tests added inside a `.forEach` handler often don't play well with editor plugins, especially with "right-click run" features.
+Another way to paramaterize tests is by generating them with a higher order function. This example is equivalent to the one above.
```
boneskull, line 723:
```suggestion
Another way to parameterize tests is to generate them with a closure. This following example is equivalent to the one above:
```

Fixed a typo above, but I think "closure" is a more widely-known term in JS

boneskull (changes requested):
Thanks! Good addition. I have a couple suggestions--happy to discuss.

boneskull:
yeah that coveralls failure doesn't make sense. we've had this problem before. unclear to me if the culprit is coveralls or nyc...

docs/index.md
```diff
@@ -684,25 +684,23 @@ Given Mocha's use of `Function.prototype.call` and function expressions to defin
...
     {args: [1, 2, 3], expected: 6},
     {args: [1, 2, 3, 4], expected: 10}
   ];
 
-  tests.forEach(function(test) {
-    it('correctly adds ' + test.args.length + ' args', function() {
-      var res = add.apply(null, test.args);
-      assert.equal(res, test.expected);
+  tests.forEach(({args, expected}) => {
+    it('correctly adds ' + args.length + ' args', () => {
```
boneskull, line 701:
```suggestion
    it(`correctly adds ${args.length} args`, function() {
```

Mocha recommends _not_ using an arrow function for callbacks because of the presence of the context object (`this`).  you _can_ do this if you're not using `this`, but it can lead to subtle bugs in future test maintenance.  best to keep it out of the official docs

docs/index.md
```diff
@@ -719,6 +717,24 @@ $ mocha
     ✓ correctly adds 4 args
 ```
 
+Tests added inside a `.forEach` handler often don't play well with editor plugins, especially with "right-click run" features.
+Another way to parameterize tests is to generate them with a closure. This following example is equivalent to the one above:
+
+```js
+describe('add()', () => {
```
boneskull, line 724:
```suggestion
describe('add()', function() {
```

docs/index.md
```diff
@@ -719,6 +717,24 @@ $ mocha
...
+  const testAdd = ({args, expected}) => () => {
```
boneskull, line 725:
```suggestion
  const testAdd = ({args, expected}) => function() {
```

docs/index.md
```diff
@@ -684,25 +684,23 @@ Given Mocha's use of `Function.prototype.call` and function expressions to defin
 Take the following example:
 
 ```js
-var assert = require('chai').assert;
+const assert = require('chai').assert;
 
 function add() {
```
boneskull, line 689:
```suggestion
function add(...args) {
```

docs/index.md
```diff
@@ -684,25 +684,23 @@ Given Mocha's use of `Function.prototype.call` and function expressions to defin
...
-  return Array.prototype.slice.call(arguments).reduce(function(prev, curr) {
-    return prev + curr;
-  }, 0);
+  return Array.prototype.slice.call(arguments).reduce((prev, curr) => prev + curr, 0);
```
boneskull, line 690:
```suggestion
  return args.reduce((prev, curr) => prev + curr, 0);
```

boneskull (changes requested):
thanks for the updates.  I have some more suggestions

boneskull:
Test failure looks like timing-related flake; unrelated

docs/index.md
```diff
@@ -684,25 +684,23 @@ Given Mocha's use of `Function.prototype.call` and function expressions to defin
...
-  return Array.prototype.slice.call(arguments).reduce(function(prev, curr) {
-    return prev + curr;
-  }, 0);
+function add(...args) {
+  return args.reduce((prev, curr) => prev + curr, 0);
 }
 
-describe('add()', function() {
-  var tests = [
+describe('add()', () => {
```
boneskull, line 693:
```suggestion
describe('add()', function() {
```

boneskull (changes requested):
missed one

juergba (approved):
@maxwellgerber thank you
The cancelled browser test is off-topic.

PR #4496, opened 2020-11-02
properly dispose Mocha instance in watch mode; closes #4495

### LINKED ISSUE #4495: mocha leaks event listeners on `process`
### Prerequisites

### Description

`Runner` does not clean up event listeners on global objects.

### Steps to Reproduce

https://github.com/eps1lon/mocha-event-leak

Removes all event listeners when destroyed.

Event listeners on `process` leak:
```bash
ℹ [mocha] waiting for changes...
(node:18921) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(node:18921) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unhandledRejection listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

`Runner.prototype.dispose` is never called in watchmode.

Every time after the ~10th run

### Versions

- The output of `node --version`: 14.4.0
- Your operating system
  - name and version: Ubuntu 18.04
  - architecture (32 or 64-bit): 64  bit
- Your shell (e.g., bash, zsh, PowerShell, cmd): fish

### Additional Information

The event handlers are attempted to be removed in `prepare`:https://github.com/mochajs/mocha/blob/7e490aa13e8ed25c78e77faa6fed0337e948b619/lib/runner.js#L1063-L1064
That doesn't have any effect after instantiation since `this.uncaught` is different between each runner instance. Whatever constructs the `Runner` instance also needs to call `dispose`. Might be better if the `Runner` took care of it e.g. with `this.once('end', () => this.dispose())` in the `constructor`.

boneskull:
- Mocha instance needed a `dispose()` call before re-instantiation
- Exposed `sleep()` in integration test helpers

Ref: #4495

### DIFF
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -415,8 +415,12 @@ async function runMochaWatchAsync(args, opts, change) {
   if (typeof opts === 'string') {
     opts = {cwd: opts};
   }
-  opts = {sleepMs: 2000, ...opts, fork: process.platform === 'win32'};
-  opts.stdio = ['pipe', 'pipe', 'inherit'];
+  opts = {
+    sleepMs: 2000,
+    stdio: ['pipe', 'pipe', 'inherit'],
+    ...opts,
+    fork: process.platform === 'win32'
+  };
   const [mochaProcess, resultPromise] = invokeMochaAsync(
     [...args, '--watch'],
     opts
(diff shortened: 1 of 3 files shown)


test/integration/helpers.js
```diff
@@ -415,8 +415,12 @@ async function runMochaWatchAsync(args, opts, change) {
   if (typeof opts === 'string') {
     opts = {cwd: opts};
   }
-  opts = {sleepMs: 2000, ...opts, fork: process.platform === 'win32'};
-  opts.stdio = ['pipe', 'pipe', 'inherit'];
+  opts = {
```
boneskull, line 418: this was preventing watch mode tests from examining STDERR

test/integration/helpers.js
```diff
@@ -539,24 +543,25 @@ function sleep(time) {
 module.exports = {
   DEFAULT_FIXTURE,
   SPLIT_DOT_REPORTER_REGEXP,
+  copyFixture,
```
boneskull, line 546: sorted this

test/integration/options/watch.spec.js
```diff
@@ -343,5 +345,31 @@ describe('--watch', function() {
...
+      this.timeout(20000);
+      const testFile = path.join(tempDir, 'test.js');
+      copyFixture(DEFAULT_FIXTURE, testFile);
+
+      return expect(
+        runMochaWatchAsync(
+          [testFile],
+          {cwd: tempDir, stdio: 'pipe'},
+          async () => {
+            const iterations = new Array(15);
```
eps1lon, line 359: Might be faster and more robust to set this to `emitter.getMaxListeners() + 1`. Maybe it works by setting the max listener count to a low number e.g. 1 or 2 which should reduce test runtime.

boneskull: would rather just use the default settings, so yeah, `emitter.getMaxListeners()` sounds better

PR #4502, opened 2020-11-10
Change CLI file parsing errors to use an error code

evaline-ju:
## Description of the Change

After the implementations of https://github.com/mochajs/mocha/pull/3467 and https://github.com/mochajs/mocha/pull/3666 and https://github.com/mochajs/mocha/pull/4464, there are a few instances in the code that do not use error codes yet.

### Alternate Designs

Open to other names for this error code

Using codes is a best practice like in Node.js, codes help identify errors as mocha errors at a glance

### Benefits

Using codes is a best practice like in Node.js, codes help identify errors as mocha errors at a glance

### Possible Drawbacks

None I'm aware of

### Applicable issues

Contributes to https://github.com/mochajs/mocha/issues/3656, semver-minor

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -18,6 +18,7 @@ const {loadConfig, findConfig} = require('./config');
 const findUp = require('find-up');
 const debug = require('debug')('mocha:cli:options');
 const {isNodeFlag} = require('./node-flags');
+const {createUnparsableFileError} = require('../errors');
 
 /**
  * The `yargs-parser` namespace
@@ -190,7 +191,7 @@ const loadPkgRc = (args = {}) => {
       }
     } catch (err) {
       if (args.package) {
-        throw new Error(`Unable to read/parse ${filepath}: ${err}`);
+        throw createUnparsableFileError(err, filepath);
       }
       debug('failed to read default package.json at %s; ignoring', filepath);
     }
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -191,7 +191,10 @@ const loadPkgRc = (args = {}) => {
       }
     } catch (err) {
       if (args.package) {
-        throw createUnparsableFileError(err, filepath);
+        throw createUnparsableFileError(
+          `Unable to read/parse ${filepath}: ${err}`,
+          filepath
+        );
       }
       debug('failed to read default package.json at %s; ignoring', filepath);
     }
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -163,7 +163,12 @@ var constants = {
    * @constant
    * @default
    */
-  TIMEOUT: 'ERR_MOCHA_TIMEOUT'
+  TIMEOUT: 'ERR_MOCHA_TIMEOUT',
+
+  /**
+   * Input file is not able to be parsed
+   */
+  UNPARSABLE_FILE: 'ERR_MOCHA_UNPARSABLE_FILE'
 };
 
 /**
@@ -496,6 +501,21 @@ function createTimeoutError(msg, timeout, file) {
 }
 
 /**
+* Creates an error object to be thrown when file is unparsable
+*
+* @public
+* @param {string} message - Error message to be displayed.
+* @param {string} filename - File name
+* @param {string} [reason] - Why the filename is unparsable.
+*/
+function createUnparsableFileError(message, filename, reason) {
+   var err = new Error(message);
+   err.code = constants.UNPARSABLE_FILE;
+   err.reason = typeof reason !== 'undefined' ? reason : `${filename} is unparsable`;
+   return err;
+ }
+
+/**
  * Returns `true` if an error came out of Mocha.
  * _Can suffer from false negatives, but not false positives._
  * @static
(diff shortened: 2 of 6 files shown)


lib/errors.js
```diff
@@ -421,6 +426,22 @@ function createInvalidPluginImplementationError(
...
 
+/**
+ * Creates an error object to be thrown when file is unparsable
+ *
+ * @public
+ * @param {string} message - Error message to be displayed.
+ * @param {string} filename - File name
+ * @param {string} [reason] - Why the filename is unparsable.
+ */
+function createUnparsableFileError(message, filename, reason) {
```
boneskull, line 437: I think we should remove `reason`, given my comments above in `lib/cli/options.js`

lib/cli/options.js
```diff
@@ -189,7 +190,7 @@ const loadPkgRc = (args = {}) => {
       }
     } catch (err) {
       if (args.package) {
-        throw new Error(`Unable to read/parse ${filepath}: ${err}`);
+        throw createUnparsableFileError(err, filepath);
```
boneskull, line 193: so, `createUnparsableFileError` wants a `string`, but here we're giving it an `Error`.  it will be coerced, of course, but we should probably avoid that

I think what we should do here is change it to something like:

```suggestion
        throw createUnparsableFileError(`Unable to read/parse ${filepath}: ${err}`, filepath);
```

my reasoning is:

1. Exceptions thrown from within parsers can be helpful, surprising, useless, etc.
1. Because of this, we should provide our own message, to provide the user some context
1. Additional information from the original error should also be provided, in case it's useful
1. Using the `reason` param for the "why" is likely going to be hidden from a user, since coercing an `Error` into a `string` shows only the `message` and formatted `stack` props.  so we should stuff everything we know about into the `message` prop.  (I think there is some sort of TC39 proposal for a nested error specification, which we should look into in the future since it applies to how we handle many errors)

test/node-unit/cli/config.spec.js
```diff
@@ -109,11 +109,15 @@ describe('cli/config', function() {
...
-        sinon.stub(parsers, 'yaml').throws();
+        sinon.stub(parsers, 'yaml').throws('goo.yaml is unparsable');
       });
 
       it('should throw', function() {
-        expect(() => loadConfig('goo.yaml'), 'to throw', /failed to parse/);
+        expect(
+          () => loadConfig('goo.yaml'),
+          'to throw',
+          'goo.yaml is unparsable'
```
boneskull, line 119: I'm kind of unsure what exactly we should be testing here.

the original code's assertion says that whenever `parsers.yaml()` throws, the error thrown by `loadConfig()` should contain the text `/failed to parse/`.

with this change, this assertion says that the `message` prop of the `Error` thrown by `parsers.yaml()` should wind up in an `Error` thrown by `loadConfig()`.  this reflects that the message is somehow re-used--which the original code's assertion _should_ have done.  so this is an improvement.

but...

- should we assert it is _not_ the _same_ `Error` as thrown by `parsers.yaml()`? 
- should we assert `loadConfig()` passes the original error message thrown by `parsers.yaml()` into  `createUnparsableFileError()`?
- should we assert the thrown error has the appropriate `code` prop?
- should we stub `createUnparsableFileError()` to return an `Error` with our _own_ `code` prop?

I don't really know the answers here yet, but I'm happy to discuss them if you have the time.  I don't think it's worth blocking the PR for, either way

boneskull (commented):
thanks Evaline!  I have some comments/suggestions... please let me know what you think.

boneskull:
There's a conflict in `lib/errors.js` because I added a `TIMEOUT` error... should be able to just use both sets of changes

boneskull:
in case anyone else was wondering, I think "unparsable" is correct and "unparseable" is not.  my spell checker dislikes both 😄

lib/cli/options.js
evaline-ju: Makes sense. I have updated based on this.

evaline-ju:
I'm not really sure if there is/was a temporary problem with the browser tests since I saw at least another PR with the same failure. At the same time, I don't know how to just re-run the test. 🤔

test/node-unit/cli/config.spec.js
evaline-ju: When I first encountered this test as a failure when I changed this code, it was a bit puzzling to just see `throws()` and `failed to parse` expected in the failure since just the empty `throws()` itself did not seem to indicate any information about what to expect. I think now with the additional changes from your suggestions to add the "unable to parse....", it seems to be clearer where this error came from.

boneskull:
hmm.  I'll look into that error.  some saucelabs weirdness.. hope it's just temporary.

this seems to have a conflict now--could you please rebase onto `master`?

evaline-ju:
hmm I'm not seeing a conflict currently, after I rebased 2 days ago?

Screen Shot 2020-11-20 at 4 40 52 PM

evaline-ju:
After a long hiatus I have tried to formally rebase - looks like the browser test on forked PRs problem is known

outsideris:
You can ignore the browser test. We are working on #4553 .

juergba (approved):
@evaline-ju thank you for this PR.

PR #4503, opened 2020-11-10
Add GH Actions workflow status badge

outsideris:
### Description of the Change

I just found we only show travis build status badge on our README.
But we already moved most our build and test to GH actions except browser tests and we will move out all of them from travis-ci.
So, travis-ci build status badge doesn't show our build status.

And I reformatted HTML for badges so that we can read HTML and diff easily.

### Benefits
We can show the current master's build status to users.

ref: https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge

스크린샷 2020-11-10 오후 7 55 35

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 <p align="center">☕️ Simple, flexible, fun JavaScript test framework for Node.js & The Browser ☕️</p>
 
 <p align="center">
+<a href="https://github.com/mochajs/mocha/actions?query=workflow%3ATests+branch%3Amaster"><img src="https://github.com/mochajs/mocha/workflows/Tests/badge.svg?branch=master" alt="GitHub Actions Build Status"></a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,10 +4,20 @@
...
+<p align="center">
+<a href="https://github.com/mochajs/mocha/actions?query=workflow%3ATests+branch%3Amaster"><img src="https://github.com/mochajs/mocha/workflows/Tests/badge.svg?branch=master" alt="GitHub Actions Build Status"></a>
+<a href="http://travis-ci.org/mochajs/mocha"><img src="https://api.travis-ci.org/mochajs/mocha.svg?branch=master" alt="Build Status"></a>
```
boneskull, line 9: We're no longer using travis

```suggestion
```

boneskull (changes requested):

README.md
outsideris: I missed it.

PR #4504, opened 2020-11-10
wip: run browser tests on gha

boneskull:
trying to get off of travis...


boneskull:
OK, this seems to be working, but is not ideal:

@christian-bromann I based this on your [example repo](https://github.com/christian-bromann/saucelabs-karma-example).

I think what I was running into was due to how we create the tunnel.  The thing is, for each build, we run karma _multiple times_; in particular, we run Mocha with different interfaces (`tdd`, `bdd`, `qunit`, etc.) and thus Karma configurations. Mocha's architecture is not such that you can change the interface on-the-fly.  So, we can't reuse a single Mocha instance (though we _can_ run most of our tests against the default interface).

Opening multiple tunnels all the time -- we often run builds in parallel, multiplying the number of tunnels -- eventually caused flake due to running up against SauceLabs' default set of concurrency restrictions.  To mitigate this, we changed the Travis config to make _Travis itself_ open the Sauce connect tunnel (there's an "add-on" for this), and we reuse that tunnel each time we run Karma (with `startConnect: false`).  SauceLabs also increased our concurrency limited after I had asked.  

So, opening a single tunnel and reusing it was what I was trying to accomplish (but couldn't).

Now, in _this_ configuration, we're back at square one: multiple tunnels per build.  It remains to be seen if this will again cause us to hit our account limits...

boneskull:
Looking to merge this asap to avoid running into Travis' build-minute wall.  I think I already disabled builds on it.

christian-bromann:
> It remains to be seen if this will again cause us to hit our account limits...

Please let me know if this happens so I can bump your account up.

Munter (approved):
Looks good to me. Pretty awesome you made it work

PR #4510, opened 2020-11-13
fix some rewiremock usages

boneskull:
using the `() => require(...)` syntax is better for refactoring/intellisense/etc


outsideris (approved):

JoshuaKGoldberg:
This is very old and has some merge conflicts. Closing out as we'll separately investigate better development-time types support.

PR #4523, opened 2020-11-26
chore(deps): upgrade to latest stable

AviVahl:
### Description of the Change

Upgrade dependencies to latest stable versions and regenerate lock file.

### Benefits

Getting the latest bug fixes and features. Better de-duping for package consumers.

### Possible Drawbacks

Regressions are always a possibility. That's why we test, right?

### Applicable issues

N/A


AviVahl:
got two node 10 windows failures. I suspect `workerpool@6.0.3` might be related, so I downgraded it back to `6.0.2`.

AviVahl:
not sure if the CI failure related to the upgrades. will re-open if I figure it out.

PR #4527, opened 2020-12-02
fix: remove ansi escape sequences from escaped xunit output

### LINKED ISSUE #4526: xunit reporter does not strip ansi escape sequences, leading to invalid XML
### Prerequisites

### Description

When testing code that is throwing exceptions that have strings formatted with `chalk` or using ANSI escape codes in some other way, the resulting XML will contain the entity reference `&x1B;` which is an invalid character in XML.

### Steps to Reproduce

Prepare minimal test file

```javascript
const chalk = require('chalk');

it('Test contains ANSI escape sequences', () => {

});
```

Run tests using mocha and xunit reporter

```bash
$  /code/.npm-global/bin/mocha -R xunit -O output=xunit.xml
Exception: /code/.npm-global/bin/mocha exited with 1
[tty 12], line 1: /code/.npm-global/bin/mocha -R xunit -O output=xunit.xml
```

Test fails as expected and creates `xunit.xml`

```xml
<testsuite name="Mocha Tests" tests="1" failures="0" errors="1" skipped="0" timestamp="Wed, 02 Dec 2020 17:59:42 GMT" time="0.003">
<testcase classname="" name="Test contains ANSI escape sequences" time="0"><failure>&#x1B;[31mthis is not ok&#x1B;[39m

      + expected - actual

      -false
      +true

AssertionError [ERR_ASSERTION]: &#x1B;[31mthis is not ok&#x1B;[39m
    at Context.it (test.js:5:9)
    at callFn (/code/mocha/lib/runnable.js:366:21)
    at Test.Runnable.run (/code/mocha/lib/runnable.js:354:5)
    at Runner.runTest (/code/mocha/lib/runner.js:677:10)
    at /code/mocha/lib/runner.js:801:12
    at next (/code/mocha/lib/runner.js:594:14)
    at /code/mocha/lib/runner.js:604:7
    at next (/code/mocha/lib/runner.js:486:14)
    at Immediate._onImmediate (/code/mocha/lib/runner.js:572:5)</failure></testcase>
</testsuite>
```

Validate XML

```bash
$ xmllint xunit.xml
xunit.xml:2: parser error : xmlParseCharRef: invalid xmlChar value 27
classname="" name="Test contains ANSI escape sequences" time="0"><failure>&#x1B;
                                                                               ^
xunit.xml:2: parser error : xmlParseCharRef: invalid xmlChar value 27
contains ANSI escape sequences" time="0"><failure>&#x1B;[31mthis is not ok&#x1B;
                                                                               ^
xunit.xml:9: parser error : xmlParseCharRef: invalid xmlChar value 27
AssertionError [ERR_ASSERTION]: &#x1B;[31mthis is not ok&#x1B;[39m
                                      ^
xunit.xml:9: parser error : xmlParseCharRef: invalid xmlChar value 27
AssertionError [ERR_ASSERTION]: &#x1B;[31mthis is not ok&#x1B;[39m
                                                              ^
Exception: xmllint exited with 1
[tty 13], line 1: xmllint xunit.xml
```

**Expected behavior:** `xunit.xml` contains valid XML

**Actual behavior:** the XML contains the invalid character &#x1B;

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 8.2.1
- The output of `node --version`: v10.22.0
- Your operating system
  - name and version: Linux b973b6e6ac1a 4.19.76-linuxkit #1 SMP Tue May 26 11:42:35 UTC 2020 x86_64 Linux
- Your shell (e.g., bash, zsh, PowerShell, cmd): elvish

trieloff:
### Requirements

### Description of the Change

The `utils.escape` function is amended to strip the escape sequence using a regular expression.

### Alternate Designs

* patch the underlying `he` library: dismissed because the direct fix is likely faster to become usable
* filter the output before running `he`: dismissed because the correct escaping in JavaScript is harder to find
* filter out all invalid XML characters: dismissed because this is not a problem observed with other characters

The `xunit` reporter is in core and should not be buggy

### Benefits

`xunit` reporter output can be consumed by tools that expect valid XML

### Possible Drawbacks

Tools that process the XML as plain (unicode) text (unlikely) can no longer see the ANSI escape sequence.

### Applicable issues

fixes #4526 

This is a bug fix (patch release)


lib/utils.js
```diff
@@ -36,7 +36,9 @@ exports.inherits = util.inherits;
  * @return {string}
  */
 exports.escape = function(html) {
-  return he.encode(String(html), {useNamedReferences: false});
+  return he
+    .encode(String(html), {useNamedReferences: false})
+    .replace(/&#x1B;/g, '');
```
outsideris, line 41: I think this is a too narrow solution. It can handle only the `#x1B;` case.
We can't replace special characters whenever we find invalid characters.

outsideris (commented):
Isn't there a way to escape or remove all invalid characters for XML?

lib/utils.js
trieloff: From https://github.com/mochajs/mocha/issues/4526#issuecomment-737443533

> `Char ::= #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]`

We could throw every hex char reference into a replacer function that checks against the spec's permitted range and strips invalid chars

```suggestion
    .replace(/&#x([0-9A-F]);/g, (match) => {
      const val = Number.parseInt(match); 
      return val === 0x9 || val === 0xA || val === 0xD || (val >= 0xE000 && val <= 0xFFFD) || (val >= 0x10000 && val <= 0x10FFFF) ? `&#x${match};` : '';
    });
```

or replace with a double escaped sequence:


```suggestion
    .replace(/&#x([0-9A-F]);/g, (match) => {
      const val = Number.parseInt(match); 
      return val === 0x9 || val === 0xA || val === 0xD || (val >= 0xE000 && val <= 0xFFFD) || (val >= 0x10000 && val <= 0x10FFFF) ? `&#x${match};` : `&amp;#x${match};`;
    });
```

trieloff: @outsideris do you have a preference?

outsideris: I prefer the first one. But I want to hear from @mochajs/core member's opinions because I don't have a specialty with xunit.

trieloff: @outsideris have you heard anything from the rest of the team?

juergba: `utils.escape` is not only used by the `xunit` reporter, at least `HTML` reporter and maybe others would be affected. What are the implications?

> I think this is a too narrow solution [...]

I agree, we can't replace each special character step by step, this would be a never ending story.
If the `he` package can't help, there must be other packages available. I don't like the idea that Mocha has to handle special characters. What about eg. the `entities` package? Any others?

juergba:
There is a related `xunit` issue #3586.

lib/utils.js
trieloff: https://github.com/mochajs/mocha/pull/4527#discussion_r551176184 has two proposals that handle the entire range of illegal characters, so it is not replacing "each special character step by step".

`entities` might provide a fix, but I wanted to avoid adding another dependency for something that is just a change of four lines of code.

https://github.com/michaelleeallen/mocha-junit-reporter as the most prominent external reporter uses `xml`, i.e. a proper XML parser, instead of just treating XML as a weird type of HTML.

trieloff:
This issue has been hanging around for 8 months now, and I'm unsure what I can do to get it moving again. There is an open thread that has multiple possible solutions to the problem, but no expression of preference from the Mocha team.

JoshuaKGoldberg:
👋 coming back to this @trieloff, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are again maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

trieloff:
Hi @JoshuaKGoldberg – that's unexpected – but not unwelcome. I'm happy to rebase the PR to the latest `main`. Do you want to reopen this PR or have me create a new one?

JoshuaKGoldberg:
Great! Go ahead and reopen this one.

We also discussed that we don't like the requirement of rebasing, so if you want to do a more casual merge that works too. Whatever's easiest on your end. 🙂

trieloff:
Ok, let me get that CLA signed, then.

lib/utils.js
JoshuaKGoldberg: Coming back to this ~3 years later, I think it's fine to have a small hardcoded list in the source code. The list of characters doesn't change often and I'd rather avoid adding in another dependency.

I'm also in favor of stripping out unsafe characters (the first one, https://github.com/mochajs/mocha/pull/4527#discussion_r560348163). Double-escaping them makes the result somewhat unreadable IME.

Sorry to repeat the cycle of asking for more input, but cc @mochajs/maintenance-crew - thoughts?

If we don't get you an answer within, say, 3 weeks from now I say let's just go with your proposal of stripping out the unsafe characters. 😄

trieloff:
- CLA has been signed (thanks, corporate overlords)
- `master` has been merged
- conflicts have been resolved

lib/utils.js
trieloff: > If we don't get you an answer within, say, 3 weeks from now I say let's just go with your proposal of stripping out the unsafe characters.

I'll take you up on this offer, @JoshuaKGoldberg and commit the fix that checks against the whole character range, but keeps stripping invalid chars.

trieloff:
@JoshuaKGoldberg can I get some eyes on this, please?

JoshuaKGoldberg:
Yes! Thanks for the prod - this slipped off my radar. Reviewing now!

test/integration/reporters.spec.js
```diff
@@ -62,6 +62,32 @@ describe('reporters', function () {
         done(err);
       });
     });
+
+    it('does not include ansi escape sequences (issue: 4526)', function(done) {
```
JoshuaKGoldberg, line 66: [Naming] Nit: behaviors and tests have a habit of morphing over time...

```suggestion
    it('does not include ansi escape sequences', function(done) {
```

(context: I think there are only two examples of a test name including something like this, from 3 someone who isn't maintaining anymore, 3 years ago - if this ever was a Mocha practice, it wasn't widely adopted)

JoshuaKGoldberg (approved):
Swell! A nice direct change, with a test case - thank you! 🙌 

Will wait for another maintainer to review. In the interim, just leaving one small nitpick, nothing particularly important.

lib/utils.js
```diff
@@ -33,7 +33,12 @@ exports.inherits = util.inherits;
  * @return {string}
  */
 exports.escape = function (html) {
-  return he.encode(String(html), {useNamedReferences: false});
+  return he
+    .encode(String(html), {useNamedReferences: false})
+    .replace(/&#x([0-9A-F]);/g, (match) => {
+      const val = Number.parseInt(match); 
+      return val === 0x9 || val === 0xA || val === 0xD || (val >= 0xE000 && val <= 0xFFFD) || (val >= 0x10000 && val <= 0x10FFFF) ? `&#x${match};` : '';
+    });
```
voxpelli, line 41: Suggestion:

Turns out that `he` already has this in itself, just doesn't expose it other than through `strict: true`, which throws when it encounters it.

Maybe we could use that regexp and try to land a PR that either exposes it or adds a `skipInvalid: true` option or such?

```suggestion
  return  he.encode(String(html).replace(regexInvalidRawCodePoint, ''), {useNamedReferences: false});
```

Requires this to be added as well:

```js
// From https://github.com/mathiasbynens/he/blob/36afe179392226cf1b6ccdb16ebbb7a5a844d93a/he.js#L53
var regexInvalidRawCodePoint = /[\0-\x08\x0B\x0E-\x1F\x7F-\x9F\uFDD0-\uFDEF\uFFFE\uFFFF]|[\uD83F\uD87F\uD8BF\uD8FF\uD93F\uD97F\uD9BF\uD9FF\uDA3F\uDA7F\uDABF\uDAFF\uDB3F\uDB7F\uDBBF\uDBFF][\uDFFE\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF]/g;
```

voxpelli (changes requested):
There is already a test that verifies that invalid XML tags are escaped:

https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/test/unit/utils.spec.js#L657-L665

This should extend those tests rather than introduce a new xunit specific one, as this change is not specific to the xunit parts of the code.

Also added a suggested code change that actually makes use of a regexp from `he`

PR #4549, opened 2021-01-11
Ensure require.resolve is only ever called with absolute filepaths

### LINKED ISSUE #4548: unloadFiles throws MODULE_NOT_FOUND in ESM context for files added with relative path
### Prerequisites

### Description
Programmatically using `mocha.addFile(file)` and `mocha.unloadFiles()` in an ES module context throws `MODULE_NOT_FOUND` when `mocha.addFile(...)` is called with a relative path.

Wrapping the module resolution logic with `path.resolve(...)` resolves this issue by ensuring the path is absolute.

### Steps to Reproduce
https://repl.it/@christiantjl/MochaFileUnloader#fails.js
`node succeeds.js` shows that an absolute path works without issue, and `node fails.js` shows that adding with a relative path throws an error.

ctjlewis:
### Requirements

### Description of the Change
`unloadFile` is only ever passed filepaths, but adding a file with `mocha.addFile(relativePath)` throws `MODULE_NOT_FOUND` in an ESM context because `require.resolve(relativeFile)` fails.

Ensuring that `unloadFile` only ever calls `require.resolve` with absolute filepaths by wrapping with `path.resolve(relativePath)` prevents this error.

### Applicable issues
- Closes #4548

### DIFF
--- a/lib/nodejs/file-unloader.js
+++ b/lib/nodejs/file-unloader.js
@@ -5,11 +5,16 @@
  * @private
  * @module
  */
+const path = require('path');
 
 /**
- * Deletes a file from the `require` cache.
+ * Deletes a file from the `require` cache. Resolve the file first to prevent
+ * MODULE_NOT_FOUND for relative paths.
+ *
+ * @see https://github.com/mochajs/mocha/issues/4548
  * @param {string} file - File
  */
 exports.unloadFile = file => {
-  delete require.cache[require.resolve(file)];
+  const absoluteModulePath = require.resolve(path.resolve(file));
+  delete require.cache[absoluteModulePath];
 };


juergba:
@ctjlewis thank you
I did some changes to our CI test workflow yesterday. I don't know why the `pull-request` event hasn't triggered our CI tests on your PR. I need to get this solved first.

ctjlewis:
No worries @juergba, I was looking for CI tests as well. I will pull from upstream when that's settled, just ping me here, no rush.

juergba:
So, I have two problems with the workflow:
I don't know your branch's name from your fork. It's not `unload-files`, nor `ctjlewis:unload-files`, nor `ctjlewis/mocha:unload-files`. I need to distinguish PR's coming from a fork against pushed ones to the original mocha repo.

Browser test: 
The github [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories) says: 
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

I guess thats the reason, why the browser test is failing. The SaucLabs login doesn't work.
Do you have any idea?

ctjlewis:
GitHub embeds it here as `ctjlewis/mocha:unload-files`. 

image

I have no idea about the CI - I've never contributed to Mocha before, I just ran into this error and wanted to submit a patch.

juergba:
The CI tests do run now on your PR, but the browser test will fail most propably.
I will have to investigate further on that one.

You could rebase locally and force-push to your fork, if you want.
Thanks for letting me abuse your PR.

ctjlewis:
No worries @juergba, feel free to use this PR to debug.

ctjlewis:
@juergba Hey, any way I can help out here? This bug does actually break most `unloadFiles()` calls in ES contexts - not trying to be pushy at all, but I'm happy to help move it along any way I can. This was the fastest response to a PR I've ever had (1hr) so I'm not complaining by any means.

juergba:
@ctjlewis If I remember correctly, the ESM loader doesn't use the `require` cache.

see our [API - unloadFiles](https://mochajs.org/api/mocha#unloadFiles)
see [Node docs](https://nodejs.org/docs/latest-v14.x/api/esm.html#esm_no_require_cache)

Even worse, AFAIK there is no way to clear the ESM cache. E.g. that's why our `watch` functionality doesn't work with ESM. There is no way to clear the cache and re-run the tests.
BTW `unloadFile` is a private function.

ctjlewis:
This PR fixes the issue. And yes but `unloadFiles`, which relies
`unloadFile`, is public.

I ran into this issue in real-world codebase, and resolving the import path
works. Mocha is imported as a CJS module so `require` works as expected
inside Mocha’s source, it can be safely ignored.

juergba:
I'm not convinced of this change. I will need a few days to look into it more closely.

Your examples just don't make any sense:
  - you never load any test files, neither the ESM way via `loadFilesAsync`, nor the CommonJS way via `mocha.run` (which calls `loadFiles`). The function `loadFiles` does call `path.resolve`.
  - `unloadFiles` does not clear the ESM cache (which is not the `require.cache`).

ctjlewis:
I guess the issue is not as obvious as I thought - I made a Repl.it here: https://repl.it/@christiantjl/MochaRequireResolve

This basically only affects programs where we import Mocha from an ESM context. ES modules *can* import CJS modules, and that `require.resolve` statement is executed within the emulated CJS module. I think under the hood it relies on `import.meta.resolve` but since Mocha is a CJS module, this is the appropriate fix, which adds support for ES interoperation.

I should have made it clear this is an ESM interop thing, where, for CJS programs, this will have no impact, but it will prevent this error from presenting when Mocha is used programmatically from inside an ES module via `import Mocha from 'mocha'`.

Apologies for not including a repro off the bat.

ctjlewis:
Tfw I *did* include a repro in that issue, and I basically recreated it a second time... I hope I have sufficiently demonstrated the issue lmao.

ctjlewis:
Hate to bug you @juergba, but did you have a chance to review the example?

juergba:
@ctjlewis None of your repros works, there are syntax errors all over when pressing the Run-Button.

Meanwhile I doubt that this issue has its cause specifically with ESM. If you just take following code snippet:
```js
mocha.addFile('./test/test-pending.js');
console.log(require.resolve('./test/test-pending.js', {paths:[process.cwd()]}));     // runs successfully
console.log(require.resolve('./test/test-pending.js'));                              // fails
```
The module resolution algorithm does not include sub-directories of the current working directory. Therefore `require.resolve` fails.

Since `loadFiles` calls `path.resolve`, it could make sense to do the same in `unloadFiles` as well.

ctjlewis:
The run button doesn't work because of the Node version Repl.it uses. Execute `node index.js` to show the behavior of `succeeds.js` and `fails.js`.

I am admittedly getting frustrated because you keep trying to tell me that this error is not real simply because you are struggling to understand it. I have tried to explain this multiple times: This presents when Mocha is imported as a CJS module in an *ESM context*.

Switch over to the shell tab, and run `node index.js` to execute the two tests.

ctjlewis:
CC:
@boneskull
@craigtaub
@Munter
@outsideris

ctjlewis:
I updated the Repl.it with a custom environment and all `.mjs` extensions to get the repro to work with the Run button rather than needing to execute it directly:

https://repl.it/@christiantjl/MochaRequireResolve

JoshuaKGoldberg:
Looks like this was not stale, and the author did provide a good amount of repro steps. Re-opening & marking the issue as accepted.

ctjlewis:
> Looks like this was not stale, and the author did provide a good amount of repro steps. Re-opening & marking the issue as accepted.

Yeah, this was one of the most incompetent responses to a bugfix PR I'd ever seen in my life. It's one thing to handle a bug report like shit, but it's especially sad when contributors are your hand and submit a fix alongside the report, plus various repros.

That guy owes me an apology. Plus everyone who uses Mocha for forcing a bug to stay in the library for 3 years.

lib/nodejs/file-unloader.js
JoshuaKGoldberg, line 1: [Testing] This change seems reasonable at first glance, but we'll want to have unit tests for it. Do you see a way to do that?

JoshuaKGoldberg (changes requested):
Seems reasonable, thanks! But we'll want to unit test this.

JoshuaKGoldberg:
👋 ping @ctjlewis, is this still something you have time for?

JoshuaKGoldberg:

PR #4557, opened 2021-01-20
Add file location when SyntaxError happens in ESM (fixes #4551)

### LINKED ISSUE #4551: If one among multiple ESM tests has a syntax error, then Mocha doesn’t report which one
### Prerequisites

### Description

If I run multiple ESM tests (e.g. via `npm t test*.mjs`), and one of them has a syntax error, then Mocha doesn’t tell me which file it is.

### Steps to Reproduce

Repository with MCVE: https://github.com/rauschma/mocha-test-repo

Execute either of the following two commands:

```
npm t test*.mjs
npm t 'test*.mjs'
```

`test2.mjs` has a syntax error.

**Expected behavior:** I’d like Mocha to tell me which file has the syntax error.

**Actual behavior:** I’m getting the following error message.

```
SyntaxError: Unexpected token '=>'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:117:18)
    at async link (internal/modules/esm/module_job.js:42:21)
```

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 8.2.1 (there is no globally installed Mocha)
- The output of `node --version`: v14.15.4

giltayar:
### Description of the Change

According to issue #4551 (I verified that it is true), when there is a syntax error in an ESM file, the error thrown by Node.js contains only the error, but not the location (file and line number) where this problem happened. 

While the error doesn't include this information, we have the filename and can "inject" it into the error artificially. At least until Node.js fixes the problem. This change is local to ESM as it is only in the ESM importing code.

### Alternate Designs

The only alternative I can see is to get Node.js to fix the problem.

Because plugins can't fix this.

### Benefits

It is currently difficult to figure out on which test file the syntax error happened. This helps a lot by at least giving the file where the syntax error is. Usually, in the IDE, the developer can figure out what the problem is.

### Possible Drawbacks

The code that "patches" the error is convoluted and can theoretically fail in weird ways. Also, we won't know when Node.js fixes the problem, so we may have both indications on location in the error once Node.js fixes it.

### Applicable issues

This is a bug fix (patch release)

### DIFF
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -3,7 +3,29 @@ const url = require('url');
 
 const formattedImport = async file => {
   if (path.isAbsolute(file)) {
-    return import(url.pathToFileURL(file));
+    try {
+      return await import(url.pathToFileURL(file));
+    } catch (err) {
+      // This is a hack created because ESM in Node.js (at least in Node v15.5.1) does not emit
+      // the location of the syntax error in the error thrown.
+      // This is problematic because the user can't see what file has the problem,
+      // so we add the file location to the error.
+      // This `if` should be removed once Node.js fixes the problem.
+      if (
+        err instanceof SyntaxError &&
+        err.message &&
+        err.stack &&
+        !err.stack.includes(file)
+      ) {
+        const newErrorWithFilename = new SyntaxError(err.message);
+        newErrorWithFilename.stack = err.stack.replace(
+          /^SyntaxError/,
+          `SyntaxError[ @${file} ]`
+        );
+        throw newErrorWithFilename;
+      }
+      throw err;
+    }
   }
   return import(file);
 };
--- a/test/integration/esm.spec.js
+++ b/test/integration/esm.spec.js
@@ -45,7 +45,10 @@ describe('esm', function() {
     const err = await runMochaAsync(fixture, args, {stdio: 'pipe'}).catch(
       err => err
     );
-    expect(err.output, 'to contain', 'SyntaxError').and('to contain', fixture);
+    expect(err.output, 'to contain', 'SyntaxError').and(
+      'to contain',
+      'esm-syntax-error.fixture.mjs'
+    );
   });
 
   it('should recognize esm files ending with .js due to package.json type flag', function(done) {
--- a/test/integration/esm.spec.js
+++ b/test/integration/esm.spec.js
@@ -1,5 +1,7 @@
 'use strict';
-var run = require('./helpers').runMochaJSON;
+var helpers = require('./helpers');
+var run = helpers.runMochaJSON;
+var runMochaAsync = helpers.runMochaAsync;
 var utils = require('../../lib/utils');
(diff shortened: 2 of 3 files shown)


giltayar:
I'm guessing the coverage decreased because I didn't add a test. Should I?

giltayar:
I added a test and coverage still decreased, so I'm guessing it's not a real problem?

juergba:
Coverage does run, without test it was -0.1, now with test -0.02. If you look at the coverage details, you see line 27 (throw err) is the evil, newly added and uncovered line. For my taste your test is sufficiant.

The browser test does not work on PR's raised from forked repos.

lib/esm-utils.js
```diff
@@ -3,7 +3,29 @@ const url = require('url');
 
 const formattedImport = async file => {
   if (path.isAbsolute(file)) {
-    return import(url.pathToFileURL(file));
+    try {
+      return await import(url.pathToFileURL(file));
```
juergba, line 7: I don't remember why we have two different calls of `import`. This one and a second one at line 30.
Why do we need this distinction again?
Why is the second `import` outside of `try`?

giltayar: I agree, but I would wait for Mocha to stop supporting Node 10, as Node 10 has the "old" ESM implementation.

outsideris: > Why is the second import outside of try?

I'm also wondering about that.

giltayar: The second import: The other `import` (line 30) is for bare specifiers (e.g. `import 'react'). As I assumed that syntax errors won't happen in packages, I didn't wrap it with try/catch: the idea was to try and minimize the surface of this hack. If you believe we should consolidate the two in the interest of consistency, I will concur.

juergba (approved):

juergba:
I restarted our CI tests, because I have problems with GH actions on another PR.
Yes, now your PR does not pass anymore. Without any changes, before it passed, now the ESLint tests fail.
I hate GH actions, it's just so unreliable and bitchy. Tomorrow is another day, we will see.

juergba:
It's even worse now, the tests are now failing due to some parallel problem. And the ESLint failure persists.
Every time I restart this GH action it's getting worse.

juergba:
@giltayar I fixed the ESLint problem. GH does not delete expired artifacts, which results in failures whenever an app is trying to download one of those. I deleted all expired artifacts and scheduled an action for this purpose as well.

But other tests are failing. Could it be the Node version? With Node 15.7. the tests passed, now with 15.8. they fail.
Could you have a look, please?

juergba:
I will test an update of `workerpool` dependency. On this branch if you don't mind.

giltayar:
@juergba Looks like it did the trick.

juergba:
@giltayar do we merge?

giltayar:
@juergba LGTM

PR #4558, opened 2021-01-20
Show where syntax error from with ESM

### LINKED ISSUE #4551: If one among multiple ESM tests has a syntax error, then Mocha doesn’t report which one

outsideris:
### Description of the Change
This is a kind of hack because node.js currently not support syntax errors in detail. See <https://github.com/nodejs/modules/issues/471>
And This is another approach to #4557 .

I found `node --unhandled-rejections=throw -e "import('syntax-error-file.mjs')"`  show syntax erroed file, linenumber, and code.

So, I try to run it with a file to import only when `SyntaxError` raised.

So, it will show like this:

```sh
$ mocha -ui qunit test*.mjs

file:///Users/outsider/Dropbox/projects/github/mocha/test2.mjs:5
test('test2a', => { // syntax error!
               ^^

SyntaxError: Unexpected token '=>'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:145:18)
    at async link (internal/modules/esm/module_job.js:47:21)
```

And `--unhandled-rejections=throw` is default since [Node.js 15.0.0](https://nodejs.org/en/blog/release/v15.0.0/#throw-on-unhandled-rejections-33021)

### Alternate Designs
See #4557

Before node.js itself support it, Users can find easily where a syntax error from.

### Possible Drawbacks
It can be broken when Node.js itself supports it.

### Applicable issues
close #4551

If we agreed with this approach, I will add tests for It. 
I want to hear your opinion about this workaround.

### DIFF
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -2,10 +2,40 @@ const path = require('path');
 const url = require('url');
 
 const formattedImport = async file => {
-  if (path.isAbsolute(file)) {
-    return import(url.pathToFileURL(file));
+  try {
+    if (path.isAbsolute(file)) {
+      return await import(url.pathToFileURL(file));
+    }
+    return import(file);
+  } catch (err) {
+    if (err.name === 'SyntaxError') {
+      // This is a hack to print where syntax error is from.
+      // In dynamic import, it doesn't contain filename linenumber in errer.stack
+      // See https://github.com/nodejs/modules/issues/471
+      // FIXME: remove this hack after nodejs suppurt
+      return await new Promise((resolve, reject) => {
+        const {spawn} = require('child_process');
+
+        const proc = spawn(process.execPath, [
+          '--unhandled-rejections=throw',
+          '-e',
+          `import('${file}')`
+        ]);
+
+        let errMsg = '';
+
+        proc.stderr.on('data', data => {
+          errMsg += data;
+        });
+
+        proc.on('exit', () => {
+          const err = new Error(errMsg);
+          err.stack = errMsg;
+          reject(err);
+        });
+      });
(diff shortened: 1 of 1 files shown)


juergba:
@outsideris IMO this solution goes too far. It's certainly a creative work-around, but we should not try to fix Node bugs in such a complex way as spawning a child-process.

I prefer the simpler approach of @giltayar which just injects the file name into the original error stack.

Btw I noticed that in Node v15.3.0 the stability of Node's EMS implementation has changed to _Stable_. Nice.

outsideris:
I also know that. I don't have a strong opinion to use this approach. 
I just made a proposal after digging to find a way where the syntax error is from.

outsideris:
Closed by #4557

PR #4563, opened 2021-01-30
Introduce the afterThis hook

Zirak:
### Requirements

### Description of the Change

The afterThis hook allows one to write cleanup logic inside the test definition,
enabling the use of variables declared inside the test itself. For example:

```js
// without afterThis hook
let resource;
beforeEach(() => {
    resource = acquireResource();
});
afterEach(() => {
    resource.dispose();
});
it('does something with a resource', () => {
    // use `resource`
});

// using afterThis hook
it('does something with a resource', () => {
    const resource = acquireResource();
    afterThis(() => resource.dispose());

    // use `resource`
});
```

The code using the resource is much closer to the initialisation/cleanup logic,
and there are less lifecycle hooks to be aware of.

This is inspired by bash's `trap` and go's `defer`.

### Alternate Designs

This design was chosen out of frustration with resource cleanup in
tests. Imagine a test in a larger suite which needs to establish a sinon stub,
like so:

```js
it('does something', () => {
    const stub = sinon.stub(fs, 'readFile').withArgs(...).returns(...);

    // something which calls fs
    sinon.assert.called(stub);

    stub.restore();
});
```

The call to `stub.restore` may never be reached, since the test may fail
beforehand. There are a few alternatives:

1) Use specific `beforeEach` and `afterEach` hooks. In case this stub is shared
between multiple tests, defining the stub in a pair of `before/afterEach` hooks
will be the way to go. However, with a test-specific stub which should /not/ be
shared between tests, you may need to set up a suite just for this test, as
such:

```js
describe('a thing', () => {
    it('does something unrelated to the stub', () => {
        // ...
    });

    describe('uses the stub', () => {
        let stub;
        beforeEach(() => {
            stub = sinon.stub(fs, 'readFile')...;
        });
        afterEach(() => {
            stub.restore();
        });

        it('does something with the stub', () => {
            // something which calls fs
            sinon.assert.called(stub);
        });
    });
});
```

This is possible, and may be the standard in some cases. However, there is a lot
of boilerplate around a single variable. When the friction of doing something is
high, it may end up not being done.

2) Use more global `beforeEach` and `afterEach` hooks. For sinon, this is
possible with `sinon.restore()`, like so:

```js
let stub;
beforeEach(() => {
    stub = sinon.stub(fs, 'readFile')...;
});
afterEach(() => {
    sinon.restore();
});
it('does something with the stub', () => {
    // something which calls fs
    sinon.assert.called(stub);
});
```

This may not be as easy for other kinds of resources, like setting up in-memory
DBs, other mocking libraries, or other mutative or otherwise stateful calls. For
more specific examples, see the "Benefits" section below.

3) A `try { ... } finally { ... }` statement, as such:

```js
it('does something with a stub', () => {
    let stub = sinon.stub(fs, 'readFile')...;

    try {
        // something which uses fs
        sinon.assert.called(stub);
    } finally {
        stub.restore();
    }
});
```

This is the closest in spirit to this commit's implementation, more akin to C#'s
`using` or Python's `with`.

This does incur confusion between test flow and program flow. On reading the
code above, the purpose of the `try...finally` may not be so clear on first
read; is this meant to be part of the test's flow, catching some kinds of
exceptions and trying out different situations, or is it test bookkeeping?

By calling a test bookkeeping function, you advertise your intention more
clearly, allowing the actual test code to dominate the bulk of the test
function.

This commit touches some of mocha's inner knobs, in a way that I don't believe
is (yet?) possible with 3rd-party packages.

An avenue I am very willing to explore is how to open up mocha so this will not
have to be in core.

### Benefits

A few more examples of resource cleanup:

```js
it('does something with a db connection', async () => {
    const db = await createMockDB();
    afterThis(db.close);

    // ...
});

it('does something with multiple resources', () => {
    const wsServer = new WebSocket.Server({
        host: 'localhost',
        port: 8888,
    });
    afterThis((done) => wsServer.close(done));

    const conn = new WebSocket('http://localhost:8888');
    afterThis((done) => {
        conn.on('close', done);
        conn.close();
    });
});

it('uses nock to persist a specific path', () => {
    const ignoreThisPath = nock('http://localhost:8888')
        .persist()
        .get('/something')
        .reply(200, 'You got something');
    afterThis(() => ignoreThisPath.persist(false));
});
```

### Possible Drawbacks

This introduces another hook, which is I believe something that hasn't been done
in a long while. This may be an undesirable precedent, or something which will
require further thought and/or collaboration with other test libraries.

The commit initially presented here may not be suitable for consumption, as I do
believe there are some flows that I've missed. It would be desirable to see if
there's interest in the feature and/or the current design before I continue work
on it.

The lack of prior art on this matter does make me wonder if I'm missing a
fundamental issue at play here.

### Applicable issues

This is at least a minor release, and may be a major release since it introduces
a new global variable to existing code bases.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -49,6 +49,7 @@ overrides:
     env:
       mocha: true
     globals:
+      afterThis: readonly
       expect: readonly
   - files:
       - bin/*
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -31,6 +31,7 @@ module.exports = function bddInterface(suite) {
     context.after = common.after;
     context.beforeEach = common.beforeEach;
     context.afterEach = common.afterEach;
+    context.afterThis = common.afterThis;
     context.run = mocha.options.delay && common.runWithSuite(suite);
     /**
      * Describe a "suite" with the given `title`
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -70,6 +70,10 @@ module.exports = function(suites, context, mocha) {
       suites[0].afterAll(name, fn);
     },
 
+    afterThis: function(name, fn) {
+      mocha.runner.suite.afterThis(name, fn);
+    },
+
     /**
      * Execute before each test case.
      *
--- a/lib/interfaces/qunit.js
+++ b/lib/interfaces/qunit.js
@@ -39,6 +39,7 @@ module.exports = function qUnitInterface(suite) {
     context.after = common.after;
     context.beforeEach = common.beforeEach;
     context.afterEach = common.afterEach;
+    context.afterThis = common.afterThis;
     context.run = mocha.options.delay && common.runWithSuite(suite);
     /**
      * Describe a "suite" with the given `title`.
--- a/lib/interfaces/tdd.js
+++ b/lib/interfaces/tdd.js
@@ -37,6 +37,7 @@ module.exports = function(suite) {
 
     context.setup = common.beforeEach;
     context.teardown = common.afterEach;
+    // XXX how should afterThis be named?
     context.suiteSetup = common.before;
     context.suiteTeardown = common.after;
     context.run = mocha.options.delay && common.runWithSuite(suite);
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -972,10 +972,10 @@ Mocha.prototype.run = function(fn) {
   var suite = this.suite;
   var options = this.options;
   options.files = this.files;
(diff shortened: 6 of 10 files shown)


AviVahl:
Whether this will be accepted or not, I suggest ensuring cleanup is done in reversed order (if I call afterThis several times within an it). We found it super important in my team's various test-kits.

AviVahl:
Similar art: https://github.com/avajs/ava/blob/master/docs/02-execution-context.md#tteardownfn

PR #4569, opened 2021-02-07
match supporter's properties with supporter.js

outsideris:
Currently, [mochajs.org](https://mochajs.org/#sponsors) doesn't show any sponsors or backers. 

I'm not sure when this problem occurred. 
[This commit](https://github.com/mochajs/mocha/commit/cf736fedfb15203824c0d21af7900112354f5804#diff-e4c53b7172f4257db299fa9e7502d3ced0348c61013f8816258218a223649186R39-R40) introduced `SPONSOR_TIER` and `BACKER_TIER` constants and then [this commit](https://github.com/mochajs/mocha/commit/df8e9e6f0108cd06925ea745bb26771ed458c7cb#diff-e4c53b7172f4257db299fa9e7502d3ced0348c61013f8816258218a223649186R229-R230) use these constants for `supporters`'s child properties name.

So, we used `supporters.sponsors` and `supporters.backers` previously but, now it changed to `supporters.sponsor` and `supporters.backer`. However we didn't change `supporters.md`.

### DIFF
--- a/docs/_includes/supporters.md
+++ b/docs/_includes/supporters.md
@@ -3,7 +3,7 @@
 Use Mocha at Work? Ask your manager or marketing team if they'd help [support](https://opencollective.com/mochajs#support) our project. Your company's logo will also be displayed on [npmjs.com](http://npmjs.com/package/mocha) and our [GitHub repository](https://github.com/mochajs/mocha#sponsors).
 
 <ul class="image-list" id="sponsors">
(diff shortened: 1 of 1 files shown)


juergba (approved):
@outsideris thanks for this speedy fix!

outsideris:
Preview build was fine, but production build was failed on Netlify. I will check it.

PR #4574, opened 2021-02-16
fix(esm): allow `import` from mocha in parallel

### LINKED ISSUE #4559: Parallel mode fails silently sometimes with tests that import mocha
### Prerequisites

### Description

IF you're using `--parallel`
AND the number of test files > number of workers (i.e. a worker is reused to run two or more test files)
AND you're using `import {it} from "mocha"` to get your `it` (note that this is not a problem in CJS)
THEN the subsequent test files that run in the reused worker will **silently** not run.

For example if the number of worker is 1, and you're running two test files, the first test file will have its tests executed, but the second test file will load, but the tests won't run.

### Steps to Reproduce

Write two test files that use `import {it} from "mocha"` and use that `it` to run tests. 

Now run them using `mocha file1.mjs file2.mjs` using `--parallel` in a machine with just one CPU. Note that `--job 1` does _not_ reproduce it. Only a machine with 1 CPU. And this is a problem mostly in CI, because dev machines usually have more than 1 CPU.

**Expected behavior:** All tests should have been executed.

**Actual behavior:** Only the tests in the first file are executed. The tests of the second file will be ignored.

**Reproduces how often:** consistent. Always.

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: `8.2.1`
- The output of `node --version`: `15.6.0`
- Your operating system
  - name and version: macOS Big Sure
- Any third-party Mocha-related modules (and their versions): nope
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): nope

### Additional Information

I dug into the code. The problem is this:

* Mocha overrides the `global.it` so that when it is called, tests are added to the current suite (in `bdd.js` in the `bddInterface` function. This also happens in the other interfacs). This is done in the `PRE_REQUIRE` phase of the tests, before the first test file is loaded.
* The first test file is loaded and when `const {it} = require('mocha')` is called, it receives this `global.it`.
* The suite gets the test functions and all is well.
* Now the second test file arrives _at the same worker_, which creates a `new Mocha` and thus a new `suite`.
* Again, in the pre-require, the `global.it` is overriden, so that when called, it will add tests **to the new suite**.
* Again, the test file is loaded, but this time, when `import {it} from 'mocha'` is called, the MJS cache happens and the test gets the original `it`, which adds the tests **to the original suite** and not the new one.
* Now when we get to executing tests, there are no tests to execute, and we're done.

The problem is that the `Mocha` class is treated as a class that can be instantiated multiple times, when it is in fact a singleton due to its use of `global.it`.

My guess is that this does not happen in CJS because in CJS you clear the CJS module cache before loading the files. Unfortunately, ESM does not have a way to clear the cache.

Solutions:
1. Clear the cache. This will work in CJS, but not in ESM that has no API (documented or otherwise) to clear the cache.
2. Make workers non-reusable.
3. Make `it` use a "current" suite that is global and set each time a new `Mocha` is instantiated.

I prefer solution 2, because this problem may just be the tip of the iceberg in terms of problems that happen because Mocha instantiated multiple times. 

One way to implement solution 2 is to use `workerpool` with `numworkers=1` and implement the parallelism by throttling calls to `workerpool.exec` by using a throttling library like `throat`.

nicojs:
### Requirements

### Description of the Change

Fixes the issue where `import { describe, it } from 'mocha'` doesn't work with `--parallel` when using es modules by proxying `describe`, `it` and friends to their actual contextual counterpart.

### Alternate Designs

There is not much else we can do I think, other than going full blown esm (for example, using [package exports](https://nodejs.org/api/packages.html#packages_exports)).

Its imported from core (bare `'mocha'`).

### Benefits

Allows `import {...} from 'mocha'` in an es module setting with `parallel`

### Possible Drawbacks

More code to maintain

### Applicable issues

Closes #4559, should be a patch release.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -112,8 +112,18 @@ exports.before = function(...args) {
 exports.describe = function(...args) {
   (currentContext.describe || currentContext.suite).apply(this, args);
 };
+exports.describe.only = function(...args) {
+  (currentContext.describe || currentContext.suite).only.apply(this, args);
+};
+exports.describe.skip = function(...args) {
+  (currentContext.describe || currentContext.suite).skip.apply(this, args);
+};
+
 exports.it = function(...args) {
   (currentContext.it || currentContext.test).apply(this, args);
+};
+exports.it.only = function(...args) {
+  (currentContext.it || currentContext.test).only.apply(this, args);
 };
 exports.xit = function(...args) {
   (
@@ -121,24 +131,13 @@ exports.xit = function(...args) {
     (currentContext.test && currentContext.test.skip)
   ).apply(this, args);
 };
-exports.setup = function(...args) {
-  (currentContext.setup || currentContext.beforeEach).apply(this, args);
-};
-exports.suiteSetup = function(...args) {
-  (currentContext.suiteSetup || currentContext.before).apply(this, args);
-};
-exports.suiteTeardown = function(...args) {
-  (currentContext.suiteTeardown || currentContext.after).apply(this, args);
-};
-exports.suite = function(...args) {
-  (currentContext.suite || currentContext.describe).apply(this, args);
-};
-exports.teardown = function(...args) {
-  (currentContext.teardown || currentContext.afterEach).apply(this, args);
-};
-exports.test = function(...args) {
-  (currentContext.test || currentContext.it).apply(this, args);
-};
+exports.it.skip = exports.xit;
+exports.setup = exports.beforeEach;
+exports.suiteSetup = exports.before;
+exports.suiteTeardown = exports.after;
+exports.suite = exports.describe;
+exports.teardown = exports.afterEach;
+exports.test = exports.it;
 exports.run = function(...args) {
   currentContext.run.apply(this, args);
 };
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -96,6 +96,53 @@ exports.Suite = Suite;
 exports.Hook = require('./hook');
 exports.Test = require('./test');
 
+let currentContext;
+exports.afterEach = function(...args) {
+  (currentContext.afterEach || currentContext.teardown).apply(this, args);
+};
+exports.after = function(...args) {
+  (currentContext.after || currentContext.suiteTeardown).apply(this, args);
+};
+exports.beforeEach = function(...args) {
+  (currentContext.beforeEach || currentContext.setup).apply(this, args);
+};
+exports.before = function(...args) {
+  (currentContext.before || currentContext.suiteSetup).apply(this, args);
+};
+exports.describe = function(...args) {
+  (currentContext.describe || currentContext.suite).apply(this, args);
+};
+exports.it = function(...args) {
+  (currentContext.it || currentContext.test).apply(this, args);
+};
+exports.xit = function(...args) {
+  (
+    currentContext.xit ||
+    (currentContext.test && currentContext.test.skip)
+  ).apply(this, args);
+};
+exports.setup = function(...args) {
+  (currentContext.setup || currentContext.beforeEach).apply(this, args);
+};
+exports.suiteSetup = function(...args) {
+  (currentContext.suiteSetup || currentContext.before).apply(this, args);
+};
+exports.suiteTeardown = function(...args) {
+  (currentContext.suiteTeardown || currentContext.after).apply(this, args);
+};
+exports.suite = function(...args) {
+  (currentContext.suite || currentContext.describe).apply(this, args);
+};
+exports.teardown = function(...args) {
+  (currentContext.teardown || currentContext.afterEach).apply(this, args);
+};
+exports.test = function(...args) {
+  (currentContext.test || currentContext.it).apply(this, args);
+};
+exports.run = function(...args) {
+  currentContext.run.apply(this, args);
+};
+
 /**
  * Constructs a new Mocha instance with `options`.
  *
@@ -351,20 +398,7 @@ Mocha.prototype.ui = function(ui) {
   bindInterface(this.suite);
 
   this.suite.on(EVENT_FILE_PRE_REQUIRE, function(context) {
-    exports.afterEach = context.afterEach || context.teardown;
-    exports.after = context.after || context.suiteTeardown;
-    exports.beforeEach = context.beforeEach || context.setup;
-    exports.before = context.before || context.suiteSetup;
-    exports.describe = context.describe || context.suite;
-    exports.it = context.it || context.test;
-    exports.xit = context.xit || (context.test && context.test.skip);
-    exports.setup = context.setup || context.beforeEach;
-    exports.suiteSetup = context.suiteSetup || context.before;
-    exports.suiteTeardown = context.suiteTeardown || context.after;
-    exports.suite = context.suite || context.describe;
-    exports.teardown = context.teardown || context.afterEach;
-    exports.test = context.test || context.it;
-    exports.run = context.run;
+    currentContext = context;
   });
 
   return this;
@@ -1299,7 +1333,7 @@ Mocha.prototype.hasGlobalTeardownFixtures = function hasGlobalTeardownFixtures()
  * A (sync) function to assert a user-supplied plugin implementation is valid.
  *
  * Defined in a {@link PluginDefinition}.
- 
+
  * @callback PluginValidator
  * @param {*} value - Value to check
  * @this {PluginDefinition}
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -118,20 +118,20 @@ exports.describe.only = function(...args) {
 exports.describe.skip = function(...args) {
   (currentContext.describe || currentContext.suite).skip.apply(this, args);
 };
-
 exports.it = function(...args) {
   (currentContext.it || currentContext.test).apply(this, args);
 };
 exports.it.only = function(...args) {
   (currentContext.it || currentContext.test).only.apply(this, args);
 };
-exports.xit = function(...args) {
+exports.it.skip = function(...args) {
   (
     currentContext.xit ||
     (currentContext.test && currentContext.test.skip)
   ).apply(this, args);
 };
-exports.it.skip = exports.xit;
+exports.xdescribe = exports.describe.skip;
+exports.xit = exports.it.skip;
 exports.setup = exports.beforeEach;
 exports.suiteSetup = exports.before;
 exports.suiteTeardown = exports.after;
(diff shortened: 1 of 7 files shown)


nicojs:
Of course, my new test is failing on node 10 😇 Will have a look at that soon (probably need to enable [`--experimental-modules`](https://nodejs.medium.com/announcing-a-new-experimental-modules-1be8d2d6c2ff))

nicojs:
I ended up skipping the tests on node 10. Importing esm style from a cjs module is supported from node 12.

giltayar (approved):

juergba:
@nicojs Are `describe`, `it`, ... still on global scope with this solution?

`import { describe, it } from 'mocha'`: we don't need an ESM entry point anymore to achieve this?

giltayar:
@juergba Oh, I didn't notice. You're right!

This is rich! What happened is interesting. Node.js has this mechanism when you `import` a CJS file in that it parses it and tries to determine what the named exports are. It uses a very basic parser that looks for standard patterns like `exports.abc = ...` and others.

What happened was that previously, that parser chocked on the mocha entry point, and defaulted to only default exports. But @nicojs's changes are making that parser very happy, so suddenly it's OK, and mocha is doing named exports from CJS.

Note that my PR is still better because each of the entry points ('/tdd`, '/qunit`...) exports only the correct test functions per-interface, and does it without the CJS parsing thing (slightly faster), whereas this PR exports them ALL.

But it's definitely progress!

nicojs:
Yeah, I think what was happening is that when you `import {describe, it} from mocha` the first time, the ESM - CJS interop layer retrieves the named exports, and caches them in ESM land. This is why, in parallel mode, the first job works and the second one does not, because the tests get added to the old suite (cached `describe` and `it`). 

This PR fixes that, because it allows `describe` and `it` to be cached forever. They will never change since they just proxy to whatever suite is loading at that moment in time. I was actually surprised to see this wasn't done already.

juergba:
@giltayar can you build your ESM entry points on top of this PR? Or does it prevent it in anyway?

> [...] whereas this PR exports them ALL

Do we have any negative implications on that?

@nicojs could you have a short look on our [watcher](https://github.com/mochajs/mocha/blob/master/lib/cli/watch-run.js#L65), please? I remember difficulties with the UI loading. Maybe we can simplify some coding?

giltayar:
@juergba of course. Once this lands, my PR will return to what it was initially: making Mocha a dual-mode library with real ESM entry points that align with the stamdard interfaces. I will remove all the crud from there that pertains to fixing the problem that this PR fixed (much more elegantly).

But that PR now becomes less urgent, and is now a PR that just fine tunes the entry points to be more "correct" and not exposes all test functions of all the interfaces, but exposes entry points for each interface that expose only the functions for that entry point. The main entry point will still expose everything, for backward compatibility.

juergba:
It would be awesome to get some feedback regarding the watcher and especially _lib/browser-entry.js_ .
Our browser tests are weak and I want to make sure to not add any regressions with this PR. Thank you.

nicojs:
Hi  @juergba 👋, I've taken a look.

- `browser-entry.js`: wow I haven't taken a proper look at that file yet. It has some hacks (no offense whatsoever),
    * It overrides mocha.ui and forces a emit of `pre-require`. 
       ```js
       mocha.ui = function(ui) {
        Mocha.prototype.ui.call(this, ui);
        this.suite.emit('pre-require', global, null, this);
        return this;
      };
       ```
        This allowed the `beforeEach`, `describe` etc to be exported in the previous setup. With this new code change, it still works the same way, except there is the proxy function in between, i.e. these functions are now cacheable.
  * However, I think I _did_ break something for both browser and non-browser environments:
   ```shell
   $ git checkout fix/import-parallel
   $ npm run start -- build
   $ node -p 'require("./mocha").setup("bdd").Mocha.describe'
    [Function]
   $ git checkout master
   $ npm run start -- build
   $ node -p 'require("./mocha").setup("bdd").Mocha.describe'
   [Function] { skip: [Function], only: [Function] }
   ```
     So apparently the old `describe`, `it` etc have functions that are no longer exposed with this PR. We will need to fix that! Is there a way to see an extensive list of these functions?
- `lib/cli/watch-run.js`: The way I understand the watch is that this feature is only supported for commonjs scenarios (there is currently no way to clear from the ESM cache, right @giltayar ?). They should still work with the proxying mechanism because test files are cleared from the `require` cache and then reloaded. If they use `const {describe} = require('mocha')` syntax, they should still be allowed to do so, at least, that's what I can deduct from this comment:
    ```js
      // this `require` is needed because the require cache has been cleared.  the dynamic
      // exports set via the below call to `mocha.ui()` won't work properly if a
      // test depends on this module (see `required-tokens.spec.js`).
      const Mocha = require('../mocha');
    ```

lib/mocha.js
```diff
@@ -96,6 +96,53 @@ exports.Suite = Suite;
...
+exports.after = function(...args) {
+  (currentContext.after || currentContext.suiteTeardown).apply(this, args);
+};
+exports.beforeEach = function(...args) {
+  (currentContext.beforeEach || currentContext.setup).apply(this, args);
+};
+exports.before = function(...args) {
+  (currentContext.before || currentContext.suiteSetup).apply(this, args);
+};
+exports.describe = function(...args) {
```
nicojs, line 112: Todo: Also proxy `describe.only`, `describe.skip`, etc.

nicojs (commented):
Don't merge this PR yet!

nicojs:
Ok, i've added proxying of `describe.only` and friends. I think I've got everything covered _except_ for `it.retries` and `describe.retries`. I'm not familiar with that part of our API. @juergba should those also be covered? What `.retries` mean in this context? 

Looking at https://mochajs.org/#retry-tests I can only find `this.retries`.

juergba:
@nicojs thank you for your comments. I try to answer some of your questions.

I understand now that we are only talking about the [require interface](https://mochajs.org/#require).
The keywords `describe` etc. will remain in global scope and therefore we don't put at risk neither our browser nor watcher nor parallel workers.

_lib/cli/watch-run.js_: yes, our watcher is limited to CJS, we have documented that somewhere. Nevertheless some users run ESM in parallel mode, which seems to work. The ESM cache is "cleared" by killing the worker processes on each re-run.

`retries(n)`: has no effect on a _passing_ test. A failing test is re-run max. n times, until it passes or n is reached. In _lib/interfaces/bdd.js_ you see that `it.retries` exists, but `describe.retries` does not. I have never used `it.retries()` yet.

I will need more time to review/test your recent additions. The topic is still fuzzy to me. Where does this `context` object comes from? IMO we don't need to export each and every alias, eg. `describe` yes, but it's alias `context` no.

nicojs:
Thanks for your explanation @juergba 

I've tried to add `it.retries` to the test but it fails because the context seems to always be undefined. This _already was the case_ before this change, so I don't think it ever worked with cjs require syntax,
image

> Where does this `context` object comes from? IMO we don't need to export each and every alias, eg. `describe` yes, but it's alias `context` no.

What do you mean by `context`? This PR keeps a local variable called `currentContext`, which is the current mocha context object for the file that is being loaded, but it doesn't export that.

I agree that not every alias needs exporting. Users who are using the `"bdd"` UI probably don't want `suite` and `test`. However, I don't think we can do anything about it without bigger changes. The way I understand it is that, for the cjs-esm bridge to work, your exports should be cacheable. Since one process could run mocha once for a `"bdd"` project and once for a `"tdd"` project in the same process (using the programmatic API), I don't see any other way. 

One thing we could do is export `suite`/`test` from `'mocha/tdd'` and export `describe`/`it` from `'mocha/bdd'`, something like that. But that would mean a bigger (breaking?) change.

juergba:
@nicojs when I test:
```js
const ste = require('mocha').describe;

ste('test suite', () => {
    it('test', () => {
        console.log('running the test');
    });
});
```
In Mocha v.8.3.0 the test passes, with this branch the result is `TypeError: ste is not a function`.
The same with `import {describe,it} from "mocha";`. v8.3.0. is working and this branch fails.
(this synthax already works in v8.3.0 @giltayar 😵😈)

Am I doing something wrong? Or do your test use the global keywords, not the required/imported ones?

juergba:
Some additional information:
- I installed this branch locally with `npm install https://github.com/mochajs/mocha#2b608c31`:  above test runs succcessfully.
- this branch in my local git clone: I changed to `const ste = require('./mocha/lib/mocha').describe;` and test also passes.
- _test/integration/fixtures/common-js-require.fixture.js:_ this test fails in local git clone
```
    1) Uncaught error outside test suite:
         Uncaught TypeError: Cannot read property 'apply' of undefined
```

juergba:
`I've tried to add it.retries [...]`: I agree, doesn't seem to work, so we just ignore it as before.

I suggest adding `xdescribe`, since we are already exporting `xit`.

Exporting named functions eg. `exports.describe = function describe(...args) { ` would make debugging easier.

giltayar:
@juergba the error in `test/integration/fixtures/common-js-require.fixture.js` is per-spec. The test in that fixture has a delayed run and according to https://mochajs.org/#delayed-root-suite, you need to add a `--delay`. Adding that option when running the test removed the failure. Not sure,BTW, why that option is needed. It is used in `mocha.js`, in this line of the prerequire hook:

```js
    context.run = mocha.options.delay && common.runWithSuite(suite);
```

If I remove the `mocha.options.delay &&`, everything works without the `--delay` flag. 

This _could_ be something that was introduced because there was no proxy? As a hack? Something that maybe isn't needed anymore?

nicojs:
The reason why I've added `--delay` was so I could test that exporting `run` would also work. The test is now testing every single export. I've added a remark in a comment to make this more clear.

> I suggest adding `xdescribe`, since we are already exporting `xit`.

I did that now. Note that this is technically a new feature since it wasn't exported before.

juergba (approved):

test/integration/fixtures/parallel/test3.mjs
```diff
@@ -0,0 +1,7 @@
+import {describe,it} from "../../../../index.js";
+
+describe('test3', () => {
+  it('should fail', () => {
+    throw new Error('expecting this error to fail');
```
joielee09, line 5: @nicojs  hello! I was reviewing your PR, and have a question for this change.
You made three differenct tests, test1, test2, test3. Can you explain about this tests?
Why test1 and test2 pass whereas test3 fail? Also Why do we need test1 and test2 each, when they have the same dir path and they both pass. Thank you 👩

Issue #4535, opened 2020-12-18
content_security_policy - EvalError: Refused to evaluate a string


goatonabicycle:
> 

**The problem:** I've had to add the following line to a chrome extension manifest file in order for the latest version of Mocha to work: ``"content_security_policy": "script-src 'self' 'unsafe-eval'; object-src 'self';",``

Without this line, mocha scripts fails to run with the following error:

```
mocha.js:13165 Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' blob: filesystem:".

    at Function (<anonymous>)
    at mocha.js:13165
    at createCommonjsModule (mocha.js:16)
    at mocha.js:12456
    at mocha.js:4
    at mocha.js:5
```

The mentioned workaround was found here: https://stackoverflow.com/a/48047578 but ideally we don't want to meddle with the content_security_policy

The error seems related to the following piece of code in `mocha.js` : 
```
	  // In sloppy mode, unbound `this` refers to the global object, fallback to
	  // Function constructor if we're in global strict mode. That is sadly a form
	  // of indirect eval which violates Content Security Policy.
	  function () {
	    return this;
	  }() || Function("return this")());
```

This looks like a regression in the newer versions of Mocha as this issue doesn't occur after downgrading to Mocha 8.1.3 . 

**What I've tried:** 
Experimented with different `devtool` settings including ` "cheap-module-source-map"` and `false`

**How I'm running Mocha:** Slightly differently because of testing a chrome extension with a specific page. But the core idea works in other projects and shouldn't be related to the described error.
```javascript
mocha.setup("bdd");
mocha.timeout(8000);
// set up all the relevant tests
mocha.run()
```

**How I'm working around this at the moment:** Downgraded to Mocha 8.1.3

outsideris:
I don't know the root cause, but we introduced it in https://github.com/mochajs/mocha/commit/12db9dbc9dd3b0ec31c182d0e41a6ec615735401 after I tracked it via `git bisect`.

snoack:
[...]

PR #4577, opened 2021-02-18
Fix EvalError caused by regenerator-runtime (#4535)

snoack:
### Description of the Change

Mocha provides a bundle that can be used directly in the browser. For compatibility with Internet Explorer 11, the bundled code is transpiled using Babel. When using `async` functions (or generator functions) as Mocha started in 12db9db (8.2.0), Babel adds the `regenerator-runtime` module to the bundle, which relies on unsafe eval to access the global scope in strict mode, which in turn makes Mocha incompatible with environments that don't allow unsafe eval.

In this PR, I implement a workaround that prevents `regenerator-runtime` from using unsafe eval. First I had to update `regenerator-runtime` since the currently used version relies on `this` in an unbound function referring to the global object which we don't have any control over, while the new version instead exploits implicit global assignments:

```javascript
try {
  regeneratorRuntime = runtime;
} catch (accidentalStrictMode) {
  // This module should not be running in strict mode, so the above
  // assignment should always work unless something is misconfigured. Just
  // in case runtime.js accidentally runs in strict mode, we can escape
  // strict mode using a global Function call. This could conceivably fail
  // if a Content Security Policy forbids using Function, but in that case
  // the proper solution is to fix the accidental strict mode problem. If
  // you've misconfigured your bundler to force strict mode and applied a
  // CSP to forbid Function, and you're not willing to fix either of those
  // problems, please detail your unique predicament in a GitHub issue.
  Function("r", "regeneratorRuntime = r")(runtime);
}
```

While this isn't working either in strict mode without unsafe eval, it allows us to work around the issue by declaring the `regeneratorRuntime` variable before this code is reached (done here via Rollup's `output.intro` setting). That way `regeneratorRuntime = runtime` is no longer an implicit definition of a global variable (which fails in strict mode), but it becomes a simple reassignment of an existing variable, and so the code in the `catch` block (which requires unsafe eval) is never reached.

### Alternate Designs

1. If it wouldn't be for Internet Explorer 11, we wouldn't need to transpile the code in the first place, and this issue wouldn't exist. There is also the option to tell Babel to just not use `transform-regenerator` in which case the transpiled code would at least remain compatible with legacy browsers that support generator functions but might lack other modern ES features.
2. Refactoring the code to not use `async`/`await`.
3. Providing two bundles, one with transpiled code that works in IE11 (but will require unsafe eval), and one that isn't transpiled.

Alternative 1 and 2 were ruled out by @juergba in the discussion in #4535. Alternative 3 seems somewhat more complicated than the workaround I ended up implementing here, and it would require users to choose the correct bundle for their environment.

Because it fixes a regression introduced in Mocha 3.2.0.

### Benefits

Mocha can be used once again in environments where unsafe eval is not allowed. This includes browser extensions, as well as websites with restrictive Content-Security-Policy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
   "devDependencies": {
     "@11ty/eleventy": "^0.11.0",
     "@11ty/eleventy-plugin-inclusive-language": "^1.0.0",
-    "@babel/preset-env": "^7.11.0",
+    "@babel/preset-env": "^7.12.17",
     "@mocha/docdash": "^3.0.1",
     "@rollup/plugin-babel": "^5.1.0",
     "@rollup/plugin-commonjs": "^14.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -503,9 +503,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.12.7",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.12.7.tgz",
-      "integrity": "sha512-YaxPMGs/XIWtYqrdEOZOCPsVWfEoriXopnsz3/i7apYPXQ3698UFhS6dVT1KN5qOsWmVgw/FOrmQgpRaZayGsw==",
+      "version": "7.12.13",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.12.13.tgz",
+      "integrity": "sha512-U/hshG5R+SIoW7HVWIdmy1cB7s3ki+r3FpyEZiCgpi4tFgPnX/vynY80ZGSASOIrUM6O7VxOgCZgdt7h97bUGg==",
       "dev": true
     },
     "@babel/core": {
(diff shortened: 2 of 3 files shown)


juergba:
@snoack thanks for this PR!

The browser tests never work on PR's from forked repos. This is a problem of secrets (saucelabs login info) not being shared with forked repos. I'm going to push your branch to Mocha's repo directly.

package.json
```diff
@@ -133,6 +133,7 @@
     "nyc": "^15.1.0",
     "pidtree": "^0.5.0",
     "prettier": "^1.19.1",
+    "regenerator-runtime": "^0.13.7",
```
juergba, line 136: Do we really need this devDependency? I'm unsure ...
`npm list regenerator-runtime` shows now:
```
mocha@8.3.0 D:\mocha-fork\mocha
+-- @babel/preset-env@7.12.11
| `-- @babel/plugin-transform-regenerator@7.12.1
|   `-- regenerator-transform@0.14.5
|     `-- @babel/runtime@7.12.5
|       `-- regenerator-runtime@0.13.7 
`-- rewiremock@3.14.3
  `-- babel-runtime@6.26.0
    `-- UNMET DEPENDENCY regenerator-runtime@0.11.1

npm ERR! missing: regenerator-runtime@0.11.1, required by babel-runtime@6.26.0
```
I'm not worrying about the error ... I think.
What do you think of upgrading _@babel/preset-env_ ? There is a minor semver.

juergba:
@munter I need your review for this PR, please.

package.json
snoack: Good idea. Done. The result seems to be the same.

juergba:
@snoack have you tested this new bundle in your environment?

snoack:
Yes, I did. Works.

juergba (approved):
@snoack thank you! 
Please be patient for some days, I will wait for more reviews.

juergba:
Babel seems to struggle a bit with publishing, see [Babel v7.12.18](https://github.com/babel/babel/releases/tag/v7.12.18).
v7.12.18 is not available yet on npm.

juergba:
@snoack I'm going to merge this branch tomorrow.

Babel is publishing new releases every few hours, some kind of weird. I don't think we need to upgrade to the newest one, what do you think?

snoack:
Sounds good.

PR #4583, opened 2021-02-24
fix(watch-run.js): pause stdin on SIGINT

jedwards1211:
### Description of the Change

I can describe this in more detail in an issue if you'd like, but I determined through experimentation that this solved funky behavior after interrupting watch mode run by yarn v1, which forwards SIGINT but doesn't wait for the child process to exit.

Even though `yarn run` exits out to the shell immediately on Ctrl-C, if you then press the **up** key while mocha teardown hooks are still running, for some reason it gets routed to mocha and causes this error:

```
node_modules/mocha/lib/runner.js:962
    throw err;
    ^

Error: read EIO
    at TTY.onStreamRead (internal/stream_base_commons.js:209:20)
Emitted 'error' event on ReadStream instance at:
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: -5,
  code: 'EIO',
  syscall: 'read'
}
```

I wonder if this is a peculiarity of macOS terminal (it seems really strange that anything would reach mocha's stdin after the parent process has been terminated?) but adding `process.stdin.pause()` prevents this issue as well as other strange behavior (`^[[A` getting printed instead of the previous command when **up** key is pressed).  There's no need to listen on stdin after mocha has received SIGINT anyway.

### Alternate Designs

Yarn and other programs should of course wait for spawned child processes to exit, but as far as what mocha has control over, this is still an improvement

This amends the core feature of allowing the user to type `rs` in watch mode.

### Benefits

Better behavior when mocha is spawned by programs that don't wait for it to exit when interrupted

### Applicable issues

minor bug fix

### DIFF
--- a/lib/cli/watch-run.js
+++ b/lib/cli/watch-run.js
@@ -211,6 +211,9 @@ const createWatcher = (
 
   let exiting = false;
   process.on('SIGINT', async () => {
+    // stop listening on stdin; otherwise there can be weird behavior if a
+    // parent process like yarn doesn't wait for mocha to exit
+    process.stdin.pause();
     showCursor();
     console.error(`${logSymbols.warning} [mocha] cleaning up, please wait...`);
     if (!exiting) {


jedwards1211:
@juergba seems something is wrong with the browser testing setup

juergba:
@jedwards1211 I don't know, the description of your issue is very fuzzy. I'm unsure wether there actually is an issue at all.
I can't merge a PR, unless I'm convinced there is an issue in first place and without clear issue  I can't review the solution.

Yes, our browser tests don't work for PR's triggered by forked repos, we already have created an issue.

jedwards1211:
@juergba so as long as I make a repo where you can reproduce the issue?

jedwards1211:
@juergba I made a video demonstrating the issue: https://drive.google.com/file/d/11I9JByijs28_gwa5ZdUxG4qkREloujGw/view?usp=sharing

Is this sufficient?  If not let me know what would be sufficient.

juergba:
@jedwards1211 thanks for the video.
It's yarn specific, the error doesn't happen if you launch Mocha directly, right?
I will do some testing on my Windows, using npm.

jedwards1211:
Although it doesn't happen when you launch mocha directly, I still wish you would merge this because it would make things work better and there's no good reason for mocha to keep listening on STDIN once you got an interrupt (you wouldn't want typing `rs` to have any effect at that point...)
Yarn devs aren't going to fix any issues in v1 and the way they made v2 work, there's no way I can use it (like many people).  So my only hope of not having to deal with this is if you merge this PR.

juergba:
I couldn't reproduce this issue using your repo, on Windows10 with _npm_.
But anyway, could you rebase and push, please. Just to be on the safe side.
Also could you repeat your manual tests using the `--parallel` option, please? (workers/subprocesses)

> So my only hope of not having to deal with this is if you merge this PR.

So bad? This isn't really a serious issue?

PR #4603, opened 2021-03-13
[WIP] test for eslint check on CI

outsider-biz:
Ignore this PR. it is just a test PR to check CI from forked PR.

PR #4606, opened 2021-03-14
Run eslint check correctly with forked PRs

outsideris:
Forked PRs have read-only permission.
So, [eslint-plus-action](https://github.com/bradennapier/eslint-plus-action#handle-forked-prs) don't run the job with forked PRs and just show `Read Only & schedule not found, we will not run on this PR` warnings. 

Therefore, ESLint check for forked PRs is always passed and meaningless.

I just run `npm start lint` as we lint our code in local. It will be failed correctly.

And, I also considered just drop [eslint-plus-action](https://github.com/bradennapier/eslint-plus-action#handle-forked-prs). But its annotation feature is helpful for beginners. 
I separated it into `eslint-comment-for-pr.yml` file because it should be scheduled to run every 15 mins. (it means there is a maximum of 15 mins delay) We removed the schedule rule in [#4497](https://github.com/mochajs/mocha/pull/4497), but I don't know the background exactly. 

I believe I can check intensively that these jobs work properly after merging this PR. I already opened [#4603](https://github.com/mochajs/mocha/pull/4603) for test purpose.


juergba:
> I just run `npm start lint` as we lint our code in local. It will be failed correctly.

It fails and shows the error log as local runs do, right? Then I don't understand why you hold on to _eslint-plus-action_. This scheduled job seems such a clumsy way to get information you already have available with the simple `npm start lint`. IMO scheduling this actions every 15m and therefore running ESLint twice is just a waste of GH action time.

.github/workflows/mocha.yml
```diff
@@ -37,19 +37,8 @@ jobs:
...
-          issueSummary: false
-          npmInstall: false
-
-  markdown:
-    name: Markdown Check
+  lint:
+    name: Linting code and markdown
     runs-on: ubuntu-latest
     needs: smoke
     steps:
```
juergba, line 44: There is a v2 release of `actions/setup-node`, should we upgrade?

outsideris:
> its annotation feature is helpful for beginners.

That's why I separate the eslint-plus-action as schedule job.
But, I'm also not sure how it helps for our contributors and I hate 15 mins delay.

> when I launch npm start lint, then also lint.markdown is executed as sub-step.

I missed it. Thank you.

.github/workflows/mocha.yml
outsideris: I checked `actions/setup-node` README. I believe the most helpful change is downloading node from node.js release cache.
I don't know how effective we can get. I'll check it separately from this issue.

outsideris:
* updated setup-node to v2 because there is no reason we don't upgrade it.
* remove eslint-plus-action job because we don't ever use annotation feature. So, we don't need it.
* remove `markdown.lint` step since `lint` run `markdown.lint`

juergba (approved):
@outsideris thank you.

Could you also remove the `purge-expired-artifacts.yml` action, which I added some weeks ago, please. The only reason for this action was `bradennapier/eslint-plus-action@v3.4.2`, so we don't need it anymore either. => see #4565

outsideris:
Sure. removed.

PR #4607, opened 2021-03-14
Add output option to JSON reporter (#4131)

### LINKED ISSUE #4131: Add output option to JSON reporter
I'd like to use test report for other applications (e.g. Slack)
So, JSON file is useful.

`package.json`:

```json
{
  "scripts": {
    "test": "mocha ./test/*-test.js --reporter json --reporter-options output=report.json",
    "posttest": "node slack.js"
  }
}
```

`slack.js`:

```js
const readJSONFile = require('./readJSONFile')
const createMessage = require('./createMessage')
const postMessage = require('./postMessage')

const report = await readJSONFile('report.json')
const message = createMessage(report)
await postMessage(message)
```

- https://github.com/i3oges/mocha-json-file-reporter

dorny:
Extends `JSON` reporter with option to write output directly to file instead of a console.

closes #4131

### Description of the Change

- If output option was provided to JSON reporter:
  - On `EVENT_RUN_END` it will write JSON output to given file - single call to `fs.writeFileSync()`
  - Directory structure will be recursively created if needed - `fs.mkdirSync()`
  - If test runs in browser it will throw unsupported error
- Unit tests has been updated
  - Verifies if reporter writes to given file same JSON content as it sets on the `runner` instance
  - Verifies if reporter throws error if output option is provided and test runs in browser
  - Diff is bigger unless you hide whitespace only changes because existing tests were put into `describe()` group
- Documentation has been updated

### Alternate Designs

`xunit` reporter uses write stream and produces output continuously during test execution.
Similar behavior is provided by `json-stream` reporter.
`json` reporter outputs a single json when the tests have completed so I've used single call to `fs.writeFileSync()` instead of creating write stream.

I've also considered refactoring of `json.spec.js`.
Ideally unit test should execute `JSONReporter` directly instead of using real `Runner` instance.
Also tests should capture standard output and run assertion on it instead of using `testResults` property on runner which doesn't look to be used anywhere else.

I've tried to rewrite tests using `createMockRunner()` from test helpers but it didn't work.
Mock runner doesn't  supports simulation of multiple events and reporter state is not preserved between multiple runs.
Therefore it was not possible to use it for `json` reporter without bigger changes.
I still think refactoring the tests would improve code quality but it is out of scope of this PR.

- It's a common feature, widely supported in other test frameworks (JavaScript or other languages)
- Same option is already supported in `xunit` reporter
- JSON output is usually processed by another tool and any noise on standard output would break it

### Benefits

Compared to redirecting report output to file using shell pipe:
- valid JSON will be written to file even if some test produces noise on standard output
- together with [mocha-multi-reporters](https://www.npmjs.com/package/mocha-multi-reporters) enables having both console output and file output

### DIFF
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -7,6 +7,10 @@
  */
 
 var Base = require('./base');
+var fs = require('fs');
+var path = require('path');
+var errors = require('../errors');
+var createUnsupportedError = errors.createUnsupportedError;
 var constants = require('../runner').constants;
 var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
 var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
@@ -38,6 +42,14 @@ function JSONReporter(runner, options) {
   var pending = [];
   var failures = [];
   var passes = [];
+  var output;
+
+  if (options && options.reporterOptions && options.reporterOptions.output) {
+    if (!fs || !fs.writeFileSync) {
+      throw createUnsupportedError('file output not supported in browser');
+    }
+    output = options.reporterOptions.output;
+  }
 
   runner.on(EVENT_TEST_END, function(test) {
     tests.push(test);
@@ -66,7 +78,13 @@ function JSONReporter(runner, options) {
 
     runner.testResults = obj;
 
-    process.stdout.write(JSON.stringify(obj, null, 2));
+    var json = JSON.stringify(obj, null, 2);
+    if (output) {
+      fs.mkdirSync(path.dirname(output), {recursive: true});
+      fs.writeFileSync(output, json);
+    } else {
+      process.stdout.write(json);
+    }
   });
 }
 
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -9,14 +9,14 @@
 var Base = require('./base');
 var fs = require('fs');
 var path = require('path');
-var errors = require('../errors');
-var createUnsupportedError = errors.createUnsupportedError;
+const createUnsupportedError = require('../errors').createUnsupportedError;
+const utils = require('../utils');
 var constants = require('../runner').constants;
 var EVENT_TEST_PASS = constants.EVENT_TEST_PASS;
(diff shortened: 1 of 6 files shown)


lib/reporters/json.js
```diff
@@ -66,7 +78,13 @@ function JSONReporter(runner, options) {
 
     runner.testResults = obj;
 
-    process.stdout.write(JSON.stringify(obj, null, 2));
+    var json = JSON.stringify(obj, null, 2);
+    if (output) {
+      fs.mkdirSync(path.dirname(output), {recursive: true});
+      fs.writeFileSync(output, json);
```
juergba, line 84: When one of these file operations fails, the error will be swallowed and nothing happens. This is an known problem with Mocha's reporters. So we could use `try/catch` and log the error to the console?

juergba (commented):
@dorny thank you for this PR.
Yes, _json.spec.js_ could benefit from some refactoring ...
I just have one remark, otherwise looks like a very diligent work.

juergba:
The root of the error being swallowed is, that we have two `EVENT_RUN_END` listeners, with incorrect order:
a) one is registered within the `runner` and sets `runner.state` to stopped (should run first)
b) one is registered within the reporter (runs first, should be second)

So I see three options:
- we use `try/catch` in `json` reporter
- we use `emitter.prependListener` for registering a), but I'm unsure about browser compability.
- we delay the execution of b) with `setImmediate()`

juergba:
Yes, and there is Mocha in parallel mode, quite a recent feature. I'm unsure about the implications, but our docs states:
`In parallel mode, reporter output is buffered; reporting will occur after each file is completed.`

We have to make sure there will be only one file created.

dorny:
@juergba thanks for your support. Right now I'm recovering from eye surgery so I've to reduce my screen time significantly for next few days. I will get back to this afterwards. Meanwhile feel free to modify PR yourself if you find solution.

lib/reporters/json.js
juergba: done

juergba (approved):
@dorny thank you

PR #4614, opened 2021-03-28
Watch for test files are crashed

### LINKED ISSUE #4580: Watcher crashes when reloading code that throws (regression mocha@7 to mocha@8)
### Prerequisites

### Description

MCVE: please see https://github.com/marcelbeumer/mocha-8-watcher-issue for a minimal test case with STR.

Since mocha@8 the watcher crashes when reloading tests and a require fails, for example when causing a syntax error. Errors that happen during test execution are handled correctly by the watcher. Experimenting with different mocha versions, it turns out it is a regression from v7 to v8.

### Steps to Reproduce

See MCVE: https://github.com/marcelbeumer/mocha-8-watcher-issue. It seems that when the watcher is reloading code that will throw when requiring, the watcher crashes.

**Expected behavior:**

The watcher does not crash when reloading test code. Behavior for the initial test run when starting the watcher is debatable but would be best if it handled both cases well.

**Actual behavior:** 

The watcher crashes when reloading test code that causes an error.

**Reproduces how often:** 

100%

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 8.30 (but also tried @8.0, and @7)
- The output of `node --version`: v15.6.0
- Your operating system
  - name and version: macOS 11.1
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): no, but same behavior with ts-node setup (see branch in https://github.com/marcelbeumer/mocha-8-watcher-issue)

outsideris:
### Description of the Change
In watch mode, the watcher crashed when there are syntax errors in test files during `require('test file')`.
It only happens in serial mode and Node v15+.

This `try - catch` statement removed [when we introduce parallel mode](https://github.com/mochajs/mocha/commit/2078c3203f55ee0e1783dc2fc9e20ac51dcb6896#diff-14a1e3aa193d6f208b9ffdfeec1876f2cbc0e51007d061a87644e9ea9df59676L94).
I think we need `try - catch` here.

And I added test cases for the crash case to prevent regression.

Watching should be worked.

### Benefits
Users can use watch mode when the test code is broken.

### Applicable issues
Fix #4580

### DIFF
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -470,6 +470,7 @@ async function runMochaWatchJSONAsync(args, opts, change) {
       // eslint-disable-next-line no-control-regex
       .replace(/\u001b\[\?25./g, '')
       .split('\u001b[2K')
+      .filter(x => x)
       .map(x => JSON.parse(x))
   );
 }
--- a/test/integration/options/watch.spec.js
+++ b/test/integration/options/watch.spec.js
@@ -47,6 +47,23 @@ describe('--watch', function() {
       });
     });
 
+    it('reruns test when watched test file is crashed', function() {
+      const testFile = path.join(tempDir, 'test.js');
+      copyFixture(DEFAULT_FIXTURE, testFile);
+
+      return runMochaWatchJSONAsync([testFile], tempDir, () => {
+        replaceFileContents(testFile, 'done();', 'done((;');
+      })
+        .then(() => {
+          return runMochaWatchJSONAsync([testFile], tempDir, () => {
+            replaceFileContents(testFile, 'done((;', 'done();');
+          });
+        })
+        .then(results => {
+          expect(results, 'to have length', 1);
+        });
+    });
+
     describe('when in parallel mode', function() {
       it('reruns test when watched test file is touched', function() {
         const testFile = path.join(tempDir, 'test.js');
@@ -57,6 +74,23 @@ describe('--watch', function() {
         }).then(results => {
           expect(results, 'to have length', 2);
         });
+      });
+
+      it('reruns test when watched test file is crashed', function() {
+        const testFile = path.join(tempDir, 'test.js');
+        copyFixture(DEFAULT_FIXTURE, testFile);
+
+        return runMochaWatchJSONAsync(['--parallel', testFile], tempDir, () => {
+          replaceFileContents(testFile, 'done();', 'done((;');
+        })
+          .then(() => {
+            return runMochaWatchJSONAsync([testFile], tempDir, () => {
+              replaceFileContents(testFile, 'done((;', 'done();');
+            });
+          })
+          .then(results => {
+            expect(results, 'to have length', 1);
+          });
       });
     });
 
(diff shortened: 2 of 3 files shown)


test/integration/helpers.js
```diff
@@ -470,6 +470,7 @@ async function runMochaWatchJSONAsync(args, opts, change) {
       // eslint-disable-next-line no-control-regex
       .replace(/\u001b\[\?25./g, '')
       .split('\u001b[2K')
+      .filter(x => x)
```
outsideris, line 473: I filtered output because the output is an empty string(`''`) when a test file is crashed.

test/integration/options/watch.spec.js
```diff
@@ -47,6 +47,23 @@ describe('--watch', function() {
       });
     });
 
+    it('reruns test when watched test file is crashed', function() {
```
juergba, line 50: reruns test when watched test file ~~is crashed~~ crashes

test/integration/options/watch.spec.js
```diff
@@ -47,6 +47,23 @@ describe('--watch', function() {
...
+        replaceFileContents(testFile, 'done();', 'done((;');
+      })
+        .then(() => {
+          return runMochaWatchJSONAsync([testFile], tempDir, () => {
+            replaceFileContents(testFile, 'done((;', 'done();');
+          });
+        })
+        .then(results => {
+          expect(results, 'to have length', 1);
+        });
```
juergba, line 64: It's not necessary to call `runMochaWatchJSONAsync`  twice. You don't process the result of the first call (good, then bad test) anyway. `results` contains only the second call (bad, then good test).

The following runs first the bad, then the good test with only one call:
```js
it('reruns test when watched test file crashes', function() {
      const testFile = path.join(tempDir, 'test.js');
      copyFixture(DEFAULT_FIXTURE, testFile);

      replaceFileContents(testFile, 'done();', 'done((;');

      return runMochaWatchJSONAsync([testFile], tempDir, () => {
        replaceFileContents(testFile, 'done((;', 'done();');
      })
        .then(results => {
          expect(results, 'to have length', 1);
        });
    });
```

test/integration/options/watch.spec.js
```diff
@@ -58,6 +75,23 @@ describe('--watch', function() {
...
+          replaceFileContents(testFile, 'done();', 'done((;');
+        })
+          .then(() => {
+            return runMochaWatchJSONAsync([testFile], tempDir, () => {
+              replaceFileContents(testFile, 'done((;', 'done();');
+            });
+          })
+          .then(results => {
+            expect(results, 'to have length', 1);
+          });
```
juergba, line 93: same

juergba (commented):
Your bug fix lgtm, but the tests are a little weird.

test/integration/options/watch.spec.js
outsideris: Smart approach!!

outsideris:
Fixed

juergba (commented):
I forgot to approve this PR. Yes, approved.

PR #4616, opened 2021-03-31
Run browser tests on forked PRs by a dedicated label

### LINKED ISSUE #4553: Github actions: browser tests fail for PR's raised from forked repos
### Description

When a PR is created/synchronized from a forked repo, a `pull_request` event is emitted.
```
In order to protect public repositories for malicious users we run all pull request workflows raised from forks
with a read-only token and no access to secrets.
```
Unfortunately we don't have access to secrets and therefore the login to Sauce Labs fails.

There is a new event `pull_request_target` which gives access to secrets, but also grants a read/write token for this repo. Which is bad and opens door for unfriendly people.

outsideris:
### Description of the Change
It is the only way I found to run browser tests on forked PR. 

After reviewing code of a forked PR, maintainers can attach [a `run-browser-test` label](https://github.com/mochajs/mocha/labels/run-browser-test).
Maintainers should check there is no malicious code to steal our secrets or do something bad on our repository.

This `Browser Tests` workflow will run based on forked PRs with writing repo and access secrets permissions. So, it can be  dangerous if the forked PR has malicious code.

I've tested `pull_request_target` event.

스크린샷 2021-03-31 오후 5 48 07

When users open a pull request, `push`, `pull_reqeust` and `pull_request_target`  will be run automatically. But `pull_request_target: types: [labeled]`(in this case `Browser Tests`) will not be run because `labeled` type.

스크린샷 2021-03-31 오후 5 48 54

If maintainers attach a label except `run-browser-test` label, `Browser Tests` workflow will be triggered, but it skipped because the label is not `run-browser-test`. In the above screenshot, `pull request safe run` workflow is `pull_request_target` workflow with `labeled` type. 

스크린샷 2021-03-31 오후 5 49 32

When maintainers attach the `run-browser-test` label, `Browser Tests` workflow will be triggered. In above screenshot, I tested it with `safe-to-run` label.

스크린샷 2021-03-31 오후 5 50 51

After that, when author of the forked PR pushed new commits, only default workflows will. be run. 

That means maintainers should remove and re-attache `run-browser-test` label to run `Browser Tests` again. 

And when the forked PR already has `run-browser-test` label, `Browser Tests` workflow will be triggered if maintainers attach a new label except `run-browser-test` label. **We have to be careful with this.**

### Alternate Designs
We can use [`workflow_dispatch` event](https://github.com/mochajs/mocha/issues/4553#issuecomment-806439992). `workflow_dispatch` is a workflow that we should trigger it manually with PR number.

I think there are some cons.

* Process is not convenient. maintainers enter PR number and trigger the workflow manually. 
* It couldn't be integrated with PR. So, we cannot the result of the workflow in GitHub's checks tab. It is hard to know whether browser tests are successful or fail. 

Therefore, I think `pull_request_target` event with `labeled` type is better than `workflow_dispatch` event.

All forked PRs failed because of browser tests.

### Applicable issues
close #4553

### DIFF
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -1,4 +1,4 @@
-name: Browser Tests
+name: Browser Tests for foked PRs
 'on':
   pull_request_target:
     types:
@@ -34,3 +34,10 @@ jobs:
         env:
           SAUCE_USERNAME: '${{secrets.SAUCE_USERNAME}}'
           SAUCE_ACCESS_KEY: '${{secrets.SAUCE_ACCESS_KEY}}'
+      - name: remove 'run-browser-test' label
+        uses: buildsville/add-remove-label@v1
+        if: ${{ always() }}
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
+          label: run-browser-test
+          type: remove
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -1,4 +1,4 @@
-name: Browser Tests for foked PRs
+name: Browser Tests for forked PRs
 'on':
   pull_request_target:
     types:
@@ -22,8 +22,6 @@ jobs:
         uses: actions/cache@v2
         with:
           path: '~/.npm'
-          # this key is different than above, since we are running scripts
-          # (builds, postinstall lifecycle hooks, etc.)
           key: "ubuntu-latest-node-full-v14-${{ hashFiles('**/package-lock.json') }}"
           restore-keys: |
             ubuntu-latest-node-full-v14-
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -1,0 +1,36 @@
+name: Browser Tests
+'on':
+  pull_request_target:
+    types:
+      - labeled
+
+jobs:
+  test-browser:
+    # TODO: configure to retain build artifacts in `.karma/` dir
+    name: 'Browser Tests'
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'run-browser-test')
+    steps:
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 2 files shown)


juergba:
> When users open a PR, `push`, `pull_request` and `pull_request_target` will be run automatically.

There shouldn't be any `push` event for users = forked PRs, right?

Your solution is awkward, but I don't see any alternative way either...

You could remove the `run-browser-test` label automatically after browser test's completion.
There is a GH action [re-run-workflow](https://github.com/marketplace/actions/re-run-workflow) which:
`once-label: When this label is added to a pull request, re-run the workflow once and remove the label again.`
What do you think about above GH re-run-workflow?
Or any other app which removes labels (search the marketplace with "remove labels").

outsideris:
> There shouldn't be any push event for users = forked PRs, right?

No, it shouldn't be triggered by forked PRs. I just mentioned `push, pull_request and pull_request_target ` because I want to explain normal workflows like our smoke, lint and tests. 

> You could remove the run-browser-test label automatically after browser test's completion.

That's a good idea. I will try it.

outsideris:
I used [Add Remove Label](https://github.com/marketplace/actions/add-remove-label) actions.

It will remove `run-browser-test` label after browser test workflow whether it succeeds or fails.

And I will more test with real PRs after merging this because there are some limitations outside of this repository.

.github/workflows/browser-test.yml
```diff
@@ -0,0 +1,43 @@
+name: Browser Tests for foked PRs
```
juergba, line 1: typo: foked => forked

.github/workflows/browser-test.yml
```diff
@@ -0,0 +1,43 @@
...
+          node-version: 14
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: 'Cache node_modules'
+        uses: actions/cache@v2
+        with:
+          path: '~/.npm'
+          # this key is different than above, since we are running scripts
+          # (builds, postinstall lifecycle hooks, etc.)
```
juergba, line 26: These comments could be removed.

juergba (approved):
I have some small comments only, lgtm.

outsideris:
fixed. 

I will merge this to test with real forked PRs.

PR #4617, opened 2021-04-02
docs: dynamically generating tests with `top-level await` and ESM test files

juergba:
### Description of the Change

With `top-level await` test data can be collected in a dynamic and asynchronous way, while the test file is loaded and before the test `runner` is created. Since ESM files are already loaded in an asynchronous way, no changes to our code base are needed.

Requisite:
- `top-level await`: since Node >= v14.8.0
- `ESM` (ECMAScript module) test file

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -684,7 +684,7 @@ Given Mocha's use of function expressions to define suites and test cases, it's 
 Take the following example:
 
 ```js
-const assert = require('chai').assert;
+const assert = require('assert');
 
 function add(args) {
   return args.reduce((prev, curr) => prev + curr, 0);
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -734,6 +734,32 @@ describe('add()', function() {
 });
 ```
 
+With `top-level await` you can collect your test data in a dynamic and asynchronous way while the test file is being loaded:
+
+```js
+// testfile.mjs
+// top-level await: Node >= v14.8.0 with ESM test file
```
outsideris, line 741: How about we provide full example to users can test it by copying this example.

```
import assert from 'assert';

function add(args) {
  return args.reduce((a, b) => a + b);
}
```

outsideris (approved):
Wow! I didn't know that.

docs/index.md
juergba: done

Issue #4453, opened 2020-09-28
mochawesome, mocha-reporter-junit, third-party reporters do not work in parallel mode


gurjeetbains:
Here is the screenshot of report in parallel mode 
![image](https://user-images.githubusercontent.com/19879766/94417432-8e7e3200-01dc-11eb-9a9a-4c90eb9e45d3.png)

Here is the screenshot of same in serial mode
![image](https://user-images.githubusercontent.com/19879766/94417809-0b111080-01dd-11eb-832b-9014ca33a4cb.png)
Here is the details of package.json files:-
`{
  "name": "playwrightmocha",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "chai": "^4.2.0",
    "mocha": "^8.1.3",
    "mochawesome": "^6.1.1",
    "playwright": "^1.4.2",
    "playwright-cli": "^0.5.2"
  },
  "mocha": {
    "parallel": true,
    "reporter": [
      "spec",
      "json",
      "mochawesome"
    ],
    "timeout": 300000
  }
}
`

boneskull:
you and/or the mochawesome maintainer(s) would need to find the root cause of the failure; once we have that, we can see if it's fixable.  the data reporters get in parallel mode is _not_ the same as serial mode; see #4403 

going to leave this open, but it's not actionable until we know more about why.

(note: mochawesome is not maintained by the mocha team)

gurjeetbains:
@boneskull  I tried with the junit-reporter for mocha and got the below error. Mochawesome I don't have error 
D:\My Practice Projects\playwrightmocha\node_modules\mocha\lib\nodejs\parallel-buffered-runner.js:269
          throw err;
          ^

TypeError: Cannot read property 'testsuite' of undefined
    at lastSuite (D:\My Practice Projects\playwrightmocha\node_modules\mocha-junit-reporter\index.js:209:46)
    at MochaJUnitReporter.<anonymous> (D:\My Practice Projects\playwrightmocha\node_modules\mocha-junit-reporter\index.js:234:5)
    at ParallelBufferedRunner.emit (events.js:327:22)
    at ParallelBufferedRunner.Runner.fail (D:\My Practice Projects\playwrightmocha\node_modules\mocha\lib\runner.js:411:8)
[...]

PR #4623, opened 2021-04-07
fix xunit reporter crash when used with parallel mode

curtisman:
### Requirements

### Description of the Change

See this comment https://github.com/mochajs/mocha/issues/4453#issuecomment-812936325

This change fixes the various issue encountered when using parallel mode with the `xunit` reporter when error happens.

All the changes are in `Test`, `Suite` and `Hook`'s serialize function
1. serialized parent fully by calling `serialize` (also call `serialize` for Hook's `ctx.currentTest` as well)
2. force serialization of the value for  `$$isPending` to `boolean` to remove the possiblity that it is `undefined` which is not serialized.
3. Add `state` to the value that is serialized in `Hook.prototype.serialized()`

Fixes broken functionality in the core

### Benefits

`xunit` reporter can be used in parallel mode

### Possible Drawbacks

I didn't fully look at all the possible reporters mentioned in #4453 and more fixes may be needed to address the full scope.

### Applicable issues

Partial bug fix to #4453


curtisman:
Oops.  Need to figure out the CLA first.  Closing the PR for now.

curtisman:
Rebased and signed the CLA

curtisman:
Reactivating PR after signing CLA

curtisman:
@juergba I believe this should fix #4624

koerbcm:
+1

juergba:
@curtisman thank you for this PR. Sorry for the delay, are you still around?

The communication between processes (IPC) with its de-/serialization is a time-costly matter. We have to be very selective with the data we are transmitting, otherwise the parallel mode turns out to be slower than a serial run.

> 1. serialized parent fully by calling `serialize` (also call `serialize` for Hook's ctx.currentTest as well)

This is too generous. I will have a look which properties we will need. At a first look, adding just `$$fullTitle: this.parent.fullTitle()` seems to be sufficient.

lib/hook.js
```diff
@@ -62,22 +62,18 @@ Hook.prototype.error = function(err) {
  */
 Hook.prototype.serialize = function serialize() {
   return {
-    $$isPending: this.isPending(),
+    $$isPending: !!this.isPending(),
```
juergba, line 65: please use `Boolean()`  here.

lib/hook.js
```diff
@@ -62,22 +62,18 @@ Hook.prototype.error = function(err) {
...
     $$titlePath: this.titlePath(),
     ctx:
       this.ctx && this.ctx.currentTest
         ? {
-            currentTest: {
-              title: this.ctx.currentTest.title,
-              [MOCHA_ID_PROP_NAME]: this.ctx.currentTest.id
-            }
+            currentTest: this.ctx.currentTest.serialize()
```
juergba, line 70: Do we need any additional data here?

lib/hook.js
```diff
@@ -62,22 +62,18 @@ Hook.prototype.error = function(err) {
...
           }
         : {},
-    parent: {
-      [MOCHA_ID_PROP_NAME]: this.parent.id
-    },
+    parent: this.parent.serialize(),
```
juergba, line 73: Adding `$$fullTitle: this.parent.fullTitle()` should be sufficient.

lib/hook.js
```diff
@@ -62,22 +62,18 @@ Hook.prototype.error = function(err) {
...
+    state: this.state,
     title: this.title,
     type: this.type,
-    [MOCHA_ID_PROP_NAME]: this.id
+    [MOCHA_ID_PROP_NAME]: this[MOCHA_ID_PROP_NAME]
```
juergba, line 77: `this.id` is the getter for `[MOCHA_ID_PROP_NAME]`, so why this change here?

lib/suite.js
```diff
@@ -594,11 +594,11 @@ Suite.prototype.serialize = function serialize() {
   return {
     _bail: this._bail,
     $$fullTitle: this.fullTitle(),
-    $$isPending: this.isPending(),
+    $$isPending: !!this.isPending(),
     root: this.root,
     title: this.title,
-    id: this.id,
-    parent: this.parent ? {[MOCHA_ID_PROP_NAME]: this.parent.id} : null
+    parent: this.parent ? this.parent.serialize() : null,
```
juergba, line 600: We have to keep the data serialized as small as possible, why should we use the entire parent here?

lib/suite.js
```diff
@@ -594,11 +594,11 @@ Suite.prototype.serialize = function serialize() {
...
+    [MOCHA_ID_PROP_NAME]: this[MOCHA_ID_PROP_NAME]
```
juergba, line 601: agreed, `id` has no setter, so: `[MOCHA_ID_PROP_NAME]: this.id`

lib/suite.js
```diff
@@ -594,11 +594,11 @@ Suite.prototype.serialize = function serialize() {
   return {
     _bail: this._bail,
     $$fullTitle: this.fullTitle(),
-    $$isPending: this.isPending(),
+    $$isPending: !!this.isPending(),
```
juergba, line 597: Boolean()

lib/test.js
```diff
@@ -92,22 +92,19 @@ Test.prototype.serialize = function serialize() {
   return {
     $$currentRetry: this._currentRetry,
     $$fullTitle: this.fullTitle(),
-    $$isPending: this.pending,
+    $$isPending: !!this.pending,
```
juergba, line 95: Boolean()

lib/test.js
```diff
@@ -92,22 +92,19 @@ Test.prototype.serialize = function serialize() {
...
     $$slow: this._slow,
     $$titlePath: this.titlePath(),
     body: this.body,
     duration: this.duration,
     err: this.err,
-    parent: {
-      $$fullTitle: this.parent.fullTitle(),
-      [MOCHA_ID_PROP_NAME]: this.parent.id
-    },
+    parent: this.parent.serialize(),
```
juergba, line 102: do we need the entire parent? IMO no

lib/test.js
```diff
@@ -92,22 +92,19 @@ Test.prototype.serialize = function serialize() {
...
     speed: this.speed,
     state: this.state,
     title: this.title,
     type: this.type,
     file: this.file,
-    [MOCHA_ID_PROP_NAME]: this.id
+    [MOCHA_ID_PROP_NAME]: this[MOCHA_ID_PROP_NAME]
```
juergba, line 108: `this.id` is ok

juergba (changes requested):

lib/hook.js
curtisman: Done.

lib/test.js
curtisman: Done

lib/test.js
curtisman: Reverted

lib/suite.js
curtisman: Done

lib/suite.js
curtisman: Done.

lib/hook.js
curtisman: Reverted.

lib/test.js
curtisman: Reverted

curtisman:
I reverted the change to fully serialized nested objects.  Although, I feel like this will just turn in the whack-a-mole, as I don't think there is a good documentation on what reporter are expected to have access to in these objects, and other reporters might us a fuller set of data here.  

I tested out my original repro in https://github.com/mochajs/mocha/issues/4453#issuecomment-812936325 with the json reporter as well, and discover a set of property that needed to be added to the `Hook` as well (`currentRetry()`, `fullTitle()`, `duration`, `file`)

lib/suite.js
curtisman: Reverted.

lib/hook.js
curtisman: Just to the consistent.  Didn't know the size of serialization is a concern. Reverted.

lib/hook.js
curtisman: Done.

juergba:
> Although, I feel like this will just turn in the whack-a-mole [...]

Yes, I feel the same way. This entire IPC implementation with its de-/serialization appears so clumsy to me. "speed" is the only argument which could justify the additional maintenance burden we have with `parallel` mode. I don't want to risk loosing any speed. 

So be prepared for more nagging on your JSON fix.

curtisman:
> I don't want to risk loosing any speed.

I totally hear you there.  For now, I am fully ok with just getting this PR in following the pattern of what is there already, to fix the bug to unblock people.

I poked around the code a little bit (specifically deserialization), and there are some pretty easy improvements that can be made to speed that part up, but I don't know whether it will have any effect on the overall scenarios that people care about. 

What scenarios are we looking at for measuring speed? Any data on which part of it causing speed issues (serialization? deserialization? IPC between worker and main process?  contention on the main process?...)

juergba:
> [...] and there are some pretty easy improvements that can be made [...]

yes please, go ahead, your PRs are very welcome.

> What scenarios are we looking at for measuring speed?

Nothing sophisticated, just the total time. The user sitting in front of his/her terminal: serial vs parallel.
The communication main process=>worker is small (just the options?)
You can't really influence or customize the IPC chanel.
So the leverage point is worker=>main, the de-/serialization of the workers' event data.
Back on main, there is little happening. Just emitting the buffered events and creating the report.

I can recommend to read:
- [essay](https://developer.ibm.com/articles/parallel-tests-mocha-v8/): by the author of parallel mode
- [mochawesome](https://github.com/adamgruber/mochawesome/pull/331): implementing parallel, with too little event data

juergba (approved):
@curtisman thank you!
No tests are ok for now. There will be more `parallel` fixes to come anyway.

PR #4626, opened 2021-04-12
Add SCSS and initial component styles

sjtrimble:
### Requirements
N/A

### Description of the Change

- Added SCSS support for documentation site
- Added initial component styles for documentation site

### Alternate Designs

- Pure CSS was considered but would require more code writing
 
- Documentation website updates

### Benefits

- Docs site refresh
- Additional SCSS features such as nesting

### Possible Drawbacks

- Contributors may feel inadequate to contribute to SCSS (although CSS works just as well)

### Applicable issues

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
   },
   "scripts": {
     "prepublishOnly": "nps test clean build",
-    "sass": "node-sass -w docs/scss/ -o docs/css/ --recursive",
     "start": "nps",
     "test": "nps test",
     "version": "nps version",
@@ -69,7 +68,6 @@
     "minimatch": "3.0.4",
     "ms": "2.1.3",
     "nanoid": "3.1.20",
-    "node-sass": "^5.0.0",
     "serialize-javascript": "5.0.1",
     "strip-json-comments": "3.1.1",
     "supports-color": "8.1.1",
@@ -145,6 +143,7 @@
     "rollup-plugin-node-globals": "^1.4.0",
     "rollup-plugin-node-polyfills": "^0.2.1",
     "rollup-plugin-visualizer": "^4.1.0",
+    "sass": "^1.32.12",
     "sinon": "^9.0.3",
     "strip-ansi": "^6.0.0",
     "svgo": "^1.3.2",
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
   },
   "scripts": {
     "prepublishOnly": "nps test clean build",
+    "sass": "node-sass -w docs/scss/ -o docs/css/ --recursive",
     "start": "nps",
     "test": "nps test",
     "version": "nps version",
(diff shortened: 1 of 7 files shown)


sjtrimble:
@boneskull here is the initial PR (testing the process more so) for adding SCSS to the docs site and some initial styles. I can keep building it once this foundation is in. I may need help to land this first PR as I get used to your process :)

outsideris:
@sjtrimble seems our designer to redesign our site and SCSS supports is needed for one's work.
We need to confirm by @boneskull .

boneskull:
Yes, @sjtrimble is our designer.

package.json
```diff
@@ -68,6 +69,7 @@
     "minimatch": "3.0.4",
     "ms": "2.1.3",
     "nanoid": "3.1.20",
+    "node-sass": "^5.0.0",
```
boneskull, line 72: `node-sass` needs to be in `devDependencies`.  so run `npm rm node-sass && npm i node-sass -D` which will update `package-lock.json` as well (I think)

I don't see where the `sass` script is actually run, though. We won't see any changes on Netlify until it runs automatically.  I think what we should do is:

1. Revert the change on L49 of `package.json`
2. Apply the following diff to `package-scripts.js`:

```diff
diff --git a/package-scripts.js b/package-scripts.js
index 213be777..430e28e8 100644
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -263,9 +263,14 @@ module.exports = {
     docs: {
       default: {
         script:
-          'nps docs.clean && nps docs.api && eleventy && nps docs.linkcheck && node scripts/netlify-headers.js docs/_site >> docs/_site/_headers',
+          'nps docs.clean && nps docs.scss && nps docs.api && eleventy && nps docs.linkcheck && node scripts/netlify-headers.js docs/_site >> docs/_site/_headers',
         description: 'Build documentation'
       },
+      scss: {
+        script: 'node-sass -w docs/scss/ -o docs/css/ --recursive',
+        description: 'Compile SCSS',
+        hiddenFromHelp: true
+      },
       production: {
         script: 'nps docs && nps docs.postbuild',
         description: 'Build docs for production'
```

We'll also need to ensure that the generated files are added to `.gitignore` if not already.

sjtrimble:
FYI I did some more research and the recommended package is actually "just" sass, the dart version.

"sass - npm" https://www.npmjs.com/package/sass

boneskull:
@sjtrimble I've switched to `sass`, updated the `.gitignore` file, and added the script to the docs build.  The `main.scss` file is empty, of course--not sure if that was intended.  See my extended commit message for usage

sjtrimble:
@boneskull Perfect, looks good on my end (although I haven't tested locally).

Yes, the `main.scss` file should be empty as I will be populating the code there and building out the general SCSS folder structure in general. I will be moving the additions on `style.scss` over there.

cddsteen (approved):

PR #4629, opened 2021-04-23
GH actions: add Node v16

juergba:
### Description of the Change

Replaces Node v15 by v16.
End-of-life of v15 is 2021-06-01.


giltayar:
Are you going to kill node 10 support? I'd like to do a pr that won't work for node 10 (esm related)

outsideris (approved):

PR #4633, opened 2021-05-07
Drop support of Node v10

juergba:
### Description of the Change

We drop support of Node v10 with end-of-life: 2021-04-30

### DIFF
--- a/.github/workflows/mocha.yml
+++ b/.github/workflows/mocha.yml
@@ -25,7 +25,6 @@ jobs:
           - ubuntu-latest
           - windows-2019
         node:
-          - 10
           - 12
           - 14
           - 16
@@ -70,7 +69,6 @@ jobs:
           - ubuntu-latest
           - windows-2019
         node:
-          - 10
           - 12
           - 14
           - 16
--- a/docs/index.md
+++ b/docs/index.md
@@ -64,7 +64,7 @@ or as a development dependency for your project:
 $ npm install --save-dev mocha
 ```
 
-> As of v8.0.0, Mocha requires Node.js v10.12.0 or newer.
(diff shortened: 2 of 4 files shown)


giltayar:
Can we make the engine minimum be 12.7.0? That version is the minimum for esm compatibility, which means that we can drop all kinds of ifs in the tests and the code around that.

I'd love to submit a pr around that, which should be part of this semver major release.

juergba:
> Can we make the engine minimum be 12.7.0?

Yes, but in _utils.supportsEsModules_ you are checking for >=v12.11. Now we take 12.7.0, sure?

> I can create it as a separate pr [...]

Please, go ahead then. We don't have enough content for a serious semver major, yet.

giltayar:
@juergba I'll get back to you on the minimum number. I'll check it while building the PR.

giltayar:
I checked, and the version that finalized support for ESM was 12.17.0 (I confused 7 and 17). Note that the version that removed the "experimental ESM" warning was 12.20.0, which is a pretty late v12 version (the latest v12 version is 12.22.1).

This means that if I implement the PR that removes the "`require` and if fails then `import`", anybody using v12 that is earlier than v12.20.0 will get a warning from ESM. I'll go ahead and implement it anyway, and we can discuss this on that PR.

juergba:
@giltayar I'm going to merge this PR, you can adjust the minimum version later on.
I hope you are alright in your country in these difficult times.

giltayar:
I would love it if we could add this new PR to the next semver-major version of Node.js, and before the semver-major version is launched: https://github.com/mochajs/mocha/pull/4635. This PR could only have been done when we dropped Node v10. The reasons are outlined in the PR.

PR #4640, opened 2021-05-29
Add new option "--dry-run"

juergba:
### Description

Adds a `--dry-run` option, which executes the `runner` and makes all `tests` pass without actually running any `hook`/`test`.

### Description of the Change

- add option `--dry-run`
- pass it to the `mocha`, then `runner` instance
- do not execute any `hook` or `test`
- other options as `skip` or `only` work as in a normal run

additionally:
- remove `utils.dQuote()` and `utils.sQuote()`
- use `template strings` instead

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -874,6 +874,12 @@ Use this option to have Mocha check for global variables that are leaked while r
 
 > _`--compilers` was removed in v6.0.0. See [further explanation and workarounds][mocha-wiki-compilers]._
 
+### `--dry-run`
+
+> _New in v9.0.0._
+
+Report tests without executing any of them, neither tests nor hooks.
+
 ### `--exit`
 
 > _Updated in v4.0.0._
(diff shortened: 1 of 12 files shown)


nicojs:
Interesting. What is the use case for this feature?

i.e. are we sure people might not misuse this in pipelines to make all tests pass?

juergba:
There is an old PR #1070 where quite a few users have been asking for this feature.

My main reasoning is: failing hooks are changing the amount of tests being executed. When a hook fails (mainly `before` and `beforeEach`) then tests get skipped and "disappear" without any reporting. This unfortunate hook pattern has been chosen years ago, so it's not a bug. With `dry-run` you can check which tests should actually run.
Furthermore it's useful for checking dynamically created tests.

> i.e. are we sure people might not misuse this in pipelines to make all tests pass?

No, how could we? This is the responsibility of the user, but we can add some kind of control later, if necessary.

0xorial:
Just out of curiosity - to me this sounds like a critical feature to integrate mocha in any IDE (when I open it I usually want to see a list of tests and be able to re-run/debug them individually). But obviously this is not the case as it was implemented only recently. How is/was it supposed to be integrated in the IDEs? (I myself got here because I was looking for a way to use TAP in vscode or WebStorm)

PR #4653, opened 2021-06-07
fix: proper version check in hasStableEsmImplementation

### LINKED ISSUE #4652: Stopped working on Node.js v10.24: "Error: Not supported" caused by ESM import
### Prerequisites

### Description

In `lib/esm-utils.js`, `hasStableEsmImplementation` returns `true` on Node.js v10.24.1 even though it must only return `true` on versions >= 12.22.

### Steps to Reproduce

In the cloned `mocha` repository:

```
$ node -e 'const path = require("path"); const utils = require("./lib/esm-utils"); utils.requireOrImport(path.resolve("./index.js"))'
```

No error

```
(node:22250) UnhandledPromiseRejectionWarning: Error: Not supported
    at formattedImport (/Users/fenster/repos/mocha/lib/esm-utils.js:7:31)
    at Object.exports.requireOrImport.hasStableEsmImplementation [as requireOrImport] (/Users/fenster/repos/mocha/lib/esm-utils.js:46:38)
    at [eval]:1:79
    at Script.runInThisContext (vm.js:122:20)
    at Object.runInThisContext (vm.js:329:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at evalScript (internal/bootstrap/node.js:590:27)
    at startup (internal/bootstrap/node.js:265:9)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
```

100%

### Versions

9.0.0

- Your operating system
  - name and version:

alexander-fenster:
Fixes #4652.

### Requirements

### Description of the Change

The version check in `hasStableEsmImplementation` returns `true` for any version that has minor component >= 22 (e.g. for version 10.24.x which is currently the latest one in the 10.x branch).  Even though Node.js v10 is officially not supported, I still see a lot of sense to fix this and make `hasStableEsmImplementation` work properly.

### Alternate Designs

None

A lot of CI tasks are still using 10.x for compatibility reasons.

### Benefits

No unintended breaking change to Node.js v10.24.x users.

### Possible Drawbacks

None known.

### Applicable issues

#4652

### DIFF
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -34,7 +34,9 @@ const hasStableEsmImplementation = (() => {
   const [major, minor] = process.version.split('.');
   // ESM is stable from v12.22.0 onward
   // https://nodejs.org/api/esm.html#esm_modules_ecmascript_modules
-  return parseInt(major.slice(1), 10) > 12 || parseInt(minor, 10) >= 22;
+  const majorNumber = parseInt(major.slice(1), 10);
+  const minorNumber = parseInt(minor, 10);
+  return majorNumber > 12 || (majorNumber === 12 && minorNumber >= 22);
 })();
 
 exports.requireOrImport = hasStableEsmImplementation


juergba (approved):

giltayar:
I would strongly recommend using Mocha v8 if you're using Node v10.

But, hey, why not accept this PR? 😊 LGTM.

But please note that accepting this PR should not mean that we support Node v10 in Mocha >= v9, as there may be other incompatibilities that we are not aware of.

alexander-fenster:
Thank you folks! Re: "using Mocha v8 if you're using Node v10" - We don't really _use_ Node v10 :)  We just want to keep it in our CI for just a little longer, and don't want to pay for it by staying on the old version. Sure, if a _real_ breaking change comes, we'll have no choice, but here in this case the trivial fix could make it live longer. So - again - thanks :)

PR #4668, opened 2021-06-25
Fixes #4665. make "-r <dir>" work with new import-first loading

### LINKED ISSUE #4665: `ts-node/register` not supported in 9.0.0

giltayar:
### Description of the Change

When falling back on CJS when loading test files, ensure that `ERR_UNSUPPORTED_DIR_IMPORT` is supported so that `-r <cjs-package>/dir` will work. See #4665 for a more detailed description.

### Benefits

This is a bug fix to return support to what was before enabled.

### Possible Drawbacks

None

### Applicable issues

#4665 

### DIFF
--- a/test/integration/fixtures/esm/test-that-uses-dir-cjs-require.fixture.js/index.js
+++ b/test/integration/fixtures/esm/test-that-uses-dir-cjs-require.fixture.js/index.js
@@ -1,0 +1,4 @@
+// See https://github.com/mochajs/mocha/issues/4665 for an explanation of this test
+it('should require a dir import', () => {
+  expect(global.testPassesIfThisVariableIsDefined, 'to be', true)
+})
--- a/lib/esm-utils.js
+++ b/lib/esm-utils.js
@@ -49,7 +49,8 @@ exports.requireOrImport = hasStableEsmImplementation
       } catch (err) {
         if (
           err.code === 'ERR_MODULE_NOT_FOUND' ||
-          err.code === 'ERR_UNKNOWN_FILE_EXTENSION'
+          err.code === 'ERR_UNKNOWN_FILE_EXTENSION' ||
+          err.code === 'ERR_UNSUPPORTED_DIR_IMPORT'
         ) {
           return require(file);
         } else {
--- a/test/integration/esm.spec.js
+++ b/test/integration/esm.spec.js
@@ -1,4 +1,5 @@
 'use strict';
+var path = require('path');
 var helpers = require('./helpers');
 var run = helpers.runMochaJSON;
 var runMochaAsync = helpers.runMochaAsync;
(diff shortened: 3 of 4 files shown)


juergba:
@giltayar could there be more cases `require` can handle, but `import` doesn't? Requiring `json` files?

giltayar:
@juergba I can check, but this is only for test files and `--require` files loaded by mocha, and not for all requires/imports, so I don't think JSON files are relevant, as even if we don't support loading JSON test files, nobody would care.

test/integration/fixtures/esm/test-that-uses-dir-cjs-require.fixture.js/index.js
```diff
@@ -0,0 +1,4 @@
+// See https://github.com/mochajs/mocha/issues/4665 for an explanation of this test
```
juergba, line 1: Why is this fixture a directory? This is the test file, not the module to be required via `index.js`.

giltayar: I wanted to check both the loading of a test file that is a directory that has `index.js`, _and_ a loading using `--require` of a similar directory. The bug is in the `--require`, but I wanted to have a test for both just to ensure that it works whatever the changes we do in how we import the test files.

juergba (approved):

jdaykin454:
https://github.com/mochajs/mocha/pull/4668

jdaykin454:
lib/esm-utils.js

jdaykin454:
a6fd8d7e4df7e62f0daab3627f99eaeea0495573

jdaykin454:
https://github.com/giltayar

jdaykin454:
92518dfd3cf2a6458a5c1e67976f1d1eb3c9d1d7

jdaykin454:
92518dfd3cf2a6458a5c1e67976f1d1eb3c9d1d7

jdaykin454:
92518dfd3cf2a6458a5c1e67976f1d1eb3c9d1d7

test/integration/fixtures/esm/test-that-uses-dir-cjs-require.fixture.js/index.js
jdaykin454: https://github.com/mochajs/mocha/pull/4666

PR #4687, opened 2021-07-10
fixes #4675: ESM swallowing MODULE_NOT_FOUND errors in case of type:module

### LINKED ISSUE #4675: improve ERR_REQUIRE_ESM error reports by including root cause exception report.
**Is your feature request related to a problem or a nice-to-have??**

## /Problem/! Saving time & money when the shit hits fan.

Actual scenario:

node v16, mocha 9.0.2, + custom markdown-it-footnote derivative under test -- which' code uses all `import`s, *zero* `require`s. Worked with mocha 9.0.1 for a while.

Confusing and /wrong/ error report by mocha 9.0.2:

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: Z:\lib\js\markdown-it-footnote\test\test.js
require() of ES modules is not supported.
require() of Z:\lib\js\markdown-it-footnote\test\test.js from Z:\lib\js\markdown-it-footnote\node_modules\mocha\lib\esm-utils.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename test.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from Z:\lib\js\markdown-it-footnote\package.json.

    at new NodeError (node:internal/errors:363:5)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1112:13)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:816:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:93:18)
    at Object.exports.requireOrImport (Z:\lib\js\markdown-it-footnote\node_modules\mocha\lib\esm-utils.js:56:18)
    at async Object.exports.loadFilesAsync (Z:\lib\js\markdown-it-footnote\node_modules\mocha\lib\esm-utils.js:75:20)
    at async singleRun (Z:\lib\js\markdown-it-footnote\node_modules\mocha\lib\cli\run-helpers.js:125:3)
    at async Object.exports.handler (Z:\lib\js\markdown-it-footnote\node_modules\mocha\lib\cli\run.js:366:5)
```

After some serious WTF?!?! we patched `mocha\lib\esm-utils.js` to find out and added one line in there immediately after the `catch` before the error location @ line 50:

```
console.error(err)
```

which gives us this **very helpful** information before mocha barfs a hairball as before:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'Z:\lib\js\markdown-it-footnote\dist\markdownitFootnote.js' imported from Z:\lib\js\markdown-it-footnote\test\t
est.js
    at new NodeError (node:internal/errors:363:5)
    at finalizeResolution (node:internal/modules/esm/resolve:321:11)
    at moduleResolve (node:internal/modules/esm/resolve:756:10)
    at Loader.defaultResolve [as _resolve] (node:internal/modules/esm/resolve:867:11)
    at Loader.resolve (node:internal/modules/esm/loader:89:40)
    at Loader.getModuleJob (node:internal/modules/esm/loader:242:28)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:73:40)
    at link (node:internal/modules/esm/module_job:72:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

This **immediately** gives us something to check: turns out there's a typo in the package.json. Problem fixed.

## My suggestion

When dumping errors (and hint messages like you do), **include the root cause exception message** because this mocha failure *is* due to some code migration happening (the markdown-it plugin at hand) but is *far removed* from what you point at in the "helpful hint" about cjs/mjs/etc., which actually **did cost us more time, rather than saving time**.

Simply keep the original exception handy when the second attempt happens to fail and include it in the mocha error report so the end user can do their diagnostic work. In this particular example, having seen the initial exception error message, I can wholly ignore the follow-up error + hint as it'll be clear to me where to *start looking* then!

## Post Mortem

I do realize the entire import vs require + npm situation is a horror show (bad design if you ask me, but this is *way* out of scope and we gotta play the cards we're dealt, sometimes): see related issues like https://github.com/mochajs/mocha/issues/4652#issuecomment-859999472, which *amazingly* kicked us in the teeth at the CI but nowhere on the local dev machines (Windows based). That kind of context being given, and error diagnostics being a hard target generally speaking, this didn't help exactly as the mocha error report was *quite misleading* in this case.

giltayar:
### Description of the Change

The move to "import-first" in Mocha generates a subtle problem in case of test files that are `.js`, yet are ESM due to the package.json include a `"type": "module"` property: if the test file includes a bad import, then Node.js throws an `ERR_MODULE_NOT_FOUND`, which we catch and try to `require` the module. Because the extension is `.js` and we have `"type": "module"`, Node.js tries to be helpful and throws an `ERR_REQUIRE_ESM`, which is what we throw back to the user.

As pointed out by @GerHobbelt, this is _totally_ confusing, as the original error with the correct information to resolve the problem is thrown away, and replaced by an error that has no bearing on the original problem.

What we do in the PR is simple: we figure out that this is the situation, and if it is, we throw the original error (`ERR_MODULE_NOT_FOUND`) instead of the new one (`ERR_REQUIRE_ESM`).

### Alternate Designs

We thought about throwing a new error with a message that includes both error call stacks. But that's ugly _and_ will also confuse the user, because they need to understand which of both errors is really relevant, and will be confused by an error who's message includes call stacks.

### Benefits

Less confusion!

### Possible Drawbacks

We're still "swallowing" errors in other situations. The alternate design would have taken care of them. We still chose the above design because it is better for the user, and currently appears to be the only problematic case.

### Applicable issues

#4675, and than you to @GerHobbelt for the great bug explanation and repo that reproduces the problem. They really helped!


GerHobbelt:
👍

test/integration/esm.spec.js
```diff
@@ -81,4 +84,22 @@ describe('esm', function() {
...
   });
+
+  it('should throw an ERR_MODULE_NOT_FOUND and not ERR_REQUIRE_ESM if file imports a non-existing module', async function() {
+    const fixture =
+      'esm/type-module/test-that-imports-non-existing-module.fixture.js';
+
+    const [, promise] = invokeMochaAsync(
+      [resolveFixturePath(fixture), '--unhandled-rejections=warn'],
+      {stdio: 'pipe'}
+    );
```
juergba, line 95: The test style you have chosen above (`should show file location when there is a syntax error in the test`) is much nicer than this one. I don't get any unhandled rejections neither.
IMO the tests are easier to read if you chose the same style for similar error test cases.
Can you change that, please?

test/integration/fixtures/esm/type-module/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
+}
```
juergba, line 3: add one line, please.

giltayar: Fixed.

test/integration/esm.spec.js
giltayar: Good idea! Fixed.

PR #4691, opened 2021-07-19
Add new option "node-option"

juergba:
### Description

Currently Mocha is parsing its own options, plus Node and V8 options as well. Mocha has to recognize those foreign options in order to pass them to a newly spawned child-process.

Mocha's options are defined in _yargs/yargs-parser_, whereas Node/V8 options:
- are not white-listed
- Node: compared with `process.allowedNodeEnvironmentFlags`, which contains most but not all flags.
- V8: compared against a beautiful regex, otherwise `v8[_-]` has to be prepended.
 
This Mocha/Node hybrid parsing has some disadvantages:
- maintenance overhead
- some options can't be forwarded at all, eg. `--prof`, `--require`
- Mocha's option object gets polluted with invalid camelCased Node/V8 options
- Mocha can't print a warning in case of incorrect user input (typos)
- we have bugs in parsing of our own kebab-case options, which are difficult to fix with the current _yargs-parser_ configuration

### Description of the Change

Mocha should not have to check/process any Node/V8 flags, but simply parse and hand them over to Node.
So we add a new flag `--node-option`/ `-n`:
- array:
  - `-n require=glob -n unhandled-rejections=strict` or 
  - `-n require=glob,unhandled-rejections=strict`
- no checks or processing, Mocha just hands the option over to Node by spawning a new child-process
- no merging: the two solutions are independent
- optional: remove current Mocha/Node hybrid solution on mid-term

in addition:
- prevent double parsing when _bin/mocha_ is the entry point
- remove package `esm` special handling. This package is abandoned and not supported in Mocha v9
- remove `gc` check and deprecation warning. This V8 option doesn't exist anymore


docs/index.md
```diff
@@ -1015,6 +1013,15 @@ Specify an explicit path to a [configuration file](#configuring-mocha-nodejs).
 
 By default, Mocha will search for a config file if `--config` is not specified; use `--no-config` to suppress this behavior.
 
+### `--node-option <name>, -n <name>`
+
+> _New in v9.1.0._
+
+For Node.js and V8 options. Mocha forwards these options to Node.js by spawning a new child-process.<br>
```
outsideris, line 1020: To clear, can we specify name should be without `--` nor `--v8-`?

lib/cli/run.js
```diff
@@ -176,6 +176,10 @@ exports.builder = yargs =>
         group: GROUPS.OUTPUT,
         hidden: true
       },
+      'node-option': {
+        description: 'Node or V8 option (no leading "--")',
```
outsideris, line 180: How about `no leading "--" or "--v8-"`

outsideris (approved):
👍🏼

docs/index.md
juergba: I added the `--` info. I prefer to not mention `--v8-` as this is a Mocha specific quirk and may be confusing.

lib/cli/run.js
juergba: I haven't added the `--v8-` part, see comment above.

PR #4714, opened 2021-08-11
Docs: `grep()` using RegExp with global flag `/g`

### LINKED ISSUE #4704: Grep function randomly skips tests
### Description

Probably this issue is my specific, but this behavior is strange  related to grep by regex. I am using mocha programmatically.
when I am using "grep" function.

The the problem is in this method:
`var match = self._grep.test(test.fullTitle());`
(node_modules\mocha\lib\runner.js )

This method outputs false, while it should be true, but only for some 'it' (tests). My describe has 2 it, one is passed, the other not visible because of condition 'false'.

Example:
this one 'it' visible:
'this is some description @testing, @testing-smoke Ensure dashFeeSensitivity exist'
while this 'it' not:
'this is some description @testing, @testing-smoke Ensure dashMaxFloor exist'

bot of them should return 'true', but within mocha, the second example returns 'false':
/@testing |@testing$|@testing-smoke |@testing-smoke$/g.test('this is some description @testing @testing-smoke Ensure dashFeeSensitivity exist')
/@testing |@testing$|@testing-smoke |@testing-smoke$/g.test('this is some description @testing @testing-smoke Ensure dashMaxFloor exist')

for testing purpose (I know my regex),  I changed ` self._grep` on `/@testing |@testing$|@testing-smoke |@testing-smoke$/g` and it works

### Versions

mocha - 8.4.0
node - 14.16.0, 10.15.3
OS - windows 10, 64
@babel/core - 7.8.4

juergba:
### Description

The global flag `/g` in a RegExp expression is behaving weirdly with Mocha's `--grep` option. The selection of tests based on the fullTitle seems wrong as tests get lost, but at the end it's just javascript.
```js
let reIgnore = /a/g;

console.log(reIgnore.test("a")); // true
console.log(reIgnore.test("a")); // false
console.log(reIgnore.test("a")); // true
console.log(reIgnore.test("a")); // false
```

### Description of the Change

Using a flag as `/g` or `/y` doesn't make any sense in this context. 
- improve `Mocha.prototype.grep` for better conversion of `string` into `RegExp`
- clarify documentation

### Alternate Designs

Instead of re-using the same RegExp instance for checking each test, we could construct new instances:
```js
let reIgnore = /a/g;

console.log(new RegExp(reIgnore).test("a")); // true
console.log(new RegExp(reIgnore).test("a")); // true
console.log(new RegExp(reIgnore).test("a")); // true
console.log(new RegExp(reIgnore).test("a")); // true
```
Mocha's `grep` selection is running in any case, with the `--grep` value or with our defaultGrep `/.*/`. Therefore I decided to not apply any fix for this edge case.

### Applicable issues

closes #4704

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1138,9 +1138,12 @@ Cause Mocha to only run tests matching the given `regexp`, which is internally c
 
 Suppose, for example, you have "api" related tests, as well as "app" related tests, as shown in the following snippet; One could use `--grep api` or `--grep app` to run one or the other. The same goes for any other part of a suite or test-case title, `--grep users` would be valid as well, or even `--grep GET`.
 
+And another option with double quotes: `--grep "groupA|groupB"`.<br>
+And for more complex criterias: `--grep "/get/i"`. Some shells as e.g. Git-Bash-for-Windows may require: `--grep "'/get/i'"`. Using flags other than the `ignoreCase /i` (especially `/g` and `/y`) may lead to unpredictable results.
+
 ```js
 describe('api', function() {
-  describe('GET /api/users', function() {
+  describe('GET /api/users groupA', function() {
     it('respond with an array of users', function() {
       // ...
     });
@@ -1148,7 +1151,7 @@ describe('api', function() {
 });
 
 describe('app', function() {
-  describe('GET /users', function() {
+  describe('GET /users groupB', function() {
     it('respond with an array of users', function() {
       // ...
     });
--- a/example/config/.mocharc.js
+++ b/example/config/.mocharc.js
@@ -1,7 +1,8 @@
 'use strict';
 
 // This is a JavaScript-based config file containing every Mocha option plus others.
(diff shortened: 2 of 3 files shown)


docs/index.md
```diff
@@ -1138,17 +1138,20 @@ Cause Mocha to only run tests matching the given `regexp`, which is internally c
 
 Suppose, for example, you have "api" related tests, as well as "app" related tests, as shown in the following snippet; One could use `--grep api` or `--grep app` to run one or the other. The same goes for any other part of a suite or test-case title, `--grep users` would be valid as well, or even `--grep GET`.
 
+And another option with double quotes: `--grep "groupA|groupB"`.<br>
```
outsideris, line 1141: `<br>` is not needed here.

outsideris (approved):

docs/index.md
juergba: I tried, there is no `end-of-line` without `<br>` in our markdown file.

Issue #4722, opened 2021-08-18
Retry timeouts only 


gita-codaio:
I would like to propose a flag that when used with retries it would only retry on timed out tests. This would help with retrying only flakey tests.

juergba:
@gita-codaio thank you for your proposition.

I don't think we need such a granular configuration on CLI/config file  level. On flaky tests you can set `retries`:
- `it('flaky test', async function() { this.retries(2) });`
- `it('flaky test', async function() {  })retries(2);`

PR #4723, opened 2021-08-18
add retrytimedout arg

gita-codaio:
### Description of the Change

Added a CLI parameter ( `--retry-timed-out` ) that when used with `--retries` would only retry tests if they had timed out
https://github.com/mochajs/mocha/issues/4722

### Alternate Designs

I was thinking of adding `this.retryTimedOut` as a suite and test level parameter as well.

This is a simple change that would help retry for flakey tests only. 

### Benefits

Help keep retries to a minimum 

### Possible Drawbacks

Not useful for everyone ¯\_(ツ)_/¯

### Applicable issues

Minor Release

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -809,6 +809,24 @@ Mocha.prototype.timeout = function(msecs) {
 };
 
 /**
+ * @summary
+ * Enable retrying only timed out tests
+ *
+ * @description
+ * A boolean to determine whether
+ *
+ * @public
+ * @see [CLI option](../#-retry-timeouts)
+ * @see [Timeouts](../#timeouts)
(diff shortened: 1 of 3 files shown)


juergba:
@gita-codaio thank you for this PR.

> @juergba sorry, do you mind looking at or giving me access to the deploy page please

I don't have access to the deploy-preview either. It's just the deploy of our docs site. I guess it fails, because you added a link which doesn't exist: `@see [CLI option](../#-retry-timeouts)`.

I'm closing this PR, see my comment in your issue.
Some feedback: most parts of adding a new option are missing, there have been PRs recently doing just that. Also documentation, tests and the licence signature are missing.

PR #4736, opened 2021-08-29
ESM: remove code for Node versions <10

juergba:
### Description

- remove ESM code which adressed Node versions <10
- move _lib/esm-utils.js_ to _lib/nodejs/esm-utils.js_


nicojs (approved):
So I guess this means a new major version?

juergba:
@nicojs  no, semver-patch.

We already have dropped Node v8 support with Mocha v8 more than 14 months ago.
Currently Mocha supports Node >=v12, and inofficially it still runs with Node v10.
So this PR just removes old code debris.

_lib/esm-utils.js_ and `lib/utils/supportsEsModules()` are not part of a public API.

outsideris (approved):

PR #4746, opened 2021-09-12
Stop using all global vars in browser entry

### LINKED ISSUE #4740: Window properties impeded using mocha
I recently tried to use mocha in a project whose runtime environment had a `window.ui` property and we had to patch our copy of mocha.js to make it work. The window property overshadows the mocha prototype property and mocha wouldn't even initialize properly. I suspect there are other critical properties that, if they appear as members of window, will cause similar problems.

I'm not sure if this is something the project wants to address directly, or if it's a "Doctor, it hurts when I do that - Don't do that" but I figured I'd report it anyway.

### Prerequisites

Actually, the last two are kind of N/A for this issue.

### Description

In this piece of code at the end of mocha.js, windows properties overwrite mocha properties, even those that are necessary for mocha to function.
```javascript
	var browserEntry = Object.assign(mocha, commonjsGlobal);
	return browserEntry;
```
In our case, the ui function was getting overwritten by a ui property introduced by the Foundry execution framework in which we were running our tests.

Once we adjusted this code to do the following, we were able to run mocha tests in our environment in the usual way:

```javascript
	// Wicked hack for Quench - 
	// get ui property out of copy of global merged with mocha before merging so it doesn't clobber mocha's ui function

	var hackedCommonJsGlobal = Object.assign({}, commonjsGlobal);
	delete hackedCommonJsGlobal.ui;
	var browserEntry = Object.assign(mocha, hackedCommonJsGlobal);
	return browserEntry;
```

It occurs to me that there may be other mocha properties that shouldn't be overshadowed by properties from window as well, and that mocha could protect itself from.

### Steps to Reproduce

Not sure how applicable this is.  Set up a globalThis.ui property before you import mocha and you should see the effect.

**Expected behavior:** Property doesn't affect mocha's ability to initialize before running tests.

**Actual behavior:** Property overshadows mocha's ui function and mocha is unable to initialize from options, even `setup('bdd')`, before running any tests. 

**Reproduces how often:** Quite consistently - if your globalThis has properties that happen to overshadow mocha properties.

### Versions

We were working with mocha 8.x and 9.x. The version of node doesn't specifically matter. 

### Additional Information

We aren't running mocha standalone from a command line in NOde, and we aren't spawning a process to run it and piping back the results, but we're running it as a library embedded in another app running in the browser. The tests need to use the APIs of the browser app in which the runner is executing.

This issue was found in the context of a module that lets us run tests within the FoundryVTT environment. This module adds the UI to the environment to kick off the tests and display the results, but it does it along with loading the core UI, the world, the system, etc.  so quite a lot of stuff for the environment has loaded, including the window.ui property, well before the imports of this specific module are invoked (including the mocha and chai imports). It is intended to be used in support of building other modules.

With this one tweak, the tests seem to run fine.

### LINKED ISSUE #4745: Stop iterating over window props
### Prerequisites

### Description

https://github.com/mochajs/mocha/blob/4860738af9de9493fade35aea3df65dc7461e100/browser-entry.js#L219

is causing chrome to show a `'window.webkitStorageInfo' is deprecated.` warning.

~To my understanding, iterating over the window properties is only for the `check-leaks` feature. Then with this feature disabled, there is no need to iterate that.~

There is no need to access all the global variables.

Thank you.

### Steps to Reproduce

Run any mocha test in a recent Chrome.

**Expected behavior:** No access to every global variables.

**Actual behavior:** At least one access to each global variable.

**Reproduces how often:** Every time you run mocha.

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 9.1.0
- The output of `node --version`: 14.17.0
- Your browser and version (if running browser tests): Chrome Dev 94.0.4595.0 with Playwright

PaperStrike:
### Description of the Change

https://github.com/mochajs/mocha/blob/4860738af9de9493fade35aea3df65dc7461e100/browser-entry.js#L219

is causing chrome to show a `'window.webkitStorageInfo' is deprecated` warning.
Using these properties may affect mocha's own functionalities, too. (See #4740)

This PR makes mocha export without using these globals.

This bug exists in mocha.

### Benefits

No `'window.webkitStorageInfo' is deprecated` warning when initializing.
No risk of breaking mocha's functionalities.

### Applicable issues

Fixes #4745.
Fixes #4740.

Is this a breaking change (major release)? **no**
Is it an enhancement (minor release)? **no**
Is it a bug fix, or does it not impact production code (patch release)? **yes**


juergba:
@PaperStrike thanks. I had a look, but havent understood everything, yet. It seems to be quite messy, and I don't want to turn it even worse.

In our Node code (for npm), these functions are static on the `Mocha` class, and therefore can be acessed by the `mocha` instance as well. In our browser code, these functions/variables are set on `global.Mocha`. Why do we need to copy them to `global.mocha` as well?

Not all these variables have to be set, only a subset depending on which interface (eg. `bdd`) has been chosen.

PaperStrike:
> Why do we need to copy them to `global.mocha` as well?

I have no idea. It's not written by me. By the comment, it's for import like in node. However, I can't get this running in real-browsers (bundled with [`playwright-test`](https://github.com/hugomrdias/playwright-test)), i.e., am not getting `global.describe` defined (let alone `mocha.describe`) before a explicit `mocha.setup`. I searched a long time for where mocha sets these to global before `mocha.setup` in test cases, tending to make them happen in the mocha instance at the same time, but no luck. Some magic is making them passing the test cases.

That's all I know.

If this is actually broken, delete / skip the required token test case for browser, export mocha without reading these globals, then everyone would be happy.

Or if I understand the situation correctly, we need a brave warrior to do a deep refactor, to better fix this.

juergba:
> I searched a long time for where mocha sets these to global [...]

It's in `mocha.ui()` => `this.suite.emit('pre-require', global, null, this);`. Note `global` is the context.
An event listener is registered in _lib/interfaces/bdd.js_.

PaperStrike:
@juergba Thanks. I know this and the problem is acturally when mocha invokes the `mocha.ui()`. In the test cases, global `describe`, `it` is always there, set by some point I don't know. But in the browser I tested, `this.suite.emit('pre-require', global, null, this);` is always invoked latter than browser entry's `export`. So the last line of `browser-entry.js` seems to be there only to make some test cases happy.

juergba:
@PaperStrike There is this special test _test\unit\required-tokens.spec.js_ already mentioned in _browser-entry.js_. Without this test, some other browser tests will fail running on `master` branch. It's weird, that this test has an influence on other browser tests.

@mantoni can I ask you a question, please?
Do you know why we need this `module.exports = Object.assign(mocha, global); ` Is it for making some useless Mocha tests happy (maybe some debris from the past)? Or something related with browserify? Or for people using Mocha's bundle for bundling their own file? Or whatever?
Why would we need: `thus, you can now do "const describe = require('mocha').describe" in a browser context (assuming browserification).` ?

mantoni:
@juergba Mocha cannot be browserified anymore since v8. I don't know why this exists, but if it explicitly mentions browserify, I guess it's legacy.

PaperStrike:
I made a aggresive commit that removes the `required-tokens` test case and put `browser-entry.js` into the setup (`test/browser-specific/setup.js`, not sure if it is the right place) for other tests that requires it.

---

And I came across a way that may fix the problem without editing tests by changing the last few lines of `browser-entry.js` to

```js

Object.keys(Mocha)
  .filter(function(key) {
    return !(key in mocha);
  })
  .forEach(function(key) {
    mocha[key] = Mocha.bind(mocha);
  });

module.exports = mocha;
```

But a weird thing happens when you use the above code, the number of test cases for `karma_no-ie11` (possibly with many others) drops from `passed: 504, ignored: 7` to `passed: 215, ignored: 2`. Emm... why? 🤔

juergba:
@mantoni thank you

PaperStrike:
It will be better to have `describe`, `it` available on `mocha` in browsers, anyway. We all agree it is broken currently as we copy from the global before a `mocha.setup`.

And we can (just tested) achieve this by copying the values from `Mocha` in the last few lines of `browser-entry.js` like:

```js
Object.keys(Mocha)
  .filter(function(key) {
    return !(key in mocha);
  })
  .forEach(function(key) {
    mocha[key] = Mocha[key];
  });

module.exports = mocha;
```

The problem is the test cases. It's hard to pass them. I tried to put a `require('../..').setup('bdd')` in `test/browser-specific/setup.js`, and found I would need to put a `.run` somewhere, too. And I don't know where to put it. (Unless set up as `'tdd'`, it passes the tests, but not gracefully.)

juergba:
@PaperStrike thank you so far, you are a big help on this issue. I feel a little better, but still this bundling story is so fuzzy.

Why do we need `describe` and `it` in `mocha` at all:
- for users?: `We all agree it is broken currently as we copy from the global before a mocha.setup`.
So you are saying for the user these functions are currently not available on `mocha`.
And nobody is complaining. Right?
- for browserify?: no => `Mocha cannot be browserified anymore since v8.`
- for our own tests?: seems to be the only reason.

So at first sight I like the idea to removing the `required-tokens` requirement:
- I haven't understood the role of `required-tokens` test case. Why does it work with, but fails without it?
- `In our Node code (for npm), these functions are static on the Mocha class, and therefore can be acessed by the mocha instance as well.` I was wrong, in Node these functions are not available on `mocha` instance either.
- `The problem is the test cases (tests in this repo, I mean).`: why? all CI tests have been passing ...
- `it will be better to have describe, it available on mocha in browsers`: why? The users currently don't have it on `mocha`.

> I made a aggresive commit that removes the required-tokens test case and put browser-entry.js into the setup (test/browser-specific/setup.js, not sure if it is the right place) for other tests that requires it.

I haven't understood, why _browser-entry.js_ in the setup makes the test pass.

Is your last proposition as an addition to the current PR? Or standalone?

PaperStrike:
@juergba 

> * I haven't understood the role of `required-tokens` test case. Why does it work with, but fails without it?

To find the cause, I shortened `required-tokens.spec.js` as tiny as possible, and found the necessary part, `require('../..')`. `../..` links to the `index.js` in the root dir, and by the `browser` field in `package.json`, it links to `browser-entry.js` when bundling. So the solution is require `browser-entry.js` somewhere. But wait, what's the root cause, and what if we only need a little piece of `browser-entry.js`? Combined with the error messages (something related to properties of `process`), and the content of `browser-entry.js`, it seems better to require `browser-entry.js` as a whole.

> * `In our Node code (for npm), these functions are static on the Mocha class, and therefore can be acessed by the mocha instance as well.` I was wrong, in Node these functions are not available on `mocha` instance either.

Yeah. Earlier this day I found mocha in Node exports `Mocha`, and mocha in browser exports its instance. Quite weird I would say.

> * `The problem is the test cases (tests in this repo, I mean).`: why? all CI tests have been passing ...

"The problem" means we _will_ have tests failed, if we use the code right above that sentence, without doing other stuff like removing `required-tokens.spec.js` and editing `test/browser-specific/setup.js`. The later sentenses in that comment start with this context, too.

Sorry I didn't made it clear.

> * `it will be better to have describe, it available on mocha in browsers`: why? The users currently don't have it on mocha.

Modules are the future. I like writting tests like:

```js
import { describe, it } from 'mocha';
describe(...);
```

which is not possible in browsers. And I wish one day mocha exports functions to global only when the user requires. It's not possible for now, and the first step to modularize would be the users being able to write like that.

> Or if I understand the situation correctly, we need a brave warrior to do a deep refactor, to better fix this.

The "deep refactor" may be to export `Mocha` in both Node and browsers. Is anyone interested?

juergba:
Ok, we are quite close now.
I prefer the solution of this PR as is, so without `describe`/`it` on `mocha`.

> import { describe, it } from 'mocha';  [...] which is not possible in browsers.

Why do you prefer having it on `mocha` then? If you can't use it anyway. Or can you use for your own bundling?

PaperStrike:
> Why do you prefer having it on mocha then? If you can't use it anyway. Or can you use it for your own bundling?

I can use it, if the code in https://github.com/mochajs/mocha/pull/4746#issuecomment-920687732 were applied. If it's applied I can write tests like:

```html
<script src="link/to/mocha/bundle"></script>
<script>
mocha.setup('bdd');
</script>
<!-- Can also link to external script files -->
<script type="module" id="my-real-tests">
import { describe, it } from 'link/to/mocha/bundle';

describe('group', () => {
  it('pass', () => {});
});
</script>
<script>
mocha.run();
</script>
```

PaperStrike:
Anyway, it seems a bit off topic now. If you prefer to delete the required-tokens test case and agree to put a `require` in `/test/browser-specific/setup.js`, merge this PR then.

juergba:
Ok then, add https://github.com/mochajs/mocha/pull/4746#issuecomment-920687732 to this PR.
Please also confirm wether #4740 will be fixed, and add it to your description.
I will do some testing tomorrow with your updated branch.

juergba:
The `required-tokens` test case will not pass, even if you add your proposition.

PaperStrike:
> The required-tokens test case will not pass, even if you add your proposition.

Yeah, that's why I mentioned "The problem is the test cases", "tried to put a `require('../..').setup('bdd')` in `test/browser-specific/setup.js`", found 0 test is running for "need to put `.run()` somewhere, too", and noticed "setting up as `setup('tdd')` magically passes the tests" while "we're using bdd".

I'm in UTC+8 and sorry I am out for some hours.

PaperStrike:
@juergba After running some own tests in real browsers without bundling beforehead, now I agreed to keep this PR as is.. Been using these TypeScript, bundlers for sometime, damn me forget the browsers don't support such syntax at all. Sorry for wasting your time.

PaperStrike:
PR description updated.

juergba:
Regarding removed test case _required-tokens.spec.js_: we still have integration tests covering this `require` interface on Node.

juergba (approved):
@PaperStrike thank you for this PR!

Issue #4748, opened 2021-09-14
Fix SNYK-JS-ANSIREGEX-1583908 and move to ESM


rafaelmaeuer:
**Is your feature request related to a problem or a nice-to-have?? Please describe.**

As of https://snyk.io/vuln/SNYK-JS-ANSIREGEX-1583908 my investigations lead to **mocha** referencing the vulnerability by peer dependencies of `"wide-align": "v1.1.3"` -> `"string-width": "^1.0.2 || 2"` -> `"strip-ansi": "^4.0.0"` -> `"ansi-regex": "^3.0.0"`.

**Describe the solution you'd like**

Forcing @iarna to accept https://github.com/iarna/wide-align/pull/57 from @jimmywarting will update `"string-width": "^5.0.1"` resulting in a fix of https://snyk.io/vuln/SNYK-JS-ANSIREGEX-1583908 with peer-dependencies `"strip-ansi": "^7.0.1"` -> `"ansi-regex": "^6.0.1"`. It will require some changes towards ESM on mocha-side too: https://github.com/mochajs/mocha/blob/27bfc7425fed7a9a9cadb6dabc536a0bd32abd7a/lib/cli/one-and-dones.js#L11

**Describe alternatives you've considered**

Alternatively the use of @jimmywarting branch https://github.com/jimmywarting/wide-align/tree/esm as dependency, or investigation of a replacement for `string-width` are thinkable.

juergba:
We get issues like this one on regular basis.
- please add a PoC explaining how this ReDoS is supposed to affect Mocha. Is there a masked tester attacking himself? How will he/she accomplish this attack? Which server/services are endangered?
- `[...] move to ESM`: no, we won't move to ESM for now. Certainly not based on a security message like this.

jimmywarting:
I have used a simpler variant in one other project of mine (not that it dose not take into consideration the string width)
```js
export function alignLeft(str, width) {
    return String(str).slice(0, width).padEnd(width)
}
export function alignRight(str, width) {
    return String(str).slice(0, width).padStart(width)
}
export function alignCenter(str, width) {
    const internalString = String(str)
    const leftPadding = Math.floor((width - internalString.length) / 2) + internalString.length
    return internalString.padStart(leftPadding).padEnd(width).slice(0, width)
}
```

jimmywarting:
honestly, now when i looked at what you used `wide-align` for in the one-and-done (which is the only place) 
where it only use align left and dumps something out in the console.log
[...]

PR #4754, opened 2021-09-23
Remove dependency 'wide-align'

juergba:
### Description

see https://github.com/mochajs/mocha/issues/4748#issuecomment-919461985 @jimmywarting

- remove dependency _wide-align_
- use `String.prototype.padEnd()` instead
- rebuild `package-lock.json`


jimmywarting (approved):
Lgtm :+1:

PR #4757, opened 2021-09-27
deps: replace node-canvas with image-size

outsideris:
### Description of the Change
We're using `node-canvas` only for getting dimensions from supporters' images.
`node-canvas` is too heavy for our purpose and it requires node-gyp build. 
So, I replace it with [image-size](https://www.npmjs.com/package/image-size). 

In my M1 MacBook, `canvas` is 8.6M and `image-size` is 192K.

Anyway, assetgraph uses node-canvas in their dependencies. So, this PR doesn't remove node-canvas from mocha at all.


juergba:
> Anyway, assetgraph uses node-canvas in their dependencies. So, this PR doesn't remove node-canvas from mocha at all.

So we have now one additional devDependency and are downloading more data than before, right? Does this make sense? We are not saving anything.

Something is weird with `package-lock.json`. Are you using npm v7? We are still on npm6 which is the latest LTS version.

outsideris:
> So we have now one additional devDependency and are downloading more data than before, right? Does this make sense? We are not saving anything.

We can't manage asssetgraph's deps. That's why we specify `canvas` in our `package.json`. 
I think we should manage our own dependency differently from theirs.

We can't keep to match versions of both our `canvas` and assetgraph's `canvas'. If one update `canvas` to new version, we will download two versions of `canvas`.  Doesn't make sense? 

> Something is weird with package-lock.json. 

What is weird? I'm using npm v7. I can update the `package-lock.json` with npm v6, but we should support npm v7 because the new LTS of Node will release in a month.

juergba:
We already had depended on `image-size` in the past which was swapped out in favour of `canvas`, see #4357.
I don't see any reason to revert this switch. Currently there is only one version of `canvas`, npm removes the double copy.

IMO this PR has no benefit. We could remove `canvas` from our direct devDependency and completely rely on `assetgraph-sprite`. There is no risk since we only use `canvas` for generating sprite images which we accomplish with mentioned `assetgraph-sprite`.

outsideris:
Agreed. I forgot #4357 .

juergba:
You could migrate our `package-lock.json` to nmp v7.
If yes, please also adapt our docs for contribution, publishing, ... else?
If we go now with nmp v7, every contributor should use it without switching back and forth v6 <-> v7.

outsideris:
I don't know compatibility issues between npm v6 and v7 in detail. And there is npm v8 now. 
I will check it.

PR #4769, opened 2021-10-14
Browser: improve 'bdd' import for bundlers

### LINKED ISSUE #4763: Unable to import mocha API functions from 'mocha'
### Prerequisites

### Description

It is no longer possible to do ES6 style imports from mocha. e.g.

```js
import { beforeEach, afterEach } from 'mocha';
```

```js
import { describe, it } from 'mocha';
```

### Steps to Reproduce

Write a test like
```js
import { describe, it } from 'mocha';

describe('A test', function () {
  it('should thing', function () {
  });
});
```

**Expected behavior:**

This test runs and passes

**Actual behavior:** 

Something like

```
TypeError: (0 , mocha__WEBPACK_IMPORTED_MODULE_2__.describe) is not a function
```

(In this case webpack is the bundler)

**Reproduces how often:**

100%

### Versions

I believe this was introduced by this PR: https://github.com/mochajs/mocha/pull/4746 and released as version 9.1.2

### Additional info

I agree with the general goal in the PR: to stop polluting the `mocha` module with the contents of `global`; but it throws the baby out with the bathwater because this was previously how you could write self-documenting import statements which keeps linters happy. `eslint:recommended` contains `no-undef` so if you are writing your tests in modules and bundling them together, you are now in a pickle.

juergba:
### Description

History: #4746

Above PR proved to be too restrictive as it's no longer possible to do ES6 style `import { describe, it } from 'mocha'` while using bundlers.

### Description of the Change

_browser-entry.js_: push all `bdd` functions from `global` to `mocha`.

We have conflicts between `global.setup` (`tdd` version of `beforeEach` hook) and `mocha.setup`(browser setup), and therefore we don't copy the `tdd` functions to `mocha`. The `tdd` interface has never/won't be available for ES6 style import. 

At the time of copying the `bdd` functions are undefined, nevertheless bundlers seem to rely on those lines. The exact logic remained unclear to me.

### Applicable issues

closes #4763

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -209,8 +209,18 @@ Mocha.process = process;
 /**
  * Expose mocha.
  */
-
 global.Mocha = Mocha;
 global.mocha = mocha;
+
+// for bundlers: enable `import {describe, it} from 'mocha'`
+// `bdd` interface only
+// prettier-ignore
+[ 
+  'describe', 'context', 'it', 'specify',
+  'xdescribe', 'xcontext', 'xit', 'xspecify',
+  'before', 'beforeEach', 'afterEach', 'after'
+].forEach(function(key) {
+  mocha[key] = global[key];
+});
 
 module.exports = mocha;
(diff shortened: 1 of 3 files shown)


juergba:
> Personally I'm looking forward to the ESM mocha.

Yes, but I'm unsure about the implications, yet.
Will we need globals anymore? I guess no, because a module has its own scope, but Mocha uses globals ever since.
There still remains the conflict on `setup()`, so evtl. we will need `import { describe, it } from 'mocha.bdd'`.

PR #4793, opened 2021-11-30
doc: fix to show sponsors in narrow view

outsideris:
In the narrow width screen, our sponsor logos disapear except the wallaby logo. 

스크린샷 2021-12-01 오전 1 09 10

Now, all sponsor logos are displayed.

스크린샷 2021-12-01 오전 1 10 52

Under 320px width, logo size is not good, but it's fine over 320px width.

### DIFF
--- a/docs/_includes/default.liquid
+++ b/docs/_includes/default.liquid
@@ -25,6 +25,15 @@
 
     <main id="content">{{ content }}</main>
 
+    <div class="sponsorship">
+      <a href="https://matomo.org/" rel="external noopener" title="Mocha is sponsored by Matomo">
+        <img src="images/matomo-logo.png?trim" loading="lazy" width="100" alt="Matomo logo">
+      </a>
+      <a href="https://wallabyjs.com" rel="external noopener" title="Mocha is sponsored by Wallaby" id="wallaby-logo">
+        <img src="https://wallabyjs.com/assets/img/logoWithText.svg" loading="lazy" width="140" alt="Wallaby logo">
+      </a>
+    </div>
+
     <footer>
       <div id="copyright-notice">
         <a rel="home" href="https://mochajs.org/">mochajs.org</a> is licensed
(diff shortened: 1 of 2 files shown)


juergba:
For my taste these logos are way too large. And the one of OpenJS Foundation belongs into the gray part, where the netlify logo shows up.

outsideris:
Could you show a screenshot?

juergba:
I mean the current docs, even before this PR. Those logos are larger than our own Mocha logo.

outsideris:
I moved all sponsorship logos into the footer because I agreed they are too big. 

스크린샷 2021-12-31 오후 7 49 15

under 400px width:

스크린샷 2021-12-31 오후 7 49 26

juergba:
OpenJS Foundation is not a sponsor, but one of Mocha's owners. That's why I proposed to move its logo to the footer where OpenJS Foundation is mentioned several times.

Imo the remaining true sponsors don't belong into the footer. I don't really like it, but it's upto you.
The size of the images seems ok to me.

outsideris:
Yes, I also don't like sponsors images bottom. It is a design issue. I will try another approach.

outsideris:
How about we remove matomo and wallaby logos?

juergba:
There are sponsor logos missing, at least the one of Saucelabs. There is a pending issue about that.
So removing the two logos mentioned doesn't solve anything.

outsideris:
I fixed them.

wide width: 

스크린샷 2022-02-04 오후 9 08 42

narrow width:
스크린샷 2022-02-04 오후 9 09 38

docs/_includes/default.liquid
```diff
@@ -25,38 +25,14 @@
...
-      >
-        <img
-          src="/images/openjsf-logo.svg"
-          loading="lazy"
-          width="300"
-          height="94"
-          alt="OpenJS Foundation Logo"
-        />
+      <a href="https://wallabyjs.com" rel="external noopener" title="Mocha is sponsored by Wallaby" id="wallaby-logo">
+        <img src="https://wallabyjs.com/assets/img/logoWithText.svg" loading="lazy" width="140" alt="Wallaby logo">
```
juergba, line 33: Some logos are loaded out of the `images` folder, some by external links.
What's the difference? Performance?

juergba (approved):

docs/_includes/default.liquid
outsideris: I think there is no reason. 
Whether we manage all assets on our side or not is only difference. Performance won't make much difference on mocha site.

PR #4795, opened 2021-12-06
chore: update markdownlint-cli

kt3k:
### Description of the Change

Update `markdownlint-cli` to the latest version.

`markdownlint-cli@^0.23.2` (which we currently use) has a vulnerable dependency `rc@~1.2.7` ( root cause: CVE-2020-7788 ). They fixed it at v0.28.0. ref: https://github.com/igorshubovych/markdownlint-cli/pull/181

We should use the latest version to avoid the security issue caused by `rc`.

Not a dependency of the core, but a dependency of the dev toolchain

### Benefits

Security in development

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,7 +85,7 @@
         "markdown-it-emoji": "^1.4.0",
         "markdown-it-prism": "^2.1.1",
         "markdown-toc": "^1.2.0",
-        "markdownlint-cli": "^0.23.2",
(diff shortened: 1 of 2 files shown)


juergba:
@kt3k thank you for this PR.

I don't share your security concerns, especially for a devDependency as _markdownlint-cli_. We will update all _markdown_ dependencies later on in one step as a group.

Issue #4767, opened 2021-10-07
Diffs over 2048 characters are truncated even when there is no performance problem


joshgoebel:
Reference: https://github.com/mochajs/mocha/pull/4638

Can this be a feature flag please so it can be disabled?  I was having NO speed issues before but now my test output is entirely broken (it doesn't show the actual diff at all) because it's a bit over 2000 characters. ☹️

This arbitrary 2048 limit does not seem a good heuristic.

_Originally posted by @joshgoebel in https://github.com/mochajs/mocha/pull/4638#issuecomment-937895384_

KsenyLa:
I have similar question - can it be disabled? Feature flag would be very helpful

jnordberg:
Also running into this, looks like its hardcoded to 2048 chars...

joshgoebel:
Correct. At the very least that should be configurable.

jnordberg:
I think something like this would help in many cases (it certainly would in mine) and will be reasonably fast even for huge inputs.

```js
// truncate beginning of actual and expected strings until first difference if they exceed maxLen
const maxLen = 2048 // TODO: make this configurable :)
const len = Math.max(actual.length, expected.length)
let p = 0
while (len - p > maxLen && actual[p] === expected[p]) {
    p++
}
actual = actual.slice(p)
expected = expected.slice(p)
```

norla:
I have run into this several times so I took the liberty to create a pull request (see above).
@jnordberg I tried including your approach, but skipping to the first diffing character caused some issues with the ouput (not showing enough context, extra diffs at the end etc) so I decided not to use it.

PR #4799, opened 2021-12-15
Configurable max diff size

norla:
### Description of the Change

Fixes https://github.com/mochajs/mocha/issues/4767

Mocha truncates large diffs before displaying a readable unified/inline comparison.
This means that they are not displayed at all for larger strings in many cases, which causes confusion.
This PR makes the limit for when diffs should be truncated configurable, so that projects that have large diffs can still get the outout they expect.

Fix for things already in core :)

### Benefits

Configurable limit ensures that each project can configure the cropping of strings before shoving diff according to their own needs.

### Possible Drawbacks

One more config flag to consider/document etc.

### Applicable issues
https://github.com/mochajs/mocha/issues/4767


nopeless:
Personally never ran into this issue, but I can see myself running into this in the future 👍

sz-piotr:
I have run into this multiple times, would really appreciate this getting more attention.

juergba:
@norla thank you for this PR.

I'm not convinced of appending an additional option for this parameter.
Nevertheless we already have a reporter specific option `reporter-option` which could be enhanced for this case. We already use it for writing test results to a file.

Usage:
- CLI: `--reporter-option diffSize=4048`
- config files: `'reporter-option': ['diffSize=4048']`

norla:
@juergba makes sense, didn't know about reporter options until now :)
The PR is updated accordingly, I named the option "maxDiffSize".

Another thing I just thought of:
The default value 2048 seems very conservative, maybe it can be increased to something like 8192 or 16384 instead?
I think that would prevent a lot of people from having to fiddle with "maxDiffSize" in the first place.

lib/reporters/base.js
```diff
@@ -188,18 +193,21 @@ function stringifyDiffObjs(err) {
...
+
+    actual = actual.slice(0, maxLen);
+    expected = expected.slice(0, maxLen);
+    let result = exports.inlineDiffs
       ? inlineDiff(actual, expected)
       : unifiedDiff(actual, expected);
+    if (skipped > 0) {
+      result = `${result}\nNOTE: Output truncated to ${maxLen} characters, see "maxDiffSize" reporter option.`;
+    }
+    return result;
```
juergba, line 210: Your solution is nice, but please leave the `... Lines skipped` part as is. Changing the output would be a breaking change, I don't want this for now.

juergba (changes requested):

lib/mocha.js
```diff
@@ -1054,6 +1054,10 @@ Mocha.prototype.run = function (fn) {
     exports.reporters.Base.useColors = options.color;
   }
   exports.reporters.Base.inlineDiffs = options.inlineDiffs;
+  exports.reporters.Base.maxDiffSixe =
+    options.reporterOptions &&
+    options.reporterOptions.maxDiffSize &&
+    Number(options.reporterOptions.maxDiffSize);
```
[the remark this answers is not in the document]
norla: Just the Number(...) part or the entire expression? If the latter, where in base.js should I put it exactly?

(I copied the how the inlineDiffs option was handled when implementing this)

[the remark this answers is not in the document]
norla: My guess is that you meant the entire expression, and that it should go into the Base() constructor which has access to the reporter options. Let me know if I'm wrong and I'll rethink it :)

lib/reporters/base.js
```diff
@@ -56,6 +56,11 @@ exports.useColors =
 
 exports.inlineDiffs = false;
 
+/**
+ * Truncate diffs longer than this value to avoid slow performance
+ */
+exports.maxDiffSize = 2048;
```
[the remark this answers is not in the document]
norla: Setting it to a NaN would lead to no limit on the diff size, does not seem right...
I'll change this so that NaN has no effect, and the default value will be used instead.

lib/reporters/base.js
norla: Sure thing, fixed this!

docs/index.md
```diff
@@ -852,6 +854,14 @@ Mocha supports the `err.expected` and `err.actual` properties of any thrown `Ass
...
 ![string diffs](images/reporter-string-diffs.png?withoutEnlargement&resize=920,9999){:class="screenshot" loading="lazy"}
 
+### Max diff size
+
+By default strings are truncated to 2048 characters before generating a diff. This is to prevent performance problems with large strings.
+
+It can however make the output harder to interpret when comparing large strings. Therefore it is possible to configure this value to suit different needs, using `--reporter-option maxDiffSize=[number]`.
+
+A value of 0 indicates no limit, default is 2048 characters.
+
```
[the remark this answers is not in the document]
norla: Done!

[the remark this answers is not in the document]
juergba: I forgot to add `> _New in v9.2.1_` before your description, sorry.
I'm going to classify this PR as a patch fix, since it's not really a new feature. If you agree.

8092 or 8192?

lib/reporters/base.js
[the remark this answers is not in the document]
juergba: Ok. Make sure that no error can be thrown in `base.js`, whatever user's input could be.
Mocha's reaction to errors thrown in a reporter is suboptimal.

lib/reporters/base.js
juergba: I'm not sure about this one, the output will still be different. We will see when doing some final tests ...

norla: The current implementation is broken in my opinion. The "... Lines skipped" part will not be displayed to the user if the diffing part of the string is after the first 2048 bytes (expected and actual will be equal then).

drGeoBN (commented):
LGTM

lib/reporters/base.js
juergba: Yes, you are right.

docs/index.md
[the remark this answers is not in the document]
norla: 8192 of course, patch will do just fine :)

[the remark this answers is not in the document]
juergba: Can you add `> _New in v9.2.1_`, please.

lib/reporters/base.js
juergba: I did some testing, the message doesn't look nice. Since current solution is broken, we have to change the output anyway.
So your first proposition is better.
`${result}\n......[mocha] output truncated to ${maxLen} characters, see "maxDiffSize" reporter-option\n`;
You have to indent the line with six(?) spaces.

lib/reporters/base.js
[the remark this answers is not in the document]
juergba: I tested and `NaN` still results in no limit.

juergba:
@norla please don't push commit by commit, our resources on the CI servers e.g. netlify are limited.
Test locally, then push occasionally.

lib/reporters/base.js
norla: Ok, will fix!

lib/reporters/base.js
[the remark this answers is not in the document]
norla: My bad, fixed now

docs/index.md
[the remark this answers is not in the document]
norla: Done

lib/reporters/base.js
norla: Done

juergba:
@norla thank you so far. I will do some testing tomorrow.
Evtl. I can find a way to simplify this `Number.isNaN` check a little bit.

You could update this issue description (reporter-option / maxDiffSize, 2048=> 8192, fix message) and squash your commits.
Then lets see what the test coverage is telling.

lib/reporters/base.js
```diff
@@ -188,18 +193,21 @@ function stringifyDiffObjs(err) {
...
   try {
-    const diffSize = 2048;
-    if (actual.length > diffSize) {
-      actual = actual.substring(0, diffSize) + ' ... Lines skipped';
-    }
-    if (expected.length > diffSize) {
-      expected = expected.substring(0, diffSize) + ' ... Lines skipped';
-    }
-    return exports.inlineDiffs
+    const maxLen = exports.maxDiffSize || Number.MAX_SAFE_INTEGER;
```
juergba, line 199: `Number.MAX_SAFE_INTEGER` is not IE11 compatible. Our (week) browser tests have passed, nevertheless I prefer to replace it, since `generateDiff()` is part of our browser bundle. You can insert a condition `if (exports.maxDiffSize > 0) { ...` otherwise don't slice anything. With this solution you also prevent weird results when `maxDiffSize` is a negativ number.

lib/reporters/base.js
```diff
@@ -318,6 +326,13 @@ function Base(runner, options) {
   this.runner = runner;
   this.stats = runner.stats; // assigned so Reporters keep a closer reference
 
+  var maxDiffSizeOpt =
+    this.options.reporterOption && this.options.reporterOption.maxDiffSize;
+  console.error();
+  if (maxDiffSizeOpt !== undefined && !Number.isNaN(Number(maxDiffSizeOpt))) {
```
juergba, line 332: The same here with `Number.isNaN()`  and IE11. You can replace it with the global `isNaN()`. Both functions ar weird, but the result should be the same in our case.

lib/reporters/base.js
```diff
@@ -318,6 +326,13 @@ function Base(runner, options) {
   this.runner = runner;
   this.stats = runner.stats; // assigned so Reporters keep a closer reference
 
+  var maxDiffSizeOpt =
+    this.options.reporterOption && this.options.reporterOption.maxDiffSize;
+  console.error();
```
juergba, line 331: remove

test/reporters/base.spec.js
```diff
@@ -168,18 +170,46 @@ describe('Base reporter', function () {
...
 
       inlineDiffsStub.value(false);
       var output = generateDiff(actual, expected);
 
-      expect(output, 'to match', / \.\.\. Lines skipped/);
+      expect(output, 'to match', /output truncated/);
     });
 
     it("should truncate overly long 'expected' ", function () {
+      var actual = 'a foo unified diff';
```
juergba, line 185: These tests don't cover the usage of `reporter-option maxDiffSize=0`. There are similar tests for the `spec` and `json` reporters, indeed terrible tests. Could you have a look and try to add one test, please?

norla:
Sure! I will try yo do it later this weekend.

Den lör 12 feb. 2022 kl 09:59 skrev Juerg B. ***@***.***>:

test/reporters/base.spec.js
juergba: Yes, I guess so. I have a look tomorrow.

juergba: Your tests are ok, `Base.maxDiffSize` gets changed/polluted for the complete test run though. I'm not happy with the entire spec file, I will try to improve it.

juergba (approved):

PR #4800, opened 2021-12-17
fixed: mocha definition is hardly readable #4787

evan-woodworth:
### Requirements

### Description of the Change

To resolve the issue "docs: mocha definition is hardly readable #4787",
changed the font-weight from 200 to 400 for #content > p:first-child, and
changed the font-weight from 100 to 200 for #tag

### Alternate Designs

Considered a heavier font-weight, but 400 and 200 seemed appropriate.

That is where the issue was posted.

### Benefits

Better readability.

### Possible Drawbacks

None come to mind for this minor fix.

### Applicable issues

This is a change to the docs, and shouldn't affect any changes to the application.

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -59,7 +59,7 @@ header {
   animation: fadein 1s forwards, slideright 1s forwards;
   color: #c29d7f;
   font-size: 30px;
-  font-weight: 100;
+  font-weight: 200;
   letter-spacing: 2px;
   margin-top: -158px;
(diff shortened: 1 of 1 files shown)


juergba:
@evan-woodworth thank you.

I'm not a CSS guy, but I don't see any difference with or without this PR. I can't confirm the hardly readable presentation either  described in the issue. I tested with Crome and Edge, the output is identical for font-weight of 100 / 200 / 400.
Why is that?

PR #4778, opened 2021-10-25
Run netlify with Node 16

outsideris:
Netlify runs on node 12, but 12 is a pretty old version.
- I updated it to Node 16
- _supporter.js_: replace `rmdir` by `rm`
- netlify settings: load netlify build image `focal` instead of `xenial`

PR #4805, opened 2021-12-31
docs: fix DEP0147 DeprecationWarning

outsideris:
### Description of the Change
Because `fs.rmdir` usage, Node 16 show [DEP0147 DeprecationWarning](https://nodejs.org/api/deprecations.html#DEP0147).

```
[DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```

* Node 16.0.0: Runtime deprecation.
* Node 14.154.0: Documentation-only deprecation.

### DIFF
--- a/docs/_data/supporters.js
+++ b/docs/_data/supporters.js
@@ -18,7 +18,7 @@
 'use strict';
 
 const {loadImage} = require('canvas');
-const {writeFile, mkdir, rmdir} = require('fs').promises;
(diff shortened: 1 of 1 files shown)


juergba:
Netlify fails because `rm` doesn't exist yet on Node v12.
Superseded by #4778

PR #4807, opened 2022-01-04
Fixes #4803: wrong error thrown if loader is used

### LINKED ISSUE #4803: Missing file/package in import statement misreported as ESM error
### Prerequisites

### Description

This lead me down the garden path: A bad import statement (where the targeted package doesn't exist) results in mocha claiming the containing file was not interpreted as ESM despite containing an `import` (and mocha swallows the actual, correct error message).

Error from Mocha:
```
/…/test.jsx:1
import {
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

Error from Node.js:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@testing-library/jest-dom'
imported from /…/test.jsx
```

The bug originates from https://github.com/mochajs/mocha/blob/v9.1.3/lib/nodejs/esm-utils.js#L51

### Steps to Reproduce

<details>
<summary>test.jsx</summary>

Without installing `@testing-library/jest-dom`:

```js
import '@testing-library/jest-dom';
```
</details>

```console
NODE_ENV=test NODE_OPTIONS="--loader=./loader.mjs" npx mocha ./test.jsx
```

(The loader live-transpiles jsx via esbuild and marks the file as esm; it's not especially important for the reproduction but it did help me find the bug in Mocha).

**Expected behavior:** Mocha reports the correct error (that the targeted package doesn't exist, as node correctly reports).

**Actual behavior:** Mocha wrongly reports a different, unrelated error.

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 9.1.3 (also occurs in 8.3.0)
- The output of `node --version`: 17.2.0
- Your operating system
  - name and version: macOS 12.1
  - architecture (32 or 64-bit): 64 bit
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): esbuild

### Additional Information

N/A

giltayar:
### Requirements

### Description of the Change

When loading test files, there is a heuristic that tries to determine whether to `import` or `require` them. This heuristic failed in the presence of a loader that recognizes filenames that are not `.js/.mjs`. In this case, when `require`-ing the file, Node.js does not throw `ERR_REQUIRE_ESM` (because of the different extension) but rather a `SyntaxError` saying that a CJSJ cannot use `import`.

I added some more code to this heuristic to deal with the problem by checking if the error is the aforementioned syntax error, and if so, throw the original error.

### Alternate Designs

Couldn't think of any.

Well... this is the code of the loading mechanism in Mocha. 

### Benefits

Devs that use ESM with loaders will get the correct error and not a weird error that they can't use to understand what the problem is.

### Possible Drawbacks

Heuristics, as they go, should be tampered with lightly, given that they may succeed in all _our_ tests, but fail in the wild world out there.

I have been careful to add tests to all weird cases I've seen, so I'm pretty confident of the change, but as I said, this may be problematic.

### Applicable issues

Not a breaking change.

### DIFF
--- a/test/integration/fixtures/esm/loader-with-module-not-found/test-that-imports-non-existing-module.fixture.mjs
+++ b/test/integration/fixtures/esm/loader-with-module-not-found/test-that-imports-non-existing-module.fixture.mjs
@@ -1,0 +1,1 @@
+import 'non-existent-package';
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -53,15 +53,30 @@ exports.requireOrImport = hasStableEsmImplementation
           err.code === 'ERR_UNSUPPORTED_DIR_IMPORT'
         ) {
           try {
+            // Importing a file usually works, but the resolution of `import` is the ESM
+            // resolution algorithm, and not the CJS resolution algorithm. So in this case
+            // if we fail, we may have failed because we tried the ESM resolution and failed
+            // So we try to `require` it
             return require(file);
           } catch (requireErr) {
-            if (requireErr.code === 'ERR_REQUIRE_ESM') {
-              // This happens when the test file is a JS file, but via type:module is actually ESM,
+            if (
+              requireErr.code === 'ERR_REQUIRE_ESM' ||
+              (requireErr instanceof SyntaxError &&
+                requireErr
+                  .toString()
+                  .includes('Cannot use import statement outside a module'))
+            ) {
+              // ERR_REQUIRE_ESM happens when the test file is a JS file, but via type:module is actually ESM,
               // AND has an import to a file that doesn't exist.
-              // This throws an `ERR_MODULE_NOT_FOUND` // error above,
+              // This throws an `ERR_MODULE_NOT_FOUND` error above,
               // and when we try to `require` it here, it throws an `ERR_REQUIRE_ESM`.
               // What we want to do is throw the original error (the `ERR_MODULE_NOT_FOUND`),
               // and not the `ERR_REQUIRE_ESM` error, which is a red herring.
+              //
+              // SyntaxError happens when in an edge case: when we're using an ESM loader that loads
+              // a `test.ts` file (i.e. unrecognized extension), and that file includes an unknown
+              // import (which thows an ERR_MODULE_NOT_FOUND). require-ing it will throw the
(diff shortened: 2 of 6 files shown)


giltayar:
Fixes #4803

JakobJingleheimer:
Thank you!

giltayar:
@JakobJingleheimer You're doing important work in Node.js on the loaders. So thank _you_. 🙏

JakobJingleheimer:
Sorry, not sure if you were waiting for me to verify the fix. Just in case: ✅

I replaced my local `node_modules/mocha/lib/nodejs/esm-utils.js` with the one from this PR and re-ran the test, and it correctly errors with

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'i-dont-exist' imported from /…/test.jsx
```

test/integration/fixtures/esm/loader-with-module-not-found/test-that-imports-non-existing-module.fixture.mjs
```diff
@@ -0,0 +1 @@
+import 'non-existent-package';
```
juergba, line 1: Is this file loaded anywhere? Your test just uses the ts file.

giltayar: It is NOT loaded. But it is the reason for the test: the test checks that if a loader is present and an import of a non-existant package happens, then Mocha shows the correct error.

juergba: We have a `*.fixture.ts` file which is to be tested. The loader resolves it to `*.fixture.js` as per your comment. But at the end it's `*.fixture.mjs` which throws the error? From `*.ts` to `*.js` then `*.mjs`?
Either your comment is incorrect, or the `*.fixture.mjs` has no purpose or I don't understand this loader test.

giltayar: The comment is incorrect. The loader resolves it to `test-that-imports-non-existing-module.fixture.mjs`, which is what throws the error.

PR #4811, opened 2022-01-12
perf: replace nanoid with hyperid

### LINKED ISSUE #4539: 'nanoid' do not define a valid './non-secure' target
### Prerequisites

### Steps to Reproduce

I have created a new test project
package.json:
```
{
  "name": "mochatest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "dependencies": {
    "mocha": "^8.2.1"
  }
}
```
test/test.js
```

describe("test",function() {
  it('should test something', function() {
    assert(1=1);
  });
});
```

running the test produces the error (during mocha load)
```
mochatest % npx mocha
Package exports for '/Users/TheFive.osm/development/mochatest/node_modules/nanoid' do not define a valid './non-secure' target
```

**Expected behavior:** 
Test should run without error.

Error Message
```Package exports for '/Users/TheFive.osm/development/mochatest/node_modules/nanoid' do not define a valid './non-secure' target```

Five tests 100 %

### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: see package.json mocha -version does not work, see error.
- The output of `node --version`: V13.1.0
- Your operating system
  - name and version: OSX 10.15.7
- Your browser and version (if running browser tests): n.a.
- Any third-party Mocha-related modules (and their versions): none

bompus:
### Description of the Change

This replaces `nanoid/non-secure` with `hyperid`, which is used by calls to `exports.uniqueID` and `export.assignNewMochaID` from `lib/utils.js`.

### Alternate Designs

There is `crypto.randomUUID`, but that only works in Node.js, and mocha needs to work in the browser as well.

ID generation and the uniqueID function is already in core.

### Benefits

hyperid is ~10x faster than nanoid/non-secure, so using hyperid will speed up test ID generation. From looking through the code and tests, there aren't any requirements for the ID apart from it being a string.

### Possible Drawbacks

If anyone is relying on a the specific format of nanoid/non-secure, the format and length of the string could change.

### Applicable issues

fixes #4539 

### Semver

I believe this can be `semver-patch`, but note Possible Drawbacks.


juergba:
@bompus Thank you.
I don't see the benefit of this PR. _nanoid_ is a very well maintained package.  Issue 4539 is closed and had its origin in Node v13.1.0 which has been end-of-life for quite some time. The performance gain will be within very few milliseconds.

PR #4813, opened 2022-01-17
Assign each worker a worker id in parallel mode.

### LINKED ISSUE #4456: parallel workers should have an ID that can be used to prevent conflicts among workers

forty:
This change is similar to https://github.com/mochajs/mocha/pull/4463 but is cleaner as is uses a new feature from workerpool (`onCreateWorker`) to override each process environment when each process is launched. I'm copying most of the description.

### Requirements

### Description of the Change

This change addresses #4456. It adds a MOCHA_WORKER_ID environment variable that is unique for each worker, so that it may be used for by any test resources that need to use global resources (network ports, filesystem paths, etc.) but ensure that these are unique in order to avoid conflicts during parallel tests.

### Alternate Designs

https://github.com/mochajs/mocha/pull/4463 is an alternate way of doing the same thing, but this one is cleaner (less hacky).

https://github.com/mochajs/mocha/pull/4463 suggest forkArgs might have been used, but this seem really unsafe and hacky.

It is common for tests to use global resources like network ports, which can conflict during parallel testing. Without a feature like this in core, users would be left on their own to create a home-brewed solution (using filesystem files as signals, sockets that elect a leader in the cluster so that one process hands out all global data, a standalone cluster orchestration system), all of which add a lot of complication to writing tests. This would also be a difference when writing tests that run in series vs. parallel, as this concern would not exist (and potentially needs to be turned off) for test in series, and would add pain points when taking existing serial tests and having them run in parallel. This makes using mocha significantly less fun.

This feature takes that pain point away. For example, allocating ports would once again be trivial, using const port = 1234 + Number(process.env.MOCHA_WORKER_ID || 0).

Other parallel testing frameworks, like jest, already use this approach: facebook/jest#2284

### Benefits

This solves common pain points for developers writing tests.

### Possible Drawbacks

Other than more complexity, I can't think of a drawback for the feature itself. The drawback noted by https://github.com/mochajs/mocha/pull/4463 is mostly addressed by this PR, by using a cleaner way to override each process environment. Another pool implementation could easily implement a similar feature.

### Applicable issues

Enter any applicable Issues here.

closes #4456

Mocha follows semantic versioning: http://semver.org

Is this a breaking change (major release)? No

Is it an enhancement (minor release)? Yes

Is it a bug fix, or does it not impact production code (patch release)? No

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1379,6 +1379,12 @@ Free-tier cloud CI services may not provide a suitable multi-core container or V
 
 It's unlikely (but not impossible) to see a performance gain from a [job count](#-jobs-count-j-count) _greater than_ the number of available CPU cores. That said, _play around with the job count_--there's no one-size-fits all, and the unique characteristics of your tests will determine the optimal number of jobs; it may even be that fewer is faster!
 
+### Parallel Mode Worker IDs
+
+> _New in v9.2.0_
+
+Each process launched by parallel mode is assigned a unique id, from 0 for the first process to be launched, to N-1 for the Nth process. This worker id may be accessed in tests via the environment variable `MOCHA_WORKER_ID`. It can be used for example to assign a different database, service port, etc for each test process.
+
 ## Root Hook Plugins
 
 > _New in v8.0.0_
--- a/lib/nodejs/buffered-worker-pool.js
+++ b/lib/nodejs/buffered-worker-pool.js
@@ -75,7 +75,23 @@ class BufferedWorkerPool {
       process.execArgv.join(' ')
     );
 
-    this.options = {...WORKER_POOL_DEFAULT_OPTS, opts, maxWorkers};
+    let counter = 0;
+    const onCreateWorker = ({forkOpts}) => {
+      return {
+        forkOpts: {
+          ...forkOpts,
+          // adds an incremental id to all workers, which can be useful to allocate resources for each process
+          env: {...process.env, MOCHA_WORKER_ID: counter++}
+        }
+      };
+    };
+
+    this.options = {
+      ...WORKER_POOL_DEFAULT_OPTS,
+      ...opts,
+      maxWorkers,
+      onCreateWorker
+    };
     this._pool = workerpool.pool(WORKER_PATH, this.options);
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "strip-json-comments": "3.1.1",
         "supports-color": "8.1.1",
         "which": "2.0.2",
-        "workerpool": "6.1.5",
+        "workerpool": "6.2.0",
         "yargs": "16.2.0",
         "yargs-parser": "20.2.4",
         "yargs-unparser": "2.0.0"
@@ -25120,9 +25120,9 @@
       }
     },
     "node_modules/workerpool": {
-      "version": "6.1.5",
-      "resolved": "https://registry.npmjs.org/workerpool/-/workerpool-6.1.5.tgz",
-      "integrity": "sha512-XdKkCK0Zqc6w3iTxLckiuJ81tiD/o5rBE/m+nXpRCB+/Sq4DqkfXZ/x0jW02DG1tGsfUGXbTJyZDP+eu67haSw=="
+      "version": "6.2.0",
+      "resolved": "https://registry.npmjs.org/workerpool/-/workerpool-6.2.0.tgz",
+      "integrity": "sha512-Rsk5qQHJ9eowMH28Jwhe8HEbmdYDX4lwoMWshiCXugjtHqMD9ZbiqSDLxcsfdqsETPzVUtX5s1Z5kStiIM6l4A=="
     },
     "node_modules/wrap-ansi": {
       "version": "7.0.0",
@@ -45239,9 +45239,9 @@
       }
     },
     "workerpool": {
-      "version": "6.1.5",
-      "resolved": "https://registry.npmjs.org/workerpool/-/workerpool-6.1.5.tgz",
-      "integrity": "sha512-XdKkCK0Zqc6w3iTxLckiuJ81tiD/o5rBE/m+nXpRCB+/Sq4DqkfXZ/x0jW02DG1tGsfUGXbTJyZDP+eu67haSw=="
+      "version": "6.2.0",
+      "resolved": "https://registry.npmjs.org/workerpool/-/workerpool-6.2.0.tgz",
+      "integrity": "sha512-Rsk5qQHJ9eowMH28Jwhe8HEbmdYDX4lwoMWshiCXugjtHqMD9ZbiqSDLxcsfdqsETPzVUtX5s1Z5kStiIM6l4A=="
     },
     "wrap-ansi": {
       "version": "7.0.0",
--- a/test/integration/fixtures/parallel/testworkerid1.mjs
+++ b/test/integration/fixtures/parallel/testworkerid1.mjs
@@ -1,0 +1,7 @@
+import assert from 'assert';
+
+describe('test1', () => {
+  it('should always run on worker with id 0', () => {
+    assert.ok(process.env.MOCHA_WORKER_ID === '0');
+  });
+});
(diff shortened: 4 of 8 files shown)


lib/nodejs/buffered-worker-pool.js
```diff
@@ -75,7 +75,23 @@ class BufferedWorkerPool {
       process.execArgv.join(' ')
     );
 
-    this.options = {...WORKER_POOL_DEFAULT_OPTS, opts, maxWorkers};
+    let counter = 0;
+    const onCreateWorker = ({forkOpts}) => {
+      return {
+        forkOpts: {
+          ...forkOpts,
```
juergba, line 82: I don't see where `process.env` is passed before this PR. Is it part of `WORKER_POOL_DEFAULT_OPTS` (as default without listing it)?
If you add more environmant variables, are they still passed to the child-processes?

test/integration/fixtures/parallel/testworkerid1.mjs
```diff
@@ -0,0 +1,7 @@
+import {describe, it} from '../../../../index.js';
```
juergba, line 1: `describe`, `it` , etc. are on the `global` object, so you can remove these imports.

test/integration/parallel.spec.js
```diff
@@ -18,4 +18,42 @@ describe('parallel run', () => {
...
+    assert.strictEqual(
+      result.tests.filter(
+        test =>
+          test.fullTitle === 'test2 should fail' &&
+          test.err.message === 'MOCHA_WORKER_ID=1'
+      ).length,
+      1
+    );
+    // test 3 may run on either worker 0 or 1 (as jobs = 2)
+    assert.strictEqual(
```
juergba, line 49: When a worker is reused after having finished its test file, it keeps its ID, right? Is this a good solution? Does it help eg. users who want to connect to a different database with each worker/test file?

test/integration/parallel.spec.js
```diff
@@ -18,4 +18,42 @@ describe('parallel run', () => {
...
+  it('should correctly set worker ids for each process', async () => {
+    const result = await runMochaJSONAsync('parallel/testworkerid3.mjs', [
+      '--parallel',
+      '--jobs',
+      '2',
+      require.resolve('./fixtures/parallel/testworkerid1.mjs'),
+      require.resolve('./fixtures/parallel/testworkerid2.mjs')
+    ]);
+    // test 1 always runs on worker 0
+    assert.strictEqual(
```
juergba, line 31: Does GH always allocates enough processes (two) to keep this test stable?

test/integration/parallel.spec.js
forty: I think the main use case is to avoid sharing resources that cannot be used in parallel (ports or DB for example), so reusing resources in each files that are run by the same worker should be fine.

There might be some use cases for having an id per file, but it would be a different use case (in particular that use case would also make sense for non parallel mode, while the current PRs is specifically to solve a problem in parallel mode)

test/integration/parallel.spec.js
forty: Yes, --jobs 2 requests `workerpool` to create 2 processes, no matter how many CPUs are available.

forty:
@juergba thanks a lot for the review! I answered every comments, let me know if something does not make sense.

test/integration/parallel.spec.js
juergba: Do you know how jest handles this ID when re-using a worker?

Do you know by chance how workerpool cleans up a process before re-use? I'm refering eg. to CJS require cache or EMS import cache?

forty: Sorry I missed your answer earlier. The way mocha uses it, workerpool launches processes using child_process.fork, which means it creating new full node processes. 

Then I don't think it's doing anything fancy. The mocha workerpool will run this file https://github.com/mochajs/mocha/blob/master/lib/nodejs/worker.js then the main node process will send the file to run and the child process will process it, then report result and get another file to run. Unless mocha does some special clean up, I think the files running in the same processes will share everything (cache & all)

forty: From what I could understand from a quick look of their code, jest seems to work similarly, though they use their own worker pool implementation, which seems slightly more complex than workerpool, so I'm not 100% sure I understood everything :)

lib/nodejs/buffered-worker-pool.js
juergba: ok, tested.

test/integration/parallel.spec.js
juergba: Thanks for your explanations. Let's go your way then, which means an unique ID for each worker,

package-lock.json
```diff
@@ -29,7 +29,7 @@
         "strip-json-comments": "3.1.1",
         "supports-color": "8.1.1",
         "which": "2.0.2",
-        "workerpool": "6.1.5",
+        "workerpool": "^6.2.0",
```
juergba, line 32: This version should be pinned to 6.2.0. Could rebase to `master` and rebuild `package-lock.json`, please?
Then run a force push to this PR.

test/integration/fixtures/parallel/testworkerid1.mjs
```diff
@@ -0,0 +1,5 @@
+describe('test1', () => {
+  it('should fail', () => {
+    throw new Error(`MOCHA_WORKER_ID=${process.env.MOCHA_WORKER_ID}`);
```
juergba, line 3: You could test for `assert.ok(process.env.MOCHA_WORKER_ID === '0')`, etc.  and then in the spec test you have to check only for three passing tests. Would be much simpler.

forty: Ah! I was not sure it was ok to assert in the fixtures, I'll do that, thanks.

package-lock.json
forty: done!

test/integration/fixtures/parallel/testworkerid1.mjs
forty: done

docs/index.md
```diff
@@ -1379,6 +1379,10 @@ Free-tier cloud CI services may not provide a suitable multi-core container or V
 
 It's unlikely (but not impossible) to see a performance gain from a [job count](#-jobs-count-j-count) _greater than_ the number of available CPU cores. That said, _play around with the job count_--there's no one-size-fits all, and the unique characteristics of your tests will determine the optimal number of jobs; it may even be that fewer is faster!
 
+### Parallel Mode Worker IDs
+
+Each process launched by parallel mode is assigned a unique id, from 0 for the first process to be launched, to N-1 for the Nth process. This worker id may be accessed in tests via the environment variable `MOCHA_WORKER_ID`. It can be used for example to assign a different database, service port, etc for each test process.
```
juergba, line 1384: I'm sorry, I forgot to add here: _New in v9.2.0_

juergba:
In your issue description above: please add the keyword `closes` before the applicable issue No. This is a GH keyword (as `fixes`).

juergba (approved):
@forty thank you

PR #2651, opened 2016-12-29
+ json ouput controls: output-file, output-object

ericprud:
It's practical to use the json reporter's output with in scripts. This patch provides to controls:

**output-object**: pass an object to the reporter so an invoking script can use the results:

```
var Mocha = require('mocha');

var ob = {};

new Mocha().
  addFile("test/toyTest").
  reporter("json", {"output-object": ob}).
  run(function (failures) {
    process.on('exit', function () {
      var s = ob.stats;
      console.log("percentage: " + s.passes/(s.passes+s.failures));
      process.exit(failures > 0 ? 1 : 0);
    });
  });
```

**output-file**: invoke mocha with a reporter option:
`mocha -R json --reporter-options output-file=rpt.json`

PR #4822, opened 2022-01-28
Add JSON Reporter output to object similar to #2651

ashleyfrieze:
### Description of the Change

Add to the existing configuration options for the JSON reporter by allowing a target object to be provided to be populated with test results.

This follows on from the existing ability to specify an output file, but suits situations where files cannot be used to communicate results to the outside world.

### Alternate Designs

We could make a duplicate of the JSON reporter, purely for writing to objects. A workaround is to hack in an alternative of the JSON reporter into the reporters in the runner. However, this is such a small addition to the core, and entirely in keeping with the behaviour of producing a JavaScript model of the test results, that it seems to belong in core.

This is a minor, useful change to an existing reporter.

### Benefits

When executing mocha within a browser test, or explicitly, there's a way to capture the outcome of the test for reporting.

The use case that drove this is where I'm using mocha as the test framework, running it in a virtual browser, driving the tests from a Java application and need to gather the test results from Mocha. I don't have the ability to send them to a file from this environment, but being able to put them into a variable means I can pull the results out at the end.

### Possible Drawbacks

This is a minor alternative to the existing options of the reporter. It should be dormant unless used.

### Applicable issues

- PR #2651

### DIFF
--- a/lib/reporters/json.js
+++ b/lib/reporters/json.js
@@ -32,7 +32,8 @@ exports = module.exports = JSONReporter;
  * @memberof Mocha.reporters
  * @extends Mocha.reporters.Base
  * @param {Runner} runner - Instance triggers reporter actions.
- * @param {Object} [options] - runner options
+ * @param {Object} [options] - runner options - the property 'output' defines a target file to write to
+ *   the property 'targetObject' an object into which the results of the tests will be written
  */
 function JSONReporter(runner, options = {}) {
   Base.call(this, runner, options);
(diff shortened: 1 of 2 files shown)


ashleyfrieze:
Disappointing.

ashleyfrieze:
I think I left a worked example. I can consult on it but may not have much time to finesse it. It seemed like an omission not to be able to do this, and made running mocha inside HTMLUnit in Java a bit tricky.

What do you need to bring this PR in? IIRC it was pretty minor

ashleyfrieze:
@JoshuaKGoldberg - the scenario I was trying to cover was to detect the outcome of a Mocha test that was being run by something other than a usual `npm` or IDE test runner.

In particular, I was running `mocha` inside a Java test. The sut code in question was the embedded javascript being served by the Java application server as part of its SSR. The project, though a Java project, essentially had some JavaScript in it, and wrapping a mocha test within a JUnit test runner required the mocha tests to be executed, but the test runner to be able to acquire the results at the end.

Using a reporter, which put the results into a target object, seemed a neat way to do this, as achieved by the original PR.

I did a remix in this PR, which I also provided a test pack for.

It's a nice to have feature, and pretty innocuous. I worked around it somehow in my project, but could imagine coming back to this and wanting it in future Java SSR situations, or in situations where some mocha tests are being run by a different test framework as part of a transition plan.

JoshuaKGoldberg:
Converting to draft pending triage of #5111.

JoshuaKGoldberg:
Closing per https://github.com/mochajs/mocha/issues/5111#issuecomment-1976653345. Thanks though, I appreciate the code reference & explanations! 🤎

PR #4826, opened 2022-02-07
Create Module not found error

sciffany:
### Requirements

### Description of the Change

In this PR, we throw the "Module Not Found" error when a non-existent module is being imported in the config file. Before this was introduced, an unhelpful message of "Unparsable File Error" is being thrown, which may confuse the users. This aims to resolve issue #4781 

### Alternate Designs

The alternative would be to ignore the "Module Not Found" error, and just display these errors as "Unparsable File". But "Module Not Found" should not be under "Unparsable File" since "Unparsable File" implies that there might be a syntax error somewhere in the program.

This should be in the core because one of the main functionalities of Mocha is to show the users the correct error messages.

### Benefits

A more helpful error message will be shown to users of Mocha, which leads to less confusion when using this package.

### Possible Drawbacks

There are no drawbacks to this change, unless maybe if some old users are not used to seeing this new error message.

### Applicable issues

Not a breaking change
This is an enhancement.
It should not impact production code.

Other notes: Hi, I am a new contributor :) Thank you very much for reviewing!

### DIFF
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -11,7 +11,10 @@ const fs = require('fs');
 const path = require('path');
 const debug = require('debug')('mocha:cli:config');
 const findUp = require('find-up');
-const {createUnparsableFileError} = require('../errors');
+const {
+  createUnparsableFileError,
+  createModuleNotFoundError
+} = require('../errors');
 const utils = require('../utils');
 
 /**
@@ -82,6 +85,12 @@ exports.loadConfig = filepath => {
       config = parsers.json(filepath);
     }
   } catch (err) {
+    if (isModuleNotFoundError(err)) {
+      throw createModuleNotFoundError(
+        `${filepath} is attempting to import a module that does not exist: ${err}`,
+        filepath
+      );
+    }
     throw createUnparsableFileError(
       `Unable to read/parse ${filepath}: ${err}`,
       filepath
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -170,7 +170,14 @@ var constants = {
    * @constant
    * @default
    */
-  UNPARSABLE_FILE: 'ERR_MOCHA_UNPARSABLE_FILE'
+  UNPARSABLE_FILE: 'ERR_MOCHA_UNPARSABLE_FILE',
+
+  /**
+   * File attempts to import non-existent module
+   * @constant
+   * @default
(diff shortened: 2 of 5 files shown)


lib/cli/config.js
```diff
@@ -82,6 +85,12 @@ exports.loadConfig = filepath => {
       config = parsers.json(filepath);
     }
   } catch (err) {
+    if (isModuleNotFoundError(err)) {
+      throw createModuleNotFoundError(
+        `${filepath} is attempting to import a module that does not exist: ${err}`,
+        filepath
+      );
+    }
```
juergba, line 93: It's not clear to me how you are going to distinguish `module not found` errors caused by invalid path (see `parser`) and by invalid imports.

lib/errors.js
```diff
@@ -516,6 +523,25 @@ function createUnparsableFileError(message, filename) {
...
+/**
+ * Creates an error object to be thrown when file attempts to import module that does not exist.
+ * @public
+ * @static
+ * @param {string} message - Error message to be displayed.
+ * @param {string} importerFilename - File name of importing module
+ * @param {string} importedFilename - File name of imported module
+ * @returns {Error} Error with code {@link constants.MODULE_NOT_FOUND}
+ */
+function createModuleNotFoundError(
```
juergba, line 535: We don't need a new Mocha error code, we can't wrap each and every Node error in a new Mocha error object. We just have to rethrow Node's error somehow or include Node's error message in Mocha's `UnparsableFileError` object.

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "mocha",
```
juergba, line 8: This is not correct here. There shouldn't be any `package-lock.json` entry in this PR.

juergba (changes requested):
I haven't understood yet how you are going to solve this issue.

juergba:
`Error: Unable to read/parse ./test/.mocharc.cjs: Error: Cannot find module './test/.mocharc.cjs'`

It's not the first part which is wrong, but the second error message. It should be:
`Error: Unable to read/parse ./test/.mocharc.cjs: Error: Cannot find module '0a9g'`

juergba:
superseded by #4832

PR #4833, opened 2022-02-17
docs: add `expect` as assertion library

SimenB:
### Description of the Change

(Jest maintainer here, for transparency 👋)

`expect` is purposefully kept usable outside the context of Jest, so linking to it makes sense so people are aware they can use it 🙂 

Example

```js
const expect = require('expect');

it('works with expect', () => {
  expect(4).toBe(5);
});
```

image

### Alternate Designs

N/A, website change

N/A, website change

### Benefits

Might be interesting to people if they want to use matchers created for Jest (such as something from `jest-extended`).

Or since mocha can run in browsers, but people might like `expect` from Jest, it's the best of both worlds. 😀 

Semi-related: we're currently working on improving our types (jest is written in typescript), so hopefully soon people will add custom matchers to `expect` directly rather than relying on `@types/jest`. At which point the TypeScript story is hopefully good for `expect` outside of Jest as well. The builtin matchers are typed, of course.

### Possible Drawbacks

One thing I'm not sure about is how to make it clear snapshot matchers doesn't work (as they require support in the test runner). They are not included when you do `require('expect')`, but the docs talk about them without making a distinction. Probably, we can improve our docs in Jest.

### Applicable issues

N/A

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -211,6 +211,7 @@ Mocha allows you to use any assertion library you wish. In the above example, we
 - [chai][] - `expect()`, `assert()` and `should`-style assertions
 - [better-assert][] - C-style self-documenting `assert()`
 - [unexpected][] - "the extensible BDD assertion toolkit"
+- [expect][] - The `expect` function that ships with Jest (except snapshots)
 
 ## Asynchronous Code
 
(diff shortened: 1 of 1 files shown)


juergba:
@SimenB thanks for your proposition.

It doesn't feel right to me to add this link to our docs. I'm suffering and have had sleepless nights. So - if there is no objection by other maintainers - I going to close this PR. You can add your link to our [wiki page](https://github.com/mochajs/mocha/wiki#assertion-libraries) though.

SimenB:
No need to suffer or lose sleep over it 😅 Why don't you want to merge it, though?

SimenB:
(I would still like to know why you don't want it. Feel free to reach out privately, contact info in profile)

PR #4835, opened 2022-02-25
feat: add MOCHA_OPTIONS env variable

### LINKED ISSUE #4330: 🚀 Feature: Support setting options via environment vars
As discussed in #4232, it can be difficult to pass mocha command-line options through `npm run` scripts.  As @boneskull said:

> yargs has a feature that supports setting options via an environment variable; we should probably take advantage of that at some point. Might even be an issue for it already.

I didn't see any open issues for this so figured I would open one.

icholy:
### Description of the Change

The `MOCHA_OPTIONS` environment variable is a space-separated list of command-line options. It essentially copies node's approach with [NODE_OPTIONS](https://nodejs.org/api/cli.html#node_optionsoptions)

The config priority is:

1. Command-line args
2. `MOCHA_OPTIONS` environment variable.
3. RC file (`.mocharc.c?js`, `.mocharc.ya?ml`, `mocharc.json`)
4. `mocha` prop of `package.json`
5. default configuration (`lib/mocharc.json`)

### Alternate Designs

Yargs has support for env variables: https://yargs.js.org/docs/#api-reference-envprefix
However, using this feature would result in a separate env variable for each option:

Ex:

* `MOCHA_BAIL`
* `MOCHA_ALLOW_UNCAUGHT`
* `MOCHA_FAIL_ZERO`
* etc ...

I decided to to go with `MOCHA_OPTIONS` because it seems less intrusive and that's how node does it too.

I don't see how it could live anywhere else.

### Benefits

My specific use-case is specifying `MOCHA_OPTIONS='--inspect-brk'` when calling `npm run test` from vim.

### Possible Drawbacks

* More code is always a liability.
* More config sources can make things more difficult to debug.

### Applicable issues

Fixes #4330.


icholy:
@juergba have you had a chance to look this over?

icholy:
@outsideris I still need this feature. Would I be able to get some feedback?

JoshuaKGoldberg:
👋 coming back to this @icholy, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

icholy:
Yes!

JoshuaKGoldberg:
Yay! We're not quite ready to start reviewing PRs yet, but hope to soon. 

Note that I don't see any linked issues. The contributing guidelines right now ask for that so that folks can discuss the change first. We'll continue to require that after https://github.com/mochajs/mocha/pull/5038. Was there any impetus for this PR besides what's in the description?

icholy:
There's no separate issue, the PR description contains all context I have. I can create an issue if you'd like though.

JoshuaKGoldberg:
That would be lovely, yes please!

Note that the three of us new maintainers are still ramping up on the project, so it might be some weeks (or even months 😞) before we have enough context to really triage the issue and review this PR. We'll do our best to get to it soon though!

icholy:
That's alright. It's been two years, what's another month or two?

lib/cli/options.js
JoshuaKGoldberg, line 1: [Docs] The new `MOCHA_OPTIONS` env variable will need to be documented. https://mochajs.org/#merging is the main place for configuration/options merging info that I know of. There might be others (someone yell at me if there are?).

lib/cli/options.js
```diff
@@ -243,7 +245,14 @@ const loadOptions = (argv = []) => {
...
 
-  args = parse(args._, mocharc, args, rcConfig || {}, pkgConfig || {});
+  args = parse(
+    args._,
+    mocharc,
+    args,
+    rcConfig || {},
+    pkgConfig || {},
+    envConfig
+  );
```
JoshuaKGoldberg, line 255: [Praise] This is really cleanly implemented. Nicely done! ❤️

JoshuaKGoldberg (changes requested):
Sweet! Really pleasing to see such a clean implementation. And with thorough tests! 👏 

Just requesting changes on docs.

lib/cli/options.js
```diff
@@ -210,7 +210,8 @@ module.exports.loadPkgRc = loadPkgRc;
  * 1. Command-line args
  * 2. RC file (`.mocharc.c?js`, `.mocharc.ya?ml`, `mocharc.json`)
  * 3. `mocha` prop of `package.json`
- * 4. default configuration (`lib/mocharc.json`)
+ * 4. `MOCHA_OPTIONS` environment variable.
+ * 5. default configuration (`lib/mocharc.json`)
```
icholy, line 214: @JoshuaKGoldberg I'm not sure about the priority ordering here. I can't decide if env vars should be higher up or not.

shadowspawn: (Drive-by comment from a Yargs and Commander contributor, but not a Mocha user so do not have domain insights.)

Environment variables can be specified on the command-line, and I think are often treated as second in priority behind the command-line args.

Yargs specifies:
> Program arguments are defined in this order of precedence:
> 1. Command line args
> 2. Env vars
> 3. Config file/objects
> 4. Configured defaults

https://yargs.js.org/docs/#api-reference-envprefix

icholy: @shadowspawn that's my gut feeling too after playing with it for a while. @JoshuaKGoldberg cool if I update the PR?

JoshuaKGoldberg: Sounds good to me 🙂 thanks for the insights @shadowspawn!

lib/cli/options.js
icholy: I've updated the docs.

icholy:
@JoshuaKGoldberg do you need anything more from my end for this PR?

docs/index.md
```diff
@@ -2214,11 +2214,16 @@ If no custom path was given, and if there are multiple configuration files in th
 1. `.mocharc.jsonc`
 1. `.mocharc.json`
 
+### Environment Variables
+
+The `MOCHA_OPTIONS` environment variable may be used to specify command line arguments. These arguments take priority over those found in configuration files.
```
JoshuaKGoldberg, line 2219: [Docs] I don't think we can expect all users to intuitively know how this setting CLI args through an env variable looks. How about a bit of explicit help like...

````suggestion
The `MOCHA_OPTIONS` environment variable may be used to specify command line arguments. These arguments take priority over those found in configuration files.

For example, setting the `bail` and `retries` options:

```bash
$ MOCHA_OPTIONS="--bail --retries 3" mocha
```
````

JoshuaKGoldberg (approved):
This looks great to me, thanks ✨ !. Just the one docs suggestion.

We should get a review by another member of @mochajs/maintenance-crew too, just to be safe.

JoshuaKGoldberg:
Btw I generally wait to be [explicitly re-requested review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#re-requesting-a-review) before taking a look - though pinging the way you did works too. That's why this took so long. Sorry for keeping you. 🙂

docs/index.md
icholy: Good call.

icholy:
cc @voxpelli for review

JoshuaKGoldberg:
Sad to say nothing 😞 we're just waiting on the other maintainers. I know they've been pretty busy this season but I've sent pings privately.

voxpelli (approved):
> I don't see how it could live anywhere else.

One can set up a `.mocharc.js` file that reads from `process.env` as exemplified here: https://github.com/mochajs/mocha/blob/master/example/config/.mocharc.js#L3-L5

But after thinking about this I see little harm in adding this option.

The one thing that we maybe would want to add down the line is a flag that opts out from `MOCHA_OPTIONS` to be read and used, but lets deal with that when the need is proven.

Approved 👍

(Sorry for slow review, have been crunching on [neostandard](https://github.com/neostandard/neostandard) + been pondering whether an environment option like this could become an injection attack of some kind, but after some googling I can't find much written that would discourage it)

icholy:
> been pondering whether an environment option like this could become an injection attack of some kind

If someone has write access to your env vars its already game over.

voxpelli:
> > been pondering whether an environment option like this could become an injection attack of some kind
> 
> If someone has write access to your env vars its already game over.

Yeah, but sometimes people try to issue vulnerability reports for the strangest of things

JoshuaKGoldberg:
This was released in [`mocha@10.5.0`](https://www.npmjs.com/package/mocha/v/10.5.0). 🚀

PR #4842, opened 2022-03-09
Fix: wrong error thrown while loading reporter

### LINKED ISSUE #3596: Correctly diagnose errors from required reporter module
### Prerequisites

### Description

When developing a third party reporter, mocha misdiagnoses an error coming from within the reporter as the reporter module not being found.

### Steps to Reproduce

[Minimal reproduction](https://github.com/aido179/mocha-err)
1. Create a parent project that uses mocha.
2. Create a custom reporter, that doesn't require mocha anywhere. 
3. Install that reporter in the parent project (in this case, as a local package using the file location).
4. When mocha `require`s the reporter, node throws an error. The `require` works correctly, but mocha erroneously `console.warn`s that the reporter wasn't found.

**Expected behavior:**

Mocha says that the reporter package caused the error.

**Actual behavior:** 

Mocha erroneously `console.warn`s that the reporter wasn't found.

### Versions

Mocha version: 5.2.0 (not installed globally)
Node version: v8.11.1
OS: MacOs High Sierra v10.13.6
Shell: iTerm2 build 3.2.4beta2

### Additional Information
As discussed in #3530, initially thought the package was not being required correctly.

juergba:
### Description

When developing a third-party reporter, Mocha misdiagnoses an error coming from within the reporter as the reporter module not being found.

### Description of the Change

- _lib/cli/run-helpers.js_: fix CLI plugin entry point
- _lib/mocha.js_: fix/simplify `reporter()` when used programmatically via Mocha API
- adapt existing tests
- adapt webpack configuration

### Applicable issues

closes #3596

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -16,7 +16,6 @@ var Suite = require('./suite');
 var esmUtils = require('./nodejs/esm-utils');
 var createStatsCollector = require('./stats-collector');
 const {
-  warn,
   createInvalidReporterError,
   createInvalidInterfaceError,
   createMochaInstanceAlreadyDisposedError,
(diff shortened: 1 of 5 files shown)


lib/mocha.js
```diff
@@ -335,35 +334,26 @@ Mocha.prototype.reporter = function (reporterName, reporterOptions) {
     }
     // Try to load reporters from process.cwd() and node_modules
     if (!reporter) {
+      let foundReporter;
       try {
-        reporter = require(reporterName);
+        foundReporter = require.resolve(reporterName);
```
AviVahl, line 339: @juergba this line caused a warning when trying to bundle `mocha/mocha.js` using webpack

juergba: @AviVahl This one: `Critical dependency: the request of a dependency is an expression`?
Is this a problem? I need to check wether the module exists, and I can't use a `fs` function either.

AviVahl: Yes, exactly that one.
The warning actually broke several internal projects we have that bundle mocha into the browser. There was infra code there that verified no errors/warnings from bundling.

If the branch is node-specific, would it be possible to completely separate it out of the browser bundle?

juergba: How can I achieve this, exclude a specific code block from the bundle?

Btw the transpiled ES5 bundle (_mocha.js_) will be dropped in a few weeks, and the ES2018 bundle (_mocha-es2018.js_ renamed to _mocha.js_) will be the one and only. If you have any input, would be welcome.

AviVahl: I can think of  several possible ways to achieve this.

I usually separate the isomorphic parts of code from the Node-specific parts.
So: the cli, config loading, reporter loading, dynamic `require`/`require.resolve`, etc would all belong to the Node-specific part.
The `mocha.js` bundle would only include the runner itself and associated APIs.

Another, perhaps trickier, way to achieve this is to specific `"browser": { "./my-node-specific-file.js": null}` which would cause bundlers to skip that file and inject an empty object instead. Runtime checks would be needed to protect against the replaced API.

Regarding es2018? My team welcomes this change. :)

scholtzm: This issue is present even now in v10. Have you found a solution @AviVahl ? I can supress these errors in webpack config but it doesn't seem like the right solution.

AviVahl: We've stopped bundling mocha from our end because of the many bundling regressions it exhibited across the years/versions. Our tool ([mocha-play](https://github.com/wixplosives/mocha-play)) uses the prebuilt bundle (in the npm package) and injects it to the page.

PR #4845, opened 2022-03-12
Drop support of Node v12

juergba:
### Description of the Change

We drop support of Node v12 which will be end-of-life: 2022-04-30

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,4 @@
-node >= 12
+node >= 14
 last 2 Chrome versions
 last 2 Edge versions
 last 2 Firefox versions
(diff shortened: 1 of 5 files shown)


giltayar:
A bit premature, no? End of April is more than a month and a half away.

There's a lot of test code that we can remove: test code that adds the `--experimental-es-modules` (or some such), and which we don't need anymore in Node 14. Cleans up things a bit.

I can do that once this lands.

juergba:
> A bit premature, no? End of April is more than a month and a half away.

No, I don't think so, our milestone `v10.0.0` has due-date of May-01-2022. There is a lot to do until then as e.g. dropping IE11, we will need those six weeks. The creation date of this PR is not identical with the release date of v10.0.0.

> There's a lot of test code that we can remove [...] I can do that once this lands.

Yes, please. And also add `.js` to the extensionless `bin/mocha` executable.

giltayar:
So LGTM

outsideris (approved):

PR #4855, opened 2022-03-21
Implement ESM cache blasting in watch mode

lehni:
### Description of the Change

Mocha [watch mode](https://mochajs.org/#-watch-w) does currently not support ES Module test files because cache blasting was not possible to implement.

This PR implements ESM cache blasting through a custom experimental loader that appends version query to each script URL and increases the version number each time the cache for a given script is blasted. The work is based on the comment of @cspotcode in https://github.com/mochajs/mocha/issues/4374#issuecomment-792326668 and the experiment by @vsnthdev in https://github.com/nodejs/modules/issues/307#issuecomment-764560656.

ESM projects are increasingly common and not being able to run mocha in watch mode in such projects is a downside.

### Benefits

Faster development when working on tests in ESM projects

### Possible Drawbacks

The Node.js loaders are still experimental and subject to future changes. We could follow `node-dev` and mirror their semver based approach to handling loaders for different node versions:

https://github.com/fgnass/node-dev/blob/4cac78e6e3faa40ed26446ed1f097613f58ecc4d/lib/index.js#L97-L108

There is a suspicion that this approach may cause memory leaks, but according to @vsnthdev in https://github.com/nodejs/modules/issues/307#issuecomment-764560656, that is not the case:

> To see if this leaks memory, I made a chain of JS files that import, and then continuously changed their contents 1000+ times, and monitored the RAM usage of `node` and it seems all normal to me. 🤷‍♂️ ESM cache seems to be cleared automatically as soon as no other file imports that particular file.

### Applicable issues

- #4374
- #4854

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -88,7 +88,7 @@ overrides:
         - selector: '*[object.object.name=global][object.property.name=/(Date|(set|clear)(Timeout|Immediate|Interval))/]:expression'
           message: *GH-237
   - files:
-      - test/**/*.mjs
+      - ./**/*.mjs
     parserOptions:
       ecmaVersion: 2018
       sourceType: module
--- a/bin/mocha
+++ b/bin/mocha
@@ -10,6 +10,8 @@
  * @private
  */
 
+const path = require('path');
+const getPackageType = require('get-package-type');
 const {loadOptions} = require('../lib/cli/options');
 const {
   unparseNodeFlags,
@@ -26,6 +28,23 @@ let hasInspect = false;
 
 const opts = loadOptions(process.argv.slice(2));
 debug('loaded opts', opts);
+
+if (
+  opts.watch &&
+  ((opts.require && opts.require.find(str => str.endsWith('.mjs'))) ||
+    // `getPackageType()` only supports files, not folders. A trick to pass in a
+    // folder is to append '/.' to cwd, but `path.join()` would resolve it, so:
+    getPackageType.sync(`${process.cwd()}${path.sep}.`) === 'module')
+) {
+  nodeArgs['no-warnings'] = true;
+  nodeArgs['experimental-loader'] = path.resolve(
+    __dirname,
+    '..',
+    'lib',
+    'nodejs',
+    'esm-loader.mjs'
+  );
+}
 
 /**
  * Given option/command `value`, disable timeouts if applicable
--- a/lib/cli/watch-run.js
+++ b/lib/cli/watch-run.js
@@ -257,9 +257,10 @@ const createRerunner = (mocha, watcher, {beforeRun} = {}) => {
   // true if a file has changed during a test run
   let rerunScheduled = false;
 
-  const run = () => {
+  const run = async () => {
     try {
       mocha = beforeRun ? beforeRun({mocha, watcher}) || mocha : mocha;
+      await mocha.loadFilesAsync();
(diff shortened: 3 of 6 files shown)


juergba:
@lehni thank you for this PR.

> There is a suspicion that this approach may cause memory leaks [...]

I support this opinion. An implementation of a memory leak by design can't be our solution of choice. I personally don't want a solution which works on small scale, but occasionally crashes on larger projects without any chance for a fix.

lehni:
Shouldn't that be tested though before it's rejected?

juergba:
I won't start any testing in that matter, afaik it's a V8 thing and there is no garbage collection.
But go ahead. Mocha's current tests don't cover such scenarios.
This PR should include the `package-lock.json` file, it will fail as is.

lehni:
So I've done some more testing and reading on the subject myself, and sadly, there are indeed memory leaks occurring.

I agree that this is not the way forward, but for the time being, it works for us as a workaround to at least having a functioning watch mode for our ESM project when we need it, and we're happy to accept the leaks as they are not a show-stopper.

juergba:
> I agree that this is not the way forward [...]

@lehni thanks for your effort, it indeed was a interesting approach.

cspotcode:
There was some relevant discussion about this in node's slack recently:
https://node-js.slack.com/archives/CEVD3CX33/p1647844098895659

lehni:
@cspotcode I belive that slack's invite only, perhaps you could quote the important bits here?

cspotcode:
That's just Slack being annoying; it's the "Node Slackers" server linked from node's website:

https://nodejs.org/en/get-involved/
[Node Slackers](https://www.nodeslackers.com/)

lehni:
@juergba a possible way forward is that jest ist doing, using the vm modules through the `--experimental-vm-modules`. I've started playing with this, and it seems like it could work because you can entirely replace the loading and caching with JS code, but my biggest obstacle now appears to be module resolution: I couldn't find an easy way to simply resolve the same way Node would, sadly. `require.resolve()` doesn't work, and I don't have access to `import.meta.resolve()`, etc. Jest seems to have an entire sub-package just for this purpose called `jest-resolve`: https://github.com/facebook/jest/tree/main/packages/jest-resolve/src

Maybe this would work, I haven't tested. Not sure you want to go there… But this seems to be the way right now. I can share my code so far, if it's of interest to you.

But since the code in this PR works for us, I'll probably stop here since it feels like a major undertaking.

cspotcode:
The jest maintainer has some concerns about the VM module.  I don't know all the details, but I think he said that it's missing some features required by jest, and he's not sure if the v8 team is interested in ever adding those features.  It might be worth reaching out to him, I'm sure he'd be willing to share what he knows.
https://github.com/facebook/jest/pull/11529#issuecomment-1027091448

marxangels:
```js

async function loadLibModule(spec) {
    let lib = ModuleCache.get(spec)
    if (lib) return lib

    let module = await import(spec)
    let names = Object.keys(module)
    lib = new vm.SyntheticModule(names, function() {
        for (let name of names) this.setExport(name, module[name])
    }, { context, identifier: spec })

    await lib.link(linker)
    await lib.evaluate()

    ModuleCache.set(spec, lib)
    return lib
}

```
@lehni Why not wrap the builtin/lib modules like this? `import` will do the resolve.

PR #4869, opened 2022-04-11
docs: fix documentation concerning glob expansion on UNIX

### LINKED ISSUE #4868: 📝 Docs: Incorrectly states the shell does glob expansion for double-quoted path specifications
A POSIX-like shell does not do glob expansion for double quoted strings. It does parameter expansion, which, while a form of expansion, does not really have much effect on glob expansion.

The documentation should be corrected.

binki:
Fixes #4868.

### Requirements

### Description of the Change

Fix documentation about how shell expansion works on UNIX.

### Alternate Designs

I removed incorrect information and added the equivalent correct information. I may have been more verbose than necessary, but it seemed like the documentation was trying to be informative here, so I continued providing the supplemental information.

This is a documentation fix. It removes erroneous information in core which should not be there.

### Benefits

The documentation will be less wrong.

### Possible Drawbacks

This may result in people being less confused.

### Applicable issues

#4868

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2332,9 +2332,12 @@ $ mocha "./spec/**/*.js"
 ```
 
 [You should _always_ quote your globs in npm scripts][article-globbing]. If you
-use double quotes, it's the shell on UNIX that will expand the glob. On the
-other hand, if you use single quotes, the [`node-glob`][npm-glob] module will
-handle its expansion.
+use double quotes or single quotes, the [`node-glob`][npm-glob] module will
+handle its expansion. Single quotes are not supported on Windows, so you should
+avoid them. Note that UNIX shells will perform
+[parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_03)
+within double quotes, so avoid using a dollar sign or backslash in any of your
+paths if you want to support UNIX.
 
 See this [tutorial][gist-globbing-tutorial] on using globs.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -2333,7 +2333,7 @@ $ mocha "./spec/**/*.js"
 
 You should _always_ quote your globs in npm scripts. If you
 use quotes, the [`node-glob`][npm-glob] module will
-handle its expansion. To support popular platforms, you must
+handle its expansion. For maximum compatibility,
 surround the entire expression with double quotes and refrain
 from `$`, `"`, `^`, and `\` within your expression.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -2331,13 +2331,11 @@ Some shells support recursive matching by using the globstar (`**`) wildcard. Ba
 $ mocha "./spec/**/*.js"
 ```
 
-[You should _always_ quote your globs in npm scripts][article-globbing]. If you
-use double quotes or single quotes, the [`node-glob`][npm-glob] module will
-handle its expansion. Single quotes are not supported on Windows.
-Note that UNIX shells will perform
-[parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_03)
-within double quotes, so avoid using a dollar sign or backslash in any of your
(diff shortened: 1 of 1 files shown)


juergba (commented):
I agree the docs is incorrect, but I'm not happy with your text either. I'm unsure about your Windows statement since we have countless shells in Windows as eg. git bash, powershell, wsl, cmd, etc. Especially powershell is a pain being so different compared to posix-like shells.
So what the docs should tell (in your own words):
- use quotes to prevent glob expansion by the shell, single/double quotes depending on shell
- it has to be Mocha (via its dependency `node-glob`) which does the glob expansion
- we recommend double quotes for portability. 

Every additional info may be confusing or incorrect, I prefer to not document on shell detail level.

binki:
> we have countless shells in Windows as eg. git bash, powershell, wsl, cmd, etc.

If you shellout in Windows, which is what `npm run` does, you get `CMD`. I do not think that will change in the foreseeable future. People who use different shells on Windows are either power users and know what they are doing or are being told to do so by other people who they can ask for help.

akashchouhan16 (approved):

docs/index.md
```diff
@@ -2332,9 +2332,12 @@ $ mocha "./spec/**/*.js"
 ```
 
 [You should _always_ quote your globs in npm scripts][article-globbing]. If you
-use double quotes, it's the shell on UNIX that will expand the glob. On the
-other hand, if you use single quotes, the [`node-glob`][npm-glob] module will
-handle its expansion.
+use double quotes or single quotes, the [`node-glob`][npm-glob] module will
+handle its expansion. Single quotes are not supported on Windows, so you should
+avoid them. Note that UNIX shells will perform
```
JoshuaKGoldberg, line 2337: Trimming down: avoiding them is implied by the fact that they're not supported. We can let folks make their own conclusion there. Less docs -> more likely to be read.

```suggestion
handle its expansion. Single quotes are not supported on Windows.
Note that UNIX shells will perform
```

docs/index.md
```diff
@@ -2332,9 +2332,12 @@ $ mocha "./spec/**/*.js"
...
+[parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_03)
```
JoshuaKGoldberg, line 2338: [Docs] This reference is pretty long and dry. I suspect devs who are newer to long technical docs and/or don't have the time to read through them will zone out quickly.

What do you think about switching to ...

```suggestion
[parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html)
```

JoshuaKGoldberg (approved):
LGTM, thanks! 🙌 

Just one suggestion around a friendlier docs link. What do you think?

docs/index.md
binki: I prefer to link to the official specification rather than Bash documentation. Bash has various extensions which are not part of the actual specification. The actual specification should be vendor neutral and describe the behavior of any compliant shell, including shells like `dash`.

binki:
> I would simplify it to:
> 
> ```
> [You should _always_ quote your globs in npm scripts][article-globbing], the [`node-glob`][npm-glob] module will handle its expansion.
> ```
> 
> Lets keep additional specifics of shells and os:es out of it.

I addressed your changes in the latest commit.

I checked out the article you are referring to and found that it is misleading. It suggests to use single quotes which are unsupported on Windows. If a folder name contains an ampersand (legal on Windows and Unix-like OSes), then the script author might create an expression which succeeds on a Unix-like OS and fails on Windows. So I removed that reference too.

Since the rationale cannot be included, I needed to add instructions to:

* use double quotes
* avoid obvious problem-causing characters within the double quotes

Please consider the latest changes.

voxpelli (commented):
Better! Thanks 🙏 And thanks for removing the erroneous article.

How would you feel about this minor wording tweak?

voxpelli (approved):

voxpelli:
Thanks @binki 🙏

PR #4872, opened 2022-04-17
CI: conditionally skip 'push' event

juergba:
### Description

When a maintainer pushes a commit to a feature branch linked to an open PR, two events `push` and `pull_request` are emitted. To prevent our workflow `Tests` from executing twice, we skipped the `pull_request` workflow.

As a consequence `coveralls` doesn't work correctly anymore since it's based somehow on the `pull_request` event.

For forked PR's there is only one event emitted which is `pull_request`. 

### Description of the Change

On a `push` event we check with `gh-find-current-pr` wether an open PR exists. If yes, then we skip the `push` workflow. So before this PR, we skipped the `pull_request` event, now we skip the `push` event.

### DIFF
--- a/.github/workflows/mocha.yml
+++ b/.github/workflows/mocha.yml
@@ -13,16 +13,28 @@ on:
 
 jobs:
   prevent-double-run:
+    # skip 'push' event when an open PR exists
     name: Prevent double run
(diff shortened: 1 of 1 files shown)


outsideris (approved):
LGTM

PR #4873, opened 2022-04-18
fix: docs parallel fgrep url fix

ea2305:
Hi, I was reading through the documentation and I noticed that `--fgrep` link was redirecting to the localhost. The link is located in the `Parallel` section, right under [Exclusive tests are disallowed](https://mochajs.org/#exclusive-tests-are-disallowed)

Screen Shot 2022-04-17 at 9 29 57 p m

### Description of the Change

- Updates the link with to link the just header without the host.
- Remove the localhost reference


outsideris (approved):
Thank you.

PR #4885, opened 2022-05-17
Add file path to `xunit` reporter

bmish:
### Description of the Change

Expose test file path for each test case in the `xunit` reporter.

```
<testsuite name="Mocha Tests" tests="2" failures="0" errors="0" skipped="0" timestamp="Tue, 17 May 2022 21:10:34 GMT" time="0.001">
<testcase classname="Array #indexOf()" name="should return -1 when the value is not present" file="/path/to/test/array.js" time="0"/>
<testcase classname="Set #has()" name="should return false when the value is not present" file="/path/to/test/set.js" time="0"/>
</testsuite>
```

The test file path was already added to other reporters in this PR:
* https://github.com/mochajs/mocha/pull/4219

It seems the file path can be added to each reporter as needed.

### Benefits

* Can be easier to find test file from file path instead of just test name, especially when overlapping/similar/identical test names are present, or in a large codebase
* In particular, it can be easier for automated tooling to find the test file from the file path as opposed to the test name when performing test analysis or other scripting

### Possible Drawbacks

* Greater size / noise of output for people who don't need the file path
* Let me know if there's any other reason not to do this

### Applicable issues

None.

This should be part of a minor release as an enhancement.

### DIFF
--- a/lib/reporters/xunit.js
+++ b/lib/reporters/xunit.js
@@ -158,6 +158,7 @@ XUnit.prototype.test = function (test) {
   var attrs = {
     classname: test.parent.fullTitle(),
     name: test.title,
+    file: test.file,
     time: test.duration / 1000 || 0
   };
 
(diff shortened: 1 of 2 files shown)


villelahdenvuo:
I would love to have this for splitting tests on the CI!

bmish:
@villelahdenvuo I just reopened the PR as #4985. Please comment or upvote the PR to express interest.

PR #4888, opened 2022-05-22
fix: support canonical module

JacobLey:
### Description of the Change

Handle canonicalizing `Module`, prevent implicit stringification failures

### Alternate Designs

* Have `canonicalType` return `'object'` for modules (not chosen as that _could_ be breaking, not sure...)
* Have `'module'` fallthrough to `'object'` with no custom logic (not chosen so we can indicate different between "object" and "module")

Current module canonicalization is broken, see https://github.com/mochajs/mocha/issues/4887

### Benefits

Tests can properly handle errors with a Module inside

### Possible Drawbacks

Small chance there is some internal/custom handling expecting canonicalization of modules to fail... Seems unlikely?

### Applicable issues

Fixes https://github.com/mochajs/mocha/issues/4887

Bug fix, patch release

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -215,7 +215,7 @@ exports.type = function type(value) {
 exports.stringify = function (value) {
   var typeHint = canonicalType(value);
 
-  if (!~['object', 'array', 'function'].indexOf(typeHint)) {
+  if (!~['object', 'array', 'function', 'module'].indexOf(typeHint)) {
     if (typeHint === 'buffer') {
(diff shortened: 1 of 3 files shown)


JacobLey:
Still an active issue

JacobLey:
Bump

JoshuaKGoldberg:
👋 coming back to this @JacobLey, are you still interested in working on this PR? As of https://github.com/mochajs/mocha/issues/5027 there are maintainers who can review it now. No worries if you no longer have time - but if you do that'd be great!

PR #4898, opened 2022-07-05
Modify npm usage instructions

weartist:
### Requirements

### Description of the Change

It looks like this usage needs to be replaced
Snip20220705_225


juergba:
@weartist thank you.
This deprecation message is shown when running npm on Windows. It has been reverted, see https://github.com/npm/cli/issues/4980.

PR #4905, opened 2022-07-31
Use standard Promise.allSettled instead polyfill

outsideris:
### Description of the Change
We've introduced [@ungap/promise-all-settled](https://www.npmjs.com/package/@ungap/promise-all-settled) in #4476. 
And Node.js support [Promise.allSettled](https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) in Node 12.9.0.

So, we don't need polyfill anymore because we support Node 14+.

We don't need polyfill anymore.

### Benefits
dependency is reduced.

### Possible Drawbacks
N/A

### DIFF
--- a/lib/nodejs/parallel-buffered-runner.js
+++ b/lib/nodejs/parallel-buffered-runner.js
@@ -6,7 +6,6 @@
 
 'use strict';
 
-const allSettled = require('@ungap/promise-all-settled').bind(Promise);
 const Runner = require('../runner');
 const {EVENT_RUN_BEGIN, EVENT_RUN_END} = Runner.constants;
 const debug = require('debug')('mocha:parallel:parallel-buffered-runner');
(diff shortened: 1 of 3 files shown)


juergba:
@outsideris I guess this PR makes sense.
One of your links points to a korean page. I couldn't find any docs about Node supporting `Promise.allSettled()`. Can you help, please?

outsideris:
Oops! Sorry for korea link. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

CleanShot 2022-08-10 at 03 10 05@2x

And you can see [12.9.0 release note](https://nodejs.org/ko/blog/release/v12.9.0/).

juergba (approved):
@outsideris thank you for your explanations and this PR.

PR #4912, opened 2022-08-17
fix: increase contrast for replay buttons

### LINKED ISSUE #4911: Inaccessible filter/replay buttons in
### Prerequisites

### Description

The ▶️ replay buttons in Mocha browser outputs are very pale against the white background. Here's a screenshot with the middle button in hover state:

Screenshot of a Mocha test suite with three example tests, showing very pale replay buttons on the right, and a middle button less pale

### Steps to Reproduce

Open an `index.html` file that runs Mocha in your browser. https://raw.githubusercontent.com/JoshuaKGoldberg/repros/mocha-browser-replay-accessibility/index.html is an example based on https://mochajs.org/#running-mocha-in-the-browser.

**Expected behavior:**

The buttons should be visible enough that we can see them easily.

**Actual behavior:**

The buttons are very difficult to see, especially in low light settings or for users with limited vision.

Active user interface components and graphical objects such as icons should have a color contrast ratio of at least 3:1 to be AA accessible. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast, a nice explainer for https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.

**Reproduces how often:**

Always.

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`: 10.0.0
- The output of `node --version`: v18.7.0
- Your operating system
  - name and version: MacOS Monterey 12.5
- Your browser and version (if running browser tests): Chrome 104
- Any third-party Mocha-related modules (and their versions): n/a
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): n/a

### Additional Information

I would be honored to send a PR! ❤️ 

Here's an example of changing the replay buttons to `opacity: 0.5` and `color: black` by default, for a 3.94:1 color contrast ratio:

image

IMO that's still kind of pale. Here's the same example with default `opacity: 0.7`, and `box-shadow: 0 0 1px inset black` added to the hover state:

image

JoshuaKGoldberg:
### Requirements

### Description of the Change

Adds the increased visual emphasis as mentioned in #4911.

<table>
<thead>
<tr>
<th>Light mode</th>
<th>Dark mode</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Light mode screenshot with more visible replay buttons
</td>
<td>
Dark mode screenshot with more visible replay buttons
</td>
</tr>
</tbody>
</table>

### Alternate Designs

n/a

n/a

### Benefits

Users can more easily see the replay buttons.

### Possible Drawbacks

n/a

### Applicable issues

* Fixes #4911
* Patch, bugfix

### DIFF
--- a/mocha.css
+++ b/mocha.css
@@ -297,7 +297,7 @@ body {
 }
 
 #mocha .test:hover a.replay {
-  box-shadow: 0 0 1px inset var(--mocha-test-replay-bg-color);
+  box-shadow: 0 0 1px inset var(--mocha-test-replay-color);
   opacity: 1;
 }
 
--- a/mocha.css
+++ b/mocha.css
@@ -17,7 +17,7 @@
   --mocha-test-html-error-color: #000;
   --mocha-box-shadow-color: #eee;
   --mocha-box-bottom-color: #ddd;
-  --mocha-test-replay-color: #888;
+  --mocha-test-replay-color: #000;
   --mocha-test-replay-bg-color: #eee;
   --mocha-stats-color: #888;
   --mocha-stats-em-color: #000;
@@ -49,7 +49,7 @@
      --mocha-test-html-error-color: #fff;
      --mocha-box-shadow-color: #444;
      --mocha-box-bottom-color: #555;
-     --mocha-test-replay-color: #888;
+     --mocha-test-replay-color: #fff;
      --mocha-test-replay-bg-color: #444;
      --mocha-stats-color: #aaa;
      --mocha-stats-em-color: #fff;
@@ -292,11 +292,12 @@ body {
   -moz-transition:opacity 200ms;
   -o-transition:opacity 200ms;
   transition: opacity 200ms;
-  opacity: 0.3;
+  opacity: 0.7;
   color: var(--mocha-test-replay-color);
 }
 
 #mocha .test:hover a.replay {
+  box-shadow: 0 0 1px inset var(--mocha-test-replay-bg-color);
   opacity: 1;
 }
 


juergba:
@JoshuaKGoldberg thank you.
IMO the result looks worse than before with the arrows being white now.

JoshuaKGoldberg:
Shoot, sorry @juergba - I switched them from `black` to `#fff` last minute, forgetting that it's `#000` for black 🤦. Updated now.

juergba:
@JoshuaKGoldberg the support for dark color scheme has been added recently. Please rebase and solve the conflicts, thanks.

mocha.css
```diff
@@ -292,11 +292,12 @@ body {
   -moz-transition:opacity 200ms;
   -o-transition:opacity 200ms;
   transition: opacity 200ms;
-  opacity: 0.3;
+  opacity: 0.7;
   color: var(--mocha-test-replay-color);
 }
 
 #mocha .test:hover a.replay {
+  box-shadow: 0 0 1px inset var(--mocha-test-replay-bg-color);
```
juergba, line 300: @JoshuaKGoldberg thank you. One more question, I'm sorry.

Does this line have any effect, since the box-shadow has the same color as the background? Just the opacity changes from 0.7 to 1?

JoshuaKGoldberg: Ah good point, I'd meant to switch this to having a slight outline color: `--mocha-test-bg-color`. Thanks!

juergba (approved):
@JoshuaKGoldberg thank you for this PR.

Issue #4903, opened 2022-07-20
🔒 Security: Upgrade yargs-parser and yargs to latest stable version


deathstar1708:
Currently the mocha@10.0.0 version has not upgraded its yarg-parser and yargs which is causing a security vulnerability  (NO-CVE: Regular Expression Denial Of Service (ReDoS)) . Please help upgrade both to the most stable version as of current date. Thank you . Attached are the vulnerability and the most stable release in the npm package library
![image](https://user-images.githubusercontent.com/109644072/179895841-a95810b7-dc70-45e4-9886-98cdc58efe6f.png)
![image](https://user-images.githubusercontent.com/109644072/179895952-0c16c07f-acb3-4bce-bce9-11bb65914235.png)

marcel-becker:
Snyk scan is also flagging Mocha ReDos as a High Risk Vulnerability:
https://security.snyk.io/vuln/SNYK-JS-MOCHA-2863123.

guimard:
See also #4938 and #4809

mahnunchik:
Any news about updating yargs-* to latest stable version?

guimard:
Any news about updating yargs-* to latest stable version?

JoshuaKGoldberg:
Marking as accepting PRs. Note that Mocha's current major version supports Node 14, so any version of a new package must also support 14.

JoshuaKGoldberg:
As with https://github.com/mochajs/mocha/issues/5148#issuecomment-2203651573:

> We talked internally and will treat this as `semver-minor`. It's a dependency of the `mocha` package and not an actual exported part of the API. We'll tackle soon! 🚀

PR #4913, opened 2022-08-17
Fix 4903

aljones15:
### Requirements

### Description of the Change
Upgrades `yargs` and `yargs-parser`

Removes security vulnerabilities found in `yargs` related deps.

### Benefits
Upgrades `yargs` to latest version.

### Possible Drawbacks
Could break some cli functionality not properly tested

### Applicable issues
Addresses: https://github.com/mochajs/mocha/issues/4903


juergba (commented):
@aljones15 thank you.

Apparently there have been breaking changes in _yargs_, which you aren't explaining. So I don't understand what you are trying to fix. I have the impression this is a try-and-error PR without deeper understanding.

Your change in the public `main()` is probably a breaking change. Your hardcoded solution in `builder` for the `reporter-option` is a red light for me. The handling of alias and camel-cased option names is a basic task of _yargs-parser_ and _yargs_. We have settings done for this.

What are the relevant changes in _yargs_, what are you trying to fix?

aljones15:
> 

You are correct, hence the draft PR. I'll try to do more research into this if time allows, if not this might provide someone else with some hints about how this upgrade should go.

- [ ] why is `process.exit(1)` needed now while previously `process.exitCode = 1` worked before
- [ ] Why are aliases not calling on `coerce` for `reporterOption`?

PR #4770, opened 2021-10-15
fix the regular expression for function `clean` in `utils.js`

yetingli:
### Description of the Change

Fix the regular expression for function `clean` in `utils.js` to avoid potential Regular Expression Denial of Service (ReDoS) vulnerabilities.

### Alternate Designs

The ReDoS vulnerabilities of the regex  `/^function(?:\s*|\s+[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\s*\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\s*\}|((?:.|\n)*))$/` are mainly due to the overlapped sub-patterns `\s+[^(]*`  and `((?:.|\n)*?)\s*`.
We can ignore `\s` first, because in the end, we can also deal with `\s` using the `trim` function (see the code below）https://github.com/mochajs/mocha/blob/8421974e8d252e2ef11602d15bed178e4f5f4534/lib/utils.js#L92

So I am willing to suggest that you replace the ReDoS-vulnerable regex  `/^function(?:\s*|\s+[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\s*\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\s*\}|((?:.|\n)*))$/` with the safe regex `/^function(?:\s*|\s[^(]*)\([^)]*\)\s*\{((?:.|\n)*?)\}$|^\([^)]*\)\s*=>\s*(?:\{((?:.|\n)*?)\}|((?:.|\n)*))$/`

### Benefits

It achieves functional equivalence and is not subject to ReDoS attacks.

### Applicable issues

See the link https://www.huntr.dev/bounties/1d8a3d95-d199-4129-a6ad-8eafe5e77b9e/

PR #4917, opened 2022-08-25
Add test to prevent ReDoS for clean()

outsideris:
### Description of the Change
In #4770 , ReDoS vulnerabilities of the regex fixed. So, this test to prevent regression ReDoS.


outsideris:
I will merge this in a few days.

juergba:
@outsideris Imo this test has no benefit.

outsideris:
I think this test prevent regression ReDoS because when we fix the RegExp, it is hard to ReDoS vulnerability. 
If you think like that I will close this.

PR #4918, opened 2022-08-30
docs: fix fragment ID for yargs' "extends" documentation

Spencer-Doak:
### Description of the Change

The fragment ID for yargs' documentation of the `extends` keyword was incorrect. I have updated the URL with the correct fragment ID.

### Alternate Designs

N/A

N/A

### Benefits

Readers who click on the link in the documentation are taken to the relevant section of yarg's docs.

### Possible Drawbacks

Less scrolling means less finger exercises for readers, which could be a drawback depending on your personal views.

### Applicable issues

No applicable open issues that I saw.

* Is it a bug fix, or does it not impact production code (patch release)? **Yes, this does not impact the testing framework's code**


JoshuaKGoldberg:
> Less scrolling means less finger exercises for readers, which could be a drawback depending on your personal views.

_chuckle_

JoshuaKGoldberg (approved):
Good catch, thanks! 👍

PR #4921, opened 2022-09-13
feat: deprecation message when an async callback is used in a Suite

### LINKED ISSUE #4920: Fatal error when async functions are passed to `describe` and `context`

I'm one of the maintainers of [a popular tool](https://github.com/nomiclabs/hardhat) that uses Mocha to run our users' tests. We get periodic reports about our tool not running any test ([example](https://github.com/NomicFoundation/hardhat/issues/3117)), and in many cases, this is because the user is passing async functions to `describe`.

I think throwing an error if executing the callback received by `describe`/`context` (I may be missing others) returns a `Promise`, explaining that the callback should be synchronous would be beneficial to Mocha's users.

For typescript users, we could partially achieve this by modifying `@types/mocha` so that the functions I mention stop accepting async functions, but that would be a partial solution, as not everyone uses typescript.

**Additional context**

This is a screenshot of a modified version of Mocha that behaves as proposed.

image

I can send a PR if there's interest in this feature.

alcuadrado:
_I'm sending this PR after having proposed this feature in #4920, getting some positive reactions, and researching the history of a similar feature that was removed_

### Description of the Change

This PR introduces a change that prints a depreciation message if an `async` callback is used in a `Suite`.

### Alternate Designs

`v6.0.0` introduced a similar feature, but it printed a deprecation warning if a `Suite` returned any value other than `undefined`, which was not practical in some cases (see #3744), so it was removed.

This PR is a specialization of that feature that prints a warning when `Promise`s are returned instead.

Also, I originally proposed throwing a fatal error, but that would be a breaking change, so I changed it to a deprecation message. This same criteria was applied to the previous version of this feature.

I don't think this can be implemented outside of core. 

As this PR benefits users who don't have a clear understanding of how a `Suite` works, taking this functionality out of core will have much lower, as there wouldn't be a clear path for those users to discover how to enable it. 

### Benefits

Many users don't realize that Suite's callbacks can't be `async`. In some cases, they get away with `async` callbacks, which further increases the confusion.

This PR is intended to clarify that confusion.

### Possible Drawbacks

I don't think this version of the feature has any serious drawback. 

It may print deprecation warnings for people who were using async callbacks in `Suite`s with some success (and a ton of luck), but those users were already at high risk of their code breaking. This change is intended to help them.

### Applicable issues

Fixes #4920
semver-patch 
(copying the semver criteria of #3550)

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -148,7 +148,7 @@ module.exports = function (suites, context, mocha) {
           const suiteResult = opts.fn.call(suite);
           if (suiteResult instanceof Promise) {
             errors.deprecate(
-              'Suite "' +
+              '[mocha] Suite "' +
                 suite.fullTitle() +
                 '" returned a Promise. ' +
                 'Asynchronous suites are not supported, use a ' +
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -145,7 +145,16 @@ module.exports = function (suites, context, mocha) {
           throw createUnsupportedError('Pending test forbidden');
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          const suiteResult = opts.fn.call(suite);
+          if (suiteResult instanceof Promise) {
+            errors.deprecate(
+              'Suite "' +
+                suite.fullTitle() +
+                '" returned a Promise. ' +
+                'Asynchronous suites are not supported, use a ' +
+                'synchronous callback instead.'
+            );
+          }
           suites.shift();
         } else if (typeof opts.fn === 'undefined' && !suite.pending) {
           throw createMissingArgumentError(
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -146,7 +146,7 @@ module.exports = function (suites, context, mocha) {
         }
         if (typeof opts.fn === 'function') {
           const suiteResult = opts.fn.call(suite);
-          if (suiteResult instanceof Promise) {
+          if (suiteResult !== undefined) {
             errors.deprecate(
               '[mocha] Suite "' +
                 suite.fullTitle() +
--- a/test/integration/fixtures/suite/suite-async-callback.fixture.js
+++ b/test/integration/fixtures/suite/suite-async-callback.fixture.js
@@ -1,3 +1,3 @@
 'use strict';
 
-describe('a suite without an async callback', async function () {});
+describe('a suite with an async callback', async function () {});
--- a/test/integration/fixtures/suite/suite-async-callback.fixture.js
+++ b/test/integration/fixtures/suite/suite-async-callback.fixture.js
@@ -1,0 +1,3 @@
+'use strict';
+
+describe('a suite without an async callback', async function () {});
--- a/test/integration/suite.spec.js
+++ b/test/integration/suite.spec.js
@@ -72,3 +72,24 @@ describe('suite returning a value', function () {
     );
   });
 });
+
+describe('suite w/async callback', function () {
+  it('should print a helpful deprecation message when a callback for suite is async', function (done) {
+    run(
+      'suite/suite-async-callback.fixture.js',
+      args,
+      function (err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(
+          res.output,
+          'to match',
+          /Asynchronous suites are not supported, use a synchronous callback instead./
+        );
+        done();
+      },
+      {stdio: 'pipe'}
+    );
+  });
+});


test/integration/fixtures/suite/suite-async-callback.fixture.js
```diff
@@ -0,0 +1,3 @@
+'use strict';
+
+describe('a suite without an async callback', async function () {});
```
fvictorio, line 3:
```suggestion
describe('a suite with an async callback', async function () {});
```

juergba:
@alcuadrado thank you, I will review within the next few days.

juergba:
@wdavidw are you still working with Mocha and CoffeeScript? 
I don't understand your comment in #3744:

> One way to please everyone is to detect if the return value found in the suite is not the same as the one returned by it.

Does this PR work with CoffeeScript? What happens if the last `it` in a `describe` returns a `Promise`?

alcuadrado:
> I certainly don't agree with the type of warning (deprecation), since we have never offered a feature "asynchronous describe".

I copied this from the previous implementation. I guess the reasoning behind that is that some users are getting away with async callbacks now, and a fatal error can be seen as breaking by them.

lib/interfaces/common.js
```diff
@@ -145,7 +145,16 @@ module.exports = function (suites, context, mocha) {
           throw createUnsupportedError('Pending test forbidden');
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          const suiteResult = opts.fn.call(suite);
+          if (suiteResult instanceof Promise) {
+            errors.deprecate(
```
juergba, line 150: On first glance I didn't like printing a **deprecation** message, since we never have supported async callbacks in `describe`. On the other hand I do like the deprecation message being cashed and printed only once. So we leave it as is.

lib/interfaces/common.js
```diff
@@ -145,7 +145,16 @@ module.exports = function (suites, context, mocha) {
...
+              'Suite "' +
```
juergba, line 151: Please add "[mocha] Suite ..." to the message as done in other parts of our code. The origin of a warning is clearer that way.

juergba (commented):
@alcuadrado I'm quite sure this change will trigger deprecation warnings to CoffeeScript users in certain circumstances. Anyway this PR is much softer than the original proposal and there is a work-around by adding a ` return` statement to the suite. Otherwise they will have to live with a single deprecation warning.

Could you please do some manual browser testing, since we don't run our integration test on browsers. And confirm the outcome, thank you.

lib/interfaces/common.js
JoshuaKGoldberg: @alcuadrado ping - can you make this change?

lib/interfaces/common.js
```diff
@@ -145,7 +145,16 @@ module.exports = function (suites, context, mocha) {
           throw createUnsupportedError('Pending test forbidden');
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          const suiteResult = opts.fn.call(suite);
+          if (suiteResult instanceof Promise) {
```
JoshuaKGoldberg, line 149: Going back to https://github.com/mochajs/mocha/pull/3550/files#diff-28d67f27fa361db19b8a3aa9181d89c5f0d5ae1089167c55f0341675386e3a28R115, I think we no longer need to prioritize supporting CoffeeScript users. The implicit returns that caused #3550 to be reverted in #3759 are not a good idea for the exact reason we saw here: that they unintentionally return values even when those values shouldn't be returned!

```suggestion
          if (suiteResult !== undefined) {
```

https://typescript-eslint.io/rules/no-confusing-void-expression is the right direction to go IMO.

JoshuaKGoldberg (changes requested):
Looks good as a general re-implementation of #3550! Heck, I even think we should go further and fully re-implement to check all truthy values - not just Promises.

Thoughts?

JoshuaKGoldberg:
👋 ping @alcuadrado, is this still something you have time for?

alcuadrado:
Hey, sorry I dropped this. I can take a look over the weekend. 

iirc I dropped this as I don't have the necessary setup to run the browser tests locally.

lib/interfaces/common.js
alcuadrado: I agree that this will be better. And agreed that I wouldn't prioritize CoffeeScript. 

I originally tried to keep the PR minimal (wrt to potential ecosystem impact) as I'm not a maintainer, but I'm happier with your version.

alcuadrado:
I rebased agains `main`, added the `"[mocha]"` tag to the deprecation message, but I can't get the tests to run, not even in `main`.

I'm using  `node@v20.15.0`, `npm@10.7.0`. Maybe using a devcontainer is bringing problems?

alcuadrado:
Sorry, maybe I was wrong. I was getting messages like this one

```
/tmp/mocha-9cPS1H/test.js:7
    done((;
          ^

SyntaxError: Unexpected token ';'
    at wrapSafe (node:internal/modules/cjs/loader:1281:20)
    at Module._compile (node:internal/modules/cjs/loader:1321:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Module._load (node:internal/modules/cjs/loader:1024:12)
    at Module.require (node:internal/modules/cjs/loader:1233:19)
    at require (node:internal/modules/helpers:179:18)
    at /workspaces/mocha/mocha/lib/mocha.js:414:36
    at Mocha.loadFiles (/workspaces/mocha/mocha/lib/mocha.js:411:14)
    at Mocha.run (/workspaces/mocha/mocha/lib/mocha.js:972:10)
    at Object.run (/workspaces/mocha/mocha/lib/cli/watch-run.js:263:22)
    at FSWatcher.<anonymous> (/workspaces/mocha/mocha/lib/cli/watch-run.js:184:14)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

and tons of

```
⚠ [mocha] cleaning up, please wait...
ℹ [mocha] waiting for changes...
ℹ [mocha] waiting for changes...
ℹ [mocha] waiting for changes...
⚠ [mocha] cleaning up, please wait...
ℹ [mocha] waiting for changes...
⚠ [mocha] cleaning up, please wait...
ℹ [mocha] waiting for changes...
ℹ [mocha] waiting for changes...
```

yet the tests seem to pass.

alcuadrado:
Regarding checking against `!== undefined`. While I like the idea, I haven't implemented that yet, as that's a bigger product change, and I don't feel in a position to make that decision. 

I'll gladly implement that and update the tests if that is a change that you (as in the maintainers) are willing to accept it.

JoshuaKGoldberg:
> can't get the tests to run

Do you mean for the GitHub PR? If so: we have to click an _Approve and run_ every time you push a new commit. I've done that just now.

alcuadrado:
> > can't get the tests to run
> 
> Do you mean for the GitHub PR? If so: we have to click an _Approve and run_ every time you push a new commit. I've done that just now.

No, sorry, the error messages during the test run confused me.

Should I do anything about the netlify deploy preview failure?

JoshuaKGoldberg:
> Should I do anything about the netlify deploy preview failure? I don't have permission to see the details

No, that's just hanging around. ou don't have to worry about that.

alcuadrado:
Anything I can do to help move this forward?

JoshuaKGoldberg:
Oh, sorry, I'd misinterpreted your messages and didn't realize this was ready for re-review! Thanks for the ping. Added back to our queue of PRs to review.

lib/interfaces/common.js
JoshuaKGoldberg: @alcuadrado is this still pending? What's up?

JoshuaKGoldberg: Ah, I see https://github.com/mochajs/mocha/pull/4921#issuecomment-2209648248. I'd personally rather we make the breaking change & have this take longer to land. cc @mochajs/maintenance-crew for more opinions.

We've been talking about moving to new major versions soon.

alcuadrado:
> Edit: and, https://github.com/mochajs/mocha/pull/4921/files#r1511274453 is still in discussion, I believe?

Yes, I think so. I'm not familiar with the decision process for things like this within mocha. But anyway, I committed your suggestion, assuming that this will probably/hopefully be accepted.

lib/interfaces/common.js
```diff
@@ -145,7 +145,16 @@ module.exports = function (suites, context, mocha) {
...
         }
         if (typeof opts.fn === 'function') {
-          opts.fn.call(suite);
+          const suiteResult = opts.fn.call(suite);
+          if (suiteResult !== undefined) {
+            errors.deprecate(
+              '[mocha] Suite "' +
+                suite.fullTitle() +
+                '" returned a Promise. ' +
+                'Asynchronous suites are not supported, use a ' +
```
JoshuaKGoldberg, line 154: > Asynchronous

...and this too.

test/integration/suite.spec.js
JoshuaKGoldberg, line 1: [Testing] This has a very clean + good unit test for the async (returning a Promise) case. But since the `common.js` change now is the more general `!== undefined` check, could you also add a unit test for a non-`undefined`, non-`Promise` value?

JoshuaKGoldberg (changes requested):
Progress! 🚀 

Now that the check is for any non-`undefined` value, rather than just `Promise`s, the messaging will need to be updated.

JoshuaKGoldberg:
👋 Ping @alcuadrado, is this still something you have energy & time for?

JoshuaKGoldberg:
Closing out to free up the backing issue. @alcuadrado if you end up seeing this and having time again, no worries, feel free to re-open the PR or send a new one.

Cheers all! 🤎

Issue #4768, opened 2021-10-08
Couldn't store data for global purposes while parallel running


NazarYermolenko:
### Description

Case: I need to collect custom data from tests while parallel running and then I need to use that.
The way I did, I tried to store and collect data in the static object but while parallel running it creates a new static object for each worker anyway, I've tried to use setup, teardown methods, global setup/teardown methods, but still couldn't resolve the issue.

### Steps to Reproduce

Create a few tests async tests (10 parallel tests) for 4 workers, and try to collect generated random numbers in each test to global object or file.

Expect:
The global tear-down setup methods can access in-memory values.
Or to have a container for data between workers 

Actual:
Global teardown/setup methods don't have access to in-memory (test) values.
It's impossible to share or collect data between workers

malammar:
As the workers are separate child processes there's no easy way to send data back to the main process. My recommendation is to have each worker/test write to a temporary file and then read that file at the end of the tests.

PR #4922, opened 2022-09-14
feat: introduce `__mocha_worker_id__` property

epszaw:
### Requirements

### Description of the Change

The PR brings `__mocha_worker_id__` property to tests, suites and hooks, just to understand on reporter level, which worker has been used and categorize multiple parallel tests on reporter level.

At this moment, many solutions are compatible only with the serial mode and when we try to run parallel one current test/suite/hook will be overwritten.

This solution doesn't bring something, which can break the runner or any solution based on it because it adds `__mocha_worker_id__` field what has `MOCHA_WORKER_ID` value (or it equals to `undefined`, e.g. in serial mode).

### Alternate Designs

I've tried to get `MOCHA_WORKER_ID` during `allure-mocha` reporter development, but it's not possible to get it due the reporter receives only IPC message.

Many people run `mocha` in parallel mode and they want to use their own or third-party reporters which are compatible only with serial mode, so we can't store any concurrent data inside. If we have worker ID, we would store running tests by it without any collision.

### Benefits

We can adapt any reporter to work in parallel mode. The main motivation – make `allure-mocha` work.

### Possible Drawbacks

`__mocha_worker_id__` can be `undefined` in the serial mode. Maybe it worth to omit the field, but from my vision, better to keep it all the time.

### Applicable issues

#4768

### DIFF
--- a/lib/hook.js
+++ b/lib/hook.js
@@ -2,7 +2,7 @@
 
 var Runnable = require('./runnable');
 const {inherits, constants} = require('./utils');
-const {MOCHA_ID_PROP_NAME} = constants;
+const {MOCHA_ID_PROP_NAME, MOCHA_WORKER_ID_PROP_NAME} = constants;
 
 /**
  * Expose `Hook`.
@@ -84,6 +84,7 @@ Hook.prototype.serialize = function serialize() {
     state: this.state,
     title: this.title,
     type: this.type,
-    [MOCHA_ID_PROP_NAME]: this.id
+    [MOCHA_ID_PROP_NAME]: this.id,
+    [MOCHA_WORKER_ID_PROP_NAME]: process.env.MOCHA_WORKER_ID
   };
 };
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -19,7 +19,7 @@ const debug = require('debug')('mocha:suite');
 const milliseconds = require('ms');
 const errors = require('./errors');
 
-const {MOCHA_ID_PROP_NAME} = utilsConstants;
+const {MOCHA_ID_PROP_NAME, MOCHA_WORKER_ID_PROP_NAME} = utilsConstants;
 
 /**
  * Expose `Suite`.
@@ -586,8 +586,9 @@ Suite.prototype.serialize = function serialize() {
     $$isPending: Boolean(this.isPending()),
     root: this.root,
     title: this.title,
+    parent: this.parent ? {[MOCHA_ID_PROP_NAME]: this.parent.id} : null,
     [MOCHA_ID_PROP_NAME]: this.id,
-    parent: this.parent ? {[MOCHA_ID_PROP_NAME]: this.parent.id} : null
+    [MOCHA_WORKER_ID_PROP_NAME]: process.env.MOCHA_WORKER_ID
   };
 };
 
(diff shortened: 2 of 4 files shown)


lamartire:
@juergba could someone check this? How it could be merged?

vovsemenv:
@juergba we really need this 👻

epszaw:
@outsideris could you please leave a review here? :)

juergba:
@epszaw I want to understand this issue before taking any decision. Adding some properties by reasoning it wouldn't break anything, doesn't convince at the moment. There are also a few custom reporters working in parallel mode. Some additional information by your side could help.

- " _[...] test/suite/hook will be overwritten_": don't we have an unique ID per suite/test/hook?
- the workers could be re-used (pool), with the same `MOCHA_WORKER_ID`: implications for your proposal?
- "_`__mocha_worker_id__` can be undefined in the serial mode_": you set this property only when serializing, right?

Could you read this issue https://github.com/mochajs/mocha/issues/4403, please? Is it the same kind of issue?

baev:
the main idea is that Allure reporter inits state and then users may modify the state within a test:

```js
it('simple test', function () {
    allure.step('check A',  function () {
        //...
    })
    allure.step('check B',  function () {
        //...
    })
});
```

Allure provides multiple features that allows users to enrich their tests reports with additional metadata such as steps, attachments, tags, labels and so on. All that is usually implemented by custom methods that interact with reporter's state. 
When tests run in parallel mode, we can create a message bus to interact from worker to the parent process (by using files, sockets, events or some other way). But we need somehow to separate events that come from different workers.

Besides that Allure offer Timeline feature -- that shows the distribution of the tests between hosts and workers:

image

epszaw:
@juergba any news here?

epszaw:
@juergba could you please provide any news here?

Rogatto:
Any news about this issue? 

Using allure methods inside a global fixtures there is no error:
```

export const mochaGlobalTeardown = async () => {
  type Environments = Record<string, string>

  const environments: Environments = {
      environment
  }

  allure.writeEnvironmentInfo(environments)
  console.log('Finishing execution!');
}
```

The problem occurs here at the test spec:

`import { allure } from 'allure-mocha/runtime'`

```
beforeEach(function () {
    allure.tms(`[${scenarioId}]`, routeTestManagement + scenarioId)
    allure.severity(Severity.CRITICAL)
    allure.feature(feature)
  })
```

```
 "before each" hook for "it must create a new product":
  TypeError: runtime_1.allure.tms is not a function
```

epszaw:
> Any news about this issue?
> 
> Using allure methods inside a global fixtures there is no error:
> 
> ```
> 
> export const mochaGlobalTeardown = async () => {
>   type Environments = Record<string, string>
> 
>   const environments: Environments = {
>       environment
>   }
> 
>   allure.writeEnvironmentInfo(environments)
>   console.log('Finishing execution!');
> }
> ```
> 
> The problem occurs here at the test spec:
> 
> `import { allure } from 'allure-mocha/runtime'`
> 
> ```
> beforeEach(function () {
>     allure.tms(`[${scenarioId}]`, routeTestManagement + scenarioId)
>     allure.severity(Severity.CRITICAL)
>     allure.feature(feature)
>   })
> ```
> 
> ```
>  "before each" hook for "it must create a new product":
>   TypeError: runtime_1.allure.tms is not a function
> ```

If you have a problem with allure reporter, please, open an issue in the [repo](https://github.com/allure-framework/allure-js)

PR #4926, opened 2022-09-24
deps: update ESLint to 8

outsideris:
### Description of the Change
[ESLint 8](https://eslint.org/blog/2021/10/eslint-v8.0.0-released/) was released a year ago. 
At that time, we couldn't update it because [eslint-config-standard](https://github.com/standard/eslint-config-standard/issues/192) didn't support ESLint. 

I fixed for object-shorthand because StandardJS 17 added [object-shorthand as warning](https://github.com/standard/standard/blob/master/CHANGELOG.md#1700---2022-04-20).

### Alternate Designs
Turn off object-shorthand in the eslint configuration.
There is no reason we don't use object-shorthand.


juergba (approved):
@outsideris thank you

Issue #4655, opened 2021-06-08
ESM Import is cached when re-using the process when using mocha programmatically


smcenlly:
Hey guys - congrats on releasing v9! 🎉

Our product ([Wallaby](https://wallabyjs.com)) uses mocha programmatically and by default reuses the node process between test runs. In the latest version of mocha (`v9`) subsequent changes to a test file are not executing after the first test run. It appears that the import only happens on the first test run which means that on subsequent calls, the context functions (`it`, `describe`, etc.) are not re-executed and test execution bails.

We are not sure if this problem is specific to how Wallaby is calling mocha but we would not have thought so. In the past, we have had to add a cache breaker in the past for other parts of Wallaby and Quokka that natively support ESM. Specifically, [this line of code](https://github.com/mochajs/mocha/blob/master/lib/esm-utils.js#L7) is causing the problem:

```javascript
  // https://github.com/mochajs/mocha/blob/master/lib/esm-utils.js#L7
  return await import(url.pathToFileURL(file));
```

For now, we have patched this ourselves by adding a cache breaker as we have in other parts of Wallaby (same technique as [described in this blog post](https://ar.al/2021/02/22/cache-busting-in-node.js-dynamic-esm-imports/#cache-invalidation-in-esm-with-dynamic-imports)). So the code becomes:

```javascript
  return await import(url.pathToFileURL(file) + "?update=" + Date.now());
```

smcenlly:
After following the technique described in the blog post, we had a problem with the change we suggested. The query string needs to be added after the `pathToFileUrl`. Have updated the original issue description so that it is now correct.

juergba:
@smcenlly thank you.

Mocha is supporting Node's native ESM support since v7.1.0. The ESM loader uses its one cache, different to `require.cache` of CJS files. Node doesn't offer any API to clear the ESM cache, and afaik they will probably never do.
For that same reason we don't support `watch` mode with ESM test files, yet. We have no solution to clear the cache either, and we probably will have to spawn a new process for each `watch` re-run.

> In the latest version of mocha (v9) [...]

There shouldn't be any changes refering to ESM cache from v8.4.0 to v9.0.0. Is there any different behavior?

juergba:
> There shouldn't be any changes refering to ESM cache from v8.4.0 to v9.0.0. Is there any different behavior?

I'm going to close this issue within the next few days.

Issue #4924, opened 2022-09-18
loadFilesAsync and unloading files


fvictorio:
**Is your feature request related to a problem or a nice-to-have?? Please describe.**
Before ESM, you could run mocha programmatically several times during the same process (see issues https://github.com/mochajs/mocha/issues/995 and https://github.com/mochajs/mocha/issues/2783). For example, this works fine:

```js
async function runTest() {
  const Mocha = require("mocha");
  const mocha = new Mocha();

  const cjsTest = path.join(__dirname, `test.js`)
  mocha.addFile(cjsTest);

  await new Promise((resolve) => {
    mocha.run(resolve)
  });

  mocha.dispose()

}

async function main() {
  await runTest()
  await runTest()
}

main()
```

If you want to support both CJS and ESM tests though, you need to add a call to `loadFilesAsync`:

```js
async function runTest() {
  const Mocha = require("mocha");
  const mocha = new Mocha();

  const cjsTest = path.join(__dirname, `test.js`)
  const esmTest =path.join(__dirname, `test.mjs`)
  mocha.addFile(cjsTest);
  mocha.addFile(esmTest);

  await mocha.loadFilesAsync()

  await new Promise((resolve) => {
    mocha.run(resolve)
  });

  mocha.dispose()

}

async function main() {
  await runTest()
  await runTest()
}

main()
```

And in that case, the tests won't run the second time. This happens because the ESM cache is used, and modules cannot be unloaded from there.

**Describe the solution you'd like**
As far as I know, there are two ways to deal with this problem:

- Run the tests in a separate process
- Use a query string or hash in the URL objects passed to the dynamic import to invalidate the cache

The first option doesn't work for us because of performance reasons, since we need to load our application again, which is kind of slow.

The second option might work. It would leak memory, but that's acceptable in our use case.

The problem is that we don't have any way to hook into the code that converts the given files to URL. If that was possible, then we would be able to pass a random hash to the resolved URLs to make sure that the cache is invalidated.

Maybe a better solution would be to accept URL objects in `Mocha.addFile`.

**Describe alternatives you've considered**
Described in the previous section.

**Additional context**
Our tool is a task runner that, among other things, has a `test` task that uses Mocha under the hood. We are working on adding support for ESM, but we need the task to work when it's programmatically called more than once during the same execution.

PR #4945, opened 2022-11-11
feat: ability to decorate esm module name before importing it

j0tunn:
### Description of the Change
This PR adds ability to decorate ESM-module name right before it will be imported by passing `esmDecorator` function to `loadFilesAsync` method.

It helps to deal with ESM cashing issues just by adding some query parameters to modue name:
```js
mocha.loadFilesAsync({
    esmDecorator: file => `${file}?foo=bar`
})
```
And it backwards compatible.
Note though that the sub-dependencies of the test files will not be reloaded.

### Applicable issues
- https://github.com/mochajs/mocha/issues/4655

It is an enhancement (minor release).

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -429,6 +429,8 @@ Mocha.prototype.loadFiles = function (fn) {
  * @see {@link Mocha#addFile}
  * @see {@link Mocha#run}
  * @see {@link Mocha#unloadFiles}
+ * @param {Object} [options] - Settings object.
+ * @param {Function} [options.esmDecorator] - Function invoked on esm module name right before importing it. By default will passthrough as is.
  * @returns {Promise}
  * @example
  *
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -437,7 +437,7 @@ Mocha.prototype.loadFiles = function (fn) {
  *   .then(() => mocha.run(failures => process.exitCode = failures ? 1 : 0))
  *   .catch(() => process.exitCode = 1);
  */
-Mocha.prototype.loadFilesAsync = function () {
+Mocha.prototype.loadFilesAsync = function ({esmDecorator} = {}) {
   var self = this;
   var suite = this.suite;
   this.lazyLoadFiles(true);
@@ -450,7 +450,8 @@ Mocha.prototype.loadFilesAsync = function () {
     function (file, resultModule) {
       suite.emit(EVENT_FILE_REQUIRE, resultModule, file, self);
       suite.emit(EVENT_FILE_POST_REQUIRE, global, file, self);
-    }
+    },
+    esmDecorator
   );
 };
 
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -6,7 +6,7 @@ const forward = x => x;
 const formattedImport = async (file, esmDecorator = forward) => {
   if (path.isAbsolute(file)) {
     try {
-      return await import(esmDecorator(url.pathToFileURL(file)));
+      return await exports.doImport(esmDecorator(url.pathToFileURL(file)));
     } catch (err) {
       // This is a hack created because ESM in Node.js (at least in Node v15.5.1) does not emit
       // the location of the syntax error in the error thrown.
@@ -29,8 +29,10 @@ const formattedImport = async (file, esmDecorator = forward) => {
       throw err;
     }
   }
-  return import(esmDecorator(file));
+  return exports.doImport(esmDecorator(file));
 };
+
+exports.doImport = async file => import(file);
 
 exports.requireOrImport = async (file, esmDecorator) => {
   if (path.extname(file) === '.mjs') {
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -1,10 +1,12 @@
 const path = require('path');
 const url = require('url');
 
-const formattedImport = async file => {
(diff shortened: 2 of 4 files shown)


juergba:
@giltayar could you have a look a this PR, please?

lib/mocha.js
```diff
@@ -437,7 +437,7 @@ Mocha.prototype.loadFiles = function (fn) {
  *   .then(() => mocha.run(failures => process.exitCode = failures ? 1 : 0))
  *   .catch(() => process.exitCode = 1);
  */
-Mocha.prototype.loadFilesAsync = function () {
+Mocha.prototype.loadFilesAsync = function ({esmDecorator} = {}) {
```
giltayar, line 440: Not sure how this works. The default `esmDecorator` is `undefined` and thus calling it will throw an exception.

giltayar (commented):
Besides the comment I added, I am not sure why we need this. What is the use case behind it? Who will use this additional parameter and for what use case.

Until we understand this, I'm not sure we should merge this PR.

lib/mocha.js
j0tunn: `esmDecorator` just passed through to the function `formattedImport` in `esm-utils.js`. And there it has default value

j0tunn:
> Who will use this additional parameter and for what use case.

Any who has problems with esm modules caching while using mocha by API:
- https://github.com/mochajs/mocha/issues/4924
- https://github.com/mochajs/mocha/issues/4655

[node.js doc](https://nodejs.org/api/esm.html#file-urls) suggests only one whay to deal with it - different queries.

There are not too many such cases for now because most of the existing community uses CJS modules for tests for now. But there will be more.
You have a bunch of code for dealing with CJS modules cache, but nothing for ESM-modules. And while this cache problem is not solved you can't say you have full support of ESM modules.

lib/mocha.js
giltayar: Right! I missed that. Me and others... so in the interest of the code documenting itself, can you move the default `forward` function in `esm-utils` to be what defines the default for `esmDecorator` in `Mocha.prototype.loadFilesAsync`.

In other words, move the default up in the call chain.

giltayar:
@j0tunn Ah. I didn't know that `Mocha.prototype.loadFileAsync` was used as an API. In that case it makes total sense given that decorating the urls is the only way to load a different instance of the test files.

Note though that the sub-dependencies of the test files will _not_ be loaded. You will probably need an ESM loader for that.

So if you change the small change that I requested, *and add appropriate tests*, this PR will be approved by me. But don't forget those tests...

lib/mocha.js
j0tunn: Unfortunately `esmUtils.requireOrImport` is used elsewhere in the code, so default value for `esmDecorator` should be set at the level of `esmUtils`.
I think in fact it pretty obvious: no decorator passed - no any decoration

lib/nodejs/esm-utils.js
```diff
@@ -27,15 +29,17 @@ const formattedImport = async file => {
       throw err;
     }
   }
-  return import(file);
+  return exports.doImport(esmDecorator(file));
 };
 
-exports.requireOrImport = async file => {
+exports.doImport = async file => import(file);
```
j0tunn, line 35: This is for the testing purposes. `rewiremock` doesn't work in combination of unexistent files and dynamic imports

j0tunn:
@giltayar done

giltayar:
@juergba from my point of view, we're good to go on this PR, which makes a lot of sense for developers using the API in an ESM setting.

juergba:
@giltayar thank you for your review. 👍 

I will have a look at this PR on coming weekend, @j0tunn please have some patience.

juergba:
@j0tunn Mocha has its own `--require` option which is able to load ESM modules as well. How does it behave with this PR?

Could you provide a short sample for me to do some testing, please?

@fvictorio would you mind reviewing this PR? Does it fix your issue?

j0tunn:
@juergba 
> @j0tunn Mocha has its own --require option which is able to load ESM modules as well. How does it behave with this PR?

As far as I understand `--require` option is about cli. This PR is about API. These are for different use cases. They are not conflicting if that is your question.

> Could you provide a short sample for me to do some testing, please?

There is a short sample in PR description. Someone can use this feature like this in its own code:
mocha.loadFilesAsync({
    esmDecorator = file => `${file}?foo=bar`
})

fvictorio:
Thanks a lot for working on this @j0tunn!

I can confirm this seems to fix my issue if I do something like this:

```typescript
        await (mocha as any).loadFilesAsync({
          esmDecorator: (file) => `${file}?invalidateCache=${Math.random()}`
        });
```

Notice the `as any`: I think this will need a PR in `@types/mocha`. I can do that, but I'm not familiar with the process. Do you have to wait until this is released before updating the types?

juergba:
> [...] I think this will need a PR in `@types/mocha` [...]

I don't know, `@types/mocha` isn't maintained by us. Normally we publish first, then the types get adapted.

> I can confirm this seems to fix my issue if I do something like this

I won't merge this PR without any testing, this could take some time though.

> Note though that the sub-dependencies of the test files will not be loaded.

@j0tunn could you add this statement to this PR's description, please?

j0tunn:
> @j0tunn could you add this statement to this PR's description, please?

Done.

juergba (approved):
@j0tunn thank you

PR #4949, opened 2022-11-23
deps: remove unused deps

outsideris:
### Description of the Change
Remove unused 3 dependencies.

* [configstore](https://www.npmjs.com/package/configstore)
    * It introduced in https://github.com/mochajs/mocha/commit/21371633e704c0735c54909f92cc3afed7eefdb6 . But It appears to have been added by mistake from the beginning.
* [through2](https://www.npmjs.com/package/through2)
    * It was introduced in https://github.com/mochajs/mocha/commit/a81e5558f086808b682b6d196b80893d2c8926f8 7 years ago for browserify. But I couldn't find where to use it in this commit.
    * It looks like removed in [this commit](https://github.com/mochajs/mocha/pull/4293/files#diff-a08fedcb4437f2d1b8f2cc51ea58b2495f52e88e2134817342fcd645f7a73e2a) when we drop the browserify.
* [watchify](https://www.npmjs.com/package/watchify)
    * It was introduced at https://github.com/mochajs/mocha/commit/c04c1d73b30979d04d94398309ee47643ede0af4 . But I couldn't find where to use it in this commit.
    * This is for browserify. Now we don't use browserify anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,6 @@
         "canvas": "^2.9.0",
         "chai": "^4.3.4",
         "coffeescript": "^2.6.1",
-        "configstore": "^5.0.1",
         "coveralls": "^3.1.1",
         "cross-env": "^7.0.2",
         "eslint": "^8.24.0",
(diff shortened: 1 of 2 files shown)


juergba:
@outsideris there is a watch script in _package-scripts.js_, I suppose it uses `watchify` for its purpose. I don't think anybody has ever used that script, so could certainly be removed.

outsideris:
@juergba Did you mean [here](https://github.com/mochajs/mocha/blob/master/package-scripts.js#L284-L287)? I think `docs.watch` doesn't use `watchify`.

Issue #4937, opened 2022-10-22
🔒 Security: Relies on minimatch@3.0.4 version with minimatch ReDoS vulnerability


creharmony:
Hi

I'm not expert of JS deps, but it seems that 10.1.0 rely on vulnerable minimatch dep version 3.0.4
```
└─┬ mocha@10.1.0
  ├─┬ glob@7.2.0
  │ └── minimatch@3.0.4

minimatch  <3.0.5
Severity: high
minimatch ReDoS vulnerability - https://github.com/advisories/GHSA-f8q6-p94x-37v3
```

cf. creharmony/node-etsy-client#56 and fix creharmony/node-etsy-client#57

It would be great to increase minimatch version [from mocha](https://github.com/mochajs/mocha/blob/master/package.json#L64) ? via `"glob": "7.2.0"` ?

NB: the [last glob version](https://github.com/isaacs/node-glob/blob/main/package.json#L23) rely on more recent version of minimatch

joewagner:
I just found this same vulnerability in minimatch <=3.0.4.  I believe this lib should upgrade glob to >=8.0.0 if possible.

WikiRik:
Duplicate of https://github.com/mochajs/mocha/issues/5009 (or rather that one is a duplicate of this earlier issue)
Seems to be mostly fixed in the master branch already, where glob is updated to 8.1.0

JoshuaKGoldberg:
Released in [`mocha@v10.6.0`](https://github.com/mochajs/mocha/releases/tag/v10.6.0). A fresh install of `npm i mocha -D` on a clean directory only has `minimatch@5` now. 🚀

PR #4970, opened 2023-03-08
fix #4837 Update glob due to vulnerability in dep

### LINKED ISSUE #4837: mocha 6.2.3 minimatch fixed version causing security scans to fail
### Prerequisites

### Description

minimatch package versions before 3.0.5 are vulnerable to Regular Expression Denial of Service (ReDoS). Fixed version of minimatch (3.0.4) for mocha version 6.2.3 is causing cloud computing scans to fail. 

In the past I've seen doing some upgrade for security reasons to older major versions so I wanted to know if I need to upgrade this service that is in maintenance mode or not. Thanks a lot in advance :) 

### Steps to Reproduce

N/A

**Expected behavior:** Security scans don't fail.

**Actual behavior:** N/A

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`: 6.2.3

jb2311:
### Description of the Change
The minimatch npm package version <=3.0.4 has a security vulnerability explained here: https://github.com/advisories/GHSA-f8q6-p94x-37v3

Mocha has the [glob package 7.2.0 as a dependency](https://github.com/mochajs/mocha/blob/master/package.json#L64), which in turn has [minimatch ^3.0.4 as a dependency](https://github.com/isaacs/node-glob/blob/v7.2.0/package.json#L23)

This results in a high security risk warning for consumers of mocha. See https://github.com/mochajs/mocha/issues/4937 for an example.

This PR changes the glob dependency to 8.1.0 which no longer depends on the insecure version of minimatch.


jb2311:
@juergba this should be ready to be tested again. 2 things I have concerns about. #1 The tests referenced coffee-script, with a hyphen, but when I ran the tests, I got a "module not found error" and it looks like the hyphen package is depreciated, and mocha's package.json references "coffeescript" with no hyphen, so I updated those tests. Additionally, I got an error during linting that   "eslint-plugin-n" is missing. I though it was my local setup, but it also happened when the tests ran in the CI process. Anyways, I added that package to the dev deps. Let me know if you have any questions or concerns

juergba:
@jb2311 thank you for this PR.

We will see wether our CI tests will pass. Tbh I don't trust your `package-lock.json`.

jb2311:
esline-plugin-node is no longer being maintained, so eslint-plugin-n (https://github.com/eslint-community/eslint-plugin-n) is a fork that is still active. However, I have no idea why it's erroring out without it suddenly. It happened on the first test run, when I did nothing but update the glob package https://github.com/mochajs/mocha/actions/runs/4368808387/jobs/7681675449

package.json
```diff
@@ -61,7 +61,7 @@
     "diff": "5.0.0",
     "escape-string-regexp": "4.0.0",
     "find-up": "5.0.0",
-    "glob": "7.2.0",
+    "glob": "8.1.0",
```
juergba, line 64: glob has published v9*. Is mocha affected by v9 breaking changes, or why are you limiting your upgrade to v8?

package.json
```diff
@@ -167,7 +167,7 @@
     "./lib/nodejs/file-unloader.js": false,
     "./lib/nodejs/parallel-buffered-runner.js": false,
     "./lib/nodejs/serializer.js": false,
-    "./lib/nodejs/worker.js": false,
+    ".  /lib/nodejs/worker.js": false,
```
juergba, line 170: please remove space.

package.json
```diff
@@ -183,4 +183,4 @@
       "pre-commit": "lint-staged"
     }
   }
-}
+}
```
juergba, line 186: is there an empty line missing at the end?

juergba (commented):
I'm not convinced completely, I'm afraid there might be some unwanted side-effects by `windowsPathsNoEscape`. Anyway let's do it.

jb2311:
@juergba updated

package.json
```diff
@@ -184,3 +184,4 @@
     }
   }
 }
+
```
juergba, line 187: Please ... I don't want to mess with linters or whatever, so please just leave this untouched. There has to be one empty line at the end. There is no need to change on this eof for this PR.

jb2311:
@juergba okay, I think the whitespace should be appropriate.

package.json
jb2311: It looks like the file in master doesnt have the empty line at the end

PR #4980, opened 2023-04-24
fix!: correct precedence of configuration extends and defaults 

### LINKED ISSUE #3916: Does the extending configuration feature work?

### LINKED ISSUE #4979: 🐛 Bug: Some Settings Cannot Be Inherited (extends keyword)
### Prerequisites

### Description

Currently, `mocha` will load default settings from `lib/mocharc.json` from this package and only after doing so process the configuration using `yargs`.

This breaks `yargs` `extends` mechanism as the values predefined in said `mocharc.json` file cannot be overridden.

This means that following settings cannot be specified using the `extends` config:
  - `diff`
  - `package`
  - `reporter`
  - `slow`
  - `timeout`
  - **`ui`**

I assume that arrays still do work as, to my understanding, yargs will concatenate arrays taken from `extends` configs.

### Steps to Reproduce
 1. Create a project with `mocha` tests written using the `tdd` ui
 2. In your project's root, create a file `base.json` with the following content:
    ```json
    {
        "ui": "tdd"
    }
    ```
 3. In your project's root, create a file called `.mocharc.json`:
    ```json
    {
        "timeout": 10000
    }
    ```
  4. Try to run your `tdd` tests by running the `mocha` command

**Expected behavior:**

`mocha` should be able to execute this test as the setting `"ui": "tdd"` is supposed to be inherited.

`mocha` will instead show the following error message:

```
ReferenceError: suite is not defined
```

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`: 10.0.2
- The output of `node --version`: 20.0.0
- Your operating system
  - name and version: arch1
  - architecture (32 or 64-bit): amd64
- Your shell (e.g., bash, zsh, PowerShell, cmd): `bash`, PowerShell, `cmd`
- ~~Your browser and version (if running browser tests):~~
- ~~Any third-party Mocha-related modules (and their versions):~~
- ~~Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version):~~

### Additional Information
Notice that default values from this package's `lib/mocharc.json` file are applied to the current config file during the execution of `import("mocha/lib/cli/options").loadOptions()` in this piece of code:

https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli/options.js#L246

Only after invoking `loadOptions()`, the `extends` setting is processed by `yargs`:

  - First, `loadOptions()` is executed - in doing so, values from `lib/mocharc.json` are applied
    https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli/cli.js#L47
  - After that, features such as `extends` are processed by `yargs`:
    https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli/cli.js#L77

Related to this question: #3916
There is also a PR addressing this issue: #4407
Furthermore, I created a PR myself: #4980

neurolag:
### Description of the Change

- [x] Addresses an existing open issue: fixes #3916
- [x] That issue was marked as [status: accepting prs](https://github.com/mochajs/mocha/issues?q=sort%3Aupdated-desc+is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md?rgh-link-date=2026-04-18T14%3A45%3A43Z) were taken

As described in issue #4979, due to `mocha` first loading its default settings and after doing so processing the arguments using `args`, there are some settings which cannot be inherited by specifying an `extends` file:

  - First, default values get loaded:
    https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli/options.js#L246
  - Sometime later, the configuration gets processed:
    https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli/cli.js#L77

Changes made in this PR processes configuration files using `yargs` right after they have been parsed. Thus, the order of events is now the following:
  - RC files are parsed
    - RC files are processed using `yargs`
  - `package.json` settings are loaded
  - Collected configs and `argv` get merged
  - `mocha`s default values (from `lib/mocharc.json`) get applied

### Alternate Designs

Configuration files could be processed right before `mocha`s defaults get applied. However, the context of the file path would get lost thus causing relative `extends` paths to possibly fail.

Applying `mocha`s default values could be removed from here:
https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli
And instead be done somewhere around here (after `yargs` has been executed):
https://github.com/mochajs/mocha/blob/37deed262d4bc0788d32c66636495d10038ad398/lib/cli/cli.js#L47

Currently, the inheritance will not work for a number of settings while other settings do. This might cause confusion and unexpected results. Thus, I'd recommend to have this change in core. 

### Benefits

All settings will be able to be inherited while `mocha`s defaults still get applied if necessary.

### Possible Drawbacks

None that I have heard of.
If people already used `extends` for their own project, there is a chance that they relied on the inheritance of `ui` to not work, so there is a small possibility they will have to perform a one-time change to their config file.

### Applicable issues
  - #4979
  - #3916
  - Original PR: #4407

* [ ] ~~Is this a breaking change (major release)?~~
* [ ] Is it an enhancement (minor release)? (Maybe?)
* [x] Is it a bug fix, or does it not impact production code (patch release)?

Or probably a minor as there is a small possibility people have to adjust their configuration if they used `extends` before

Fixes #4979
Fixes #3916

Co-authored-by: Pat Herlihy @mf-pherlihy

### DIFF
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -11,6 +11,7 @@ const fs = require('fs');
 const path = require('path');
 const debug = require('debug')('mocha:cli:config');
 const findUp = require('find-up');
+const yargs = require('yargs/yargs');
 const {createUnparsableFileError} = require('../errors');
 const utils = require('../utils');
 
@@ -76,6 +77,10 @@ exports.loadConfig = filepath => {
     } else {
       config = parsers.json(filepath);
     }
+
+    config = yargs(undefined, path.dirname(filepath))
+      .parserConfiguration(require('./options').YARGS_PARSER_CONFIG)
+      .config(config);
   } catch (err) {
     throw createUnparsableFileError(
       `Unable to read/parse ${filepath}: ${err}`,
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -78,7 +78,7 @@ exports.loadConfig = filepath => {
       config = parsers.json(filepath);
     }
 
-    const {$0, ...options} = yargs(undefined, path.dirname(filepath))
+    const {$0, ...options} = yargs(config._, path.dirname(filepath))
       .parserConfiguration(require('./options').YARGS_PARSER_CONFIG)
       .config(config).argv;
 
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -78,9 +78,11 @@ exports.loadConfig = filepath => {
       config = parsers.json(filepath);
     }
 
-    config = yargs(undefined, path.dirname(filepath))
+    const {$0, ...options} = yargs(undefined, path.dirname(filepath))
       .parserConfiguration(require('./options').YARGS_PARSER_CONFIG)
-      .config(config);
+      .config(config).argv;
+
+    config = options;
   } catch (err) {
     throw createUnparsableFileError(
       `Unable to read/parse ${filepath}: ${err}`,
--- a/lib/cli/config.js
+++ b/lib/cli/config.js
@@ -11,7 +11,6 @@ const fs = require('fs');
 const path = require('path');
 const debug = require('debug')('mocha:cli:config');
 const findUp = require('find-up');
-const yargs = require('yargs/yargs');
 const {createUnparsableFileError} = require('../errors');
 const utils = require('../utils');
 
@@ -77,12 +76,6 @@ exports.loadConfig = filepath => {
     } else {
       config = parsers.json(filepath);
     }
-
-    const {$0, ...options} = yargs(config._, path.dirname(filepath))
-      .parserConfiguration(require('./options').YARGS_PARSER_CONFIG)
-      .config(config).argv;
-
-    config = options;
   } catch (err) {
     throw createUnparsableFileError(
       `Unable to read/parse ${filepath}: ${err}`,
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -8,7 +8,9 @@
  */
 
 const fs = require('fs');
+const path = require('path');
 const ansi = require('ansi-colors');
+const yargs = require('yargs/yargs');
 const yargsParser = require('yargs-parser');
 const {
   types,
@@ -212,7 +214,7 @@ const parse = (args = [], defaultValues = {}, ...configObjects) => {
 const loadRc = (args = {}) => {
   if (args.config !== false) {
     const config = args.config || findConfig();
-    return config ? loadConfig(config) : {};
+    return config ? loadRcFile(loadConfig(config), config) : {};
   }
 };
 
@@ -254,7 +256,7 @@ const loadPkgRc = (args = {}) => {
       const pkg = JSON.parse(configData);
       if (pkg.mocha) {
         debug('`mocha` prop of package.json parsed: %O', pkg.mocha);
-        result = pkg.mocha;
+        result = loadRcFile(pkg.mocha, filepath);
       } else {
         debug('no config found in %s', filepath);
       }
@@ -270,6 +272,20 @@ const loadPkgRc = (args = {}) => {
 };
 
 module.exports.loadPkgRc = loadPkgRc;
+
+/**
+ * Loads the inherited configuration of the rc file located at the specified {@linkcode filePath}.
+ *
+ * @param {Object} config - The parsed configuration of the rc file.
+ * @param {string} filePath - The name of the file containing the parsed configuration.
+ */
+const loadRcFile = (config, filePath) => {
+  const {$0, ...options} = yargs(config._, path.dirname(filePath))
+    .parserConfiguration(configuration)
+    .config(config).argv;
+
+  return options;
+};
 
 /**
  * Priority list:
--- a/test/integration/fixtures/config/mocharc-extended/package-lock.json
+++ b/test/integration/fixtures/config/mocharc-extended/package-lock.json
@@ -1,0 +1,6 @@
+{
+    "mocha": {
+        "extends": "./extends.json",
+        "grep": "package"
+    }
+}
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -42,4 +42,16 @@ describe('options', function () {
     expect(extended.reporter, 'to equal', 'html');
     expect(extended.slow, 'to equal', 30);
   });
+
+  it('Should support extended options using package.json', function () {
+    var extended = loadOptions([
+      '--package',
+      path.join(workspaceDir, 'package-lock.json')
+    ]);
+    expect(extended.require, 'to equal', ['foo', 'bar']);
+    expect(extended.bail, 'to equal', true);
+    expect(extended.reporter, 'to equal', 'html');
+    expect(extended.slow, 'to equal', 30);
+    expect(extended.grep, 'to equal', 'package');
+  });
 });
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -11,6 +11,7 @@ describe('options', function () {
     'config',
     'mocharc-extended'
   );
+  const configFile = path.join(workspaceDir, 'extends.json');
 
   beforeEach(function () {
     workingDirectory = process.cwd();
@@ -21,11 +22,13 @@ describe('options', function () {
     process.chdir(workingDirectory);
   });
 
+  it('Should load `_`-option properly', function () {
+    var extended = loadOptions(['--config', configFile]);
+    expect(extended._, 'to equal', ['**/*.spec.js']);
+  });
+
   it('Should support extended options using --config parameter', function () {
-    var extended = loadOptions([
-      '--config',
-      path.join(workspaceDir, 'extends.json')
-    ]);
+    var extended = loadOptions(['--config', configFile]);
     expect(extended.require, 'to equal', ['foo', 'bar']);
     expect(extended.bail, 'to equal', true);
     expect(extended.reporter, 'to equal', 'html');
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -45,6 +45,7 @@ describe('options', function () {
 
   it('Should support extended options using package.json', function () {
     var extended = loadOptions([
+      '--no-config',
       '--package',
       path.join(workspaceDir, 'package-lock.json')
     ]);
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -1,0 +1,23 @@
+'use strict';
+
+var path = require('path');
+var loadOptions = require('../../lib/cli/options').loadOptions;
+
+describe('options', function () {
+  it('Should support extended options', function () {
+    var configDir = path.join(
+      __dirname,
+      'fixtures',
+      'config',
+      'mocharc-extended'
+    );
+    var extended = loadOptions([
+      '--config',
+      path.join(configDir, 'extends.json')
+    ]);
+    expect(extended.require, 'to equal', ['foo', 'bar']);
+    expect(extended.bail, 'to equal', true);
+    expect(extended.reporter, 'to equal', 'html');
+    expect(extended.slow, 'to equal', 30);
+  });
+});
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -43,6 +43,13 @@ describe('options', function () {
     expect(extended.slow, 'to equal', 30);
   });
 
+  it('Should support standalone options using standalone rc file', function () {
+    var extended = loadOptions(['--config', configFileStandalone]);
+    expect(extended.require, 'to equal', ['foo', 'bar']);
+    expect(extended.bail, 'to equal', true);
+    expect(extended.reporter, 'to equal', 'dot');
+    expect(extended.slow, 'to equal', 60);
+  });
   it('Should support extended options using package.json', function () {
     var extended = loadOptions([
       '--no-config',
--- a/test/integration/options.spec.js
+++ b/test/integration/options.spec.js
@@ -4,17 +4,36 @@ var path = require('path');
 var loadOptions = require('../../lib/cli/options').loadOptions;
 
 describe('options', function () {
-  it('Should support extended options', function () {
-    var configDir = path.join(
-      __dirname,
-      'fixtures',
-      'config',
-      'mocharc-extended'
-    );
+  var workingDirectory;
+  const workspaceDir = path.join(
+    __dirname,
+    'fixtures',
+    'config',
+    'mocharc-extended'
+  );
+
+  beforeEach(function () {
+    workingDirectory = process.cwd();
+    process.chdir(workspaceDir);
+  });
+
+  afterEach(function () {
+    process.chdir(workingDirectory);
+  });
+
+  it('Should support extended options using --config parameter', function () {
     var extended = loadOptions([
       '--config',
-      path.join(configDir, 'extends.json')
+      path.join(workspaceDir, 'extends.json')
     ]);
+    expect(extended.require, 'to equal', ['foo', 'bar']);
+    expect(extended.bail, 'to equal', true);
+    expect(extended.reporter, 'to equal', 'html');
+    expect(extended.slow, 'to equal', 30);
+  });
+
+  it('Should support extended options using rc file', function () {
+    var extended = loadOptions([]);
     expect(extended.require, 'to equal', ['foo', 'bar']);
     expect(extended.bail, 'to equal', true);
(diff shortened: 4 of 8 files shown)


manuth:
Just pinging to have the `stale` label removed
I'd really love to see this implemented

test/integration/options.spec.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+var path = require('path');
+var loadOptions = require('../../lib/cli/options').loadOptions;
+
+describe('options', function () {
+  var workingDirectory;
```
JoshuaKGoldberg, line 7: [Question] Is there a reason not to make this set only once? Does it have to be set before each test?

```suggestion
  const workingDirectory = process.cwd();
```

test/integration/options.spec.js
JoshuaKGoldberg, line 1: [Praise] As with #4407's tests, these are nice! I like that this also has an added second one for both `--config` and file coverage.

lib/cli/config.js
```diff
@@ -76,6 +77,10 @@ exports.loadConfig = filepath => {
     } else {
       config = parsers.json(filepath);
     }
+
+    config = yargs(undefined, path.dirname(filepath))
+      .parserConfiguration(require('./options').YARGS_PARSER_CONFIG)
+      .config(config);
```
JoshuaKGoldberg, line 83: Noting https://github.com/mochajs/mocha/pull/4407/files#r1509449877 here: would be nice to have discussion on which API option to use. I'm leaning towards this one, but want to hear if there are advantages to `applyExtends` I missed.

test/integration/options.spec.js
JoshuaKGoldberg, line 1: [Non-Actionable] These tests look like they're a continuation of the code in #4407. I.e. their code was partially written by @mf-pherlihy in #4407. I added in a `Co-authored-by` attribution at the bottom of the PR description. Just a heads up. 🙂

JoshuaKGoldberg (dismissed):
Seems good and reasonable, nice! 🔥 

Marking as requesting changes just so we can chat about the two possible implementations a bit.

JoshuaKGoldberg:
Marking as `semver-major` because some projects might accidentally be relying on the current, arguably broken behavior.

test/integration/options.spec.js
manuth: Whenever I change directories for performing tests, I use to always restore the previous working directory in order to not have it interfer with other tests and `beforeEach` functions.

I could entirely drop it or replace it with `beforeAll`/`afterAll`

lib/cli/config.js
manuth: I'm not entirely sure anymore why I decided to use this function instead.
I remember that when using a different approach, some options were overridden by default.

I will read through my commit messages during the upcoming weekend. I really hope I can reproduce my thoughts this way.

king-2929 (approved):
Fix it

king-2929 (approved):
Fix it

king-2929 (approved):

JoshuaKGoldberg:
👋 just checking in @manuth, anything I can help with?

manuth:
Alright, so I made a little digging.
Of course, I might be wrong with my conclusions, so I'm open to feedback.

As described in issue #4979, the underlying issue that causes Mocha to ignore certain settings from `extends`-files is, that Mocha first applies its defaults to the provided configuration and, only after that, processes the config using yargs. Let's call this behaviour "Late Yargs Processing", which is explained in detail in it's own section.

In order to fix this, yargs must be run early or more precisely:
Yargs must run ***before*** Mocha's global defaults are applied.

My solution was to run `yargs` while Mocha is loading a configuration file:
https://github.com/mochajs/mocha/blob/f131e9a6d05c425459d717c6ffc009bb5cf15770/lib/cli/config.js#L81-L83

The solution proposed by @mf-pherlihy is to process the config using `yargs` ***not*** while loading the configuration but rather ***after*** loading in the `parse` function, which is still before Mocha applies its global defaults:

https://github.com/mochajs/mocha/blob/f07d42739990bf7a51ffb418ff97751616ba9a84/lib/cli/options.js#L130-L142

### Conclusion
Of course, I might be biased as former solution is written by me while latter isn't.
I consider my solution more clean and future-proof. Especially considering that `loadRc` is an ***exported*** function and using it might re-introduce the issue if my solution weren't implemented.

Furthermore, it uses yarg's `yargs`-function which seems like "more public" API to me than the `yargs/helpers`=> `applyExtends` function (I guess?).

However, there are more...

### Things to Consider
I have just noticed that RC config files aren't the only place where the `extends` option reasonably might occur. One more spot is the `package.json` file.

I'd recommend adding a function (like `processConfig(config, filePath)` - recommendations welcome, though) to `lib/cli/options.js` which processes the config using `yargs` (which also expands the `extends` option) that we can use for both RC config files and the `package.json` file.

### Late Yargs Processing
Mocha processing the configuration using `yargs` at a late stage causes certain settings from `extends`-files to be ignored.
I will describe a case in some greater detail here.

Let's imagine a config file:
***.mocharc.json:***
```json
{
  "extends": "./base.json",
  "timeout": 1000
}
```

And base config:
***base.json:***
```json
{ "ui": "tdd" }
```

First, at this line, Mocha will apply the global defaults to the config:
https://github.com/mochajs/mocha/blob/75dcf8c6c40ed1ce134ae5e174b6f4c4ca4d8c42/bin/mocha.js#L27
or, when called through `_mocha`:
https://github.com/mochajs/mocha/blob/75dcf8c6c40ed1ce134ae5e174b6f4c4ca4d8c42/lib/cli/cli.js#L51

Making the runtime-configuration at this point look like this:

***Runtime-Config After `loadOptions`:***
```json
{
  "_": [],
  "config": false,
  "package": false,
  "extends": "./base.json",
  "timeout": 1000,
  "diff": true,
  "extension": [
    "js",
    "cjs",
    "mjs"
  ],
  "reporter": "spec",
  "slow": 75,
  "ui": "bdd",
  "watch-ignore": [
    "node_modules",
    ".git"
  ]
}
```
> ***Note:***
> Take especial notice that the `ui` setting now is already set to Mocha's global default `bdd` loaded from the `mocharc` variable here:
> https://github.com/mochajs/mocha/blob/75dcf8c6c40ed1ce134ae5e174b6f4c4ca4d8c42/lib/cli/options.js#L315-L322

After that, `yargs` is executed on the config here:
https://github.com/mochajs/mocha/blob/75dcf8c6c40ed1ce134ae5e174b6f4c4ca4d8c42/lib/cli/cli.js#L81

You will notice, that the configuration remains unchanged. That's, ofc, as you might remember from the previous Runtime-Config dump, because of the fact that `ui` is now set in the `config` object, which overrides the `ui` set in the `extends`-file. Thus, the `ui` option and more cannot be set using `extends`.

lib/cli/config.js
manuth: Closing this convo as I've addressed this in my latest comment:
https://github.com/mochajs/mocha/pull/4980#issuecomment-2569169363

Sorry for letting you wait for so long.

manuth:
So... uh... looks like I did a little oupsie in this PR:
I forgot to load the config of the `yargs` result from the `argv` property, effecitvely causing the majority of the tests not to run at all and instead printing the `mocha` version number.

I fixed this in commit a41006d497705c0031bd4e98d97084f3e88d5d8f by adding `argv` to the end of the statement.

Furthermore, as discussed in my previous comment, I added a new function `loadRcFile` which processes the configuration of a config file or `package.json`-file using `yargs`, which takes care of the `extends` option.

Now, `extends` works properly for both `package.json` and config files.

I'm ready for a review, I guess

lib/cli/options.js
JoshuaKGoldberg, line 1: > `yargs` must run _**before**_ Mocha's global defaults are applied.

ACK, that seems reasonable to me. As I understand it: Mocha-specific global defaults are meant to tack on if nothing is specified in Yargs-land. Having Yargs-land finish before Mocha-land starts seems right.

test/integration/options.spec.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('Should support extended options using rc file', function () {
+    var extended = loadOptions([]);
+    expect(extended.require, 'to equal', ['foo', 'bar']);
+    expect(extended.bail, 'to equal', true);
+    expect(extended.reporter, 'to equal', 'html');
+    expect(extended.slow, 'to equal', 30);
+  });
+
```
JoshuaKGoldberg, line 45: [Testing] When I first looked at this file I was confused that nothing ever tested the original values from `base.json`. I manually added this test case locally to confirm things still work. Any reason not to including it?

```suggestion

  it('Should support standalone options using standalone rc file', function () {
    var extended = loadOptions(['--config', configFileStandalone]);
    expect(extended.require, 'to equal', ['foo', 'bar']);
    expect(extended.bail, 'to equal', true);
    expect(extended.reporter, 'to equal', 'dot');
    expect(extended.slow, 'to equal', 60);
  });
```

...with above:

```js
const configFileStandalone = path.join(workspaceDir, 'base.json');
```

JoshuaKGoldberg (dismissed):
🔥 Whoohoo! I love how the changes in `lib/cli/options.js` are super straightforward now - just one extracted function for the two loading places.

I tried this out in the https://github.com/mochajs/mocha-examples packages that use a mocha config file and nothing was broken (that wasn't already broken...). So that's great.

I also think that because this is a breaking change for the next behavior-changing major, we'll have an alpha/beta/rc out in public for folks to try it out with. So I say 🚀 great! The only blocker now is waiting for us to get to being able to working on that major.

Marking as approved, with the notes that:

* Ideally if someone else in @mochajs/maintenance-crew has time, they should review too
* We're blocked from merging until the work on the next major gets going

Thanks @manuth! 🙌

manuth:
Awesome! Thank you so much for reviewing - and also thanks for the feedback, I really appreciate it 😄

mark-wiemer:
Adding this to my PR backlog, looks like a simple fix but I want to understand more. Additionally, v12 will be a minimal release so this'll be part of v13.

test/integration/options.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  it('Should support standalone options using standalone rc file', function () {
+    var extended = loadOptions(['--config', configFileStandalone]);
```
JoshuaKGoldberg: [blocking] (high confidence) `configFileStandalone` is never defined, so this test fails with `ReferenceError: configFileStandalone is not defined` (the file runs 4 passing / 1 failing). This one's on me - the suggestion I posted omitted the declaration, sorry! The fix that matches the assertions below (reporter `'dot'`, slow `60` from `base.json`):

```js
const configFileStandalone = path.join(workspaceDir, 'base.json');
```

CI never ran on this push, which is why nothing caught it (`no-undef` is also off in our ESLint config).

lib/cli/options.js
```diff
@@ -271,6 +273,20 @@ const loadPkgRc = (args = {}) => {
...
 module.exports.loadPkgRc = loadPkgRc;
 
+/**
+ * Loads the inherited configuration of the rc file located at the specified {@linkcode filePath}.
+ *
+ * @param {Object} config - The parsed configuration of the rc file.
+ * @param {string} filePath - The name of the file containing the parsed configuration.
+ */
+const loadRcFile = (config, filePath) => {
+  const {$0, ...options} = yargs(config._, path.dirname(filePath))
```
JoshuaKGoldberg: [blocking] (high confidence) Instantiating a full `yargs()` here re-parses `config._` as CLI argv and applies yargs' untyped value handling to every rc file and `package.json` config, `extends` or not. Concrete regressions vs the merge-base, all reproduced locally:

- `{"grep": "123"}` -> coerced to number -> `TypeError: self._grep.test is not a function`
- `{"fgrep": "1000"}` -> usage dump + `ERROR: null` (exit 1)
- `{"_": ["--bail", "extra.spec.js"]}` -> `--bail` swallows `extra.spec.js` as its value; both are lost
- `{"_": ["123"]}` -> `TypeError: arg.split is not a function` instead of our friendly unsupported-option error
- `{"--": ["x"]}` -> the `--` key is dropped and its contents moved into `_`
- `{"help": true}` -> yargs prints usage and calls `process.exit(0)` inside `loadRc` (the CLI behaved this way before too, but now the exported `loadOptions`/`loadRc` API never returns)
- a `.mocharc.js` exporting a function/array -> raw `YError` instead of a wrapped Mocha error

`applyExtends(config, path.dirname(filePath), true)` from `yargs/helpers` does only the `extends` expansion and leaves everything else byte-identical - it avoids every regression above. (Largely moot now that yargs is gone from `main`, but it's the key lesson for the reimplementation: expand `extends`, don't re-parse the config.)

lib/cli/options.js
```diff
@@ -271,6 +273,20 @@ const loadPkgRc = (args = {}) => {
 
 module.exports.loadPkgRc = loadPkgRc;
 
+/**
+ * Loads the inherited configuration of the rc file located at the specified {@linkcode filePath}.
+ *
+ * @param {Object} config - The parsed configuration of the rc file.
+ * @param {string} filePath - The name of the file containing the parsed configuration.
+ */
+const loadRcFile = (config, filePath) => {
```
JoshuaKGoldberg: [question] (high confidence on the behavior, open question on the resolution) yargs only treats `extends` values matching `.json`/`.*rc` as file paths; `.yml`/`.yaml`/`.js`/`.cjs` targets are silently ignored (the dangling `extends` key then leaks into the merged argv), and `.jsonc` crashes on yargs' raw `JSON.parse`. Since #3916 explicitly says the reporter tried js and yaml formats, should the resolution go through Mocha's own `loadConfig` recursively instead, so every config format we support can be an `extends` target? That's pre-existing yargs behavior rather than a regression, but closing #3916 while yaml/js targets stay silently broken feels like it'll generate a follow-up issue immediately.

test/integration/fixtures/config/mocharc-extended/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+    "mocha": {
```
JoshuaKGoldberg: [nit] (high confidence) This fixture is named `package-lock.json` but contains a `package.json`-style `{"mocha": ...}` config. npm, Renovate, and security scanners glob for `**/package-lock.json` expecting lockfile contents. Other fixtures use a real `package.json` in a dedicated directory (e.g. `test/integration/fixtures/config/mocha-config/package.json`) - could this follow that pattern?

test/integration/options.spec.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+var path = require('path');
+var loadOptions = require('../../lib/cli/options').loadOptions;
+
+describe('options', function () {
```
JoshuaKGoldberg: [nit] (medium confidence) This is an in-process unit test of `loadOptions` rather than an integration test (nothing spawns the mocha binary), and our integration suite runs with `--parallel`, where the `process.chdir` in `beforeEach` mutates shared worker state. `test/node-unit/cli/options.spec.js` is where the existing `loadOptions` coverage lives - moving these there (with fixtures instead of `chdir`) would also avoid the confusing name collision with the `test/integration/options/` directory and the existing `describe('options')` suite.

PR #4989, opened 2023-05-31
feat: add option to use posix exit code upon fatal signal

### LINKED ISSUE #3559: 🚀 Feature: Add option to exit with standard exit codes
### Description

Mocha's exit code is the number of failed tests (up to 255).
This is _nonstandard_ and should be more like Unix/POSIX programs.

`process.exitCode` should be one of:
* `0` for _success_
* `1` for _failure_ (e.g., internal error, if _any_ tests fail, no tests were run)
* `2` for _failure_ due to invalid command line usage
* `126` if "bin/_mocha" not executable (handled by shell)
* `128+n` where `n` is signal# (e.g., `SIGINT`=2, so rc=`130`)

where:
* `0` is EXIT_SUCCESS (POSIX)
* `1` is EXIT_FAILURE (POSIX)
* rest are shell scripting maxims (e.g., sh, ksh, [bash](http://tldp.org/LDP/abs/html/exitcodes.html)) (UNIX)

This should be _explicitly_ added to the documentation.

### Notes

Exit codes have a range from 0-255.
An exit value greater than 255 returns an exit code modulo 256.

The `126` exit code can _only_ happen on UNIX and would be provided by the shell itself.

Mocha's current scheme leads to ambiguous error codes.
For example, there is no means to distinguish between `130` test failures and interrupt by **\<Ctrl\>-C**.

It was mentioned that this _would_ certain CI systems relied on this exit code abuse.
I fail to see how this could affect them; a test failure from above would still return `1`, just not a count.

### Related

#2445
#2438 (same problem, but fix only addressed clamping #failures)

73rhodes:
### Requirements

* [x] Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* [x] All new code requires tests to ensure against regressions.

### Description of the Change
Mocha uses the number of failed tests as an exit code, which is unconventional and has led to a number of issues when trying to integrate mocha into ci/cd pipelines. 

To resolve these issues, this PR introduces a `--posix-exit-codes` boolean command-line option. If this option is specified, a fatal signal (eg. `SIGABRT`, et al) will cause the process to consistently exit with a standard posix exit code (128 + the numeric ID of the signal) when mocha is run as a child process (which is the case when passing node options).  This helps to solve issues for toolchains that expect standard posix exit codes, for example by preventing out-of-memory crashes from being silently ignored.

The [GNU libc manual page](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) provides additional context on why using the number of errors as an exit status is problematic:

> Warning: Don’t try to use the number of errors as the exit status. This is actually not very useful; a parent process would generally not care how many errors occurred. Worse than that, it does not work, because the status value is truncated to eight bits. Thus, if the program tried to report 256 errors, the parent would receive a report of 0 errors—that is, success. For the same reason, it does not work to use the value of errno as the exit status—these can exceed 255.

Providing the option to exit with standard posix shell exit codes avoids these problems and solves a number of downstream issues listed below.

### Alternate Designs

The alternatives considered were not in the scope of the mocha project.

This option is implemented in the core repository where signal handling occurs.

### Benefits

This PR provides a solution for #3559 and various downstream issues reported by mocha consumers; it preserves non-zero exit codes when mocha is spawned as a child process via various CI/CD or reporting tools and prevents silently swallowing out-of-memory errors.

### Possible Drawbacks

Introduces another option. Requires docs. Leaves non-standard exit code behavior to remain as the default.

### Applicable issues

fixes #3559 
#3893
#2445
#2438
https://github.com/istanbuljs/nyc/issues/798
https://github.com/cypress-io/cypress/issues/24695

* No breaking changes; this is an enhancement (minor release)


JoshuaKGoldberg:
We should probably disable the stale action, pending picking up reviewing PRs.

bin/mocha.js
```diff
@@ -109,7 +110,12 @@ if (mochaArgs['node-option'] || Object.keys(nodeArgs).length || hasInspect) {
   proc.on('exit', (code, signal) => {
     process.on('exit', () => {
       if (signal) {
-        process.kill(process.pid, signal);
+        if (mochaArgs['posix-exit-codes'] === true) {
+          process.exitCode = 128 + os.constants.signals[signal];
+          process.exit(process.exitCode);
```
JoshuaKGoldberg, line 115: [Question] From https://nodejs.org/api/process.html#processexitcode_1:

> Specifying a code to [`process.exit(code)`](https://nodejs.org/api/process.html#processexitcode) will override any previous setting of process.exitCode.

Is there a reason to set both?

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+var helpers = require('../helpers');
+var runMocha = helpers.runMocha;
+
+describe('--posix-exit-codes', function () {
+  // subprocess
```
JoshuaKGoldberg, line 7: [Style] Nit: No need, I think. It's implied.

```suggestion
```

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,36 @@
...
+  before(function () {
+    process.on('SIGINT', killSubprocess);
+  });
+
+  after(function () {
+    process.removeListener('SIGINT', killSubprocess);
+  });
+
+  describe('when enabled with node options', function () {
+    it('should exit with code 134 on SIGABRT', function (done) {
```
JoshuaKGoldberg, line 24: [Testing] To be thorough, could you add another test for a different signal? Just to make sure we're covering a couple, as the code should change between signals.

Also, there should be corresponding tests for what happens when `--posix-exit-codes` _isn't_ provided.

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  // these two handlers deal with a ctrl-c on command-line
+  before(function () {
+    process.on('SIGINT', killSubprocess);
+  });
+
+  after(function () {
+    process.removeListener('SIGINT', killSubprocess);
+  });
```
JoshuaKGoldberg, line 21: [Question] I don't follow, why include these two handlers? If I remove them locally this test still passes.

lib/cli/run.js
JoshuaKGoldberg, line 1: [Docs] This will need to be documented in `index.md`. https://mochajs.org/#command-line-usage have a section for each option.

JoshuaKGoldberg (changes requested):
Nice and clean implementation, thanks! ✨ 

Requesting changes on a few things. But let me know, please, if I'm off base here 🙂.

test/integration/options/posixExitCodes.spec.js
73rhodes: Unfortunately, I've lost the context on why these were originally added; it does seem like they can be removed. 👍

73rhodes: Now that I think about it, these were most likely added as a developer convenience to be able to kill the mocha process if there was a problem in the signal handling test suite that caused it to hang (eg. subprocess not responding to the signal as intended). These handlers would allow the developer to ctrl-c out of the mocha process without having to go to another shell prompt, find the PID and manually kill it.  I wanted to make sure that if this test suite didn't work as intended on someone else's operating system, they would still be able to ctrl-c out of the test suite and get back to their command line, but I think that risk is small enough that they're not essential.

73rhodes:
@JoshuaKGoldberg Thanks for the review! Hopefully that latest update addresses your questions, but let me know if there's anything else you'd like to see.   Looking froward to not using our own fork of mocha! 😆

docs/index.md
```diff
@@ -936,6 +936,10 @@ Define a global variable name. For example, suppose your app deliberately expose
 
 By using this option in conjunction with `--check-leaks`, you can specify a whitelist of known global variables that you _expect_ to leak into global scope.
 
+### `--posix-exit-codes`
+
+Exits with standard POSIX exit codes instead of the number of faled tests.
```
JoshuaKGoldberg, line 941:
```suggestion
Exits with standard POSIX exit codes instead of the number of failed tests.
```

bin/mocha.js
```diff
@@ -109,7 +110,13 @@ if (mochaArgs['node-option'] || Object.keys(nodeArgs).length || hasInspect) {
...
-        process.kill(process.pid, signal);
+        const numericSignal =
+          typeof signal === 'string' ? os.constants.signals[signal] : signal;
+        if (mochaArgs['posix-exit-codes'] === true) {
+          process.exit(128 + numericSignal);
+        } else {
+          process.kill(process.pid, signal);
+        }
       } else {
         process.exit(code);
```
JoshuaKGoldberg, line 121: [Bug] Sorry, I might have missed this in my first review: when I run the "hello world" example from https://github.com/mochajs/mocha-examples/pull/73, `signal` is `null` and `code` is the number of failed tests. Modifying the tests to fail 49 and pass 1:

```js
describe("example", () => {
  for (let i = 0; i < 50; i += 1) {
    it("2 + 2 (" + i + ")", () => {
      assert.equal(2 + 2, i);
    });
  }
});
```

...then `code` is `49` and `signal` is `null`.

Shouldn't the `posix-exit-codes` logic also handle the case of `process.exit(code)`? Am I missing something?

```suggestion
        process.exit(code + (mochaArgs['posix-exit-codes'] ? 128 : 0));
```

Note that my suggestion snippet is only part of the way there. There's also a complication of having over 127 (255 - 128) tests. That exit code should be capped to 255, I think. Fun fact: `node -e "process.exit(270)"; echo $?` reports an exit code of 14 (270 % 256).

test/integration/options/posixExitCodes.spec.js
JoshuaKGoldberg, line 1: [Testing] Can you also have tests for the proper exit code when there are test failures? Per https://tldp.org/LDP/abs/html/exitcodes.html (found via https://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux), if there are test failures, the exit should be `128` + the number of them, capped at 255, I believe?

JoshuaKGoldberg (changes requested):
Progress! Thanks for adding in the tests. I think we'll need to capture some more cases?

bin/mocha.js
73rhodes: The posix exit codes are a numerical representation of the signal that terminated the process, so if the process ran normally and wasn't pre-empted by some signal, the posix-compliant exit code of the process would be `0`.  In this case, no signal was raised, so the choice was between exiting with code of `0` or continuing with the legacy behavior of exiting with the number of failed tests. I chose to continue the legacy behavior but looking at this again now, I think the correct response would be to: 
```
process.exit(mochaArgs['posix-exit-codes'] ? 0 : code);
```

test/integration/options/posixExitCodes.spec.js
73rhodes: As per my previous comment, if no signal terminated the process and it exited normally, the posix-compliant exit code would be 0. Whether or not there a test failures, the process could exit normally or it could be terminated by a signal in some cases (out of memory error, etc) and if `--posix-exit-codes` are requested, the number of failed tests would be irrelevant to the exit code (since the posix exit code purely represents the signal that terminated the process, or zero if it exited normally). But I can add codes to cover the case where `--posix-exit-codes` is _not_ specified, and there are some test failures, to assert that the original behavior remains unchanged unless you provide the new command-line option.

docs/index.md
```diff
@@ -936,6 +936,10 @@ Define a global variable name. For example, suppose your app deliberately expose
 
 By using this option in conjunction with `--check-leaks`, you can specify a whitelist of known global variables that you _expect_ to leak into global scope.
 
+### `--posix-exit-codes`
+
+Exits with standard POSIX exit codes instead of the number of failed tests.
```
JoshuaKGoldberg, line 941: [Docs] Users will probably be curious about this:

```suggestion
Exits with standard POSIX exit codes instead of the number of failed tests.

Those exit codes are:

* `0`: if all tests passed
* `1`: if any test failed
* `128 + <signal>` if given a signal, such as:
  * 134: `SIGABRT` (`128 + 6`)
  * 143: `SIGTERM` (`128 + 15`)
```

JoshuaKGoldberg (changes requested):
Looking good! I think just the one point on `0` vs `1` to hammer out? 🙌

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 143);
+        done();
+      });
+    });
+
+    it('should exit with code 0 even if there are test failures', function (done) {
```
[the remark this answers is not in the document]
73rhodes: I think it's a good idea. 👍

lib/cli/run.js
```diff
@@ -190,6 +190,10 @@ exports.builder = yargs =>
         description: 'Run tests in parallel',
         group: GROUPS.RULES
       },
+      'posix-exit-codes': {
+        description: 'Use posix exit codes for fatal signals',
```
plroebuck, line 194: [Description is wrong](description: 'Use POSIX and shell scripting exit codes as Mocha's return value',)

plroebuck: https://itsfoss.com/linux-exit-codes/

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+var helpers = require('../helpers');
+var runMocha = helpers.runMocha;
+
+describe('--posix-exit-codes', function () {
+  describe('when enabled, and with node options', function () {
+    var args = ['--no-warnings', '--posix-exit-codes'];
```
plroebuck, line 8: var signalOffset = 128;

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+    it('should exit with code 134 on SIGABRT', function (done) {
+      var fixture = 'signals-sigabrt.fixture.js';
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 134);
```
plroebuck, line 16: expect(res.code, 'to be', signalOffset + os.constants.signals.SIGABRT);

plroebuck: Resist hardcoded values outside 0, 1 and the empty string.

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+var helpers = require('../helpers');
+var runMocha = helpers.runMocha;
+
+describe('--posix-exit-codes', function () {
+  describe('when enabled, and with node options', function () {
+    var args = ['--no-warnings', '--posix-exit-codes'];
+
+    it('should exit with code 134 on SIGABRT', function (done) {
```
plroebuck, line 10: it('should exit abnormally with POSIX shell SIGABRT code',

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 134);
+        done();
+      });
+    });
+
+    it('should exit with code 143 on SIGTERM', function (done) {
```
plroebuck, line 21: as above

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      var fixture = 'signals-sigterm.fixture.js';
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 143);
```
plroebuck, line 27: as aabove

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      });
+    });
+
+    it('should exit with code 0 even if there are test failures', function (done) {
+      var fixture = 'failing.fixture.js';
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 0);
```
plroebuck, line 38: expect(res.code, 'to be', EXIT_SUCCESS);

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    it('should exit with the number of failed tests', function (done) {
+      var fixture = 'failing.fixture.js'; // one failing test
+      var args = ['--no-warnings'];
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 1);
```
plroebuck, line 64: expect(res.code, 'to be', EXIT_FAILURE);

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', null);
+        done();
+      });
+    });
+
+    it('should exit with the number of failed tests', function (done) {
```
plroebuck, line 57: it('should exit abnormally with EXIT_FAILURE when any tests fail (or if none performed)',

test/integration/options/posixExitCodes.spec.js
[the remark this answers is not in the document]
plroebuck: it('should exit normally with EXIT_SUCCESS',

test/integration/options/posixExitCodes.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      runMocha(fixture, args, function postmortem(err, res) {
+        if (err) {
+          return done(err);
+        }
+        expect(res.code, 'to be', 0);
+        done();
+      });
+    });
+  });
+
```
plroebuck, line 43: Don't understand why L44-55 is desired -- shouldn't it just do like it (unhelpfully) did before?

It doesn't seem helpful to get a NULL return code from a signal.

plroebuck: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html

Use table above to ensure all the signals being handled are as expected.

test/integration/options/posixExitCodes.spec.js
plroebuck: btw, not sure this vs what expected processed, but signal numbers are only somewhat constant across various platforms.

https://stackoverflow.com/questions/44792204/why-are-several-signal-numbers-architecture-dependent-on-linux

https://unix.stackexchange.com/questions/155829/why-does-kill-l-not-list-signal-numbers-of-32-and-33

https://www-uxsup.csx.cam.ac.uk/courses/moved.Building/signals.pdf

test/integration/options/posixExitCodes.spec.js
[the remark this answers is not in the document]
73rhodes: I think the `normal` vs `abnormal` nomenclature was dropped from the posix nomenclature [See: https://pubs.opengroup.org/onlinepubs/009695299/functions/exit.html].

If we go with @JoshuaKGoldberg 's suggestion here, it would be `it(should exit with EXIT_FAILURE)`.  That lines up with what I was able to find on the standard use of 0 vs 1 as exit codes[*]:

  * 0: Success—Indicates that the command or program executed successfully without any errors.
  * 1: General Error—A catch-all exit code for a variety of general errors. Often used when the command or program encounters an error, but no specific exit code is available for the situation.

test/integration/options/posixExitCodes.spec.js
73rhodes: Added test case for test failures, exiting with code `EXIT_FAILURE` ( = 1 ).

test/integration/options/posixExitCodes.spec.js
73rhodes: Good call.

test/integration/options/posixExitCodes.spec.js
73rhodes: When we don't specify `--posix-exit-codes` (which might better be called `--unix-exit-codes` based on the discussion here)... we want to leave the legacy behavior unchanged. The reason why we `expect(res.code).to.be(1)` is because there's exactly one failing test in that suite.  I should probably change it to be several failing tests and a few passing ones just to avoid confusion, because the `1` here isn't meant to be synonymous with the EXIT_FAILURE (ie. 1) general error code.

test/integration/options/posixExitCodes.spec.js
73rhodes: > Don't understand why L44-55 is desired -- shouldn't it just do like it (unhelpfully) did before?
> 
> It doesn't seem helpful to get a NULL return code from a signal.

This is to assert that the existing behavior doesn't change when the new option isn't given and ensure there's no unexpected breaking change.

73rhodes: > https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html
> 
> Use table above to ensure all the signals being handled are as expected.

In this clause of the suite, the signal shouldn't matter since we're testing the behavior _without_ the `--posix-exit-codes` flag and want to ensure that mocha continues to exit with the number of failed tests. In the previous clause we test scenarios where `--posix-exit-codes` is present, but there isn't really any special handling to be done beyond ensuring that the appropriate `exit_code` is set. We have test cases there to cover a couple of common signals... adding a test for every signal in `os.constants.signals` would be overkill imho and not adding any real value since they're all simply passed on to the exit call with a +128 offset.

73rhodes: PS the null code is just the existing mocha behavior, we didn't change that her; we're just asserting that the old behavior keeps working the same way.

73rhodes:
@JoshuaKGoldberg do you want to take a look at this again? I went with your suggestion to exit with 1 like Jest and Vitest when you've asked for posix/unix shell exit codes and you've got test failures. @plroebuck raised the point that there's a distinction between posix exit codes and unix shell exit codes but I think the naming is okay given the following documentation from the [GNU libc manual page](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) on exit status:

>>> **Warning**: Don’t try to use the number of errors as the exit status. This is actually not very useful; a parent process would generally not care how many errors occurred. Worse than that, it does not work, because the status value is truncated to eight bits. Thus, if the program tried to report 256 errors, the parent would receive a report of 0 errors—that is, success.
>>>
>>> For the same reason, it does not work to use the value of errno as the exit status—these can exceed 255.
>>>
>>> **Portability note**: Some non-POSIX systems use different conventions for exit status values. For greater portability, you can use the macros EXIT_SUCCESS and EXIT_FAILURE for the conventional status value for success and failure, respectively. They are declared in the file stdlib.h.

test/integration/options/posixExitCodes.spec.js
plroebuck: > > https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html
> > Use table above to ensure all the signals being handled are as expected.
> 
> [SNIP] We have test cases there to cover a couple of common signals... adding a test for every signal in `os.constants.signals` would be overkill imho and not adding any real value since they're all simply passed on to the exit call with a +128 offset.

Fine. I didn't see where you were setting signal handlers. The table was to ensure you only did so for signals with either a T or A action (ignoring SIGKILL obviously).

> PS the null code is just the existing mocha behavior, we didn't change that her; we're just asserting that the old behavior keeps working the same way.

Perhaps that null code aspect was added when Mocha went parallel then. But all this comes back to the shell expecting an integer return -- not sure there's such a thing as a null code -- not being converted to zero (no failure), is it?

test/integration/options/posixExitCodes.spec.js
plroebuck: having multiple failures would have made this more apparent.

73rhodes: Yep, added more failures in the test fixture.

test/integration/options/posixExitCodes.spec.js
73rhodes: `process.exit(null)` results in a `0` exit code in the shell, yes.

plroebuck: Are you saying the existing behavior is to terminate with successful return code (0) when a signal handler is (abnormally) invoked?

73rhodes: Yes, exactly :-) ... that existing behavior is the root cause of the issues in the PR description, hence the need for an option to tell mocha _not_ to do this, and use a standard posix / shell exit code instead: "This helps to solve issues for toolchains that expect standard posix [/ unix shell] exit codes, for example by preventing out-of-memory crashes from being silently ignored."

73rhodes: PS. Even though the GNU libc manual warns against using exit codes to indicate the number of failed tests ("Don’t try to use the number of errors as the exit status... it does not work"), this PR isn't meant to express an opinion on whether or not mocha's current behavior is good.  From the linked issues, it seems that some tools now depend on mocha's existing behavior, and to avoid breaking those, this command line option should preserve the original behavior while providing a solution for toolchains that expect standard posix / shell exit codes.

plroebuck: Preserving so many different broken behaviors will be its own maintenance nightmare. My original was only to separate Mocha #errors from its exit code, while providing a sane alternative. I certainly understand sometimes  things need be done piecemeal to work them into the codebase, but IMHO deprecation warnings should be given for compatible-but-[borked](https://cheezburger.com/3878624512/borked) stuff like that (and the entire nonsense be removed upon the next major).

73rhodes: I agree and I think what you suggested will be a reasonable path forward as a next step if the maintainers want to make this the default behavior. For this PR, changing the default behavior would be out of scope. Any remaining concerns on this code section in particular?

JoshuaKGoldberg:
Will take a look, thanks!

test/integration/options/posixExitCodes.spec.js
73rhodes: I can go ahead and remove this test case - in a way it's ensconcing a behavior that probably should never have existed and would make it just a bit harder to fix that behavior going forward.

docs/index.md
```diff
@@ -936,6 +936,18 @@ Define a global variable name. For example, suppose your app deliberately expose
 
 By using this option in conjunction with `--check-leaks`, you can specify a whitelist of known global variables that you _expect_ to leak into global scope.
 
+### `--posix-exit-codes`
```
JoshuaKGoldberg, line 939: From https://github.com/mochajs/mocha/pull/4989#discussion_r1566158084:

> (which might better be called `--unix-exit-codes` based on the discussion here)

Yeah, the requested behavior has evolved a good bit. I wonder if a better name would be `--standard-exit-codes`? That way the word "standard" is intentionally a little ambiguous, to match how this adheres more to the general community standard of how test runners should exit, rather than the precise posix or unix standards...

This isn't a blocker for me. The long-term plan is to make this the default in a future Mocha version anyway. So I'll happily defer to you & the rest of the team. 😄

JoshuaKGoldberg (approved):
This looks wonderful to me, thanks so much for all the discussion and iteration @73rhodes! 🙌 

Marking this as having one team member approval. We'll want another team member to take a look too. Ping @voxpelli, IIRC you were particularly interested in making Mocha more standards-compliant?

docs/index.md
```diff
@@ -936,6 +936,18 @@ Define a global variable name. For example, suppose your app deliberately expose
...
+Exits with standard POSIX exit codes instead of the number of failed tests.
+
+Those exit codes are:
+
+* `0`: if all tests passed
+* `1`: if any test failed
+* `128 + <signal>` if given a signal, such as:
+  * 134: `SIGABRT` (`128 + 6`)
+  * 143: `SIGTERM` (`128 + 15`)
+
```
73rhodes, line 950:
```suggestion
- `0`: if all tests passed
- `1`: if any test failed
- `128 + <signal>` if given a signal, such as:
  - 134: `SIGABRT` (`128 + 6`)
  - 143: `SIGTERM` (`128 + 15`)

```

docs/index.md
73rhodes: I've no strong opinion on the naming and can change it if that's needed to get final approval, but I'll leave it as-is for now considering that the gnu docs I referenced mention this kind of exit code is standard for posix systems... lmk if I should change it though.

73rhodes: Marking as resolved since there seems to be no strong objection to the naming.

73rhodes:
@JoshuaKGoldberg just circling back to see if there's anything you need from me to unblock this enhancement? lmk 🙏

JoshuaKGoldberg:
Thanks for the ping! Nothing yet. The other maintainers have been swamped (so much happening in the ecosystem!) but I'm still holding onto hope someone will be able to review. I've personally time boxed this one to within June.

voxpelli (dismissed):
As far as I can see this only changes the flow in this flow:

https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/bin/mocha.js#L80

Not in the other flow:

https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/bin/mocha.js#L139-L141

Is this really the right place for this fix?

Eg: The non-standard error codes are set here:

https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/lib/cli/run-helpers.js#L26-L45

And in the old-school case where one forces the process to be killed once all tests are done it happens within that:

https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/lib/cli/run-helpers.js#L52

---

I overall find this PR to be a mix of two changes:

1. Stop mocha from abusing the exit code as an error count reporter
2. Add signal specific exit codes

The latter of the two I would want to read up on more, but the first of them should be an easier fix.

voxpelli:
Node.js should already be setting the signal exit codes, I wonder if we can piggy back on that instead? https://nodejs.org/api/process.html#exit-codes

73rhodes:
@voxpelli Thanks for the review. That's right, the option only applies "when mocha is run as a child process". The reasoning was to try and be as unobtrusive as possible and address specifically the problem of fatal errors (OOM et al) getting swallowed silently in a CI/CD pipeline where mocha is spawned by something like nyc.  It sounds like there's general agreement that mocha shouldn't use the exit code to report the number of errors though, so I can update this PR to support the option in both cases. As for exit codes other than 0, 1 and 128+signal - I guess we'd need to know what the exit code scheme should be and could be something for another PR....

voxpelli:
@73rhodes Is there a need to handle 128+signal codes manually or is that already [handled by node.js itself](https://nodejs.org/api/process.html#exit-codes)? Reason why I ask is since its typically not something I have seen handled manually by other CLI tools built with node.js

73rhodes:
@voxpelli ... I just updated the PR with a commit ([this PR](https://github.com/zendesk/sunco-mocha/pull/11)) that adds the signal-handling behavior for both the child-process and in-process modes of running mocha, as requested.  As far as piggybacking on node's signal handling, we're basically working with these inputs:

```
proc.on('exit', (code, signal) => {
```
and 
```
const exitMochaLater = code => {
```
where, based on what I've seen and implemented in the tests, we need to check the `code` and `signal` and decide on the desired exit code ourselves. Maybe it's possible that at some deeper layer of mocha internals there's an alternative way to implement this, but this is the most intuitive way I managed to come up with...  and it's been working in production for quite a while...

73rhodes:
Just circling back here to see if there are any specific changes being requested here to move forward. We would prefer to use the original mocha package but proper signal handling is critical for us. Please let me know if there's anything I can do to unblock this, thanks!

voxpelli:
For me I’m still at:

> I overall find this PR to be a mix of two changes:
> 
> * Stop mocha from abusing the exit code as an error count reporter
> * Add signal specific exit codes
> 
> The latter of the two I would want to read up on more, but the first of them should be an easier fix.

And I’m also torn on shipping this as an option rather than shipping it as a breaking change in the next major – I would like Mocha to gain fewer options and less complex code base, not the opposite

bin/mocha.js
```diff
@@ -109,7 +113,15 @@ if (mochaArgs['node-option'] || Object.keys(nodeArgs).length || hasInspect) {
   proc.on('exit', (code, signal) => {
     process.on('exit', () => {
       if (signal) {
-        process.kill(process.pid, signal);
+        const numericSignal =
+          typeof signal === 'string' ? os.constants.signals[signal] : signal;
```
voxpelli, line 117: According to the documentation the `signal` argument of the `exit` event on `child_process.spawn()` is [always a string](https://nodejs.org/api/child_process.html#event-exit) – when is it ever something else?

bin/mocha.js
```diff
@@ -109,7 +113,15 @@ if (mochaArgs['node-option'] || Object.keys(nodeArgs).length || hasInspect) {
...
+        if (mochaArgs['posix-exit-codes'] === true) {
+          process.exit(SIGNAL_OFFSET + numericSignal);
+        } else {
+          process.kill(process.pid, signal);
+        }
```
voxpelli, line 122: If the `process.kill(process.pid, signal)` is important because it will trigger handles like the `process.on('SIGINT'`, then that's what always should be used here.

Else the `process.exit()` variant should always be used.

Though the logic of Mocha here is kind of weird – the `signal` comes from the child process, and the `SIGNAL_OFFSET + numericSignal` should be sent if the process itself, not one of its childs, gets sent the `signal` – so I would say that `process.exit(SIGNAL_OFFSET + numericSignal)` here is wrong and that the original `process.kill(process.pid, signal)` is weird (as its essentially bubbling up the process signal? Is that common practice?)

bin/mocha.js
```diff
@@ -109,7 +113,15 @@ if (mochaArgs['node-option'] || Object.keys(nodeArgs).length || hasInspect) {
...
+      } else if (code !== 0 && mochaArgs['posix-exit-codes'] === true) {
+        process.exit(EXIT_FAILURE);
       } else {
         process.exit(code);
```
voxpelli, line 126: Could probably make it this:

```suggestion
      } else {
        process.exit(Math.min(code, mochaArgs['posix-exit-codes'] ? 1 : 255));
```

That reuses the `Math.min(code, 255)` logic that's in other places.

Not sure the `EXIT_FAILURE` helps much here, the direct `1` makes it possible to use the `Math.min` like this

lib/cli/run-helpers.js
```diff
@@ -25,7 +26,10 @@ const PluginLoader = require('../plugin-loader');
  */
 const exitMochaLater = code => {
   process.on('exit', () => {
-    process.exitCode = Math.min(code, 255);
+    const usePosixExitCodes = process.argv.includes('--posix-exit-codes');
+    const exitCode =
+      usePosixExitCodes && code > 0 ? EXIT_FAILURE : Math.min(code, 255);
+    process.exitCode = exitCode;
```
voxpelli, line 32:
```suggestion
    process.exitCode = Math.min(code, process.argv.includes('--posix-exit-codes') ? 1 : 255);
```

lib/cli/run-helpers.js
```diff
@@ -37,7 +41,9 @@ const exitMochaLater = code => {
  * @private
  */
 const exitMocha = code => {
-  const clampedCode = Math.min(code, 255);
+  const usePosixExitCodes = process.argv.includes('--posix-exit-codes');
+  const clampedCode =
+    usePosixExitCodes && code > 0 ? EXIT_FAILURE : Math.min(code, 255);
```
voxpelli, line 46:
```suggestion
  const clampedCode = Math.min(code, process.argv.includes('--posix-exit-codes') ? 1 : 255);
```

voxpelli (commented):
One thing that making this a breaking change instead would do is that instead of still having `exitMochaLater`, `exitMocha`, `proc.on('exit'` be getting sent the number errors, that could be cleaned up.

And the current setup of a `process.on('exit'` being set in a `proc.on('exit'` and then doing a `process.kill` on itself – its all very messy and hard to follow and rather than adding more if-statements and more paths to it I would vote for trying to make it less messy.

@JoshuaKGoldberg shared with me how eg. `vitest` handles some of this, and its much much cleaner and easier to follow: https://github.com/vitest-dev/vitest/blob/f9a628438a5462436b59dd9bdeffddada19a9e81/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts#L191

voxpelli:
Moving from a confusing setup like:

```js
  proc.on('exit', (code, signal) => {
    process.on('exit', () => {
      if (signal) {
        process.kill(process.pid, signal);
      } else {
        process.exit(code);
      }
    });
  });
```

To something like:

```js
  proc.on('exit', (code, signal) => {
    process.on('exit', () => {
      if (signal) {
        const numericSignal =
          typeof signal === 'string' ? os.constants.signals[signal] : signal;
        if (mochaArgs['posix-exit-codes'] === true) {
          process.exit(SIGNAL_OFFSET + numericSignal);
        } else {
          process.kill(process.pid, signal);
        }
      } else if (code !== 0 && mochaArgs['posix-exit-codes'] === true) {
        process.exit(EXIT_FAILURE);
      } else {
        process.exit(code);
      }
    });
  });
```

It doesn't really make it easier to consume 🤔

voxpelli:
And for reference: The original introduction of that listener in a listener happened 12 years ago in https://github.com/mochajs/mocha/pull/571 with the PR description:

> relay signals from the "_mocha" process to the main "mocha" bin
> 
> This makes is so that if your test case cases node to crash with a SIGSEGV
or SIGBUS for example, than "mocha" will have a proper exit code (instead of 0).

And might have been somewhat invalidated by #3827 which makes it so that mocha does `not fork if no node flags present`.

JoshuaKGoldberg:
👋 Ping @73rhodes, do you have any thoughts on VoxPelli's review?

73rhodes:
@JoshuaKGoldberg Sorry for the slow response. So this is still an issue preventing us from using the main mocha distribution, and yes we'd still like to stop using a fork.  This came up again recently, since we seem to be having some issues with mocha in a new environment and now need to go through the whole process of rebasing, which we'd really rather avoid and just use the main distro.  I can address the PR review comments if you reopen this.

voxpelli:
@73rhodes Reopening for you :)

lib/cli/run-helpers.js
```diff
@@ -16,6 +16,7 @@ const {format} = require('util');
 const {createInvalidLegacyPluginError} = require('../errors');
 const {requireOrImport} = require('../nodejs/esm-utils');
 const PluginLoader = require('../plugin-loader');
+const EXIT_FAILURE = 1;
```
73rhodes, line 19:
```suggestion
```

bin/mocha.js
73rhodes: It can be passed via `process.kill` as a string or number [[node docs](https://nodejs.org/docs/latest/api/process.html#processkillpid-signal)] and although I don't recall the exact circumstances, I did see both cases during development, and that's why I needed to add this check. If I find the steps to reproduce it I'll add that as a comment.

73rhodes:
> It doesn't really make it easier to consume 🤔

I've refactored this a bit to simplify. I avoided compressing this into the smallest possible number to aid readability.

The test criteria are unchanged and can be run as:
```
bin/mocha.js test/integration/options/posixExitCodes.spec
```

bin/mocha.js
73rhodes: I believe exit has the benefit of explicitly setting the exit code instead of explicitly stating `process.exitCode = <code>`, but I've refactored this to simplify a bit.

bin/mocha.js
73rhodes: To demonstrate how `kill` can be used to send either a string or numerical representation of a signal that must be properly handled, I've added a test case to cover the scenario of `kill` sending a numeric signal.

bhavikpatel-zd (approved):

73rhodes:
> And I’m also torn on shipping this as an option rather than shipping it as a breaking change in the next major – I would like Mocha to gain fewer options and less complex code base, not the opposite

I appreciate the need to reduce command line options. Ideally, this posix-compliant behavior will be the default at some point. Adding it as an option for now will allow folks to begin using and experimenting with it, prior to making it the default behavior as a breaking change in a future release. I believe this is how node has introduced certain experimental features as well.

mark-wiemer:
Looks excellent @73rhodes, subscribing to help review :)

mark-wiemer (approved):

mark-wiemer:
@mochajs/maintenance-crew and @voxpelli as a previous change requestor. This looks good to me, I see the concerns about more complex code but I agree with the Darren (PR author) that adding this as an option early will allow folks to experiment with it before it becomes a breaking change. Plus if there are bugs in this approach we can find them now rather than before a major version upgrade. I'm in favor of merging this now that tests pass and the code looks clean enough to me (adding utility functions for DRYness is my only thought, but that almost makes it less readable in this case)

voxpelli (commented):
I don’t want to be a blocker for this, if @mark-wiemer thinks this is good to go then I’m okay with that :)

JoshuaKGoldberg (approved):
🚀

test/integration/helpers.js
```diff
@@ -358,6 +359,18 @@ function createSubprocess(args, done, opts = {}) {
...
+   * Emulate node's exit code for fatal signal. Allows tests to see the same
+   * exit code as the mocha cli.
+   */
+  mocha.on('exit', (code, signal) => {
+    if (signal) {
+      mocha.exitCode =
+        SIGNAL_OFFSET +
+        (typeof signal == 'string' ? os.constants.signals[signal] : signal);
+    }
+  });
```
JoshuaKGoldberg, line 372: Just noting for future reference: I was confused at first that `exitCode` isn't newly referenced elsewhere. It's some built-in process plumbing. A couple of integration tests fail if you comment this out locally.

73rhodes: Yeah, this was a bit puzzling, but for mocha self-testing IIRC it was necessary to mimic the way node's exit codes work (which is already posix compliant). I'll resolve this thread, but if you want additional detail here in addition to what is already in the doc block above, please let me know.

mark-wiemer:
I was very careful to make sure this PR actually passed tests, but I clearly misread the report. Fortunately, the only test that failed is a known flaky issue on main branch that I haven't formally reported yet. I'm busy this weekend but planning to report it on Tuesday.

PR #4995, opened 2023-07-08
fixes #4994 - Preserves error.name during serialization process

### LINKED ISSUE #4994: error.name missing when errors are serialized when running tests in parallel - Causes test results to be incorrectly reported in allure-mocha
### Prerequisites

- [X] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend that you _not_ install Mocha globally.

### Description

Some reporters use the type of error to determine how to treat a test that fails.  Allure is one example.  If a test fails due to an AssertionError, then the test is marked in red, as failed.  All other errors are treated as broken tests.  These results are marked in yellow, as broken.

When running mocha in parallel, the error objects are serialized and then later deserialized.  During the serialization process, we are preserving the `error.message` but not the `error.name`.  Because of this, it's impossible for Allure to determine the type of error by using the error.name property.

### Steps to Reproduce

**To Reproduce**
1. Clone this repo: https://github.com/jamesmortensen/allure-js-reporter-issues#when-run-in-parallel-mode-with-hooks-hookerror-is-not-a-function and follow the instructions to run the tests and generate the report.

The example has us run tests in sequential mode as well as parallel mode.  The README also has screenshots which show the issue.

**Expected behavior**
- For AssertionErrors, results should appear in the report in RED.  

**Actual behavior**
- In parallel mode, they're reported in YELLOW.

**Screenshots**

- Good (shows red):  https://github.com/jamesmortensen/allure-js-reporter-issues/blob/master/screenshots/sequential-mode-report.png
- Bad (shows yellow):  https://github.com/jamesmortensen/allure-js-reporter-issues/blob/master/screenshots/parallel-mode-report.png

**Reproduces how often:** 

Reproduces 100% of the time.

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`:  10.2.0
- The output of `node --version`:  18.3.0
- Your operating system
  - name and version:  macOS Ventura 13.3.1
  - architecture (32 or 64-bit):  arm64
- Your shell (e.g., bash, zsh, PowerShell, cmd):  bash
- Your browser and version (if running browser tests):  N/A
- Any third-party Mocha-related modules (and their versions):  allure-mocha v2.4.0
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version):  N/A

**Additional information:**

Issue originally reported here:  https://github.com/allure-framework/allure-js/issues/485 but I traced the root cause to Mocha.

jamesmortensen:
### Requirements

### Description of the Change

When error objects are serialized, the name is stripped from the object. The name contains the type of the error, whether it's an AssertionError, TypeError, ReferenceError, or just Error.  Some test reporters use the type of the error to determine how to process the failure.  For example, Allure treats AssertionErrors as failures and all other errors as broken tests.

The change involved modifying serializer.js so that when the error object is serialized, it includes the name property so that when it's later deserialized, the name is still present.  Currently, only the stack and message properties were preserved.

### Alternate Designs

One other alternative would have been to attempt to patch this in the reporter directly by extracting the error type from the stack trace.  This seems messy and would have involved adding conditional logic to determine whether mocha is running in parallel or sequentially.  Since it shouldn't matter -- the results should be the same way no matter what -- I decided it would be best to make the change upstream, in mocha.  This also allows other reporters to capitalize on this change.

Since no name property existed in mocha, this should be a non-breaking change.  

Sending test results to reporters is a core function of mocha.  What to do with that data is of course the responsibility of the reporter, but it's mocha's responsibility to make sure there's consistency in how the data is being sent to the reporters.  When running tests sequentially, the error object contains the name property. I believe this happens because no serialization is required.  Reporters receive this information when running sequentially.

But when running in parallel, Mocha's behavior of how errors are passed to reporters changes.  Important information, such as error type, is not included in the name property.  Including it in the name property creates consistency for those developing reporters.

### Benefits

**Users of Allure reporter can run mocha in parallel mode.**

Since Mocha released parallelism in version 8.0.0, users of allure were not able to capitalize on the feature without complicated hacks like this:  https://github.com/allure-framework/allure-js/issues/245#issuecomment-1176254785.  Fixing this issue makes Allure more usable.

**Maintenance of reporters using the error object is easier.**

Hacking the stack object would have been cool, but it might have been more difficult to understand or maintain than changing one line of code in mocha, which represents an additive change, not a breaking change.

### Possible Drawbacks

The only thing I can think of is if some reporter was, for whatever reason, looking for the error.name to be undefined, and it ended up being empty string or null, that could potentially impact the behavior of their reporter.  However, I cannot imagine why someone would do that.  Additionally, if the error.name is already undefined, then the name property should continue to be undefined.  I do not see this as being high risk change.

### Applicable issues

- I lean more towards considering this as a patch update, a "bug fix", since it corrects inconsistent behavior between singular mode and parallel mode.  
- Issue was originally reported here:  https://github.com/allure-framework/allure-js/issues/485

### DIFF
--- a/lib/nodejs/serializer.js
+++ b/lib/nodejs/serializer.js
@@ -204,7 +204,8 @@ class SerializableEvent {
       value = Object.assign(Object.create(null), value, {
         stack: value.stack,
         message: value.message,
-        __type: 'Error'
+        __type: 'Error',
+        name: value.name
       });
       parent[key] = value;
       // after this, set the result of type(value) to be `object`, and we'll throw


jamesmortensen:
Hi.  I wanted to check in on the status of this pull request?  Please let me know if there are any other details you need in order to evaluate this small change.  Thank you!

JoshuaKGoldberg:
Marking as draft and blocked on an isolated reproduction for #4994. cc @jamesmortensen

JoshuaKGoldberg:
Closing out as #4994 was closed out too.

PR #5001, opened 2023-07-21
fix: ensure that the xunit reporter emits well-formed XML

malloryavvir:
Edit: I have updated this PR to fix the bug in addition to creating a failing test.

### Description of the Change

From the PR:

    XML does not allow certain characters to appear in documents. This
    change ensures that those restricted characters are replaced by legal
    characters.

    This bug means that when a test ends up containing, for example, ANSI
    color codes (i.e. `&#x1B;[36m&#x3C;div&#x1B;[39m`) then the generated
    XML file is invalid and some consumers (such as CircleCI) are unable to
    parse the contents.

    Link: https://www.w3.org/TR/2006/REC-xml11-20060816/#charsets

### Alternate Designs

This is a lossy conversion. An alternative to displaying the replacement character would be to display some message explaining what character has been replaced. This is a lot less verbose and will be easier to read when your tests are outputting ANSI control codes, and if the original text is needed you can rerun the tests locally.

This is a bug. Mocha is generating malformed XML documents that cannot be parsed.

### Benefits

XML documents generated by the XUnit reporter are now well-formed and thus can be parsed.

### Possible Drawbacks

It's a lossy conversion. Probably 99% of the time it's going to be an ANSI control code being removed.

### Applicable issues

This is a bug fix.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -15,6 +15,7 @@ var util = require('util');
 var he = require('he');
 
 const MOCHA_ID_PROP_NAME = '__mocha_id__';
+const RESTRICTED_CHAR = /[\x01-\x08\x0B-\x0C\x0E-\x1F\x7F-\x84\x86-\x9F]/g;
 
 /**
  * Inherit the prototype methods from one constructor into another.
@@ -35,6 +36,7 @@ exports.inherits = util.inherits;
  * @return {string}
  */
 exports.escape = function (html) {
+  html = html.toString().replaceAll(RESTRICTED_CHAR, "�");
   return he.encode(String(html), {useNamedReferences: false});
 };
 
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -35,6 +36,7 @@ exports.inherits = util.inherits;
  * @return {string}
  */
 exports.escape = function (html) {
+  html = html.toString().replaceAll(RESTRICTED_CHAR, "�");
```
JoshuaKGoldberg, line 39: 🤔 this looks like a duplicate of #4527. @malloryavvir do you have thoughts on https://github.com/mochajs/mocha/pull/4527/files#r1511131215 ?

JoshuaKGoldberg:
👋 ping @malloryavvir, is this still something you have time for?

malloryavvir:
@JoshuaKGoldberg I've been pretty busy lately. I'm afraid I don't have the capacity to look at this. Thanks for reaching out though.

JoshuaKGoldberg:
No worries, thanks for letting us know! I'll close this out so that anybody who has time + energy can approach it. 

If a PR is sent based on code form this one, it should include a co-authored-by attribution.

Cheers!

PR #5015, opened 2023-10-12
feat: use <progress> and <svg> for browser progress indicator instead of <canvas>

### LINKED ISSUE #5113: 🚀 Feature: Switch from <canvas> to native <progress> element in browser reporter
### Feature Request Checklist

### Overview

Filing a backing issue for #5015: right now, the Mocha browser reporter uses a `<canvas>` element to visualize a progress circle. That's a holdover from when the reporter was written long ago and it was hard to do progress visualizations in CSS.

### Suggested Solution

https://github.com/mochajs/mocha/pull/5015#discussion_r1511451520: Now, many years later, the [<progress> element is supported in all major browsers](https://caniuse.com/progress). Even IE 10! (which Mocha no longer supports). Proposal: let's remove the canvas altogether, and switch to a full HTML/CSS `<progress>`?

### Alternatives

Maintaining two code paths for the visualization is an extra maintenance burden. Not a fan of the idea.

### Additional Info

_No response_

yourWaifu:
### Description of the Change

It's possible for HTMLCanvasElement to not be available. In the HTML reporter, if canvas isn't available, there's no fallback to see the progress percentage.

here's a picture:
image

### Alternate Designs

Using CSS flex-basis to create a progress bar, this is good alternative and can also work. However, to ensure compatibility, I decided to use text as the fallback because I know it'll work.

Seems to be a bug that makes it very hard to see the progress of test.

### Benefits

The progress can be seen regardless if canvas is available

### Possible Drawbacks

Very niche, 99% of cases, canvas is supported

### Applicable issues

Fixes #5113.


yourWaifu:
I'll ask my employer about this easyCLA thing

lib/reporters/html.js
```diff
@@ -84,6 +85,12 @@ function HTML(runner, options) {
     ctx = canvas.getContext('2d');
     ctx.scale(ratio, ratio);
     progress = new Progress();
+  } else {
+    // On some broswers, canvas might be unavailable for whatever reason.
+    // As such, we need a text version as a fallback
+    var progressTextFallback = fragment('<li class="progress-text">progress: <em>0</em>%</li>');
+    progressText = progressTextFallback.getElementsByTagName('em')[0];
+    items[0].replaceWith(progressTextFallback);
```
JoshuaKGoldberg, line 93: [Refactor] It's not great to have to maintain two code paths for the same feature. Also, the [`<progress>` element is supported in all major browsers](https://caniuse.com/progress). Even IE 10! (which Mocha no longer supports)

Proposal: let's remove the canvas altogether, and switch to a full HTML/CSS `<progress>`?

JoshuaKGoldberg (changes requested):
I like this direction! In fact, I'd even say let's go with a full switch, rather than a fallback? 🔥

yourWaifu:
Sure, I can do that and say that it's not work for my previous employer. They don't support canvas very well, but progress element is supported, so this does sound like a better solution.

yourWaifu:
Update Screenshot: Edge top left, Firefox bottom left, Webkit right
image

lib/reporters/html.js
yourWaifu: Should I also be removing progress.js?

JoshuaKGoldberg: If the code is no longer used, then 🔪 let's remove it!

JoshuaKGoldberg:
Just to be explicit: I don't think we should change the visual design in this PR. We should keep to the same progress visualization as was being done with canvas.

yourWaifu:
Keeping both the <progress> element and the design will be tricky, but possible with multiple CSS transforms and progress elements to warp the bars into a circle. However, changing the rounded ends of the bar might need to use nonstandard CSS to achieve. We could instead replace <progress> element with HTML elements with JS to set CSS gradients to show progress. However, doing that would move up the browser version requirements: https://caniuse.com/?search=conic-gradient

JoshuaKGoldberg:
An alternative could be to use an `<svg>` to get access to more fancy UI features while visually hiding the `<progress>`.  A quick Google for _svg radial progress bar_ found me https://stackoverflow.com/questions/66990496/simple-svg-css-progress-circle and https://stackoverflow.com/questions/72676042/set-svg-progress-bar-percentage-in-javascript. [`<clipPath>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/clipPath) might be useful too.

As for the rounded bar ends, where do you see those? This is what I get locally:

https://github.com/mochajs/mocha/assets/3335181/57209fb1-74ed-4939-a2de-ef5892374960

Note the lack of animation or rounded-ness.

yourWaifu:
Now using SVG for the progress ring
image

JoshuaKGoldberg:
Just a heads up @yourWaifu, I merged the latest `master` into this branch. You'll want to `git pull` locally to bring those changes in.

mocha.css
```diff
@@ -334,20 +334,47 @@ body {
...
+  text-align: center;
+  position: absolute;
+  width: 40px;
+  display: block;
+  top: 11px; /* padding */
+}
+
+#mocha-stats .progress-ring {
+  width: 40px;
+  height: 40px;
```
JoshuaKGoldberg, line 361: [Style] Later down there are comments like `/* half of width */` to indicate the relationships between values. How about using CSS variables to make the relationships explicitly stated in code?

```suggestion
  --ring-size: 40px;
  width: var(--ring-size);
  height: var(--ring-size);
```

...and later down, like this:

```css
  cx(calc(var(--ring-size) / 2);
```

Come to think of it, I bet `--ring-size` could actually be put in the `#mocha-stats` so that `.progress-element`, `progress-text`, etc. can use it.

mocha.css
```diff
@@ -334,20 +334,47 @@ body {
...
+  fill: var(--mocha-bg-color);
+  stroke-width: 2px;
+}
+
+#mocha-stats .ring-whole {
+  stroke: var(--mocha-stats-color);
+}
+
+#mocha-stats .ring-highlight {
+  stroke: var(--mocha-stats-em-color);
```
JoshuaKGoldberg, line 377: [Bug] I like that you're using a CSS variable here 😄 but this isn't the same color as before. The deleted `Progress.prototype.draw` shows which colors should be used in dark and light mode: . `--mocha-stats-em-color` is `#000` in light mode and `#fff` in dark mode.

I don't think all those color matchings are represented in existing CSS variables. You might need to make some new one(s).

lib/reporters/html.js
```diff
@@ -38,7 +37,7 @@ exports = module.exports = HTML;
 
 var statsTemplate =
   '<ul id="mocha-stats">' +
-  '<li class="progress"><canvas width="40" height="40"></canvas></li>' +
+  '<li class="progress-contain"><progress class="progress-element" max="100" value="0"></progress><svg class="progress-ring"><circle class="ring-whole"/><circle class="ring-highlight" stroke-dasharray="0%,100%"/></svg><div class="progress-text"><em>0</em>%</div></li>' +
```
JoshuaKGoldberg, line 40: > `<em>`

[Bug] [`<em>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em) indicate _emphasis_. I don't think that's the right semantic implication here. I'm not 100% sure about this, so if you did choose it intentionally I'd love to know why 😄.

Either way, unintended change: the text wasn't italic before, and is getting `font-style: italic` by default now. The old progress bar didn't italicize its text, so the new progress bar shouldn't either.

JoshuaKGoldberg (changes requested):
🙌 Progress! This is getting pretty close to merge, I think. Thanks for all the iteration!

Requesting changes on a few things that changed unintentionally. Please make sure the new visual thing is as close to the same as the old one as is straightforward to go for, including approximate border widths, colors, and font styles.

Once you think it's ready for re-review please re-request my review (as in, with the [re-request button on GitHub](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#re-requesting-a-review)) and I'll be excited to take another look.

lib/reporters/html.js
yourWaifu: I can't remember exactly, but from what I remember, it's to match with the text values in the template.

lib/reporters/html.js
```diff
@@ -38,7 +37,7 @@ exports = module.exports = HTML;
 
 var statsTemplate =
   '<ul id="mocha-stats">' +
-  '<li class="progress"><canvas width="40" height="40"></canvas></li>' +
+  '<li class="progress-contain"><progress class="progress-element" max="100" value="0"></progress><svg class="progress-ring"><circle class="ring-whole"/><circle class="ring-highlight" stroke-dasharray="0%,100%"/></svg><div class="progress-text">0%</div></li>' +
```
JoshuaKGoldberg, line 40: Comparing visual changes:

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Progress circle with 11px off-center font
</td>
<td>
Progress circle with 12px font
</td>
</tr>
</tbody>
</table>

Ignoring that the text was slightly off-center before, I still see two unintended visual changes:
* The font size looks like it went from `11px` to `12px`
* The stroke thickness doubled

Let me know if you need help wrestling with the CSS to get those changes undone.

JoshuaKGoldberg (changes requested):
Cool! I think we can simplify the code here a bit more, and get the visual changes reduced. Progress! 🚀

mocha.css
```diff
@@ -334,20 +343,53 @@ body {
...
+  cy: var(--ring-radius);
+  r: calc(var(--ring-radius) - var(--stroke-thickness)); /* radius - stroke */
+  fill: hsla(0, 0%, 0%, 0);
+  stroke-width: calc(var(--stroke-thickness) * 2);
+}
+
+#mocha-stats .ring-whole {
+  stroke: var(--mocha-progress-ring-color);
+  stroke-width: calc(var(--stroke-thickness) * 1.8);
+  /* slightly smaller to fix strange edge issue */
```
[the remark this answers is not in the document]
yourWaifu: To be sure, I'm not going crazy. There's seems to be a difference between rendering both rings, and only parts to make a whole ring.

 strange-edge-issue
Left to right: rendering both, rendering one, difference, difference amplified

I have come up with a different solution then the above, and it's simple, render both as parts that make a whole ring. I doubt many would notice, but it was bothering me.

lib/reporters/html.js
yourWaifu: Is this close enough? 
difference
left to right: before with canvas, after with SVG, difference, difference text removed.

The color should be the same, looking at code, but it seems to appear a bit different.

lib/reporters/html.js
```diff
@@ -234,8 +218,22 @@ function HTML(runner, options) {
...
+      text(
+        progressText,
+        percent.toFixed(Math.min(Math.max(decmalPlaces, 0), 100)) + '%'
+      );
+    }
+    if (progressRing) {
+      var radius = 19; // to do, figure out how to match with css
+      var wholeArc = Math.PI * 2 * radius;
+      var highlightArc = percent * (wholeArc / 100);
+      progressRing.style['stroke-dasharray'] = `${highlightArc}px,${wholeArc}px`;
```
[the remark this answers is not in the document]
yourWaifu: Yeah, I noticed this too, but I was unsure if this is part of the standard or not. It does work on all modern browsers from testing. I decided to use getComputedStyle and getPropertyValue as an alternative.

lib/reporters/html.js
JoshuaKGoldberg: LGTM, thanks!

Yeah the little differences are ... close enough. And I like the less pixellated circle 😄

mocha.css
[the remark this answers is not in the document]
JoshuaKGoldberg: Ha, yeah, nice spot - and good idea on the difference & amplified images! I think teeny pixel differences like this one are fine.

JoshuaKGoldberg (approved):
Whoohoo, thanks for this @yourWaifu! I really appreciate you iterating with me on it 🤝. The code is looking clean and the page better supports dark mode too. Plus this will help us avoid annoying the `canvas` package dependency in developing Mocha. A win all around!

Zach Galifianakis flying through the air with a stern look on his face. Caption: "#WINNING"

JoshuaKGoldberg:
This was released in [`mocha@10.5.0`](https://www.npmjs.com/package/mocha/v/10.5.0). 🚀

PR #5017, opened 2023-10-17
feat: added optional jsw format to json reporter

### LINKED ISSUE #4889: JSON reporter with no formatting

The problem is the Mocha JSON reporter is hardcoded to output the JSON object over multiple lines. This makes it impossible to use with Splunk which requires the object to just be on one line.

This is the line of code I am referring to: https://github.com/mochajs/mocha/blob/master/lib/reporters/json.js#L81

```
...
var json = JSON.stringify(obj, null, 2);
...
```

I would like the spacing in the JSON reporter output to be an option so we can configure this as required.

We could add a option to the reporter which allows the user to set the whitespace, but defaults to the current value if it's not provided so the change is backwards compatible. 

For example, the line I linked above would be changed to something like this:

```
var json = JSON.stringify(obj, null, options.jsonStringifyWhitespace ?? 2);
```

An alternative is forking the code and creating my own reporter, but it is such a small change and I believe the community would benefit from this extra option so I would rather make the change here.

CheadleCheadle:
### Requirements

### Description of the Change
Added an optional CLI argument ('jsw' or 'jsonStringifyWhitespace') that allows for the user to set the whitespace of the JSON reporter. If it isn't provided it defaults to the current value (2).

This change allows for more customization of the JSON output!
Closes #4889

### Alternate Designs
An alternative approach might involve using a boolean CLI argument to toggle the JSON reporter and remove whitespace. However, a more precise level of customization can be achieved by using a CLI argument to control the amount of whitespace instead.

It's a relatively simple change that simplifies the parsing of the JSON output."

### Benefits
This feature would prove valuable when parsing JSON test results. For example, Splunk mandates JSON formatting with no whitespace, and having the option to generate whitespace-free JSON would enable Splunk to directly use the data without additional processing.

### Possible Drawbacks
There shouldn't be any drawbacks to making this modification. It allows for further customization but is kept default if not specified.

### Applicable issues
N/A


JoshuaKGoldberg (changes requested):
Marking as waiting for changes per https://github.com/mochajs/mocha/issues/4889#issuecomment-1930894308.

CheadleCheadle:
I addressed the changes requested in https://github.com/mochajs/mocha/issues/4889#issuecomment-1930894308.

Instead of `jsonStringifyWhiteSpace`, I went with `indentSize`. It is not only shorter and easier to type, but clearer in what is used for!

lib/reporters/json.js
```diff
@@ -78,7 +78,16 @@ function JSONReporter(runner, options = {}) {
 
     runner.testResults = obj;
 
-    var json = JSON.stringify(obj, null, 2);
+    var optionsIndentSize;
+    if (options.reporterOption && 'indentSize' in options.reporterOption) {
+      optionsIndentSize = options.reporterOption.indentSize;
+    } else {
+      optionsIndentSize = 2; // Default Indentation size
+    }
```
JoshuaKGoldberg, line 86: 🧹✨ 

```suggestion
    var optionsIndentSize= options.reporterOption?.indentSize ?? 2;
```

We support Node >=14, which supports [`??` nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#browser_compatibility) and [`?.` optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility).

lib/reporters/json.js
```diff
@@ -78,7 +78,16 @@ function JSONReporter(runner, options = {}) {
...
+
+    var indentSize = parseInt(optionsIndentSize, 10); // Cast string to int
```
JoshuaKGoldberg, line 88: [Bug] What if someone passes in `'\t'`? I'd think it'd be nice to allow that. And no real reason not to, right?

```suggestion
```

...which makes me thing: how about just calling this `options.indentation`?

JoshuaKGoldberg (changes requested):
Progress! 🚀

lib/reporters/json.js
CheadleCheadle: I like it!

lib/reporters/json.js
CheadleCheadle: I agree. Handling tabs is a good addition

bin/mocha.js
JoshuaKGoldberg, line 1: @CheadleCheadle please revert this file, again.

lib/reporters/json.js
JoshuaKGoldberg, line 1: [Docs] I just realized the new reporter option isn't documented  anywhere. Could you add a note to [`### JSON`](https://github.com/mochajs/mocha/blob/8711e72b9dd573dff6940061dd5643986c02f38a/docs/index.md#L1965)?

JoshuaKGoldberg (changes requested):
Getting there, but I think we might have missed each other on the reporter option name & usage?

test/reporters/json.spec.js
```diff
@@ -229,5 +229,19 @@ describe('JSON reporter', function () {
         'file output not supported in browser'
       );
     });
+
+    it('should set options.indentation correctly', function () {
```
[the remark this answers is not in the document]
CheadleCheadle: Gotcha. I'll make some changes.

bin/mocha.js
CheadleCheadle: I'm not entirely sure why there are permissions changes on bin/mocha. Any advice?

JoshuaKGoldberg:

JoshuaKGoldberg:
Closing out to free up the backing issue. @CheadleCheadle if you end up seeing this and having time again, no worries, feel free to re-open the PR or send a new one.

Cheers all! 🤎

Issue #4982, opened 2023-05-01
🚀 Feature: Support printing each error within AggregateErrors


domenic:
**Is your feature request related to a problem or a nice-to-have?? Please describe.**

Sometimes I'm using Mocha to test something which can fail multiple ways. This is what [`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) is designed for.

However, when an `AggregateError` is thrown in a Mocha test, nothing nice prints:

```js
it("should work with AggregateErrors", () => {
  const err1 = new Error("1");
  const err2 = new Error("2");
  const aggErr = new AggregateError([err1, err2], "2 errors");

  throw aggErr;
});
```

gives

```
  1) should work with AggregateErrors:
     AggregateError: 2 errors
      at Context.<anonymous> (x.test.js:4:18)
      at process.processImmediate (node:internal/timers:478:21)
```

**Describe the solution you'd like**

It should probably print something for each error, as well as the AggregateError itself. Maybe

```
     AggregateError: 2 errors
      at Context.<anonymous> (x.test.js:4:18)
      at process.processImmediate (node:internal/timers:478:21)

      Error: 1
        at Context.<anonymous> (x.test.js:2:16)
        at process.processImmediate (node:internal/timers:478:21)

      Error: 2
        at Context.<anonymous> (x.test.js:3:16)
        at process.processImmediate (node:internal/timers:478:21)
```        

**Describe alternatives you've considered**

I tried making a fake `AggregateError`:

```js
it("should work with fake AggregateErrors", () => {
  const err1 = new Error("1");

  const err2 = new Error("2");

  const aggErr = new Error(`2 errors:\n\n${err1.stack}\n\n${err2.stack}`);

  throw aggErr;
});
```

This seems to confuse Mocha's stack trace defection code and result in printing the error twice:

```
  1) should work with fake AggregateErrors:
     2 errors:

Error: 1
    at Context.<anonymous> (C:\Users\Domenic\Dropbox\Programming\WIP\jest-aggregateerror-test\x.test.js:2:16)
    at callFn (C:\Users\Domenic\AppData\Local\npm-cache\_npx\508606763866ae01\node_modules\mocha\lib\runnable.js:366:21)
    at Runnable.run (C:\Users\Domenic\AppData\Local\npm-cache\_npx\508606763866ae01\node_modules\mocha\lib\runnable.js:354:5)
[...]

PR #4829, opened 2022-02-09
feat: include `.cause` stacks in the error stack traces

voxpelli:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions.

### Description of the Change

Appends the full error stack chain for errors with causes.

### Alternate Designs

Prior to the Error cause standardization, there was `VError` which I did a PR for supporting some years ago #2381. That one as primarily closed because an aversion to supporting random module hooks, which this now being a standard changes.

So I think this is now more a matter of:

* Should this be on by default?
* How should this be formatted?

### Why should this be in core?

All current browsers supports this property and Node.js as well since `16.x`.

See also eg:

* [V8 blog post](https://v8.dev/features/error-cause)
* [My blog post describing my ponyfill + helpers for it](https://dev.to/voxpelli/pony-cause-1-0-error-causes-2l2o)
* `pino` merging support for it: https://github.com/pinojs/pino-std-serializers/pull/78

### Benefits

Error causes are a great way to enrich errors and to ensure that one can catch all relevant places across an async workflow. See also eg. this old Joyent article explaining [how they uses VError](https://www.joyent.com/node-js/production/design/errors#5-if-you-pass-a-lower-level-error-to-your-caller-consider-wrapping-it-instead).

### Possible Drawbacks

* It can become a bit too verbose if a lot of unnecessary causes gets printed, but is that an issue to take into consideration here?

### Applicable issues

* This would be a breaking change for everyone that relies on specific stack output for current error object's containing a stack possessing `cause` property, else it would be a minor. So depends on whether the specific stack output is seen as an API that should be stable or whether it's simply a presentation detail.

Fixes #2735.

PR #5018, opened 2023-10-19
feat: support reporting AggregateErrors 

CheadleCheadle:
### Requirements

### Description of the Change
Addresses #4982
If an AggregateError is thrown, the base reporter will print out the AggregateError's errors. Before this change, only the Aggregate Error itself was printed. So, if the current error being printed is an Aggregate error the errors it contains will also be formatted and printed.

### Alternate Designs
I could have completely restructured the Base.list method to only loop over the failures and call a helper function that would create the format(fmt), msg, and title and then print the error, but it was much simpler to just add a conditional check to see if the current error being printed was an AggregateError and if it contained multiple errors and just 

Currently, nothing nice prints. When dealing with AggregateErrors it would be better if each subsequent error was printed and formatted as well.

### Benefits
The errors within the AggregateError will be nicely formatted when printed.

### Possible Drawbacks
If an AggregateError contains a lot of errors, then a lot of errors will be printed!

### Applicable issues
Nothing of note.


lib/cli/run-option-metadata.js
```diff
@@ -49,7 +49,7 @@ const TYPES = (exports.types = {
     'sort',
     'watch'
   ],
-  number: ['retries', 'jobs'],
+  number: ['retries', 'jobs', 'jsonStringifyWhitespace'],
```
voxpelli, line 52: Minor feedback: This seems to be an unrelated addition

lib/reporters/base.js
```diff
@@ -246,6 +246,7 @@ exports.list = function (failures) {
     if (test.err && test.err.multiple) {
       if (multipleTest !== test) {
         multipleTest = test;
+
```
voxpelli, line 249: Preferably avoid needless styling changes like this 🙏

lib/reporters/base.js
```diff
@@ -301,11 +302,37 @@ exports.list = function (failures) {
       }
       testTitle += str;
     });
-
```
voxpelli, line 304: Preferably avoid needless styling changes like this 🙏

lib/reporters/base.js
```diff
@@ -301,11 +302,37 @@ exports.list = function (failures) {
...
     Base.consoleLog(fmt, i + 1, testTitle, msg, stack);
+
+    // Handles formatting and printing Aggregate errors
+    if (test.err.errors) {
+      for (const error of test.err.errors) {
+        const testTitle = addIndent(`Error ${error.message}`, 0);
+        const stack = addIndent(error.stack, 3);
+
+        Base.consoleLog(fmt, error.message, testTitle, '', stack);
```
voxpelli, line 313: I think the formatting of aggregated errors should use the same formatting as non-aggregated errors, so rather than formatting here locally, send `test.err.errors` to `Base.list` or such?

Helps with:

- Aggregated errors that are nested in aggregated errors
- Aggregated errors with causes in them, pending: https://github.com/mochajs/mocha/pull/4829

package-lock.json
```diff
@@ -109,10 +109,6 @@
       },
       "engines": {
         "node": ">= 14.0.0"
-      },
-      "funding": {
-        "type": "opencollective",
-        "url": "https://opencollective.com/mochajs"
```
voxpelli, line 115: Unrelated change, would suggest removing

lib/reporters/json.js
```diff
@@ -153,6 +153,7 @@ function cleanCycles(obj) {
  */
 function errorJSON(err) {
   var res = {};
+
```
voxpelli, line 156: Styling change, would suggest removing

voxpelli (changes requested):
Some initial feedback

lib/cli/run-option-metadata.js
CheadleCheadle: This relates to #5017 and involves introducing an optional --jsw flag that allows users to specify the amount of whitespace in JSON reporter output. But, I'll exclude this from the current pr to focus on addressing one feature addition.

lib/reporters/base.js
CheadleCheadle: Good idea! If a test contains an aggregate error it now recursively calls Base.list with those errors.

wesleytodd:
Plus one on this. I just ran into it. Is anyone working on this anymore? It looks like maybe the feedback was addressed?

nwalters512:
Somewhere between Node 16 and Node 20, Node's own APIs started throwing `AggregateError` in certain cases, e.g. network failures when `autoSelectFamily === true` (the default). Without proper formatting, it's very difficult to debug such errors when they occur in Mocha tests. Is there anything we can do to help this along?

CheadleCheadle:
I addressed all the changes and sent an email to voxpelli to see if he could approve it.

voxpelli:
I'll try to get around to it soon, else eg. @JoshuaKGoldberg can chime in?

lib/reporters/base.js
```diff
@@ -292,17 +295,24 @@ exports.list = function (failures) {
...
-    test.titlePath().forEach(function (str, index) {
-      if (index !== 0) {
-        testTitle += '\n     ';
-      }
-      for (var i = 0; i < index; i++) {
-        testTitle += '  ';
-      }
-      testTitle += str;
-    });
+    // Incase test is an error object when recursively listing aggregate errors
```
wesleytodd, line 298: I am not super familiar with `mocha` internals, but this comment and the line below is suspicious to me.

The comment implies this check is for when `test` is an instance of `AggregateError` but then the check is for `titlePath`. I am not saying the code is wrong, but I am saying that whatever is going on here is not clear and could probably be commented or written in a way that was more clearly addressing the issue.

wesleytodd (approved):

lib/reporters/base.js
CheadleCheadle: I went ahead and modified that if statement to check if the error is an instance of an Aggregate Error. It accomplishes the same thing but is much more clear why it is doing it.

voxpelli:
I'm running out of time to review and merge this today, but this one together with these two are my main priority right now:

* https://github.com/mochajs/mocha/pull/5074
* #4829

bin/mocha.js
JoshuaKGoldberg, line 1: [Question] Why the change in file permissions?

lib/reporters/base.js
```diff
@@ -292,17 +295,24 @@ exports.list = function (failures) {
...
-    test.titlePath().forEach(function (str, index) {
-      if (index !== 0) {
-        testTitle += '\n     ';
-      }
-      for (var i = 0; i < index; i++) {
-        testTitle += '  ';
-      }
-      testTitle += str;
-    });
+    // Incase test is an AggregateError object when recursively listing errors
```
JoshuaKGoldberg, line 298: [Docs] Nit: I think the `instanceof AggregateError` implies what this comment is trying to say. 🙂 

```suggestion
```

lib/reporters/base.js
```diff
@@ -544,7 +554,6 @@ var objToString = Object.prototype.toString;
 function sameType(a, b) {
   return objToString.call(a) === objToString.call(b);
 }
-
```
JoshuaKGoldberg, line 547: [Style] Nit: could you undo this unrelated change & any others please?

test/reporters/base.spec.js
```diff
@@ -507,21 +507,34 @@ describe('Base reporter', function () {
...
 
-    it('should let you stub out console.log without effecting reporters output', function () {
-      Base.list([]);
-      baseConsoleLog.restore();
+    var errOut = stdout.join('\n').trim();
 
-      expect(baseConsoleLog, 'was called');
-      expect(console.log, 'was not called');
-    });
+    expect(errOut, 'to contain', ' Error: 1', 'Error: 2');
```
JoshuaKGoldberg, line 520: [Testing] Aggregate / multiple test error reports can be tricky. I think we'd want to test more than just whether these two lines are included. Could you test more directly what the actual output is? I.e. that it includes the proper newlines, spacing, any details/messages/stacks (or not) as expected?

test/reporters/base.spec.js
```diff
@@ -507,21 +507,34 @@ describe('Base reporter', function () {
     );
   });
 
-  describe('when reporter output immune to user test changes', function () {
-    var baseConsoleLog;
+  it('should list all the errors within an AggregateError', function () {
+    var err1 = new Error('1');
+    var err2 = new Error('2');
+    var aggErr = new AggregateError([err1, err2], '2 errors');
```
JoshuaKGoldberg, line 513: [Testing] A few more test cases that should be covered...

* An `AggregateError` that only has one error
* An `AggregateError` where the constructor receives something that isn't an `Error` - which is possible per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError/AggregateError#errors

JoshuaKGoldberg (changes requested):
Good start! 🙌 

I'd also like to hear from @voxpelli, but leaving some thoughts in the meantime too.

bin/mocha.js
CheadleCheadle: Must have been accidental because I don't recall modifying bin/mocha.

lib/reporters/base.js
CheadleCheadle: Yeah, that one speaks for itself, you're right!

lib/reporters/base.js
CheadleCheadle: Will do

test/reporters/base.spec.js
CheadleCheadle: I went ahead and added test cases for these.

bin/mocha.js
JoshuaKGoldberg, line 1: Still the permissions change? Is there an automatic modification happening on save or something like that?

CheadleCheadle: It's possible. I'll look into it this week.

JoshuaKGoldberg (changes requested):
Whoops, I'd meant to post this a bit back, sorry - but looks like there are failing tests?

JoshuaKGoldberg:

nwalters512:
@JoshuaKGoldberg I'm very interested in seeing this land. What's the best way for me to help, open PRs against _this_ branch? Or open a brand-new PR with these same changes?

JoshuaKGoldberg:
Yeah, this has been stale for a while. I think it's safe to close out.

If you end up using code from this PR, please add a co-author attribution. Otherwise 🚀 free to make a new one!

PR #5020, opened 2023-10-31
chore: drop node 19, add node 20 and 21 to workflows

Uzlopak:
Using lts for browser tests. Removing node 19 and add node 20 and 21 to the workflows.

### DIFF
--- a/.github/workflows/mocha.yml
+++ b/.github/workflows/mocha.yml
@@ -45,7 +45,8 @@ jobs:
           - 14
           - 16
           - 18
-          - 19
+          - 20
+          - 21
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/mocha.yml
```diff
@@ -45,10 +47,12 @@ jobs:
           - 14
           - 16
           - 18
-          - 19
+          - 20
```
JoshuaKGoldberg, line 50: Might make sense to have adding Node 21 as a followup issue? https://nodejs.org/en/blog/announcements/v21-release-announce

JoshuaKGoldberg (approved):
Looks great to me at a quick glance! I can't give a full review just now but I trust you very much over myself on these matters 😄

.github/workflows/mocha.yml
Uzlopak: There are currently issues with node21 and actions/setup-node. Times for some reason out. I had first node 21 but because of that i removed it.

PR #5023, opened 2023-11-05
chore: remove touch as dev dependency

Uzlopak:
Remove touch as dev dependency.

### DIFF
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -495,6 +495,7 @@ function touchFile(filepath) {
     fs.closeSync(fd);
   }
 }
+
 /**
  * Synchronously replace all substrings matched by `pattern` with
  * `replacement` in the contents of file at `filepath`
--- a/test/integration/helpers.js
+++ b/test/integration/helpers.js
@@ -7,7 +7,6 @@ const {format} = require('util');
 const path = require('path');
 const Base = require('../../lib/reporters/base');
 const debug = require('debug')('mocha:test:integration:helpers');
-const touch = require('touch');
 
 /**
  * Path to `mocha` executable
@@ -479,6 +478,8 @@ async function runMochaWatchJSONAsync(args, opts, change) {
   );
 }
 
+const touchRef = new Date();
+
 /**
  * Synchronously touch a file. Creates
  * the file and all its parent directories if necessary.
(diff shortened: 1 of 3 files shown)


test/integration/helpers.js
```diff
@@ -2,12 +2,12 @@
 
 const escapeRegExp = require('escape-string-regexp');
 const os = require('os');
-const fs = require('fs-extra');
+const fs = require('fs');
+const fsExtra = require('fs-extra');
```
voxpelli, line 6:
```suggestion
const {
  utimesSync,
} = require('fs');
const {
  ensureDirSync,
  readFileSync,
} = require('fs-extra');
```

Thinking out loud: How about destructuring here?

Also: Doesn't `fs-extra` re-export everything from `fs`? What's the reason for introducing `fs` in addition to `fs-extra`?

voxpelli (commented):
Some comments

Generally I like a focused module like `touch` more than a huge one like `fs-extra`, but since `touch` hasn't seen an update in 6 years but `fs-extra` has gotten one in the last year, then 👍 

I do think we can leave `fs-extra` being imported as `fs` though?

test/integration/helpers.js
Uzlopak: i didnt know it is re exporing. 
I want to make as few as possible changes.

PTAL

JoshuaKGoldberg (approved):
🔪

test/integration/helpers.js
```diff
@@ -487,9 +488,13 @@ async function runMochaWatchJSONAsync(args, opts, change) {
...
   fs.ensureDirSync(path.dirname(filepath));
-  touch.sync(filepath);
+  try {
+    fs.utimesSync(filepath, touchRef, touchRef);
+  } catch (e) {
+    const fd = fs.openSync(filepath, 'a');
+    fs.closeSync(fd);
+  }
 }
-
```
JoshuaKGoldberg, line 492: [Style] Was removing this blank line intentional? (not at all a blocker for me, though I do prefer it there a bit 😄)

test/integration/helpers.js
```diff
@@ -487,9 +488,13 @@ async function runMochaWatchJSONAsync(args, opts, change) {
...
 function touchFile(filepath) {
   fs.ensureDirSync(path.dirname(filepath));
-  touch.sync(filepath);
+  try {
+    fs.utimesSync(filepath, touchRef, touchRef);
+  } catch (e) {
+    const fd = fs.openSync(filepath, 'a');
+    fs.closeSync(fd);
+  }
 }
```
Uzlopak, line 497:
```suggestion
}

```

PR #5024, opened 2023-11-05
chore: remove nanoid as dependency

Uzlopak:
trivial replacement of nanoid

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -8,8 +8,6 @@
 /**
  * Module dependencies.
  */
-
-const {nanoid} = require('nanoid/non-secure');
 var path = require('path');
 var util = require('util');
 var he = require('he');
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -615,11 +613,20 @@ exports.constants = exports.defineConstants({
...
  * @returns {string} Unique identifier
  */
-exports.uniqueID = () => nanoid();
+exports.uniqueID = () => {
+  let id = '';
+  for (let i = 0; i < 21; i++) {
+    id += uniqueIDBase[(Math.random() * 64) | 0];
+  }
+  return id;
+};
```
voxpelli, line 629: Can we mention in the comment that this is essentially a copy of `nanoid/non-secure`?

Especially as that explains why `Math.random` is okay whereas it typically shouldn't be used for unique ID:s (I wonder where in the code this is used?)

voxpelli (approved):

PR #5038, opened 2023-12-02
docs: overhaul contributing and maintenance docs for end-of-year 2023

### LINKED ISSUE #5036: Overhaul issue and PR templates
It's been a while since the issue and PR templates here were looked at. [GitHub Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) are now publicly available too. We'll go through the existing docs and touch them up.

This is part of #5027.

### LINKED ISSUE #5037: Revisit and overhaul existing maintenance docs
It's been a while since the code of conduct, contributing guidelines, development guidelines, and maintenance guide here were looked at. We'll go through the existing docs and touch them up.

This is part of #5027.

JoshuaKGoldberg:
## Overview

Applies what I've personally found to be reasonable and usable contributing and maintenance processes as an iteration to the existing docs. Includes:

* General docs:
  * Splitting a new `.github/DEVELOPMENT.md` out of `.github/CONTRIBUTING.md`
  * Renaming "JS Foundation" to "OpenJS Foundation", per the foundation's rename
  * Split sentences to one per line (as if https://github.com/JoshuaKGoldberg/sentences-per-line were enabled)
* Issues:
  * Added two new issue types: _documentation_ and _repository tooling_
  * Disabled filing a new blank issue
  * Streamlined the contents of the issue templates a bit, and added some emojis
  * Used the new [GitHub Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) instead of `.md` templates
* Pull requests:
  * Streamlined the template
* Labels:
  * Renaming labels to include `area: ` or `type: ` prefixes
  * Switching from needing an explicit `unconfirmed-bug` to opting into `status: ...` labels after triage
* Switching from the old Gitter to a new Discord
* Switching from requiring rebasing to merging

Does not include likely followups:

* Renaming the `master` branch to `main`
* Enforcing any formatting/spellchecking/style with any new tooling
* Set up commit/PR linting to enforce PR titles and nothing else, per squash merging

Fixes #5036. Fixes #5037.


README.md
```diff
@@ -1,5 +1,5 @@
 <p align="center">
-  <img src="https://cldup.com/xFVFxOioAU.svg" alt="Mocha test framework"/>
+  <img src="https://cldup.com/xFVFxOioAU.svg" alt="Mocha test framework logo"/>
```
JoshuaKGoldberg, line 2: Accessibility nitpick: making it more clear that this is a _logo_.

README.md
```diff
@@ -22,10 +22,11 @@
 
 ## Links
 
-- **[Documentation](https://mochajs.org/)**
+- **[Documentation](https://mochajs.org)**
```
JoshuaKGoldberg, line 25: A nitpick on my end: unnecessary trailing `/` slashes at the end of URLs irritate me. 😄

plroebuck: Web server always assumes it's serving a file, unless the URL ends in a directory separator.

If the URL is to a directory, the server will append the directory index file (typically 'index.html') to the URL before it tries. If the server fails to return a file, it appends the directory separator to the path and retries as a directory.

Thus doing it this way causes *twice* the number of network calls needed for every directory URL without a trailing directory separator.

[Apache: Mapping URLs to Filesystem Locations #DocumentRoot](https://httpd.apache.org/docs/2.4/urlmapping.html#documentroot)

voxpelli: @plroebuck Thats something that eg Apache does internally – it’s not something done through multiple requests – and eg plain Node.js doesn’t do anything like that at all

@JoshuaKGoldberg Not sure I like this though, the one with the slash in the end is the more “correct” one 😜

voxpelli:
@JoshuaKGoldberg We should maybe investigate adding eg https://github.com/JoshuaKGoldberg/sentences-per-line later? To document / enforce such stylistic choices and making it easier for contributors to follow them?

README.md
JoshuaKGoldberg: Haha how do we define "correct" here... per _some_ specs the trailing slash is more correct. But end-users are generally told to visit URLs without saying the slash.

Uzlopak:
Didnt we want a "Checkbox I want to provide a PR"

MAINTAINERS.md
```diff
@@ -211,108 +199,164 @@ Examples of a breaking changes might be:
...
 
-`master` is the only maintained branch in `mochajs/mocha` or any of the other repos. **`master` is the only branch to which force-pushing is disallowed.**
+`master` is the only maintained branch in `mochajs/mocha` or any of the other repos.
+**`master` is the only branch to which force-pushing is disallowed.**
 
 Maintainers may push new branches to a repo, as long as they remove them when finished (merging a PR will prompt to do so).
 
 Please _please_ **_please_** delete old or unused branches.
 
 ## Merging PRs
```
JoshuaKGoldberg, line 356: We're discussing this now. Our consensus is:

1. We don't believe the benefits of a pristine Git history to be worth the cost of asking everyone to maintain that history
2. Squash merging PRs using their title + description preserves a relatively clean history
3. We'll switch to squash merging always, and only enforcing conventions on the PR title
4. We can therefore remove commit linting for anything other than file formatting

.github/ISSUE_TEMPLATE/01-bug.yml
JoshuaKGoldberg, line 1: @Uzlopak moving https://github.com/mochajs/mocha/pull/5038#issuecomment-1841139467 here so we can thread:

> Didnt we want a "Checkbox I want to provide a PR"

I don't see any benefit to having those. IME issue claiming systems often result in accidental ["licked cookie"](https://devblogs.microsoft.com/oldnewthing/20091201-00/?p=15843) situations where contributors claim an issue but run out of time or energy trying before sending a PR. And other than claiming the issue I don't see any reason to add a checkbox folks have to read & make a decision on?

voxpelli: I don't see it as claiming, more as being open to pay it forward, I'm 👍 on it

JoshuaKGoldberg: I'll add it in 🚀

Uzlopak (approved):
LGTM

.github/ISSUE_TEMPLATE/02-documentation.yml
```diff
@@ -0,0 +1,26 @@
...
+      required: true
+  - attributes:
+      description: Any additional info you'd like to provide.
+      label: Additional Info
+    type: textarea
+description: Report a typo or missing area of documentation
+labels:
+  - "area: documentation"
+name: 📝 Documentation
+title: "📝 Documentation: <short description of the request>"
```
JoshuaKGoldberg, line 26: 💡 shorter name good!

```suggestion
name: 📝 Docs
title: "📝 Docs: <short description of the request>"
```

Issue #2116, opened 2016-02-18
Please allow describe can do async with done/promise


Thaina:
In addition to --delay flag. Describe should also able to use done/promise flow to do async before using it

Such as, I want to do dynamic test for all user in database. I need to query user in database with async first, then each of user will emit their own test case in block

```
describe(function(done){
   query(function(users){
       users.forEach(function(user,i){
             it("Test : " + user.ID,function() {
                  if(!user.ID)
                        throw new Error("user " + i + has no ID);
             });
       });

       done();
   }); 
});
```

ps. This thread is copied from #2085 . It was ruin by a someone and was locked. And I need to make it reopen to ask for something

@boneskull I know that it would not be implemented immediately. That's what software development process is

What I would like you to do is consider you will never do it. Or will be doing it in which version. So I would consider move to another tool or make it by myself

ps2 @boneskull if you just not hurry lock that thread I would not need to waste another thread like this

boneskull:
@Thaina I locked it because you were becoming belligerent.   It's difficult to infer tone in written comments, but I'm sensing some of the same here.  Please be polite when posting in issues. 

We may be able to implement this in the next major (nothing imminent).  I'm not sure it's wise to merge any feature PRs at this point, however.  Feel free to implement it yourself in a fork, or use another tool.

jdmarshall:
I've been using describe->before(done) to accomplish this, and it seems to be working just fine.

mmiszy:
It is working, but it's inconvenient. `describe('', async () => {} )` we definitely be a step forward!

mmiszy:
Related: #2975

pawelngei:
:+1:

plroebuck:
Enable this `eslint-plugin-mocha`[rule](https://github.com/lo1tuma/eslint-plugin-mocha/blob/d36c1499133ab688bd81a650e19f0d0cd11ea248/docs/rules/no-async-describe.md).

We backed out the deprecation warning in #3744 due to problems dealing with CoffeeScript's implicit return, but `describe`'s function is _just a namespace_ -- **don't add `async` to it!**

jdmarshall:
I'm not too keen on this model.  And by 'not too keen' I mean I'd be deeply upset and disappointed in Mocha if this was added.

Before, After, and It can all deal with async code, promises, and callbacks.  The only use I can see for adding 'done' to suites is to run them in parallel.

[...]

Issue #2085, opened 2016-02-02
Please allow describe can do async with done/promise


Thaina:
In addition to --delay flag. Describe should also able to use done/promise flow to do async before using it

Such as, I want to do dynamic test for all user in database. I need to query user in database with async first, then each of user will emit their own test case in block

```
describe(function(done){
   query(function(users){
       users.forEach(function(user,i){
             it("Test : " + user.ID,function() {
                  if(!user.ID)
                        throw new Error("user " + i + has no ID);
             });
       });

       done();
   }); 
});
```

ORESoftware:
You should put the database query in a before() hook and use done() with the before; there is no done with describe, because describe is designed to register all functions synchronously.

this is what before() hooks are for!

EcherHet:
@ORESoftware  You don't understand the use case 
Before Cannot make dynamic it. Because as you said, describe is normally synchronous

Register test case asynchronously Is what I want. lt should be possible. And it actually already possible with --delay

 I was just want to propose additional way to do it with done in describe. so it is the same code with it(name, done). describe(name, done) should also available

Thaina:
@ORESoftware You don't understand the use case 
Before Cannot make dynamic it. Because as you said, describe is normally synchronous

Register test case asynchronously Is what I want. lt should be possible. And it actually already possible with --delay

I was just want to propose additional way to do it with done in describe. so it is the same code with it(name, done). describe(name, done) should also available

ORESoftware:
I see what you mean. You want something like this:

```js

describe('foor',function(){

    var arr = [];

    before(function(done){

        setTimeout(function(){
            arr.push(1);
            arr.push(2);
            arr.push(3);
            done();
        },1000);

    });

    describe('bar',function(){

        arr.forEach(function(item){

            it('[test]' + item, function(){

            });

        });

    });
    
});

```

in the above, no tests will be registered...yeah I don't know. Someone enlighten us.

Thaina:
@ORESoftware I know it cannot work like that from the start. before() is called only with it(). And nothing could be done to delay registering it() in describe() except --delay flag

mocha really doesn't have this funtionality so I need to create this issue to propose new feature. The enlightening you could get is accept that this is missing feature

ORESoftware:
Sure, you may have a point, I can't say for sure since I am not a core contributor to mocha

ORESoftware:
[...]

PR #5046, opened 2023-12-12
docs: Add warning about async callback for describe

mcdurdin:
### Requirements

### Description of the Change

This is a documentation only change.

I couldn't find any reference to `describe` not supporting an async callback. It seems like a natural conclusion to draw, given mocha specializes in async tests, and given `it` and hook do.

I spent considerable time trying to figure out why an async test was failing in one of my modules before I discovered the reason deep in an issue discussion (https://github.com/mochajs/mocha/issues/2975#issuecomment-1004176440)!

This is a documentation change.

### Benefits

I hope that this helps others design their test suites (I saw a number of comments about spending 'lots of time' trying to figure this out).

### Applicable issues

* #2975 
* #2116 
* #2085

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -296,6 +296,10 @@ describe('#find()', function () {
 });
 ```
 
+### Limitations of asynchronous callbacks
+
+You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because, in the "parsing" cycle all `describe` callbacks are excecuted and the test hierarchy (`runner`) is created, before any tests are run.
+
 ## Synchronous Code
 
 When testing synchronous code, omit the callback and Mocha will automatically continue on to the next test.
--- a/docs/index.md
+++ b/docs/index.md
@@ -298,7 +298,7 @@ describe('#find()', function () {
 
 ### Limitations of asynchronous callbacks
 
-You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because, in the "parsing" cycle all `describe` callbacks are excecuted and the test hierarchy (`runner`) is created, before any tests are run.
+You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because in the "parsing" cycle all `describe` callbacks are executed and the test hierarchy (`runner`) is created before any tests are run.
 
 ## Synchronous Code
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -298,7 +298,8 @@ describe('#find()', function () {
 
 ### Limitations of asynchronous callbacks
 
-You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because in the "parsing" cycle all `describe` callbacks are executed and the test hierarchy (`runner`) is created before any tests are run.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -296,6 +296,10 @@ describe('#find()', function () {
 });
 ```
 
+### Limitations of asynchronous callbacks
+
+You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because, in the "parsing" cycle all `describe` callbacks are excecuted and the test hierarchy (`runner`) is created, before any tests are run.
```
JoshuaKGoldberg, line 301:
```suggestion
You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because in the "parsing" cycle all `describe` callbacks are executed and the test hierarchy (`runner`) is created before any tests are run.
```

JoshuaKGoldberg (approved):
LGTM! Just a couple small touchups for typos.

JoshuaKGoldberg:
Note that #4921 should also help with this.

docs/index.md
```diff
@@ -296,6 +296,10 @@ describe('#find()', function () {
 });
 ```
 
+### Limitations of asynchronous callbacks
+
+You can use all three asynchronous patterns (`done`, `Promise`, and `async`/`await`) in callbacks for `it()` and for all hooks (`before()`, `after()`, `beforeEach()`, `afterEach()`), but `describe()` will not work correctly with an asynchronous callback -- it must be synchronous. This is because in the "parsing" cycle all `describe` callbacks are executed and the test hierarchy (`runner`) is created before any tests are run.
```
voxpelli, line 301:
```suggestion
You can use all asynchronous callbacks (`done`, `Promise`, and `async`/`await`) in callbacks for `it()`, `before()`, `after()`, `beforeEach()`, `afterEach()`) but not `describe()` -- it must be synchronous.
```

voxpelli (changes requested):
Can we simplify it a bit?

docs/index.md
mcdurdin: OK for all changes except that I feel like understanding the 'why' really helped me here.

voxpelli: Yeah, I'm thinking its maybe getting a bit too much into the current implementation of things, I'll have a think about it.

If we include it then maybe as a new paragraph, like `Technically this is caused by...`? Thoughts @mochajs/maintenance-crew?

voxpelli: Yeah, and it can always be found by looking in the GitHub history of that documentation line as it will reference this PR 😌

JoshuaKGoldberg:
👋 @mcdurdin is this still something you have time for? Just checking in!

docs/index.md
mcdurdin: I am still keen on including a technical reason -- it's a fundamental design in mocha that is seriously not obvious, and it took me (and others, read the linked issues) a _long_ time to find out why my tests weren't working properly. (You can always remove the explanation in the unlikely event of mocha redesign making it unimportant...)

IMO the docs are for technical readers; they shouldn't be dumbed down to the point of 'do this because we tell you to'. That's really not helpful for users; they end up using mocha without understanding it, and get things wrong -- like this.

Little details in technical documentation are the things that make a project stand out.

But if you really don't want that extra detail, I'll pull it out.

JoshuaKGoldberg: How about:

```suggestion
You can use all asynchronous callbacks (`done`, `Promise`, and `async`/`await`) in callbacks for `it()`, `before()`, `after()`, `beforeEach()`, `afterEach()`) but not `describe()` -- it must be synchronous.
See [#5046](https://github.com/mochajs/mocha/pull/5046) for more information.
```

> IMO the docs are for technical readers

I think the difference here isn't from whether the docs are for technical readers, it's to what extent readers want those technical bits from high-level docs. Many readers on a top-level page like https://mochajs.org are just seeking the _"do this because we tell you to"_. For those users, the choice is to either give that to them, or have them zone out and find something else.

Ideally we should have deeper docs to link to that explain things like architecture and design choices. I'd be very in favor of a site reorg that encompasses that. That'd let us have both the surface information _and_ deeper bits. +1 to the urge to give users the option to actually understand Mocha!

mcdurdin: > Ideally we should have deeper docs to link to that explain things like architecture and design choices. I'd be very in favor of a site reorg that encompasses that.

+1 on this, agree that would be the ideal.

Happy with your suggestion -- strikes a good compromise for those who want to understand why.

JoshuaKGoldberg (approved):
I'll go ahead and take the liberty of merging this now, as it's a docs change. Thanks for the productive conversation @mcdurdin! 🙌

PR #5057, opened 2023-12-28
fix: set test total properly with --fail-zero and --delay

### LINKED ISSUE #4950: --fail-zero always fails the test run when --delay is also used
### Prerequisites

### Description

Running mocha with --fail-zero and --delay options, and calling run(). The exit code is always 1, even when some tests did run.

### Steps to Reproduce

I created a gist with the files used in the examples below: https://gist.github.com/tonimellodic/525d29054744ca6d602145b81eb57e80

**Expected behavior:** Running with `--fail-zero` and `--delay` produces `exit 0` when at least one test run.

**Actual behavior:** Running with `--fail-zero` and `--delay` always produces `exit 1`.

Running with `--fail-zero` only:

```
% npx mocha --fail-zero test/no-delay.js     

  No delay example
    ✔ runs a test

  1 passing (3ms)

% echo $?
0
```

Running with `--fail-zero` and `--delay`:

```
% npx mocha --fail-zero --delay test/delay.js

  Delay example
    ✔ runs a delayed test

  1 passing (5ms)

% echo $?
1
```

**Reproduces how often:** Always

### Versions

mocha 10.1.0
node 16.13.1
MacOS 12.6.1
zsh shell

joandino:
## PR Checklist

- [X] Addresses an existing open issue: fixes #4950

## Description of the change

When running --fail-zero along with --delay the result always return 1 even when at least one test runs.
Apparently the total variable in the runner was not being updated after the event delay ends and the tests runs.

On the emit of the `EVENT_DELAY_END` we make sure to update the total of tests ran.

## Why should this be in core?

To prevent tests that run with `--fail-zero` and `--delay` always fail.

## Applicable issues

This will be a fix/patch release


JoshuaKGoldberg:

PR #5074, opened 2024-01-08
fix: harden error handling in `lib/cli/run.js`

### LINKED ISSUE #5085: 🐛 Bug: Errorhandling fails when node:module registered errors are thrown
### Bug Report Checklist

### Expected

When running a test that throws a null error i expect to get information about what happened and the errorhandler to give useful output.

### Actual

When an null error is thrown when running the tests the output from the mocha script is the program option list and a line:
`✖ ERROR: null`

I created a PR for this error - thinking this was only applicable for Typescript errors, but as stated in the PR: all null errors it will cause this to happen.

Nevertheless, the example created for the original PR shows the error:
https://github.com/stalet/mocha-typescript-testproject

### Versions

10.2.0

### Additional Info

This error was reported as part of PR #5074 - but since it's not just applicable for Typescript errors the PR was closed - and I am opening this issue instead.

stalet:
Before this patch the errorhandling would fail with the error: "ERROR: null". Debug showed that the error "caught error sometime before command handler: TypeError: Cannot convert object to primitive value"

## PR Checklist

- [x] Addresses an existing open issue: fixes #5085

## Overview
When the error in the runner is a typescript error the current script fails with the error: "TypeError: Cannot convert object to primitive value" - leaving the output as a null value.
It is very hard to see the actuall error. With this change the output will be something like this.

`Error: undefined test/testdata.ts(59,14): error TS2739: Type '{ loginId: number; spidId: number; sid: string; amr: never[]; }' is missing the following properties from type 'Identity': sub, nbf, exp, iat`

Before the fix the output was
`✖ ERROR: null`
Debug output contailed this
`  mocha:cli:cli caught error sometime before command handler: TypeError: Cannot convert object to primitive value
  mocha:cli:cli     at exports.handler (/home/stalet/code/account/min-finn-frontend/node_modules/.pnpm/mocha@10.2.0/node_modules/mocha/lib/cli/run.js:372:65) +859ms
`


voxpelli:
Can you provide a reproduction? Sounds odd that it would swallow all TypeScript errors currently

Other PR:s related to error reporting are eg:

- https://github.com/mochajs/mocha/pull/4829

stalet:
I have created a small project that reproduces the error here: https://github.com/stalet/mocha-typescript-testproject

barak007:
It's a general thing with errors
```ts
console.error('\n' + (err.stack || `Error: ${err.message || err}`));
```
this line from mocha fail if I throw null as an error
```ts
try {
 throw null;
} catch (err) { 
  console.error('\n' + (err.stack || `Error: ${err.message || err}`));
}
```
the error will be
```
Uncaught TypeError: Cannot read properties of null (reading 'stack')
```

JoshuaKGoldberg:
👀

> - [ ] Addresses an existing open issue: fixes #000

See [`.github/CONTRIBUTING.md`](https://github.com/mochajs/mocha/blob/53a4bafbdeb32576440b0a21787f2525585411c0/.github/CONTRIBUTING.md). We ask that PRs address an open accepted issue first. A few benefits:

* Issues are much more discoverable for folks looking to discuss & report suggested changes.
* The issue templates ask for the info needed to properly triage the suggestion. 
* We don't want unhappy situations where someone spends their time on a PR that can't be accepted (which would have been caught by an issue's discussion).

Closing this PR so it doesn't show up on our review queue. But if it is determined in an issue that we can take this in, we'd be happy to re-open and review. Thanks for sending!

tl;dr: please file an issue so we can discuss there 🙂.

lib/cli/run.js
```diff
@@ -369,7 +369,7 @@ exports.handler = async function (argv) {
   try {
     await runMocha(mocha, argv);
   } catch (err) {
-    console.error('\n' + (err.stack || `Error: ${err.message || err}`));
+    console.error('\n' + (err.stack || `Error: ${err && err.message}`), err);
```
voxpelli, line 372:
```suggestion
    if (!err) {
      console.error('\n Undefined error:`, err);
    } else {
      console.error('\n' + (err.stack || `Error: ${err.message}`), err);
    }
    
```

This makes sense, nothing in `catch (err)` guarantees that `err` is an `Error` and a guard like this would be good

lib/cli/run.js
```diff
@@ -369,7 +369,12 @@ exports.handler = async function (argv) {
   try {
     await runMocha(mocha, argv);
   } catch (err) {
-    console.error('\n' + (err.stack || `Error: ${err.message || err}`));
+    if (!err) {
+      console.error('\n Undefined error:`, err);
+    } else {
+      console.error('\n' + (err.stack || `Error: ${err.message}`), err);
```
barak007, line 375: Sometimes error does not have either stack or message. In that case it would be nice to `String(err)`

barak007: in this approach you would see double message when a real error thrown

lib/cli/run.js
```diff
@@ -369,7 +369,12 @@ exports.handler = async function (argv) {
   try {
     await runMocha(mocha, argv);
   } catch (err) {
-    console.error('\n' + (err.stack || `Error: ${err.message || err}`));
+    if (!err) {
+      console.error('\n Undefined error:', err);
+    } else {
+      console.error('\n' + (err.stack || `Error: ${err.message}`), err);
```
voxpelli, line 375:
```suggestion
          console.error('\n' + (err.stack || `Error: ${err.message || err}`));
```
Meant to have it work as the original

lib/cli/run.js
```diff
@@ -369,7 +369,12 @@ exports.handler = async function (argv) {
   try {
     await runMocha(mocha, argv);
   } catch (err) {
-    console.error('\n' + (err.stack || `Error: ${err.message || err}`));
+    if (!err) {
+      console.error('\n Undefined error:', err);
+    } else {
+      console.error('\n' + (err.stack || `Error: ${err.message || err}`));
+    }
```
JoshuaKGoldberg, line 376: See https://github.com/mochajs/mocha/issues/5085#issuecomment-1931191033. This handling wouldn't fix the originally reported error. 😎

JoshuaKGoldberg (changes requested):
Requesting changes to handle the `node:module` case of errors that don't have a `toString()` method. Please let me know if I've gone down the totally wrong track though!

lib/cli/run.js
stalet: You are correct. I did not verify the latest changes with the project. I have checked now - and it does not work.
I therefore added a test the object had to either be falsy or that is lacks the `toString` method - and then it works.
I have pushed it to this PR.

barak007 (commented):
LGTM.

now we only need to support `AggregateError` and Error `cause` and it will be perfect reporting
I think they have separated issues

voxpelli:
Yep, and I think they are all waiting on me, I'll try to squeeze them all in this Monday:

* https://github.com/mochajs/mocha/pull/5018
* https://github.com/mochajs/mocha/pull/4829

lib/cli/run.js
voxpelli: @JoshuaKGoldberg @stalet I guess the `err.toString` is targeted at the `err` in `Error: ${err.message || err}`? Which gets converted to a string due to the template literal?

Thinking about this even more: I think its better to simply defer to the `console.error` to format the errors. Browsers as well as Node.js does better formatting of errors than what is attempted here anyways (eg: They usually nowadays support outputting of error causes, which would otherwise require similar adaptions as #4829)

```
      console.error('\n Mocha run failed:', err);
```

voxpelli (commented):
Proposing to simply do this instead of doing our own error formatting: `console.error('\n Mocha run failed:', err);`

Thoughts @JoshuaKGoldberg @stalet?

Also: Sorry for repeated change requests and delaying of time on this. I promise I'm not trying to be a pain in the ass even though I'm probably failing at that 🙈

voxpelli:
Changed title to work with our new PR compliance check, which helps changelog generation and automated releases, since that's entirely on us 🙈

lib/cli/run.js
JoshuaKGoldberg: I think we want to handle four cases well:

1. The "happy path" of a regular `Error` - which is already handled well now
2. Standard literals/primitives like `null` or `"thrown string"` that doesn't have a `.message` or `.stack`, but can `toString`/`JSON.toString`/similar to a nice error message by default
3. Standard objects like `{ customValue: "not an error" }` that doesn't have a `.message` or `.stack`, and doesn't `toString`/string-coerce to anything useful
4. Fancy cross-domain objects like what's mentioned in https://github.com/mochajs/mocha/issues/5085#issuecomment-1931191033 that don't stringify well, can't be `toString`/`JSON.stringify`/similar into anything useful

I think this code will handle all of that roughly as best as we can:

```js
function printError(error) {
  console.error("");

  if (!error) {
    console.error("Caught error trying to run Mocha:", error);
    return;
  }

  if (error instanceof Error) {
    console.error("Caught error trying to run Mocha:");
    console.error(error);
    return;
  }

  console.error("Unknown error thrown trying to run Mocha:");
  console.error(util.inspect(error));
}
```

Trying it out with a few values:

```js
printError(null);
printError({});
printError("wat");
printError(["abc", 123]);
printError(new Error("Outside!", { cause: new Error("Inside!") }));
```

```plaintext
Caught error trying to run Mocha: null

Unknown error thrown trying to run Mocha:
{}

Unknown error thrown trying to run Mocha:
'wat'

Unknown error thrown trying to run Mocha:
[ 'abc', 123 ]

Caught error trying to run Mocha:
Error: Outside!
    at Object.<anonymous> (/Users/josh/repos/mocha-typescript-testproject/node_modules/.pnpm/mocha@10.2.0/node_modules/mocha/lib/cli/run.js:388:12)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    ... 7 lines matching cause stack trace ...
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10) {
  [cause]: Error: Inside!
      at Object.<anonymous> (/Users/josh/repos/mocha-typescript-testproject/node_modules/.pnpm/mocha@10.2.0/node_modules/mocha/lib/cli/run.js:388:43)
      at Module._compile (node:internal/modules/cjs/loader:1376:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
      at Module.load (node:internal/modules/cjs/loader:1207:32)
      at Module._load (node:internal/modules/cjs/loader:1023:12)
      at Module.require (node:internal/modules/cjs/loader:1235:19)
      at require (node:internal/modules/helpers:176:18)
      at Object.<anonymous> (/Users/josh/repos/mocha-typescript-testproject/node_modules/.pnpm/mocha@10.2.0/node_modules/mocha/lib/cli/commands.js:13:15)
      at Module._compile (node:internal/modules/cjs/loader:1376:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
}
```

...and using it in here:

```plaintext
Unknown error thrown trying to run Mocha:
test/index.test.ts(5,1): error TS2582: Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.
test/index.test.ts(6,5): error TS2582: Cannot find name 'it'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.
test/index.test.ts(9,44): error TS2322: Type 'string' is not assignable to type 'number'.
```

voxpelli: Solving with my suggestion then and merging 👍 
Double checked with @JoshuaKGoldberg over chat 🙂

voxpelli (approved):

JoshuaKGoldberg (approved):
🚀

stalet:
> Proposing to simply do this instead of doing our own error formatting: `console.error('\n Mocha run failed:', err);`
> 
> Thoughts @JoshuaKGoldberg @stalet?

Yeah i guess a simple error-handler is easier to maintain and understand - so I am all for it :+1:

voxpelli:
Seems like that broke some tests and I didn't have time to look into it 😝

lib/cli/run.js
JoshuaKGoldberg, line 1: [Testing] I think this is functionally looking great. But we should really include some unit test coverage of the changed logic. I would find it spooky if we could make so many changes to it and never have to update unit tests. 👻 

Per https://github.com/mochajs/mocha/actions/runs/8016731201/job/21906168806?pr=5074 at least one unit test is indirectly failing now.

JoshuaKGoldberg (changes requested):
I think just waiting on tests now?

JoshuaKGoldberg (approved):
🚀

voxpelli:
Included in [v10.4.0](https://github.com/mochajs/mocha/releases/tag/v10.4.0) :tada:

PR #5075, opened 2024-01-11
Test Sharding

### LINKED ISSUE #4958: 🚀 Feature: Test Sharding for CI Matrix Purposes
I use CodeceptJS which uses Mocha under the hood to programmatically run end-to-end tests for browsers. We've been increasing the amount of tests slowly and we're running into issues where tests are taking too long to run on a single machine.

There is a concept of test sharding that I propose we add to Mocha. You can read more about this in a few other test runners such as [Playwright](https://playwright.dev/docs/test-parallel#shard-tests-between-multiple-machines) and [Codeception](https://codeception.com/docs/ParallelExecution) .

So if we had 100 tests, and I split into groups of 4, then tell one machine to execute group 1, it should execute 25 tests. One method of determining which test are part of which groups are simply running every 4th test.

We can then use the `matrix` strategy in Github Actions to execute each shard on a different machine in parallel. 

1. We've considered using parallel workers but the machines we're using in CI are very weak but we can spin up a lot of them.
2. We've also considered marking each test with a grep such as "Shard 1: Test Name" and then targetting "Shard 1" with the --grep option. This gives us the flexibility to determine which tests are in which shard, but it is a large burden on the developer to need to decide this all the time.

**Additional context**
I've implemented this in a fork: https://github.com/mochajs/mocha/compare/master...discatech:mocha:test-sharding of Mocha already and would love to get some feedback on whether this is a good change for Mocha itself. I will probably continue to use my forked Mocha until we reach a decision. Thank you so much for considering this!

amerryma:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4958 

## Overview

This strategy uses file-based sharding. The entire set of test files is divided into different groups or "shards". Each shard contains a subset of the test files. The shards use a round robin approach using a simple modulus calculation. The way this works in the code is it simply filters out the tests by the shard value before proceeding to the rest of the mocha code.

A shard option has been added to the command line interface for specifying which shard to run and how many shards there are in total. This value is validated to ensure that bad values aren't inputted.

Sharding does not work when parallel mode is enabled.


JoshuaKGoldberg:
Marking as draft pending discussion in #4958.

JoshuaKGoldberg:
We've got a lot of old PRs, including drafts, in the backlog. It's getting hard to manage 😅. Pending https://github.com/mochajs/mocha/issues/4958#issuecomment-1931094685 I'm going to go ahead and close this old one out. But we can definitely reopen later! Thanks all ❤️

PR #5080, opened 2024-01-18
fix: use original Error object for uncaught errors when provided in browser

### LINKED ISSUE #958: error stack should be displayed for uncaught exceptions in async tests
Currently exceptions in async tests lose their stack and display unfriendly message like:

````
Chrome 28.0.1500 (Linux) simple test FAILED
	Error: Uncaught AssertionError: expected 1 to equal 2 (/home/janotav/anywhere/web-ide/client/core/components/chai/chai.js:505)
	    at global.onerror (/home/janotav/anywhere/web-ide/server/node_modules/mocha/mocha.js:5343:23)
````

This has been addressed recently in the HTML5 spec and can produce nice output like:

````
Chrome 31.0.1605 (Windows 7) simple test FAILED
	AssertionError: expected 1 to equal 2
	    at Assertion.assertEqual (/home/janotav/anywhere/web-ide/client/core/components/chai/chai.js:907:12)
	    at Assertion.ctx.(anonymous function) [as equal] (/home/janotav/anywhere/web-ide/client/core/components/chai/chai.js:3113:25)
	    at /home/janotav/anywhere/web-ide/client/tests/spec/idb-filesystem-spec.js:43:34
````

If the global handler in mocha re-uses the existing Error object like this:

````javascript
    global.onerror = function(err, url, line, colno, erroObj){
        fn(erroObj || new Error(err + ' (' + url + ':' + line + ')'));
    };
````

Currently this extra arguments work in Canary, other browsers are likely to follow. Related blog post:

https://mikewest.org/2013/08/debugging-runtime-errors-with-window-onerror

### PR DESCRIPTION
**Pull request by karyon**
## PR Checklist

- [x] Addresses an existing open issue: fixes ~958 (which is closed but not really fixed)~ #5106

## Overview

In case of uncaught errors, mocha would print only the error message, which is essentially the first line of the stack trace. The actual stack trace was unused. This is essentially https://github.com/mochajs/mocha/pull/2403 which was closed for inactivity.

Here's a fiddle demonstrating the problem: https://jsfiddle.net/Lru1q4n6/ (updated from the PR above). Unfortunately, I don't know how to demonstrate the fix there. Of course it does work locally, I used this fix to debug our tests.

Co-authored-by: @mgiuffrida

### DIFF
--- a/browser-entry.js
+++ b/browser-entry.js
@@ -71,8 +71,11 @@ process.listenerCount = function (name) {
 
 process.on = function (e, fn) {
   if (e === 'uncaughtException') {
-    global.onerror = function (err, url, line) {
-      fn(new Error(err + ' (' + url + ':' + line + ')'));
+    global.onerror = function (err, url, line, colno, error) {
+      if (!error) {
+        error = new Error(err + ' (' + url + ':' + line + ')');
+      }
+      fn(error);
       return !mocha.options.allowUncaught;
     };
     uncaughtExceptionHandlers.push(fn);

### CODE REVIEW

File: browser-entry.js [hunk 1]
```diff
@@ -71,8 +71,11 @@ process.listenerCount = function (name) {
 
 process.on = function (e, fn) {
   if (e === 'uncaughtException') {
-    global.onerror = function (err, url, line) {
-      fn(new Error(err + ' (' + url + ':' + line + ')'));
+    global.onerror = function (err, url, line, colno, error) {
+      if (!error) {
+        error = new Error(err + ' (' + url + ':' + line + ')');
+      }
+      fn(error);
```
JoshuaKGoldberg, line 78: Aha! I'd filed #5106 before seeing this PR. Great. I'll close that one since this once came first.

[Feature] What do you think about including `':' + col` too, per https://github.com/mochajs/mocha/pull/5107/files#diff-c282db01f80ee266af40ae81278d9d99177c763ae242fdeb38ee33c603598a2fR75 ?

Request adding it in as a new nicety, unless you know of a reason not to.

browser-entry.js
JoshuaKGoldberg, line 1: [Testing] Could you add a unit test too, please?

#5107 has a reference if that's helpful.

JoshuaKGoldberg (changes requested):
Lovely! 🙌 

Requesting changes on the unit test & including column.

karyon:
I appreciate your concern, but it would probably much simpler if you just merged your PR. Just close mine, no worries :)

PR #5086, opened 2024-01-24
fix: Add error handling for nonexistent file case with --file option

khoaHyh:
## 🔎 Overview 

Fixes https://github.com/mochajs/mocha/issues/4047

- Added validation to the file(s) passed in to the `--file` option by resolving the argument to an absolute path and check for its existence. We now log a warning and exit if the file does not exist.
  - the errors for `--require` bubble up to the middleware in `lib/cli/run.js`. See [here](https://github.com/mochajs/mocha/blob/master/lib/cli/run.js#L355). This specific error occurs in `lib/cli/run-helpers.js` in the `singleRun` method when we load files asynchronously. Since we were lacking error handling there, the errors appeared the way they previously did before these changes.
- Added a test to assert our changes

Screenshot from 2024-02-07 00-34-30


test/integration/options/file.spec.js
JoshuaKGoldberg, line 1: [Testing] Since the implementation uses `require.resolve`, I think it'd be good to make sure there is at least one test each for CJS and ESM. Using module-specific APIs such as `require` or `import` scares me, even for ones that should always work... Thoughts?

lib/cli/collect-files.js
```diff
@@ -52,6 +53,22 @@ module.exports = ({
...
+    try {
+      require.resolve(fileAbsolutePath);
+    } catch (err) {
+      // this error doesn't bubble up to the middleware like how
+      // --require handles it so we log a warning here and exit
+      const warningMessage = ansi.yellow(
+        `${symbols.warning} Warning: Cannot find test file "${file}" at the path "${fileAbsolutePath}"`
+      );
+      console.warn(`${warningMessage}\n`);
+      process.exit(1);
```
JoshuaKGoldberg, line 68: [Refactor] I don't love having a `process.exit` deep within code. [`n/no-process-exit`](https://github.com/eslint-community/eslint-plugin-n/blob/4778ae86c398f90da13248c2d31bfc4a83cd2dea/docs/rules/no-process-exit.md) has good reasoning around it.

It's also not really a "warning" yellow if the whole thing exists. At that point it's an error (red).

I think a more future-facing, approach would be either:

* (less work, but still partially flawed) have this function throw a special error that any higher-up handler(s) know how to log nicely
* (more work, but more understandable long-term) have this function return an object explaining its results, and have higher-up handler(s) work nicely with it

I haven't looked deeply enough to know whether either or both is feasible. Before we go into a whole deep dive, what do you think about this direction?

JoshuaKGoldberg (changes requested):
Cool! 😎 

Requesting changes on adding a bit more testing. But also let's talk about the direction a bit - I feel nervous adding in `process.exit`/`throw`s deep within code. Thanks for sending!

lib/cli/collect-files.js
khoaHyh: Good to learn about the `no-process-exit`, thanks! I think the latter where we return an object and having the higher-up handler(s) deal with it sounds like the approach here. The error wasn't handled in `run-helpers` so maybe we can add something so that bubbles up to `run.js` like the other methods?

test/integration/options/file.spec.js
khoaHyh: I thought to use `require.resolve` here because the original stack trace showed `require` being used. Alternatively, I could use `fs.existsSync` like how our `--require` flag does it.

lib/cli/collect-files.js
JoshuaKGoldberg: Yeah that approach sounds like a good one to try!

lib/cli/collect-files.js
```diff
@@ -30,7 +32,7 @@ module.exports = ({
   sort,
   spec
 } = {}) => {
-  const unmatched = [];
+  const unmatchedSpecFiles = [];
```
khoaHyh, line 35: Renamed variable for semantic clarity

test/integration/options/file.spec.js
khoaHyh: I removed the use of `require.resolved` and instead opted to use the `path` and `fs` modules to check for file existence instead.

JoshuaKGoldberg: Wait, `require.resolve` is a good thing to use though! If someone is in CJS mode and passes `path/to/place` they'll expect that both `path/to/place.js` and `path/to/place/index.js` would work. Not to mention `.js` vs `.cjs` shenanigans. Right?

I just mean to ask for more testing, not a refactor.

khoaHyh: Oops I misunderstood, yeah that makes sense. I'll add the tests and revert back to using `require.resolve`.

khoaHyh: Before I commit my changes, what do we think of this output?

image

JoshuaKGoldberg: That looks pretty good to me!

Though, would it be a lot of work to reduce it to two lines? That screenshot has two saying nothing found and one mentioning the absolutePath. I wonder if it would be cleaner for users to reduce down to, say...

```plaintext
No test file(s) found with the given pattern(s), exiting with code 1
  absolutePath: /home/ka9/dev/mocha/nonexistent.js
```

...if that's a lot of work to reorder to, then I'd say let's treat that as a followup issue and not a blocker here.

khoaHyh: I reduced it to two lines but I kept the absolute path on the same line as the pattern statement in case there are multiple files passed in. Here the `--file` argument is treated as an array when the values are comma-separated.

image

JoshuaKGoldberg: Beautiful! I'm happy with that. Nice 🔥

khoaHyh: Pushed changes and added tests for esm + cjs!

khoaHyh (commented):
@JoshuaKGoldberg tagging for re-review just in case this disappeared into the backlog abyss 😅

JoshuaKGoldberg:
Oh! It did, thank you!

lib/cli/collect-files.js
```diff
@@ -92,4 +118,10 @@ module.exports = ({
  * @property {string[]} file - List of additional files to include
  * @property {boolean} recursive - Find files recursively
  * @property {boolean} sort - Sort test files
+ * @typedef {Object} UnmatchedFile - Diagnostic object containing unmatched files
```
JoshuaKGoldberg, line 121: [Docs] I believe it's standard to have a separate JSDoc tag for each object declaration?

```suggestion
 */
 
/**
 * @typedef {Object} UnmatchedFile - Diagnostic object containing unmatched files
```

lib/cli/run-helpers.js
```diff
@@ -140,9 +157,27 @@ const singleRun = async (mocha, {exit}, fileCollectParams) => {
...
+        ansi.yellow(
+          `Warning: Cannot find any files matching pattern "${pattern}" at the absolute path "${absolutePath}"`
+        )
+      );
+    });
+    console.log(
+      'No test file(s) found with the given pattern, exiting with code 1'
+    );
+    return mocha.run(exitMocha(1));
+  }
```
JoshuaKGoldberg, line 174: [Refactor] This is about the same as the previous block. Up for extracting out a function?

Having two areas of code that are roughly the same isn't the worst thing in the world, but given how they're pretty much exactly identical - I think it'd make the code more readable & easy to understand. Plus if we have to change it later, it'll be easy to forget one.

test/integration/options/file.spec.js
```diff
@@ -64,4 +66,82 @@ describe('--file', function () {
...
+          `Warning: Cannot find any files matching pattern`
+        ).and('to contain', nonexistentTestFileArg);
+        done();
+      },
+      {stdio: 'pipe'}
+    );
+  });
+
+  it('should provide warning for nonexistent cjs file extensions', function (done) {
+    const nonexistentCjsArg = 'nonexistent.test.js';
```
JoshuaKGoldberg, line 105: [Naming] This confused me:

> nonexistent cjs file extensions
> `.test.js`

Is the intent that it's generally testing CommonJS files (`.js`)? I think the name could be a bit more clear. Maybe ...

```suggestion
  it('should provide warning for nonexistent js file extensions', function (done) {
    const nonexistentCjsArg = 'nonexistent.test.js';
```

test/integration/options/file.spec.js
```diff
@@ -64,4 +66,82 @@ describe('--file', function () {
...
+    runMochaJSON(esmFile, testArgs, function (err, res) {
+      if (err) {
+        return done(err);
+      }
+      expect(res, 'to have passed');
+      done();
+    });
+  });
+
+  it('should log a warning if a nonexistent file is specified', function (done) {
```
JoshuaKGoldberg, line 83: [Naming]

```suggestion
  it('should log a warning if a nonexistent file with an unknown extension is specified', function (done) {
```

JoshuaKGoldberg (approved):
This looks great to me, thanks! I have no gripes about the code changes themselves and appreciate the clarity of the new `unmatchedSpecFiles` name. 😄

Will wait for another maintainer to review. In the interim, just leaving a few small nitpicks, nothing I'm passionate about.

lib/cli/collect-files.js
khoaHyh: TIL, thanks! Made changes to have separate tags for each object declaration.

lib/cli/run-helpers.js
khoaHyh: Yeah that makes sense to me 👍🏻. I extracted the common code to a function `handleUnmatchedFiles` that is now called both by `singleRun` and `parallelRun` instead of the previous duplication of code.

test/integration/options/file.spec.js
khoaHyh: Yep just general testing of common JS files. Changed!

lib/cli/collect-files.js
```diff
@@ -44,14 +45,34 @@ module.exports = ({
...
     }
   }, []);
 
+  // check that each file passed in to --file exists
+
+  const unmatchedFiles = [];
+  fileArgs.forEach(file => {
+    const fileAbsolutePath = path.resolve(file);
+    try {
+      require.resolve(fileAbsolutePath);
```
voxpelli, line 62: I would suggest this comment to explain the difference to https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/lib/cli/lookup-files.js#L76-L77

```suggestion
      // Used instead of fs.existsSync to ensure that file-ending less files are still resolved correctly
      require.resolve(fileAbsolutePath);
```

voxpelli (approved):
Would be lovely to get the docs updated to properly mention how `--file` works as well (somewhat outside of scope, I know): https://github.com/mochajs/mocha/blob/2f3fedcc41cbb9d3e503d84098fcc07d7c3c49f1/docs/index.md#L1064

As docs claim `--file` supports directories and globs as well, which it doesn't

lib/cli/collect-files.js
khoaHyh: For the docs: I removed the `directory|glob` portion so that it only shows `--file <file>` now

- not sure if you were thinking of adding anything else so I only removed the unsupported file arguments

For the comment in `mocha/lib/cli/collect-files.js`: I added the comment you suggested

voxpelli (approved):
Looks good! Thanks 👍 🎉 🙏

voxpelli (changes requested):
Looks like one test is failing, can you take a look?

```
expected '\n Exception during run: TypeError: Cannot read properties of undefined (reading \'length\')\n    at handleUnmatchedFiles (/home/runner/work/mocha/mocha/lib/cli/run-helpers.js:119:22)\n    at singleRun (/home/runner/work/mocha/mocha/lib/cli/run-helpers.js:151:12)\n    at exports.runMocha (/home/runner/work/mocha/mocha/lib/cli/run-helpers.js:231:10)\n    at exports.handler (/home/runner/work/mocha/mocha/lib/cli/run.js:370:11)\n    at /home/runner/work/mocha/mocha/node_modules/yargs/build/index.cjs:443:71\n'
to contain 'Warning: Cannot find any files matching pattern'


 Exception during run: TypeError: Cannot read properties of undefined (reading 'length')
    at handleUnmatchedFiles (/home/runner/work/mocha/mocha/lib/cli/run-helpers.js:119:22)
    at singleRun (/home/runner/work/mocha/mocha/lib/cli/run-helpers.js:151:12)
    at exports.runMocha (/home/runner/work/mocha/mocha/lib/cli/run-helpers.js:231:10)
    at exports.handler (/home/runner/work/mocha/mocha/lib/cli/run.js:370:11)
    at /home/runner/work/mocha/mocha/node_modules/yargs/build/index.cjs:443:71
```

khoaHyh:
> Looks like one test is failing, can you take a look?

That was my bad, I forgot to pass the mocha instance as a parameter to the `handleUnmatchedFiles` helper function. I also see the `netlify/mocha/deploy-preview` is failing, anything on my end I should do?

JoshuaKGoldberg:
>  I also see the netlify/mocha/deploy-preview is failing, anything on my end I should do?

You can ignore that - it's a separate issue from this PR on our end.

JoshuaKGoldberg (approved):
Looks great, thanks for the fix!

David Levinson in Independence Day performing a checkmate in chess and saying "Checkmate"

As before, the deploy preview failure is unrelated. It's 429 errors from us running more CI jobs than the service is used to.

JoshuaKGoldberg:
Note: I switched this to a `fix:` as it's more of a fix around improper inputs than a fully fledged feature. It doesn't specifically need a minor version bump. A patch would suffice.

voxpelli:
Thanks for stepping in and merging @JoshuaKGoldberg 🙏 🙂

Issue #5055, opened 2023-12-27
🛠 Repo: Set up Renovate


JoshuaKGoldberg:
### Tooling Suggestion Checklist

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have pulled the latest `master` branch of the repository.
- [X] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/master/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/master/.github/CONTRIBUTING.md)
- [X] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [X] I want to provide a PR to resolve this

### Overview

There are quite a few issues on this repo right now tracking out-of-date packages. We're going to enable [Renovate](https://github.com/renovatebot/renovate) to automatically create PRs to update packages.

Notes from @boneskull:

> Looks reasonable. The only thing I’d suggest with the automated tooling (something like Renovate, yeah?) is to have a human in the loop for production dependency upgrades—check them closely. Mocha has been bitten several times due to semver violations in its dependency tree. I would also recommend using https://socket.dev/ as well (should be free for OSS). While Mocha hasn’t been hit by a malicious dep (to my knowledge), historical performance is not a guarantee of future results. 😄

JoshuaKGoldberg:
We'd looked at this a few times last year. Unfortunately, the old docdash-based website has a ton of outdated dependencies that make this rather difficult. If we were to set up Dependabot/Renovate today then we'd have to exclude a heck of a lot of dependencies from it.

#5207 -> #5246 tracks making a new website on a more modern stack (Astro). Once the new website fully replaces the old one then we'll be free of the ancient docdash dependencies. In the meantime, marking this as blocked. 🛑

mark-wiemer:
Assigning myself to track this one, I'm OK excluding dependencies in the short-term if it means at least some automated upgrade PRs are made available :) @JoshuaKGoldberg OK marking as unblocked? We can update the "archive old site" issue to say "and also cleanup the Renovate ignore list" or similar

JoshuaKGoldberg:
That sounds reasonable to me.

[...]

Issue #5027, opened 2023-11-14
Project Status: Maintenance Reboot


JoshuaKGoldberg:
👋 Hi all! @Uzlopak, @voxpelli, and I are happy to announce that we've picked up maintenance of the Mocha organization and packages. That primarily includes the [`mocha`](https://www.npmjs.com/package/mocha) test framework.

Shoutout and thanks to @boneskull for inviting us on and helping set up many of the documentation and processes still in use today, as well as the other recent Mocha core team members: @giltayar, @juergba, @Munter, and @outsideris.

## Context

Mocha was not actively maintained for most of 2023. It was maintained off-and-on through 2022. This is natural in open source projects: because they are run by volunteers, maintenance often is dependent on personal voluntary availability in out-of-work hours.

The three of us "reboot" maintainers are joining on to:

* Get the project back into active maintenance, including regularly triaged issues and reviewed pull requests
* Update the project's tooling to reduce the overhead of maintenance
* Update Mocha as needed to stay up-to-date with the surrounding JavaScript ecosystem

### Project Charter

Mocha has been a core part of the JavaScript ecosystem for over a decade. Per its [`PROJECT_CHARTER.md`](https://github.com/mochajs/mocha/blob/ee2b70e8e82b0b67347d41028fa45521d43633d9/PROJECT_CHARTER.md):

> **Mocha** is a unopinionated, general-purpose testing framework for the JavaScript community. **Mocha** favors flexibility over rigidity, stability over disruption, and deliberation over agility. **Mocha** aims to do one thing, and do it well. **Mocha** will strive to evolve with the needs of the community, but will make every effort not to abandon its current users in doing so.

We intend to keep to this project charter. Mocha is positioned differently from other popular test frameworks in the ecosystem. Notably:

* Higher-level frameworks [Jest](https://jestjs.io) and [Vitest](https://vitest.dev) provide more abstractions and utilities on top of testing
* [Node's built-in runner](https://nodejs.org/api/test.html) integrates closely with Node itself at a lower level

We aim to keep Mocha positioned as a stable test framework, minimizing breaking changes or significant new feature additions. Our priority is for the existing users who do not want to be broken by or have to spend time reacting to changes.

## Next Steps

Our current plan is to:

- [x] Discuss this pinned issue and announce it
- [ ] Tidy up the existing contributing and maintenance documentation and processes, such as issue and PR templates
[...]

PR #5090, opened 2024-02-08
chore: update chokidar to 3.6.0

orgads:
## Overview

Update chokidar to 3.6.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "ansi-colors": "4.1.1",
         "browser-stdout": "1.3.1",
-        "chokidar": "3.5.3",
+        "chokidar": "^3.6.0",
         "debug": "4.3.4",
         "diff": "5.0.0",
         "escape-string-regexp": "4.0.0",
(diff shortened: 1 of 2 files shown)


voxpelli:
@orgads There are lots of outdated dependencies in this project, to help us prioritize, can you explain the reasoning for updating this specific one?

JoshuaKGoldberg:
I think the not using `^` versions is the more curious point to investigate. Filed #5114.

Marking this one as draft pending discussion there.

Thanks for pointing this out!

voxpelli:
> nest-cli updated to 3.6.0 (pinned), and it uses mocha.

@orgads And the benefit of updating here is that the `chokidar` in `nest-cli` and in `mocha` can be deduped by npm and thus you get fewer individual versions in your `node_modules`?

Or any more urgent issue?

orgads:
It's been a month 😅

IIRC there was a Typescript compatibility issue too, but I'm unsure.

voxpelli:
All dependencies will be updated in due time as we move further in our reboot of the project (https://github.com/mochajs/mocha/issues/5027) and eventually be done so with automated PR:s: https://github.com/mochajs/mocha/issues/5055

But if any dependency is causing urgent issues then it of course requires more urgent attention, hence my questions 🙂

voxpelli:
Closing, we will update dependencies together with Mocha 11, #5216, no need for separate PR:s.

PR #5095, opened 2024-02-15
chore: revert #5069 to restore Netlify builds

voxpelli:
`assetgraph-builder` is used through at least the `buildProduction` command

This reverts commit 53a4bafbdeb32576440b0a21787f2525585411c0.

## PR Checklist

- [x] ~~Addresses an existing open issue: fixes #000~~
- [x] ~~That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)~~

## Overview

Reverts #5069 to restore Netlify builds by adding back `assetgraph-builder`

Downside: Installs on non-x86 machines again require `--ignore-scripts`: https://github.com/mochajs/mocha/issues/4828#issuecomment-1035204506


package.json
```diff
@@ -81,7 +81,9 @@
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-multi-entry": "^4.0.1",
     "@rollup/plugin-node-resolve": "^13.1.3",
+    "assetgraph-builder": "^9.0.0",
     "autoprefixer": "^9.8.6",
+    "canvas": "^2.11.2",
```
JoshuaKGoldberg, line 86: Do we need to add `canvas` back too? Can we just revert the `assetgraph-builder` removal?

JoshuaKGoldberg (changes requested):
I think we can keep the removal of `canvas`? Otherwise LGTM (alas)

package.json
voxpelli: `assetgraph-builder` includes the `canvas` as well, so it didn't make any difference in terms of annoyance

JoshuaKGoldberg: Blurgh, 👍

JoshuaKGoldberg (approved):

JoshuaKGoldberg (changes requested):
Or, actually, wait - `docs/*` doesn't need to be reverted. It doesn't need to use `canvas`.

PR #5099, opened 2024-02-22
spec-reporter prints retries

### LINKED ISSUE #5014: Should the spec reporter report retries?
### Prerequisites

### Description
We have a few flaky tests and we would like to be aware of them. Currently, the spec reporter does not make a distinction between a test that succeeded first try and one that needed to be retried. We are probably going to create a custom reporter that extends `Mocha.reporters.Spec` to report the flaky tests at the end, but if it can be upstreamed, that would be ideal.

If the idea has good reception with mocha maintainers, I could try making a PR.

CheadleCheadle:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5014

## Overview

I've implemented a function called logRetries() which displays the number of the current retry along with the total number of retries, provided that the total number of retries is greater than zero. This output is formatted below the test title and highlighted in yellow as requested for ecosystem consistency. I would greatly appreciate any feedback on this addition!
Example:
retries


lib/reporters/spec.js
```diff
@@ -67,20 +67,35 @@ function Spec(runner, options) {
...
+      logRetries();
     } else {
       fmt =
         indent() +
         color('checkmark', '  ' + Base.symbols.ok) +
         color('pass', ' %s') +
         color(test.speed, ' (%dms)');
       Base.consoleLog(fmt, test.title, test.duration);
+      logRetries();
     }
```
JoshuaKGoldberg, line 99: [Refactor] Both the `if` and the `else` blocks end with `logRetries()`. So the first can be removed:

```suggestion
    }
    
    logRetries();
```

lib/reporters/spec.js
```diff
@@ -67,20 +67,35 @@ function Spec(runner, options) {
   });
 
   runner.on(EVENT_TEST_PASS, function (test) {
+    function logRetries() {
+      if (test._retries > 0) {
```
JoshuaKGoldberg, line 71: [Refactor] Inside this `if`, `Base.consoleLog` only gets called when `test._currentRetry > 0`. Why not make that the condition here?

Then there'd be no need for the `? :` ternary or the `if (retryMessage)`.

bin/mocha.js
JoshuaKGoldberg, line 1: [Question] Why the permissions change?

My guess is it's unintentional & unnecessary for this PR.

lib/reporters/spec.js
JoshuaKGoldberg, line 1: [Design] From https://github.com/mochajs/mocha/issues/5014#issuecomment-1902514428 -> https://github.com/vitest-dev/vitest/pull/1929, Vitest has a bit more succinct of a report:

```ts
import { it } from "vitest";

let value = 0;

it(
	"example",
	() => {
		if (!value++) {
			throw new Error("nah");
		}
	},
	{ retry: 2 },
);
```

```plaintext
 ✓ src/greet.test.ts (1)
   ✓ example (retry x1)
```

I'd meant to suggest that this should use the same display as Vitest's. Vitest's doesn't add an extra line, which is nice for keeping the display succinct.

Thoughts?

JoshuaKGoldberg (changes requested):
Hello again! 😄 

Left some thoughts - I'm excited to see this come in!

lib/reporters/spec.js
CheadleCheadle: I believe I added it on a new line because I would have to modify some tests to expect it. I can make the change, however.

lib/reporters/spec.js
CheadleCheadle: Good catch!

CheadleCheadle (commented):
Refactored to include retry on same line`(retry x1)`.

lib/reporters/spec.js
JoshuaKGoldberg, line 1: @CheadleCheadle There are some unnecessary whitespace changes to this file, could you please revert them to make it easier to review?

lib/reporters/spec.js
```diff
@@ -45,46 +45,60 @@ function Spec(runner, options) {
...
-  runner.on(EVENT_TEST_PASS, function (test) {
+  runner.on(EVENT_TEST_PASS, function(test) {
+    function logRetries() {
+      if (test._currentRetry > 0) {
+        /*
+          Included message within format so it can be passed into `Base.consoleLog` as one argument
+        */
+        var retryFmt = color(
+          'bright yellow',
+          `  (retry x${test._currentRetry})`
```
JoshuaKGoldberg, line 77: [Question] Is it intentional that this has two spaces in front of it, not one?

JoshuaKGoldberg (changes requested):
Wasn't sure if you're ready to re-request review, but just in case 🙂

lib/reporters/spec.js
CheadleCheadle: It was a stylistic choice I made. However, one space seems to look better. Will change.

lib/reporters/spec.js
```diff
@@ -67,20 +67,34 @@ function Spec(runner, options) {
   });
 
   runner.on(EVENT_TEST_PASS, function (test) {
+    function logRetries() {
+      if (test._currentRetry > 0) {
+        /*
+          Included message within format so it can be passed into `Base.consoleLog` as one argument
+        */
+        var retryFmt = color(
+          'bright yellow',
```
voxpelli, line 76: To really stick with the theme I think we should add a `'retries'` color here:

https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/base.js#L64-L90

And possibly stick even more with the style of eg. durations and resolve to the color in `base.js`:

https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/base.js#L369-L377

And maybe do a similar style and have two levels: `'some retries'` / `'many retries'` With only the latter resolving to bright yellow

lib/reporters/spec.js
```diff
@@ -67,20 +67,34 @@ function Spec(runner, options) {
   });
 
   runner.on(EVENT_TEST_PASS, function (test) {
+    function logRetries() {
+      if (test._currentRetry > 0) {
```
voxpelli, line 71: The underscore in `_currentRetry` is an old-school practice for indicating that its a private property, so we shouldn't rely on it here I think as its not intended for outside consumption.

~~If we want it to be exposed we should make it an explicit public property, like `duration`, and probably name it `retries` or `retryCount` or such~~

We should use the public [`.currentRetry()`](https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/runnable.js#L185-L195) instead.

voxpelli (changes requested):
Some feedback on the implementation.

On top of that: The [`bin/mocha.js`](https://github.com/mochajs/mocha/pull/5099/files#diff-6ecfc053a45484b7eac23ba8a40b9644f57dc2fc84f807038bb0f5ce5eeeff09) has an unrelated permission change.

JoshuaKGoldberg:

JoshuaKGoldberg:
Closing out to free up the backing issue. @CheadleCheadle if you end up seeing this and having time again, no worries, feel free to re-open the PR or send a new one.

Cheers all! 🤎

PR #5110, opened 2024-03-03
chore: switch two-column list styles to be opt-in

### LINKED ISSUE #3702: 🛠️ Repo: Two-column lists in the docs should not be the default style for lists
This should not be the default list style:

image

Change the stylesheet to display a "normal" list unless we apply a specific class for two-column display.  I think it's OK to use in some places, but not for *all* lists.

unless there's a way to specify the class in our markdown compiler (not sure), just use html for the ones we keep.  features and table of contents are good to keep... not sure about the rest.

For implementors:

To see changes locally, run `npm start docs` and open up `docs/_dist/index.html` with your browser. You can also do `npm start docs.watch` to rebuild as you work (you'll still have to reload).

marjys:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3702

## Overview

This PR:
- Adds the two-column class with the expected style for two-column unordered lists, making sure to keep the two-column style at Features and Table of Contents
- Removes column style from ul element scope (unneeded style)


docs/css/style.css
```diff
@@ -180,15 +180,18 @@ a.direct-link {
 
 ul {
   box-sizing: content-box;
-  column-count: 2;
-  column-gap: 30px;
   margin-top: 20px;
   padding: 0 15px;
 }
 
+h2#table-of-contents + ul,
```
JoshuaKGoldberg, line 187: [Refactor] The `table-of-contents` ID is auto-generated. We won't want to take a dependency on it existing as-is. It's also kind of coincidental to assume the list immediately following the h2 table of contents should be two-column.

```suggestion
```

Could you add the `two-column` class to the list instead? I.e. add another `{:.two-column}`?

docs/css/style.css
```diff
@@ -180,15 +180,18 @@ a.direct-link {
 
 ul {
   box-sizing: content-box;
-  column-count: 2;
-  column-gap: 30px;
   margin-top: 20px;
   padding: 0 15px;
```
JoshuaKGoldberg, line 184: [Refactor] The `.single-column` class below overrides the `margin-top` and `padding` styles too. Is there a reason not to move these specific values into `.two-column` as well?

JoshuaKGoldberg (changes requested):
Functionally looks good, nicely done! 🎨 

Requesting changes on a bit of refactoring to make the styles more clear.

docs/css/style.css
marjys: Well noticed, thanks! Fixed on [64d8788](https://github.com/mochajs/mocha/pull/5110/commits/64d8788f01a4348a9c7788a39fe2b79e2bef5936)

marjys:
@JoshuaKGoldberg The 'Deploy Preview' check has failed even before I committed the requested changes (when I merged master to my branch), but I'm not authorized to open Netlify to check the error. Is there another way for me to verify that? Thanks!

JoshuaKGoldberg:
...huh, that's a new one.

```plaintext
8:01:04 PM: 2024-03-06T01:01:04.355Z mocha:docs:data:supporters supporter image pull completed
8:01:04 PM: [11ty] Problem writing Eleventy templates:
8:01:04 PM: [11ty] 1. Having trouble rendering liquid template ./docs/api/reporters_base.js.html (via TemplateContentRenderError)
8:01:04 PM: [11ty] 2. unexpected token at ": string, msg...", file:./docs/api/reporters_base.js.html, line:273, col:12 (via ParseError)
8:01:04 PM: [11ty] 3. unexpected token at ": string, msg..." (via AssertionError)
8:01:04 PM: [11ty] Wrote 0 files in 3.01 seconds (v1.0.2)
8:01:04 PM: The script called "docs" which runs "nps docs.clean && nps docs.api && eleventy && nps docs.linkcheck" failed with exit code 1 https://github.com/sezna/nps/blob/master/other/ERRORS_AND_WARNINGS.md#failed-with-exit-code
```

Does that happen for you? I'm traveling right now and blocked from trying it locally on https://github.com/mochajs/mocha/issues/5113.

marjys:
I can see this error locally now! It seems to be happening due to Eleventy not being able to parse the properties of an object on the Jsdocs api at `/lib/reporters/base.js`, line 230 (didn't catch it before as I was running `npm start docs.watch` and it does not build before serving, so I was serving an old version of my build, my fault).
I was thinking about some ways to fix this:
1. Simply replace `{ message: string, msg: string, stack: string }` with `Object`
2. As we know what the Object structure is, we could add a type definition to it with something like the following, and then replacing the object with FullErrorStack:
```javascript
 * @return {FullErrorStack}

...

/**
 * An object containing the full error stack results
 * @private
 * @typedef {Object} FullErrorStack
 * @property {string} message
 * @property {string} msg
 * @property {string} stack
 */
```

So please let me know your preferences and whether we should be opening those changes in another PR or we could just apply them on this one.
Thanks!

JoshuaKGoldberg:
Nice investigation! 👏 

Does that happen for you on the `master` branch? If does (i.e. is not unique to this PR) then yeah, that'd be a separate issue & PR. Yes please!

If it's unique to this branch (?!) then it'd be something to fix here.

marjys:
The error happens on master too:

Screenshot 2024-03-07 at 11 29 39

Ok, I'll be opening an issue for this. Thanks a lot!

marjys:
Requested changes were applied and all checks are passing after fix on #5116.

JoshuaKGoldberg (approved):
Michael Fassbender calmly saying "perfection"

JoshuaKGoldberg:
Thanks @marjys!

Issue #5114, opened 2024-03-04
📦 Package: Switch dependency versions to ^ ranges


JoshuaKGoldberg:
Spinning out of #5090: @orgads noted that the `package.json` versions of `dependencies` are all pinned to specific versions like `4.1.1` rather than "caret" `^` ranges like `^4.1.1`:

https://github.com/mochajs/mocha/blob/3345eff154d40802ba4470b73fd0040c15f6c4f9/package.json#L53-L56

Why is that?

I'm accustomed to `^` ranges to help consumers deduplicate packages. E.g. if a consumer's package requirements are `chokidar@^3.5.2` and `chokidar@^3.6.0`, us specifying `chokidar@^3.5.3` would mean they could all resolve to the same package version.

voxpelli:
I'm a big 👍 to this. It was different in the pre-`package-lock.json` era, that's when it was good practice to try and lock down dependencies this way, now its better handled by the `package-lock.json` in our and other's projects.

Maybe implement this on a dependency by dependency basis when we update them? That way we will test that no breakage will occur

JoshuaKGoldberg:
No comments for a while, and two 👍 votes. Accepting PRs!

Let's have a single PR for each dependency so we can test them separately - and revert separately if needed.

the-sinner:
Hi, as far i understand, you want to add a caret symbol after each dependency below (except chowkidar) in separate PRs ?
https://github.com/mochajs/mocha/blob/472a8be14f9b578c8b1ef3e6ae05d06fc2d9891b/package.json#L53-L74

the-sinner:
So, while reading about the caret symbol in package.json, i found this stack overflow comment 

> Posting here to hopefully catch people that don't quite think this through, but both ^ and ~ assumes you can trust minor and point releases from your dependencies. If you are publishing a library and want other people to trust you, DO NOT BLINDLY ACCEPT DOWNSTREAM DEPENDENCIES. A bad dot release from your dependency can cause a chain reaction upstream, and will have people knocking at YOUR door when things go pear shaped. This is another huge reason to use npm shrinkwrap on your production code. - tehfoo

[Comment](https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json#comment45166739_22345808)

So, it should be fine, right ?

JoshuaKGoldberg:
[...]

PR #5150, opened 2024-05-30
feat: allow ^ versions for character encoding packages

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes part of #5114

## Overview

Adds a `^` range to the following packages related to character encoding in logs:

* `ansi-colors`: new minor versions available, up to `4.1.3`
* `browser-stdout`: hasn't been updated in 6 years
* `chokidar`: a new minor version available, `3.6.0`
* `supports-color`: no new _minor_ versions available - just a new major


voxpelli (changes requested):
The one change looks indeed harmless: https://github.com/doowb/ansi-colors/compare/4.1.1...4.1.3

But can you do a new `npm install ansi-colors@^4.1.3` to ensure that our lockfile actually also references it?

package-lock.json
```diff
@@ -24087,9 +24087,9 @@
       }
     },
     "ansi-colors": {
-      "version": "4.1.1",
-      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-4.1.1.tgz",
-      "integrity": "sha512-JoX0apGbHaUJBNl6yF+p6JAFYZ666/hhCGKN5t9QFjbJQKUU/g8MNbFDbvfrgKXvI1QpZplPOnwIo99lX/AAmA=="
+      "version": "4.1.3",
+      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-4.1.3.tgz",
+      "integrity": "sha512-/6w/C21Pm1A7aZitlI5Ni/2J6FFQN8i1Cvz3kHABAAbw93v/NlvKdVOqz7CCWz/3iv/JplRSEEZ83XION15ovw=="
```
JoshuaKGoldberg, line 24092: @voxpelli from https://github.com/mochajs/mocha/pull/5150#pullrequestreview-2091274231:

> The one change looks indeed harmless: [doowb/ansi-colors@4.1.1...4.1.3](https://github.com/doowb/ansi-colors/compare/4.1.1...4.1.3)
> 
> But can you do a new `npm install ansi-colors@^4.1.3` to ensure that our lockfile actually also references it?

This brings up an interesting point: we can either:
* include the newer `4.1.3`, and miss out on testing `4.1.1`
* keep the older `4.1.1`, and miss out on testing `4.1.3`

🤷 I don't mind either way, and do like keeping to the latest internally. Just thought I'd note it. I think we're on the same page 😄

voxpelli: If we do `^4.1.3` we get an update as well, then we need a lockfile maintenance job to ensure we continuously test against what our users are getting: https://docs.renovatebot.com/configuration-options/#lockfilemaintenance

package.json
```diff
@@ -51,7 +51,7 @@
     "test:smoke": "node ./bin/mocha --no-config test/smoke/smoke.spec.js"
   },
   "dependencies": {
-    "ansi-colors": "^4.1.3",
+    "ansi-colors": "^4.1.1",
```
voxpelli, line 54: Why keep it the same? I think we should bump it?

JoshuaKGoldberg: Oh, oops, will revert any actual version drops.

JoshuaKGoldberg: Discussed with @Uzlopak in a regular meeting: yes, we want to make sure the version we test with (listed in our `package-lock.json`) is the minimum version users receive. 

I was initially hesitant about this because I didn't want to force users to upgrade in case that's a blocker for some and/or causes duplicate package versions. But the making sure we're not using a newer version that lacks some of the features of a user's older version convinced me. 😄

package-lock.json
JoshuaKGoldberg: Updated to the latest. 👍

JoshuaKGoldberg:
Released in [`mocha@v10.6.0`](https://github.com/mochajs/mocha/releases/tag/v10.6.0). 🚀

PR #5157, opened 2024-06-07
chore: allow blank issues

voxpelli:
I personally think we should allow for blank issues even though we might prioritize them lower. Not all issues fit into pre-defined templates and I know myself that I get very frustrated when I'm being forced to use a template that might not fit the issue that I need to open with the team.

I can take it upon myself to triage all the blank issues if that's the sticking point.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: Documentation Website
     about: Please read our documentation website before filing new issues.


voxpelli:
> [Coverage Status](https://coveralls.io/builds/67954295)
> 
> coverage: 94.359%. remained the same when pulling **[1292e99](https://github.com/mochajs/mocha/commit/1292e99ec32667fcdff0bcaf1c390bb149472bb4) on voxpelli/allow-blank-issues** into **[5b7af5e](https://github.com/mochajs/mocha/commit/5b7af5eec2098f094fe1601b0c5b85499fa67828) on master**.

We should move the coverage check to be an actual GitHub check and not a comment. Its less spammy

JoshuaKGoldberg (approved):
👍 I'm up for this!

Issue #5156, opened 2024-06-04
🐛 Bug: Hit an error with CJS named exports


khteh:
### Bug Report Checklist

- [X] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/master/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/master/.github/CONTRIBUTING.md)
- [X] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [X] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

No runtime error

### Actual

I have the following `src/infrastructure/db.ts` file:
```
@injectable()
export class Database {
}
```
and `src/infrastructure/index.ts`:
```
import config from 'config';
export * from "./db"
```
This file is referenced in `src/webapi/routes/api.ts` and hit the following error using rollup:

```
 Exception during run: file:///usr/src/Node.JSRestAPI/src/webapi/routes/api.ts:13
import { StudentRepository, TeacherRepository, LoggerImpl, DatabaseTypes, Database } from "infrastructure";
                                                                          ^^^^^^^^
SyntaxError: Named export 'Database' not found. The requested module 'infrastructure' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'infrastructure';
const { StudentRepository, TeacherRepository, LoggerImpl, DatabaseTypes, Database } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:171:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:254:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:475:24)
    at async formattedImport (/usr/src/Node.JSRestAPI/node_modules/mocha/lib/nodejs/esm-utils.js:9:14)
    at async Object.exports.requireOrImport (/usr/src/Node.JSRestAPI/node_modules/mocha/lib/nodejs/esm-utils.js:42:28)
[...]

PR #5160, opened 2024-06-19
Update esm-utils.js

musanmaz:
Fix CommonJS named export error in esm-utils (Issue #5156)

This PR addresses the issue described in #5156, where a named export error occurs with CommonJS modules. The following changes have been made:

- Modified `formattedImport` function in `esm-utils.js` to ensure proper handling of CommonJS module imports.
- Updated `requireOrImport` to use `dealWithExports` for CommonJS modules.

### DIFF
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -50,7 +50,7 @@ exports.requireOrImport = async (file, esmDecorator) => {
         // Importing a file usually works, but the resolution of `import` is the ESM
         // resolution algorithm, and not the CJS resolution algorithm. We may have
         // failed because we tried the ESM resolution, so we try to `require` it.
-        return require(file);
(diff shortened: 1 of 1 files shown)


Uzlopak:
I dont see the changes you mention in your original post

JoshuaKGoldberg (changes requested):
👋 the linked issue isn't accepting PRs, so we can't take this PR in. You can read more about this in our [`.github/CONTRIBUTING.md` file](https://github.com/mochajs/mocha/blob/3a4c4b71759b3ca6cd80a31052ea606ff4475ace/.github/CONTRIBUTING.md). Marking as a draft in the meantime.

FWIW I think this fix makes sense if we can have a clean reproduction on the issue... though we will also need a test to validate it actually fixes the issue. 🙂

PR #5177, opened 2024-07-15
ci: github actions workflow improvements

voxpelli:
## Overview

The major benefit is that the workflows are lot DRY:er now – main definition is only 40 lines of code since all the repetitive boilerplate is ripped out into a reusable part: https://github.com/mochajs/mocha/blob/19ae1455109e7cd32db73ef6c3714e36dc045c51/.github/workflows/mocha.yml#L16-L56

This also ensures that the caching, installation etc will behave consistently in all the places and makes it easy to evolve it in a consistent manor.


.github/workflows/mocha.yml
```diff
@@ -1,205 +1,56 @@
...
+    uses: ./.github/workflows/npm-script.yml
+    with:
+      os: 'ubuntu-latest,windows-2019'
+      node-versions: '14,16,18,20,22'
+      npm-script: test:smoke
 
   test-node:
-    name: 'Node.js [v${{ matrix.node }} / ${{ matrix.os }}]'
+    # TODO: Restore "mocha-github-actions-reporter" style reporting without relying on third party module
+    uses: ./.github/workflows/npm-script.yml
```
JoshuaKGoldberg, line 31: [Question] Why remove the `MOCHA_REPORTER: mocha-github-actions-reporter`?

.github/workflows/npm-script.yml
JoshuaKGoldberg, line 1: [Praise] Big +1 from me on making a shared sript for this.

.github/workflows/npm-script.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Coveralls Parallel
+        if: always() && inputs.coverage
+        uses: coverallsapp/github-action@v2
+        with:
+          flag-name: run-${{ join(matrix.*, '-') }}
+          parallel: true
+
+  upload-coverage:
+    needs: script
+    if: always() && inputs.coverage
```
JoshuaKGoldberg, line 88: [Question] Why the `always() &&`? I'm seeing discussions (https://github.com/orgs/community/discussions/25789, https://github.com/orgs/community/discussions/26303) around how it makes the job un-cancellable, but nothing else on what it actually does for us?

JoshuaKGoldberg (approved):
🚀 This is fantastic, thanks for the cleanup!

Nothing blocking from me. Just a couple of questions.

.github/workflows/mocha.yml
voxpelli: Because it was being installed within the workflow rather than being a dev dependency of the `package.json`, which makes versioning of it non-straightforward, almost felt like a backdoor

Also felt a bit odd to have all the other kinds of reporters built in but not this – if we need it then so do others, we should dogfood it (and possibly offer to upstream it)

.github/workflows/npm-script.yml
voxpelli: Because it should always upload coverage reports, no matter if any of the tests failed. Else it would only update coverage report when all tests passes

PR #5185, opened 2024-08-05
ci: update to `windows-latest` in actions

voxpelli:
Ever since the original move to GitHub Actions almost 4 years ago its been `windows-2019` that's been in use.

I'm thinking we should always use `-latest` when possible


Uzlopak (approved):

Uzlopak:
I think so too.

PR #5191, opened 2024-08-06
docs: improve filtering

voxpelli:
Went through and created a somewhat better filter. A good stopgap measure at least.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -46,8 +46,6 @@ module.exports = function (eleventyConfig) {
 
   eleventyConfig.setLibrary('md', markdown);
 
-  eleventyConfig.setUseGitIgnore(false);
-
   return {
     passthroughFileCopy: true,
     dir: {
--- a/docs/_data/supporters.js
+++ b/docs/_data/supporters.js
@@ -27,7 +27,7 @@ const blocklist = new Set(require('./blocklist.json'));
  * In addition to the blocklist, any account slug matching this regex will not
  * be displayed on the website.
  */
-const BLOCKED_STRINGS = /(?:[ck]a[sz]ino|seo|slots|gambl(?:e|ing)|crypto)/i;
+const BLOCKED_STRINGS = /(?:[ck]a[sz]ino|seo|slot|gambl(?:e|ing)|crypto|follow|buy|cheap|instagram|hacks|tiktok|likes|youtube|subscriber|boost|deposit|mushroom|bingo|broker|promotion|bathroom|landscaping|lawn care|groundskeeping|remediation|esports|links|coupon|review|refer|promocode|rabattkod|jämför|betting|reddit|hire|fortune|equity|download|marketing|comment|rank|scrapcar|lawyer|celeb|concrete|firestick|playground)/i;
 
 /**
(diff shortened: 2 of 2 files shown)


docs/_data/supporters.js
```diff
@@ -27,7 +27,7 @@ const blocklist = new Set(require('./blocklist.json'));
  * In addition to the blocklist, any account slug matching this regex will not
  * be displayed on the website.
  */
-const BLOCKED_STRINGS = /(?:[ck]a[sz]ino|seo|slots|gambl(?:e|ing)|crypto)/i;
+const BLOCKED_STRINGS = /(?:[ck]a[sz]ino|seo|slot|gambl(?:e|ing)|crypto|follow|buy|cheap|instagram|hacks|tiktok|likes|youtube|subscriber|boost|deposit|mushroom|bingo|broker|promotion|bathroom|landscaping|lawn care|groundskeeping|remediation|esports|links|coupon|review|refer|promocode|rabattkod|jämför|betting|reddit|hire|fortune|equity|download|marketing|comment|rank|scrapcar|lawyer|celeb|concrete|firestick|playground)/i;
```
JoshuaKGoldberg, line 30: Some of these could be legitimate names. Examples:

* `boost`: is a C++ library and is a pretty generalized term
* `review`: very general term
* `playground`: common term for tooling sites, e.g. typescriptlang.org/play

JoshuaKGoldberg (changes requested):
I like the idea, but I think it goes a bit too far?

docs/_data/supporters.js
voxpelli: This list is just for words included in the slug, name and url – I think its unlikely to give a false positive.

* `boost` – used by tools that helps you boost your social media presence
* `review` – used by websites that make money through affiliate links
* `playground` – used by adult-related websites
*
That eg. TypeScript Playground  or similar would sponsor us with that in their name, slug or URL feels unlikely, and if they do they can reach out we can tweak it then.

voxpelli: I have verified that there are no false negatives right now

JoshuaKGoldberg: > right now

... I suppose that's the operative term.

I've you're confident this won't stick around long enough to accidentally stop good actors, I won't try to stand in its way. 🙂

JoshuaKGoldberg (approved):

.eleventy.js
```diff
@@ -46,8 +46,6 @@ module.exports = function (eleventyConfig) {
 
   eleventyConfig.setLibrary('md', markdown);
 
-  eleventyConfig.setUseGitIgnore(false);
```
JoshuaKGoldberg, line 49: unrelated?

voxpelli: `fix: do not reload eleventy infinitely` – else when I tried this out it infinitely retried in the background until Open Collective ratelimited me

docs/_data/supporters.js
voxpelli: I mean, its a best effort to show people on the site, I think people are more concerned about showing up among bad actors than about not showing up at all :)

PR #5198, opened 2024-08-14
fix: handle case of invalid package.json with no explicit config

### LINKED ISSUE #5141: 🐛 Bug: mocha fails silently on invalid `package.json` section
### Bug Report Checklist

### Expected

I made an error in my `package.json` (an extra comma) because JSON is a bad format for configuration files ;-) like this:
```json
"mocha": {
    "spec": "./*.spec.js",
}
```
I expected `mocha` to give me an error, ideally a bit friendlier than the one from `npm`, but something like that:
```
$ npm install
npm ERR! code EJSONPARSE
...
npm ERR! JSON.parse Note: package.json must be actual JSON, not just JavaScript.
```

### Actual

Mocha just failed as if I hadn't configured it at all.  This was very confusing:
```
$ npx mocha
Error: No test files found: "test"
```

See "expected" above but basically:
```
npm install -D mocha
```
then add the offending section to `package.json`, then create `foo.spec.js`, then try to run `mocha`.

### Versions

`
10.4.0
10.4.0
v20.11.0
`

### Additional Info

I could definitely provide a PR if you could direct me to the appropriate place in the code (with which I am not familiar)

dhdaines:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5141

## Overview

Handle the corner case which can happen if:

- You had a perfectly good `package.json` which npm was happy with
- You decided to put `mocha` configuration inside your `package.json`
- You left a stray comma in there, because you are used to more sensible configuration file formats ;-)
- Instead of running `npm test` you ran `mocha` directly (with `npx` or just `mocha.js`)

As detailed here: https://github.com/dhdaines/mocha-5141

### DIFF
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -195,6 +195,11 @@ const loadPkgRc = (args = {}) => {
           `Unable to read/parse ${filepath}: ${err}`,
           filepath
         );
+      } else if (err.toString().includes('SyntaxError')) {
+        throw createUnparsableFileError(
+          `Unable to parse ${filepath}: ${err}`,
+          filepath
+        );
       }
       debug('failed to read default package.json at %s; ignoring', filepath);
     }
--- a/lib/cli/options.js
+++ b/lib/cli/options.js
@@ -181,8 +181,24 @@ const loadPkgRc = (args = {}) => {
   result = {};
   const filepath = args.package || findUp.sync(mocharc.package);
   if (filepath) {
+    let configData;
     try {
-      const pkg = JSON.parse(fs.readFileSync(filepath, 'utf8'));
+      configData = fs.readFileSync(filepath, 'utf8');
+    } catch (err) {
+      // If `args.package` was explicitly specified, throw an error
+      if (filepath == args.package) {
+        throw createUnparsableFileError(
+          `Unable to read ${filepath}: ${err}`,
+          filepath
+        );
+      } else {
+        debug('failed to read default package.json at %s; ignoring',
+              filepath);
+        return result;
+      }
+    }
+    try {
+      const pkg = JSON.parse(configData);
       if (pkg.mocha) {
         debug('`mocha` prop of package.json parsed: %O', pkg.mocha);
         result = pkg.mocha;
@@ -190,18 +206,11 @@ const loadPkgRc = (args = {}) => {
         debug('no config found in %s', filepath);
       }
     } catch (err) {
-      if (args.package) {
-        throw createUnparsableFileError(
-          `Unable to read/parse ${filepath}: ${err}`,
-          filepath
-        );
-      } else if (err.toString().includes('SyntaxError')) {
-        throw createUnparsableFileError(
-          `Unable to parse ${filepath}: ${err}`,
-          filepath
-        );
-      }
-      debug('failed to read default package.json at %s; ignoring', filepath);
+      // If JSON failed to parse, throw an error.
+      throw createUnparsableFileError(
+        `Unable to parse ${filepath}: ${err}`,
+        filepath
+      );
     }
   }
   return result;
--- a/test/node-unit/cli/options.spec.js
+++ b/test/node-unit/cli/options.spec.js
@@ -206,8 +206,7 @@ describe('options', function () {
             // package.json
             readFileSync
               .onFirstCall()
-              // Note the extra comma
-              .returns('{"mocha": {"retries": 3, "_": ["foobar.spec.js"],}}');
+              .returns('{definitely-invalid');
             findConfig = sinon.stub().returns('/some/.mocharc.json');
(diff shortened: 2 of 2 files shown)


dhdaines:
(signed the CLA!)

lib/cli/options.js
```diff
@@ -195,6 +195,11 @@ const loadPkgRc = (args = {}) => {
           `Unable to read/parse ${filepath}: ${err}`,
           filepath
         );
+      } else if (err.toString().includes('SyntaxError')) {
+        throw createUnparsableFileError(
```
voxpelli, line 199: (On my phone) Will throwing here result in a warning? Feels more like an error to me?

lib/cli/options.js
```diff
@@ -195,6 +195,11 @@ const loadPkgRc = (args = {}) => {
           `Unable to read/parse ${filepath}: ${err}`,
           filepath
         );
+      } else if (err.toString().includes('SyntaxError')) {
```
voxpelli, line 198: If we want to separate the parsing from the reading errors, then I would suggest instead splitting this try/catch into an initial one for reading followed by a second one for parsing – that way we will not have to check the error as we know it's `JSON.parse()` that fails

Also gives more granular error for the existing use case – different error message depending on whether it can't read or can't parse

voxpelli:
Thanks for the PR! Gave some feedback

lib/cli/options.js
dhdaines: Seems like a good idea.  Also, would it be useful to give an error for parsing errors more like the one npm gives, i.e. "make sure your package.json is JSON and not Javascript"?

voxpelli: I don’t think it’s needed, just a warning saying “could not check package.json due to error while parsing” is probably the clearest

lib/cli/options.js
dhdaines: It will fail with an error, these exceptions are not caught (we could, possibly, catch them at the top level of the CLI and exit cleanly in order to avoid barfing stack traces on the user...)

lib/cli/options.js
dhdaines: I've simply split it into "Unable to read" and "Unable to parse" for the moment, not sure if there is a better wording.

JoshuaKGoldberg: Looks great to me 🙂 I like it!

test/node-unit/cli/options.spec.js
```diff
@@ -199,6 +199,37 @@ describe('options', function () {
...
 
+        describe('when path to package.json unspecified and package.json exists but is invalid', function () {
+          beforeEach(function () {
+            const filepath = '/some/package.json';
+            readFileSync = sinon.stub();
+            // package.json
+            readFileSync
+              .onFirstCall()
+              // Note the extra comma
+              .returns('{"mocha": {"retries": 3, "_": ["foobar.spec.js"],}}');
```
JoshuaKGoldberg, line 210: [Testing] Nit: no need to have a difficult-to-read JSON that needs a comment, when we can have something more straightforward:

```suggestion
              .returns('{definitely-invalid');
```

JoshuaKGoldberg (approved):
LGTM, nicely done! Agreed with the strategy inside 👍 thanks!

Since @voxpelli was looking at this before, leaving open for re-review.

JoshuaKGoldberg:
Btw @dhdaines, just a heads up, the `- [ x ]`s in the description can be `- [x]`s (no spacing) to work with [GitHub's task lists](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists#creating-task-lists). I edited them just now. 🙂

dhdaines:
> Btw @dhdaines, just a heads up, the `- [ x ]`s in the description can be `- [x]`s (no spacing) to work with [GitHub's task lists](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists#creating-task-lists). I edited them just now. 🙂

Ohhh!  Thank you, I have been wondering about that for ages now! (suppose I should have RTF the Markdown manual...)

test/node-unit/cli/options.spec.js
dhdaines: Ah, yes, that is a lot clearer (the most common error people will make in JSON is obviously the trailing comma, but that isn't very relevant for this test case)

dhdaines:
Ah, I suspect that the more obviously invalid JSON will break the too-specific assertion in the test...  will fix it in a second.

dhdaines:
> Ah, I suspect that the more obviously invalid JSON will break the too-specific assertion in the test... will fix it in a second.

Done!

JoshuaKGoldberg (approved):
🙌 Looks great to me! Just waiting on @voxpelli. Thanks!

PR #5204, opened 2024-08-29
fix: ignore all node_modules and .git dirs by default

### LINKED ISSUE #5203: 🚀 Feature: Change default value for `--watch-ignore` to `**/node_modules`, `**/.git`
### Feature Request Checklist

- [X] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)

### Overview

In monorepos and other projects that have `node_modules` and/or `.git` directories deeper than the root, `mocha --watch` watches the contents of these directories by default.

(You may say, but in a monorepo you're just running tests in the directory for each individual package, where its `node_modules` lives, right?  But sometimes you want to run integration tests that combine multiple packages in the monorepo.)

This generally causes `mocha --watch` to take a long time to start up, with no clear indication why.

### Suggested Solution

If we changed the default `watch-ignore` to `["**/node_modules", "**/.git"]` it would automatically ignore all such directories

### Alternatives

Of course we can override this in userland, but I think it would benefit everyone to just ignore all `node_modules` and `.git` directories by default.

### Additional Info

_No response_

jedwards1211:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5203 

## Overview

Just adds `**/` before the default ignore patterns so that all `node_modules` and `.git` dirs get ignored, not just ones in the root folder.

### DIFF
--- a/lib/mocharc.json
+++ b/lib/mocharc.json
@@ -6,5 +6,5 @@
   "slow": 75,
   "timeout": 2000,
   "ui": "bdd",
-  "watch-ignore": ["node_modules", ".git"]
+  "watch-ignore": ["**/node_modules", "**/.git"]
 }


JoshuaKGoldberg:
Putting in draft per https://github.com/mochajs/mocha/issues/5203#issuecomment-2316648237. Sorry for being misleading!

lib/mocharc.json
```diff
@@ -6,5 +6,5 @@
   "slow": 75,
   "timeout": 2000,
   "ui": "bdd",
-  "watch-ignore": ["node_modules", ".git"]
+  "watch-ignore": ["**/node_modules", "**/.git"]
```
JoshuaKGoldberg: [blocking] (high confidence) Per CONTRIBUTING.md, tests are required for any code change — and right now the watch suite passes identically with and without this line. The existing default-ignore test (`test/integration/options/watch.spec.cjs`, `ignores files in "node_modules" and ".git" by default`) only creates root-level `node_modules`/`.git`, which both the old and new defaults ignore. Could you extend it (or add a sibling case) with nested dirs, e.g. `tempDir/packages/a/node_modules/file.xyz` and `tempDir/packages/a/.git/file.xyz`? That's the behavior this PR actually changes, and it would keep a future refactor of the watch matcher from silently reverting it.

PR #5210, opened 2024-09-10
docs: add `SECURITY.md` pointing to Tidelift

voxpelli:
To fulfill the lifting at Tidelift we need to add a security policy (which we should have anyways)


JoshuaKGoldberg (approved):

Uzlopak (approved):
LGTM

Issue #5148, opened 2024-05-28
🚀 Feature: Update glob from 8.1.0 which is no longer supported


einsteinsfool:
### Feature Request Checklist

- [X] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/master/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/master/.github/CONTRIBUTING.md)
- [X] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [ ] I want to provide a PR to resolve this

### Overview

mocha uses glob@8.1.0 which is not supported and also depends on inflight which is also not supported (any version).
```
$ mkdir mochatest
$ cd mochatest
$ npm install mocha
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported
...
```

I do realise it's been reported in #4988 but I'm not reporting it as security vulnerability but feature request.
It would be nice not to have these warnings. Especially that a lot of packages depend on mocha (like appium from which I came).

### Suggested Solution

Updating glob dependency to v9.0.0 or maybe even v10.4.1.

### Alternatives

Leave glob v8.1.0 which is no longer supported and ignore warnings.

### Additional Info

https://www.npmjs.com/package/inflight
https://www.npmjs.com/package/glob/v/8.1.0

JoshuaKGoldberg:
Agreed, this would be great. There's no reason to keep `glob` -or any other dependency- on an outdated major version - especially one that's no deprecated!

#5114 tracks allowing `^` (minor) version matching on packages. That's a `semver-minor` for us. 

[...]

PR #5216, opened 2024-09-16
fix!: adapt new engine range for Mocha 11

### LINKED ISSUE #5206: 🧹 Versioning: Bump minimum Node.js version from 14.0.0 to 18.18.0
## Overview

We're gearing up for a new major version of Mocha 11. That means we can bump the minimum Node engine version! Hooray! This is great for a couple of reasons:

* It unblocks bumping old package versions, i.e. #5148
* It means we can use newer JavaScript + platform features

We'd like to bump to, at the most strict, [Node LTS](https://endoflife.date/nodejs). An added pressure is that Mocha is still used by large projects such as ESLint and TypeScript, which have older version support requirements.

Our current thought is to go with:

* Node LTS at time of v11 release: should all be supported
* #5148: whatever is needed for a reasonably recent `glob` version
* If a major consumer such as ESLint requires a _slight_ extension of support range for its devDependencies, we could allow that too

voxpelli:
As we prepare for the next step of Mocha its time to get rid of some of the baggage that's stopping us from updating dependencies and using modern language features.

As a dev dependency that other open source projects rely on we can't be too aggressive with this bump – eg. dropping Node 20 is sadly too soon.

To make as little hassle as possible for users it makes sense to align our version range with other common dev dependencies, and one of the newest such releases is ESLint 9, which has a version range of [`^18.18.0 || ^20.9.0 || >=21.1.0`](https://github.com/eslint/eslint/blob/6d4484d9c19e4132f3dee948174a543dbbb5d30f/package.json#L199)

This PR also sets up `release-please` for pre-releases and adds `installed-check` (which I maintain) to validate the engine range against that of our dependencies (running `npx installed-check --engine-check --verbose` it shows which modules that are missing published engine ranges and which it can't check)

Fixes #5206


package.json
```diff
@@ -53,8 +53,9 @@
     "format:eslint": "eslint --fix . 'bin/*'",
     "format:prettier": "prettier --write '!(package*).json' '.*.json' 'lib/**/*.json' '*.yml'",
     "format": "run-s format:*",
-    "lint:knip": "knip --cache",
     "lint:code": "eslint . 'bin/*' --max-warnings 0",
+    "lint:installed-check": "installed-check --engine-check",
+    "lint:knip": "knip --cache",
```
voxpelli, line 58: Alphabetic sorting is 🙌

voxpelli:
Lets due some more due diligence before dropping older Node.js versions, as eg. TS uses Mocha and has a wider engine range than suggested here, see comment here: https://github.com/mochajs/mocha/issues/5148#issuecomment-2353604097

Uzlopak (approved):
LGTM

package.json
```diff
@@ -40,7 +40,7 @@
     "test": "./test"
   },
   "engines": {
-    "node": ">= 14.0.0"
+    "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
```
JoshuaKGoldberg, line 43: TypeScript still supports Node 14.17. I asked about their plans in https://discord.com/channels/508357248330760243/640177505915109377/1293239882323591200.

JoshuaKGoldberg: Confirmed we're 👍 to break TypeScript.

Related: @voxpelli noted in a private chat that bumping from 14 to 20 is _two_ even major version bumps. And (still quoting here) it could possibly also be that since Node 18 introduced nodes test runner, that a maintained test runner for older versions are in extra high demand.

👍 agreement from me - it'd be nice to ease into updating version requirements.

```suggestion
    "node": "^16.16.0 || ^18.18.0 || ^20.9.0 || >=21.1.0"
```

JoshuaKGoldberg (changes requested):
🚀

voxpelli:
We need to decide on if we feel we have enough data to support a decision to bump to 16 or 18

For 16 we should double check what that means in terms of bumping dependencies, if we still remain on outdated unsupported dependencies then we have kind of bumped in vain and are not much better of with 16 than 14?

If 18 means dropping far too many projects, then that will be complex as well

I tried pulling down all the engine ranges for dependents from ecosyste.ms using my list-dependents-cli, but I think I failed

And that data has been proven
More unreliable than I would have hoped as it turns out that some heavyweight projects refuse to use engine declarations, making automated calculations less representative https://github.com/fastify/fastify/pull/5667

voxpelli:
If we don’t have time or want to do more investigation to pick between 16 and 18, then I say we stick with the original proposal and update to 18. Maybe it’s too cutting edge for some, but it’s simpler for us and we have no data to show that it would cause a lot of problems for the community.

JoshuaKGoldberg:
👍 I don't have the energy to push back on this and don't feel strongly beyond "we should do this next week". So if nobody has the energy to investigate, and you're ok with 18, then sure - 18 it is.

JoshuaKGoldberg (approved):
🚀

package.json
JoshuaKGoldberg, line 1: [Docs] Just noting, we'll also want to do a find-all for any reference to Node.js 14. For example, `.github/DEVELOPMENT.md`.

jedwards1211:
It would help if the changelog message were more straightforward, like:

BREAKING CHANGE: drop support for versions of Node below ^18.18.0 || ^20.9.0 || >=21.1.0

Is this actually what the breaking change was?  The wording in this PR is vague

PR #5218, opened 2024-09-29
Highlight browser failures

mark-wiemer:
Picking up where #4148 left off as I don't have push permissions to that fork


mocha.sass
```diff
@@ -342,6 +342,16 @@ body {
   margin: 0;
   color: var(--mocha-stats-color);
   z-index: 1;
+  background-color: #00e600;
+  border: solid  #00e600;
+  border-radius: 10px;
+}
+
+#mocha-stats.fail {
+ @extend #mocha-stats;
```
mark-wiemer, line 351: `@extend` won't work now that we've removed sass. I personally don't have experience with sass but I'm good with CSS, I'll dig in and see if I can replace this

[From JoshuaKGoldberg on the original PR](https://github.com/mochajs/mocha/pull/4148/files/c9b2939c98789cf23bccfd17e23682601c7a8ca2#r1509367853):

[Style] I don't think this @extend is really needed. We can achieve the same result in raw CSS, such as with CSS variables or different class names.

Thinking in terms of semantic class names, I think it'd make sense to have a class like .fail, .pass, etc. that each define their own colors.

mocha.sass
```diff
@@ -342,6 +342,16 @@ body {
   margin: 0;
   color: var(--mocha-stats-color);
   z-index: 1;
+  background-color: #00e600;
```
mark-wiemer, line 345: From [JoshuaKGoldberg on the original PR](https://github.com/mochajs/mocha/pull/4148/files/c9b2939c98789cf23bccfd17e23682601c7a8ca2#r1509399168):

A few things that make me not a fan yet:

The new colors are hard to read on top of, to the point of not having [accessible color contrast](https://webaim.org/resources/contrastchecker/?fcolor=000000&bcolor=CC0000).
The colors also don't match the existing aesthetic of the page. The page right now has very minimalist colors and a subtle design. This new green and red is pretty glaring. We should go with something that doesn't clash so much.
It's a little inconsistent in how its laid out: more padding on the top vs. the bottom, and less horizontal padding than other UI elements.
I'm not sure what would be the "right" approach. A few things come to mind, but I haven't thought very deeply:

Using much thinner border lines and/or much less saturated colors
Even switching to an [nprogress](https://github.com/rstacruz/nprogress)-style bar on the stats header or the page
Instead of a background or border, using bold & colored text
(I'm not convinced any of that is better, just ideating)

mark-wiemer (commented):
Pending changes from original PR

mark-wiemer:
FYI, I'm very new to contributing to mocha, so right now I'm still trying to just reproduce the issue. I've installed Chrome and `npm run test-browser-run` is passing, which is great, but I don't know yet how to keep the browser open to inspect test results. I always use the CLI :D

mark-wiemer:
image

Still struggling to get a passing case working with a local build of mocha.js, I'm only used to the default CLI reporter! But I think outline is a better way to go than background color, and some clear text/shape indicating pass or fail will work for colorblind folks as well. Not sure if we want to add an announcement to the HTML reporter if tests fail, but that sounds beyond the scope of this PR anyway.

I'll continue working to get a passing example, fixup padding, and make any requested changes as well :)

mocha.sass
mark-wiemer: - [ ] New colors: Agreed we should use existing colors, the new bright colors aren't necessary
- [ ] Padding: alignment is in progress :)
- [ ] Thinner border lines: will investigate
- [ ] nprogress-style bar: will investigate
- [ ] Colored text vs background vs border: will investigate

mark-wiemer:
The styles are kinda tough with `display: block`, I'm going to see if I can change to `display: flex`. Would that be OK or is it strongly not recommended?

Here is the current appearance, btw:

image

image

image

image

mark-wiemer:
Was able to make things work with `display: block;`, if a slightly taller header is OK. Final styles until review:

image

image

---

image

image

mocha.sass
mark-wiemer: - New colors used
- Padding fixed up
- Border lines look OK to me
- Loading indicator feels out-of-scope for this PR
- Background color + icon feels the most accessible and subtle to me

Resolving this one, happy to work on new comments :)

mocha.css
```diff
@@ -352,7 +377,7 @@ body {
 #mocha-stats :is(.progress-element, .progress-text) {
   width: var(--ring-container-size);
   display: block;
-  top: 12px;
+  top: 17px;
```
JoshuaKGoldberg, line 380: 😬 I was really hoping we wouldn't move UI elements around for this change. The taller this section, the less fits on the page in smaller viewports. A lot of people are accustomed to the placement as-is. Which is why I was hoping to discuss a bit in the backing issue (https://github.com/mochajs/mocha/pull/4148#pullrequestreview-1911745911) before having a PR author such as yourself go through the effort of implementing a more complex design. I'll post there.

JoshuaKGoldberg (changes requested):
Seems like a reasonable implementation for the proposed design at first. But I think we'll want to discuss a bit before settling on the design. Thanks for sending! ❤️

mocha.css
mark-wiemer: No worries, just wanted to get something looking good, let's discuss in the backing issue :)

mark-wiemer:
Sorry, I had forked a fork instead of the main repo, it was becoming difficult to manage. Will open this as a separate PR shortly, don't worry :)

PR #5221, opened 2024-10-08
feat: add ESM export

### LINKED ISSUE #5211: 🚀 Feature: Ability to import mocha from ESM in the browser
### Feature Request Checklist

### Overview

Inability to import as ESM module from the browser

### Suggested Solution

Include an ESM build

### Alternatives

Use Rollup on the file but burdensome for each project. Continue adding separate script tags for Mocha and other projects.

### Additional Info

_No response_

brettz9:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5211 

## Overview

Adds an ESM export through a new `mocha.mjs` file (making the regular file to be ESM would be a breaking change). It is not entirely modular, as the base script still sets a global `Mocha` and `mocha`, but it does export `mocha`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,9 @@ docs/images/supporters
 docs/api
 mocha.js
 mocha.js.map
-mocha.mjs
-mocha.mjs.map
+esm/mocha.js
+esm/mocha.js.map
+esm/package.json
 .karma/
 !bin/mocha.js
 !lib/mocha.js
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ docs/images/supporters
 docs/api
 mocha.js
 mocha.js.map
+mocha.mjs
+mocha.mjs.map
 .karma/
 !bin/mocha.js
 !lib/mocha.js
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -4,6 +4,9 @@ docs/images/supporters
 docs/api
 mocha.js
 mocha.js.map
+esm/mocha.js
+esm/mocha.js.map
+esm/package.json
```
JoshuaKGoldberg, line 9: Oh 😬 I didn't think adding support for ESM would require an entirely new output file. This file is pretty huge: 600k+ characters, 620K on disk. We can't add such a large addition to package size for all users.

I had assumed the entry point would be a minimal shim, i.e.:

```js
import "./mocha.js"

const { mocha } = globalThis;

const describe = mocha.describe.bind(mocha);
const it = mocha.it.bind(mocha);
const setup = mocha.setup.bind(mocha);

export { describe, it, setup };
```

Is that not doable for some reason?

JoshuaKGoldberg (changes requested):
Thanks for sending this! Let's talk about the direction?

.gitignore
brettz9: That makes sense. I'm unfortunately dealing with some health problems, so I think I will just nix this PR.

PR #5227, opened 2024-10-12
fix: remove `:is()` from `mocha.css` to support older browsers (#5225)

### LINKED ISSUE #5225: `mocha.css` contains `:is()` and broke in older browsers
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:is#browser_compatibility), `:is()` is supported since Chromium 88 and Firefox 78.

Not sure what browser versions do mocha intend to support. Our project currently targets Chromium >= 73 and Firefox >= 68, and this CSS issue prevents us from upgrading `mocha` library (from `v10.3.0`).

danny0838:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5225

## Overview

Remove `:is` to prevent CSS breaking in older browsers.

This is also more consistent with older existent CSS which did not use `:is()` in the same file, such as [L318-L319](https://github.com/mochajs/mocha/blob/v10.7.3/mocha.css#L318-L319).

### DIFF
--- a/mocha.css
+++ b/mocha.css
@@ -349,7 +349,8 @@ body {
   padding: 0;
 }
 
-#mocha-stats :is(.progress-element, .progress-text) {
+#mocha-stats .progress-element,
+#mocha-stats .progress-text {
   width: var(--ring-container-size);
   display: block;
   top: 12px;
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
👍 Eventually it'd be nice to use more modern tooling that transpiles down to a supported browser range. But this makes sense in the interim. Thanks!

PR #5212, opened 2024-09-12
fix: support errors with circular dependencies in object values with --parallel

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5188; fixes #5209
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Only calling `breakCircularDeps` on `result.error` means any circular references in other properties of `result` still go through. `SerializableEvent._serialize` can generally visit them fine, but they still end up being infinitely added to `pairs`. This PR adds a `seenPairs` `Set` to make sure they only get serialized once each. This fixes #5209.

Also only writes to properties with `.writable` object descriptors. This fixes #5188.

Issue #5249, opened 2024-11-11
📌 Mocha 11 Release Plan


JoshuaKGoldberg:
**Edit (December 2nd, 2024): [`mocha@11.0.1`](https://www.npmjs.com/package/mocha/v/11.0.1) is released as `latest` on npm.**

---

We're working on a new major version of Mocha! 📦 

[Mocha 11](https://github.com/mochajs/mocha/milestone/65) will be the first new major version of Mocha in 3 years. It will contain **no new bug fixes or features** compared to v10. The only user-facing changes in Mocha 11 will be:

* #5148
* #5206

Our plan is to:
- [x] Monday, November 11th: release a "pre-release" version of Mocha 11 today to npm (_not_ as stable)
    * [`mocha@11.0.0`](https://www.npmjs.com/package/mocha/v/11.0.0) is published as `mocha@next` on npm 👍 
- [x] Wait a few business days to merge the two PRs, then >=1 business day after
- [ ] Release Mocha 11 as stable on npm

As mentioned in #5027, we're a new volunteer maintenance team and want to tread carefully with breaking changes to Mocha. We're holding off an adding any additional breaking changes until the next major version, Mocha 12, which we will not release until some time in 2025 at the earliest. We'll first learn see how the Mocha 11 release goes and learn from it. As careful as we are planning on being with Mocha 11, there is always some risk with publishing a new major version of a project. Mocha is depended upon by _thousands upon thousands_ of consuming projects, many of which are not open source or built in ways we're familiar with as a team. 

Alternating major versions between ones that purely raise engine targets and ones that add breaking feature changes is a strategy used by some other ecosystem projects, such as [typescript-eslint](https://typescript-eslint.io/blog/announcing-typescript-eslint-v7#breaking-changes).

> 👉 In the meantime, we're also working on a new docs site at **https://mocha-docs-next.netlify.app**.

mark-wiemer:
>  It will contain no new bug fixes or features compared to the 11.

Sorry, what exactly does this mean?

JoshuaKGoldberg:
Whoops, that should say "to 10". Fixed. Thanks :)

jftanner:
> Wait a few business days to merge the two PRs

How're you feeling about merging #5148? It'd be nice to be able to kick the tires on `mocha@next` for a few days before a lot of US folk are away for the Thanksgiving week.

kraenhansen:
[...]

PR #5254, opened 2024-11-14
fix: catch exceptions setting Error.stackTraceLimit

kevinoid:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5212

## Overview

When node is run with [--frozen-intrinsics], a `TypeError` is thrown when any intrinsic objects or their properties are modified.  This occurs when attempting to set `Error.stackTraceLimit`.  To avoid exiting due to an uncaught exception, catch the exception, debug log it, and continue.

[--frozen-intrinsics]: https://nodejs.org/api/cli.html#--frozen-intrinsics

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -42,7 +42,11 @@ exports.main = (argv = process.argv.slice(2), mochaArgs) => {
     module.paths.push(cwd(), path.resolve('node_modules'));
   }
 
-  Error.stackTraceLimit = Infinity; // configurable via --stack-trace-limit?
+  try {
+    Error.stackTraceLimit = Infinity; // configurable via --stack-trace-limit?
+  } catch (err) {
+    debug('unable to set Error.stackTraceLimit = Infinity', err);
+  }
 
   var args = mochaArgs || loadOptions(argv);
 


kevinoid:
Note: Tests which load `unexpected` can't currently run with `--frozen-intrinsics` due to https://github.com/petkaantonov/bluebird/issues/1717 (present in `unexpected-bluebird`) causing:

    ✖ ERROR: TypeError <Object <Object <[Object: null prototype] {}>>>: Cannot define property __BluebirdErrorTypes__, object is not extensible
        at Object.defineProperty (<anonymous>)
        at notEnumerableProp (/path/to/mocha/node_modules/unexpected-bluebird/js/main/util.js:106:9)
        at Object.<anonymous> (/path/to/mocha/node_modules/unexpected-bluebird/js/main/errors.js:99:5)
        at Module._compile (node:internal/modules/cjs/loader:1469:14)
        at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
        at Module.load (node:internal/modules/cjs/loader:1288:32)
        at Module._load (node:internal/modules/cjs/loader:1104:12)
        at Module.require (node:internal/modules/cjs/loader:1311:19)
        at require (node:internal/modules/helpers:179:18)
        at module.exports (/path/to/mocha/node_modules/unexpected-bluebird/js/main/promise.js:30:14)

Since `.mocharc.yml` requires `test/setup.js` which requires `unexpected`, nearly all invocations of `mocha` will fail.

We could smoke test this feature by calling `node --frozen-intrinsics bin/mocha.js --version` or do more thorough testing with a different `--config` which doesn't require `unexpected`, but I'm not sure the value is justified by the effort.  It may be preferable to wait until the test dependencies support running under `--frozen-intrinsics`.  What do you think?

JoshuaKGoldberg (approved):
LGTM, thanks!

JoshuaKGoldberg:
This is really more of a bugfix than a feature - so it only necessitates a patch-level version bump.

We're also about to release major version 11 and don't have a separate branch set up to differentiate v10 and v11 releases or support backporting. So this should be good to go only after v11 is released this month.

JoshuaKGoldberg:
Quick status update: Mocha 11 is stable and I'm just waiting this week in case something breaks. I'd be surprised if something does, but you never know! #5249 

Assuming nothing goes wrong, I plan on merging this early next week. 🙂

JoshuaKGoldberg:
Lloyd Bridges as Izzy Mandelbaum in Seinfeld aggressively telling Jerry: "IT'S GO TIME"

JoshuaKGoldberg:
Oh, and:

> ... It may be preferable to wait until the test dependencies support running under --frozen-intrinsics. What do you think?

I think there's no harm in Mocha fixing its support now. Eventually the test dependencies will too. No need for us to be the ones still blocking.

Issue #5207, opened 2024-09-03
📝 Docs: Overhaul website generation


JoshuaKGoldberg:
### Documentation Request Checklist

- [X] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [X] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [X] I want to provide a PR to resolve this

### Overview

Back in #5027, we stated:

> We don't intend to overhaul any currently-useful resources such as https://mochajs.org/ or https://github.com/mochajs/mocha/wiki. Though, we will happily accept issue reports if those resources need improvement.

We still don't want to overhaul the _contents_ of the Mocha website. However, the _tooling_ behind the website is ... not great to work with. It's built with the very old https://github.com/mochajs/mocha-docdash which comes with a ton of old dependencies that are hard to update (#5071).

We'd like to redo the website's _tooling_ to be something more modern. I'm personally partial to Astro Starlight. I've had great experiences with Docusaurus too.

### Additional Info

This is roughly blocking #5055. If we enable Renovate, we'll get a bunch of update PRs for the same packages touched by #5071. Irksome.

JoshuaKGoldberg:
FYI, I'm going to be live streaming making a new Mocha website tomorrow (Thursday) at 10am EST. 👻 
https://www.twitch.tv/joshuakgoldberg/schedule?segmentID=4090b3ff-45c0-46ad-b1a3-4f200204251b

PR #5263, opened 2024-11-25
Fix 5028 // Numerical arguments to cli throw uncaught error

### LINKED ISSUE #5028: `TypeError: arg.split is not a function` when passed a numerical argument
### Prerequisites

### Description

With similar symptoms to https://github.com/mochajs/mocha/issues/3861, Mocha fails when given a numerical argument:

```
>npx mocha 1234                                                                 
/Users/nf/.npm/_npx/508606763866ae01/node_modules/mocha/lib/cli/options.js:113
      const pair = arg.split('=');
                       ^

TypeError: arg.split is not a function
    at /Users/nf/.npm/_npx/508606763866ae01/node_modules/mocha/lib/cli/options.js:113:24
    at Array.reduce (<anonymous>)
    at parse (/Users/nf/.npm/_npx/508606763866ae01/node_modules/mocha/lib/cli/options.js:111:67)
    at loadOptions (/Users/nf/.npm/_npx/508606763866ae01/node_modules/mocha/lib/cli/options.js:246:10)
    at Object.<anonymous> (/Users/nf/.npm/_npx/508606763866ae01/node_modules/mocha/bin/mocha.js:27:14)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)

Node.js v18.12.1
[1]    79408 exit 1     npx mocha 1234
```

**Expected behavior:**
Mocha to proceed as expected, or to throw a more descriptive error message. 

**Actual behavior:** 
A crash and a stacktrace.

**Reproduces how often:** 
100%

### Versions

Mocha 10.2.0
Node v18.12.1
macOS Sonoma 14.0
zsh 5.9 (x86_64-apple-darwin23.0)

### Additional Information

It can be expected that `argv` is a string array, but it appears that `yargsParser.Arguments._` which is fed back into `parse` at https://github.com/mochajs/mocha/blob/48002bc60144c25060bf91359620c4a0284e8284/lib/cli/options.js#L246 will contain numbers.

Dinika:
## PR Checklist

- [X] Addresses an existing open issue: fixes #5028 

## Overview

This PR adds error handling for "TypeError: arg.split is not a function" which is thrown when an unexpected numerical argument is passed to mocha cli. Hopefully the thrown error message is clearer to understand (I'm open to suggestions to further improve it).

This custom error is thrown in following cases:

- A numerical argument was passed to cli (such as `mocha 123`)
- A numerical argument was passed as a value to a mocha option that does not accept numerical values (such as `mocha --delay 123`)

The error is not thrown in if a numerical value is passed to mocha flag which can handle numerical values (such as `mocha --retries 2`) or if numerical value is passed to node options.

*PS* - Apologies for posting this PR before receiving a reply to [this comment](https://github.com/mochajs/mocha/issues/5028#issuecomment-2494261947) I posted in the issue.


lib/cli/options.js
```diff
@@ -110,15 +110,19 @@ const parse = (args = [], defaultValues = {}, ...configObjects) => {
...
   const nodeArgs = (Array.isArray(args) ? args : args.split(' ')).reduce(
     (acc, arg) => {
-      const pair = arg.split('=');
-      let flag = pair[0];
-      if (isNodeFlag(flag, false)) {
-        flag = flag.replace(/^--?/, '');
-        return arg.includes('=')
-          ? acc.concat([[flag, pair[1]]])
-          : acc.concat([[flag, true]]);
+      try {
```
JoshuaKGoldberg, line 113: [Refactor] `try...catch` is a very blunt tool that addresses a symptom, not a root cause of something being wrong. I think a more direct fix would be to check that the type of the arg is what we expect: a string.

I wrote up an explainer here with examples: https://www.joshuakgoldberg.com/blog/try-catch-as-little-as-possible

lib/cli/options.js
```diff
@@ -192,8 +196,7 @@ const loadPkgRc = (args = {}) => {
           filepath
         );
       } else {
-        debug('failed to read default package.json at %s; ignoring',
-              filepath);
+        debug('failed to read default package.json at %s; ignoring', filepath);
```
JoshuaKGoldberg, line 199: [Style] Please keep these and other unrelated style changes out of the PR. It makes it harder to know what's actually different.

test/node-unit/cli/options.spec.js
```diff
@@ -676,5 +674,48 @@ describe('options', function () {
...
+      it('does not throw error if numerical value is passed to a node options', function () {
+        expect(
+          () =>
+            loadOptions(
+              `--secure-heap-min=${numericalArg} --conditions=${numericalArg}`
+            ),
+          'not to throw'
+        );
+      });
+    });
```
JoshuaKGoldberg, line 719: [Praise] These are super thorough, nicely done! 🔥

JoshuaKGoldberg (changes requested):
Thanks for getting this started! 🙌 The unit tests in particular are lovely.

Requesting changes on a different strategy for detecting invalid args.

Dinika:
@JoshuaKGoldberg Thanks so much for the review. I've revised the PR as per the suggestions.

lib/cli/options.js
```diff
@@ -110,6 +110,9 @@ const parse = (args = [], defaultValues = {}, ...configObjects) => {
   // 4. we can then reapply the values after yargs-parser is done.
   const nodeArgs = (Array.isArray(args) ? args : args.split(' ')).reduce(
     (acc, arg) => {
+      if (typeof arg !== 'string') {
+        throw new Error(`Invalid option ${arg} passed to mocha cli`);
```
JoshuaKGoldberg, line 114: [Feature] If there are a lot of flags, it'd be nice to know _which_ is wrong. Also, in the case of a very long arg, it'd be easier to put the flag first:

```suggestion
        throw new Error(`Mocha flag '--${flag}' given invalid option: '${arg}'`);
```

(or something like that: I haven't run the code to check)

JoshuaKGoldberg (changes requested):
This looks great, thank you! The approach seems very reasonable and straightforward. Just requesting changes on including a bit more information in the thrown error.

The [Style] nit is definitely not blocking if you want to assertively say that it's _numerical_, not _numeric_ 😄 

Thanks!

JoshuaKGoldberg:
Btw, whenever you want a re-review, the [re-request review button](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review#requesting-reviews-from-collaborators-and-organization-members) works.

Dinika:
I do not have permissions to check why the "netlify/mocha/deploy-preview" job is failing. Lemme know if there's something I need/can to do to fix it. Thanks!

lib/cli/run-option-metadata.js
JoshuaKGoldberg, line 1: The Netlify deploy previews are failing with:

```plaintext
5:35:48 AM: [11ty] Problem writing Eleventy templates:
5:35:48 AM: [11ty] 1. Having trouble rendering liquid template ./docs/api/cli_run-option-metadata.js.html (via TemplateContentRenderError)
5:35:48 AM: [11ty] 2. unexpected token at ":string[]", file:./docs/api/cli_run-option-metadata.js.html, line:55, col:10 (via ParseError)
5:35:48 AM: [11ty] 3. unexpected token at ":string[]" (via AssertionError)
5:35:48 AM: [11ty] Wrote 0 files in 12.18 seconds (v1.0.2)
5:35:48 AM: ERROR: "docs:site" exited with 1.
```

I get it trying to run `npm run docs` locally:

```plaintext
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble rendering liquid template ./docs/api/cli_run-option-metadata.js.html (via TemplateContentRenderError)
[11ty] 2. unexpected token at ":string[]", file:./docs/api/cli_run-option-metadata.js.html, line:55, col:10 (via ParseError)
[11ty] 3. unexpected token at ":string[]" (via AssertionError)
[11ty] 
[11ty] Original error stack trace: AssertionError: unexpected token at ":string[]"
[11ty]     at new AssertionError (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:1201:28)
[11ty]     at assert (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:1236:15)
[11ty]     at Tokenizer.readFilter (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:2245:9)
[11ty]     at Tokenizer.readFilters (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:2234:31)
[11ty]     at new Value (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:2788:34)
[11ty]     at new Output (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:4472:23)
[11ty]     at Parser.parseToken (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:4541:24)
[11ty]     at Parser.parseTokens (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:4531:33)
[11ty]     at Parser.parse (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:4525:21)
[11ty]     at Liquid.parse (/Users/josh/repos/mocha/node_modules/liquidjs/dist/liquid.node.cjs.js:4663:28)
[11ty] Benchmark    973ms  77%     1× (Data) `./docs/_data/supporters.js`
[11ty] Benchmark    110ms   9%     1× (Data) `./docs/_data/usage.js`
[11ty] Wrote 0 files in 1.21 seconds (v1.0.2)
ERROR: "docs:site" exited with 1.
```

There's some other log spam earlier about `TypeError: received html and expected a png; outage likely`. You can ignore that.

lib/cli/run-option-metadata.js
```diff
@@ -114,3 +105,23 @@ const ALL_MOCHA_FLAGS = Object.keys(TYPES).reduce((acc, key) => {
 exports.isMochaFlag = flag => {
   return ALL_MOCHA_FLAGS.has(flag.replace(/^--?/, ''));
 };
+
+/**
+ * Returns expected yarg option type for a given mocha flag.
+ * @param {string} flag - Flag to check (can be with out without leading "--"")
+ * @returns {string | undefined} - If flag is a valid mocha flag, the expected type of argument for this flag is returned, otherwise undefined is returned.
+ */
```
JoshuaKGoldberg, line 113: This shouldn't be mentioned in the docs site, so:

```suggestion
 * @returns {string | undefined} - If flag is a valid mocha flag, the expected type of argument for this flag is returned, otherwise undefined is returned.
 * @private
 */
```

And 🤷 cool hey this fixes the docs generator for me locally! Must be that Mocha's docdash version is so old it can't really parse TS syntax normally. Yet another motivation for #5207.

lib/cli/run-option-metadata.js
JoshuaKGoldberg: Aha! Accidentally stumbled onto it: https://github.com/mochajs/mocha/pull/5263/files#r1867700143

test/node-unit/utils.spec.js
```diff
@@ -45,5 +45,16 @@ describe('utils', function () {
         );
       });
     });
+    describe('isNumeric()', function () {
```
JoshuaKGoldberg, line 48: [Tests] I tried removing the `isFloat()` from `isNumeric`, making it just `return !isNan(input)`. These three unit tests still passed but `npm run test-node` had a failure:

```plaintext
Error: Option  is unsupported by the mocha cli
    at createUnsupportedError (/Users/josh/repos/mocha/lib/errors.js:124:101)
    at /Users/josh/repos/mocha/lib/cli/options.js:52:523
    at Array.reduce (<anonymous>)
    at parse (/Users/josh/repos/mocha/lib/cli/options.js:52:186)
    at loadOptions (/Users/josh/repos/mocha/lib/cli/options.js:85:354)
    at Object.<anonymous> (/Users/josh/repos/mocha/bin/mocha.js:9:674)
    at Module._compile (node:internal/modules/cjs/loader:1467:14)
    at Module.replacementCompile (/Users/josh/repos/mocha/node_modules/append-transform/index.js:60:13)
    at Module._extensions..js (node:internal/modules/cjs/loader:1551:10)
    at Object.<anonymous> (/Users/josh/repos/mocha/node_modules/append-transform/index.js:64:4) {
  code: 'ERR_MOCHA_UNSUPPORTED'
}
```

So, there must be some case missing from tests. Requesting that be tested too please.

test/node-unit/cli/options.spec.js
```diff
@@ -294,7 +295,6 @@ describe('options', function () {
 
             result = loadOptions('--no-diff --no-config');
           });
-
           it('should return parsed args, default config and package.json', function () {
```
JoshuaKGoldberg, line 298: [Style] Nit: unrelated change?

```suggestion

          it('should return parsed args, default config and package.json', function () {
```

test/node-unit/cli/options.spec.js
```diff
@@ -526,7 +526,7 @@ describe('options', function () {
           loadOptions('--timeout 500'),
           'to have property',
           'timeout',
-          '500'
+          500
```
JoshuaKGoldberg, line 529: [Non-Actionable] I suppose this is an unrelated change, but there's no reason why 500 should have been a string before. 👍 from me 😄

lib/cli/run-option-metadata.js
```diff
@@ -114,3 +105,23 @@ const ALL_MOCHA_FLAGS = Object.keys(TYPES).reduce((acc, key) => {
...
+exports.expectedTypeForFlag = flag => {
+  const normalizedName = flag?.replace(/^--?/, '');
```
JoshuaKGoldberg, line 116: > `flag?.`

[Refactor] `flag` is marked as `@param {string}`, not `@param {string | undefined}`. So in theory this `?.` can be replaced with a `.`:

```suggestion
  const normalizedName = flag.replace(/^--?/, '');
```

I tried it out locally and no new tests failed. So I think this is a good replacement to make? But if it isn't, then the JSDoc type should be switched to `string | undefined`.

test/node-unit/cli/mocha-flags.spec.js
```diff
@@ -0,0 +1,27 @@
...
+const {
+  types,
+  expectedTypeForFlag
+} = require('../../../lib/cli/run-option-metadata');
+
+describe('mocha-flags', function () {
+  describe('expectedTypeForFlag()', function () {
+    it('returns expected type for all mocha flags', function () {
+      Object.entries(types).forEach(([dataType, flags]) => {
+        flags.forEach(flag => {
```
JoshuaKGoldberg, line 12: [Tests] The problem with doing >=1 `.forEach` / loop in general _inside_ a test is that you only get one complaint if _any_ of the cases fail. But it's generally nicer to have a test for each of the cases. That way you get a nice detailed log of all the cases and which ones fail.

We can get that with a cute lil reorder and templatization of the `it` name:

```suggestion
    Object.entries(types).forEach(([dataType, flags]) => {
      flags.forEach(flag => {
        it(`returns expected ${flag}'s type as ${dataType}`, function () {
```

test/node-unit/cli/options.spec.js
JoshuaKGoldberg, line 1: [Praise] Super thorough unit testing here & elsewhere, thank you! 🙌

lib/cli/options.js
```diff
@@ -104,24 +114,41 @@ const nargOpts = types.array
...
+      );
+    }
+
+    const pair = arg.split('=');
+    let flag = pair[0];
+    if (isNodeFlag(flag, false)) {
+      flag = flag.replace(/^--?/, '');
+      return arg.includes('=')
+        ? acc.concat([[flag, pair[1]]])
+        : acc.concat([[flag, true]]);
```
JoshuaKGoldberg, line 148: [Style] This is totally optional, no worries if you hate it and close this conversation without doing anything, but you this `.concat` can be deduplicated:

```suggestion
      return acc.concat([[flag, arg.includes('=') ? pair[1] : true]]);
```

Again, not a problem at all if you don't want this 😄

lib/cli/options.js
JoshuaKGoldberg, line 1: [Praise] I appreciate how thoroughly the new behavior is implemented! It really captures the edge cases and how Mocha did things before. Nice. 🔥

lib/cli/options.js
```diff
@@ -104,24 +114,41 @@ const nargOpts = types.array
...
-      }
-      return acc;
-    },
-    []
-  );
+  const allArgs = Array.isArray(args) ? args : args.split(' ');
+  const nodeArgs = allArgs.reduce((acc, arg, index, allArgs) => {
+    const maybeFlag = allArgs[index - 1];
+
+    if (isNumeric(arg) && (!maybeFlag || !isMochaFlag(maybeFlag))) {
```
JoshuaKGoldberg, line 125: [Question] Why `isNumeric(arg) &&`? Wouldn't we want to check this for other string flags too?

Unless I'm missing something, could this be trimmed down to?:

```suggestion
    if (!maybeFlag || !isMochaFlag(maybeFlag)) {
```

lib/cli/options.js
```diff
@@ -104,24 +114,41 @@ const nargOpts = types.array
...
+      isNumeric(arg) &&
+      isMochaFlag(maybeFlag) &&
+      expectedTypeForFlag(maybeFlag) !== 'number'
+    ) {
+      throw createInvalidArgumentTypeError(
+        `Mocha flag '--${maybeFlag}' given invalid option: '${arg}'`,
+        Number(arg),
+        expectedTypeForFlag(maybeFlag)
+      );
+    }
```
JoshuaKGoldberg, line 140: [Feature] This check seems legitimate and good. I like it. But, it feels odd to me that it's only implemented for expected numeric types. There are booleans and strings too. I think it'd be weird as a user to only get this report for an incorrect number, but not incorrect booleans or strings.

What do you think about generalizing it to all argument types? 

I.e. instead of:

```python
if numeric and expected_type !== 'number':
  throw
```

Doing roughly:

```python
if actual_type !== expected_type:
  throw
```

JoshuaKGoldberg (changes requested):
Wow this really ballooned in scope! On the one hand, I feel bad that what looked like a really small-scoped good first issue is turning into a lot more work. On the other hand, the implementation is looking great and is touching on some edge cases Mocha doesn't yet handle well. So this is looking like a really great new feature addition! 🚀

lib/cli/run-option-metadata.js
Dinika: Removing the `?` since it's not needed.

test/node-unit/utils.spec.js
Dinika: Thanks for the catch. Done!

lib/cli/options.js
Dinika: I've made a rather annoying mistake :disappointed:. I forgot to "submit" my replies to your earlier comments before requesting the second round of review :facepalm: . Really sorry about that.

I will wait for your feedback to [the previous comment](https://github.com/mochajs/mocha/pull/5263#discussion_r1869971177) before addressing this comment because I think they are related.

lib/cli/options.js
Dinika: Same here. I'll wait for your reply to [the previous comment](https://github.com/mochajs/mocha/pull/5263#discussion_r1869971177) before addressing this.

Dinika: I realized that this means that mocha does not currently accept filenames which only has numeric characters (example `mocha 404` is not accepted).

lib/cli/options.js
Dinika: Based on this and [the previous comment](https://github.com/mochajs/mocha/pull/5263#discussion_r1867727259), I've summarized the current behaviour of mocha for different flags and values data-type. 

Columns show flags and rows show values passed to flags.

Red dot 🔴 is placed on cells where mocha currently crashes without a good error message. This PR should definitely address these cases.
Orange dot 🟠 is placed where the behaviour could be improved (imo) perhaps. Mocha does not crash in these situations.

| Value / Flag | No flags  | number ⚑   | boolean ⚑ | array ⚑ |  string ⚑  |
|--------------|-----------|-------------|----------|----------| -----|
| number value  | 🔴 **Eg**: `mocha 123` - Mocha crashes with error "arg.split()"    | **Eg**: `mocha --retries 123` - 123 is assigned to `retries`   | 🔴 **Eg**: `mocha --delay 123`: `delay` is assigned value "true". 123 is treated as positional arg. Mocha crashes with error "arg.split()" | **Eg**: `mocha --file 123`: assigns value "123" to the flag. Appropriate error/warning is thrown if flag can't handle value (eg. file not found).   | **Eg**: `mocha --grep 123` assigns "123" to flag. Appropriate action is taken. |
| boolean value | **Eg**: `mocha true`: "true" is treated like a positional value for filename.    | **Eg**: `mocha --retries true`: "NaN" is assigned to flag. This value is then ignored.      | **Eg**: `mocha --delay true`: works same as `mocha --delay`. true is assigned to flag.   | **Eg** `mocha --spec true`: stringified "true" is added to array of flag values. Appropriate action is taken.   | **Eg** `mocha --ui true`. Stringified "true" is assigned to flag. |
| array value | **Eg**: `mocha v1,v2`: "v1,v2" is treated like a positional value for a filename   | 🟠 **Eg**: `mocha --retries v1,v2`  "NaN" is assigned to flag. This value is then ignored.    | **Eg**: `mocha --delay v1,v2`: value true is assigned to flag. "v1,v2" is treated like a positional filename.   | **Eg**: `mocha --spec v1,v2` Flag has value ["v1", "v2"]  | **Eg**: `mocha --grep v1,v2` String "v1,v2" is assigned to flag |
| string value | **Eg**: `mocha abc`: "abc" is treated like a positional value for a filename.    | 🟠 **Eg**: `mocha --retries abc`: "NaN" is assigned to flag. This value is then ignored.   | **Eg**: `mocha --delay abc`: value `true` is assigned to flag. "abc" is treated like positional value for a filename.   | **Eg**: `mocha --spec abc`. "abc" is added to array of flag values.   | **Eg**: `mocha --grep abc`: Works as expected. |
| No value | - | **Eg**: `mocha --retries`. Error `Not enough arguments following: retries` is thrown | **Eg**: `mocha --delay`: Flag has value `true`| **Eg**: `mocha --spec` Error `Not enough arguments following: spec` is thrown | **Eg**: `mocha grep`: Error `Not enough arguments following: grep` is thrown |


Based on the table above, I think only numeric values (or flags) seem to be causing the problem right now. That being said, I don't mind creating a more generic error handling approach. Lemme know what you prefer. I'll make the changes.

lib/cli/options.js
Dinika: Thanks - really appreciate it :blush:

JoshuaKGoldberg (approved):
🚀 Wonderful! No more notes from me, this is all looking great. 

Since it's a nontrivial change, leaving it open for a bit in case others in @mochajs/maintenance-crew have time to review.

lib/cli/options.js
JoshuaKGoldberg: What a great table 😄. Very helpful. I think the 🟠 ones can be ignored for now, agreed.

Dinika:
Yayy!! Thanks for the thorough and quick review @JoshuaKGoldberg 😊

PR #5294, opened 2025-02-06
fix: resolve ERR_REQUIRE_CYCLE_MODULE for never installed modules

### LINKED ISSUE #5290: Misleading error message when attempting to import from module if module was never installed
### Bug Report Checklist

### Expected

I forgot to npm install a library, and tried to test with Mocha. I expected an error about the missing module.

### Actual

The error claims the missing module is a cyclical import. It is not cyclical, it is just missing

Here is a gist that reproduces the problem using bash: https://gist.github.com/MaxKessler5/9d53a143df67ce1cd4dff9e9f03d2db4

The contents: 

npm init -y
sed -i 's/"type": "commonjs"/"type": "module"/' package.json
echo "// Entry point for the application" > index.js
echo "import { cloneDeep } from \"lodash-es\";" >> index.js
sed -i 's/"test":.*/"test": "mocha"/' package.json
npm install mocha
mkdir test
echo "import \"../index.js\";" > test/test.js
npm test > output.txt

### Versions

$ npx mocha --version
11.1.0

node_modules/.bin/mocha --version
(I'm on windows, not sure how to do this one)

node --version
v22.12.0

### Additional Info

_No response_

Donbasilpeter:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5290

## Overview

The requireOrImport function was originally set up to try importing a file as an ES module first, and if that failed, it would fall back to require. However, since Node.js version 22.12, the require function now automatically handles both ES modules (ESM) and CommonJS (CJS). This caused an issue where calling require after a failed import could lead to a cyclic dependency error.

in the updated code I have used require  first and if fails with requireErr.code === 'ERR_REQUIRE_ESM'  which only happens if version is <22.12.0 and  and target file is esm ,then use the older method.

https://nodejs.org/en/blog/release/v22.12.0

https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js

### DIFF
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -39,47 +39,33 @@ exports.requireOrImport = async (file, esmDecorator) => {
     return formattedImport(file, esmDecorator);
   }
   try {
-    return dealWithExports(await formattedImport(file, esmDecorator));
-  } catch (err) {
+    return require(file);
+  } catch (requireErr) {
     if (
-      err.code === 'ERR_MODULE_NOT_FOUND' ||
-      err.code === 'ERR_UNKNOWN_FILE_EXTENSION' ||
-      err.code === 'ERR_UNSUPPORTED_DIR_IMPORT'
+      requireErr.code === 'ERR_REQUIRE_ESM' ||
+      (requireErr instanceof SyntaxError &&
+        requireErr
+          .toString()
+          .includes('Cannot use import statement outside a module'))
     ) {
+      // In Node.js environments after version 22.11, the `loadESMFromCJS` function 
+      // is used within the `require` function to handle cases where the target file 
+      // is in ESM (ECMAScript Module) format. If the Node.js version is after 22.11, 
+      // the code will import the module without any issues. For older versions, 
+      // this `if` statement is required to properly handle ESM modules.
+      // This `if` statement can be removed once all Node.js environments with current 
+      // support include the `loadESMFromCJS` function.
       try {
-        // Importing a file usually works, but the resolution of `import` is the ESM
-        // resolution algorithm, and not the CJS resolution algorithm. We may have
-        // failed because we tried the ESM resolution, so we try to `require` it.
-        return require(file);
-      } catch (requireErr) {
-        if (
-          requireErr.code === 'ERR_REQUIRE_ESM' ||
-          (requireErr instanceof SyntaxError &&
-            requireErr
-              .toString()
-              .includes('Cannot use import statement outside a module'))
-        ) {
-          // ERR_REQUIRE_ESM happens when the test file is a JS file, but via type:module is actually ESM,
-          // AND has an import to a file that doesn't exist.
-          // This throws an `ERR_MODULE_NOT_FOUND` error above,
-          // and when we try to `require` it here, it throws an `ERR_REQUIRE_ESM`.
-          // What we want to do is throw the original error (the `ERR_MODULE_NOT_FOUND`),
-          // and not the `ERR_REQUIRE_ESM` error, which is a red herring.
-          //
-          // SyntaxError happens when in an edge case: when we're using an ESM loader that loads
-          // a `test.ts` file (i.e. unrecognized extension), and that file includes an unknown
-          // import (which throws an ERR_MODULE_NOT_FOUND). `require`-ing it will throw the
-          // syntax error, because we cannot require a file that has `import`-s.
-          throw err;
-        } else {
-          throw requireErr;
-        }
+        return dealWithExports(await formattedImport(file, esmDecorator));
+      } catch (err) {
+        throw err;
       }
     } else {
-      throw err;
+      throw requireErr;
     }
   }
 };
+
 
 function dealWithExports(module) {
   if (module.default) {
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -42,7 +42,7 @@ exports.requireOrImport = async (file, esmDecorator) => {
     return require(file);
   } catch (requireErr) {
     if (
-      requireErr.code === 'ERR_REQUIRE_ESM' ||
+      requireErr.code === 'ERR_REQUIRE_ESM' ||  requireErr.code === 'ERR_INTERNAL_ASSERTION' ||
       (requireErr instanceof SyntaxError &&
         requireErr
           .toString()
--- a/test/integration/plugins/root-hooks.spec.js
+++ b/test/integration/plugins/root-hooks.spec.js
@@ -136,47 +136,38 @@ describe('root hooks', function () {
     });
 
     describe('support ESM via .js extension w/o type=module', function () {
-      describe('should fail due to ambiguous file type', function () {
-        const filename =
-          '../fixtures/plugins/root-hooks/root-hook-defs-esm-broken.fixture.js';
-        const noDetectModuleRegex = /SyntaxError: Unexpected token/;
-        const detectModuleRegex = /Cannot require\(\) ES Module/;
-
-        it('with --no-experimental-detect-module', function () {
+      const filename =
+        '../fixtures/plugins/root-hooks/root-hook-defs-esm-broken.fixture.js';
+      const [major, minor] = process.version.slice(1).split('.').map(Number);
+      if (major > 22 || (major === 22 && minor >= 7)) {
+        it('It should not fail since nodejs can recognize the file based on syntax', function () {
+          return expect(
+            runMochaForHookOutput([
+              '--require=' + require.resolve(filename),
+              require.resolve(
+                '../fixtures/plugins/root-hooks/root-hook-test.fixture.js'
+              )
+            ]),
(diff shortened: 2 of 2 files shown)


lib/nodejs/esm-utils.js
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
+        throw err;
       }
     } else {
-      throw err;
+      throw requireErr;
     }
   }
 };
 
+
```
JoshuaKGoldberg, line 69: [Style] Nit: unrelated change, let's revert.

```suggestion
```

lib/nodejs/esm-utils.js
JoshuaKGoldberg, line 1: [Testing] This is important code, could you please add some tests? I'm not confident reviewing without the new behavior being tested.

You can find examples of tests covering this file's behaviors in the last few PRs that touched this file in the Git blame.

JoshuaKGoldberg (changes requested):
Thanks for getting this started! 🚀

lib/nodejs/esm-utils.js
Donbasilpeter: sure will do that . Thank you

Donbasilpeter:
@JoshuaKGoldberg I have looked in to the test cases.  

The only  test case that fails is when module=type is not present. 

node js consider this as Ambiguous file but from node 2.7 this is not the case.

https://nodejs.org/en/blog/release/v22.7.0#module-syntax-detection-is-now-enabled-by-default

the function doesn't  handle this case and returns an ERR_REQUIRE_CYCLE_MODULE Error,

the test case just expects this error and passes.  but it is not the actual expected behaviour.

i jsut did a camparison which can be found below :

https://docs.google.com/spreadsheets/d/1T9AsIWK5tLsnqicjGuh9miyPGetemOgxLzGzAK3ZBJA/edit?usp=sharing

please review and feedback so that I can start updating the test case

JoshuaKGoldberg:
😄 It's been a while since I've seen test cases sent in a spreadsheet. Those seem reasonable, thanks for thinking so deeply on them. It's hard to say for sure until we see them implemented in PR review though.

Donbasilpeter:
I also wanted to add that the test case to produce the specific issue is already present in the code:

https://github.com/mochajs/mocha/blob/main/test/integration/esm.spec.js

It can be reproduced by running with v22.12.0

lib/nodejs/esm-utils.js
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
-      err.code === 'ERR_UNSUPPORTED_DIR_IMPORT'
+      requireErr.code === 'ERR_REQUIRE_ESM' ||  requireErr.code === 'ERR_INTERNAL_ASSERTION' ||
+      (requireErr instanceof SyntaxError &&
+        requireErr
+          .toString()
+          .includes('Cannot use import statement outside a module'))
     ) {
+      // In Node.js environments after version 22.11, the `loadESMFromCJS` function 
+      // is used within the `require` function to handle cases where the target file 
+      // is in ESM (ECMAScript Module) format. If the Node.js version is after 22.11,
```
JoshuaKGoldberg, line 53: [Docs] `loadESMFromCJS` is an internal implementation of Node.js. Nothing would stop them from renaming it or refactoring to something else. We in userland don't really care about it anyway; we just care about the behavior of Node.js across versions:

```suggestion
      // When trying to require() an ESM module in Node.js >= 22.11, 
```

lib/nodejs/esm-utils.js
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
+      (requireErr instanceof SyntaxError &&
+        requireErr
+          .toString()
+          .includes('Cannot use import statement outside a module'))
     ) {
+      // In Node.js environments after version 22.11, the `loadESMFromCJS` function 
+      // is used within the `require` function to handle cases where the target file 
+      // is in ESM (ECMAScript Module) format. If the Node.js version is after 22.11, 
+      // the code will import the module without any issues. For older versions, 
+      // this `if` statement is required to properly handle ESM modules.
```
JoshuaKGoldberg, line 55: [Docs] Phrases like _"required to properly handle ..."_ are often indications the docs are unclear. The fact that the code's author believed it required to handle something is implied by the code existing. What readers really need to know is _why_ it's required.

```suggestion
      // the code will import the module without any issues.
      // For older Node.js versions, we fall back to trying an ESM `import()`.
      // Note that the `import()` must come after the `require()` to avoid a cyclic dependency.
      // See https://github.com/mochajs/mocha/issues/5290.
```
```

File: lib/nodejs/esm-utils.js [hunk 4]
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
+      // This `if` statement can be removed once all Node.js environments with current 
+      // support include the `loadESMFromCJS` function.
```
JoshuaKGoldberg, line 57:
```suggestion
      // This `if` statement can be removed once Mocha only supports Node.js versions
      // that allow require(ESM): Node.js ^20.19.0 || >=22.12.0
```

That `20.19.0` is a guess, the wonderful https://github.com/nodejs/node/pull/56927 hasn't been released yet.

lib/nodejs/esm-utils.js
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
+  } catch (requireErr) {
     if (
-      err.code === 'ERR_MODULE_NOT_FOUND' ||
-      err.code === 'ERR_UNKNOWN_FILE_EXTENSION' ||
-      err.code === 'ERR_UNSUPPORTED_DIR_IMPORT'
+      requireErr.code === 'ERR_REQUIRE_ESM' ||  requireErr.code === 'ERR_INTERNAL_ASSERTION' ||
+      (requireErr instanceof SyntaxError &&
+        requireErr
+          .toString()
+          .includes('Cannot use import statement outside a module'))
```
JoshuaKGoldberg, line 49: [Question] Why not go off of `requireErr.code` the way the other checks do?

If there is some reason why it should go off other properties in the error, I'd think checking just `requireErr.message` would be preferable to stringifying the whole thing. In theory someone could name a file `test/Cannot use import statement outside a module.just-to-be-annoying.spec.js`.

lib/nodejs/esm-utils.js
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
-          // import (which throws an ERR_MODULE_NOT_FOUND). `require`-ing it will throw the
-          // syntax error, because we cannot require a file that has `import`-s.
-          throw err;
-        } else {
-          throw requireErr;
-        }
+        return dealWithExports(await formattedImport(file, esmDecorator));
+      } catch (err) {
+        throw err;
       }
```
JoshuaKGoldberg, line 62: [Question] Why a `catch(err)` that only has a `throw err`?

test/integration/plugins/root-hooks.spec.js
```diff
@@ -136,47 +136,38 @@ describe('root hooks', function () {
     });
 
     describe('support ESM via .js extension w/o type=module', function () {
-      describe('should fail due to ambiguous file type', function () {
-        const filename =
-          '../fixtures/plugins/root-hooks/root-hook-defs-esm-broken.fixture.js';
-        const noDetectModuleRegex = /SyntaxError: Unexpected token/;
-        const detectModuleRegex = /Cannot require\(\) ES Module/;
-
-        it('with --no-experimental-detect-module', function () {
```
JoshuaKGoldberg, line 145: [Testing] Why remove the `--no-experimental-detect-module` tests? That's still something we'll have to support.

lib/nodejs/esm-utils.js
```diff
@@ -39,48 +39,34 @@ exports.requireOrImport = async (file, esmDecorator) => {
...
   }
   try {
-    return dealWithExports(await formattedImport(file, esmDecorator));
-  } catch (err) {
+    return require(file);
+  } catch (requireErr) {
     if (
-      err.code === 'ERR_MODULE_NOT_FOUND' ||
-      err.code === 'ERR_UNKNOWN_FILE_EXTENSION' ||
-      err.code === 'ERR_UNSUPPORTED_DIR_IMPORT'
```
JoshuaKGoldberg, line 47: [Testing] This code explicitly handles modules not being found, unknown file extensions, and unsupported ESM directory import errors. For each of those three codes, could you please either reply with the existing test or add a new one? I just want to make sure we're on the same page of what test coverage there is.

JoshuaKGoldberg (changes requested):
Thanks for continuing on this! I have some questions.

JoshuaKGoldberg:
Closing out as it's been a month since last ping. If anybody wants to send a PR tackling this issue, please do. Be sure to add a co-author attribution if it takes code from this PR. Cheers! 🤎

PR #5319, opened 2025-03-26
Build new site

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5315 
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

"Force" Netlify to use `npm` instead of `pnpm` to get around our weird build issues.

Note this issue is only visible in Netlify so non-trusted contributors won't be able to see a lot of the context here.

Ref #5318 for links to a sample build and snippets of logs for context

Issue #5315, opened 2025-03-21
🛠 Repo: New site isn't deployed to mochajs.org/next


mark-wiemer:
### Tooling Suggestion Checklist

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have pulled the latest `main` branch of the repository.
- [X] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [X] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

mochajs.org/next should host the new docs site

### Additional Info

Build logic is hosted in Netlify, so most contributors can't help here :/

mark-wiemer:
(I've been granted access and am poking around now, won't post screenshots as I'm new to Netlify and unsure what might be considered a security risk)

mark-wiemer:
Netlify initializes its build environment by installing dependencies via corepack's `pnpm`. We developers were using the non-corepack `npm` to install. When using corepack's pnpm on my machine, I am able to reproduce the issue:

1. Get a fresh clone of the repo (make sure there are no `node_modules` installed or anything else)
2. Use Node v20 (that's what our Netlify uses, v20.19.0 currently)
3. `npm i -g corepack@latest` (ref https://github.com/nodejs/corepack/issues/612)
4. `cd docs-next`
5. `pnpm install`
6. `npm run docs`

```
$ npm run docs
...
 generating static routes 
18:34:37 ▶ @astrojs/starlight/routes/static/404.astro
18:34:37   └─ /404.html (+29ms)
18:34:37 ▶ @astrojs/starlight/routes/static/index.astro
18:34:37   ├─ /declaring/dynamic-tests/index.html (+18ms)
...
18:34:38   ├─ /features/timeouts/index.html (+13ms)
18:34:38   ├─ /getting-started/index.htmlNamed export 'DecodingMode' not found. The requested module 'entities/lib/decode.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

[...]

PR #5322, opened 2025-03-28
chore: Revert "chore: "Force" Netlify to use npm to build new site"

mark-wiemer:
Reverts mochajs/mocha#5319

This change didn't help, let's clean up where we can :)

Ref #5315

### DIFF
--- a/package.json
+++ b/package.json
@@ -204,6 +204,5 @@
   },
   "overrides": {
     "webdriverio": "^7.33.0"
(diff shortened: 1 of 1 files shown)


mark-wiemer:
Disabled deploy previews in Netlify as they don't work with forks

JoshuaKGoldberg (approved):
It was worth a try at least!

Issue #5048, opened 2023-12-18
mocha failed and shows ERROR: null


Rustam108:
I'm using ts-mocha and put mistaken code like this:
`this.logger?.error(e)`

ts-mocha is failed, and after debugging code of mocha I've detected stupid reaction like this
`
(node:13116) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("ts-node/esm", pathToFileURL("./"));'
(Use `node --trace-warnings ...` to show where the warning was created)
mocha inspect [spec..]

Run tests with Mocha

Rules & Behavior
      --allow-uncaught       Allow uncaught errors to propagate        [boolean]
  -A, --async-only           Require all tests to use a callback (async) or
                             return a Promise                          [boolean]
  -b, --bail                 Abort ("bail") after first test failure   [boolean]
      --check-leaks          Check for global variable leaks           [boolean]
      --delay                Delay initial execution of root suite     [boolean]
      --dry-run              Report tests without executing them       [boolean]
      --exit                 Force Mocha to quit after tests complete  [boolean]
      --fail-zero            Fail test run if no test(s) encountered   [boolean]
      --forbid-only          Fail if exclusive test(s) encountered     [boolean]
      --forbid-pending       Fail if pending test(s) encountered       [boolean]
      --global, --globals    List of allowed global variables            [array]
  -j, --jobs                 Number of concurrent jobs for --parallel; use 1 to
                             run in serial
                                   [number] [default: (number of CPU cores - 1)]
  -p, --parallel             Run tests in parallel                     [boolean]
      --retries              Retry failed tests this many times         [number]
  -s, --slow                 Specify "slow" test threshold (in milliseconds)
                                                          [string] [default: 75]
  -t, --timeout, --timeouts  Specify test timeout threshold (in milliseconds)
                                                        [string] [default: 2000]
  -u, --ui                   Specify user interface    [string] [default: "bdd"]

Reporting & Output
  -c, --color, --colors                     Force-enable color output  [boolean]
      --diff                                Show diff on failure
                                                       [boolean] [default: true]
      --full-trace                          Display full stack traces  [boolean]
      --inline-diffs                        Display actual/expected differences
                                            inline within each string  [boolean]
  -R, --reporter                            Specify reporter to use
                                                      [string] [default: "spec"]
  -O, --reporter-option,                    Reporter-specific options
[...]

PR #5344, opened 2025-04-27
fix: print error when internal cli error happens (more helpful than ERROR: null)

### LINKED ISSUE #5078: 🐛 Bug: mocha 10.2.0 prints out command help text instead of running successfully
### Bug Report Checklist

### Expected

Successfully run tests with mocha 10.2.0

### Actual

```
$ ./node_modules/.bin/mocha -r ts-node/register tests/**/*.ts --reporter mocha-junit-reporter --reporter-options mochaFile=test_reports/mocha/test-results.xml --exit
(node:75854) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("ts-node/esm", pathToFileURL("./"));'
mocha inspect [spec..]

Run tests with Mocha

Rules & Behavior
      --allow-uncaught       Allow uncaught errors to propagate        [boolean]
  -A, --async-only           Require all tests to use a callback (async) or
                             return a Promise                          [boolean]
  -b, --bail                 Abort ("bail") after first test failure   [boolean]
      --check-leaks          Check for global variable leaks           [boolean]
      --delay                Delay initial execution of root suite     [boolean]
      --dry-run              Report tests without executing them       [boolean]
      --exit                 Force Mocha to quit after tests complete  [boolean]
      --fail-zero            Fail test run if no test(s) encountered   [boolean]
      --forbid-only          Fail if exclusive test(s) encountered     [boolean]
      --forbid-pending       Fail if pending test(s) encountered       [boolean]
      --global, --globals    List of allowed global variables            [array]
  -j, --jobs                 Number of concurrent jobs for --parallel; use 1 to
                             run in serial
                                   [number] [default: (number of CPU cores - 1)]
  -p, --parallel             Run tests in parallel                     [boolean]
      --retries              Retry failed tests this many times         [number]
  -s, --slow                 Specify "slow" test threshold (in milliseconds)
                                                          [string] [default: 75]
  -t, --timeout, --timeouts  Specify test timeout threshold (in milliseconds)
                                                        [string] [default: 2000]
  -u, --ui                   Specify user interface    [string] [default: "bdd"]

Reporting & Output
  -c, --color, --colors                     Force-enable color output  [boolean]
      --diff                                Show diff on failure
                                                       [boolean] [default: true]
      --full-trace                          Display full stack traces  [boolean]
      --inline-diffs                        Display actual/expected differences
                                            inline within each string  [boolean]
  -R, --reporter                            Specify reporter to use
                                                      [string] [default: "spec"]
  -O, --reporter-option,                    Reporter-specific options
  --reporter-options                        (<k=v,[k1=v1,..]>)           [array]

Configuration
      --config       Path to config file   [string] [default: (nearest rc file)]
  -n, --node-option  Node or V8 option (no leading "--")                 [array]
      --package      Path to package.json for config                    [string]

File Handling
      --extension          File extension(s) to load
                                           [array] [default: ["js","cjs","mjs"]]
      --file               Specify file(s) to be loaded prior to root suite
                           execution                   [array] [default: (none)]
      --ignore, --exclude  Ignore file(s) or glob pattern(s)
                                                       [array] [default: (none)]
      --recursive          Look for tests in subdirectories            [boolean]
  -r, --require            Require module              [array] [default: (none)]
  -S, --sort               Sort test files                             [boolean]
  -w, --watch              Watch files in the current working directory for
                           changes                                     [boolean]
      --watch-files        List of paths or globs to watch               [array]
      --watch-ignore       List of paths or globs to exclude from watching
                                      [array] [default: ["node_modules",".git"]]

Test Filters
  -f, --fgrep   Only run tests containing this string                   [string]
  -g, --grep    Only run tests matching this string or regexp           [string]
  -i, --invert  Inverts --grep and --fgrep matches                     [boolean]

Positional Arguments
  spec  One or more files, directories, or globs to test
                                                     [array] [default: ["test"]]

Other Options
  -h, --help             Show usage information & exit                 [boolean]
  -V, --version          Show version number & exit                    [boolean]
      --list-interfaces  List built-in user interfaces & exit          [boolean]
      --list-reporters   List built-in reporters & exit                [boolean]

✖ ERROR: null
```

https://github.com/khteh/Node.JSRestAPI

### Versions

`10.2.0`
`v21.5.0`

### Additional Info

_No response_

Rob--W:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5078 / #5048 / others (the `ERROR: null` part of it)

## Overview

There have been issues where the reported error is "ERROR: null", which is very unhelpful. Here is an analysis of a specific example: https://github.com/mochajs/mocha/issues/5048#issuecomment-2833619212

Although the trigger for that error was fixed by #5074, the unhelpfulness of "ERROR: null" was not addressed.

To help with debugging, this patch prints the original error when this stage is unexpectedly reached.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -61,6 +61,7 @@ exports.main = (argv = process.argv.slice(2), mochaArgs) => {
       debug('caught error sometime before command handler: %O', err);
       yargs.showHelp();
       console.error(`\n${symbols.error} ${pc.red('ERROR:')} ${msg}`);
+      console.error(err);
       process.exit(1);
     })
     .help('help', 'Show usage information & exit')


JoshuaKGoldberg:
Switching to draft pending https://github.com/mochajs/mocha/issues/5078#issuecomment-2845231557. We can't review without a reproduction, and it might be that all buggy behavior was already fixed?

Thanks for the PR in the interim though!

Rob--W:
> Switching to draft pending https://github.com/mochajs/mocha/issues/5078#issuecomment-2845231557. We can't review without a reproduction, and it might be that all buggy behavior was already fixed?

For clarity, this is not about fixing whatever that triggered the error, but about improving the logged diagnostics information when an error (unexpectedly) happens.

While the specific trigger was fixed, it is still possible to end up in this case if some other internal error occurs. If you want to reproduce locally, you could throw an error from https://github.com/mochajs/mocha/blob/abf3dd921544b45c4c09eef8f7c9c3c4481a3d66/lib/cli/run.js#L377

There are probably other ways to trigger this issue, but I haven't looked.

If you are concerned about logging too much information (when an error message is already present), I could add a check to only log the error when the message is null.

lib/cli/cli.js
```diff
@@ -61,6 +61,7 @@ exports.main = (argv = process.argv.slice(2), mochaArgs) => {
       debug('caught error sometime before command handler: %O', err);
       yargs.showHelp();
       console.error(`\n${symbols.error} ${pc.red('ERROR:')} ${msg}`);
+      console.error(err);
```
JoshuaKGoldberg: [Bug] Most of the, the `msg` is actually enough on its own. So I don't think we'd want to log the full `err` - that'd be a lot of extra unnecessary output.

How about only logging `err` if `msg` is falsy? Treating it as a kind of targeted backup?

```suggestion
      console.error(`\n${symbols.error} ${pc.red('ERROR:')} ${msg || err}`);
```

Rob--W: > [Bug] Most of the, the `msg` is actually enough on its own. So I don't think we'd want to log the full `err` - that'd be a lot of extra unnecessary output.

Not logging the error when `msg` is set sounds reasonable to me. I called that out as an option before, at https://github.com/mochajs/mocha/pull/5344#issuecomment-2845274208

> How about only logging `err` if `msg` is falsy? Treating it as a kind of targeted backup?
> 
> ```suggestion
>       console.error(`\n${symbols.error} ${pc.red('ERROR:')} ${msg || err}`);
> ```

The above could work but would hide the fact that this is an internal error, and also loses the stack trace (which is valuable for debugging). I was thinking of the following:

```suggestion
    console.error(`\n${symbols.error} ${pc.red('ERROR:')} ${msg}`);
    if (!msg) {
      console.error(err);
    }
```

I did intentionally not stringify the error (or even assume that it is an `Error` instance). We are in very bad territory, and dumping the raw value (an Error preferably, but could be any promise rejection or thrown value in theory) allows the reader to debug the issue.

JoshuaKGoldberg: Yeah that makes sense. I'm in favor :)

Lightning00Blade: +1 to the having stack trace, as that made it possible to debug https://github.com/mochajs/mocha/issues/5562 much simpler.

mark-wiemer:
Hey @Rob--W, could you add a new unit test for this? Something that would've failed on main but passes with your changes?

Lightning00Blade:
@mark-wiemer Hey if you take another look at the PR I have linked in the comments (https://github.com/mochajs/mocha/issues/5562) You can see that this is happening on internal Mocha Error.
With that one would need to exploit a such failure to reproduce this and have a test, but then the bug should be fixed rendering the test failing.

Can you explain why exactly in this very niche issue (mostly for people to report/fix mocha issue) a test would be required?

mark-wiemer:
Hey @Lightning00Blade, I know this is a pretty niche issue from what we can see, but it's hard to know just how often it occurs to end-users with wonky setups. Mocha specifically had a huge userbase a decade ago, and many folks using it today have patched together build pipelines over the years that are surprisingly weird. So even if an edge case is extremely rare, we want a unit test to cover it. That way, if we need to change behavior later, we can better guarantee that we didn't break any "promises" we made with previous fixes. In short, we test everything because it's best practice :)

Rob--W:
Do you have pointers / examples of preferred practices to use for testing this? I did not find something close when I looked for existing ones at https://github.com/mochajs/mocha/tree/main/test/node-unit/cli

To test this effectively, we would have to inspect the serialized output of the cli execution. The next best alternative is to spy on the console global to detect console log calls.

I'm leaning towards the latter so that I can simply check the raw error instead of having to interpret Node's interpretation of the error serialization.

mark-wiemer:
Hey @JoshuaKGoldberg, can you assist here? I don't have capacity right now :/

JoshuaKGoldberg:
Yeah, this is a tricky one. Normally I'd suggest extracting the code that includes logging into a function in a separate file, and then unit test that function. But this specific area of code is pretty inter-tangled with the yargs stuff.

This is also a single-line change that is easy to verify manually. So I feel a little more comfortable suggesting: how about we merge as-is, and file a followup task for adding test coverage? @mark-wiemer

mark-wiemer:
Sounds good. I'll manually verify later today, merge this if it works, and open a chore to cover this scenario with automated tests

mark-wiemer (approved):
Was able to repro the issue and get the additional information thanks to this PR. See backing issue for details. Thank you very much!

JoshuaKGoldberg (approved):
🚢

Issue #5025, opened 2023-11-06
🐛 Bug: remove util.inherits / convert to classes


jimmywarting:
**Is your feature request related to a problem or a nice-to-have?? Please describe.**
It would be nice to get rid of some NodeJS specific things that makes it better suited for browsers.
one such thing could be the option to remove `util.inherits` and replacing it with classes that can extends things.
resulting in fewer dependencies and smaller code bundle that won't require the browserified version of [util](https://npmgraph.js.org/?q=util#select=exact%3Autil%400.12.5) that has an additional of 20 modules and 12 kb

NodeJS docs have marked it as legacy and has this to say about `util.inherits`

> Usage of util.inherits() is discouraged. Please use the ES6 class and extends keywords to get language level inheritance support

**Describe the solution you'd like**
Change prototype based classes into es-classes

voxpelli:
Fully agree!

Some classes that needs to be converted:

* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/hook.js#L29
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/runnable.js#L55
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/suite.js#L99
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/test.js#L39
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/dot.js#L79
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/landing.js#L114
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/list.js#L76
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/min.js#L50
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/nyan.js#L78
* https://github.com/mochajs/mocha/blob/24560c1532b8e67d1254b489fcc6f84c4033c0e1/lib/reporters/progress.js#L102
[...]

Issue #4228, opened 2020-04-14
lint the jsdoc


boneskull:
I'm getting bit with failing documentation builds because I'm writing docstrings that JSDoc doesn't like (specifically, TS-style).  But there's no linting happening before getting to Netlify. 

We may be able to make JSDoc just check the syntax and exit (and make it part of the `lint` scripts).

This is moot if we want to move forward with maintaining our own TS types (for which there is an open PR that I cannot find atm).  That will most likely require dumping JSDoc in lieu of a TS-based API doc generator, because the types aren't really compatible.

There may be another ESLint-based solution, but I'm not sure.  If someone else thinks this is worth doing, please add `help wanted` label, but I don't want anyone to pick this up unless there's agreement that it's worth bothering with (if we're going to eventually have to drop JSDoc).

JoshuaKGoldberg:
Marking as accepting PRs in lieu of #4154. Even if the whole repo were to be converted to TypeScript it'd still be good to have some linting for JSDocs (i.e. [`eslint-plugin-jsdoc`](https://github.com/gajus/eslint-plugin-jsdoc)).

PR #5351, opened 2025-05-01
chore: move callback and object typedefs to a new types.d.ts

### LINKED ISSUE #5347: 🛠 Repo: Split type declarations into standalone .d.ts files

### Overview

Mocha defines all code types with JSDoc comments in `.js` files right now. For example, this is how `PluginDefinition` is declared:

https://github.com/mochajs/mocha/blob/945d6e3bf5a9de19c3aa26fbdac966a721006b58/lib/mocha.js#L1309-L1316

That type block becomes https://mochajs.org/api/global#PluginDefinition.

We'd like to move to `.d.ts` files for declaring these typedefs for a few reasons:

* It's a cleaner, less verbose, more readable syntax
* It makes #4154 easier
* It makes #4228 easier
* It makes #5309 easier

### Additional Info

We're not committing to converting existing `*.js` files to `*.ts` - runtime source code will stay in JS-land for now. That's a separate discussion

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5347

## Overview

Creates a new `lib/types.d.ts` and moves every `@callback` and `@typedef {Object}` to it. A new `tsconfig.json` & `tsc` script can be run to verify that the types in that file can be type checked.

Does not yet validate types in `.js` files (#4154, #4228). You can preview those by switching `tsconfig.json`'s `allowJs` to `checkJs`. I made sure there are no more _"`Cannot find name ...`"_ errors but fixing the rest of the ~650+ will be a lot more work. Much of the code will be much easier to get type-safe when it's ported from manual function prototypes to classes (#5025).  I also held off porting all the info from `@types/mocha` there to keep the changes minimal.

Moves error constants from `errors.js` to `error-constants.js` to avoid circular dependency errors.

### DIFF
--- a/lib/types.d.ts
+++ b/lib/types.d.ts
@@ -1,0 +1,378 @@
+/**
+ * These types are a preliminary step towards moving Mocha's TypeScript types into Mocha.
+ * They are not yet complete and are not yet fully used in the codebase.
+ * For now, if you're an external user, you should use the types from @types/mocha.
+ */
+
+import type {FSWatcher} from 'chokidar';
+
+import type {constants} from './error-constants.js';
+import type Mocha from './mocha.js';
+import Runner from './runner.js';
+
+/**
+ * Command-line options
+ */
+export interface MochaOptions {
+  /** Propagate uncaught errors? */
+  allowUncaught?: boolean;
+
+  /** Force `done` callback or promise? */
+  asyncOnly?: boolean;
+
+  /** Bail after first test failure? */
+  bail?: boolean;
+
+  /** Check for global variable leaks? */
+  checkLeaks?: boolean;
+
+  /** Color TTY output from reporter? */
+  color?: boolean;
+
+  /** Delay root suite execution? */
+  delay?: boolean;
+
+  /** Show diff on failure? */
+  diff?: boolean;
+
+  /** Report tests without running them? */
+  dryRun?: boolean;
+
+  /** Fail test run if tests were failed? */
+  passOnFailingTestSuite?: boolean;
+
+  /** Fail test run if zero tests? */
+  failZero?: boolean;
+
+  /** Test filter given string. */
+  fgrep?: string;
+
+  /** Tests marked `only` fail the suite? */
+  forbidOnly?: boolean;
+
+  /** Pending tests fail the suite? */
+  forbidPending?: boolean;
+
+  /** Full stacktrace upon failure? */
+  fullTrace?: boolean;
+
+  /** Variables expected in global scope. */
+  global?: string[];
+
+  /** Test filter given regular expression. */
+  grep?: RegExp;
+
+  /** Display inline diffs? */
+  inlineDiffs?: boolean;
+
+  /** Invert test filter matches? */
+  invert?: boolean;
+
+  /** Disable syntax highlighting? */
+  noHighlighting?: boolean;
+
+  /** - Reporter name or constructor. */
+  reporter?: string | Reporter;
+
+  /** Reporter settings object. */
+  reporterOption?: Object;
+
+  /** Number of times to retry failed tests. */
+  retries?: number;
+
+  /** Slow threshold value. */
+  slow?: number;
+
+  /** Timeout threshold value. */
+  timeout?: number | string;
+
+  /** Interface name. */
+  ui?: string;
+
+  /** Run jobs in parallel. */
+  parallel?: boolean;
+
+  /** Max number of worker processes for parallel runs. */
+  jobs?: number;
+
+  /** Hooks to bootstrap the root suite with. */
+  rootHooks?: MochaRootHookObject;
+
+  /** Pathname of `rootHooks` plugin for parallel runs. */
+  require?: string[];
+
+  /** Should be `true` if `Mocha` process is running in a worker process. */
+  isWorker?: boolean;
+
+  watch?: boolean;
+  extension?: string[];
+  recursive?: boolean;
+  sort?: boolean;
+  file?: string[];
+  spec?: string[];
+  ignore?: string[];
+}
+
+/**
+ * Callback to be invoked when test execution is complete.
+ *
+ * @private
+ * @param failures - Number of failures that occurred.
+ */
+export type DoneCB = (failures: number) => void;
+
+export interface RunnerOptions {
+  /** Files to run */
+  files?: string[];
+  /** Command-line options */
+  options?: object;
+}
+
+/**
+ * An alternative way to define root hooks that works with parallel runs.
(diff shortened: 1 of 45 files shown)


voxpelli:
Why everything in one large `.d.ts`?

Also: You can try to use (still sometimes buggy) the `@import` to import instead of using typedef, eg:

```js
/** @import { Foo } from './bar.js' */
```

JoshuaKGoldberg:
> Why everything in one large `.d.ts`?

I don't have preferences on where to organize types, since this feels like a transient _"we'll clean things up over time anyway"_ kind of situation. Do you have a preference for where things should go?

JoshuaKGoldberg:
ping @voxpelli did you mean for https://github.com/mochajs/mocha/pull/5351#issuecomment-2852296087 to be a full review? I'm not clear on whether you're requesting changes and waiting on me, or just posting thoughts pending a full review?

lib/types.d.ts
```diff
@@ -0,0 +1,378 @@
...
+  /** Display inline diffs? */
+  inlineDiffs?: boolean;
+
+  /** Invert test filter matches? */
+  invert?: boolean;
+
+  /** Disable syntax highlighting? */
+  noHighlighting?: boolean;
+
+  /** - Reporter name or constructor. */
```
mark-wiemer:
```suggestion
  /** Reporter name or constructor. */
```

lib/types.d.ts
```diff
@@ -0,0 +1,378 @@
...
+  reporter?: string | Reporter;
+
+  /** Reporter settings object. */
+  reporterOption?: Object;
+
+  /** Number of times to retry failed tests. */
+  retries?: number;
+
+  /** Slow threshold value. */
```
mark-wiemer:
```suggestion
  /** Slow threshold value, in milliseconds. */
```

lib/types.d.ts
```diff
@@ -0,0 +1,378 @@
...
+  slow?: number;
+
+  /** Timeout threshold value. */
```
mark-wiemer:
```suggestion
  /** Timeout threshold value, in milliseconds. */
```

lib/types.d.ts
```diff
@@ -0,0 +1,378 @@
...
+  ignore?: string[];
+}
+
+/**
+ * Callback to be invoked when test execution is complete.
+ *
+ * @private
+ * @param failures - Number of failures that occurred.
+ */
+export type DoneCB = (failures: number) => void;
```
mark-wiemer: I know in JS all params are implicitly optional, but I think we should be explicit for clarity

```suggestion
export type DoneCB = (failures?: number) => void;
```

lib/types.d.ts
```diff
@@ -0,0 +1,378 @@
...
+  recursive?: boolean;
+  /** Sort test files */
+  sort?: boolean;
+}
+
+/**
+ * Diagnostic object containing unmatched files
+ */
+export interface UnmatchedFile {
+  /** A list of unmatched files derived from the file arguments passed in */
```
mark-wiemer: List? How are entries separated? Or should this be an array? Same for next entry

mark-wiemer: I think the comments for this UnmatchedFile type were copy-pasted wrong :)

package.json
```diff
@@ -163,6 +168,7 @@
     "rollup-plugin-polyfill-node": "^0.8.0",
     "rollup-plugin-visualizer": "^5.6.0",
     "sinon": "^9.0.3",
+    "typescript": "^5.8.3",
```
mark-wiemer, line 174: TS famously doesn't follow semantic versioning, but I believe they only introducing breaking changes in minor versions, never in patches

```suggestion
    "typescript": "~5.8.3",
```

test/unit/errors.spec.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var errors = require('../../lib/errors');
+var constants = require('../../lib/error-constants');
```
mark-wiemer, line 4: If we are adding a reference to error-constants, can we remove the references to `errors`?

package.json
JoshuaKGoldberg, line 1: Fun fact, we include both `@types/glob` and `@types/minimatch` in our nested dev dependencies. Which is problematic because both `glob` and `minimatch` now define their own types! I spent some time digging in locally to debug what was essentially https://www.github.com/DefinitelyTyped/DefinitelyTyped/issues/68125. Irksome.

No code changes in this PR because clearing `node_modules/` and re-installing fixed everything. But I did end up sending:

* https://www.github.com/DefinitelyTyped/DefinitelyTyped/pull/73158
* https://www.github.com/DefinitelyTyped/DefinitelyTyped/pull/73157

package.json
JoshuaKGoldberg: I don't think this is going to net reduce our types / versioning pain 😁. The changes that come out in minor versions are generally improvements to type checking. They save the big config changes for *.0 versions. 

If a TypeScript type system change does introduce new errors for us, they'll show up in CI & block the build. Which'll give us a chance to massage the lockfile. I don't think there's enough benefit to being more restrictive with versions to justify breaking the common `^` convention.

Fun fact, I have had _thoughts_ on the matter: https://www.learningtypescript.com/articles/why-typescript-doesnt-follow-strict-semantic-versioning (not required reading for this PR at all). I think it depends on what you mean by "breaking changes". 😄

test/unit/errors.spec.js
JoshuaKGoldberg: Sorry, I don't follow - what do you mean? `errors.createInvalid*Error` functions are used here a lot, no?

lib/types.d.ts
JoshuaKGoldberg: D'oh! Thanks 🙂

JoshuaKGoldberg: Oh, _ha_, they were copy-pasted correctly. 😂 

https://github.com/mochajs/mocha/blob/c12d85d4ccaae8b2a5a8c0a9de1eaf165657a291/lib/cli/collect-files.js#L127-L128

Fixed in df3bac28.

test/unit/errors.spec.js
mark-wiemer: Ah, missed that somehow, resolving

package.json
mark-wiemer: Fair enough, just wanted to make sure we're aware, resolving this one

mark-wiemer (approved):

mark-wiemer:
@JoshuaKGoldberg  I'm good to merge this if you are

PR #5353, opened 2025-05-01
Feature: ESM configuration file

### LINKED ISSUE #5049: ESM based configuration file
To continue the journey in [supporting ESM in Mocha](https://github.com/mochajs/mocha/releases/tag/v9.0.0), it would be nice to have support for ESM based configuration files.  Currently this is documented as a [known limitation](https://mochajs.org/#current-limitations)
> _[Configuration file](https://mochajs.org/#configuring-mocha-nodejs) can only be a CommonJS file (`.mocharc.js` or `.mocharc.cjs`)_

Allow ESM based configuration files in projects, either with (`.mocharc.js`) or without  (`.mocharc.mjs`) a `type="module"` field in _package.json_
```js
export default {
  timeout: 30000
}
```

Only other current alternative in Node is CJS.

**Additional context**
N / A

ByteB4rb1e:
## PR Checklist

- [X] Addresses an existing open issue: fixes #5049

## Overview

Added support for ESM module (*Michael Jackson Script*) RC configuration file. As the nature of ESM module resolution is asynchronous, this requires the entire call tree to be asynchronous. Though I considered many optimizations along the way, this PR solely focuses on adding support for RC ESM module support.

The PR is purposefully in Draft state, as maybe it makes sense to do some refactoring along the way? I've added my thoughts in the affected commits.


ByteB4rb1e:
As far as I can tell, at least some test case failures are also applicable to the `main` branch at the moment? I will try to configure test-reporting as to get a JSON report for each test suite on my branch as well as `main` and diff them.

JoshuaKGoldberg:
Yeah, if it's also on the `main` branch then you don't have to worry about it.

bin/mocha.js
```diff
@@ -24,119 +24,123 @@ const mochaArgs = {};
...
-
-/**
- * If `value` begins with `v8-` and is not explicitly `v8-options`, remove prefix
- * @param {string} [value] - Value to check
- * @returns {string} `value` with prefix (maybe) removed
- * @ignore
- */
-const trimV8Option = value =>
-  value !== 'v8-options' && /^v8-/.test(value) ? value.slice(3) : value;
+async function main() {
```
JoshuaKGoldberg, line 27: `main` is part of Mocha's public API: https://mochajs.org/api/module-lib_cli#.main. Switching it to return a `Promise` would be a pretty major breaking change. We're really, _really_ hesitant to tackle those.

OTOH, Node.js `^20.19.0 || >=22.12.0` supports require(ESM). So the code could still all be synchronous! I think it'd be fine and reasonable to tell users that ESM config files are only newly supported on Node.js versions with require(ESM). WDYT?

Aside: Node.js 18 just went EOL this week and it's looking likely that Mocha 12 will drop raise Mocha's minimum supported Node.js version to something at least as recent as `20.19.0`. That means Mocha 12 might be able to fully rely on `require(ESM)`. Very exciting.

ByteB4rb1e: > main is part of Mocha's public API

I share your reservations... It seemed to me that somebody needed to get the conversation on the feature request going and yes, this is also a pretty painful PR for me... The diff on `bin/mocha.js` looks horrendous... If the PR leads to no resolution, this is fine with me as well.

> Node.js ^20.19.0 || >=22.12.0 supports require(ESM). So the code could still all be synchronous!

Under the historic context of discussions around the Node.js ecosystem, I'm not quite sure where the compatibility layer between CommonJS and ESM in Node.js will be heading in the future... Migrating towards ESM, to me, seems *all or nothing*. Once migrated towards ESM, *CommonJS* backward-compatibility is straightforward, but the forward compatibility solution provided by Node.js, which inherently changes the behavior of `require`, seems like an unstable solution, catering towards *CommonJS* proponents. It's not ideal and also hinders the adoption of ECMAScript conventionality, IMO. Keeping in mind that *mocha* might find broader adoption with other runtimes, such as Deno, depending on a feature, which is essentially a workaround, might cause bigger issues in the future.

[This](https://medium.com/the-node-js-collection/an-update-on-es6-modules-in-node-js-42c958b890c) is a very interesting read on the historic origins of this subject matter by one of the core contributors to Node.js.
>  ...It would be far too disruptive a change to the ecosystem for us to modify the semantics of require() to allow it to do asynchronous loading...

Yet here we are...

https://github.com/nodejs/node/blob/f552c86fecd6c2ba9e832ea129b731dd63abdbe2/lib/internal/modules/esm/loader.js#L362

All in all, I'm not sold on the approach provided by Node.js. There are a few caveats, still some to-dos left - as far as I can tell from the commit history and comments. I'd rather encourage users to commit to ESM fully, though I can see the challenge in doing so as a popular project such as *mocha*.

> I think it'd be fine and reasonable to tell users that ESM config files are only newly supported on Node.js versions with require(ESM). WDYT?

I agree, though I don't know your userbase, so can't really have an informed opinion on that. [According to Repology](https://repology.org/project/nodejs/versions), adoption of Node.js ver. >= 20 is good, though e.g. Ubuntu 24.04 LTS is still on v18 and some popular CI services use that Ubuntu version for their container images...

ByteB4rb1e:
I've written an idempotent script for comparing the failing test suites of my branch and `main` and additional failures became apparent, which makes me question the feasibility of implementing this feature request with native ESM compatibility, in a manner that does not introduce breaking changes to mocha.
I agree with @JoshuaKGoldberg that the usage of the `require()` compatibility layer for loading ESM modules synchronously might be less disruptive... Sadly... Really would have loved to see native ESM adoption.

How should I proceed from here?

```sh
#!/usr/bin/env sh
get_test_report() {
    workdir="$1"
    remote_url="$2"
    branch="$3"
    spec_basedir="$4"
    output_path="$5"

    cd "$workdir"

    sh -cx "git clone --depth 1 --single-branch --branch '$branch' '$remote_url' ."

    sh -cx "npm install"

    sh -cx "node bin/mocha.js '$spec_basedir'/**/*.spec.js --reporter json-stream" \
    | tee "$output_path"
}

compare_test_reports() {
    spec_basedir="$1"
    diff_out_file="$2"
    workdir="$(mktemp -d)"

    cd "$workdir"

    # define the working directories for execution beforehand so that we can
    # normalize spec paths and make them canonical for comparison.
    test_report_1_workdir="$(mktemp -d)"
    test_report_2_workdir="$(mktemp -d)"

    # paths to output the test reports to
    test_report_1_path="$workdir/main-test-report.json"
    test_report_2_path="$workdir/byteb4rb1e-feature-5049-test-report.json"

    # execute integration test suites on main branch
    get_test_report "$test_report_1_workdir" \
                    'https://github.com/mochajs/mocha.git' \
                    'main' \
                    "$spec_basedir" \
                    "$test_report_1_path" &
    jid="$!"

    # execute integration test suites on PR branch
    get_test_report "$test_report_2_workdir" \
                    'https://github.com/byteb4rb1e/mocha.git' \
                    'feature/5049' \
                    "$spec_basedir" \
                    "$test_report_2_path" &
    jid="$!"

    wait $jid

    # make the path pretty for NT environments
    # only applicable for Cygwin/MSYS2 - in my case
    #diff_out_file="$(
    #    cygpath -w "$diff_out_file" | sed 's|\\|\\\\|g'
    #)"
    #test_report_1_workdir="$(
    #    cygpath -w "$test_report_1_workdir" | sed 's|\\|\\\\|g'
    #)"
    #test_report_2_workdir="$(
    #    cygpath -w "$test_report_2_workdir" | sed 's|\\|\\\\|g'
    #)"
    #
    #test_report_1_path="$(
    #    cygpath -w "$test_report_1_path" | sed 's|\\|\\\\|g'
    #)"
    #test_report_2_path="$(
    #    cygpath -w "$test_report_2_path" | sed 's|\\|\\\\|g'
    #)"

    node << EOF
import { open } from 'node:fs/promises';
import fs from 'node:fs';
import path from 'node:path';
import { createHash } from 'node:crypto';

const testReport1Workdir = '$test_report_1_workdir';
const testReport2Workdir = '$test_report_2_workdir';

const testReport1Path = '$test_report_1_path';
const testReport2Path = '$test_report_2_path';

const diffOutputPath = '$diff_out_file';

const mainResultHashes = {
    pass: [],
    fail: []
};

(async () => {
    const f1 = await open(testReport1Path);
    const f2 = await open(testReport2Path);

    // iterate over every streamed JSON object, extract the case state and
    // metadata, calculate a canonicalized hash and store it in an array for
    // lookup when iterating over the other test results
    for await (const line of f1.readLines()) {
        const [caseState, caseMeta] = JSON.parse(line);

        // skip anything that has nothing to do with a concrete test case
        if (!['fail', 'pass'].includes(caseState)) continue;

        //canonicalize the test suite file path
        const canonFilePath = path.relative(testReport1Workdir, caseMeta.file);

        // calculate a hash of the test case
        const hash = createHash('sha256');
        hash.update(canonFilePath);
        hash.update(caseMeta.fullTitle);
        const hashDigest = hash.copy().digest('hex');

        mainResultHashes[caseState].push(hashDigest);
    }

    for await (const line of f2.readLines()) {
        const [caseState, caseMeta] = JSON.parse(line);

        // skip anything that has nothing to do with a concrete test case
        if (!['fail', 'pass'].includes(caseState)) continue;

        //canonicalize the test suite file path
        const canonFilePath = path.relative(testReport2Workdir, caseMeta.file);

        // calculate a hash of the test case
        const hash = createHash('sha256');
        hash.update(canonFilePath);
        hash.update(caseMeta.fullTitle);
        const hashDigest = hash.copy().digest('hex');

        if (!mainResultHashes[caseState].includes(hashDigest)) {
            var msg  = \`$spec_basedir: mismatched, (is) '\${caseState}': \`;
                msg += JSON.stringify(caseMeta, null, 4);
                msg += '\\n\\n';

            console.log(msg);

            fs.appendFileSync(diffOutputPath, msg);
        }
    }
})()
EOF
}

out_file="$(pwd)/mocha-5094-test-reports-diff";

compare_test_reports 'test/integration' "$out_file"
compare_test_reports 'test/unit' "$out_file"
compare_test_reports 'test/node-unit' "$out_file"
```

[mocha-5094-test-reports-diff.txt](https://github.com/user-attachments/files/20019299/mocha-5094-test-reports-diff.txt)

bin/mocha.js
JoshuaKGoldberg: Unfortunately, we can confidently say that Mocha will perpetually have a large number of users on the exact oldest version of Node.js and browsers it supports. They'll have to stay supported.

> I'm not quite sure where the compatibility layer between CommonJS and ESM in Node.js will be heading in the future...

I'll step in with more confidence then 🙂 - the present & future is:

* Node.js LTS versions can all use require(ESM)
* `require(ESM)` is the only way to synchronously load ESM modules
  * Tip (maybe not needed here): [`Module.createRequire()`](https://nodejs.org/api/module.html#modulecreaterequirefilename) is a good way to use it in ESM code if needed

https://github.com/nodejs/nodejs.org/issues/7441 has some more links.

So, sticking with `require()` is the way to go for this PR.

ByteB4rb1e:
Pardon for going against the PR requirement of running the test suite locally first, but it gives more context to my comment.

@JoshuaKGoldberg, I'm not sure how to weigh convenience against convention, in this case. Would you require the user to understand how to export defaults from an ESM module that aligns with how exports are accessible with CommonJS, or should the config loader treat all possible cases of ESM exports?

I can either adapt the test case to pass, or extend the config loader to treat all cases of ESM exports.

For context:

```javascript
#!/usr/bin/env sh

TMPDIR=$(mktemp -d)

cat << EOF > $TMPDIR/explicit-default.mjs
export default {
  require: ['foo', 'bar'],
  bail: true,
  reporter: 'dot',
  slow: 60
};
EOF

cat << EOF > $TMPDIR/implicit-default.mjs
const config = {
  require: ['foo', 'bar'],
  bail: true,
  reporter: 'dot',
  slow: 60
};

export { config as 'module.exports' }
EOF

cat << EOF > $TMPDIR/explicit.mjs
export const require = ['foo', 'bar'];
export const bail = true;
export const reporter = 'dot';
export const slow = 60;
EOF

for case in 'explicit-default' 'implicit-default' 'explicit'; do

    path="$TMPDIR/$case.mjs"

    command -v cygpath
    test $? -eq 0 && path="$(cygpath -w "$path" | sed 's|\\|\\\\|g')"


    node << EOF
console.log('$path', require('$path'));
EOF
done
```

yields

```
$> sh 5353-demo.sh
C:\msys64\tmp\tmp.Qc1TS3umA2\explicit-default.mjs [Module: null prototype] {
  __esModule: true,
  default: { require: [ 'foo', 'bar' ], bail: true, reporter: 'dot', slow: 60 }
}
C:\msys64\tmp\tmp.Qc1TS3umA2\implicit-default.mjs { require: [ 'foo', 'bar' ], bail: true, reporter: 'dot', slow: 60 }
C:\msys64\tmp\tmp.Qc1TS3umA2\explicit.mjs [Module: null prototype] {
  bail: true,
  reporter: 'dot',
  require: [ 'foo', 'bar' ],
  slow: 60
}
```

> When an ES Module contains both named exports and a default export, the result returned by require() is the [module namespace object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object), which places the default export in the .default property, similar to the results returned by import() [1]

* `implicit-default.mjs` does not yield the result as described in the Node.js documentation [1]
* `explicit-default.mjs` requires an extra step.
* `explicit.mjs` yields the expected result, but this way seems inconvenient in comparison to the CommonJS export.

[1] https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require

JoshuaKGoldberg:
> Would you require the user to understand how to export defaults from an ESM module that aligns with how exports are accessible with CommonJS, or should the config loader treat all possible cases of ESM exports?

That's a great question... We don't have to get this _exactly_ right in the first version - we can always add in support for more edge cases later on. Maybe, whatever can be easily supported in just 1-2 lines of code / checks?

Vaguely:

```js
const rawValue = require(...);
const esmValue = rawValue.default ?? rawValue;
```

How does that strike you?

ByteB4rb1e:
> How does that strike you?

That's reasonable. Pardon for all the back and forth...

I've kept the changes minimal, though moved the `js` function to avoid code duplication and the pollution of the config object itself. This way a `default` property may still be introduced to the config and regressions (in regards to a new `default` property) can only be introduced for the new ESM module loader. I hope that's fine with you.

I've also added a new unit test case, which I seem to have missed.

ByteB4rb1e:
This looks promising and should resolve itself after https://github.com/mochajs/mocha/issues/5358. If there's anything I can support with in the meantime, please let me know.

lib/cli/config.js
```diff
@@ -24,30 +24,45 @@ const utils = require('../utils');
...
-    } catch (err) {
-      if (cwdFilepath) throw err;
+  js: loadModule,
+  mjs: filepath => {
+    const module = loadModule(filepath);
 
-      debug('parsers: retry load as module-relative path: "%s"', filepath);
-      return require(filepath);
-    }
+    return module.default ?? module;
```
JoshuaKGoldberg, line 65: +1 to me on the `module.default ?? module` change. I think it might make sense to do it for `js` too. If the package itself is ESM then users will expect to be able to write `export default` in a `mocha.config.js`.

Thinking out loud: that _might_ be considered a breaking change. If someone happen to have a `module.exports.default = ...` in their `mocha.config.js` then that would change their behavior. [This Sourcegraph search for `default` in Mocha config files](https://sourcegraph.com/search?q=context:global+default+file:mocharc%5C..*+lang:JavaScript+&patternType=keyword&sm=0) shows that nobody in public seems to be doing this. So I think it's fine.

tl;dr, requesting changes: that the `.default ??` should always apply.

test/integration/config.spec.js
```diff
@@ -12,10 +12,12 @@ describe('config', function () {
     var configDir = path.join(__dirname, 'fixtures', 'config');
     var js = loadConfig(path.join(configDir, 'mocharc.js'));
     var cjs = loadConfig(path.join(configDir, 'mocharc.cjs'));
+    var mjs = loadConfig(path.join(configDir, 'mocharc.mjs'));
     var json = loadConfig(path.join(configDir, 'mocharc.json'));
     var yaml = loadConfig(path.join(configDir, 'mocharc.yaml'));
     expect(js, 'to equal', json);
     expect(js, 'to equal', cjs);
+    expect(mjs, 'to equal', cjs);
```
JoshuaKGoldberg, line 20: [Bug] Typo?

```suggestion
    expect(mjs, 'to equal', mjs);
```

mark-wiemer:
Didn't notice this one while reviewing active PRs, but it's the same as nodejs/nodejs.org#5397 so I'm going to close this as a duplicate. Thanks for the effort though and feel free to comment on that PR with any changes you think are necessary!

PR #5365, opened 2025-05-12
refactor(deps): replace find-up with empathic

### LINKED ISSUE #5346: 📦 Package: Replace find-up with empathic
### Feature Request Checklist

### Overview

`find-up` is a relatively [large package](https://npmgraph.js.org/?q=mocha#zoom=w&select=exact%3Afind-up%405.0.0) for what it provides, using quite a few transitive dependencies.

### Suggested Solution

Replace `find-up` with the smaller and [faster](https://github.com/lukeed/empathic/blob/main/benchmarks.md) [`empathic`](https://github.com/lukeed/empathic)

### Alternatives

There is also `find-up-simple` or using `locate-path` directly, but they are both bigger and slower than `empathic` while providing less features.

### Additional Info

I can open a PR for this, it's not a very difficult refactor.

### LINKED ISSUE #5359: perf: switch to empathic for config resolution
We can get an easy performance bump by using empathic for config resolution (traversal up the fs tree).

We shouldn't be traversing often but this is still a gain, even if small in mocha's case. We're also moving an increasing number of tools to use empathic, so this will de-dupe well over time as it allows us to consolidate many packages across the ecosystem into using just this one.

PR ready to go if you want it 👍

beeequeue:
fixes #5346
fixes #5359

## PR Checklist

- [x] Addresses an existing open issue

## Overview

This PR swaps `find-up` for the faster and smaller alternative `empathic`, see #5346 and #5359 for more details.


lib/cli/config.js
```diff
@@ -92,7 +92,7 @@ exports.loadConfig = filepath => {
  * @returns {string|null} Filepath to config, if found
  */
 exports.findConfig = (cwd = utils.cwd()) => {
-  const filepath = findUp.sync(exports.CONFIG_FILES, {cwd});
+  const filepath = find.any(exports.CONFIG_FILES, {cwd});
```
mark-wiemer, line 95: In one call, `findUp.sync` is replaced with `find.any`, in another, it's replaced with `find.up`. Is that intentional?

mark-wiemer: Ref https://github.com/lukeed/empathic/blob/main/src/find.ts, they look different, can we stick with `find.up` for now?

mark-wiemer (changes requested):

lib/cli/config.js
beeequeue: It is intentional, `up` searches for a single path while `any` searches for multiple:
`find.up(path: string)` vs. `find.any(paths: string[])`

JoshuaKGoldberg:
Moving to draft pending discussion in #5346.

voxpelli (changes requested):
Unfortunately I have reviewed this at depth and it would cause regressions as well as having dubious performance gains: https://github.com/mochajs/mocha/issues/5346#issuecomment-2892045608

I see no reason to explore this further. At least not until the upstream package has become a drop-in replacement.

Issue #5372, opened 2025-05-30
🚀 Feature: begin running tests immediately instead of waiting for watcher ready event


jedwards1211:
### Feature Request Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

In my large project with thousands of files, mocha watch mode takes many seconds longer to start the first test run than non-watch mode, unnecessarily wasting time.  I wish it would start running the tests immediately instead of waiting for the watcher to be ready.

### Suggested Solution

To run immediately but still be able to rerun on a file change, we just need to turn off `ignoreInitial` and distinguish between initial `'add'` events and paths added after test startup.  We can compare the `mtime` of the path to the test startup time to decide if the change should trigger a rerun.

I am ready to make a PR for this, the change looks like:

```js
  const startTime = Date.now();
  let ready = false;
  watcher.on('ready', () => {
    ready = true;
  });

  watcher.on('all', async function handleEvent(event, filePath, stat) {
    if (exiting) return;
    if (ready || event === 'unlink' || (stat && stat.mtime > startTime)) {
      if (event === 'add') {
        tracker.regenerate();
      }
      if (tracker.has(filePath)) {
        rerunner.scheduleRun();
      }
      return;
    }
    fs.stat(filePath, (err, stat) => {
      if (stat) handleEvent(event, filePath, stat);
    });
  });

  (async () => {
    if (!globalFixtureContext) {
      debug('triggering global setup');
      try {
        globalFixtureContext = await mocha.runGlobalSetup();
      } catch (err) {
        console.error(err);
      }
    }
    rerunner.run();
  })();
```

### Alternatives

Right now I'm `pnpm patch`ing `mocha` in my projects that have this issue.

JoshuaKGoldberg:
Oh! I've been bit by this too and meant to file an issue. What an annoying performance cut. Yes, my vote is 👍 as a non-breaking feature request.

[...]

Issue #5361, opened 2025-05-12
🛠️ Repo: CI run failures (test failures)


mark-wiemer:
### Tooling Suggestion Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

2026-06-16:

- Merged #6058 to resolve #6037, which should address most of the run failures. We'll keep chipping away at child issues, which are now the best source of truth for up-to-date info :)

2026-03-29:

- "--timeout should allow numeric value" failed once on main: [Run 5193](https://github.com/mochajs/mocha/actions/runs/23710882090/job/69070535853)

2026-03-23:
- "reruns test when file under --watch-files changes" fails rarely in #5409: [Run 4859](https://github.com/mochajs/mocha/actions/runs/22022724150/job/63634561899?pr=5409)

2026-02-16: Codecov has been pretty stable recently, and #5409 fixes some reliability issues with `--watch` tests (see also #5714). 2 out of 20 runs (4875 to 4894) failed, with 3 jobs failing and 3 total tests failing:

- [4880 - reporters](https://github.com/mochajs/mocha/actions/runs/22079249414/job/63801388986):
  - XUnit reporter when 'reporterOptions.output' is provided when fileStream cannot be created when given an invalid pathname
    - "before each" hook for "should throw system error": `Timeout of 1000ms exceeded`
      - new failure, likely a rare timeout issue, we can double timeout here and monitor
      - possibly new because Mocha previously didn't report hook failures? https://github.com/mochajs/mocha/issues/1955 
    - "after each" hook for "should throw system error": `TypeError: cleanup is not a function`
      - new failure, likely connected to the before each hook failing
- [4880 - Codecov](https://github.com/mochajs/mocha/actions/runs/22079249414/job/63801390095): no test failures
[...]

Issue #5357, opened 2025-05-06
📌 Mocha 12 Release Plan


JoshuaKGoldberg:
We're working on another new major version of Mocha! 📦

[Mocha 11](https://github.com/mochajs/mocha/issues/5249) was the first new major version of Mocha in 3 years. It was released on December 2nd, 2024. It contained no new bug fixes or features; it only updated the `glob` dependency (#5148 -> #5250) and raised the minimum supported Node.js version from 14.0.0 to 18.18.0 (#5206 -> #5206). The small changes were because, as mentioned in #5027, we're a new volunteer maintenance team and want to tread carefully with breaking changes.

[Mocha 12](https://github.com/mochajs/mocha/milestone/66) will also be a careful step forward. It will also contain minimal intentional breaking changes in Mocha's code. Mocha 12 will be used to:

* Bump the minimum Node.js version to `^20.19.0 || >=22.12.0`: which will allow Mocha to rely on `require(ESM)` 
* Bump all production dependencies to their latest versions - many of which are ESM-only
* Remove long-deprecated options and any other known dead code
* Remove usage of [`util.inherits`](https://nodejs.org/docs/latest/api/util.html#utilinheritsconstructor-superconstructor) and convert all manual prototype management to ECMAScript classes

You can see the full list of issues in scope under [milestone/66 v12.0.0](https://github.com/mochajs/mocha/milestone/66).

This version upgrade is intentionally meant to be "boring" for end-users. However, it's very exciting for us as maintainers -- in that it will enable us to produce new features that will be very exciting for end-users later on. That includes:

* [milestone/67 v13.0.0](https://github.com/mochajs/mocha/milestone/67): the _next_ major version, where we will start releasing long-requested user-facing breaking changes
* [projects/8 New Website](https://github.com/orgs/mochajs/projects/8): work tracking revamping our documentation website: https://mochajs.org/next

We're looking forward to continuing Mocha's modernization through 2025 and beyond. If there's anything you feel we should work on that isn't already on [Mocha's issue tracker](https://github.com/mochajs/mocha/issues), please do file it there or ask in [Mocha's Discord](https://discord.gg/KeDn2uXhER). We'd love to hear from you. 🤎 

## Remaining work

(Refresh the page for the latest)

- [x] Switch release-please to betas
- [x] Close https://github.com/mochajs/mocha/milestone/66
- [x] Switch release-please to release candidates (#6083)
[...]

PR #5379, opened 2025-06-18
fix: watch mode using chokidar v4

### LINKED ISSUE #5355: 🐛 Bug: watching files outside of CWD is broken
### Bug Report Checklist

### Expected

Prior to https://github.com/mochajs/mocha/pull/5256, I was able to:
- watch files outside of the CWD
- watch files specified with an absolute path
 
I expect to still be able to do this.

### Actual

I am only able to watch files inside the CWD, and they must be specified as relative paths to the CWD.

MCVE: https://github.com/zsalzbank/mocha-watch-mcve

### Versions

mocha: 11.2.2
node: v20.13.1
os: ubuntu
shell: bash

### Additional Info

`chokidar` used to get passed the `watchFiles`, but now it only gets passed the current directory.

### LINKED ISSUE #5374: 🐛 Bug: Not able to watch files with huge node_modules anymore after bumping chokidar to v4
### Bug Report Checklist

### Expected

`mocha --watch` should work with big node_modules codebases

### Actual

`mocha --watch` stopped to work with big node_modules codebases

After merging https://github.com/mochajs/mocha/pull/5256 I am not able to watch files anymore and instead I am receiving this error:

```
Error: EMFILE: too many open files, watch
    at FSWatcher._handle.onchange (node:internal/fs/watchers:207:21)
Emitted 'error' event on FSWatcher instance at:
    at FSWatcher._handleError (/Users/nescalante/dev/exchange-ui/node_modules/mocha/node_modules/chokidar/index.js:540:18)
    at NodeFsHandler._boundHandleError (/Users/nescalante/dev/exchange-ui/node_modules/mocha/node_modules/chokidar/handler.js:302:49)
    at /Users/nescalante/dev/exchange-ui/node_modules/mocha/node_modules/chokidar/handler.js:149:9
    at foreach (/Users/nescalante/dev/exchange-ui/node_modules/mocha/node_modules/chokidar/handler.js:81:9)
    at fsWatchBroadcast (/Users/nescalante/dev/exchange-ui/node_modules/mocha/node_modules/chokidar/handler.js:148:5)
    at FSWatcher.<anonymous> (/Users/nescalante/dev/exchange-ui/node_modules/mocha/node_modules/chokidar/handler.js:197:17)
    at FSWatcher.emit (node:events:517:28)
    at FSWatcher._handle.onchange (node:internal/fs/watchers:213:12) {
  errno: -24,
  syscall: 'watch',
  code: 'EMFILE',
  filename: null
}

Node.js v18.20.8
```

This is because now the watcher is being initialized this way:

```
const watcher = chokidar.watch('.', {
  ignoreInitial: true
});
```

and that is causing chokidar to do a watch in the whole directory

### Versions

Edit by @mark-wiemer:
Mocha ^11.2.2
Node 18.20.8

### Additional Info

_No response_

Arnesfield:
Glob paths are no longer supported by chokidar starting at v4.

This update works around this by resolving `watchFiles` and `watchIgnore` to valid paths and creating a list of matchers to determine if the files should be allowed or ignored through the chokidar `ignored` match function.

This commit reverts changes from 8af0f1a9005a948fbefeb19be618a64dd910d39f so that the watched file changes are not fixed to the current directory.

Additional note: when a `watchFile` path is removed while chokidar is watching it, recreating the `watchFile` path does not trigger events from chokidar to rerun the tests.

## PR Checklist

- [x] Addresses an existing open issue: fixes #5355, fixes #5374 (@mark-wiemer added 5374)


Arnesfield:
Might need to rethink this a bit. Currently, running into some issues:
- The test `reruns test when file matching --watch-files is added` doesn't pass
- The `watchFiles` for this test are: `**/*.xyz`
- From what I understand, the actual test result looks like it has an additional rerun (3 instead of the expected 2) which I think is probably due to the directory being created (`lib/` then `lib/file.xyz`)
- But ignoring the directory in chokidar means its contents are also ignored
- Need a way to not ignore a directory (`lib`) but also only allow contents to trigger a rerun (`*.xyz`)

Arnesfield:
Hi @JoshuaKGoldberg, thanks for the response.

Actually, the changes are really only meant for handling glob paths for `watchFiles` and `watchIgnore` and not so much about the `EMFILE` issue or crashing chokidar. I'm thinking that those issues should have their own pull requests and set of changes instead of here.

To recap, since chokidar v4 [dropped support for globs](https://github.com/paulmillr/chokidar?tab=readme-ov-file#changelog), this caused issues where files are not correctly being watched. Before 8af0f1a9005a948fbefeb19be618a64dd910d39f, the `watchFiles` were being passed to chokidar directly, essentially entrusting the glob path handling to chokidar. After 8af0f1a9005a948fbefeb19be618a64dd910d39f, which upgrades chokidar to v4, these occurred:

- Glob paths for `watchFiles` and `watchIgnore` no longer work as expected.
- The watcher instance is now only watching file changes on the current directory (`chokidar.watch('.')`) instead of the patterns from `watchFiles`.
- The watcher instance no longer ignores patterns from `watchIgnore`.
- Watch mode become noticeably slower to start, probably due to the `glob.sync()` calls, especially for the `watchIgnore` patterns which would normally include `node_modules`.

These are the issues that these changes aim to address which are mostly fixes for the watched files, and I think that the `EMFILE` issue or testing the limits of chokidar would be out of scope for this (even though they're still more or less related to the watch mode).

While the `watchIgnore` patterns being ignored with this update could probably help with the `EMFILE` issue, the issues mentioned could still occur if the user's machine has a lower resource limit or if it does have too many files being read (e.g., by another program), and those could be caused by something outside of mocha/chokidar.

Please let me know what you think. Thanks!

JoshuaKGoldberg:
The assorted recent watch mode issues for reference:

* #5372
* #5374
* #5355

If it's easiest to fix >=1 of these all together in one PR (as in, the changed areas of code intersect a lot). I'd say go for it. 🙂

Arnesfield:
The list of issues seem about right to me, where the first 2 are more or less about the watch mode taking some time to start than before and the last being the main issue. I think the changes here now should already address these issues, unless I missed something. Let me know should we need to update a few things or if there's anything to do on my end, thanks!

JoshuaKGoldberg:
Great! Can you add test(s) to verify the changed behavior? https://github.com/mochajs/mocha/pull/5379#pullrequestreview-2979149465

Arnesfield:
Added the following tests:

- Test for watched files from outside the current working directory
- Test reruns with a thousand watched files

Though, I'm not too confident with the approach for the last added test. Let me know if we should change how we go about it, thanks.

**Edit**: Well, the last test seems to have failed since it kind of relied on the machine's performance. I'll work on a fix for it and limit the change to a single file.

mark-wiemer:
Next step here will be ensuring the test failures are nothing new (ref https://github.com/mochajs/mocha/issues/5361), then I can review the code and if I approve I'll ping Josh to do the same :D

Arnesfield:
Looks like the test for 1000 watched files is failing due to a timeout error (tests have a timeout of 10000). I tried reducing it to 100 files and while it seems to pass, it does not really test the limits of possibly huge projects with thousands of watched files, so it has been removed instead.

Though, the test for watched files outside the current working directory is kept as it is one of the main behaviors that we probably would want to keep in check.

mark-wiemer:
Going to stress test this one as we saw failures in 5 out of 30 runs on main branch with our `--watch` tests, ref #5361.

Ran [20 test runs on my branch](https://github.com/mochajs/mocha/actions/workflows/mocha.yml?query=branch%3Amark-wiemer-fix-5355-watch-mode)

[5/20 runs failed](https://github.com/mochajs/mocha/actions/workflows/mocha.yml?query=branch%3Amark-wiemer-fix-5355-watch-mode+is%3Afailure):
- **[One known issue with `--watch` (run 4048)](https://github.com/mochajs/mocha/actions/runs/16240288768/job/45855917660)**
    - This fails even more on main branch so this PR is an improvement if not a complete fix, still investigating
- **[One newly-discovered error with `--parallel` (run 4051)](https://github.com/mochajs/mocha/actions/runs/16240290292/job/45855933205)**
   - Given we've only done 30 runs before this batch of 20 runs, I'm not too concerned about this, but will check it out
- Two fails due to a known unrelated issue ([run 4039](https://github.com/mochajs/mocha/actions/runs/16240281505/job/45855836865) and [run 4049](https://github.com/mochajs/mocha/actions/runs/16240289254/job/45855919051))
- [One downstream error (Coveralls, run 4046)](https://github.com/mochajs/mocha/actions/runs/16240287025/job/45855913652)

Next steps:

- Learn more about Chokidar code and Mocha test architecture
- Try to reproduce these issues and fixes locally
- Review this PR in more depth (I read it but will need to read it a few more times given the significance of the changes)

mark-wiemer:
@JoshuaKGoldberg your last review asks to write a test when Chokidar crashes, but this PR doesn't fix that issue. The PR does now have a test for watching files outside of cwd which matches the main issue fixed here. The other issue is a new feature and I'm not sure how we'd programmatically test it, open to ideas. My focus right now is still learning the code and validating correctness of the bugfix

lib/cli/watch-run.js
```diff
@@ -137,40 +139,169 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+ * Normalizes the glob path to extract out the path without the glob part
+ * and with the full glob path equivalent. This can return more than one
+ * normalized glob path as `magicalBraces` would have their own patterns
+ * (e.g., `pkg/{lib,dist}` would have the patterns `pkg/lib` and `pkg/dist`).
+ * @param {string} globPath The glob path to normalize.
+ * @returns {NormalizedGlob[]} The normalized glob paths.
+ * @ignore
+ * @private
+ */
+function normalizeGlob(globPath) {
```
mark-wiemer: Requesting changes to ensure full code coverage of each newly-added function:

- pass in a watch file path that isn't a glob
- pass in one that is a glob
- pass in one where `typeof entry === 'string'` but `!currentPattern.isString()` (I'm not sure when we'd get to the second term of this condition)

lib/cli/watch-run.js
```diff
@@ -137,40 +139,169 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+
+/**
+ * Creates a list of path patterns from the provided normalized glob paths.
+ * @param {NormalizedGlob[]} nglobs The normalized glob paths to create the path patterns.
+ * @param {string} basePath The path where mocha is run (e.g., current working directory).
+ * @returns {PathPattern[]} The list of patterns for matching.
+ * @ignore
+ * @private
+ */
+function createPathPatterns(nglobs, basePath) {
```
mark-wiemer: Would it be simpler to reason about this function if it was `createPathPattern(nglob, basePath)` (singular instead of plural)?

lib/cli/watch-run.js
```diff
@@ -204,11 +335,56 @@ const createWatcher = (
...
+    const patterns = createPathPatterns(nglobs, basePath);
+    allowPatterns.push(...patterns);
+
+    // each path of the normalized glob would be watched by chokidar
+    for (const nglob of nglobs) {
+      watchPaths.add(path.resolve(basePath, nglob.path));
+    }
+  }
+
+  // note: can be shortened to `flatMap()`
```
mark-wiemer: What is this `note` comment? Can we remove it or use `flatMap` if it's actually better?

lib/cli/watch-run.js
```diff
@@ -137,40 +139,169 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+  const patterns = [];
+
+  for (const nglob of nglobs) {
+    if (nglob.globPath === null) {
+      // if not a glob, then match the exact path and its contents
+      // by creating an absolute path glob pattern ending with '**/*'.
+      // if the path to match is not a directory,
+      // then the first non-glob path pattern should have matched it
+      const absPath = path.resolve(basePath, nglob.path);
+      patterns.push(
```
mark-wiemer: For our automated tests, can we somehow compare against the return value of this `createPathPatterns` function to ensure it's working correctly? Just spitballing before I forget

lib/cli/watch-run.js
Arnesfield: Thanks @mark-wiemer, I'll look into adding those for the tests. For that last item, yeah I think the second part is kind of unreachable if we already verify `entry` is a type of `string`. I included it there "just in case". We can probably safely remove that and just use `typeof entry !== 'string'` for the condition if that sounds good.

lib/cli/watch-run.js
Arnesfield: Right now, `createPathPatterns` can return an array of 2 objects for 1 `nglob`, so I think it should be kept plural in this case. We could just pass in 1 `nglob` but we'd have to loop for `nglobs` twice (one for `watchFiles` and another for `watchIgnore`) so I figured to just pass in the `nglobs` as is.

lib/cli/watch-run.js
Arnesfield: Yeah, I searched the code for an existing `flatMap` usage but didn't find any, so, just in case, I opted to use a standard `for` loop instead. I'll probably just remove the comment and keep it as is so it's consistent with the previous loop.

lib/cli/watch-run.js
Arnesfield: I'll have to look into this a bit. From what I'm gathering, we'd want to expose/export `createPathPatterns` so we can test it against some known results.

Not sure if we'd want to test internals though since testing `watchFiles` and `watchIgnore` (with the [list](https://github.com/mochajs/mocha/pull/5379#discussion_r2209108612) you've provided) should more or less test the correctness of `createPathPatterns` (which could change implementation if needed). If somehow something breaks in `createPathPatterns` then I think the tests for `watchFiles` and `watchIgnore` should catch that instead.

Let me know what you think, thanks.

Arnesfield:
After taking a look at this again, I now realize that my current solution is flawed, as the pattern `**/foo` would match files and directories named `foo` but would not match anything under it, say `./foo/bar`. Only found out about it after seeing a proposal here to change the default `watchIgnore`s to something like `**/node_modules`.

Hope you don't mind if I end up drastically changing the current implementation. Maybe not that much, but we'll see.

mark-wiemer:
If you think a drastic change is necessary / for the best, go ahead :) with the complexity of this code you're welcome to extract it into its own function and add unit tests to ensure it works as well. All of this glob parsing used to be handled downstream so we didn't have tests explicitly for it

lib/cli/watch-run.js
mark-wiemer: If flatMap is shorter with the same behavior, you're welcome to use it, I just wanted to confirm. But of course if you are changing your implementation and this comment becomes obsolete that's fine too :)

lib/cli/watch-run.js
mark-wiemer: Yeah that makes sense, just want to make sure our code coverage for this section is high given the complexity of the changes and new behaviors present in Mocha that used to be covered in Chokidar v3

lib/cli/watch-run.js
mark-wiemer: I'll review any changes, I'm always learning and forgetting about globs anyway, but I'll see how the official glob library handles things and go from there based on this code, along with manual practical tests. Thanks for keeping at this :)

Arnesfield:
Pushed up an update to handle file paths under the provided `watchFiles` and additional tests for matching glob paths.

- The implementation has been refactored now and `PathPattern` has been updated to have `Set`s for the paths and globs (instead of an `isGlob` boolean).
- Allowed and ignored file path results are now cached to a map (with a size limit of 1000) to skip re-matching already matched file paths.
- Also added `is-path-inside` as a dependency to check for parent paths in the chokidar `ignored` match function.

Let me know should there be any changes needed or if I missed a thing or two, thanks.

**Edit**: I'm not sure why the netlify deploy preview is failing and how to resolve it. Clicking on it takes me to the netlify page with this permission error:

```text
7:51:56 PM: You do not have permissions to view logs for this resource
```

mark-wiemer:
Thanks! You can ignore Netlify, that's the host of mochajs.org, so we only use it for changes to the website, and I think only contributions by trusted contributors will pass the deploy preview. It's not a blocker to approval or merging in any sense :)

mark-wiemer:
I'll review this tomorrow :)

mark-wiemer:
Hey @Arnesfield, I've been digging into this one and I can't get over how complex it all feels. Even our current (broken) glob parsing in this file is dozens of lines, it seems we should be using the glob library more directly in some way to handle all of this for us. The [Chokidar v4 upgrade docs](https://github.com/paulmillr/chokidar#upgrading) have this:

```js
// other way
import { glob } from 'node:fs/promises';
const watcher = watch(await Array.fromAsync(glob('**/*.js')));
```

Is something like that not possible for us? If not, why? I'll continue researching, right now I'm looking at how Chokidar v3 handled globs for a reference point.

Ultimately, if we do need this complex code, I'm going to ask for a lot of unit tests. Because Chokidar used to handle glob parsing for us, we never stress tested it, at least that's what I'm seeing from our current coverage.

mark-wiemer:
Looks like Chokidar v3.5.3 accomplished its glob support with [`anymatch`](https://www.npmjs.com/package/anymatch?activeTab=readme), which hasn't been updated in 3 years but still gets 58M installs a week and has 99% code coverage. I'm not aware of any security issues with it from searching the GitHub Advisory Database, and it only has two dependencies. @JoshuaKGoldberg what are your thoughts on using `anymatch` instead of custom code here? I'm in favor of it given the complexity and low test coverage we currently have.

Arnesfield:
Hey @mark-wiemer, thanks for the response.

Yeah, using `fs.promises.glob('**/*.js')` would retrieve the file paths that match the `**/*.js` pattern and chokidar would watch those paths. Unfortunately, it wouldn't be watching any directories, and so if a new file is added like `src/file.js`, the watcher wouldn't emit events for it, and thus wouldn't rerun the tests. Using `'**/*'` would match directory paths, but for some reason, it won't emit events for file changes that are in the current directory. Additionally, doing `fs.promises.glob()` wouldn't be ignoring paths from `watchIgnore`, so it would even load files in `node_modules`, which can get pretty slow. Only filtering them afterwards just wouldn't be great. There's all that while also needing to be compatible with existing configs and I don't think we'd want projects updating their `watchFiles` and `watchIgnore` or anything like that.

What I think makes the chokidar v3 API very much different from the v4 API is that it seems they kind of [prefer watching directories](https://github.com/paulmillr/chokidar#upgrading) (or exact files) and only ignore the other paths via an `ignored` function (e.g., ignore paths that do not have a `.js` extension). At least, that's the impression I got.

Personally, I'm alright with any solution that would be better than this as long as the issue gets resolved. There's also the option to downgrade back to chokidar v3 until a major release like mocha 12, but I'll leave the decision up to the team here.

lib/cli/watch-run.js
```diff
@@ -137,40 +140,228 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+   * @type {PathMatcher}
+   */
+  const matcher = {
+    allow(filePath) {
+      let allow = allowCache.get(filePath);
+      if (allow !== undefined) {
+        return allow;
+      }
+
+      allow = matchPattern(filePath, allowed.match);
```
mark-wiemer, line 322: Why not call this and quickly check within the folders before calling `minimatch`?

```suggestion
      allow = matchPattern(filePath, allowed.match, true);
```

mark-wiemer:
Hey @Arnesfield, pushed a bit of cleanup and some more docs. Globs are new to me and this is my first big PR on this repo, but I'm happy to approve now :) pinging @mochajs/committers for a second opinion, and of course we'll give you time to review the changes I made to this branch. Thanks for the help, this is a great bugfix and actually a boon to performance! (See #5374 for perf notes)

mark-wiemer (approved):

mark-wiemer:
Test failures are known issues unrelated to this PR, ref #5361

Arnesfield:
Hi @mark-wiemer, thanks. I looked through the changes and I think I'm fine with them. This is probably just me being subjective and nitpicky, but some things I'd like to point out:

1. The `globPaths.reduce()` change could probably be simplified into a `flatMap()` call. Although, I'm guessing the idea behind this change was to reduce the nested loops? I'm alright with that. Though, it does create an additional array and function which introduces some overhead, albeit probably negligible.
2. The `Array.from(pattern.globs).some()` change in the `matchPattern()` function I'm also fine with, just that it kinda creates an additional array and function whenever the `matchPattern()` call reaches that point (which can be called for every watched file change).
3. I noticed the functions `createPathFilter()` and `createPathMatcher()` were added to `exports`? Honestly, I'm still okay with that, but I think these functions should be kept internal to the file unless it's actually needed somewhere else.

Well, that's just me trying to optimize the little things, hope you don't mind.

I am a bit curious about the performance and what causes the ~5-second delay. Not sure when I'll have time to look into that though. Seems interesting nonetheless.

---

On another topic, something is kinda bothering me with the changes I've made. It's a small edge case that I've thought about regarding the directory globs and it's not even really an issue because of how it's written. Right now, if we have a `watchFile` like `src/package*/src/**/*.ts`, the resulting directory paths to match (parsed by `createPathFilter()`) would be `src` and `src/**/*` (meaning that `src` and all directory paths under it _would not be ignored_ by the chokidar `ignored` match function, so a directory path like `src/utils` wouldn't be ignored).

I think the directory patterns to match (if we're using `src/package*/src/**/*.ts`) should be something like:

- `src`
- `src/package*`
- `src/package*/src`
- `src/package*/src/**/*`

This way, paths like `src/utils` would be ignored by the chokidar `ignored` match function.

It's a small thing that doesn't really affect how it's intended to work, since the allowed patterns wouldn't allow `src/utils` and its contents anyway.

I'm probably just looking into it too deeply. Let me know should we want to handle this case too, though it probably works just fine as it is now without going into too much detail.

mark-wiemer:
Thanks @Arnesfield. Perf is definitely important here but I focused on readability first, hence the change to `reduce` and `some`. I didn't notice any significant perf decrease with those changes, so I think we should leave them for now and come back to them once we have more data about the main bottlenecks here.

Regarding adding these functions to `exports`, yeah, let me revert that change. Originally I was going to write unit tests specifically for those functions but I never did that, so that change has no impact other than confusion!

Regarding your edge case concern, are you able to write a test case that reproduces this? If the behavior in your PR is significantly different than Mocha 11.2.1 (with Chokidar 3) or our current code (without your changes), it's definitely important to document. With so many users of Mocha, any potential breaking change should be deeply investigated.

That said, for the record, Mocha 11.2.2 introduces a behavioral change where `--watch-files [ab].js` no longer watches a file literally named `[ab].js`, only files named `a.js` and `b.js`. This matches our documentation (which says the arg is a glob) and if users do want to watch a file of that name, they can pass in `[[]ab[]].js` as the arg instead. But I added a test for clarity to reduce the chance of regression.

Arnesfield:
Looks good, thanks. For the edge case I brought up, the behavior of the watch mode wouldn't change even if it were to be addressed. It only ever applies when the `watchFiles` array includes a regexp before a globstar (e.g., `src/reg*exp/**`). It would be a kind of small, internal improvement rather than something higher up that we can test.

mark-wiemer:
@Arnesfield gotcha, so that edge case indicates a potential perf improvement and refactor, not a potential bug in the behavior that anyone using Mocha would see?

Arnesfield:
Yeah. Actually, maybe "improvement" isn't the right term and more like handling that edge case makes it more _precise_ instead of being just _good enough_. Now that I think about it, it's hard to say if there'd be any performance improvements since being more precise with the checks means more patterns to match.

Anyway, I only brought it up so I wouldn't be the only one that's aware of it, and it could become something worth looking into in the future if it ever becomes an issue (probably not). For now, I'd say the current checks should be good enough.

Let me know should there be anything to do on my end. Thanks!

**Edit:** Meant _precise_ and not _concise_*

mark-wiemer:
Thanks @Arnesfield. Just FYI, we're marking this one as semver-major, but Mocha 12 is planned soon (#5357). The likelihood of breaks with this change is just too high.

lib/cli/watch-run.js
```diff
@@ -137,44 +139,226 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+
+  /**
+   * Performs a `map.set()` but will clear the map whenever the limit is reached.
+   * @param {Map<string, boolean>} map The map to use.
+   * @param {string} key The key to use.
+   * @param {boolean} value The value to set.
+   */
+  function cache(map, key, value) {
+    if (map.size >= MAX_CACHE_SIZE) {
+      map.clear();
```
jedwards1211: Why clear the map instead of just removing the first key from the keys() iterator here?

lib/cli/watch-run.js
```diff
@@ -137,44 +139,226 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
-    const globOpts = {
+  // for checking if a path ends with `/**/*`
+  const globEnd = path.join(path.sep, '**', '*');
+  
+    /**
+     * The current glob pattern to check.
+     * @type {typeof Pattern[]} The `Pattern` class is not exported by the `glob` package.
+     * Ref [link](../../node_modules/glob/dist/commonjs/pattern.d.ts)
+     */
+  const patterns = globPaths.reduce((acc, globPath) => {
```
jedwards1211: I think there were comments about this, but .flatMap is a better way to do this

lib/cli/watch-run.js
```diff
@@ -137,44 +139,226 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+
+  if (matchParent) {
+    for (const childPath of pattern.paths) {
+      if (isPathInside(childPath, filePath)) {
+        return true;
+      }
+    }
   }
+
+  return Array.from(pattern.globs).some(globPath =>
```
jedwards1211: This code will be run on potentially lots of files, it would be better to just iterate the set looking for a glob that matches instead of creating an array every time we test a path

lib/cli/watch-run.js
```diff
@@ -198,36 +382,46 @@ const createWatcher = (
...
+
+  const basePath = process.cwd();
+  const allowed = createPathFilter(watchFiles, basePath);
+  const ignored = createPathFilter(watchIgnore, basePath);
+  const matcher = createPathMatcher(allowed, ignored, basePath);
+
+  // Chokidar has to watch the directory paths in case new files are created
+  const watcher = chokidar.watch(Array.from(allowed.dir.paths), {
+    ignoreInitial: true,
+    ignored: matcher.ignore
```
jedwards1211, line 434: We could also ignore directories that don't match a prefix of any watch pattern

jedwards1211 (commented):
I think we should reduce unnecessary CPU and Memory usage in a few spots since the number of files X number of watch paths may get large.

jedwards1211:
I think we could afford to make those match caches bigger, too, like up to 10000

lib/cli/watch-run.js
jedwards1211: `n` may not be large here, but reduce-concat is `O(n^2)` whereas flatmap is just `O(n)`

lib/cli/watch-run.js
```diff
@@ -137,44 +139,226 @@ exports.watchRun = (mocha, {watchFiles, watchIgnore}, fileCollectParams) => {
...
+     * @type {typeof Pattern | null} The `Pattern` class is not exported by the `glob` package.
+     */
+    let currentPattern = pattern;
+    let isGlob = false;
+
+    do {
+      // save string patterns until a non-string (glob or regexp) is matched
+      const entry = currentPattern.pattern();
+      const serializedEntry = JSON.stringify(entry);
+      debug('processing glob subpattern: %s', serializedEntry);
```
jedwards1211: You can use `%j` in place of `%s`/`JSON.stringify` here

lib/cli/watch-run.js
Arnesfield: This was a `for` loop previously (https://github.com/mochajs/mocha/pull/5379/commits/5514fe288584a2c29b97fe5092e6bbe5c13f4186, iterating `globPaths`). We probably don't even need the array even if it's only a small performance cost for a slight readability improvement. Still, I'd prefer the `for` loop over an extra array (with `.flatMap()`/`.reduce()`) despite the additional nesting. Thoughts on reverting this change or going for `flatMap` @mark-wiemer?

lib/cli/watch-run.js
Arnesfield: Directory paths not part of or parent of `watchFiles` are already ignored in https://github.com/mochajs/mocha/pull/5379/files/8af5d0bf89d1e31a61a7c364a67c4e8390a78f14#diff-14a1e3aa193d6f208b9ffdfeec1876f2cbc0e51007d061a87644e9ea9df59676R347-R350

lib/cli/watch-run.js
Arnesfield: Good point. The change would be something like:

```js
// also make sure to update the doc comment for this function
// ...
if (map.size >= MAX_CACHE_SIZE && !map.has(key)) {
  map.delete(map.keys().next().value);
}
// ...
```

Where `!map.has(key)` makes sure to skip the delete if the key already exists (which wouldn't increase the map size).

lib/cli/watch-run.js
Arnesfield: Agreed. We can revert this back to the previous `for` loop (https://github.com/mochajs/mocha/pull/5379/commits/caca6507ce8eae9c6ba4d53263088601b4ff5a05), but I'll let @mark-wiemer decide what we'd want to do.

Thanks for the feedback @jedwards1211.

lib/cli/watch-run.js
jedwards1211: Ah, gotcha

jedwards1211:
@mark-wiemer 

> Perf is definitely important here but I focused on readability first, hence the change to reduce and some

I don't see how reduce-concat is so much more readable than a nested for loop (or especially flatMap?) to justify its O(n^2) cost (where n is the size of the final array).  It's a good pattern in FP languages based on linked lists where it's O(n); not such a good pattern in languages like JS that aren't based on linked lists.

Array.from(set).some(...) is more readable but it creates a array from the set for every call, which is a needless waste for a slight increase in readability (a for loop isn't that much harder to read).

lib/cli/watch-run.js
mark-wiemer: I had similar thoughts but didn't want to prematurely optimize :) if this has the same perf I'm fine with it, still very clean code :)

lib/cli/watch-run.js
mark-wiemer: If flatMap works without introducing new dependencies let's go for it. If it's a whole new dependency just for this optimization, I'd have to see some numbers when this would have any significant impact. Given it's a simple iteration over patterns (not even matched files) I can't see it saving more than 1ms per run. That said, I do like clean code, and if flatMap is simpler / easier to maintain I'm happy to consider it. I'd just have to see them side-by-side :)

lib/cli/watch-run.js
mark-wiemer: Ah yeah I think my change here hurts performance, and since this is the number of files it could be significant. Feel free to revert if it improves perf and is still readable, maybe adding a comment like "use raw data to save time instead of copying into array"

lib/cli/watch-run.js
mark-wiemer: resolving this one :)

lib/cli/watch-run.js
mark-wiemer: Nice, good to know moving forward, others are welcome to make this change but since it only affects debug builds I'm not gonna prioritize it :D

mark-wiemer:
> @mark-wiemer
> 
> > Perf is definitely important here but I focused on readability first, hence the change to reduce and some
> 
> I don't see how reduce-concat is so much more readable than a nested for loop (or especially flatMap?) to justify its O(n^2) cost (where n is the size of the final array). It's a good pattern in FP languages based on linked lists where it's O(n); not such a good pattern in languages like JS that aren't based on linked lists.
> 
> Array.from(set).some(...) is more readable but it creates a array from the set for every call, which is a needless waste for a slight increase in readability (a for loop isn't that much harder to read).

Yeah, I may have been over-valuing readability and prioritizing my preferences as "objective." Happy to revert some of my changes, see other comments for details :)

Arnesfield:
Pushed up changes to address the recent comments. Let me know if I missed anything, thanks.

Some notes:

- Opted for array `flatMap` over the `for` loop to replace the `reduce` call.
- Removed `JSON.stringify(entry)` since `entry` can be a symbol (for globstar) and `JSON.stringify(symbol)` returns `undefined`.
- I would have preferred `const MAX_CACHE_SIZE = 10_000;` but it seems to fail during lint since it's using ES2020 ([numeric separators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators) seem to be a feature under ES2021).

lib/cli/watch-run.js
```diff
@@ -258,9 +260,16 @@ function matchPattern(filePath, pattern, matchParent) {
     }
   }
 
-  return Array.from(pattern.globs).some(globPath => 
-    minimatch(filePath, globPath, {dot: true, windowsPathsNoEscape: true})
-  );
+  // loop through the set of glob paths instead of converting it into an array
+  for (const globPath of pattern.globs) {
```
mark-wiemer, line 277:
```suggestion
  // provides minor perf increase
  for (const globPath of pattern.globs) {
```

dylan-chong:
any progress on getting this merged? for the last 3 months i've been using mocha 11.1.0 just so i can use the watch feature...

i know youre risking a breaking change to a core feature, but it's already broken...

mark-wiemer:
> any progress on getting this merged? for the last 3 months i've been using mocha 11.1.0 just so i can use the watch feature...
> 
> i know youre risking a breaking change to a core feature, but it's already broken...

I'll work with @JoshuaKGoldberg on this one. I would like to add it to v12 personally but I haven't looked at this in the past couple weeks and don't have a ton of expertise with glob matching, so I'm not confident I'd be able to identify subtle bugs here.

dylan-chong:
Hey guys, any progress here?

This watch is super important to my workflow, without it, life is so much slower!
(Also my old mocha 11.1.0 stopped working due to some ESM/CJS shenans so I can't use that anymore nor do I want to try fix that)

The solution doesn't have to work perfectly. You'll have a chance to iron out the kinks if bug reports come in, but right now the watch mode is just a dead feature on most projects.

test/integration/options/watch.spec.js
```diff
@@ -267,6 +407,30 @@ describe('--watch', function () {
       });
     });
 
+    // New breaking change in Mocha 11.2.2
```
JoshuaKGoldberg: [Docs] The versioning isn't relevant, just the root issue. I'm also overly sensitive to _"breaking change"_ being applied to accidental bug introductions. That implies the public stated contract was broken -- not the same as a bug in the contract's implementation. /rant

```suggestion
```

JoshuaKGoldberg:
Good timing: Mark and I are now both more free than we've been in a while, and focusing heavily on getting Mocha 12 through the door (#5357). It'd be nice to have this important fix in 11 - so it's top of my queue now.

lib/cli/watch-run.js
```diff
@@ -4,16 +4,24 @@ const logSymbols = require('log-symbols');
...
+const {minimatch} = require('minimatch');
 const Context = require('../context');
 const collectFiles = require('./collect-files');
-const glob = require('glob');
 
 /**
  * @typedef {import('chokidar').FSWatcher} FSWatcher
+ * @typedef {import('glob').Glob['patterns'][number]} Pattern
+ * The `Pattern` class is not exported by the `glob` package.
+ * Ref [link](../../node_modules/glob/dist/commonjs/pattern.d.ts).
```
JoshuaKGoldberg, line 17: Could we file an issue on glob to export it?

jedwards1211: They probably consider this internal API.  I think we should ask them to export a public API like `Glob.test(file: string): boolean` that we can use instead of copying so much of its internal logic into `createPathFilter` here.

jedwards1211: I filed this issue, let's see what they say: https://github.com/isaacs/node-glob/issues/632

mark-wiemer: The comment here talks about exporting the Pattern class, but the issue talks about exporting a `glob.test` function, I think there was a disconnect here. Regardless, not a blocker to this PR :)

mark-wiemer:
@JoshuaKGoldberg still LGTM, I fixed the lint in #5475 and we'll work on test failures in #5361 of course. Ready to merge when you are :) just not sure if we're including this in v11 or if it'll be the first PR of v12

JoshuaKGoldberg:
Let's ship now so v11 users can benefit from the important bug fix. 🚢 !

lib/cli/watch-run.js
jedwards1211: Yes, I was saying above, the Pattern class seems internal (I doubt they would be willing to export it).  You're welcome to ask them if they'll export it though

Arnesfield: I posted a comment on the issue (https://github.com/isaacs/node-glob/issues/632#issuecomment-3358662007) and it looks like the logic that uses the `Pattern` class for path parsing could be improved by using `minimatch` directly. Not sure when I'll have time to refactor it (I'll probably create a separate issue and PR), but I think for now the current implementation should still work as intended.

jedwards1211: One thing to consider: the logic copied from glob for dealing with patterns is a bit nontrivial, and glob has hopefully tested all possible cases that can happen with it, but we probably aren't testing them all here.

This is part of why I think the logic deserves to live in a reusable package.

If I have time soon I may make a proper package for working with multiple patterns, that puts them into a tree for more efficient matching and gives us a `getTopDirs` method, and has thorough unit tests.

Arnesfield: I made some changes to the path parsing logic to use Minimatch instead of Glob `Pattern` and while the tests seem to pass (after handling only a few of the differences in behavior), I came to the conclusion that using Glob `Pattern`s is better as it produces more accurate path segments, especially for Windows paths.

It adds slashes for drives (e.g., `C:` to `C:/`). Getting the path segments of `C:\Users\user` for example:

- Minimatch path segments: `[ 'C:', 'Users', 'user' ]`
- Glob `Pattern` path segments: `[ 'C:/', 'Users', 'user' ]`

> Small note: Glob uses slashes `/` even for Windows.

`path.resolve()` results:

```javascript
// via NodeJS Windows

// using path segments from minimatch
console.log(path.resolve("C:\\a\\b\\c", "C:", "Users", "user"));
// output: C:\a\b\c\Users\user

// using path segments from glob pattern (more accurate)
console.log(path.resolve("C:\\a\\b\\c", "C:/", "Users", "user"));
// output: C:\Users\user
```

UNC paths also produce different path segments. Using `\\host\share\a\b\c` for example:

- Minimatch path segments: `[ '', '', 'host', 'share', 'a', 'b', 'c' ]`
- Glob `Pattern` path segments: `[ '//host/share/', 'a', 'b', 'c' ]`

Glob `Pattern` treats `//host/share/` as one segment. `path.resolve()` seems to handle this case as well:

```javascript
// via NodeJS Windows

// using path segments from minimatch
console.log(path.resolve('C:', '', '', 'host', 'share', 'a', 'b', 'c'));
// output: C:\host\share\a\b\c

// using path segments from glob pattern (more accurate)
console.log(path.resolve('C:', '//host/share/', 'a', 'b', 'c'));
// output: \\host\share\a\b\c
```

Refactoring to Minimatch as I previously suggested isn't the right call. Using UNC paths as watch paths probably won't make sense, but I guess the option is there so long as Chokidar is capable of watching those paths.

There may be more cases that Glob `Pattern` handles that I haven't looked into. In my opinion, we can keep the current way `Pattern` is referenced. It's hard to say if `Pattern` needs to be exported from the Glob package since we really only need a reference to the _type_ and not the _class_ itself just for explicitly typing a few variables.

Issue #5380, opened 2025-06-18
🐛 Bug: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './esm-register' is not defined by "exports


erayhanoglu:
### Bug Report Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

Mocha should run as expected.

### Actual

After updating to versiyon 11.7.0 mocha fails with this error. If i downgrade to version 11.6.0 it works as expected.

```sh
✖ ERROR: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './esm-register' is not defined by "exports" in /Users/ehanoglu/dev/oslib/async-events/node_modules/@swc-node/register/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:322:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:613:13)
    at resolveExports (node:internal/modules/cjs/loader:636:36)
    at Module._findPath (node:internal/modules/cjs/loader:716:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1198:27)
    at Module._load (node:internal/modules/cjs/loader:1043:27)
    at Module.require (node:internal/modules/cjs/loader:1298:19)
    at require (node:internal/modules/helpers:182:18)
    at requireModule (/Users/ehanoglu/dev/oslib/async-events/node_modules/mocha/lib/nodejs/esm-utils.js:94:12)
    at exports.handleRequires (/Users/ehanoglu/dev/oslib/async-events/node_modules/mocha/lib/cli/run-helpers.js:97:34)
    at async /Users/ehanoglu/dev/oslib/async-events/node_modules/mocha/lib/cli/run.js:358:25 {
[...]

PR #5381, opened 2025-06-18
fix: revert "feat: use require to load esm (#5366)"

legendecas:
## PR Checklist

- [x] Addresses an existing open issue:
  - https://github.com/mochajs/mocha/issues/5380
  - https://github.com/mochajs/mocha/pull/5366#issuecomment-2983939313

## Overview

This reverts commit 41e24a242944da0cfc9d4d6989dede85f648cb40.


voxpelli:
Since pinning to previous version works I would say that we skip revert unless we have an estimated double digit percentage regressions (almost impossible to know though) or the estimated fix may be complex / impossible (if we can't get a fix out within a week)

Thoughts?

legendecas:
From the issue tracker, there are non trivial amount of workflows were broken with #5366. I would prefer a revert and re-land it with proper test coverages (https://github.com/mochajs/mocha/issues/5382).

thatsmydoing:
Another argument for reverting is `engines` claims to support nodejs 18 but I don't think `require` for ESM is supported there at all https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require (and you get a different error when you try to do this).

It also needs a flag `--experimental-require-module` prior to nodejs 20.19 and 22.12 which is also newer than the listed engines.

legendecas:
> Another argument for reverting is `engines` claims to support nodejs 18.

This was covered by [`process.features.require_module`](https://nodejs.org/api/process.html#processfeaturesrequire_module), which is only present after v20.19.0.

voxpelli:
> Another argument for reverting is `engines` claims to support nodejs 18 but I don't think `require` for ESM is supported there at all https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require (and you get a different error when you try to do this).
> 
> It also needs a flag `--experimental-require-module` prior to nodejs 20.19 and 22.12 which is also newer than the listed engines.

@thatsmydoing It is properly feature flagged and a progressive enhancement:

https://github.com/mochajs/mocha/blob/5d617f3591979dfac68d8a76dfec4b33e9ea78f0/lib/nodejs/esm-utils.js#L106-L110

---

@legendecas Of course, I don't want to put undue stress on eg. you, but I also do think that we can eg. fix some of this by what I mentioned in https://github.com/mochajs/mocha/issues/5380#issuecomment-2985041278, in other words add a `|| err.code === 'ERR_PACKAGE_PATH_NOT_EXPORTED'` to here:

https://github.com/mochajs/mocha/blob/5d617f3591979dfac68d8a76dfec4b33e9ea78f0/lib/nodejs/esm-utils.js#L96-L98

thatsmydoing:
Ah sorry, I had missed that. That said, it does seem like `require` ESM interacts differently with loaders across node versions. For example, https://github.com/mochajs/mocha/issues/5388 seems to work fine in node 22 and up but not in node 20.19.2 so there seem to be other complicating factors.

voxpelli:
Yes, it's tracked across #5380, #5382, #5387, #5388 as well as a few PR:s

If someone can unify them that would be great

JoshuaKGoldberg:
I'm trying to fix the root issue. I agree with @voxpelli that if we can fix this quickly, we should just push that as a fix. Will post back soon with some findings.

Borewit:
> Ah sorry, I had missed that. That said, it does seem like `require` ESM interacts differently with loaders across node versions. For example, #5388 seems to work fine in node 22 and up but not in node 20.19.2 so there seem to be other complicating factors.

Issues is not present in Node version ≥ [22.15.0](https://nodejs.org/en/blog/release/v22.15.0). It is reproducable with [22.14.0](https://nodejs.org/en/blog/release/v22.14.0) and older.

JoshuaKGoldberg:
Confirmed that this would fix the reported issues. However, since #5384 fixes the issue without a revert, I think we should go with #5384.

legendecas:
Closing as https://github.com/mochajs/mocha/pull/5384 landed.

JoshuaKGoldberg:
@legendecas I do want to say though, you've been really helpful & very proactive - we really appreciate it. Thanks for helping us with this very tricky area of code! ❤️

Issue #5382, opened 2025-06-18
🛠 Repo: Add tests for issues discovered with native import() requiring


JoshuaKGoldberg:
### Tooling Suggestion Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

Mocha 11.7.0 included #5366, using require to load ESM when supported in the running Node.js version. This was a great feature add for fixing #5314 & #5317, as well as working towards unblocking https://github.com/nodejs/node/pull/57298. 

However, it uncovered at least 2-3 user-reported issues:

* #5380
* https://github.com/mochajs/mocha/pull/5366#issuecomment-2983753452: [asynchronous Node.js loaders/hooks](https://nodejs.org/docs/latest/api/module.html#customization-hooks)
* https://github.com/mochajs/mocha/pull/5366#issuecomment-2983939313: MDN's browser-compat data tests

The root issue here is that the Mocha repo doesn't test those cases. Which is a problem. We should really have test coverage for these core user flows IMO.

### Additional Info

IMO a good next step would be to:

1. Fill in any examples in https://github.com/mochajs/mocha-examples
2. Set up "ecosystem" style tests for PRs to make sure all examples are fixed

I think that can be separately tracked / out of scope for this specific issue.

PR #5383, opened 2025-06-19
test: add esm loader test

legendecas:
## PR Checklist

- [x] Addresses an existing open issue: https://github.com/mochajs/mocha/issues/5382

## Overview

Add module loader test. The new test depends on https://github.com/mochajs/mocha/pull/5381.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -69,6 +69,16 @@ module.exports = [
     }
   },
   {
+    files: [
+      'test/compiler-esm/*.js'
+    ],
+    languageOptions: {
+      sourceType: 'module',
+      // For top-level await support.
+      ecmaVersion: 2022,
+    }
+  },
+  {
     files: ['test/**/*.{js,mjs}'],
     languageOptions: {
       globals: {
--- a/test/compiler-esm/package.json
+++ b/test/compiler-esm/package.json
@@ -1,0 +1,4 @@
+{
+  "//": "Setting default type for .ts files as ESM as well",
+  "type": "module"
+}
--- a/test/compiler-esm/test-tla.js
+++ b/test/compiler-esm/test-tla.js
@@ -1,0 +1,11 @@
+const obj = { foo: 'bar' };
+
+describe('esm written in esm with top-level-await', () => {
+  it('should work', () => {
+    expect(obj, 'to equal', { foo: 'bar' });
+  });
+});
+
+await undefined;
+
+export const foo = 'bar';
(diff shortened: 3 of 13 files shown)


legendecas:
Hmm, it seems like the new test fails starting with Node.js v22.12.0 (which should be addressed by the revert).

voxpelli:
@legendecas Can we add the tests independently of  #5381?

JoshuaKGoldberg (approved):
Thanks, this looks solid to me!

I think this PR in conjunction with #5384 is a very good solution to the introduced 11.17.0 issues.

legendecas:
Rebased with the latest main branch.

JoshuaKGoldberg:
We can ignore the compliance bot, it's just reporting that this doesn't _fix_ a linked issue. This is fine.

legendecas:
I added two new test cases to cover CJS format with module loaders..

JoshuaKGoldberg:
Fantastic, even better! I'm ready to merge this whenever you are. 👍 

Edit: assuming all test failures are also happening on `main` (i.e. are not introduced by this PR)

legendecas:
Seems like the new tests fails on v18. I'm going to split them up to a follow-up PR.

legendecas:
This should be ready to be reviewed again. ~~The compliance failure seems due to no associated issue.~~ Update: done

jdmarshall:
This test breaks Node 18 compatibility.

Module.register doesn't exist until Node 20 and then it's a candidate feature.

JoshuaKGoldberg:
> This test breaks Node 18 compatibility.

What do you mean?

jdmarshall:
> What do you mean?

**Module.register doesn't exist until Node 20** and then it's a candidate feature.

You're relying on effectively experimental APIs that don't even exist in the set of Node versions in package.json->engines

JoshuaKGoldberg:
> Module.register doesn't exist until Node 20 

It's in 18: https://nodejs.org/docs/latest-v18.x/api/module.html#moduleregisterspecifier-parenturl-options. We can also tell it's in 18 because the integration tests on this PR, then on `main`, passed: e.g. https://github.com/mochajs/mocha/actions/runs/15851891395/job/44687301757

> candidate feature
> effectively experimental APIs

Well, at the time of authoring in 18 and 20, they were. But now that 18 is EOL and 20 is no longer under active development, they're roughly stable. It's quite unlikely that 20 would take in a change that alters `Module.register`. But, if it does, this is just internal test code -- we can always change it internally. Nothing in this PR impacts end users.

Issue #5376, opened 2025-06-14
🛠️ Repo: Mocha's integration tests fail locally, and then tests won't run at all afterward


jdmarshall:
### Bug Report Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

Trying to run 'npm run test' locally in order to possibly create a PR to fix another issue:

  333 passing (2m)
  1 pending
  3 failing

...

And then all subsequent runs of the tests don't even start, even after an 'npm run clean'. For some reason nuking node_modules and starting over makes it work again.

### Actual

```plaintext
  333 passing (2m)
  1 pending
  3 failing

  1) esm
       should throw an ERR_MODULE_NOT_FOUND and not ERR_REQUIRE_ESM if file imports a non-existing module:
     
[...]

PR #5384, opened 2025-06-20
always fallback to import if requiring esm module fails

mshima:
## PR Checklist

- [ ] Addresses an existing open issue: fixes #000

### DIFF
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -93,13 +93,8 @@ const requireModule = async (file, esmDecorator) => {
   try {
     return require(file);
   } catch (err) {
-    if (
-      err.code === 'ERR_REQUIRE_ASYNC_MODULE'
-    ) {
-      // Import if the module is async.
-      return formattedImport(file, esmDecorator);
-    }
-    throw err;
+    // Import if require fails.
+    return formattedImport(file, esmDecorator);
   }
 }
 


voxpelli:
@mshima Can you sign the CLA? Also: See #5382 and #5383

lib/nodejs/esm-utils.js
```diff
@@ -93,13 +93,8 @@ const requireModule = async (file, esmDecorator) => {
...
   } catch (err) {
-    if (
-      err.code === 'ERR_REQUIRE_ASYNC_MODULE'
-    ) {
-      // Import if the module is async.
-      return formattedImport(file, esmDecorator);
-    }
-    throw err;
+    // Import if require fails.
+    return formattedImport(file, esmDecorator);
```
JoshuaKGoldberg:
```suggestion
    return dealWithExports(await formattedImport(file, esmDecorator));
```

mshima:
Seems to fix the error locally using node 22, could not reproduce the same error that happened in CI using node 24..

voxpelli:
Would be good to get at least some level of tests in that verifies this, to ensure it stays fixed.

How comfortable are you with adding tests for this? See eg. https://github.com/mochajs/mocha/issues/5382 and https://github.com/mochajs/mocha/pull/5383

JoshuaKGoldberg:
#5383's added testing passes with this change for me on Node.js 22.10.0, 22.12.0, and 24.2.0.

I'm still testing out other issue repros, but if all goes well I'd say we merge this PR then, merge #5383 right after.

Aside: #5376 tracks existing integration test failures. Those aren't something we need to address here.

JoshuaKGoldberg:
Trying this out on other reported bug repros:

* ✅ #5380: https://github.com/JoshuaKGoldberg/repros/tree/swc-node-mocha-11.7.0) 
* ✅ #5387: https://github.com/WJSoftware/dr-fetch/tree/JP/mocha-v11.7.0
* ✅ #5388: https://github.com/Borewit/music-metadata/pull/2435)

Nice!

jdmarshall:
I will check this out later today and see how it goes.

voxpelli:
Only reason not to do this would be to fail fast in cases we know will fail in both – and optimizing to fail fast feels like a very low priority.

So I'm 👍👍

PR #5395, opened 2025-06-27
feat!: detect illegally nested tests

### LINKED ISSUE #4525: Throw on nested it call
It would be nice for Mocha to throw if it founds a nested `it` block. Currently, such blocks are silently ignored, but developers mistakenly can assume that they are called and test something.

I'm talking about such code:
```js
describe('some function', () => {
  it('should work parent', () => {
    console.log('parent it block')

    // Accidentally created nested it block
    it('should work nested', () => {
      console.log('nested it block')
      assert.equal([1, 2, 3].indexOf(4), -1)
    })
  })
})
```

In such a simple example it's easy to spot the error but if you have a complex test structure, it's easy to accidentally create such a structure. And you would think that tests really test your code unless you add logs in nested it block and find out that none of your tests are really called.

Example of a more complex test structure:
```js
const sumWithCondition = (condition, input) => {
  if (condition === 1 || condition === 2) {
    return input + 1
  }
  return input + 2
}

const runTest = (input, expected) => {
  testConditions = [1, 2]

  testConditions.forEach(condition => {
    it('should work with condition', () => {
      console.log('called')
      expect(someFunction(condition, input)).to.be.deep.equal(expected)
    })
  })
}

describe('some complex function', () => {
  it('should work with input 1', () => {
    runTest(1, 2)
  })

  it('should work with input 2', () => {
    runTest(2, 3);
  })
})
```

Throw an error if find nested it block

You could also call nested it blocks, but it would create a strange test structure and it would be consistent with other test libraries (e.g. jest)

**Additional context**
It jest such feature was implemented in https://github.com/facebook/jest/pull/9828
repo to check out with mocha and jest: https://github.com/DimaDK02/mocha-opportunity-to-improve/
Run `yarn jest` to run jest on this test and `yarn mocha` to run mocha

mmorearty:
Throws if there is an it() call inside an it() call. Same for test().

## PR Checklist

- [x] Addresses an existing open issue: fixes #4525

## Overview

Adds some state to keep track of whether we are currently inside a test. If we are, and we enter a new test, throws an error.

### helpful failure messages

If a nested test is detected, the failure that is logged shows the name of both
the outer and inner test. For example:

> Error: Nested test "my inner test" detected inside test "my outer test". Nested tests are not allowed.

### new tests

I added a bunch of tests:

* Both BDD and TDD
* Three kinds of async tests: callback, promise, and async/await
* Test that the test names are included in the error message, as noted above

### global variable

I could find no existing state that was visible and allowed me to determine
whether we were already inside a test. Therefore, this PR creates a new global
variable, `global._mochaExecutionState`. If there is a better, non-global place
to store this, that would be great.

### old failing tests

I don’t know why, but `npm test` had some failing tests even before my changes.
They are unrelated to my proposed changes.

### minor code duplication and code complexity

Before my change, in lib/interfaces/{tdd.js,bdd.js} the main test functions,
context.test and context.it, were super short and simple. I don’t love the fact
that my change adds a bit of complexity to both (and it’s the same code in both
-- but on the other hand, those two functions already contain identical code).

### risk of breaking existing tests

I want to point out that even though this is a desirable feature, I think there
is significant risk of breaking people’s current tests. I became aware of this
nested-test issue when I worked at Asana, and over our very large codebase,
there were at least 20 tests that had this problem. It took a lot of work to
fix them all.

If Mocha just made this change, this breakage could cause people to avoid
upgrading to more recent versions.

Mitigation options to consider:

* Add an option to disable the nested-test checks. My suggestion would be that
  nested-test checks are ON by default, but can be turned off. This way, when
  someone upgrades Mocha, they may get errors, but they can still keep the new
  Mocha without having to stop and invest a lot of time in fixing their tests.
  Also, on by default means that if anyone uses Mocha in a new project, they
  will get the nested-test checks.
* Decide what Semver version bump is appropriate. If the new checks are on by
  default, then this could be considered a breaking change, which would mean a
  major version bump.

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -51,6 +51,22 @@ module.exports = function (suites, context, mocha) {
     },
 
     /**
+     * Check for nested test registration.
+     *
+     * @param {Suite} suite The suite to which a test is being added.
+     * @param {string} title The title of the test being added.
+     */
+    checkForNestedTest: function checkForNestedTests(suite, title) {
+      // Check for nested test registration
+      var currentRunnable = suite.rootSuite().currentRunnable;
+      if (currentRunnable) {
+        throw createUnsupportedError(
+          'Nested test "' + title + '" detected inside "' + currentRunnable.title + '".'
+        );
+      }
+    },
+
+    /**
      * Execute before running tests.
      *
      * @param {string} name
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -56,8 +56,7 @@ module.exports = function (suites, context, mocha) {
      * @param {Suite} suite The suite to which a test is being added.
      * @param {string} title The title of the test being added.
      */
-    checkForNestedTest: function checkForNestedTests(suite, title) {
-      // Check for nested test registration
+    checkForNestedTest: function checkForNestedTest(suite, title) {
       var currentRunnable = suite.rootSuite().currentRunnable;
       if (currentRunnable) {
         throw createUnsupportedError(
--- a/lib/interfaces/tdd.js
+++ b/lib/interfaces/tdd.js
@@ -3,6 +3,8 @@
 var Test = require('../test');
 var EVENT_FILE_PRE_REQUIRE =
   require('../suite').constants.EVENT_FILE_PRE_REQUIRE;
+var errors = require('../errors');
+var createUnsupportedError = errors.createUnsupportedError;
 
 /**
  * TDD-style interface:
@@ -84,6 +86,18 @@ module.exports = function (suite) {
       if (suite.isPending()) {
         fn = null;
       }
+
+      // Check for nested test registration
+      if (global._mochaExecutionState && global._mochaExecutionState.currentRunnable) {
+        var currentRunnable = global._mochaExecutionState.currentRunnable;
+        if (currentRunnable.type === 'test') {
+          throw createUnsupportedError(
+            'Nested test "' + title + '" detected inside test "' + currentRunnable.title + '". ' +
+            'Nested tests are not allowed.'
+          );
+        }
+      }
+
       var test = new Test(title, fn);
       test.file = file;
       suite.addTest(test);
--- a/lib/interfaces/tdd.js
+++ b/lib/interfaces/tdd.js
@@ -87,13 +87,7 @@ module.exports = function (suite) {
         fn = null;
       }
 
-      // Check for nested test registration
-      var currentRunnable = suite.rootSuite().currentRunnable;
-      if (currentRunnable) {
-        throw createUnsupportedError(
-          'Nested test "' + title + '" detected inside "' + currentRunnable.title + '".'
-        );
-      }
+      common.checkForNestedTest(suite, title);
 
       var test = new Test(title, fn);
       test.file = file;
--- a/lib/interfaces/tdd.js
+++ b/lib/interfaces/tdd.js
@@ -88,14 +88,11 @@ module.exports = function (suite) {
       }
 
       // Check for nested test registration
-      if (global._mochaExecutionState && global._mochaExecutionState.currentRunnable) {
-        var currentRunnable = global._mochaExecutionState.currentRunnable;
-        if (currentRunnable.type === 'test') {
-          throw createUnsupportedError(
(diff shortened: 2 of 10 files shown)


lib/interfaces/tdd.js
```diff
@@ -84,6 +86,18 @@ module.exports = function (suite) {
       if (suite.isPending()) {
         fn = null;
       }
+
+      // Check for nested test registration
+      if (global._mochaExecutionState && global._mochaExecutionState.currentRunnable) {
```
JoshuaKGoldberg: > `global.`

😬 We really don't want to add to the global scope if at all possible. Globals have a bad habit of getting mucked around with by other tools and exposing internal state.  Even when they're prefixed with a tool-specific name like `_mocha`. Plus they become very tricky to work with in concurrent/parallel scenarios.

Have you tried adding to the current test context? Searching on `context` and `currentContext` in `lib/` shows a few places that that's used to set up state. I would think populating a property like `currentContext.currentTest` would be a good path here.

In fact, searching for `currentTest` shows that it already exists! So something like `common`, `context`, or `currentTest` should be a place you can put things like this.

WDYT?

JoshuaKGoldberg (changes requested):
Thanks for starting on this! I think there should be a way to avoid globals.

mmorearty:
Absolutely -- I didn't want to use a global variable, but didn't understand the structure of Mocha well enough to avoid it. But I think I have it working now.

When the test runner is called, it is passed a suite. I now store a `currentRunnable` on the suite. (I could not find an existing variable that worked for what we're doing here.)

Also, I tweaked it so that it also detects tests nested inside of hooks. For example, this is now caught (although this is far less likely to occur than a test inside a test):

```js
before('my hook', () => {
  it('should not have this test inside a hook', () => {});
});
```

Thanks for the feedback; please let me know if any changes are needed.

lib/interfaces/tdd.js
```diff
@@ -84,6 +86,15 @@ module.exports = function (suite) {
...
         fn = null;
       }
+
+      // Check for nested test registration
+      var currentRunnable = suite.rootSuite().currentRunnable;
+      if (currentRunnable) {
+        throw createUnsupportedError(
+          'Nested test "' + title + '" detected inside "' + currentRunnable.title + '".'
+        );
+      }
```
JoshuaKGoldberg: [Refactor] This TDD error has the same text as the BDD one. Have you considered making a shared function function for at least the error throwing? Maybe even the checking of the current runnable?

I'm on the fence here and don't mind either way. Personally I normally try to deduplicate when something is repeated 3x. This is only 2x. 🤷 whatever you prefer!

JoshuaKGoldberg (approved):
Looks great to me, nice! I fiddled with it a bunch and couldn't find

No blocking review comments from me. Just one small _"it's fine either way"_ consideration. But, since the backing issue #4525 is a semver-major breaking change, we're going to have to wait to merge this until we're prepping a major version with breaking behavioral changes.

lib/interfaces/tdd.js
mmorearty: I agree, I'll pull this out into a separate function in the file common.js, which contains code that is common to both bdd.js and tdd.js.

mmorearty:
Hi @JoshuaKGoldberg just checking in to make sure you saw that last week I addressed the point you raised about duplicate code :)

lib/interfaces/common.js
```diff
@@ -50,6 +50,22 @@ module.exports = function (suites, context, mocha) {
...
     },
 
+    /**
+     * Check for nested test registration.
+     *
+     * @param {Suite} suite The suite to which a test is being added.
+     * @param {string} title The title of the test being added.
+     */
+    checkForNestedTest: function checkForNestedTests(suite, title) {
+      // Check for nested test registration
```
JoshuaKGoldberg: [Naming] The names didn't match between the export and the function declaration. Plus the comment said roughly the same thing a third time.

```suggestion
    checkForNestedTest: function checkForNestedTest(suite, title) {
```

JoshuaKGoldberg (approved):
Lovely!

No blockers from me other than just waiting for a major version. Thanks!

mark-wiemer:
Calling this out for #5357  :)

Issue #5396, opened 2025-06-30
📝 Docs: Mention top level ESM errors get swallowed and do not fail the suite in certain Node.js versions


samuelhnrq:
### Bug Report Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

ESM errors thrown at top level gets silently swallowed (hidden) and does not fail the suite (as I assume is expected). Mocha ignores the error and keeps going. Makes it really annoying to debug whats wrong. Any `it` tests defined before the error do get "registered" normally though. 

### Actual

Nothing, simply ignores the error and keeps going

### Minimal, Reproducible Example

given a test `test/add.test.ts` suite like this 

```ts
import assert from 'node:assert/strict'

throw new Error()

it('should add stuff', () => {
    const b = 1 + 1;
    assert.equal(2, b)
})
```

Then `mocha --extension ts` will simply output

```
(node:64751) ExperimentalWarning: Type Stripping is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)


  0 passing (0ms)
```

instead of exposing the error.

### Versions

Mocha 11.7.1
Node v24.3.0

### Additional Info

Conversely if you change it to
```ts
import assert from 'node:assert/strict'

it('should add stuff', () => {
    const b = 1 + 1;
    assert.equal(2, b)
})

throw new Error()
```

it will then output
```
Running main test suite...
(node:64877) ExperimentalWarning: Type Stripping is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)


  ✔ should add stuff

  1 passing (1ms)
```

Still no mention of the error.

JoshuaKGoldberg:
[...]

PR #5401, opened 2025-07-10
docs: explain node import swallowing error

samuelhnrq:
Fixes: https://github.com/mochajs/mocha/issues/5396

## PR Checklist

- [x] Addresses an existing open issue: Fixes: #5396 

## Overview

The fix landed in node js main nodejs/node#58957 I belive we now should update the ESM support section of the documentation since this error has been affecting even node LTS. I took the liberty to guess the next node.js version, shouldn't matter if they skip the 24.4.1 patch straight to a new minor, the point will still stand.

I was intentionally vague in the problem description and proposed workaround, and left a link to the original issue so if the person is curious to learn more / understand how the fix works. I want to optimize for the quick answers crowd, the curious crowd can delight itself reading the original issue.

### DIFF
--- a/docs-next/src/content/docs/explainers/nodejs-native-esm-support.mdx
+++ b/docs-next/src/content/docs/explainers/nodejs-native-esm-support.mdx
@@ -29,5 +29,7 @@ More information can be found in the [Node.js documentation](https://nodejs.org/
 - [Watch mode](../running/cli#--watch--w) does not support ES Module test files
 - [Custom reporters](../reporters/third-party) and [custom interfaces](../interfaces/third-party) can only be CommonJS files
 - [Configuration file](../running/configuring) can only be a CommonJS file (`.mocharc.js` or `.mocharc.cjs`)
+- Mocha in Node.js version 24.4.0 or older [silently ignored top level errors in ESM files](https://github.com/mochajs/mocha/issues/5396).
+  If you cannot upgrade to a newer Node.js version, you can add `--no-experimental-require-module` to the `NODE_OPTIONS` environment variable.
 - When using module-level mocks via libs like `proxyquire`, `rewiremock` or `rewire`, hold off on using ES modules for your test files.
   You can switch to using `testdouble`, which does support ESM.
--- a/docs/index.md
+++ b/docs/index.md
@@ -2131,6 +2131,8 @@ More information can be found in the [Node.js documentation](https://nodejs.org/
 - [Custom reporters](#third-party-reporters) and [custom interfaces](#interfaces)
   can only be CommonJS files
 - [Configuration file](#configuring-mocha-nodejs) can only be a CommonJS file (`.mocharc.js` or `.mocharc.cjs`)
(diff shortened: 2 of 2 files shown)


docs-next/src/content/docs/explainers/nodejs-native-esm-support.mdx
```diff
@@ -29,5 +29,8 @@ More information can be found in the [Node.js documentation](https://nodejs.org/
 - [Watch mode](../running/cli#--watch--w) does not support ES Module test files
 - [Custom reporters](../reporters/third-party) and [custom interfaces](../interfaces/third-party) can only be CommonJS files
 - [Configuration file](../running/configuring) can only be a CommonJS file (`.mocharc.js` or `.mocharc.cjs`)
+- There was a [bug](https://github.com/nodejs/node/issues/58945) in node which caused top level errors in ESM files to be ignored. 
+  The fix landed in node 24.4.1, but if your tests seem to fail silently or dont seem to be detected you can add `--no-experimental-require-module`
+  to the `NODE_OPTIONS` environment variable.
```
JoshuaKGoldberg: > [bug](https://github.com/nodejs/node/issues/58945) 

[Accessibility] https://www.w3.org/WAI/WCAG21/Techniques/general/G91 - links should be descriptive in just their text. Some users use link texts to navigate through the page.

Applying a bit of wordcrafting, how does this look to you?

```suggestion
- Mocha in Node.js versions prior to 24.1.1 [silently ignored top level errors in ESM files](https://github.com/mochajs/mocha/issues/5396).
  If you cannot upgrade to that Node.js version or newer, you can add `--no-experimental-require-module`  to the `NODE_OPTIONS` environment variable.
```

JoshuaKGoldberg (changes requested):
A good start! WDYT about these touchups?

docs-next/src/content/docs/explainers/nodejs-native-esm-support.mdx
samuelhnrq: Thanks will adopt your version, I like framing the version in the past, `versions prior to` but do you mean 24.1.1 literally? 24.4.0 is out and still has the bug :grimacing:

JoshuaKGoldberg: I don't personally have a preference, what you changed it to makes sense to me. I suppose we'll know for sure what the range is once they release the next version?

JoshuaKGoldberg (approved):
🙌

docs-next/src/content/docs/explainers/nodejs-native-esm-support.mdx
samuelhnrq: Yeah, I tried to be precise with the wording to be clear the mentioned version is *still* broken, but it would be simpler if we knew the good version, do we merge like this and change the mentioned version later?

Some people might be impacted by this right now and they can reach for the node flag while we wait for the node release. I'll try later to see if its possible to back-port the fix into LTS too...

mark-wiemer (approved):

mark-wiemer:
@JoshuaKGoldberg I think we can merge this, the "compliance" check is failing but things look OK to me, it looks like it was a transient issue

PR #5409, opened 2025-07-30
fix: begin running tests immediately instead of waiting for watcher

### LINKED ISSUE #5372: 🚀 Feature: begin running tests immediately instead of waiting for watcher ready event

jedwards1211:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5372 

## Overview

Turn off `ignoreInitial` and distinguish between initial `'add'` events and paths added after test startup. We can compare the `mtime` of the path to the test startup time to decide if the change should trigger a rerun.


jedwards1211:
@mark-wiemer is using `new Date()` allowed in the CLI code?  An ESLint rule complains about it.  If not is there any way I can get the current time to compare to file timestamps?

Also, let me know if you have any opinions about if/how I should test this change.  It would be nice if I could inject some kind of chokidar mock so I can control timing of the file events to be able to verify in a test that it runs before the ready event is emitted...but the only easy way I think of to do that would be to call things in `watch-run.js` directly from the test process instead of executing `mocha --watch` in a separate process.

lib/cli/watch-run.js
```diff
@@ -215,23 +216,43 @@ const createWatcher = (
...
-      rerunner.scheduleRun();
+    if (!ready && !stat) {
+      fs.stat(filePath, (err, stat) => {
+        if (stat) handleEvent(event, filePath, stat);
+      });
     }
   });
 
+  (async () => {
+    if (!globalFixtureContext) {
```
mark-wiemer: The earlier versions of Mocha using Chokidar 3 still waited until `watcher.on('ready')` to call `mocha.runGlobalSetup()`, but I'm not sure why. A quick look at the code isn't too revealing, but I see why we'd want to call this right away. I'm curious if we can add a test case that has a complex global setup to ensure things still work?

jedwards1211: I'm not exactly familiar with what that global setup is (loading required modules?) but I can try to figure out what goes into it

jedwards1211: Okay Chokidar doesn't seem to specify whether `'change'`/`'unlink'` events can get emitted before `'ready'` if changes occur during its initial scan, but if so, it could cause the preexisting code here to start running tests before it's performed global setup.

Obviously the intention of the preexisting code is to wait for the `'ready'` event to run everything for the first time (first global setup, then the first test run).

I've fixed this in my PR now, if it gets an event before the global fixture context has been created it won't schedule a rerun.

(I still need to write tests for this)

mark-wiemer:
Marking as draft until tests are added

jedwards1211:
@mark-wiemer I want to run some changes by you before I dive into them...

I realized that `GlobFilesTracker` synchronously globs for the watched/ignored files, which mostly defeats the purpose of starting tests before the watcher is ready, since all that globbing would take about as long.  So I think we should make it glob asynchronously.

Also...
- It globs each watch pattern and each ignore pattern separately...but it seems like we can just pass all of the watch and ignore patterns to one glob call?
- The existing code re-globs everything when the watcher gets a path added event, which could be a lot of redundant work...instead I think we should check if the added path is already in the watched or ignored set, and if not and it's a directory, initiate a glob on just that directory?

mark-wiemer:
Hey @jedwards1211, the current glob code is not ideal for sure for exactly the reasons you mentioned and probably more. The best discussion around this is #5379 and in short we don't want to touch this code until v12. We as maintainers don't have a ton of experience with glob matching and don't want to break anything, but we do want to keep Chokidar at v4. Mocha 11.2.1 used Chokidar v3, we upgraded to v4 in Mocha 11.2.2, but Chokidar v4 dropped glob support so we had to use our own. Learning more about globs is definitely a priority, but I'm still getting the ropes on the rest of Mocha as a new maintainer, so I won't have a ton of updates here.

In short, feel free to update or change or even rewrite the glob handling. Just keep #5379 in mind as another alternative and add as many tests as you can to help me and other maintainers verify any changes :)

Related issues include #5355 and #5374

jedwards1211:
Yeah I kind of thought about doing the same thing as that PR.

One question is how to change `blastCache` - if we start and possibly even rerun before the watcher is finished scanning, we can't rely on getting the watched paths from chokidar.  I already had to fix this in this PR, by getting the paths to delete from the GlobFilesTracker.

Instead we would have to delete all `require.cache` paths matching the filter.  I worry a bit about how well that would perform compared to just deleting the list of matched paths we know about though.  If there's a long list of watch patterns it could be a lot of work determining matches.

mark-wiemer:
I wouldn't worry about performance too much in the abstract. We had a concrete issue with #5374 that has effectively been resolved with #5379, and in practice I can't imagine folks are using many regexes for watch paths. If we get a bug report about perf, we can look into it, but as long as our basic stress tests pass, I'd prefer simpler changes first.

On another note, @mochajs/maintenance-crew I'm thinking of marking this as "semver-major" given the complexity impacted and possibility of bugs. Since we want Mocha 12 to have minimal changes we'll probably hold this one for Mocha 13. (#5357)

jedwards1211:
Okay I guess I'll eventually piggyback this on #5379 and blast the cache with the new filtering from that PR.

Also, I'm not sure what to do about getting the launch time to compare file mtimes to.  Date etc could get monkeypatched by test utils like Sinin fake timers... do you have any thoughts?

I guess it won't get monkeypatched before we run global setup though.  But I got an ESLint error about using Date, it must be for good reason.

mark-wiemer:
I'm not sure why our linter warns about using Date. Could you see if that's a recommended rule? There's a real chance that our linter is very out of date. We've had some discussions about it but haven't prioritized updates there for a while.

jedwards1211:
@mark-wiemer looks like the rule was explicitly added to guard against exactly the danger of fake timers I was talking about: https://github.com/mochajs/mocha/issues/237

(This issue is explicitly linked in your ESLint config)

Note, it would probably be possible to grab a reference to the original `Date` constructor, `Date.now()` before it can get monkeypatched, and use that if necessary throughout Mocha.

As I was saying though, in this case I only need to get the launch date before running any global setup where userland code might monkeypatch `Date`, so I guess it's okay to suppress the error in this case

mark-wiemer:
Thanks for researching. The core concern still applies, but I think the workaround does too: just have `Date = global.Date` at the top-level and then call `Date()` within your function as needed. Sinon mocks `global.Date` once the tests start running, but if we do this at top-level, it's done before the tests are run. As long as we never call `global.Date()` within a function we are good. I think... once the code is ready I'll definitely ping other maintainers to review it!

JoshuaKGoldberg:
👋 Just checking in @jedwards1211, are you waiting on us for anything?

jedwards1211:
No, I'll pick this back up soon now that that related PR is merged

jedwards1211:
@mark-wiemer okay, to test timing-sensitive cases, I used an IPC channel to send mock `chokidar` events and tell the mocha process when to pretend global setup is done.  I turned this mode on by passing a `MOCHA_TEST_WATCH` environment variable, but I'm not sure if you'd want this...if not how do you want me to accomplish the mocking?

lib/cli/watch-run.js
```diff
@@ -385,6 +386,12 @@ function createPathMatcher(allowed, ignored, basePath) {
   return matcher;
 }
 
+const TEST_MODE = Boolean(process.env.MOCHA_TEST_WATCH);
```
mark-wiemer: I don't think this `TEST_MODE` variable is a good idea, we need to ensure we're testing the same behavior that we're shipping. I'll research this one sometime this week, but I don't want to add any branches to our code that only execute in test mode.

mark-wiemer (dismissed):

lib/cli/watch-run.js
mark-wiemer: There are two processes here:
- Mocha global setup
- Chokidar watcher setup

My first thought to get around this is to write a test that adds artificial delays to one setup event so we can 99.999% ensure that the other completes first. I don't know how to implement that off the top of my head but the goal would be:

- Keep prod code the same, that is, `watch-run.js` shouldn't change to support testing
- Add delays in the test via spies or some other mock approach

jedwards1211: How would we add artificial delays to the events coming from Chokidar to test the case where it takes longer to get ready than global setup?

I don't think there's any great way.  We could add or touch files late but we can't control when chokidar emits the ready signal.

Maybe we could add a large number of files to make sure chokidar takes a long time to get ready but I don't know for sure that global setup wouldn't be slowed down by the chokidar scanning if we do that.

I never like when tests flake out once in a blue moon due to random variations in fs or network timing so I try to implement more guaranteed control in testing.

It's don't understand the doubt that simulating the timing of events happening would fail to test the behavior if the real production components create the same timing.

jedwards1211: I would want to at least get concrete information out of the process somehow to verify that the intended timing actually happened, so that we're not just *assuming* it happened how we intended.  Maybe I could turn on some debug messages to check that?

mark-wiemer: Yes, debug messages would be great, we use the `debug` package for that :)

> It's don't understand the doubt that simulating the timing of events happening would fail to test the behavior if the real production components create the same timing.

Assuming you meant `I don't understand ...`, my concern is that these `TEST_MODE` branches are being executed when tested, but the non-test-mode branches are being executed when users actually use the code. How do we know that the non-test-mode branches work if we're not executing them in tests?

jedwards1211: @mark-wiemer plenty of the tests do execute those other branches.  For the tests that do this mocking, the same production event handler code runs as in non-test mode; the only difference is we're triggering the events instead of creating Chokidar and crawling the fs.

> so we can 99.999% ensure that the other completes first

One test I wrote provides a 100% guarantee that the code for handling the ready event runs before the code that runs after global setup is done.  Another test provides 100% guarantee that they run in the opposite order.  Another test provides 100% guarantee that our code for a change event gets triggered before our code for a ready event.

If you really want weaker guarantees I can give you that, but be careful what you wish for.  I can't 100% guarantee that we'll always produce the desired event order with other techniques.

jedwards1211: Actually another case I still need to test is that adding a new file after global setup is done should trigger a rerun even if the watcher isn't ready (is still scanning the initial files).  This is an order of events that's hard to guarantee if I'm just touching files to trigger chokidar events.  I would probably have to create a large number of files so that chokidar is unlikely to finish scanning before I change one of them.  And what seems like a large enough number of files on my machine might prove insufficient on another machine.  Why leave it up to chance?

mark-wiemer: If the other tests do cover it, this is something I can consider. 100% is obviously ideal, but if we really have 99.99% then I won't complain. Curious for @JoshuaKGoldberg 's thoughts on this, but I know he's been busy.

Sorry for the delay, I've been super sick. I'm healthy now! I'm catching up on notifications now and will continue thinking about this one.

jedwards1211: Oh no, hope it was just a one time thing!

Sorry I was in a bit of a grouchy mood about the strategy.  I'm just thinking it will be a lot tricker to do testing without these mocks, I'll have to see how many thousands of dummy files I'd have to create to make Chokidar scan take long enough.  And probably a bunch of different subdirectories would take longer than the same number of files in one directory?

My biggest concern was having an environment variable left in the code that users could accidentally turn on.  If it were up to me I would configure build tools to strip testing hooks out when building the code to publish, but I assume y'all don't want something like that, so I'm trying to think if there's another way to trigger these test hooks or at least decide on an env var name that I can't imagine anyone using for anything outside this project.

jedwards1211: @mark-wiemer the subprocess helpers [merge together `stdout` and `stderr`](https://github.com/mochajs/mocha/blob/9a7053349589344236b20301de965030eaabfea9/test/integration/helpers.js#L349) into `RawResult.output` if piping `stderr` is requested.

This doesn't seem very helpful to me because if I `runMochaJSON` but also turn on  `mocha:cli:watch` `debug` output and pipe `stderr` to the `output` for test assertions, the mixed contents of `stderr` break the JSON parsing.  I ran into this awhile back and was confused at first.  I ended up not piping `stderr` in my new tests I've committed so far, but if we avoid the mocking I'm doing I'd want to inspect debug output to verify that the intended timing occurred during the tests.

I think putting the piped `stderr` into `RawResult.stderr` instead of mixing it into `RawResult.output` would be less confusing for everyone in the long run.  Alternatively I could add an option like `separateStderr` to work around this, but I think that would be less straightforward than just getting rid of the merging altogether.

But, I'd have to refactor all of the tests that assert some `stderr` output.  Would you be open to me making a separate PR to change that?

mark-wiemer: I am always in favor of PRs to clean things up! Shorter, cleaner, focused PRs get reviewed faster. The only catch is urgent bugfixes take top priority of course ;)

I think you've convinced me on the use of these test branches. Before I approve, I'll confirm that the non-test branches are covered by other tests. That will reassure me a lot more.

Thank you very much for the continued patience and detailed explanations :) and no worries about the grouchiness, I'd rather someone be honest than exude toxic positivity. And yes the illness was a onetime thing :)

mark-wiemer:
@jedwards1211 new tests are frequently timing out, do you think we can refactor them to complete in under 10 seconds?

jedwards1211:
@mark-wiemer huh, there may be a race condition causing the new tests to time out, I'll see what I can figure out.  Theoretically they should run especially quickly since they aren't designed to rely on any arbitrary delays.  (Many of the watch tests do use delays of a second or so, which makes running the entire watch test suite quite slow).

jedwards1211:
@mark-wiemer is there a way I can run CI with SSH to debug if the tests keep timing out?  It's not happening to me locally

jedwards1211:
oh I think I know what was causing the timeouts, they were probably sometimes missing the initial `{ listening: true }` event sent over IPC because of the default `sleepMs: 2000` before running the test-supplied `change` function.  I set `sleepMs: 0` in my tests now

mark-wiemer:
@jedwards1211 seems the tests are consistently failing on Windows, are you able to look into it?

jedwards1211:
Yup, I will tomorrow

jedwards1211:
@mark-wiemer is there a way I can run the Windows CI tests with SSH?  I can't reproduce locally on a Windows machine.  It seems to be some issue with IPC on Windows CI, but I'll need to either SSH into the tests or turn on some kind of debug logging to figure it out.

mark-wiemer:
I'm going to re-run the tests just to make sure this wasn't some weird fluke. And if they fail again, I'll try to repro from Windows on my side. I'll note that you can run the tests locally with the DEBUG flag on (ref #5538) , but if you can't repro, I understand those debug logs don't mean much.

Ref https://github.com/mochajs/mocha/actions/runs/19623994930 for test run

jedwards1211:
@mark-wiemer okay I temporarily turned on more debug logging in these tests, hopefully that will help me figure out what's going on in Windows CI

jedwards1211:
@mark-wiemer lol, the Windows issue is that `os.tmpdir()` seems to return a goofy [8.3 name](https://superuser.com/questions/1666765/how-to-resolve-tilde-completed-paths-in-windows-cmd-powershell):

```
2025-11-24T17:33:44.460Z mocha:cli:watch clean path: C:\Users\RUNNER~1\AppData\Local\Temp\mocha-JJqlNx\.git
2025-11-24T17:33:44.460Z mocha:cli:watch absolute path: C:\Users\RUNNER~1\AppData\Local\Temp\mocha-JJqlNx\.git
```

In the new watch tests I'm `realpath`-ing temp file paths for the mock watcher events because on macOS at least, `os.tmpdir()` contains symlinks.  Well, in Windows, that resolves `C:\Users\RUNNER~1` to `C:\Users\runneradmin`:

```
2025-11-24T17:33:44.469Z mocha:cli:watch got message {
  watcher: [
    'all',
    'add',
    'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\mocha-JJqlNx\\test.js'
  ]
}
```

I changed the `createTempDir` helper to `realpath` the directory it returns for all tests that use it instead, hopefully it should solve all these problems and not break anything else.

I guess I should also confirm that if Chokdiar is watching the contents of `C:\Users\RUNNER~1` on Windows, it also uses the 8.3 name in the file paths it emits, instead of the realpath.

Also -- I think all `require.cache` entries are realpaths, but I need to double check.  We're deleting modules from `require.cache` that match the watch path filters before rerunning, so if one is a realpath and the other isn't, it won't work as intended.

jedwards1211:
Okay this stuff about Windows short names is really aggravating.  Here's what I've discovered...

- `require('fs').realpathSync('C:\\Users\\RUNNER~1')` returns `C:\Users\RUNNER~1`
- `require('fs').realpathSync.native('C:\\Users\\RUNNER~1')` returns `C:\Users\runneradmin`
- `require.cache` does contain `'C:\\Users\\RUNNER~1\\...'` entries when the CWD is `C:\Users\RUNNER~1`
- Fun fact: `require('fs/promises').realpath` is actually based upon `require('fs').realpath.native`.  [I'm not the only one who found this confusing](https://github.com/nodejs/node/issues/37737)

I guess leaving the `fs/promises` `realpath`ed temp dirs for testing should work, but maybe I should also add tests that don't realpath it on windows...

jedwards1211:
@JoshuaKGoldberg can y'all retrigger a CI run?

mark-wiemer:
@jedwards1211 done, good luck :)

jedwards1211:
yaaaay, windows still 🤡

mark-wiemer:
Hi @jedwards1211, moved this to 12.x instead of 12.0, would still love to merge it but just clarifying that it isn't necessary for us to release Mocha 12. Are you able to continue working on this?

jedwards1211:
Yeah, if there's any way I could have permission to retrigger CI it would help me vet it done sooner

JoshuaKGoldberg:
One strategy I've done for situations like this is to fork the upstream repo and send a PR within that repo to itself. That way you don't have to wait on upstream things.

jedwards1211:
Oh I forgot I could do that, thanks for the tip!

mark-wiemer:
@jedwards1211 we also have our test workflow set to `workflow_dispatch` which allows you to manually trigger runs on any branch. For you that only applies to your fork, of course, but it's nice if you want to detect a flaky issue :) feel free to reach out in [the Discord](https://discord.gg/KeDn2uXhER) for detailed guidance!

mark-wiemer:
@jedwards1211  congrats on getting the tests passing!! I'll check this one out soon. @JoshuaKGoldberg  I think we can put this in 12.0, what do you think?

jedwards1211:
In CI within my fork I noticed a preexisting test flaking out once, only getting one of two expected test runs in the output.  I think this is the brittleness of sleep timing based approaches.  If you want I could rewrite more of the existing watch tests to wait for messages from the mocha process instead of sleeping to mitigate the flakiness

(A few of my tests mock chokidar to control timing of even its ready event, but I would leave existing tests using real chokidar.  I would just tell the mocha process to send messages when it receives chokidar events, finishes a test run, etc and make tests wait for those events instead of just sleeping.)

jedwards1211:
Btw, the final change that fixed the tests was checking if any of the `mtime`, `ctime` or `birthtime` of a file is after the test start time.  Before I was only checking the `mtime`, but that can actually be before the `birthtime` on Windows (`fs.cpSync` may preserve `mtime` but set a new `birthtime`).

mark-wiemer:
> I could rewrite more of the existing watch tests

You're welcome to open a second PR for this if you'd like, but let's keep this PR to its current scope :)

mark-wiemer:
Just FYI, might be a few days before I have enough energy to dedicate time to reviewing this, it's a very sensitive set of changes

jedwards1211:
Okay sounds good

lib/cli/run-helpers.js
```diff
@@ -106,7 +106,9 @@ exports.handleRequires = async (
       modpath = path.resolve(mod);
       debug("resolved required file %s to %s", mod, modpath);
     }
-    const requiredModule = await requireOrImport(modpath);
+    const requiredModule = forceRequire
```
mark-wiemer, line 115: Changing `handleRequires` feels like it can be split into its own small PR with tests, am I correct there? Curious to track details about what problem this solves (e.g. create an issue for this showing why the change is needed)

test/integration/helpers.js
```diff
@@ -542,7 +541,9 @@ function copyFixture(fixtureName, dest) {
  * @returns {Promise<CreateTempDirResult>} Temp dir path and cleanup function
  */
 const createTempDir = async () => {
-  const dirpath = await fsP.mkdtemp(path.join(os.tmpdir(), "mocha-"));
+  const dirpath = await fsP.realpath(
```
mark-wiemer, line 550: Can you elaborate why realpath is needed here, maybe with a code comment?

test/integration/helpers.js
```diff
@@ -564,6 +565,43 @@ function sleep(time) {
   });
 }
 
+/**
+ * Waits for an `EventEmitter` to emit the given `event`.
+ * @param {object} [opts]
+ * @param {EventFilter} [opts.filter] - predicate that will filter which event payloads will resolve the Promise
```
mark-wiemer: Can never be too clear!

```suggestion
 * @param {EventFilter} [opts.filter] - predicate that will filter which event payloads will resolve the Promise (defaults to `() => true`)
```

test/integration/helpers.js
```diff
@@ -564,6 +565,43 @@ function sleep(time) {
   });
 }
 
+/**
+ * Waits for an `EventEmitter` to emit the given `event`.
```
mark-wiemer:
```suggestion
 * Returns a promise that resolves when the `emitter` 
 * emits the `event` with properties matching the `filter`
```

test/integration/helpers.js
```diff
@@ -658,3 +697,10 @@ module.exports = {
  * @callback RemoveTempDirCallback
  * @returns {void}
  */
+
+/**
+ * Predicate for `gotEvent` that will filter which event payloads will resolve the Promise
+ * @callback EventFilter
+ * @param {...?} payload - the payload emitted for the event
```
mark-wiemer: I think this function only ever takes one arg, right?


```suggestion
 * @param {*} payload - the payload emitted for the event
```

lib/cli/watch-run.js
```diff
@@ -429,31 +455,89 @@ const createWatcher = (
   const matcher = createPathMatcher(allowed, ignored, basePath);
 
   // Chokidar has to watch the directory paths in case new files are created
-  const watcher = chokidar.watch(Array.from(allowed.dir.paths), {
-    ignoreInitial: true,
-    ignored: matcher.ignore,
-  });
+  let watcher, globalSetupBarrier;
```
mark-wiemer: maybe I'll find it soon, but I'm not sure how `watcher` gets initialized outside of `TEST_MODE`

mark-wiemer: haha there's an else block. my bad, consider this resolved. (using VS Code to write comments, can't find resolve button here)

lib/cli/watch-run.js
```diff
@@ -429,31 +455,89 @@ const createWatcher = (
...
+    if (TEST_MODE) sendMessage({ received: ["all", event, filePath] });
+    if (!matcher.allow(filePath)) {
+      debug("event doesn't match filter: %s, %s", event, filePath);
+      return;
     }
+    function doHandle(event, filePath, stat) {
+      if (exiting) return;
+      if (
+        event === "unlink" ||
+        (stat
```
mark-wiemer: when would we get into a situation where `stat` is not available but the watcher is still `ready`?

lib/cli/watch-run.js
```diff
@@ -429,31 +455,89 @@ const createWatcher = (
...
-    ignoreInitial: true,
-    ignored: matcher.ignore,
-  });
+  let watcher, globalSetupBarrier;
+  /* istanbul ignore next */
+  if (TEST_MODE) {
+    const EventEmitter = require("node:events");
+    watcher = new EventEmitter();
+    let resolveGlobalSetup;
+    globalSetupBarrier = new Promise(
```
mark-wiemer: can you comment the purpose of `globalSetupBarrier`, just to be clear that we're on the same page? What problem does it solve?

lib/cli/watch-run.js
```diff
@@ -613,14 +682,17 @@ const eraseLine = () => {
 
 /**
  * Blast all of the watched files out of `require.cache`
- * @param {FSWatcher} watcher - Chokidar FSWatcher
+ * @param {PathMatcher} matcher - `PathMatcher` instance
```
mark-wiemer:
```suggestion
 * @param {PathMatcher} matcher - only delete keys allowed by this matcher
```

lib/cli/watch-run.js
```diff
@@ -429,31 +455,89 @@ const createWatcher = (
...
+    ready = true;
+    /* istanbul ignore next */
+    if (TEST_MODE) sendMessage({ received: ["ready"] });
   });
 
-  watcher.on("all", (_event, filePath) => {
-    // only allow file paths that match the allowed patterns
-    if (matcher.allow(filePath)) {
-      rerunner.scheduleRun();
+  watcher.on("all", function handleEvent(event, filePath, stat) {
```
mark-wiemer: note to self: review scenario test coverage here (not just lines of code)

lib/cli/watch-run.js
```diff
@@ -509,27 +593,30 @@ const createWatcher = (
...
 
   // true if a file has changed during a test run
   let rerunScheduled = false;
 
-  const run = () => {
+  const run = async () => {
     try {
-      mocha = beforeRun ? beforeRun({ mocha, watcher }) || mocha : mocha;
+      if (TEST_MODE) sendMessage({ run: true });
+      mocha = (await beforeRun({ mocha, watcher })) || mocha;
```
mark-wiemer, line 598: note to self: how are we guaranteeing there is always a `beforeRun` callback now?

lib/cli/watch-run.js
```diff
@@ -509,27 +593,30 @@ const createWatcher = (
  *
  * @param {Mocha} mocha - Mocha instance
  * @param {FSWatcher} watcher - Chokidar `FSWatcher` instance
- * @param {Object} [opts] - Options!
- * @param {BeforeWatchRun} [opts.beforeRun] - Function to call before `mocha.run()`
+ * @param {PathMatcher} matcher - `PathMatcher` instance
+ * @param {Object} opts - Options!
+ * @param {BeforeWatchRun} opts.beforeRun - Function to call before `mocha.run()`
```
mark-wiemer:
```suggestion
 * @param {BeforeWatchRun} opts.beforeRun - Function to call before `mocha.run()`. Must be provided.
```

lib/cli/watch-run.js
mark-wiemer: answer: existing code already guaranteed this, private functions and whatnot can trace to only two possible `beforeRun` callbacks. consider this resolved.

.knip.jsonc
```diff
@@ -11,6 +11,7 @@
   "ignore": [
     "test/integration/fixtures/esm/type-module/test-that-imports-non-existing-module.fixture.js",
     "test/integration/fixtures/options/watch/test-with-dependency.fixture.js",
+    "test/integration/fixtures/options/watch/test-with-dependency-and-delay.fixture.js",
```
mark-wiemer:
```suggestion
    // import is only resolvable while running, not on a clean checkout
    "test/integration/fixtures/options/watch/test-with-dependency-and-delay.fixture.js",
```

package.json
```diff
@@ -83,6 +83,7 @@
     "test-node:only:globalBdd": "npm run -s test-node-run-only -- --ui bdd test/only/global/bdd.spec --no-parallel",
     "test-node:only:globalQunit": "npm run -s test-node-run-only -- --ui qunit test/only/global/qunit.spec --no-parallel",
     "test-node:only:globalTdd": "npm run -s test-node-run-only -- --ui tdd test/only/global/tdd.spec --no-parallel",
+    "test-node:only:integration:watch": "npm run -s test-node-run-only -- --no-parallel --timeout 10000 --slow 3750 \"test/integration/options/watch.spec.js\"",
```
mark-wiemer: can you clarify why this new command is needed?

test/integration/fixtures/options/watch/test-with-dependency-and-delay.fixture.js
```diff
@@ -0,0 +1,8 @@
+const dependency = require('./lib/dependency');
+
+it('checks dependency', async () => {
```
mark-wiemer:
```suggestion
// sleep and then check for existing depencency.fixture.js
it('checks dependency', async () => {
```

test/integration/options/watch.spec.js
```diff
@@ -475,6 +477,287 @@ describe("--watch", function () {
...
+              mochaProcess.send({ watcher: args }),
+            ]);
+
+          await gotMessage((msg) => msg.listening);
+          await change(mochaProcess, { gotMessage, sendWatcherEvent });
+        },
+      );
+    }
+
+    it("works when watcher is ready before global setup finishes", async function () {
```
mark-wiemer: Let's clean up these tests. I am thinking something like "before watcher is ready" becomes its own suite.

e.g. `handles files added before watcher is ready with timestamp after start`

becomes `before watcher is ready > files added > handles timestamp after start`

that should help us de-dupe boilerplate here as well.

Let me know if you have questions. Goal is to keep test names short, even if they're deeply nested. Then we can better review if all scenarios are covered.

lib/cli/watch-run.js
```diff
@@ -128,9 +133,17 @@ exports.watchRun = (mocha, { watchFiles, watchIgnore }, fileCollectParams) => {
...
+        // disable requireOrImport, since if the module gets loaded via ESM
+        // the cache won't get busted between runs.
+        // watch serial mode doesn't currently support ESM anyway.
+        forceRequire: true,
+      });
+
       // ... and now that we've gotten a new module, we need to use it again due
       // to `mocha.ui()` call
-      const newMocha = new Mocha(mocha.options);
+      const newMocha = new Mocha(Object.assign({}, mocha.options, plugins));
```
mark-wiemer: Should be the same, just rewritten in a way I find a bit more readable

```suggestion
      const newMocha = new Mocha({
        ...mocha.options,
        plugins
      });
```

mark-wiemer: Hmm, it looks like this second suggestion is actually the same behavior, but I'm not sure why we're spreading `...plugins`


```suggestion
      const newMocha = new Mocha({ ...mocha.options, ...plugins });
```

mark-wiemer: let's make sure we have a unit test that covers when plugins are present vs not present

mark-wiemer (changes requested):

mark-wiemer (changes requested):
Very close, I promise! If we clean up tests I'll have much more confidence in this :)

lib/cli/watch-run.js
mark-wiemer: @jedwards1211  do we still need tests for this?

lib/cli/run-helpers.js
mark-wiemer: haha with the scale of this PR, breaking out these few lines won't change much. a brief comment to cover why `forceRequire` is necessary would be appreciated here :)

lib/cli/watch-run.js
mark-wiemer: I'm thinking it is meant to prevent the test watcher from firing too early?

lib/cli/watch-run.js
mark-wiemer: resolving this one in favor of the general "test cleanup" thread

mark-wiemer:
Other reviewers are welcome to check my architecture notes: https://gist.github.com/mark-wiemer/e98f4aa9e49fd749519e6789d6ae564d and search `PR 5409:` for changes

lib/cli/run-helpers.js
jedwards1211: yes I will comment, this was to make sure changed root hooks modules actually get re-`required` instead of imported, because if imported the ESM wrapper breaks our cache busting via `delete require.cache[...]`.

And I judged it okay to do this because ES modules aren't officially supported in watch mode anyway (serial mode at least, parallel mode seems to work)

test/integration/helpers.js
jedwards1211: actually, since writing this I discovered that Node has builtin `events.once()` for this purpose since well before v20, I should refactor

jedwards1211: wait...it doesn't support a `filter` parameter though

test/integration/helpers.js
jedwards1211: oh right, I was thinking of what to do in case there are multiple arguments to `emit(...)` but decided not to handle it

test/integration/helpers.js
jedwards1211: but, I could still use `events.once()` in the tests anyway

.knip.jsonc
jedwards1211: I'm confused, how is the file not there on a clean checkout?

jedwards1211: whatever is the case for the existing `test-with-dependency.fixture.js` is also the case for this new `test-with-dependency-and-delay.fixture.js` I added

package.json
jedwards1211: Oh I added it just for working on specific tests with `.only` during development.  I did notice some of the `only` scripts seem to be run in CI, what is the purpose of running them there?

test/integration/options/watch.spec.js
jedwards1211: Personally, I hate putting a bunch of tests in one file in my own projects, I find it much easier to work with if each test is in its own file.

jedwards1211: Btw, it's not very ergonomic to me that putting a bunch of tests in one file is the easiest way to accomplish shared setup and teardown boilerplate with Mocha.

`vitest`'s `setupFiles` option provides a better means of applying shared setup and teardown to multiple test files, and using its `projects` setting makes it possible to do that without having the `setupFiles` be global to absolutely all tests

In my own projects with Mocha I have to use separate mocha configs or launch scripts for unit, integration, e2e etc since each category has different shared setup and teardown

lib/cli/watch-run.js
jedwards1211: maybe I can pick a better name, this is for testing, it allows some tests to control when it's pretending global setup files finish.

If you prefer, instead of this I guess I could make those tests put in a real global setup file that does something like

```
const { once } = require('events');
const { before } = require('mocha');

before(async function () {
  let message;
  do {
    message = await once(process, 'message');
  } while (!message?.resolveGlobalSetup);
})

File: lib/cli/watch-run.js (hunk 9 above)
Comment (jedwards1211) [reply lib/cli/watch-run.js]: Before the watcher is ready, we need to `stat` files so we can see whether they were created after the test process was started, rather than just initial `add` events for preexisting files.

After the watcher is ready, we don't need to `stat` any more, we can just assume any `add` or `change` happened after startup and should trigger a rerun.

I could use `chokidar`'s `alwaysStat: true` option instead if you prefer, I just didn't because that would be unnecessary `stat`ing after the watcher is ready

File: lib/cli/watch-run.js (hunk 19 above)
Comment (jedwards1211) [reply lib/cli/watch-run.js]: we do, the plugins are present in my tests for changing root hooks files, and (well, I assume) not present in tests that don't have root hooks.

`plugins` is an object containing Mocha constructor options like `rootHooks`, rather than a Mocha constructor option itself

File: test/integration/helpers.js (hunk 4 above)
Comment (jedwards1211) [reply test/integration/helpers.js:550]: yeah I'll add a comment soon, without this it's more work for some of the new watch tests to get the right paths in mock file watcher events

**Comment by mark-wiemer** [comment]
Sorry for the delay, I am still watching this one :) should have some time this weekend :)

**Comment by jedwards1211** [comment]
Okay same!  I'm still trying to decide how to recategorize the tests you asked for, it's not quite so simple, since there are bunch of possible orderings of ready signal, file changes, and global setup/test runs starting or finishing

**Comment by jedwards1211** [comment]
It would simplify the code and tests if I make chokidar always stat, ignore the ready event, and just always compare the file timestamps to decide whether to rerun.  So I could do that if you want, it's probably harmless to just stat all changes

**Comment by mark-wiemer** [comment]
> It would simplify the code and tests if I make chokidar always stat, ignore the ready event, and just always compare the file timestamps to decide whether to rerun. So I could do that if you want, it's probably harmless to just stat all changes

If you think it'd be better, you're more than welcome to do it :) I am not too too concerned about perf at this point, obv this PR solves a lot of the perf problems anyway. Worst case, we refactor later :)

**Comment by jedwards1211** [comment]
@mark-wiemer @JoshuaKGoldberg I'm not sure how things were managing to behave, but I wanted to point out some cruft I noticed and fixed.

I was occasionally seeing an error get logged without failing some watch tests from `mocha.unloadFiles()` saying the mocha instance had already been disposed.

This call to `mocha.unloadFiles()` is unnecessary because `beforeRun` calls `mocha.dispose()` a moment later, which does the same thing.  The call to `mocha.unloadFiles()` is ancient, `mocha.dispose()` was added later but someone seemingly forgot to remove `mocha.unloadFiles()`.

I also discovered that if an error is thrown during `mocha.run()` (for instance if a user test has a syntax error), the mocha instance is permanently left in `RUNNING` state.  And `dispose()` throws if the mocha instance is still in `RUNNING` state.

So I added a `try`/`catch` to `mocha.run()` to set the state back to `IDLE` before throwing the error, and removed the obsolete `mocha.unloadFiles()` call, and it seems to be behaving now.

**Comment by jedwards1211** [comment]
@mark-wiemer I am still suffering from windows ci failures after refactoring the tests and I am extremely fed up with it, so please accept the test structure I have now, I don't want to have to go through another round of this.

**Comment by jedwards1211** [comment]
Ugh part of those occasional errors I was seeing had to do with a race condition when file events happen in rapid succession while idle due to the fact I had to make `beforeRun` async.

But regardless, an issue with the current design is that, while idle, it starts a new run immediately on the next file event, even if there are going to be more file events on the next tick that abort that run.  I added a 10ms debounce to `rerun` to address this.

**Comment by jedwards1211** [comment]
Okay I finally got everything ironed out, it was painful but rewriting the tests did uncover some old and new issues.

Some things that would make it less painful:
- Typescript (I think this is planned right?)
- replacing callbacks with promises in more places like `Mocha.run`
- moving each test into a separate file, defining a vscode hotkey to run the current test file so we don't have to spam `.only` everywhere during development
- logging test titles as they pass/fail instead of at the very end so that it's possible to tell where debug output associated with a specific test is in CI runs
- defining an explicit prettier config in the project (so that we can keep the prettier-vscode **require config** option on - ask me if you don't understand the point of this)
- and/or setup `lint-staged` to format and lint on commit
- fixing stack traces - i think there's some kind of source mapping issue, the line numbers were all off in errors thrown during testing
- some way to run specific tests in CI without making messy commits to the workflows?  Maybe a magic environment variable and workflow conditions?

If you're okay with any of these there's some low hanging fruit I could quickly make PRs for.

**Comment by mark-wiemer** [comment]
Thank you very much! I love the feedback too. Please search for and open issues for anything you have energy to cover :) I'm on mobile right now but I'll be able to check this PR later this weekend.

File: lib/cli/watch-run.js (hunk 9 above)
Comment (mark-wiemer) [reply lib/cli/watch-run.js]: sounds good, no change needed here :)

File: lib/cli/watch-run.js (hunk 10 above)
Comment (mark-wiemer) [reply lib/cli/watch-run.js]: I think it's good as-is :)

File: lib/cli/watch-run.js [hunk 20]
```diff
@@ -429,29 +454,70 @@ const createWatcher = (
...
+    watcher = chokidar.watch(Array.from(allowed.dir.paths), {
+      ignoreInitial: true,
+      ignored: matcher.ignore,
+      alwaysStat: true,
+    });
+  }
 
-  const rerunner = createRerunner(mocha, watcher, {
-    beforeRun,
+  // if files change during global setup, we need to blast the cache before the first
```
mark-wiemer, line 474: note to self: make sure we have a test for this scenario

lib/mocha.js
```diff
@@ -974,79 +974,88 @@ Object.defineProperty(Mocha.prototype, "version", {
...
+    // no "catch" here is intentional. errors coming out of
+    // Runner#run are considered uncaught/unhandled and caught
+    // by the `process` event listeners.
+    // also: returning anything other than `runner` would be a breaking
+    // change
+    runAsync(runner).then(done);
+
+    return runner;
+  } catch (error) {
+    this._state = mochaStates.INIT;
```
mark-wiemer, line 1056: note to self: make sure we have a new test for this error case

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,463 @@ describe("--watch", function () {
...
+     * @param {ChokidarMockChangeCallback} change - function to perform test-specific actions
+     * once the `mocha` process has started.
+     */
+    async function runMochaWatchWithChokidarMock(args, opts, change) {
+      let runScheduled = false;
+      let running = false;
+      const result = await runMochaWatchJSONAsync(
+        args,
+        {
+          stdio: ["pipe", "pipe", "inherit", "ipc"],
```
mark-wiemer, line 523:
```suggestion
          // stdin: pipe, stdout: pipe, stderr: inherit, ipc: create IPC channel for messages
          stdio: ["pipe", "pipe", "inherit", "ipc"],
```

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+        args,
+        {
+          // stdin: pipe, stdout: pipe, stderr: inherit, ipc: create IPC channel for messages
+          stdio: ["pipe", "pipe", "inherit", "ipc"],
+          env: {
+            ...process.env,
+            __MOCHA_WATCH_MOCK_CHOKIDAR: "1",
+            DEBUG: "mocha:cli:watch",
+          },
+          sleepMs: 0,
```
mark-wiemer, line 529: should we not spread `opts` first? Or do we actively want callers to override `stdio`, `env`, and `sleepMs` if set?

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+          ...opts,
+        },
+        async (mochaProcess) => {
+          mochaProcess.on("message", (msg) => {
+            if (!msg) return;
+            if (msg.run) {
```
mark-wiemer, line 535:
```suggestion
            if (msg.runStarted) {
```

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+              running = true;
+              runScheduled = false;
+            }
+            if (msg.runFinished) running = false;
+            if (msg.scheduleRun) runScheduled = true;
```
mark-wiemer, line 540: This PR introduces this msg concept, I think we should rename following this convention:

- past tense verb first
- object comes after

so:

- startedRun
- finishedRun
- scheduledRun

lib/cli/watch-run.js
mark-wiemer: we have one :)

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+          // this version of the module gets loaded by global setup,
+          // but it should get replaced before the test is run
+          replaceFileContents(
+            dependency,
+            "module.exports.testShouldFail = false",
+            "module.exports.testShouldFail = true",
+          );
+
+          const dependency2 = path.join(tempDir, "lib", "dependency2.js");
+          if (event !== "add")
```
mark-wiemer, line 784: can you explain the role of this if block?

- why exclude add events?
- why add dependency2.js if it's not referenced?

mark-wiemer: OK, I see that it is referenced later, but still not sure why we exclude add events

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+
+          mochaProcess.send({ resolveTest: true });
+          await gotMessage((msg) => msg.runFinished);
+
+          copyFixture("options/watch/dependency", dependency3);
+          await sendWatcherEvent.change(dependency, -600);
+          await sendWatcherEvent.add(dependency3, -600);
+        },
+      );
+      expect(results, "to have length", 1);
```
mark-wiemer, line 709: if we're repeating the same batch of `expect`s in each test, we can just add that to them to the new `runMochaWatchWithChokidarMock` func, and we can rename it `testWatchWithMockChokidar`

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+          "--require",
+          globalSetupFile,
+        ],
+        { cwd: tempDir },
+        async (mochaProcess, { gotMessage, sendWatcherEvent }) => {
+          await sendWatcherEvent.add(testFile, -1000);
+          await sendWatcherEvent.add(dependency, -1000);
+          await sendWatcherEvent.change(dependency, -500);
+          // this should not get picked up by the test file
+          // it should already have the old version in memory
```
mark-wiemer, line 742: please clarify which version is "old" (`= true` or `= false`) for exteme clarity

mark-wiemer:
```suggestion
          // it should already have the `= true` version in memory
```

mark-wiemer:
```suggestion
          // it should already have the `= false` version in memory
```

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+        async (mochaProcess, { gotMessage, sendWatcherEvent }) => {
+          await sendWatcherEvent.add(testFile, -1000);
+          await sendWatcherEvent.add(dependency, -1000);
+          await sendWatcherEvent.change(dependency, -1000);
+
+          mochaProcess.send({ resolveGlobalSetup: true });
+          await gotMessage((msg) => msg.testStarted);
+
+          copyFixture("options/watch/dependency", dependency2);
+          await sendWatcherEvent.change(dependency, -800);
```
mark-wiemer, line 663: this test is handling two scenarios: when "old" events are received during a test run, and when "old" events are received after a test run. I'll clarify that an old event is one that happened before the test run was started.

I'm hoping we can break this test up into two smaller pieces to make it easier to identify potential issues when this outer test fails*

test/integration/options/watch.spec.js
mark-wiemer: I see that we want to send `event` for this dep2 file in every case (line 732, `await sendWatcherEvent[event](dependency2, 1000)`, but I'm still not quite sure why we need to exclude this `copyFixture` step when `event === "add"`)

mark-wiemer: is it because the file must not exist before the run starts? if so, wouldn't the check in line 723 be sufficient? I'm worried that the current test for add may be a false negative. Let me walk through this:

1. Add test file and dep 1. If unchanged, dep 1 will force tests to fail.
1. Update dep 1 to allow tests to pass
1. send an event of the relevant type (add, change, unlink) to the Mocha subprocess
   - If Mocha subprocess blasted cache, then it'll get the updated version of dep 1, and inner test will pass
   - If Mocha subprocess fails to blast cache, then it'll use the old version of dep 1, and inner test will fail
1. Expect inner test to have passed, etc.

OK, not a false negative. Definitely hard to reason with but I don't have any better ideas off the top of my head.

Overall, I would love for our test logic to be simpler. Who tests the tests, you know? I'll have to think about all this.

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+              if (event !== "add") {
+                await sendWatcherEvent.add(dependency2, 0);
+                await sendWatcherEvent.add(dependency3, 0);
+              }
+
+              console.error(0);
+              mochaProcess.send({ resolveGlobalSetup: true });
+              console.error(1);
+              await gotMessage((msg) => msg.runFinished);
+              console.error(2);
```
mark-wiemer: I appreciate the associated the commit message that added these err logs, even as a Microsoft employee 🫠

```suggestion
              mochaProcess.send({ resolveGlobalSetup: true });
              await gotMessage((msg) => msg.runFinished);
```

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+              if (event === "add") {
+                copyFixture("options/watch/dependency", dependency2);
+                copyFixture("options/watch/dependency", dependency3);
+              } else if (event === "unlink") {
+                await Promise.all([
+                  fsPromises.unlink(dependency2),
+                  fsPromises.unlink(dependency3),
+                ]);
+              }
+              console.error(3);
```
mark-wiemer:
```suggestion
```

test/integration/options/watch.spec.js
```diff
@@ -475,6 +478,464 @@ describe("--watch", function () {
...
+                  console.error(5),
+                ),
+                gotMessage((msg) => msg.runFinished).then(() =>
+                  console.error(6),
+                ),
+              ]);
+            },
+          );
+
+          console.error(require("util").inspect(results, { depth: 10 }));
```
mark-wiemer:
```suggestion
```

test/integration/options/watch.spec.js
mark-wiemer: Ah, another ambiguity is "the test file", which could be this `watch.spec.js` but seems to really refer to the inner test runner process. Because we're telling the inner process that the file changed 500 ms before that inner process started, the inner process will think "oh, that doesn't matter, I don't need to learn about that change." I'll update the code comments in this file accordingly.

mark-wiemer: OK, I can't really do that, because I see a bigger issue here: shouldn't the actual change to the `dependency` file be done after the `sendWatcherEvent.add(dependency, ...)` call and before the `sendWatcherEvent.add(dependency, ...)` call?

test/integration/options/watch.spec.js
mark-wiemer: for now, I'm OK to merge this. It took some getting used to but it's not so so bad, all things considered. Testing a test runner is hard.

lib/mocha.js
mark-wiemer: I'm not seeing one 😬

mark-wiemer:
@jedwards1211 finished my review, super duper cool to see all these tests passing, just a bit more cleanup. I'll probably do it myself later today or tomorrow if you're unavailable, just FYI

@mochajs/maintenance-crew ready for review by a second maintainer, this one is a beast but a valuable beast!

test/integration/options/watch.spec.js
jedwards1211: I think you're overthinking this, my intention for `runMochaWatchWithChokidarMock` is to give these tests and potentially new tests in the future an open-ended API to verify whatever is needed, if we assume all tests are going to make these assertions it's less flexible in the long run

test/integration/options/watch.spec.js
jedwards1211: yes I had in mind letting callers override things if necessary in the future.  I would need to make sure to keep the `ipc` channel though.
On that note I should be spreading `...opts?.env` into `env`

test/integration/options/watch.spec.js
jedwards1211: `scheduleRun` is dispatched when `scheduleRun` is called, whether or not a run is already scheduled.  If we rename it to `scheduledRun`, then you might want the semantics to change to be when a new run is actually scheduled, and that change might break existing tests, I'm not 100% sure.

lib/mocha.js
jedwards1211: The existing tests for syntax errors in user test files should be reaching this catch block, and if you remove the catch block those tests should fail now that I removed the superfluous `mocha.unloadFiles()` from `beforeRun`, so I consider this tested

test/integration/options/watch.spec.js
jedwards1211: oh and I didn't repeat the same exact assertions in each of these new tests anyway, some have just one run, some have multiple

test/integration/options/watch.spec.js
jedwards1211: I mean, I guess we don't necessarily have to test that the cache doesn't get blasted in this case; if we simply blasted before every run the behavior would also be correct.
But, I did intentionally avoided blasting before the very first run if we don't need to, in case it's a bit slow to check every entry in `require.cache`.
Since I have the logic to decide whether it's necessary to blast before the very first run, it's necessary to cover both branches.

In this case we start with `testShouldFail = false`, mocha boots up and requires that file, then we change it to `testShouldFail = true`.
But the mock event timestamps are before mocha started watching so we should decide *not* to blast the cache, and it should run with `testShouldFail = false`.

jedwards1211: I wasn't thinking clearly about the `change` event though, a `change` event timestamped before mocha started watching isn't realistic, only an `add` event timestamped before mocha started watching could reasonably happen

test/integration/options/watch.spec.js
jedwards1211: Also i can get rid of the `DEBUG` setting here now too

test/integration/options/watch.spec.js
jedwards1211: okay I'll split it up

jedwards1211:
@mark-wiemer sigh why can't you be satisfied with my test changes already, my attempts to change things are causing failures on Windows CI *AGAIN*, I am so sick of this 😭

jedwards1211:
@mark-weimer well, actually it was a preexisting test flaking out in CI on my fork, probably due to race conditions from waiting arbitrary amounts of time in the watch tests.
Anticipating this kind of flakiness is why it was very important to me to set up these mocks that can fully control timing, at least for some of the tests.

test/integration/options/watch.spec.js
jedwards1211: also the existing events in `lib` have a topic + infinitive convention so maybe we should follow that...

```
    /**
     * Emitted when {@link Suite} execution begins
     */
    EVENT_SUITE_BEGIN: "suite",
    /**
     * Emitted when {@link Suite} execution ends
     */
    EVENT_SUITE_END: "suite end",
    /**
     * Emitted when {@link Test} execution begins
     */
    EVENT_TEST_BEGIN: "test",
    /**
     * Emitted when {@link Test} execution ends
     */
    EVENT_TEST_END: "test end",
```

I think it's pretty common to use noun + verb so that things sort and group visually by topic

jedwards1211:
sorry I shouldn't complain, you did say you might make the changes yourself, I wish there were just faster ways to iterate to get this done :/

test/integration/options/watch.spec.js
mark-wiemer: looks good, resolving, this was a bit of a nit anyway :)

test/integration/options/watch.spec.js
mark-wiemer: this was a bit of a nit, we can change these names later, resolving

mark-wiemer (approved):
This is so so cool!! Mocha is gonna be fast and awesome thanks to your dedicated work here :) I apologize for starting to nitpick 5 months into the review but we're finally over the line :) only thing left is for another maintainer to give it a look, given how critical this change is

mark-wiemer:
Only test failure is a known unrelated issue:

```
  365 passing (4m)
  2 pending
  1 failing

  1) --parallel
       pool shutdown
         during operation with --bail
           should not leave orphaned processes around:
     
expected Promise when fulfilled to have items satisfying false
  expected [ false, true, true ] to have items satisfying false

  [
    false,
    true, // should equal false
    true // should equal false
  ]

      at process.processImmediate (node:internal/timers:505:21)
  From previous event:
      at Context.<anonymous> (test\integration\options\parallel.spec.js:545:16)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      set UNEXPECTED_FULL_TRACE=true to see the full stack trace
```

jedwards1211:
Yeah I've been using this change in my biggest project and it's been great!

Sorry for my impatience :/

Btw I've been thinking we could use a similar strategy with ipc to fix the flakiness in the other watch tests and make them faster (use real chokidar, but report chokidar events and test runs up to the test process over ipc)

mark-wiemer:
I'm always open to improvements!

jedwards1211:
okay I'll outline more of a proposal in an issue soon

mark-wiemer:
https://github.com/mochajs/mocha/actions/runs/22022724150/job/63634561899?pr=5409 technically shows one new failure, but I'm not concerned. As always, ref #5361.

jedwards1211:
huh okay, I just merged main again, if it still fails I'll see if I can figure out why

mark-wiemer:
All checks passed :) I've reached out to @mochajs/maintenance-crew to prioritize this one, thanks for keeping an eye on it!

JoshuaKGoldberg:
> On another note, @mochajs/maintenance-crew I'm thinking of marking this as "semver-major" given the complexity impacted and possibility of bugs. 

I agree, this is a lot of really intricate stuff. Mocha 11 is very stable. No matter how thoroughly this PR is tested, the nature of its deep changes is that some folks will be broken by it.

> Since we want Mocha 12 to have minimal changes we'll probably hold this one for Mocha 13. (https://github.com/mochajs/mocha/issues/5357)

How about we wait until Mocha 12.1 or 12.2? Given that this is technically only bugfixes (albeit with real behavioral changes), it'd be nice to land it sooner. Giving folks a 12.0.x line with the first round of 12 stability, then following up with a subsequent minor release, means we'd get it in front of folks sooner.

lib/cli/watch-run.js
```diff
@@ -385,6 +397,19 @@ function createPathMatcher(allowed, ignored, basePath) {
...
+const TEST_MODE = Boolean(process.env.__MOCHA_WATCH_MOCK_CHOKIDAR);
+
+// grab Date.now before anything like sinon.createFakeTimers can monkeypatch it
+// eslint-disable-next-line no-restricted-globals
+const dateNow = Date.now;
+if (TEST_MODE) debug("in test mode");
+
+const sendMessage = (msg) => {
+  if (!TEST_MODE) return;
+  process.send(msg);
```
JoshuaKGoldberg, line 409: [Question] How bad would it be if Mocha always sent these messages? As in, removing `TEST_MODE` altogether?

jedwards1211: Hypothetically, a user could be spawning Mocha with IPC and sending their own messages for some reason, so I wouldn't want this to unexpectedly interfere with that.

mark-wiemer: Resolving as approved and thumbs-upped

mark-wiemer:
@jedwards1211 we finally did it :) thank you so much for the initiative and patience here, this is great work that will be a major celebration in the Mocha 12 announcement :)

mark-wiemer:
Oh no, @JoshuaKGoldberg I did kinda merge this one, would you like me to revert? 😬 I checked for a `status: blocked` tag or similar and didn't see one, we should align on how to mark PRs as "approved but not ready to merge for X reason," my bad!

mark-wiemer:
I'm going to revert to reduce the risk of merge conflicts overall. We will get this in for 12.x pretty early, I will handle merge conflicts, assuming I'm not super sick again. Feels bad to revert but it's for the greater good!

JoshuaKGoldberg:
Oop! Yeah a revert makes sense. Agreed.

jedwards1211:
Yay!  Thank y'all for being open to this :)

I've been pretty swamped with work lately, so I haven't had time to work on the CJS/ESM errors yet.  Hopefully soon...

PR #5413, opened 2025-08-09
Fixes #5411.

### LINKED ISSUE #5411: 🐛 Bug: Unhelpful error message when a default import is made from a module without a default export
### Bug Report Checklist

### Expected

When a default import is made from a module without a default export, I expect an error message such as:

```
SyntaxError: The requested module './index.js' does not provide an export named 'default'
```

This is also what Mocha until version `11.7.0` gives.

### Actual

In Mocha version `11.7.1`, I get the following error message instead:

```
Exception during run: Error [ERR_INTERNAL_ASSERTION]: Unexpected status of a module that is imported again after being required. Status = 0
This is caused by either a bug in Node.js or incorrect usage of Node.js internals.
Please open an issue with this stack trace at https://github.com/nodejs/node/issues
```

@papandreou bisected the commits and found out that the issue was introduced in https://github.com/mochajs/mocha/pull/5384.

https://github.com/martinslota/mocha-default-import-error

### Versions

11.7.1

### Additional Info

_No response_

martinslota:
Fail with an informative error message on a file with a broken default import.

## PR Checklist

- [x] Addresses an existing open issue: fixes #5411 

## Overview

Fail with an informative error message on a file with a broken default import.


mark-wiemer:
Nice :) cursory review looks good, I'll dive deep into this tomorrow as I'm still a new maintainer

lib/nodejs/esm-utils.js
```diff
@@ -93,11 +94,20 @@ const requireModule = async (file, esmDecorator) => {
   try {
     return require(file);
   } catch (err) {
-    // Import if require fails.
-    return dealWithExports(await formattedImport(file, esmDecorator));
+    debug('requireModule caught err: %O', err.message);
+    if (
+      err instanceof SyntaxError &&
+      err.message.includes("does not provide an export named 'default'")
```
JoshuaKGoldberg: [Question] I'm a _little_ suspicious of this because I know some projects do fancy/weird things around [`esModuleInterop`](https://www.typescriptlang.org/tsconfig/#esModuleInterop) and similar massaging of default exports/imports. Maybe a project fails to import with `import()` but succeeds with `require()`. Is that a case we should test?

Tentatively I think we should make sure this works in that case, but I'm out right now and can't dive in 😅. Does anybody else have thoughts?

martinslota: > Maybe a project fails to import with import() but succeeds with require().

The code here attempts `require` first, so it would succeed.

But maybe you meant the other way around, a case when `require` fails but `import` would succeed? I think a challenge here is that in the specific case of a missing default export, Node.js seems to get into a funky state when the initial `require` fails, making the subsequent import [fail with an internal assertion](https://github.com/nodejs/node/blob/abccbb438b9175036e8b6bd1cc545a1ba5cda7e9/lib/internal/modules/esm/module_job.js#L463-L486) that shows `Status = 0`, which I guess might mean that the module ended up in [the uninstantiated state](https://github.com/nodejs/node/blob/abccbb438b9175036e8b6bd1cc545a1ba5cda7e9/src/module_wrap.cc#L1438).

But I do think you have a good point, and I certainly do not have a full picture of all the various cases. I attempted to address your concern by [adjusting the change](https://github.com/mochajs/mocha/pull/5413/commits/f689dbc790d597515ba19447e414556e7879b00a) such that the `import()` is always attempted and, if it does fail, and the import error is `ERR_INTERNAL_ASSERTION`, the original error from the `require` call is thrown instead.

> Is that a case we should test?

I'd be happy to add another test around some TypeScript code transpiled with `esModuleInterop` being set to both `false` and `true` - but I'm not entirely sure which cases are really worth testing. Would it be more stuff around default imports, or just some sanity tests showing that `requireModule` successfully loads the files?

mark-wiemer: @JoshuaKGoldberg any updates here?

mark-wiemer:
@martinslota can you resolve merge conflicts?

lib/nodejs/esm-utils.js
JoshuaKGoldberg: Ah thanks - I think you got what I was trying to convey 😅. Great!

PR #5429, opened 2025-08-16
fix: load mjs files correctly

### LINKED ISSUE #5425: 🐛 Bug: .mjs test files can be loaded with `require()` which breaks custom module loaders
### Bug Report Checklist

### Expected

The recent changes in mocha@11.7.x for deciding whether to load a mocha test file as CommonJS or ESM should *not* use CommonJS for `.mjs` files. Specifically I mean for the code path that runs `requireModule()` in lib/nodejs/esm-utils.js.

### Actual

"esm-utils.js" includes:

```js
if (process.features.require_module) {
  exports.requireOrImport = requireModule;
```

which leads to `require(file)` being used to load a test file if running with a version of node that supports the `require(esm)` feature.  However:

1. Versions of Node.js v20 starting with v20.19.0 [include `require(esm)` support](https://nodejs.org/en/blog/release/v20.19.0#requireesm-is-now-enabled-by-default) (and hence pass the `if (process.features.require_module)` condition), but do *not* include https://github.com/nodejs/node/pull/55698 (and possibly other Node.js module hook handling changes, I'm not perfectly clear on which Node.js commits are relevant). The result is that a custom module loader is not used for that loaded file, or the files it later imports.
2. All versions of Node.js that support `require(esm)` *and* have the additional module hook handling changes (e.g. Node.js ~22.15.0, some minor version of 24.x),  **will** run custom module loader hooks for the `import`s in a file loaded by `require(some-esm-file)`, but **will NOT** run the custom module loader hooks for the `some-esm-file` itself.

**tl;dr: Even in the latest versions of Node.js, doing `require("./some-esm-file.mjs")` has different behaviour with custom module loaders compared to doing `await import("./some-esm-file.mjs")`**

https://github.com/trentm/repro-mocha-11-mjs-import-hooks-issue

This is a small example with a test file that (a) uses `.mjs` to indicate to Node that it should be loaded as an ES module, and (b) depends on a loader hook (`--experimental-loader=./hook.mjs`) to be run when the test file is loaded for the tests to pass.

```bash
NODE_NO_WARNINGS=1 node --experimental-loader=./hook.mjs ./node_modules/.bin/mocha 'test/**/*.test.mjs'
```

The README shows that this test passes when using mocha@11.6.0, but fails with mocha@11.7.x.  You must run with Node.js v20.19.x or v22.14.0 to see both test cases fail.  Running with v22.15.0 or later will fail one of the two test cases.

### Versions

```bash
% npx mocha --version
11.7.1
% node --version
v20.19.1
```

### Additional Info

A believe a good fix would be to make this change:

```diff
diff --git a/lib/nodejs/esm-utils.js b/lib/nodejs/esm-utils.js
index 3cea8af2..773f0829 100644
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -90,6 +90,10 @@ const tryImportAndRequire = async (file, esmDecorator) => {
 // and `require.cache` effective, while allowing us to load ESM modules
 // and CJS modules in the same way.
 const requireModule = async (file, esmDecorator) => {
+  if (path.extname(file) === '.mjs') {
+    return formattedImport(file, esmDecorator);
+  }
+
   try {
     return require(file);
   } catch (err) {
```

This does the same ".mjs" guard as was already added for the `tryImportAndRequire` function in the same file. The idea is that the ".mjs" extension is a clear sign that the file should be loaded with `import()` and *not* with `require()`.

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5425

## Overview

mjs files were sometimes loaded like cjs files despite their extension. In some versions of Node (including Node 20.19.4) this causes breaking issues caught by our tests. A separate PR will bump Node versions in CI.

### DIFF
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -1,5 +1,6 @@
 const path = require('node:path');
 const url = require('node:url');
+const debug = require('debug')('mocha:esm-utils');
 
 const forward = x => x;
 
@@ -99,8 +100,12 @@ const requireModule = async (file, esmDecorator) => {
     // Import if require fails.
     return dealWithExports(await formattedImport(file, esmDecorator));
   }
-}
+};
 
+// We only assign this `requireOrImport` function once based on Node version
+// We check for file extensions in `requireModule` and `tryImportAndRequire`
+debug('Node version: %s', process.versions.node);
+debug('process.features.require_module: %s', process.features.require_module);
 if (process.features.require_module) {
   exports.requireOrImport = requireModule;
 } else {
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -90,6 +90,9 @@ const tryImportAndRequire = async (file, esmDecorator) => {
 // and `require.cache` effective, while allowing us to load ESM modules
 // and CJS modules in the same way.
 const requireModule = async (file, esmDecorator) => {
+  if (path.extname(file) === '.mjs') {
+    return formattedImport(file, esmDecorator);
+  }
   try {
     return require(file);
   } catch (err) {


trentm (approved):
FWIW, this LGTM. :)

lib/nodejs/esm-utils.js
```diff
@@ -90,14 +91,21 @@ const tryImportAndRequire = async (file, esmDecorator) => {
...
     // Import if require fails.
     return dealWithExports(await formattedImport(file, esmDecorator));
   }
-}
+};
 
+// We only assign this `requireOrImport` function once based on Node version
+// We check for file extensions in `requireModule` and `tryImportAndRequire`
+debug('Node version: %s', process.versions.node);
+debug('process.features.require_module: %s', process.features.require_module);
```
JoshuaKGoldberg: 🤔 I'm not sure this is valuable. Running debug lines in the start of files as a side effect means file import order will change runtime log behavior. Plus anybody running Node.js in CI, etc. should already have logs for these things.

mark-wiemer: I'm OK with runtime log behavior being a bit inconsistent as long as the logs are present. Right now it's very hard to tell whether a Node version has `require_module` feature without looking it up somewhere online, I feel it's a helpful debug log for issues like this one. I know Node version is a bit easier to find, I think it's convenient to log but if we are worried about being too verbose in debug logs I can remove it

mark-wiemer:
@JoshuaKGoldberg what would the Node folks respond with? The documentation is clear that files with `mjs` are sent to the ESM loader, so we should do the same.

lib/nodejs/esm-utils.js
JoshuaKGoldberg: I would like to vote for removing it.

JoshuaKGoldberg:
🤷 No specific requests, just that I know they were working on this with us recently. I would feel off making non-trivial logic changes without including them.

trentm:
> the Node folks

By "the Node folks", do you mean for example @legendecas who was involved with the recent https://github.com/mochajs/mocha/pull/5366 changes?

legendecas (commented):
Would you mind updating the test in https://github.com/mochajs/mocha/blob/main/test/integration/compiler-esm.spec.js as well?

I think the test might be ineffective to catch the case described in https://github.com/mochajs/mocha/issues/5425.

mark-wiemer:
+1 to test coverage, not sure how I missed that!

Edit: lolol I was the one that opened this PR, I thought someone else wrote this code. Expect an updated test soon :)

lib/nodejs/esm-utils.js
mark-wiemer: Sure, we can remove it

mark-wiemer:
@legendecas, thanks for calling out test coverage, I think the new tests are good but I'd appreciate your review. I refactored existing ones to use a helper function so that the new tests don't triply-duplicate code.

legendecas (approved):
Thank you!

test/integration/compiler-esm-only-loader.spec.js
```diff
@@ -5,13 +5,18 @@ var path = require('node:path');
 
 // Regression test for https://github.com/mochajs/mocha/issues/5380
 describe('support ESM only module loader packages', function () {
-  it('should support ESM .js extension', function (done) {
+  /**
+   * Load the filename using an ESM-only loader and expect things to work
+   * @param {string} filename Name of the file to load
+   * @param {Function} done The done callback
```
JoshuaKGoldberg: [Docs] Nit: I think this clashes with the spirit of [`jsdoc/informative-docs`](https://github.com/gajus/eslint-plugin-jsdoc/blob/3bf911c48105a760d1d90a87a817522696eaf8f0/docs/rules/informative-docs.md#readme).

```suggestion
```

(here and elsewhere)

JoshuaKGoldberg (approved):
Looks great to me! I like the test flow, it's very clear. ✨

JoshuaKGoldberg:
By the way:

> ✅ All checks have passed

This is lovely to see in PRs again. 🥲

mark-wiemer:
Failures are just Coveralls, going to merge this one :) (we will be moving to Codecov, ref #5436 )

PR #5489, opened 2025-10-03
perf: cli/lookup-files.js, refactor, add jsdoc, reduce intermediary variables

Uzlopak:
`lookupFiles` repeats various operations over and over. We can avoid alot of this overhead and improve the performance of big test suites.

To improve the performance, we can normalize the file extensions once at the beginning and dont need to do it again if we have a recursive call. `normalizeFileExtensions` is a new function, which takes care of it once per run.

This means that we can avoid again normalizing the array in `hasMatchingFileExtension` (before: `hasMatchingExtname`). Also we avoid the performance bottleneck of Array.helper functions like `.map` and `.some()`.

`fs.statSync` is called with `throwIfNoEntry`, which exists since node 14.17.0, set to `false`. We can avoid "expensive" errors and the try catch blocks. We just check if stat is undefined, and if so, well, the path has an issue.

Added `jsdoc` for better type resolution. Also if you would put `// @ts-check` at the beginning of the file, it would pass.


lib/cli/lookup-files.js
```diff
@@ -24,30 +27,60 @@ const debug = require('debug')('mocha:cli:lookup-files');
...
+  if (!exts) {
+    return [];
+  }
+
+  for (let i = 0; i < exts.length; i++) {
+    if (exts[i][0] !== '.') {
+      exts[i] = `.${exts[i]}`;
+    }
+  }
+  return exts;
```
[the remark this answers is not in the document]
Uzlopak: We are processing the file paths internally. So we can dog food this function as we like.

lib/cli/lookup-files.js
JoshuaKGoldberg, line 1: > performance

@Uzlopak this looks like a lot of good optimizations, but I'm not aware of any known performance woes with file lookups. Can you demonstrate that there are issues for users? A reproduction case + measurement script, etc.?

lib/cli/lookup-files.js
```diff
@@ -1,18 +1,22 @@
...
-var errors = require('../errors');
-var createNoFilesMatchPatternError = errors.createNoFilesMatchPatternError;
-var createMissingArgumentError = errors.createMissingArgumentError;
+const fs = require('node:fs');
+const path = require('node:path');
+const glob = require('glob');
+const {
+  createNoFilesMatchPatternError,
+  createMissingArgumentError
+} = require('../errors');
```
JoshuaKGoldberg, line 15: Can you separate out these unrelated changes, like `var` -> `const`? I'm in favor of them in general but the diff is already pretty big.

If you want to holistically change most/all `var`s to `const`s separately I'd be very in favor.

lib/cli/lookup-files.js
```diff
@@ -24,30 +28,54 @@ const debug = require('debug')('mocha:cli:lookup-files');
...
+ * @private
+ * @param {FileExtension[]|string[]|undefined|null} exts
+ * @returns {FileExtension[]}
+ */
+const normalizeFileExtensions = (exts) => {
+  if (!exts) {
+    return [];
+  }
+
+  for (var i = 0; i < exts.length; i++) {
```
mark-wiemer, line 51:
```suggestion
  for (let i = 0; i < exts.length; i++) {
```

lib/cli/lookup-files.js
```diff
@@ -24,30 +28,54 @@ const debug = require('debug')('mocha:cli:lookup-files');
...
 /**
- * Determines if pathname has a matching file extension.
+ * Normalize file extensions to ensure they have a leading period character.
  *
- * Supports multi-part extensions.
+ * @private
+ * @param {FileExtension[]|string[]|undefined|null} exts
+ * @returns {FileExtension[]}
+ */
+const normalizeFileExtensions = (exts) => {
```
mark-wiemer, line 46: is the expected behavior for `normalizeFileExtensions(['hello.world'])` really to return `[.hello.world']`? extensions should just be the stuff after the last `.` . Maybe we are careful about this in the callers but it's something to consider.

That said, I haven't read the rest of the PR yet!

lib/cli/lookup-files.js
```diff
@@ -59,31 +87,30 @@ const hasMatchingExtname = (pathname, exts = []) =>
...
+ * @returns {string[]} An array of paths.
  * @throws {Error} if no files match pattern.
- * @throws {TypeError} if `filepath` is directory and `extensions` not provided.
+ * @throws {TypeError} if `filepath` is directory and `fileExtensions` not 
+ * provided or an empty array.
  */
 module.exports = function lookupFiles(
   filepath,
-  extensions = [],
+  fileExtensions,
```
mark-wiemer, line 99: let's make the code match the comment. I know `normalizeFileExtensions` does this for us but explicit is a bit more readable

```suggestion
  fileExtensions = [],
```

lib/cli/lookup-files.js
```diff
@@ -106,50 +133,51 @@ module.exports = function lookupFiles(
...
-    if (
-      !stat.isFile() ||
-      !hasMatchingExtname(pathname, extensions) ||
-      isHiddenOnUnix(pathname)
-    ) {
-      return;
+    // Handle directory
+    const dirEnts = fs.readdirSync(filepath, { recursive, withFileTypes: true });
+
+    for (var i = 0; i < dirEnts.length; i++) {
```
mark-wiemer, line 157:
```suggestion
    for (const dirEnt of dirEnts) {
```

lib/cli/lookup-files.js
```diff
@@ -106,50 +133,51 @@ module.exports = function lookupFiles(
...
+      const dirEnt = dirEnts[i];
+
+      const pathname = dirEnt.parentPath
+        ? path.join(dirEnt.parentPath, dirEnt.name)
+        : path.join(filepath, dirEnt.name);
```
mark-wiemer, line 162:
```suggestion
      const pathToJoin = dirEnt.parentPath ?? filepath
      const pathname = path.join(pathToJoin, dirEnt.name)
```

lib/cli/lookup-files.js
mark-wiemer: This is my main concern as well. Reviewing a PR like this takes a lot of time. I don't want this work to languish but I don't want to ignore potentially bigger issues with Mocha (pipeline failures, known bugs, easy requested features, etc.) I think we should follow the [perf issue template](https://github.com/mochajs/mocha/blob/main/.github/ISSUE_TEMPLATE/04-performance.yml) and discuss before taking up significant work like this so we're on the same page with review timelines and priority

mark-wiemer (commented):
First pass looks good, I'll have to review code coverage and triple check things before I'm comfortable signing off though

mark-wiemer:
Hey @Uzlopak, going to close this one for now. You're welcome to reopen it as you see fit :)

PR #5490, opened 2025-10-03
chore: remove "he" dependency

Uzlopak:
We use of the 101 kb javascript code of the "he" only about 500 bytes of trivial code. Implementing our own `escape` function to escape html strings. Simplyfying the code and using more descriptive names for the variables and helper functions.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -10,7 +10,6 @@
  */
 var path = require('node:path');
 var util = require('node:util');
-var he = require('he');
 
 const MOCHA_ID_PROP_NAME = '__mocha_id__';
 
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (changes requested):
This is a lot of added logic for us without corresponding tests. HTML encoding is tricky. If `he` is big (and last updated 7 years ago) I think it'd be preferable to find a more modern alternative. Have you checked with the [e18e](https://e18e.dev) folks about `he`?

124 kB unpacked for [`he@1.2.0`](https://www.npmjs.com/package/he/v/1.2.0) isn't _great_, but IMO it's worth not having to do this all ourselves. Plus, it's a pretty widely used package. I would expect lots of older users of Mocha already have it in their dependency chain. https://e18e.dev/learn/advocacy.html#dependency-tree and all.

mark-wiemer (commented):
Concurring opinion: I wouldn't call this trivial, but I am happy to have it in our codebase. It just needs text and maybe some explanation: what does `&#x` represent? why subtract* `0xD800`? lots of magic numbers make this a bit of a risky change, tests would help explain this

mark-wiemer (changes requested):
see above

mark-wiemer:
Hey @Uzlopak , closing this one for now, feel free to reopen as you see fit

PR #5491, opened 2025-10-06
ES6 Class Declarations (Part One)

### LINKED ISSUE #5025: 🐛 Bug: remove util.inherits / convert to classes

Offroaders123:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5025

## Overview

This is the (nearly) complete migration to the ES6 class syntax, from the original constructor function implementation.

This was originally intended to include changes for two more files, but they both were having challenges with passing tests. Those were `lib/mocha.js`, and `lib/reporters/base.js`. As they seemingly will require more formatting than simple syntax changes, I am leaving them for a separate commit.

*A concern of mine:*
For the goal of diff clarity, I did not perform a Prettier format step. I am happy to add a second commit, which does the formatting separately!

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -21,14 +21,7 @@ var EVENT_RUN_BEGIN = constants.EVENT_RUN_BEGIN;
 var EVENT_RUN_END = constants.EVENT_RUN_END;
 var EVENT_TEST_PENDING = constants.EVENT_TEST_PENDING;
 var EVENT_TEST_END = constants.EVENT_TEST_END;
-var inherits = require('../utils').inherits;
 var sprintf = util.format;
-
-/**
- * Expose `TAP`.
- */
-
-exports = module.exports = TAP;
 
 /**
  * Constructs a new `TAP` reporter instance.
@@ -37,11 +30,14 @@ exports = module.exports = TAP;
  * @class
  * @memberof Mocha.reporters
  * @extends Mocha.reporters.Base
+ */
+class TAP extends Base {
+/**
  * @param {Runner} runner - Instance triggers reporter actions.
  * @param {Object} [options] - runner options
  */
-function TAP(runner, options) {
-  Base.call(this, runner, options);
+constructor(runner, options) {
+  super(runner, options);
 
   var self = this;
   var n = 1;
@@ -79,11 +75,13 @@ function TAP(runner, options) {
     self._producer.writeEpilogue(runner.stats);
   });
 }
+}
 
 /**
- * Inherit from `Base.prototype`.
+ * Expose `TAP`.
  */
-inherits(TAP, Base);
+
+exports = module.exports = TAP;
 
 /**
  * Returns a TAP-safe title of `test`.
@@ -143,14 +141,14 @@ function createProducer(tapVersion) {
  * @private
  * @constructor
  */
-function TAPProducer() {}
+class TAPProducer {
 
 /**
  * Writes the TAP version to reporter output stream.
  *
  * @abstract
  */
-TAPProducer.prototype.writeVersion = function () {};
+writeVersion() {};
 
 /**
  * Writes the plan to reporter output stream.
@@ -158,7 +156,7 @@ TAPProducer.prototype.writeVersion = function () {};
  * @abstract
  * @param {number} ntests - Number of tests that are planned to run.
  */
-TAPProducer.prototype.writePlan = function (ntests) {
+writePlan(ntests) {
   println('%d..%d', 1, ntests);
 };
 
@@ -169,7 +167,7 @@ TAPProducer.prototype.writePlan = function (ntests) {
  * @param {number} n - Index of test that passed.
  * @param {Test} test - Instance containing test information.
  */
-TAPProducer.prototype.writePass = function (n, test) {
+writePass(n, test) {
   println('ok %d %s', n, title(test));
 };
 
@@ -180,7 +178,7 @@ TAPProducer.prototype.writePass = function (n, test) {
  * @param {number} n - Index of test that was skipped.
  * @param {Test} test - Instance containing test information.
  */
-TAPProducer.prototype.writePending = function (n, test) {
+writePending(n, test) {
   println('ok %d %s # SKIP -', n, title(test));
 };
 
@@ -191,7 +189,7 @@ TAPProducer.prototype.writePending = function (n, test) {
  * @param {number} n - Index of test that failed.
  * @param {Test} test - Instance containing test information.
  */
-TAPProducer.prototype.writeFail = function (n, test) {
+writeFail(n, test) {
   println('not ok %d %s', n, title(test));
 };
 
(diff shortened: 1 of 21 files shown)


lib/reporters/tap.js
```diff
@@ -143,22 +141,22 @@ function createProducer(tapVersion) {
...
  *
  * @abstract
  */
-TAPProducer.prototype.writeVersion = function () {};
+writeVersion() {};
 
 /**
  * Writes the plan to reporter output stream.
  *
  * @abstract
```
mark-wiemer, line 158: Tiny scope creep: the function isn't actually `abstract`, the docs have been wrong this whole time, let's fix it while we're here

```suggestion
```

lib/reporters/tap.js
```diff
@@ -201,14 +199,15 @@ TAPProducer.prototype.writeFail = function (n, test) {
...
```
mark-wiemer:
```suggestion
```

lib/runnable.js
```diff
@@ -32,10 +30,14 @@ module.exports = Runnable;
...
  * @public
+ */
+class Runnable extends EventEmitter {
+/**
  * @param {String} title
  * @param {Function} fn
  */
-function Runnable(title, fn) {
+constructor(title, fn) {
+  super();
```
mark-wiemer: should this call `super`?

lib/runnable.js
```diff
@@ -254,7 +251,7 @@ Runnable.prototype.resetTimeout = function () {
  * @private
  * @param {string[]} globals
  */
-Runnable.prototype.globals = function (globals) {
+globals(globals) {
```
mark-wiemer: Let's take this opportunity to un-shadow this variable by renaming it

```suggestion
globals(globalArgs) {
```

lib/runner.js
```diff
@@ -238,15 +238,6 @@ class Runner extends EventEmitter {
...
   }
-}
-
-/**
- * Wrapper for setImmediate, process.nextTick, or browser polyfill.
- *
- * @param {Function} fn
- * @private
- */
-Runner.immediately = global.setImmediate || process.nextTick;
```
mark-wiemer, line 249: please move this back up to line ~250, a tiny part of me is worried these values might be changed somewhere in the next 1000 lines :D

lib/runner.js
```diff
@@ -1221,9 +1212,18 @@ Runner.prototype.isParallelMode = function isParallelMode() {
...
-Runner.prototype.workerReporter = function () {
+workerReporter() {
   throw createUnsupportedError('workerReporter() not supported in serial mode');
 };
+}
+
+/**
+ * Wrapper for setImmediate, process.nextTick, or browser polyfill.
+ *
+ * @param {Function} fn
```
mark-wiemer: Docs were wrong the whole time

```suggestion
```

lib/suite.js
```diff
@@ -48,17 +51,12 @@ Suite.create = function (parent, title) {
...
- * @class
- * @extends EventEmitter
- * @see {@link https://nodejs.org/api/events.html#events_class_eventemitter|EventEmitter}
  * @param {string} title - Suite title.
  * @param {Context} parentContext - Parent context instance.
  * @param {boolean} [isRoot=false] - Whether this is the root suite.
  */
-function Suite(title, parentContext, isRoot) {
+constructor(title, parentContext, isRoot) {
+  super();
```
mark-wiemer: just checking if we need to explicitly call `super` here as well. I might be missing something but I don't think we need it

lib/suite.js
```diff
@@ -668,3 +662,9 @@ var constants = defineConstants(
 );
 
 Suite.constants = constants;
+
+/**
+ * Expose `Suite`.
+ */
+
```
mark-wiemer: We **could** get rid of all comments of this nature in this PR, but it's very low priority. However this one bothers me for some reason so I'm mentioning it here :D

```suggestion
```

lib/test.js
```diff
@@ -28,21 +29,16 @@ function Test(title, fn) {
       'string'
     );
   }
+  super(title, fn);
   this.type = 'test';
```
mark-wiemer: Let's err on the side of caution ;)

```suggestion
  this.type = 'test';
  super(title, fn);
```

mark-wiemer (changes requested):
Looks very good, just a few small comments and questions :) thanks for this and for keeping the PR focused on one core issue!

lib/runnable.js
Offroaders123: For ES6 classes, a constructor that extends a class requires a `super()` call, unfortunately, or else it will error when constructed.

Super Constructor Demo

lib/suite.js
Offroaders123: Was feeling the same as well, haha.

lib/test.js
Offroaders123: This one is a similar issue as to the `super()` call from earlier, it might technically be the same issue in the grand scheme of things. Accessing `this` before calling `super()` is a `ReferenceError`.

Accessing This Before Super Demo

JoshuaKGoldberg:
> For the goal of diff clarity, I did not perform a Prettier format step. I am happy to add a second commit, which does the formatting separately!

Blurgh, yeah, this has been bugging me. I've been meaning to get us to a clean baseline for a while. Thanks for the reminder 😄. Sent: #5492 -> #5493.

lib/runner.js
Offroaders123: Ideally it would be able to stay there, I would wished it could have, too. But since ES2022 class field syntax isn't supported by the tooling we currently have, I'm not sure if there are any other ways to define them without moving it out of the ES6 class syntax scope (normal variable assignments aren't valid there).

lib/test.js
mark-wiemer: I don't think this will be a problem

lib/runnable.js
```diff
@@ -251,7 +251,7 @@ resetTimeout() {
  * @private
  * @param {string[]} globals
```
mark-wiemer, line 255:
```suggestion
 * @param {string[]} globalArgs
```

Offroaders123:
Been intending to finalize the last few things here for this thread, but been a bit busy with other things as well. Going on a trip for the next few days, so I plan to finish everything once I'm back home 😃 [edit: back on Saturday]

mark-wiemer:
@Offroaders123 I am healthy now and happy to review this PR when you have a chance to clean it up. Hope your travels were fun :)

Offroaders123:
I'm a bit lost on where to carry forward next. Did I do things in the wrong order? I think I may have merged into my changes branch too soon, and I (now clearly) should have been checking the results of the tests while updating the code itself, rather than trying to fix the code to work afterwards. A worthy learning experience 🙂

Offroaders123:
Whoops, agh. Okay, so I was hoping to tidy things up and fix the upstream changes. I don't plan to close this itself, but moreso try to fix some of the errors that are stuck there. I will open another PR to start fresh then. I didn't think that this would close, it makes sense though. The branches for the development of everything up to now are all still available in my fork. I just wanted to start new on my `main` branch itself.

Offroaders123:
I'll move back to the main thread for figuring out what steps are next.

PR #5496, opened 2025-10-08
feat!: change the default of --forbid-only to check for process.env.CI

### LINKED ISSUE #5299: 🚀 Feature: Exit with code 1 when .only is used
### Feature Request Checklist

### Overview

It's an unfortunately common error to check in a test file with `.only`. This can hide failures especially when run in a CI system where the number of tests is not reported or easily viewable.

### Suggested Solution

Automatically fail the test run when a `.only` is found and emit exit code 1.

(This is the behavior of newer test runners such as the builtin Node.js and Deno runners.)

### Alternatives

Using a different testing framework. :smile:

### Additional Info

_No response_

RobinVdBroeck:
## PR Checklist

- [X] Addresses an existing open issue: fixes #5299

## Overview

I've changed the default value of `--forbid-only` to be the truthy value of process.env.CI.

Since this is a major change, I've added documentation for the behaviour before v12 (the next major release) and after. Let me know if this is not how docs are written for mocha, and how I should change it to be inline with mocha's standards.

I could use some pointers on where to add tests. I've only found integration level tests for testing CLI flags, but that seems kinda overkill for adding a default? Should I write an unit test somewhere, and if so, where?

### DIFF
--- a/.github/DEVELOPMENT.md
+++ b/.github/DEVELOPMENT.md
@@ -16,3 +16,32 @@ If you are having trouble, don't be afraid to [ask for help](./CONTRIBUTING.md#â
    - Do not use `yarn install` or `pnpm install`.
    - Some optional dependencies may fail; you can safely ignore these unless you are trying to build the documentation.
    - If you're sick of seeing the failures, run `npm install --ignore-scripts`.
+
+## Developing Mocha
+
+When you contribute to Mocha, you will probably want to try to run your changes on the test suite of another project. You can (and should) run the test suite of Mocha itself before committing, but also confirming that your changes give the expected result on another project.
+
+For example, [WebSocket.io](https://github.com/LearnBoost/websocket.io/):
+
+    $ git clone https://github.com/LearnBoost/websocket.io.git
+
+Retrieve websocket.io's dependencies, which will include the stable version of Mocha:
+
+    $ cd websocket.io/
+    $ npm install
+
+Replace the Mocha dependency by the current git repository:
+
+    $ cd node_modules/
+    $ mv mocha/ mocha.save
+    $ git clone https://github.com/mochajs/mocha.git
+
+Install Mocha's dependencies for the development version:
+
+    $ cd mocha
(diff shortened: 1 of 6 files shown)


mark-wiemer:
Thanks @RobinVdBroeck for this one :) unfortunately from what I'm seeing, our `exports.builder` function is a pretty tight wrapper around `yargs`, which I'm familiar with but not an expert in. Right now, I don't know how to test custom values being passed into this builder, it seems to always read from `process.argv`, so we can't test it in a unit test without significant refactoring, and I agree that an integration test feels like overkill.

Curious if @mochajs/maintenance-crew has any insights here, otherwise this may sit for a bit while I work on other bugfixes and think about this one

JoshuaKGoldberg:
Yeah, I think it's fine to omit testing for this one. The behavioral change is one line. Tests would be ideal but 🤷 it's old code locked into a design pattern we don't control.

RobinVdBroeck:
I've pushed 2 new commits:
- One fixes the existing integration tests that depend on .only, since these now fail while being run in CI. I missed this in my original PR since I did not run tests with `CI=true` :sweat_smile: 
- I've added that --forbid-only is the default in CI environments to the error message. Let me know if this is too much, I was on the fence myself, and I'll drop the commit again.

test/integration/options/delay.spec.js
```diff
@@ -5,11 +5,7 @@ var helpers = require('../helpers');
 var runMochaJSON = helpers.runMochaJSON;
 
 describe('--delay', function () {
-  var args = [];
-
-  before(function () {
-    args = ['--delay'];
-  });
```
RobinVdBroeck, line 12: This before block was doing nothing, so I removed it and put it directly on the args array.

mark-wiemer:
Hi @RobinVdBroeck would you mind resolving the new conflicts? They should just be format fixes, `npm run format:prettier` should resolve most of them :)

The new message looks good to me, thank you

PR #5408, opened 2025-07-29
fix: use original require() error for TS files if ERR_UNKNOWN_FILE_EXTENSION

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5393
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Adds in a special-case `try`/`catch` -> `throw`. If the file being imported is a `.cts`, `.mts`, or `.ts` _and_ the caught error from `import()` is for an unknown file extension, we know it's from a Node.js version that doesn't support those file extensions. In that case we'll want to throw the original informative `require()` error per #5393.

💖

Issue #5393, opened 2025-06-26
🐛 Bug: original require() errors are masked with an unrelated `ERR_UNKNOWN_FILE_EXTENSION`


lkho:
### Bug Report Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [x] I want to provide a PR to resolve this (but after community discussion first)

### Expected

in 11.7.0, if `require()` fails, the error is thrown and output to console. e.g.:

```
 $ mocha --require ts-node/register --require source-map-support/register 'a.ts'
 
  Exception during run: .../a.ts:61:66 - error TS2345: Argument of type '{ address: string; }' is not assignable to parameter of type ...
 
 $ mocha --version
 11.7.0
```

### Actual

after this change https://github.com/mochajs/mocha/pull/5384#diff-30714d24f4a4acd4826e5f7e92f58922ed0ab41b1f0062ae2a46df6c5c899eb4L93-L102
the `err` is masked and returned an `ERR_UNKNOWN_FILE_EXTENSION` which is misleading. 

```
 $ mocha --require ts-node/register --require source-map-support/register 'a.ts'
 
  Exception during run: TypeError: Unknown file extension ".ts" for /.../a.ts
     at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:219:9)
     at defaultGetFormat (node:internal/modules/esm/get_format:245:36)
     at defaultLoad (node:internal/modules/esm/load:120:22)
     at async ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:580:32)
     at async ModuleJob._link (node:internal/modules/esm/module_job:116:19) {
   code: 'ERR_UNKNOWN_FILE_EXTENSION'
 }
 
 $ mocha --version
 11.7.1
```

[...]

Issue #5235, opened 2024-10-25
🚀 Feature: if loading a file as CJS and ESM both fail, display both errors instead of just ESM


jedwards1211:
### Feature Request Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

Ever since `chai` v5 dropped support for CJS, we get errors like this when we accidentally install chai 5+:

```
$ mocha 

 Exception during run: TypeError: Unknown file extension ".ts" for /Users/user/src/battalion/storage-site-sim/src/__tests__/SeededRNG.test.ts
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:160:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:203:36)
    at defaultLoad (node:internal/modules/esm/load:141:22)
    at ModuleLoader.load (node:internal/modules/esm/loader:409:7)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:291:45)
    at link (node:internal/modules/esm/module_job:76:21) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```

This threw us for a loop at first, because normally our tests work fine with `@babel/register` handling `.ts` files.

What's happening is first Mocha tries to load our file as CJS, which make it as far as `require('chai')` which fails with `ERROR_REQUIRE_ESM`.  Then Mocha tries to load our file as ESM instead, which triggers the above error.

### Suggested Solution

If Mocha would print out the error from loading our file as CJS in addition to the above error, we would have seen what caused the problem more quickly.

### Alternatives

I've thought about adding code to our toolchain to check for chai v5+ and explicitly warn...but this same issue could occur with any package we're using in tests that drops CJS support

I guess another solution would be a mocharc option to force Mocha to only try one of CJS or ESM, not both.

[...]

PR #5498, opened 2025-10-10
fix: swallow more `require` errors from *ts files

### LINKED ISSUE #5494: 🐛 Bug: Mocha v11.7.3 and v11.7.4 fails with error ERR_REQUIRE_ESM
### Bug Report Checklist

### Expected

Unit tests runs as expected

### Actual

Unit tests fail with error ERR_REQUIRE_ESM

After some investigation, this problem is the same as #5387 which was fixed in 11.7.1 and 11.7.2 but is broken again in 11.7.3 and 11.7.4 .
The problem has been re-introduced with PR #5408 and the issue might be in `lib/nodejs/esm-utils.js` line 108 in the PR, the condition reads
```javascript
      if (
        /\.(cts|mts|ts)$/.test(file) || // <- or
        importErr.code === 'ERR_UNKNOWN_FILE_EXTENSION'
      ) {
        throw requireErr;
      }
```
but in the PR description it is

If the file being imported is a .cts, .mts, or .ts *and* the caught error from import() is for an unknown file extension, we know it's from a Node.js version that doesn't support those file extensions.

**or** in the code, **and** in description.

Note : While making the minimal reproducible example, it appears that if the imported module has no async bit the tests are not run and fail silently.

```typescript
// src/main.test.ts
import "./other.js";

describe("the test", function () {
  it("should work", function () {
    console.log("ok");
  });
});
```

```typescript
// src/other.ts
export const ok = () => {
  console.log("ok");
};

// there must be an async involved somewhere or the error does not trigger
await Promise.resolve("something");

// simulate an error in a module
throw new Error("stop");
```

```bash
npx mocha '--import=data:text/javascript,import{register}from"node:module";import{pathToFileURL}from"node:url";register("ts-node/esm",pathToFileURL("./"));' --enable-source-maps --inspect 'src/*.test.ts'
```

### Versions

v11.7.3, v11.7.4
Nodejs v24.3.0

### Additional Info

_No response_

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5494, works on #5361

## Overview

In #5494, users reported `requireErr` being thrown when `importErr` should've been thrown instead. Recent change #5408 accidentally throws `requireErr` when the file is a *ts file **_or_** `importErr.code === 'ERR_UNKNOWN_FILE_EXTENSION'`. This condition should've been **_and_**, because we don't want to throw the earlier `requireErr` when the later `importErr` is meaningful (e.g. from a *ts file but **not** an unknown file extension).

Per research into https://github.com/mochajs/mocha/issues/5361#issuecomment-3368708708, we know Node's loader / "customization hook" behavior changes whether the parent package is in module mode or not. Because our CI tests run through `nyc`, Node was always "primed" to import modules, so `importErr` and `requireErr` were the same in tests that expected them to be different. This is why we didn't detect the issue earlier.

This PR adds `test-node-integration:bare` to the only test that needs it. Bare tests are run directly via `npx mocha` to avoid accidentally "priming" Node to import modules.

(This is a PR into the new `v11.x` branch, so we'll have to find a way to push 11.7.5 instead of 12.0.0-beta-1. Not sure if we want this to be a one-off for now or work towards a high-resilience deployment system. We haven't defined our support of old Mocha versions, but I believe we should be generous in backporting given our userbase)


mark-wiemer:
See [run 4329](https://github.com/mochajs/mocha/actions/runs/18408396533) for test results

mark-wiemer:
No new failures, ready for review. Though Coveralls was down so all test jobs failed 😭  I checked the integration ones and nothing went wrong there!

.github/DEVELOPMENT.md
```diff
@@ -5,6 +5,11 @@ If you are having trouble, don't be afraid to [ask for help](./CONTRIBUTING.md#
 
 ## Prerequisites
 
+<!--
+todo update this section and setup section
+https://github.com/mochajs/mocha/issues/5497
+-->
+
```
JoshuaKGoldberg:
```suggestion
```

.github/DEVELOPMENT.md
```diff
@@ -16,3 +21,11 @@ If you are having trouble, don't be afraid to [ask for help](./CONTRIBUTING.md#
    - Do not use `yarn install` or `pnpm install`.
    - Some optional dependencies may fail; you can safely ignore these unless you are trying to build the documentation.
    - If you're sick of seeing the failures, run `npm install --ignore-scripts`.
+
+## Tests
+
+Running tests via `npm test` should just work on any operating system with any supported version of Node. If they don't, please file an issue or contact us in [our Discord](https://discord.gg/KeDn2uXhER).
```
JoshuaKGoldberg: Nit: "just" implies this is always easy and straightforward, which isn't the case for many newcomers.

```suggestion
Running tests via `npm test` should work on any operating system with any supported version of Node. If they don't, please file an issue or contact us in [our Discord](https://discord.gg/KeDn2uXhER).
```

test/integration/esm.spec.js
```diff
@@ -82,7 +82,7 @@ describe('esm', function () {
     );
   });
 
-  it('should throw an ERR_MODULE_NOT_FOUND and not ERR_REQUIRE_ESM if file imports a non-existing module with a loader', async function () {
+  it('should throw an ERR_MODULE_NOT_FOUND and not ERR_REQUIRE_ESM if file imports a non-existing module with a loader @bare', async function () {
```
JoshuaKGoldberg, line 85: [Non-Actionable] Amusingly, this is very much what #928 is asking for. Which has been open for >12 years and accepting PRs for about 5.

JoshuaKGoldberg (approved):
Thank you for so thoroughly covering this silly oversight of mine! +1 to the new form of test coverage.

mark-wiemer:
## Postmortem

I never fully wrapped my head around the exact problem here. Why was Node loading modules in a different way based on whether `nyc` had been enabled? How can we track that to diagnose future issues?

I'll be updating this comment as I reflect more, I plan to reproduce the issue in excruciating detail so that I can summarize it precisely.

## Open questions

- What is a customization hook?
- How does Node's module loading system work?
- How can I get docs for an exact patch version of Node, not just LTS?

## References

- 2024-10-25: #5235
- 2025-06-25: #5393
- 2025-07-29: #5408
- 2025-10-04: [Research into local machine failures when CI passes (#5361)](https://github.com/mochajs/mocha/issues/5361#issuecomment-3368708708)
- 2025-10-07: #5494 
- 2025-10-10: https://github.com/mochajs/mocha/tree/2f405a9c2409c0faa19c92b4d4ab6861ad06eb6f/sandbox/issue-5494

---

- [Node.js 22 LTS docs on customization hooks](https://nodejs.org/api/module.html#customization-hooks)

synjurnimoolla-hash (approved):

mark-wiemer:
@copilot+claude-opus-4.6 draft a report about why this error happened, it's quite complex. Organize sources so that the report is readable. Try to answer the open questions in this comment thread.

mark-wiemer:
New homepage for this is #5494

Issue #5483, opened 2025-10-02
🛠 Repo: Bump eslint from 8 to 9


JoshuaKGoldberg:
### Tooling Suggestion Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

Mocha has a devDependency right now on a 2-year-old, deprecated version of [`eslint`](https://www.npmjs.com/package/eslint/v/8.56.0):

https://github.com/mochajs/mocha/blob/b2985b3428b4b88ca220a14a26e9eb7139e8d445/package.json#L139

This isn't a problem on its own. `eslint@8` works as-is. But it pulls in old dependencies that bloat our local node_modules size. For example, in #5484, we see it pulling in `strip-json-comments@3`:

```plaintext
$ npm ls strip-json-comments@3
mocha@11.7.4 /Users/josh/repos/mocha
├─┬ eslint@8.56.0
│ └─┬ @eslint/eslintrc@2.1.4
│   └── strip-json-comments@3.1.1
├─┬ jsdoc@3.6.11
│ └── strip-json-comments@3.1.1
└─┬ markdownlint-cli@0.30.0
  └─┬ run-con@1.2.11
    └── strip-json-comments@3.1.1
```

I think it'd be a good idea in general to bump it to the latest major version. 

### Additional Info

In theory it'd be nice to automatically upgrade _all_ devDependencies. #5055 is blocked for now: https://github.com/mochajs/mocha/issues/5055#issuecomment-2633897266.

Also note that this issue is _not_ about significantly changing the list of rules we use. It's just about bumping the version of ESLint (and, if necessary, plugins / other related devDependencies). Rule & plugin changes are tracked separately.

See also [ESLint 9 Migration Guide](https://eslint.org/docs/latest/use/migrate-to-9.0.0)

💖

[...]

PR #5510, opened 2025-10-21
repo: update Rollup to v4

Lightning00Blade:
## PR Checklist

- [x] Addresses an existing open issue: works on #5483 (edited by @mark-wiemer)
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22) (edited by @mark-wiemer)

## Overview

Updates rollup to v4.
This change is needed for the EsLint v9 bump - https://github.com/mochajs/mocha/issues/5483.

Notable changes drops `rollup-plugin-node-global` and uses `rollup-plugin-polyfill-node` unconditionally.


eslint.config.js
```diff
@@ -176,6 +176,9 @@ module.exports = [
       "test/integration/fixtures/**",
       ".karma/**",
       "mocha.js",
+      // TODO: EsLint v8 parser can't parse Import attributes
```
mark-wiemer: Can you clarify what this comment means?

rollup.config.mjs
```diff
@@ -3,22 +3,22 @@ import commonjs from "@rollup/plugin-commonjs";
...
 import nodePolyfills from "rollup-plugin-polyfill-node";
-import globals from "rollup-plugin-node-globals";
 
 // Debugging tools
 import { visualizer } from "rollup-plugin-visualizer";
 
-import pickFromPackageJson from "./scripts/pick-from-package-json";
-import { version } from "./package.json";
+import pickFromPackageJson from "./scripts/pick-from-package-json.mjs";
+import packageJson from "./package.json" with { type: "json" };
```
mark-wiemer, line 11: why change this one? does the old `import { version } from "./package.json";` no longer work? would it work if we used `import { version } from "./package.json" with { type: "json" };` ?

mark-wiemer (commented):
Just commenting some first thoughts for now, I'll need to review Rollup in more detail and test this locally to see what's going on. Might be a week-ish, I usually prioritize bugfixes first.

rollup.config.mjs
Lightning00Blade: I have to double check the first part, that was something but for the Import Attributes only Default exports are possible 
```
Note: JSON modules only have one default export. You cannot do named imports from them (like import { name } from "data.json").
```
[MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with#:~:text=JSON%20modules%20only%20have%20one%20default%20export.)

eslint.config.js
Lightning00Blade: If you use [Import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) they can't be parsed by the AST builder used in EsLint v8 and it crashes the lint process.

Lightning00Blade: So `import packageJson from "./package.json" with { type: "json" };` result is something like 
```
Error: unexpected character `{` in X
```

rollup.config.mjs
Lightning00Blade: And to answer the first question, no you get a Error (Tested Node v20,v22,v24)
```bash
TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module "./package.json" needs an import attribute of "type: json"
```

mark-wiemer: Thanks for the followup. This one is still low on the priority list but I appreciate the PR being opened and ready for review :)

mark-wiemer:
PR is actually quite small and tests are still passing. (99% of lines changed are in package-lock.json)

Curious for @JoshuaKGoldberg's thoughts on this, I'll review tmrw in more detail.

eslint.config.js
JoshuaKGoldberg:
```suggestion
      // TODO: ESLint v8 parser can't parse Import attributes
```

PR #5526, opened 2025-11-05
fix: correct assertion import syntax in getting-started guide

### LINKED ISSUE #5517: 📝 Docs: Change node:assert import from '{ assert }' to 'assert'
### Documentation Request Checklist

### Overview

The test file on this URL: https://mochajs.org/next/getting-started/

Suggests that { assert } is to be imported from 'node:assert'. 

However, I get the following error:

> Exception during run: file:///C:/Users/ACER/Strongest/test/test/test1.mjs:1
> import { assert } from "node:assert";
>          ^^^^^^
> SyntaxError: The requested module 'node:assert' does not provide an export named 'assert'

I went into the actual code of node:assert and found that 'assert' is being exported as an object inside a variable and not a class or a function. 
That's why I am receiving this error. 

To resolve the issue, I changed the statement to:

> import assert from "node:assert";

The code then works perfectly.

While I know that on my PC, only 'assert' also works but I figured it's because I'm using general JavaScript and not TypeScript. I recall how when using the Deno framework it is required to call NodeJS libraries using 'node:<lib_name>'.

### Additional Info

Here's a screenshot of the code that I looked into from which I found out that 'assert' is being exported as an object inside a variable.

Screenshot of code that shows 'assert' exported as an object

Further, Here's a full screenshot of the error I discussed receiving:

Screenshot of error when using current import statement

Here's the screenshot of code working perfectly when the import statement is updated.

Screenshot of output when import statement is updated

saranganet:
Change named import to default import for node:assert in docs-next/src/content/docs/getting-started.mdx to match Node’s default export and avoid runtime error.

fixes #5517

### DIFF
--- a/docs-next/src/content/docs/getting-started.mdx
+++ b/docs-next/src/content/docs/getting-started.mdx
@@ -21,7 +21,7 @@ Create the following `example.test.js` file under a `test/` directory:
 
 ```js
 // test/example.test.js
-import { assert } from "node:assert";
+import assert from "node:assert";
 
 describe("Array", function () {
   describe("#indexOf()", function () {


mark-wiemer (approved):
Thank you! I've created https://github.com/mochajs/mocha/issues/5528 as well.

PR #5527, opened 2025-11-05
feat: add mocha.mjs export

### LINKED ISSUE #5211: 🚀 Feature: Ability to import mocha from ESM in the browser

brettz9:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5211

## Overview

Adds an ESM export file for use in the browser.

The other non-underscored (private) keys I saw on the global `mocha` object which I wasn't sure which I might include were the following:

`files`, `options`, `suite`, `isWorker`,  `throwError`,  `ui`

---

2026-08-27: updating to stop using brittle filename (already changed to mocha.js):

https://github.com/googleapis/release-please#how-can-i-fix-release-notes

BEGIN_COMMIT_OVERRIDE
feat: add ESM export for browsers
END_COMMIT_OVERRIDE

### DIFF
--- a/mocha.mjs
+++ b/mocha.mjs
@@ -1,4 +1,4 @@
-// this build file is not committed to the repo
+// this built file is not committed to the repo
 /* eslint-disable-next-line n/no-missing-import */
 import "./mocha.js";
 
--- a/mocha.mjs
+++ b/mocha.mjs
@@ -2,27 +2,50 @@
 /* eslint-disable-next-line n/no-missing-import */
 import "./mocha.js";
 
-const { mocha } = globalThis;
+const {mocha, Mocha} = globalThis;
 
-mocha.ui();
+const mochaExports = {
+  afterEach: (...args) => globalThis.afterEach?.(...args),
+  after: (...args) => globalThis.after?.(...args),
+  beforeEach: (...args) => globalThis.beforeEach?.(...args),
+  before: (...args) => globalThis.before?.(...args),
+  describe: (...args) => globalThis.describe?.(...args),
+  it: (...args) => globalThis.it?.(...args),
+  xdescribe: (...args) => globalThis.xdescribe?.(...args),
+  xit: (...args) => globalThis.xit?.(...args),
+  setup: (...args) => mocha.setup?.(...args),
+  suiteSetup: (...args) => globalThis.suiteSetup?.(...args),
+  suiteTeardown: (...args) => globalThis.suiteTeardown?.(...args),
+  suite: (...args) => globalThis.suite?.(...args),
+  teardown: (...args) => globalThis.teardown?.(...args),
+  test: (...args) => globalThis.test?.(...args),
+  xspecify: (...args) => globalThis.xspecify?.(...args),
+  specify: (...args) => globalThis.specify?.(...args),
+  context: (...args) => globalThis.context?.(...args),
+  xcontext: (...args) => globalThis.xcontext?.(...args),
+  run: mocha.run.bind(mocha)
+};
 
-const after = globalThis.after.bind(mocha);
-const afterEach = globalThis.afterEach.bind(mocha);
-const before = globalThis.before.bind(mocha);
-const beforeEach = globalThis.beforeEach.bind(mocha);
-const context = globalThis.context.bind(mocha);
-const describe = globalThis.describe.bind(mocha);
-const it = globalThis.it.bind(mocha);
-const specify = globalThis.specify.bind(mocha);
-const xcontext = globalThis.xcontext.bind(mocha);
-const xdescribe = globalThis.xdescribe.bind(mocha);
-const xit = globalThis.xit.bind(mocha);
-const xspecify = globalThis.xspecify.bind(mocha);
+export default Mocha;
 
-const run = mocha.run.bind(mocha);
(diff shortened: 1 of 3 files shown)


mark-wiemer:
For now let's exclude the private keys, they shouldn't be necessary for this export.

mark-wiemer (approved):

brettz9:
Thank you for the review!

I saw the linting failed after the merge, so I added my addition of `/mocha.mjs` to the eslint config file. There's also still a failure with `/esm/mocha.js`, a file that predated this PR, but with that file not being published, I wasn't sure what that was about. It technically is not even allowed with the ".js" extension since the `package.json` is `type: 'commonjs'`. Don't know if that file should be converted to `.mjs`, removed, treated as ESM by the linter anyways, or what?

mark-wiemer:
Fixed the build error, but not sure how to manually test this one. Going to unassign myself as this isn't a focus area for me compared to bugfixes and the v12 release, but hopefully after v12 we can come back to this. Thanks for getting this started and continuing to follow up!

mark-wiemer:
@JoshuaKGoldberg if you want to help me clean up PRs you can help review this one ;)

mocha.mjs
```diff
@@ -0,0 +1,36 @@
+// this built file is not committed to the repo
+/* eslint-disable-next-line n/no-missing-import */
+import "./mocha.js";
+
+const { mocha } = globalThis;
+const after = mocha.after.bind(mocha);
```
JoshuaKGoldberg: [Bug] I tried this PR's branch out on https://github.com/brettz9/mocha-no-esm-minimal-test-repro, but got an error on this line:

```plaintext
Uncaught TypeError: Cannot read properties of undefined (reading 'bind')
    at mocha.mjs:6:27
```

Looking at `globalThis.mocha`:

```plaintext
Mocha {
  after: undefined
  afterEach: undefined
  before: undefined
  beforeEach: undefined
  context: undefined
  describe: undefined
  files: []
  isWorker: false
...
```

Does this PR actually work as intended? Am I holding it wrong?

JoshuaKGoldberg (changes requested):
This _looks_ good but I can't get it to work - is something missing?

mocha.mjs
brettz9: My apologies... You're right. I very wrongly assumed the methods would be available. I've updated.

I can remove the additional exports I added that are not desired.

mocha.mjs
```diff
@@ -0,0 +1,68 @@
+// this built file is not committed to the repo
+/* eslint-disable-next-line n/no-missing-import */
+import "./mocha.js";
+
+const {mocha, Mocha} = globalThis;
+
+const mochaExports = {
+  after: (...args) => globalThis.after?.(...args),
```
JoshuaKGoldberg, line 8: I don't think Mocha should fail silently when the global `after` doesn't exist. That would indicate the page hasn't run `mocha.setup("bdd")` or an equivalent. Which would be something they'd want an error for.

```suggestion
  after: (...args) => after(...args),
```

In fact, I think it'd be useful to even throw a specific error:

```suggestion
  after: (...args) => {
    if (typeof after === "undefined") {
      throw new Error("`after` is not yet available. `mocha.setup("bdd\")` needs to be run first.");
    }
  },
```

WDYT?

JoshuaKGoldberg (changes requested):
Confirmed this actually does work now. Great! ✨ 

Requesting changes to make it a little more informative if users mess this up. IME forgetting to run `mocha.setup("bdd")` (or running it out of order) is pretty easy to do.

mocha.mjs
brettz9: The regular UMD mocha implementation silently fails when `mocha.setup` is not called. I would think such error handling would be desirable to have within the UMD implementation too. I've run into missing that as well. If separate handling is then necessary in ESM, I think we could add it there also.

I'd be happy to add it in ESM alone, but it seems to me it might be better something for the whole project. WDYT?

While on the topic, I'd personally favor `bdd` just working as the default if no prior setup call has been made (it is the default as far as no arguments passing, but a call to `mocha.setup` is always required in the browser). I also wonder if `mocha.run()` could be run by default as well (maybe based on a timer after the first `describe` call)--it'd be nice not to have separate behaviors required for the browser. Even the `<div id="mocha">` could be auto-appended if not found.

JoshuaKGoldberg: Good point! Yes, I think you're right. It's fine as-is. We can treat all of those suggestions as followup issues. 👍

JoshuaKGoldberg (approved):
Looks great to me, thanks! Bouncing back to @mark-wiemer for re-review as it's changed since Mark's last review.

At some point I think we'll want to revamp the browser-testing docs. But they're already pretty out of date separately from this, and will be captured by general docs cleanups.

mark-wiemer (approved):

mark-wiemer:
Created #5578. Thanks for the PR, @brettz9 !

PR #5542, opened 2025-11-23
feat: include test duration in timeout error messages

### LINKED ISSUE #5543: 🚀 Feature: Include Actual Test Duration in Timeout Error Messages
### Feature Request Checklist

### Overview

Right now, timeout errors look like:

`Error: Timeout of 2000ms exceeded.`

This message does not tell developers whether the test ran only slightly over (e.g., 2100ms) or was massively slow (e.g., 15000ms).
Both scenarios require different debugging strategies.

Add the actual test runtime to timeout error messages when available, such as:

`Error: Timeout of 2000ms exceeded. Test ran for 2450ms.`

### Suggested Solution

I propose extending the timeout error logic so that the actual duration of the test is included whenever available. This involves:
Tracking the start time of each runnable test.
Passing the computed duration into _timeoutError() when a timeout occurs or when a test completes after the timeout threshold.
Updating the error message format to include something like:
“Test ran for Xms” (only when duration data exists).

### Alternatives

-Keeping current behavior: Developers manually instrument slow tests, use custom timers, or rely on external tools, but this adds friction and is not ideal in CI environments.

-Adding a verbose/debug mode: Only include durations when a debug flag is active. This reduces the risk of reporter output changes but makes helpful information harder to discover.

### Additional Info

_No response_

Jeet-Srivastava:
Enhance timeout error messages to include the actual test duration when available. This provides developers with more context when debugging slow tests that exceed their timeout threshold.

Changes:
- Modified _timeoutError() to accept optional duration parameter
- Track start time on runnable instance to calculate duration
- Include duration in error message when timeout fires or test finishes late
- Add comprehensive tests for duration inclusion in timeout errors

The error message now includes: 'Test ran for Xms' when duration is available, making it easier to understand how much longer the test took than the timeout.

- fixes #5543

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -67,6 +67,7 @@ Runnable.prototype.reset = function () {
   this.pending = false;
   delete this.state;
   delete this.err;
+  delete this._startTime;
 };
 
 /**
@@ -243,7 +244,11 @@ Runnable.prototype.resetTimeout = function () {
     if (self.timeout() === 0) {
       return;
     }
-    self.callback(self._timeoutError(ms));
+    var duration;
+    if (self._startTime) {
+      duration = new Date() - self._startTime;
+    }
+    self.callback(self._timeoutError(ms, duration));
     self.timedOut = true;
   }, ms);
 };
@@ -270,6 +275,7 @@ Runnable.prototype.globals = function (globals) {
 Runnable.prototype.run = function (fn) {
   var self = this;
   var start = new Date();
+  this._startTime = start;
   var ctx = this.ctx;
   var finished;
   var errorWasHandled = false;
@@ -305,7 +311,7 @@ Runnable.prototype.run = function (fn) {
     self.duration = new Date() - start;
     finished = true;
     if (!err && self.duration > ms && ms > 0) {
-      err = self._timeoutError(ms);
+      err = self._timeoutError(ms, self.duration);
     }
     fn(err);
   }
@@ -420,11 +426,15 @@ Runnable.prototype.run = function (fn) {
  * Instantiates a "timeout" error
  *
  * @param {number} ms - Timeout (in milliseconds)
+ * @param {number} [duration] - Actual duration (in milliseconds), if available
  * @returns {Error} a "timeout" error
  * @private
  */
-Runnable.prototype._timeoutError = function (ms) {
+Runnable.prototype._timeoutError = function (ms, duration) {
   let msg = `Timeout of ${ms}ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.`;
+  if (typeof duration === "number" && duration > 0) {
+    msg += ` Test ran for ${duration}ms.`;
+  }
   if (this.file) {
     msg += " (" + this.file + ")";
   }
--- a/test/unit/runnable.spec.js
+++ b/test/unit/runnable.spec.js
@@ -623,6 +623,46 @@ describe("Runnable(title, fn)", function () {
           });
         });
       });
+
+      it("should include duration in timeout error message when timeout fires", function (done) {
+        var runnable = new Runnable("foo", function (done) {
+          setTimeout(done, 50);
+        });
+        runnable.timeout(20);
+        runnable.run(function (err) {
+          expect(err, "to satisfy", { code: TIMEOUT, timeout: 20 });
+          expect(err.message, "to match", /Test ran for \d+ms/);
+          done();
+        });
+      });
+
+      it("should include duration in timeout error message when test finishes late", function (done) {
+        var runnable = new Runnable("foo", function (done) {
+          setTimeout(function () {
+            done();
+          }, 30);
+        });
+        runnable.timeout(10);
+        runnable.run(function (err) {
+          expect(err, "to satisfy", { code: TIMEOUT, timeout: 10 });
+          expect(err.message, "to match", /Test ran for \d+ms/);
+          done();
+        });
+      });
+
+      it("should include duration in timeout error message for promise-based tests", function (done) {
+        var runnable = new Runnable("foo", function () {
+          return new Promise(function (resolve) {
+            setTimeout(resolve, 30);
+          });
+        });
+        runnable.timeout(10);
+        runnable.run(function (err) {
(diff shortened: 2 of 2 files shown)


saranganet:
seems good! 
I myself was gonna do that, great tho u did it.

mark-wiemer:
@Jeet-Srivastava can you open an issue so we can consider the use-cases here? Changing reporter output is usually a semver-major change which would require some thought as to "do we make this a CLI flag first?" and other questions

Jeet-Srivastava:
Hi @mark-wiemer, as you can see @saranganet has raised the issue, and i have replied to him there.
You can see that and then we can discuss further on it.

lib/runnable.js
```diff
@@ -243,7 +244,11 @@ Runnable.prototype.resetTimeout = function () {
     if (self.timeout() === 0) {
       return;
     }
-    self.callback(self._timeoutError(ms));
+    var duration;
+    if (self._startTime) {
+      duration = new Date() - self._startTime;
+    }
+    self.callback(self._timeoutError(ms, duration));
```
JoshuaKGoldberg: [question] (high confidence) In this timer-fired path the duration is structurally `~timeout + timer jitter`: Mocha aborts the runnable at the deadline, so a hung async test can never report a large number here — probing 15 hanging runnables at `timeout(60)`, every message said 61-63ms. Two consequences:

- the "slightly over vs. massively slow" discrimination from #5543 can't materialize in this branch — a test that would have finished 110ms late and one 1500ms late produce byte-identical messages;
- "Test ran for Xms" reads as a completed runtime, but at this point the body is still running (a probe whose body kept executing to ~400ms is reported as "ran for 53ms").

The branch where the number is genuinely informative is the late-finish one in `done()` below. Would it make sense to append the duration only there? That would keep the most common timeout message byte-identical to today (which also shrinks the reporter-output/compat surface discussed on the thread), and the sentence would only appear when it says something true and useful.

Related skew worth knowing about: `this.timeout(x)` mid-test re-arms the timer, but `_startTime` stays pinned at `run()` start, so the two numbers can describe different windows (e.g. "Timeout of 100ms exceeded... Test ran for 223ms" when Mocha did wait exactly 100ms after the re-arm).

lib/runnable.js
```diff
@@ -305,7 +311,7 @@ Runnable.prototype.run = function (fn) {
     self.duration = new Date() - start;
     finished = true;
     if (!err && self.duration > ms && ms > 0) {
-      err = self._timeoutError(ms);
+      err = self._timeoutError(ms, self.duration);
```
JoshuaKGoldberg: [question] (medium confidence) Since `createTimeoutError` already carries structured `err.timeout` and `err.file`, would an `err.duration` field be worth adding (alongside or instead of the message change)? Tools then wouldn't have to regex the prose for the number, and reporters could opt in to formatting it. It could also simplify the implementation: `self.duration` is computed right above on both paths, so decorating the error here in `done()` would avoid the `_startTime` plumbing and the extra parameter entirely.

lib/runnable.js
```diff
@@ -420,11 +426,15 @@ Runnable.prototype.run = function (fn) {
...
  * @param {number} ms - Timeout (in milliseconds)
+ * @param {number} [duration] - Actual duration (in milliseconds), if available
  * @returns {Error} a "timeout" error
  * @private
  */
-Runnable.prototype._timeoutError = function (ms) {
+Runnable.prototype._timeoutError = function (ms, duration) {
   let msg = `Timeout of ${ms}ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.`;
+  if (typeof duration === "number" && duration > 0) {
+    msg += ` Test ran for ${duration}ms.`;
```
JoshuaKGoldberg: [nit] (high confidence) Two wording points:

- `Runnable` is also the base class of `Hook`, so a timing-out `before`/`beforeEach` now reports "Test ran for Xms" — note the adjacent sentence deliberately says "For async tests *and hooks*". A noun-neutral phrasing (e.g. "took Xms") avoids that.
- Placement: the new sentence lands after the ~100-character advice sentence, separating the two numbers a reader wants to compare — and in the late-finish case (the one where the number is informative) the combined message contradicts itself: the duration proves `done()` *was* called, right after advice to "ensure done() is called". Something like `Timeout of ${ms}ms exceeded (took ${duration}ms).` up front might read better.

test/unit/runnable.spec.js
```diff
@@ -623,6 +623,46 @@ describe("Runnable(title, fn)", function () {
...
+        });
+        runnable.timeout(20);
+        runnable.run(function (err) {
+          expect(err, "to satisfy", { code: TIMEOUT, timeout: 20 });
+          expect(err.message, "to match", /Test ran for \d+ms/);
+          done();
+        });
+      });
+
+      it("should include duration in timeout error message when test finishes late", function (done) {
```
JoshuaKGoldberg: [blocking] (high confidence) This test doesn't exercise the late-finish branch its title describes: with `timeout(10)` and `done` firing at 30ms, the 10ms timer always wins, so it goes through `resetTimeout`'s path exactly like the test above (the reported duration is ~12ms, not ~30ms). The same is true of the promise-based test below. As a result the `err = self._timeoutError(ms, self.duration)` branch in `done()` has no unit coverage at all — reverting just that line still passes all 58 tests in this file — and `/Test ran for \d+ms/` also passes if the implementation hardcodes `duration = 1`.

Could you replace this one with a synchronous overrun, which deterministically reaches the `done()` branch, and pin the number with a lower bound? For example:

```js
var runnable = new Runnable("foo", function () {
  var end = Date.now() + 40;
  while (Date.now() < end) {}
});
runnable.timeout(10);
runnable.run(function (err) {
  expect(err, "to satisfy", { code: TIMEOUT, timeout: 10 });
  var match = /Test ran for (\d+)ms/.exec(err.message);
  expect(match, "to be truthy");
  expect(Number(match[1]), "to be greater than or equal to", 40);
  done();
});
```

(An integration-level home also exists: `test/integration/fixtures/timeout.fixture.js` already reaches this branch, but `timeout.spec.js` only asserts stats, never the message.)

Issue #5555, opened 2025-12-01
🐛 Bug: "Cannot find module" if module contains ts-node style 'TSError'


sam-super:
### Bug Report Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

When i have TS error in a module included in a test, it should show the TS error

### Actual

When there is a TS error it shows "Cannot find module".

### Minimal, Reproducible Example

I have what i think is a pretty simple typescript setup but at some point the compilation error handling has started causing a very hard to debug "Cannot find module", error rather than the TS error.

This is a stripped down repo which still exhibits the issue:
https://github.com/sam-super/mocha-ts-node-bug

### Versions

mocha@11.7.5
node@v24.11.0

### Additional Info

E.g. from the reference example repo:

```
npm run test -- test/included-module-ts-error.test.ts

Exception during run: Error: Cannot find module '[...]/test/helpers/bad-module' imported from [...]/test/included-module-ts-error.test.ts
```

Possibly related issue is it manages to execute a test with a TS error in the top-level test file/suite without complaint:
```
npm run test -- test/top-level-compilation.test.ts

  should error with ts error
    ✔ but passes when it should have failed
```

mccmrunal:
can  I take this one if anyone else hasn't started working

mark-wiemer:
@mccmrunal you can always work on any issue, but "status: in triage" bugs are not necessary accepted by maintainers, "status: accepting PRs" are more likely to be reviewed. That said, draft PRs for triaged bugs are always welcome as they can be helpful for triaging issues.

sam-super:
@mark-wiemer i can take a look at this if you think https://github.com/mochajs/mocha/pull/5557 was on the right lines?

[...]

PR #5557, opened 2025-12-02
fix: preserve TypeScript compilation errors instead of masking as MOD…

hk2166:
#5555 
When ts-node or similar TypeScript compilers encounter a type error,
the error was being caught and masked by fallback module loading logic,
resulting in unhelpful "Cannot find module" messages instead of the
actual TypeScript diagnostics.

This fix detects TypeScript compilation errors by checking for TS-specific
patterns in error messages and throws them immediately without attempting
fallback mechanisms, preserving the full diagnostic information for users.

Fixes issue where TypeScript errors in imported modules showed as
"Cannot find module" rather than displaying the actual type error.

### DIFF
--- a/lib/nodejs/esm-utils.js
+++ b/lib/nodejs/esm-utils.js
@@ -46,6 +46,20 @@ const tryImportAndRequire = async (file, esmDecorator) => {
   try {
     return dealWithExports(await formattedImport(file, esmDecorator));
   } catch (err) {
+    // Check if this is a TypeScript compilation error (from ts-node or similar)
+    // These often have detailed diagnostic information that we should preserve
+    const isTypescriptError =
+      err.message &&
+      (err.message.includes("TS") ||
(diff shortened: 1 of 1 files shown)


mark-wiemer (changes requested):
Off the bat, let's:
- de-dupe this code
- add unit tests
- see if we can narrow the `message.includes` section as these are pretty vague. We should identify which types of errors are being thrown

All this said, I'm still reviewing the underlying bug. Thanks for opening this promptly though :)

mark-wiemer:
RE "narrow the message.includes" section, we should be able to look at the error code instead. If the import error code is `MODULE_NOT_FOUND` and the require error is something more meaningful, we throw that instead. Ref https://github.com/mochajs/mocha/pull/5498/files#diff-30714d24f4a4acd4826e5f7e92f58922ed0ab41b1f0062ae2a46df6c5c899eb4 which also looks at the error code.

PR #5572, opened 2025-12-18
fix: surface ts-node compile errors

### LINKED ISSUE #5555: 🐛 Bug: "Cannot find module" if module contains ts-node style 'TSError'

sam-super:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5555

## Overview

Prevents "Cannot find module" when there are TS errors by prevent fallback to 'import()' when it errors.

The downside is this is tightly coupled to how ts-node throws errors when loading modules. However, this change seems unlikely to have any unexpected side-effects, so the risk is that it wont help if someone is using a different TS loader, but it's still a net gain for most.


lib/nodejs/esm-utils.js
```diff
@@ -97,6 +97,9 @@ const requireModule = async (file, esmDecorator) => {
   try {
     return require(file);
   } catch (requireErr) {
+    if (requireErr.name === 'TSError') {
```
JoshuaKGoldberg, line 101: [Refactor] Hmm. _Hmm_. Hardcoding `'TSError'` for `ts-node` only solves for `ts-node`. Ideally I don't think we should hardcode specific userland names into the general Mocha library. But I'm not sure there's a way to do that...

To recap, the root cause as I understand it is that the logic right now results in the wrong error getting shown to users:

1. First, `requireModule` attempts to `require(file)`
2. If that fails, it attempts to `formattedImport(file)`
3. If that also fails, the error from `formattedImport` is shown

...but in this case, it's really the error from `require` that I think we'd want.

This is similar to the previous #5235 -> #5393 -> #5408 -> #5498.

Anyway, for this new entry in the saga, I really don't see a better way of doing things. `ts-node` is the only popular userland library I know of that type-checks code as part of imports. That practice as a whole has been out of favor for a while - both as a _"this is slow, do it separately"_ concern and especially now that Node.js supports `.ts` files natively. So I don't think we'll see new libraries with new names to add.

@mark-wiemer I _think_ this change is reasonable, but I want to get your thoughts on it too. Is there some nice strategy I'm missing?

lib/nodejs/esm-utils.js
```diff
@@ -97,6 +97,9 @@ const requireModule = async (file, esmDecorator) => {
...
+      throw requireErr;
+    }
     debug("requireModule caught err: %O", requireErr.message);
```
JoshuaKGoldberg: [Bug] Let's move the new code to after this `debug()` line. `DEBUG=* npm run test` is a nice helpful thing users do.

test/node-unit/esm-utils.spec.js
JoshuaKGoldberg, line 1: [Praise] Very clean new test. Nice 🙂

JoshuaKGoldberg (changes requested):
I agree with the strategy, but would really like a spot check from another maintainer to make sure there's no way we can get around it.

Just requesting changes on moving the `debug()` line.

Thanks!

lib/nodejs/esm-utils.js
sam-super: will do!

lib/nodejs/esm-utils.js
sam-super: I agree it's not perfect.

To add something else to the mix i also sent this bug:
https://github.com/mochajs/mocha/issues/5579
The PR i will put up for that will be similar to this PR and will add another condition like so:
image

With a growing list of exceptions, it could point to a the default behavior (of a require error falling back to an import ) being the wrong way round. I wondered if it should only fall back in specific scenarios based on the error (inverting the default so it doesn't fall back), but i appreciate that might be an even worse case of whack-a-mole.

lib/nodejs/esm-utils.js
sam-super: done: 4a25ae7

lib/nodejs/esm-utils.js
mark-wiemer: I think the current PR is reasonable here. Swapping default behavior is more likely to cause more harm than good, we don't have a comprehensive test suite on which errors should go in which buckets, so let's not rock the boat too much.

JoshuaKGoldberg (approved):
LGTM, thanks! 🙌

mark-wiemer (approved):

mark-wiemer:
Test failure is known transient issue (#5361), merging. Thank you!

SchroederSteffen:
Thanks for the fix. Could this be downported to v11?

mark-wiemer:
Created https://github.com/mochajs/mocha/issues/6029 :)

PR #5608, opened 2026-01-04
build: initial file implementation for ocotoguide

### LINKED ISSUE #5573: 🛠 Repo: Switch from pr-compliance-action to OctoGuide

### Overview

Right now, this repo uses http://github.com/mtfoley/pr-compliance-action to enforce PRs have proper titles, resolve an issue, etc. That action is generally good for what it does. However:

* There are other things I'd want to enforce, such as alt text on images and user comments not just being _"+1"_
* pr-compliance-action isn't actively maintained right now

I made [OctoGuide](https://octo.guide) as a kind of "next step" action that does what pr-compliance-action does and a few more things. I propose we use it in Mocha.

### Additional Info

OctoGuide is pretty configurable (shoutout @cylewaitforit for adding in a bunch of features) so I imagine this will help with #5426 if we want to commit to that.

lennonnikolas:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5573 

## Overview
This change will replace the current pr-compliance-action to OctoGuide in the Github Actions for PR/issue management.


lennonnikolas:
I didn't define any rules. I would like to discuss with the maintainers if there are any rules to consider and what the text output of those rules should be if needed. As far as I can tell, the defaults from the Octoguide should be sufficient to migrate from the compliance.yml action to the octoguide.yml action.

mark-wiemer:
Feel free to add questions to the core issue :)

.github/workflows/octoguide.yml
```diff
@@ -0,0 +1,39 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          comment-header: "Octobot Response:"
+          include-bots: "false"
+          include-associations: "FIRST_TIMER,FIRST_TIME_CONTRIBUTOR,CONTRIBUTOR,MEMBER"
+          pr-body-descriptive: "true"
+          comment-meaningful: "true"
+          pr-title-conventional: "true"
+          pr-linked-issue: "true"
+          pr-task-completion: "true"
+          text-image-alt-text: "true"
```
JoshuaKGoldberg: Interesting: why did you fill all these out rules settings explicitly? They default to true and the docs on octo.guide don't suggest to.

As in - separate from the comments asking about which issues to enable. I'm saying these already default to `"true"` so I'm curious about what led you to set it there anyway.

(as an OctoGuide maintainer, understanding how folks interact with it / set it up is of high interest to me)

.github/workflows/octoguide.yml
JoshuaKGoldberg, line 1: > will replace the current pr-compliance-action

Just noting, you'll need to delete `compliance.yml` too

.github/workflows/octoguide.yml
mark-wiemer: TIL Josh is an OctoGuide maintainer! But +1 to his question, I am a fan of leaving defaults implicit to reduce the size of config files :)

lennonnikolas: Hi! So the reason I explicitly set rules values is because the documentation doesn't explicitly state what rules are enabled by default, only that you can customize predefined rules and set a flag to enable/disable. 

If necessary and/or wanted I can remove the explicit rules. This I believe is up to your discretion. However, on the converse, if the documentation did explicitly state the defaults for each of these rules,I believe it to be useful to have them here defined as a way to quickly understand what this github action is doing instead of relying and navigating to core documentation.

.github/workflows/octoguide.yml
lennonnikolas: Right! Baby steps. That will be my next commit.

.github/workflows/octoguide.yml
JoshuaKGoldberg: > TIL

I should also note I'm the _creator_ of OctoGuide, too. That's what I meant by _"I made [OctoGuide](https://octo.guide)"_ in #5573. Full disclaimer 🙂 

> documentation doesn't explicitly state what rules are enabled by default

Err, can I push back on that?

* https://octo.guide/configs#rules-table: explicitly states what rules are enabled in which config by default
* https://octo.guide/rules/comment-meaningful and other rule pages: explicitly state which config(s) the rule is enabled in

I suppose https://octo.guide/get-started#enabling-and-disabling-individual-rules only mentions _"in your chosen config"_, and doesn't explicitly link to them.

lennonnikolas: No worries, I'll remove the explicits. Thanks!

PR #5612, opened 2026-01-04
fix: allow chain call timeout to override nested items timeout

### LINKED ISSUE #5422: 🐛 Bug: timeout() chain call on `describe` not working as expected
> @mark-wiemer Thanks for your reply. "Copying the docs works" is because you have `.timeout(1000)` on `it()`, delete that and you'll find out the issue. I give you this snippet for example, by default mocha timesout 2000, so case failing for timeout has to be longer than that.
> 
> ```javascript
> describe('my suite', () => {
>   it('my test', async () => {
>     await new Promise((resolve) => {
>       setTimeout(resolve, 4000);
>     });
>   });
> }).timeout(5000);
> ```
> The issue is because mocha runs the "hook function" immediately, in which the test cases get created with a default timeout 2000. After that `suite.timeout(5000)` changes `suite._timeout` but has no effect on cases inside it.
> I'd open a new issue if you consider it a bug and would fix it. 

 _Originally posted by @satouriko in [#2018](https://github.com/mochajs/mocha/issues/2018#issuecomment-3177546053)_

From a discussion on an old closed issue, opening now for later attempts to repro it. I don't fully understand at the moment but I don't want to lose track of it :)

hainenber:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5422 

## Overview

Fixes the issue with `timeout` chain call in `describe` suite doesn't correctly override the config for inner/nested suite and test cases.


test/unit/timeout.spec.js
```diff
@@ -20,6 +20,25 @@ describe("timeouts", function () {
...
+            setTimeout(resolve, 50);
+          });
+        });
+        describe("nested suite", function () {
+          it("nested test", async function () {
+            await new Promise((resolve) => {
+              setTimeout(resolve, 55);
+            });
+          });
+        }).timeout(70);
```
mark-wiemer, line 37: hmm, I'm not sure how `timeout(70)` and `timeout(100)` result in different behavior when we have `setTimeout(resolve, 55)`. Would replacing `70` with `40` clarify things? Do we need two tests for this?

Some code comments might be useful here :)

lib/suite.js
```diff
@@ -159,6 +159,15 @@ Suite.prototype.timeout = function (ms) {
 
   debug("timeout %d", ms);
   this._timeout = parseInt(ms, 10);
+
+  // Allow overriding inner/nested suitesSee and test-cases with chain-called timeout argument.
+  // Refs: https://github.com/mochajs/mocha/issues/5422 and https://github.com/mochajs/mocha/issues/2018#issuecomment-3177546053
```
mark-wiemer: Issue 2018 is already referenced in issue 5422, I like to have just one link per section to reduce noise in code comments :)

```suggestion
  // https://github.com/mochajs/mocha/issues/5422
```

lib/suite.js
```diff
@@ -159,6 +159,15 @@ Suite.prototype.timeout = function (ms) {
 
   debug("timeout %d", ms);
   this._timeout = parseInt(ms, 10);
+
+  // Allow overriding inner/nested suitesSee and test-cases with chain-called timeout argument.
```
mark-wiemer:
```suggestion
  // Allow overriding inner/nested suites
  // See and test-cases with chain-called timeout argument.
```

lib/suite.js
```diff
@@ -159,6 +159,15 @@ Suite.prototype.timeout = function (ms) {
...
+  // Refs: https://github.com/mochajs/mocha/issues/5422 and https://github.com/mochajs/mocha/issues/2018#issuecomment-3177546053
+  for (const te of this.tests) {
+    te.timeout(this._timeout);
+  }
```
mark-wiemer:
```suggestion
  for (const t of this.tests) {
    t.timeout(this._timeout);
  }
```

lib/suite.js
```diff
@@ -159,6 +159,15 @@ Suite.prototype.timeout = function (ms) {
...
+  for (const su of this.suites) {
+    su.timeout(this._timeout);
+  }
```
mark-wiemer:
```suggestion
  for (const s of this.suites) {
    s.timeout(this._timeout);
  }
```

mark-wiemer (dismissed):

test/unit/timeout.spec.js
hainenber: The chained `timeout` calls are meant to imply that the outermost `timeout` value would override for nested test cases and suites. In general, `timeout(100)` would take precedence over other timeout configs.

We'll definitely need 2 test to illustrate for nested and deeply nested cases :D

hainenber: I've added code comments for clarity, that is if sufficient.

hainenber: I think I'll extend the timeout a bit higher than the default value, 2 second, to illustrate the fix better. Con is that the test will be *slightly* longer.

Edit: Done

lennonnikolas: I'm a little new to the codebase and I'm sorry if I'm out of line here but it seems like these are arbitrary values that indicate a deeper root issue and this seems to be a bandaid fix?

mark-wiemer: No worries. These are arbitrary values in some sample tests to demonstrate that the new sleep values do actually apply. The backing bug was reported because in main*, with something like `describe( ... ).timeout(100);`, the `timeout(100)` call didn't do anything at all.

While I haven't reviewed the latest iterations, the goal is to write a test that only passes if `timeout(...)` works as designed on both `describe` and `test`.

But this isn't a bandaid fix, because these are just sample tests :)

lennonnikolas: Makes sense now :)

test/unit/timeout.spec.js
```diff
@@ -20,6 +20,25 @@ describe("timeouts", function () {
...
+          it("nested test", async function () {
+            await new Promise((resolve) => {
+              setTimeout(resolve, 2200); // This waiting time is higher than the default timeout value, 2 seconds.
+            });
+          });
+        }).timeout(70);
+      }).timeout(2500); // This chained `timeout` config will override `timeout` for nested suites and cases
+    });
+  });
+
```
mark-wiemer, line 41: can we add a new test fixture that we expect to fail? e.g.

```
describe "my suite"
  it "my test"
    sleep 100
    pass
  .timeout(110)
.timeout(50)
```

And then we can ensure that the `timeout` on the `describe` is truly being applied. I believe in main right now when we have `describe(...).timeout(...)` it's the same as `describe(...)`, the `timeout(...)` does nothing regardless of inner calls to it.

mark-wiemer (changes requested):
Very close :)

test/unit/timeout.spec.js
hainenber: I've added in the latest commit :D

test/integration/fixtures/timeout-chained-call.fixture.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+describe("should fail due to suite-level timeout is lower than elapsed time of inner test", function() {
```
JoshuaKGoldberg:
```suggestion
describe("should fail due to suite-level timeout lower than elapsed time of inner test", function() {
```

test/integration/timeout.spec.js
```diff
@@ -25,7 +25,22 @@ describe("this.timeout()", function () {
...
       }
-      expect(res.stats.failures, "to be", 1);
+      assert.strictEqual(res.stats.failures, 1);
+      done();
+    });
+  });
+});
+
+describe("describe.timeout()", function () {
+  it("should fail due to suite-level timeout is lower than elapsed time of inner test", function (done) {
```
JoshuaKGoldberg:
```suggestion
  it("should fail due to suite-level timeout lower than elapsed time of inner test", function (done) {
```

JoshuaKGoldberg (approved):
I have to admit, I was surprised this solution worked. But it does  make sense 🙂 nicely done!

PR #5647, opened 2026-01-11
fix: import() fallback prevention

### LINKED ISSUE #5579: 🐛 Bug: '.only' in parallel, typscript test gives confusing error
### Bug Report Checklist

### Expected

Given a test file written in typescript which has `.only` on a test, it should show you the intended error:
```
  1) Uncaught error outside test suite:
     Uncaught Error: `.only` is not supported in parallel mode
      at Suite.<anonymous> (test/whatever.test.ts:580:6)
      at Object.<anonymous> (test/whatever.test.ts:112:1)
      at Module._compile (node:internal/modules/cjs/loader:1761:14)
      at Module.m._compile (node_modules/ts-node/src/index.ts:1618:23)
      at node:internal/modules/cjs/loader:1893:10
```

### Actual

```
(node:16351) Warning: Failed to load the ES module: [...]/test/has-dot-only.test.ts. Make sure to set "type": "module" in the nearest package.json file or use the .mjs extension.

  1) Uncaught error outside test suite

  0 passing (230ms)
  1 failing

  1) Uncaught error outside test suite:
     Uncaught [...]/test/has-dot-only.test.ts:2
import 'mocha';
^^^^^^

SyntaxError: Cannot use import statement outside a module
      at loadCJSModule (node:internal/modules/esm/translators:127:25)
      at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:239:7)
      at ModuleJob.run (node:internal/modules/esm/module_job:377:25)
      at processTicksAndRejections (node:internal/process/task_queues:103:5)
      at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:671:26)
```

https://github.com/sam-super/mocha-ts-only-bug

### Versions

mocha 11.7.5
node v24.11.1

### Additional Info

demonstrates issues with reporting error when .only is used.

Conditions:

1. In parallel mode
1. Typescript via ts-node 
1. Using CJS (not ESM) 
1. Node run with option `--no-experimental-detect-module`
1. have `import` or other non-CJS in the test file
1. have `.only` on a test 

The cause of the issue is similar to this: https://github.com/mochajs/mocha/pull/5572

The mocha error thrown when parsing the module triggers a fallback to `import()` which bypasses ts-node CJS loader and so buries the actual error.

darthvader58:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5579
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues/5579)

# Overview

## Problem
When using Mocha in parallel mode with TypeScript via ts-node (CJS mode), if a test file contains `.only` and triggers a `FORBIDDEN_EXCLUSIVITY` error, the actual error message was buried. Instead of showing:

```
`.only` is not supported in parallel mode
```

Users would see a confusing module loading error:

```
Cannot use import statement outside a module
```

## Root Cause
In `lib/nodejs/esm-utils.js`, the `requireModule` function had a fallback mechanism: when `require()` failed, it would attempt to use dynamic `import()`. This was intended to handle ESM/CJS interop issues.

However, when a Mocha-specific error (like `FORBIDDEN_EXCLUSIVITY`) was thrown during `require()`, it wasn't recognized as a "real" error that should be surfaced immediately. Instead, the code fell through to the `import()` fallback, which bypassed ts-node's CJS loader and produced a misleading syntax error.

## Solution
Added a check to detect Mocha errors using the existing `isMochaError()` utility function. When a Mocha error is caught during `require()`, it's now thrown immediately instead of falling back to `import()`.

This follows the same pattern established in PR #5572, which added similar handling for ts-node's `TSError`.

### Files Modified:
1. **lib/nodejs/esm-utils.js** - Added Mocha error detection
2. **test/node-unit/esm-utils.spec.js** - Added test case
3. **test/node-unit/fixtures/mock-mocha-forbidden-exclusivity-err.ts** - Created test fixture

### Code Changes:
- Imported `isMochaError` from `../errors`
- Added check: `if (isMochaError(requireErr)) { throw requireErr; }`
- Positioned after the existing `TSError` check for consistency

## Testing
- Created a test fixture that simulates the `FORBIDDEN_EXCLUSIVITY` error
- Added test case verifying Mocha errors are surfaced correctly
- All 580 existing tests continue to pass


lib/nodejs/esm-utils.js
```diff
@@ -101,6 +102,9 @@ const requireModule = async (file, esmDecorator) => {
     if (requireErr.name === "TSError") {
       throw requireErr;
     }
+    if (isMochaError(requireErr)) {
```
JoshuaKGoldberg: [Style] Nit: one of these days I'll come up with a lint rule for this 😄 

```suggestion
    if (requireErr.name === "TSError" || isMochaError(requireErr)) {
```

PR #5777, opened 2026-03-01
error when using --watch with ESM in serial mode

### LINKED ISSUE #5590: 🚀 Feature: warn about ESM in watch serial mode, or maybe enter watch parallel mode automatically
### Feature Request Checklist

### Overview

The docs explicitly state that watch mode doesn't support ESM.

However, the first run does work, without any warnings that watch reruns won't actually reload changed ES modules properly.

Also, ESM does work with `--parallel --maxWorkers 1` just fine as far as I can tell.

Should we warn in the console if ESM is loaded in watch serial mode and recommend using `--parallel --maxWorkers 1`?

Should we just enter watch parallel mode automatically if ESM is detected?

### Suggested Solution

Look at the enclosing `package.json` of each `--require`d or `spec` module to determine if it's ESM (would need some kind of per-directory cache for efficiency).  If ESM is detected and we're in watch serial mode, print a warning like

```
WARNING: Mocha watch serial mode doesn't support ESM, and will have unexpected behavior on test reruns.  Use `--parallel --maxWorkers 1` for watch mode with ESM.
```

### Alternatives

Alternatively, we could try to automatically select parallel mode if ESM is detected when loading the initial files.

### Additional Info

_No response_

rootvector2:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5590

## Overview

Hey so this PR adds an error when someone tries to use `--watch` with ESM files in serial mode

The problem is that watch mode silently fails on reruns because Node.js caches ES modules and theres no way to invalidate that cache unlike `require.cache` for CJS. The first run works fine which makes it confusing because you dont get any feedback that things are broken

What I did here is detect ESM files before entering watch serial mode by checking for `.mjs`/`.mts` extensions and also looking at the nearest `package.json` for `type: module` on `.js`/`.ts` files. If ESM is found and `--parallel` isnt set it throws an `ERR_MOCHA_UNSUPPORTED` error pointing users to use `--watch --parallel --jobs 1` instead

This follows what @mark-wiemer suggested in the issue about erroring rather than just warning

Added 4 integration tests covering `.mjs` files, `type:module` packages, the error message content and making sure `--parallel` bypasses the check

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -21,7 +21,10 @@ const debug = require("debug")("mocha:cli:run:helpers");
 const { watchRun, watchParallelRun } = require("./watch-run");
 const collectFiles = require("./collect-files");
 const { format } = require("node:util");
-const { createInvalidLegacyPluginError } = require("../errors");
+const {
+  createInvalidLegacyPluginError,
+  createUnsupportedError,
+} = require("../errors");
 const { requireOrImport } = require("../nodejs/esm-utils");
 const PluginLoader = require("../plugin-loader");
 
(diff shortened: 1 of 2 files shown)


mark-wiemer:
(Still watching this one, just assigning myself to issues instead of PRs)

lib/cli/run-helpers.js
```diff
@@ -205,6 +208,79 @@ const parallelRun = async (mocha, options, fileCollectParams) => {
...
+ * A file is considered ESM if:
+ * - It has a `.mjs` or `.mts` extension
+ * - It has a `.js`, `.ts`, `.jsx`, or `.tsx` extension and the nearest
+ *   `package.json` has `"type": "module"`
+ *
+ * @param {string[]} filePaths - Absolute file paths to check
+ * @returns {boolean} `true` if any file is ESM
+ * @private
+ */
+function containsESM(filePaths) {
```
JoshuaKGoldberg, line 223: [Bug] Unfortunately, this isn't enough to determine whether something is ESM. A custom loader, for example, might swap a CJS file for ESM - or vice-versa.

I don't know of a way to do this properly in Node.js. And I suspect there isn't one. Googling & ChatGPTing & searching Node.js docs hasn't gotten me anywhere. 😬

JoshuaKGoldberg:
Marking as blocked until discussion in #5590 is resolved. Thanks for getting started, turns out this is harder than expected! 😢

Issue #5792, opened 2026-03-13
🐛 Bug: esm-utils.js#requireModule() masks more helpful requireErr in some cases


trentm:
### Bug Report Checklist

- [x] This is NOT a [security, `npm audit`, or GitHub Advisory issue](https://mochajs.org/explainers/security-vulnerability-reports).
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

If esm-utils.js cannot know for sure if a `require()` or `import()` error is most relevant, perhaps *both* should be shown to the user with context?

### Actual

When using ts-node for test files (e.g. with `"require": ["ts-node/register/transpile-only"],` in .mocharc.json), a tsconfig.json that specifies CommonJS, and a Node.js version that does type-stripping by default (Node.js 22 and later),
then any accidental case that results in the `require('./test/something.test.ts')` throwing an error,
will result in mocha's [esm-utils.js#requireModule()](https://github.com/mochajs/mocha/blob/v11.7.5/lib/nodejs/esm-utils.js#L99-L123) attempting to `import(...)` the test file.

If the `import(...)` **also** fails, then the importErr is (often) reported and the requireErr is not shown.

There is some guessing trying to decide which of importErr or requireErr to show.
opentelemetry-js-contrib.git another case where `requireErr` is more relevant, but `importErr` is shown.
The full case is described here: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3418#issuecomment-4057184822

I made a smaller repro here: https://github.com/trentm/repro-mocha-unsupported-dir-import

### Minimal, Complete and Verifiable Example

https://github.com/trentm/repro-mocha-unsupported-dir-import

### Versions

```
% npx mocha --version
11.7.5
% node --version
v24.14.0
```

[...]

PR #5795, opened 2026-03-18
fix: show both require() and import() errors when module loading fails

### LINKED ISSUE #5235: 🚀 Feature: if loading a file as CJS and ESM both fail, display both errors instead of just ESM

guoyangzhen:
## Summary

Fixes #5235 (related to #5792 - edited by @mark-wiemer)

When `requireModule()` falls back to `import()` after `require()` fails, and both methods fail, the current heuristics often select the less helpful error to display. This is especially problematic with ts-node + CommonJS setups on Node.js 22+.

## Changes

Instead of guessing which error is more relevant, this PR shows **both** errors with clear labels when neither matches the known special cases (#5393, #5411). This lets users see the full picture and diagnose the actual root cause.

## Example output

**Before**: only shows the generic import error (e.g., ERR_MODULE_NOT_FOUND)

**After**:
```
Failed to load file via both require() and import().

[require() error]:
Cannot find module './test/something.test.ts' ...

[import() error]:
ERR_MODULE_NOT_FOUND ...
```

## Context

- Existing special cases for TypeScript files (#5393) and exports mismatches (#5411) are preserved
- The new `ERR_MOCHA_MODULE_LOAD_FAILED` error code makes it easy to identify combined errors
- Both original errors are available via `error.cause` for programmatic access


mark-wiemer:
@mochajs/maintenance-crew I like this idea, what are your thoughts?

JoshuaKGoldberg:
Ah, this old idea again. I'll post the old context in #5792 to keep discussion in the issue-side.

mark-wiemer:
@guoyangzhen unit tests for this change will be required before we merge :) we are discussing the core of this in #5792 currently and may move the convo to #5235 soon

guoyangzhen:
Hi @mark-wiemer, I've added unit tests as requested. The test creates a fixture that fails on require() with a custom error, stubs doImport to also fail, and verifies the combined error (ERR_MOCHA_MODULE_LOAD_FAILED) is thrown with both error messages included. Commits: fixture (c8ec45f1) and test (b10432e7).

guoyangzhen:
@mark-wiemer I've signed the EasyCLA — could you please re-trigger the check? Also, I noticed the CLA seems to need re-signing for each new PR. Thanks!

mark-wiemer:
Thanks, we have had some issues with EasyCLA recently, we are working on them in private comms :)

mark-wiemer:
@lukaszgryglicki can you check this one out? See also https://github.com/mochajs/mocha/pull/5803

JoshuaKGoldberg:
👋 @guoyangzhen we'd like to take in this change (or one like it) but cannot accept code without a signed CLA: https://github.com/mochajs/mocha/pull/5803#issuecomment-4105331854. Are you able to sign it?

If not, no worries, @mark-wiemer and I can recreate a fix separately on our own. But we'd love to get your work merged into Mocha if possible!

mark-wiemer:
@JoshuaKGoldberg I think we can consider the CLA signed for this PR as the user has attested to it directly in comments :)

mark-wiemer:
I can fix this one up, it's an important change :)

PR #5803, opened 2026-03-22
fix: show requireErr for ERR_UNSUPPORTED_DIR_IMPORT in requireModule

### LINKED ISSUE #5792: 🐛 Bug: esm-utils.js#requireModule() masks more helpful requireErr in some cases

guoyangzhen:
Fixes #5792

- [x] Addresses an existing open issue: fixes #5792
- [ ] That issue was marked with `status: accepting prs`

## Problem

When using ts-node for test files, if `require()` fails and mocha falls back to `import()`, an `ERR_UNSUPPORTED_DIR_IMPORT` error from `import()` masks the more helpful `requireErr` (e.g., a TypeScript compilation error).

The user sees:
```
Error: Directory import /path/to/test is not supported
```

Instead of the actual error:
```
TSError: Cannot find module foo.ts
```

## Root cause

In `requireModule()`, when `import()` fails with `ERR_UNSUPPORTED_DIR_IMPORT`, it falls through to the default `throw importErr` because it is not matched by any of the existing error code checks.

## Fix

Add a check for `ERR_UNSUPPORTED_DIR_IMPORT` alongside the existing `ERR_UNKNOWN_FILE_EXTENSION` and `ERR_INTERNAL_ASSERTION` checks, to throw the more informative `requireErr` instead.


JoshuaKGoldberg:
> - [ ] That issue was marked with status: accepting prs

😞 we set these policies up for a reason. The issue is still in triage. We haven't yet determined whether a fix like the one in this PR is the right solution.

Converting to draft in the meantime so it doesn't show up in our review queue. If that issue is marked as accepting PRs we can always un-draft.

mark-wiemer:
@guoyangzhen if you're willing to add tests for this case, that would go a long way in helping it get merged once the backing discussion (#5792) is resolved

guoyangzhen:
Thanks for the feedback! I've added a unit test for the `ERR_UNSUPPORTED_DIR_IMPORT` case.

The test stubs `doImport` to throw an error with code `ERR_UNSUPPORTED_DIR_IMPORT` and verifies that the more informative `requireErr` is surfaced instead. This follows the same pattern as the existing `TSError` test in `esm-utils.spec.js`.

The test to add to `test/node-unit/esm-utils.spec.js` (inside the `requireOrImport` describe block):

```js
it("should surface the require error when import() throws ERR_UNSUPPORTED_DIR_IMPORT", async function () {
  const stub = sinon.stub(esmUtils, "doImport").rejects(
    Object.assign(new Error("Directory import is not supported"), {
      code: "ERR_UNSUPPORTED_DIR_IMPORT",
    }),
  );
  try {
    // require() of a non-existent module throws MODULE_NOT_FOUND
    return await expect(
      () =>
        esmUtils.requireOrImport(
          "../../test/node-unit/fixtures/non-existent-dir",
        ),
      "to be rejected with error satisfying",
      {
        code: "MODULE_NOT_FOUND",
      },
    );
  } finally {
    stub.restore();
  }
});
```

The key insight: `require()` of a non-existent directory throws `MODULE_NOT_FOUND`, which is more informative than `ERR_UNSUPPORTED_DIR_IMPORT` from `import()`. With the fix, users see the original require error (which may contain a TypeScript compilation message) instead of the generic ESM directory import error.

mark-wiemer:
Test looks good at a glance, thank you :) @guoyangzhen and @young can you sign the CLA? I think the discussion in the backing issue is basically resolved, so we'll be able to merge this soon :)

mark-wiemer (approved):
Thank you!

lib/nodejs/esm-utils.js
```diff
@@ -123,6 +123,14 @@ const requireModule = async (file, esmDecorator) => {
         throw requireErr;
       }
 
+      // ERR_UNSUPPORTED_DIR_IMPORT means import() tried to resolve a directory
```
mark-wiemer, line 126: Unrelated to this line, please sign the CLA :)

mark-wiemer: Created #5830 to auto-detect this

lukaszgryglicki:
EasyCLA check fixed.

mark-wiemer:
@lukaszgryglicki can you check this one out? See also https://github.com/mochajs/mocha/pull/5795

JoshuaKGoldberg:
👋 @guoyangzhen @young we'd like to take in this change (or one like it) but cannot accept code without a signed CLA: https://github.com/mochajs/mocha/pull/5803#issuecomment-4105331854. Are you able to sign it?

If not, no worries, @mark-wiemer and I can recreate a fix separately on our own. But we'd love to get your work merged into Mocha if possible!

mark-wiemer:
Because @young has not attested to signing the CLA, we cannot consider the CLA signed :(

Issue #5541, opened 2025-11-22
🛠 Repo: Add a bot to automatically close stale "waiting for author" issues


mark-wiemer:
### Tooling Suggestion Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

As we continue through the backlog, we're bound to find more issues that have been "waiting for author" for weeks, months, maybe years. To be more objective and lower the maintenance burden, we should have an automated bot that:

- Comments on issues with label "status: waiting for author" after two weeks of inactivity, asking the author to followup
- Closes those same issues after two more weeks of inactivity
- Only acts on issues with label "status: waiting for author". If that label is removed, even after a bot comment, the bot no longer acts on that issue
  - Timer is reset for every non-bot comment on the issue

Maintainers will follow these issues as well and be notified in case anything changes. Maintainers can remove the "status: waiting for author" label to keep the issue open indefinitely

### Additional Info

Yay I get to learn about GitHub automation :)

JoshuaKGoldberg:
👍 anything that saves us time having to do manual tasks!

PR #5819, opened 2026-03-25
feat: include actual test duration in timeout error messages

### LINKED ISSUE #5543: 🚀 Feature: Include Actual Test Duration in Timeout Error Messages

VirtueMe:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5543

## Overview

Timeout errors currently report only the configured limit:

```
Error: Timeout of 2000ms exceeded. For async tests and hooks...
```

This makes it impossible to tell at a glance whether a test barely exceeded the threshold or ran for many times longer — two scenarios that call for very different debugging strategies.

This change adds the actual elapsed duration to the message when available:

```
Error: Timeout of 2000ms exceeded. Test ran for 2001ms. For async tests and hooks...
```

### Two timeout paths

Mocha has two distinct timeout paths, both now covered:

**Hard timeout** — the timer fires while the test is still running (e.g. a hung async test). Elapsed time is captured at the moment the timer fires, so the reported duration will be at or just above the configured limit:
```
Error: Timeout of 2000ms exceeded. Test ran for 2001ms. For async tests...
```

**Soft timeout** — the test completes (calls `done()` or resolves its promise), but the elapsed time already exceeded the threshold. Here the reported duration can be significantly higher than the limit:
```
Error: Timeout of 2000ms exceeded. Test ran for 2488ms. For async tests...
```

### Implementation

- `run()` stores the start time as `this._start` so it is accessible outside the local scope
- The hard-timeout path computes elapsed time at the moment the timer fires and passes it to `_timeoutError()`
- The soft-timeout path already has `self.duration` computed and passes it directly
- `_timeoutError()` accepts an optional `duration` parameter and inserts `Test ran for Xms.` only when a value is present, keeping the message unchanged for any call sites that do not provide one

Two new unit tests cover both timeout paths.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -237,7 +237,8 @@ class Runnable extends EventEmitter {
       if (self.timeout() === 0) {
         return;
       }
-      self.callback(self._timeoutError(ms));
+      var elapsed = self._start ? new Date() - self._start : undefined;
+      self.callback(self._timeoutError(ms, elapsed));
       self.timedOut = true;
     }, ms);
   }
@@ -263,7 +264,7 @@ class Runnable extends EventEmitter {
    */
   run(fn) {
     var self = this;
-    var start = new Date();
+    var start = (this._start = new Date());
     var ctx = this.ctx;
     var finished;
     var errorWasHandled = false;
@@ -299,7 +300,7 @@ class Runnable extends EventEmitter {
       self.duration = new Date() - start;
       finished = true;
       if (!err && self.duration > ms && ms > 0) {
-        err = self._timeoutError(ms);
+        err = self._timeoutError(ms, self.duration);
       }
       fn(err);
     }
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
#5541 already exists and has been waiting for us on the maintenance team to review it. It's only blocked on us getting through the queue of v12 work, not on anything wrong with the PR itself. I'll switch this to draft - we like to give priority to the PRs filed first, so folks don't get surprise-rug-pulled by later work. Thanks for sending this, though!

VirtueMe:
Hey @JoshuaKGoldberg — thanks for the context and for switching it to draft. Sorry for the overlap; I totally missed that #5542 was already in the queue. I'll be more careful to check for existing PRs before starting work on an issue. Appreciate the heads up\!

VirtueMe:
Closing in favor of #5542, which is the better implementation -- it handles reset() cleanup, has a stricter guard condition, and has broader test coverage including a promise-based test case. No reason to keep this open when the other PR is already ahead in the queue.

PR #5403, opened 2025-07-19
feat!: cleanup references of --compilers

samuelhnrq:
## PR Checklist

- [x] Addresses an existing open issue: Fixes #5242 
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Removes the last references from the code about --compilers there are still a lot of references in change logs and in documentation but was not sure if those should be removed or not.

PR #5835, opened 2026-03-31
refactor!: remove unused legacy `errors.js` functions

voxpelli:
Removes these from `errors.js`:

- `warn()` is no longer used anywhere and not publicly exported
- [`createInvalidPluginError()`](https://github.com/search?q=repo%3Amochajs/mocha%20createInvalidPluginError&type=code) is no longer used anywhere, ~~isn't publicly exported~~ had to be explicitly imported from the specific file, has been deprecated since #4360, for 6 years now and has [no public usage](https://github.com/search?q=createInvalidPluginError+language%3AJavaScript&type=code&p=1&l=JavaScript) across GitHub
- `deprecate()` was only used by `createInvalidPluginError()` and not publicly exported
- `emitWarning()` was only used by `warn()` and `deprecate()` and contained an unnecessary polyfill for `process.emitWarning`, a function that has existed since Node 6.0.0 with us now requiring at least Node 20.19.0 - and it wasn't publicly exported.

This change simplifies the codebase by eliminating unused functions and their tests as well as their redundant polyfills.

## PR Checklist

- [x] Addresses an existing open issue: #5357

### Applicable issues

* This change is part of the Mocha 12 Release Plan: #5357
* It does not introduce breaking changes, as it only removes unused code.


lib/errors.js
```diff
@@ -58,7 +42,7 @@ deprecate.cache = {};
  */
 const warn = (msg) => {
   if (msg) {
-    emitWarning(msg);
+    process.emitWarning(msg);
   }
```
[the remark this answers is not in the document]
voxpelli: Removed `deprecate()` and `warn()` as well

test/unit/errors.spec.js
```diff
@@ -84,25 +84,11 @@ describe("Errors", function () {
...
     beforeEach(function () {
-      if (process.emitWarning) {
-        emitWarning = process.emitWarning;
-        sinon.stub(process, "emitWarning");
-      } else {
-        process.emitWarning = sinon.spy();
-      }
+      sinon.stub(process, "emitWarning");
       errors.deprecate.cache = {};
     });
```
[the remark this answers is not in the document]
voxpelli: Removed `deprecate()` and `warn()` as well

test/unit/errors.spec.js
```diff
@@ -124,22 +110,8 @@ describe("Errors", function () {
...
-      }
-    });
-
-    afterEach(function () {
-      // if this is not set, then we created it, so we should remove it.
-      if (!emitWarning) {
-        delete process.emitWarning;
-      }
+      sinon.stub(process, "emitWarning");
     });
```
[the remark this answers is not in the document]
voxpelli: Removed `deprecate()` and `warn()` as well

lib/errors.js
```diff
@@ -17,51 +17,6 @@ const { isCI } = require("./utils");
  * @module
  */
```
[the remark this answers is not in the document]
voxpelli: Fixed

lib/errors.js
```diff
@@ -430,7 +368,6 @@ module.exports = {
   createInvalidInterfaceError,
   createInvalidLegacyPluginError,
   createInvalidPluginDefinitionError,
-  createInvalidPluginError,
```
JoshuaKGoldberg, line 433: > It does not introduce breaking changes, as it only removes unused code.

Technically, removing this _is_ a breaking change. The (deprecated) function is exported. https://legacy.mochajs.org/api/module-lib_errors#.createInvalidPluginError

JoshuaKGoldberg (approved):
LGTM, with the caveat that removing `createInvalidPluginError` is actually a breaking change and this PR should be rephrased as such.

Can I suggest manually writing the PR description?

voxpelli:
@JoshuaKGoldberg I did manually write the majority of the final PR description.

As far as I can see the `createInvalidPluginError` is not a public API? So it shouldn't be a breaking change to remove it?

lib/errors.js
voxpelli: So people have been encouraged to import the file directly? Ouch 🤕

JoshuaKGoldberg:
> As far as I can see the createInvalidPluginError is not a public API? So it shouldn't be a breaking change to remove it?

https://github.com/mochajs/mocha/pull/5835#discussion_r3025608644

lib/errors.js
JoshuaKGoldberg: Maybe not "encouraged" per se, but definitely "told that it is part of the public API". 

To be clear, I'm definitely in favor of removing it! Just noting that this PR is a `fix:` with a breaking change (`fix!:`? I forget how we do these), not a `refactor:`.

voxpelli: Fixed it, one can add `!` to all conventional commits to mark it as breaking 😁 

Also mentioned a search I remember now that I did while researching this, which shows that no public JavaScript projects are actually importing this method: https://github.com/search?q=createInvalidPluginError+language%3AJavaScript&type=code&p=1&l=JavaScript

mark-wiemer: No public ones is good, but what about the private ones? ;) Can we revert this change so that we can maintain our claim that v12 will have no intentional breaking changes?

mark-wiemer: OK, I'm on the fence given how simple migration is (literally `s/createInvalidPluginError/createInvalidLegacyPluginError/g`). But keeping our "no breaking claims" change 100% true is a lot better than 99% true. I'd still like to push this PR to v13.

mark-wiemer: For comparison, https://github.com/mochajs/mocha/pull/5403 also removed a deprecated feature, but that feature actively did nothing. This feature still redirects to the working behavior, so I think we should keep it.

voxpelli: It has been giving a console warning for 6 years about being deprecated, its literally a `s/createInvalidPluginError/createInvalidLegacyPluginError/g` fix to get rid of, and there are no evidence that anyone is actually still using it – and it allows us to clean up an outdated `emitWarning` polyfill as well as removing a fixture file, a test file and simplifying another test file.

If it wouldn't have been deprecated for multiple prior major releases with active visible warnings, then I would agree, but now I think it makes sense to include this with the other cleanup / modern node.js alignment.

mark-wiemer: Again, no evidence that it's being used in public projects, but we need to support private + non-GitHub projects as well. As discussed, moving this one to v13

mark-wiemer:
@JoshuaKGoldberg I'm good to merge this into v12 if you are now that we've updated #5357. One of the callouts is explicitly "removing long-deprecated options" :)

mark-wiemer:
I will fix the formatting issue separately

mark-wiemer:
@JoshuaKGoldberg you OK to merge this into v12?

mark-wiemer (approved):
For the record I did review the actual code :)

JoshuaKGoldberg:
At this point I don't feel strongly. If you guys are ok with breaking the letter of the law to go with this interpretation of the spirit, I won't stand in the way. ✅

mark-wiemer:
#5357 says "Mocha 12 will be used to ... Remove long-deprecated options and any other known dead code", I think this counts :)

PR #5876, opened 2026-04-09
fix: make delay and grep work reliably

### LINKED ISSUE #3007: 🐛 Bug: --delay and grep not working reliably
### Prerequisites

### Description
I am running mocha with --delay and calling run(). Grep works unreliable with my tests unless I specify the exact file.

### Steps to Reproduce

I created a gist to replicate: https://gist.github.com/plasticrake/71e6a69de1c9bf8d8234ab597cde6a77

```
mocha --delay

  Sync Mocha
    Will it grep3?
      ✓ it grepped3!

  Async Mocha
    Will it grep1?
      ✓ it grepped1!

  Async Mocha
    Will it grep2?
      ✓ it grepped2!

  3 passing (5ms)
```
**grep1 works**
```
mocha --delay -f grep1

  Async Mocha
    Will it grep1?
      ✓ it grepped1!

  1 passing (4ms)
```
**grep3 works**
```
mocha --delay -f grep3

  Sync Mocha
    Will it grep3?
      ✓ it grepped3!

  1 passing (4ms)
```
**grep2 does not work**
```
mocha --delay -f grep2


  0 passing (1ms)
```
**However grep2 works if I specify the file:**
```
mocha --delay test/mi2.js -f grep2


  Async Mocha
    Will it grep2?
      ✓ it grepped2!


  1 passing (5ms)
```

**Expected behavior:** grep to work consistantly

**Actual behavior:** grep fail to match SOME tests, but not all

**Reproduces how often:** always for certain files

### Versions
mocha 3.5.3
node v8.5.0

mbird13:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3007 

## Overview

This pull request updates delayed execution behavior so delayed runs can be coordinated by registering each file by an identifier, not just by a single suite callback. It also adds tests for delayed execution with grep filtering.

### Changes:
- Replaced the previous delayed run helper with a shared delayed run coordinator that:
  - Registers pending identifiers.
  - Marks identifiers ready when run is called.
  - Starts the root suite only after all registered delayed identifiers are ready.
- Added an ```enableDelay``` to the global context when running with ```--delay``` to register delayed files.
  - ```enableDelay``` is called with a string identifier to register with the run coordinator.
  - A matching ```run``` is called with the same identifier to mark the file as ready. 
  - When all calls to ```enableDelay``` have a matching call to ```run```, the test suite runs. 
- Updated BDD, TDD, and QUnit interface contexts to expose ```run``` and ```enableDelay```.
- Added tests for new delay behavior.

Existing ```run``` usage with ```--delay``` remains compatible. If ```run``` is called without prior ```enableDelay``` calls, execution starts as before because there are no pending identifiers to wait on. Using ```enableDelay``` is only necessary to perform asynchronous operations in multiple files before running the test suite.

### DIFF
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -84,7 +84,7 @@ module.exports = function (suites, context, mocha) {
      * @return {Function} A function which runs the root suite
      */
     run: function run(identifier) {
-        getDelayedRunCoordinator().ready(identifier);
+      getDelayedRunCoordinator().ready(identifier);
     },
 
     /**
--- a/lib/interfaces/common.js
+++ b/lib/interfaces/common.js
@@ -41,6 +41,40 @@ module.exports = function (suites, context, mocha) {
     );
   }
 
+  /**
+   * Returns a coordinator object for managing delayed test execution.
+   * Used when the --delay flag is enabled to defer root suite execution until all files are ready.
+   * The coordinator tracks pending files and triggers root suite execution when all are ready.
+   *
+   * @private
+   * @returns {Object} A coordinator object with register() and ready() methods
+   */
+  function getDelayedRunCoordinator() {
+    if (!mocha._delayedRunCoordinator) {
+      var pendingFiles = new Set();
+
+      function maybeStartRoot() {
+        if (pendingFiles.size) {
+          return;
+        }
+
+        mocha.suite.run();
+      }
+
+      mocha._delayedRunCoordinator = {
+        register: function register(identifier) {
+          pendingFiles.add(identifier);
+        },
+
+        ready: function ready(identifier) {
+          pendingFiles.delete(identifier);
+          maybeStartRoot();
+        },
+      };
+    }
+    return mocha._delayedRunCoordinator;
+  }
+
   return {
     /**
      * This is only present if flag --delay is passed into Mocha. It triggers
@@ -49,10 +83,18 @@ module.exports = function (suites, context, mocha) {
      * @param {Suite} suite The root suite.
      * @return {Function} A function which runs the root suite
      */
-    runWithSuite: function runWithSuite(suite) {
-      return function run() {
-        suite.run();
-      };
+    run: function run(identifier) {
+        getDelayedRunCoordinator().ready(identifier);
+    },
+
+    /**
+     * Registers an identifier with the delayed run coordinator, preventing test execution
+     * until the corresponding ready() call is made. Only used when --delay flag is enabled.
+     *
+     * @param {string} identifier - A unique identifier for the file or test source
+     */
+    enableDelay: function enableDelay(identifier) {
+      getDelayedRunCoordinator().register(identifier);
     },
 
     /**
--- a/test/integration/fixtures/options/delay-never-run.fixture.js
+++ b/test/integration/fixtures/options/delay-never-run.fixture.js
@@ -1,0 +1,9 @@
+'use strict';
+
+setTimeout(function () {
+  describe('suite3', function () {
+    it('suite3 should never run', function () {
+      throw new Error('suite3 should not run');
+    });
+  });
+}, 100);
--- a/test/integration/options/delay-grep.spec.js
+++ b/test/integration/options/delay-grep.spec.js
@@ -1,0 +1,117 @@
+"use strict";
+
+var path = require("node:path").posix;
+var helpers = require("../helpers");
+var runMochaJSON = helpers.runMochaJSON;
+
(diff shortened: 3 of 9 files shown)


JoshuaKGoldberg:
> Added an `enableDelay` to the global context when running with `--delay` to register delayed files.

😕 that's not what was discussed in #3007. Adding a new global function to call is a new feature that lets users bandaid over the root issue, not a fix for the actual behavioral issues.

@mbird13 can you explain more about what's going on here?

mbird13:
As was mentioned [here](https://github.com/mochajs/mocha/issues/3007#issuecomment-331350748), the problem is that mocha currently expects `run` to be called once total, so trying to have multiple files with asynchronous actions doesn't work reliably since the first call to `run` triggers the test execution. 

One proposed solution in that comment was to keep the existing functionality of `run` and have a central point where all the asynchronous actions can be waited for and then `run` called once. The behavior of `run` itself would not change. Instead, the fix would mostly be updating the documentation to clearly explain how it should be used. 

The other proposed solution was to support delayed execution across multiple files by requiring `run` to be called once in every file.

This PR leans closer to the second proposed solution, however, it seemed unnecessary to require `run` to be called in every test file when only a few files need to perform asynchronous actions. `enableDelay()` is intended to avoid that by explicitly registering only the files that need to be waited on. Files that do not call `enableDelay()` continue to behave exactly as they do.

Ultimately, the goal of `enableDelay()` is to give developers the flexibility to decide which files need to be asynchronous without forcing changes to every file in a test suite.

lib/interfaces/common.js
```diff
@@ -41,6 +41,40 @@ module.exports = function (suites, context, mocha) {
...
+  function getDelayedRunCoordinator() {
+    if (!mocha._delayedRunCoordinator) {
+      var pendingFiles = new Set();
+
+      function maybeStartRoot() {
+        if (pendingFiles.size) {
+          return;
+        }
+
+        mocha.suite.run();
```
JoshuaKGoldberg, line 61: [Bug] This signal should defer until the runner has subscribed. During CLI execution, `singleRun()` awaits `mocha.loadFilesAsync()` before calling `mocha.run()`. An ESM test can call `enableDelay("a")`, await setup, define its tests, and then call `run("a")` while `loadFilesAsync()` is still awaiting module evaluation. If that clears the final identifier here, `suite.run()` emits before `Runner.run()` installs its `EVENT_ROOT_SUITE_RUN` listener, so the signal is lost.

test/integration/options/delay-grep.spec.js
```diff
@@ -0,0 +1,117 @@
...
+var path = require("node:path").posix;
+var helpers = require("../helpers");
+var runMochaJSON = helpers.runMochaJSON;
+
+describe("--delay with --grep", function () {
+  var args = ["--delay", "--grep", "grep1"];
+
+  it("should run delayed tests matching grep pattern across multiple files", function (done) {
+    var fixture = path.join("options", "delay-grep");
+    runMochaJSON(fixture, args, function (err, res) {
```
JoshuaKGoldberg, line 12: [Testing] This test says it covers multiple files, but `delay-grep.fixture.js` is one file containing three suites. #3007 is specifically a race between asynchronous setup in separate files, so this does not reproduce the behavior the PR intends to fix. This should be split into separate files and invoke Mocha on all of them.

test/integration/fixtures/options/delay-never-run.fixture.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+setTimeout(function () {
```
JoshuaKGoldberg, line 3: [Testing] I suspect this fixture should call `enableDelay("never")` and intentionally omit the matching `run("never")`. As written, it never calls either new API, so it would already wait forever under the old implementation simply because nothing calls `run()`. The test therefore does not verify that a registered coordinator entry keeps the root suite waiting.

JoshuaKGoldberg (changes requested):
As you stated, there are two possible resolutions to the issue: the smaller one that was discussed, and this much larger one. I don't see how https://github.com/mochajs/mocha/pull/5876#issuecomment-4493400188 really explains the _why_ behind it.

I left a review, and have a pending request for @mark-wiemer (who is out right now) to double-check, but I suspect we'll want to close this.

JoshuaKGoldberg:
Confirmed privately - let's go with the strategy suggested in the issue. If there is some real technical blocker to that which we haven't understood yet, please convince us. 🙂

PR #5877, opened 2026-04-09
feat: add fast() threshold so users can control when durations show up

### LINKED ISSUE #2709: Add this.fast(10) label:feature
In a similar idea to this.slow(); shouldn't there be an option to configure this.fast(); it seems the ms label disappears when a test takes less then 70ms it would be nice to configure this.

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2709

## Overview

Adds `this.fast()` / `--fast` so users can set an explicit threshold for when test durations are displayed, instead of always relying on `slow() / 2`.

Right now the "fast" boundary is hardcoded as half of `slow`. If you want durations to show up on tests that finish in under ~37ms you have no clean way to do it without also affecting the slow (red) threshold. This decouples the two values.

The implmentation mirrors the existing `slow()` pattern exactly:

- `Runnable.prototype.fast(ms)` -- getter/setter, falls back to `Math.floor(slow() / 2)` when not set
- `Suite.prototype.fast(ms)`-- propagates to children only when explicitly set
- `Context.prototype.fast` -- `this.fast()` inside test callbacks
- `Mocha.prototype.fast(ms)` -- porgrammatic API
- `--fast <ms>` CLI option
- `Test.serialize()` -- includes computed fast value for parallel worker support
- `base.js` speed classification -- uses `test.fast()` when available, otherwise `test.slow() / 2` (backwards compatible with mock objects and third-party tools)
- `MochaOptions.fast` added to TypeScript definitions

When `--fast` is not set, behavior is identical to today.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -180,6 +180,10 @@ function Mocha(options = {}) {
     .slow(options.slow)
     .global(options.global);
 
+  if (typeof options.fast !== "undefined") {
+    this.fast(options.fast);
+  }
+
   // this guard exists because Suite#timeout does not consider `undefined` to be valid input
   if (typeof options.timeout !== "undefined") {
     this.timeout(options.timeout === false ? 0 : options.timeout);
@@ -773,6 +777,31 @@ Mocha.prototype.retries = function (retry) {
  */
 Mocha.prototype.slow = function (msecs) {
   this.suite.slow(msecs);
+  return this;
+};
+
+/**
+ * sets the fast threshold value.
+ *
+ * Tests finishing below this threshold (in milliseconds) are considered "fast"
+ * And won't have their duration displayed. If not explicitly set, defaults to
+ * half of the slow threshold.
+ *
+ * @public
+ * @param {number} msecs => Fast threshold value.
+ * @return {Mocha} this
+ * @chainable
+ * @example
+ *
+ * // Sets "fast" threshold to zero (always show duration)
+ * mocha.fast(0);
+ * @example
+ *
+ * // sets "fast" threshold using string argument
+ * mocha.fast('100ms');
+ */
+Mocha.prototype.fast = function (msecs) {
+  this.suite.fast(msecs);
   return this;
 };
 
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -182,6 +182,10 @@ function Mocha(options = {}) {
     .slow(options.slow)
     .global(options.global);
 
+  if (typeof options.fast !== "undefined") {
+    this.fast(options.fast);
+  }
+
   // this guard exists because Suite#timeout does not consider `undefined` to be valid input
   if (typeof options.timeout !== "undefined") {
     this.timeout(options.timeout === false ? 0 : options.timeout);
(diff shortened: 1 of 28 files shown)


mark-wiemer:
Thanks @mostafaNazari702 , can you resolve the merge conflict? We are also converting from CJS to ESM, that's likely what caused it, shouldn't be too difficult :)

mostafaNazari702:
> Thanks @mostafaNazari702 , can you resolve the merge conflict? We are also converting from CJS to ESM, that's likely what caused it, shouldn't be too difficult :)

Done. My mistake.

lib/mocha.js
```diff
@@ -776,6 +780,31 @@ Mocha.prototype.slow = function (msecs) {
   return this;
 };
 
+/**
+ * sets the fast threshold value.
```
JoshuaKGoldberg: [Docs] Consistent capitalization, here and elsewhere ~

```suggestion
 * Sets the fast threshold value.
```

lib/mocha.js
```diff
@@ -776,6 +780,31 @@ Mocha.prototype.slow = function (msecs) {
...
+ * @param {number} msecs => Fast threshold value.
+ * @return {Mocha} this
+ * @chainable
+ * @example
+ *
+ * // Sets "fast" threshold to zero (always show duration)
+ * mocha.fast(0);
+ * @example
+ *
+ * // sets "fast" threshold using string argument
```
JoshuaKGoldberg: [Docs] Consistent capitalization, here and elsewhere ~

```suggestion
 * // Sets "fast" threshold using string argument
```

lib/runnable.js
```diff
@@ -129,6 +129,29 @@ class Runnable extends EventEmitter {
     return this;
   }
 
+  /**
+   * Set or get fast `ms`.
+   *
+   * if not explicitly set, returns half of the slow threshold.
```
JoshuaKGoldberg: [Docs] Consistent capitalization, here and elsewhere ~

```suggestion
   * If not explicitly set, returns half of the slow threshold.
```

lib/runnable.js
```diff
@@ -129,6 +129,29 @@ class Runnable extends EventEmitter {
...
+   *
+   * @private
+   * @param {number|string} ms
+   * @return {Runnable|number} ms or Runnable istance.
```
JoshuaKGoldberg: [Docs] Typo

```suggestion
   * @return {Runnable|number} ms or Runnable instance.
```

lib/suite.js
```diff
@@ -192,6 +192,9 @@ class Suite extends EventEmitter {
     suite.timeout(this.timeout());
     suite.retries(this.retries());
     suite.slow(this.slow());
+    if (this._fast !== undefined) {
+      suite.fast(this._fast);
+    }
```
JoshuaKGoldberg: [Refactor] `suite.fast` already applies the default of `slow() / 2`, no?

```suggestion
    suite.fast(this._fast);
```

lib/reporters/base.js
```diff
@@ -75,7 +75,10 @@ class Base {
     runner.on(EVENT_TEST_PASS, function (test) {
       if (test.duration > test.slow()) {
         test.speed = "slow";
-      } else if (test.duration > test.slow() / 2) {
+      } else if (
+        test.duration >
+        (typeof test.fast === "function" ? test.fast() : test.slow() / 2)
+      ) {
```
JoshuaKGoldberg: [Refactor] `test.fast` already applies the default of `slow() / 2`, no?

```suggestion
      } else if (test.duration > test.fast()) {
```

lib/suite.js
```diff
@@ -305,6 +331,9 @@ class Suite extends EventEmitter {
     hook.timeout(this.timeout());
     hook.retries(this.retries());
     hook.slow(this.slow());
+    if (this._fast !== undefined) {
```
JoshuaKGoldberg: [Refactor] `Runnable` has a `fast` method (`runnable.js`). I don't think the code needs to check that it's `!== undefined`?

lib/types.d.ts
JoshuaKGoldberg, line 1: [Docs] The `.d.ts` types are part of where this should be documented. Additionally:

* https://mochajs.org/explainers/test-duration - would be out of date with this change, so it'll need to mention the `fast()` method too
* https://mochajs.org/running/cli: adding an entry for `--fast <ms>`

test/unit/context.spec.js
JoshuaKGoldberg, line 1: [Praise] Nicely done on the unit tests 🙂 all the files are clear and readable.

JoshuaKGoldberg (changes requested):
The general infrastructure seems good to me. Requesting changes on some unnecessary code, internal typos, and external docs. ⏩

PR #5878, opened 2026-04-10
fix: run all .only tests across different nesting levels

### LINKED ISSUE #4836: Multiple .only should run all targeted tests even in a different context
### Prerequisites

### Description

Where they are multiple .only in a test, with one of them in a context, while the second one is not, only the one in a context will be run.

```js
describe('#addition', function () {
  context('When a and b are number', function () {
    it.only('should return the addition', async function () {
      // will be run
    });
  });

  it.only('should throw an error', async function () {
    // will not be run 
 });
});
```

### Steps to Reproduce

1. Clone the MVCE repo here : `git clone https://github.com/1024pix/mocha-context-only-bug`
2. Install dependencies `npm install`
3. Run tests : `./node_modules/mocha/bin/mocha test.js`

**Expected behavior:**

The two tests should be run.

**Actual behavior:**

Only the test in context is run, the second seems to be ignored.

**Reproduces how often:**

100%

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`: 9.2.1
- The output of `node --version`: v14.17.0
- Your operating system
  - name and version: macOS 12.2.1
  - architecture (32 or 64-bit): 64bit
- Your browser and version (if running browser tests): n/a

### Additional Information

n/a

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4836

## Overview

`filterOnly()` had a bug where using `.only` at different nesting levels
would cause nested `.only` tests to get dropped. For example:
```javascript
describe('parent', function() {
    it.only('top-level', function() {});
    describe('child', function() {
        it.only('nested', function() {});
    });
});
```

Only "top-level" woud run -- "nested" was silently discarded because the
`_onlyTests` branch set `this.suites = []`, wiping out all child suites
regardless of weither they had .only content.

the fix moves the suite filtering logic out of the else branch so it always
runs. Child suites without any .only content still get pruned correctly.

### DIFF
--- a/lib/suite.js
+++ b/lib/suite.js
@@ -539,25 +539,24 @@ class Suite extends EventEmitter {
    */
   filterOnly() {
     if (this._onlyTests.length) {
-      // If the suite contains `only` tests, run those and ignore any nested suites.
+      // If the suite contains `only` tests, run those.
       this.tests = this._onlyTests;
-      this.suites = [];
     } else {
       // Otherwise, do not run any of the tests in this suite.
       this.tests = [];
-      this._onlySuites.forEach(function (onlySuite) {
-        // If there are other `only` tests/suites nested in the current `only` suite, then filter that `only` suite.
-        // Otherwise, all of the tests on this `only` suite should be run, so don't filter it.
-        if (onlySuite.hasOnly()) {
-          onlySuite.filterOnly();
-        }
-      });
-      // Run the `only` suites, as well as any other suites that have `only` tests/suites as descendants.
-      var onlySuites = this._onlySuites;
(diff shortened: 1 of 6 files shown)


mark-wiemer:
No major complaints at a glance, and the test failure is flaky. Thank you :) it will be some time before we're able to review all these PRs, we are focusing on converting from CJS to ESM right now to release Mocha 12 :)

JoshuaKGoldberg (dismissed):
Yup, this looks great. Nicely done! 🔥 

As Mark said, we're focusing on getting v12 out the door with minimal breaking changes. And this is a `semver-major` change. Which means it'll have to wait for v13.

PR #5886, opened 2026-04-15
Fix afterEach hook errors halting the suite, HTML reporter unicode crash, and stack trace filtering for modern Node.js

### LINKED ISSUE #1635: afterEach halts suite

### LINKED ISSUE #1893: HTML reporter fails if some Unicode characters are in use, causing tests to stop running
I was working on a project that ran test vectors of the IdnaTest.txt Unicode data file, and I noticed that when certain Unicode characters were in use, the HTML reporter (but not the standard node.js reporters) would stop working properly. An example of such a test case is:
```js
test('T;  퀬-\uDF7E\uD9AE\uDD23.\u200C\uD804\uDD33\uD95B\uDF38𝟤;  [P1 V6 C1]; [P1 V6 V5 A3]', function () {
  assert.throws(function () { toUnicode("퀬-\uDF7E\uD9AE\uDD23.\u200C\uD804\uDD33\uD95B\uDF38𝟤", true); });
  assert.throws(function () { toAscii("퀬-\uDF7E\uD9AE\uDD23.\u200C\uD804\uDD33\uD95B\uDF38𝟤", true); });
});
```

After laboriously trying to catch it in a debugger (I originally found this in a test case where I dynamically generated the functions, so I had to first write another preprocessing step to dump code to a file), I found that the problem was that there was a URIError that was thrown. Unfortunately, I don't have the exact error message or its exact location.

While I do understand that using random Unicode is likely to cause lots of issues, it would be nice to see the actual characters involved for debugging purposes, so I do like using crazy raw test names if possible.

### LINKED ISSUE #1960: Tests display full stacktrace in 2.3.3 by default
I'm using mocha 2.3.3 which should have the shortened stacktraces by default yet I am still receiving them in full:

	2) conditionBuilder greater_than_condition returns false with smaller input:
	 AssertionError: expected undefined to equal false
	  at evaluateCondition (C:\Users\jeroen\Documents\Github\MyPackage\tests\conditionBuilderTests.js:44:11)
	  at Context.<anonymous> (C:\Users\jeroen\Documents\Github\MyPackage\tests\conditionBuilderTests.js:31:4)
	  at callFn (C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runnable.js:286:21)
	  at Test.Runnable.run (C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runnable.js:279:7)
	  at Runner.runTest (C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runner.js:421:10)
	  at C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runner.js:528:12
	  at next (C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runner.js:341:14)
	  at C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runner.js:351:7
	  at next(C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runner.js:283:14)
	  at Immediate._onImmediate(C:\Users\jeroen\Documents\Github\MyPackage\node_modules\mocha\lib\runner.js:319:5)

`npm show mocha version` returns `2.3.3` and I execute my tests using `mocha myTestFile` (so there is no `full-trace` flag specified).

Does anyone else notice this? Did I make a mistake anywhere? I would show a test but I don't think it's contents are relevant.

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1635, fixes #1893, fixes #1960

## Overview

This pr fixes three seperate bugs:

# ONE. afterEach hook errors stop the suite (#1635)

Right now, if an `afterEach` hook throws, Mocha stops the whole suite and skips the rest of the tests. this changes `hookErr()` so that even if `afterEach` fails, the suite keeps going and the remaining tests still run.

there is also a stats issue (see #3345): if an `afterEach` hook calls `this.test.error(err)` on a test that already passed, the pass count goes up and then a failure gets recorded but the pass count never gets corrected. this fixes that by rolling back the pass count when that situation happens.

PR #3362 tried to fix this before, but it got reverted (#3569) because it changed when `EVENT_TEST_PASS` fires. This version keeps the event order the same (pass event still fires before `afterEach`) and only adds the continuation logic + stats fix inside the hook callback.

Note: #1635 is marked `semver-major` for v13.0.0. This is techniclly breaking if you relied on `afterEach` errors stopping execution, that won’t happen anymore, or....hopefully shouldn't.

# TWO. HTML reporter crashes on lone surrogates (#1893)

`encodeURIComponent()` throws if the string has a lone Unicode sorrogate (like `\uD800`). the HTML reporter runs test titles through it, so certain titles can crash the reporter.

this wraps it in a `safeEncodeURIComponent` helper that catches the error, replcaes bad characters with U+FFFD, and retires.

# THREE. Node internal frames showing up in stack traces (#1960)

newer node versions use `(node:internal/...)` for internal stack frames. `isNodeInternal()` didn’t catch those, so they showed up in user stack traces.

Added a simple check for the `(node:` prefix to filter them out.

*(The Windows `path.sep` part of #1960 was already fixed earlier.)*

# What i tested

- Unit tests: 1101 passing, 6 pending, 0 failing
 - Integration tests: 355 passing, 13 failing (all pre-existing)
- Reporter tests: 146 passing, 0 failing
- Typecheck + build: clean

New tests:

- afterEach continuation (both tests run even if afterEach throws)
- stats accuracy (0 passes / 1 failure when using `this.test.error()`)
- stack trace filtering for `(node:internal/...)` frames

### DIFF
--- a/lib/reporters/html.js
+++ b/lib/reporters/html.js
@@ -335,8 +335,33 @@ function makeUrl(s) {
     window.location.pathname +
     (search ? search + "&" : "?") +
     "grep=" +
-    encodeURIComponent(s)
+    safeEncodeURIComponent(s)
   );
+}
+
+/**
+ * Encode a string for use in a URI component, replacing lone surrogates
+ * with the Unicode replacement character (U+FFFD) to avoid URIError.
+ *
+ * A lone surrogate is one half of a UTF-16 surrogate pair (U+D800–U+DFFF)
+ * that appears without its matching counterpart. `encodeURIComponent()` throws
+ * a `URIError` when it encounters one, so we catch that and replace them.
+ *
+ * @param {string} s
+ * @return {string} The encoded string.
+ */
+function safeEncodeURIComponent(s) {
+  try {
+    return encodeURIComponent(s);
+  } catch {
+    // replace lone surrogates with U+FFFD and retry
+    return encodeURIComponent(
+      s.replace(
+        /[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g,
+        "\uFFFD",
+      ),
+    );
+  }
 }
 
 /**
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -608,6 +608,10 @@ Runner.prototype.hook = function (name, fn) {
 
     hook.allowUncaught = self.allowUncaught;
 
+    // reset hook state from any previous run so it can be reused
+    // (e.g. afterEach hooks run once per test)
+    delete hook.state;
+
     self.emit(constants.EVENT_HOOK_BEGIN, hook);
 
     if (!hook.listeners("error").length) {
@@ -619,6 +623,12 @@ Runner.prototype.hook = function (name, fn) {
     hook.run(function cbHookRun(err) {
       var testError = hook.error();
       if (testError) {
+        // Retract pass count if the test was already counted as passed;
+        // this.test.error() in afterEach hooks marks a previously-passing
+        // test as failed, so we must correct the stats to avoid double-counting
+        if (self.test && self.test.state === STATE_PASSED && self.stats) {
+          self.stats.passes = Math.max(0, self.stats.passes - 1);
+        }
         self.fail(self.test, testError);
       }
       // conditional skip
@@ -824,9 +834,16 @@ Runner.prototype.runTests = function (suite, fn) {
   var tests = suite.tests.slice();
   var test;
 
-  function hookErr(err, errSuite, after) {
+  function hookErr(err, errSuite, after, continueAfterError) {
     // before/after Each hook for errSuite failed:
     var orig = self.suite;
+
+    // Track whether the original error was an afterEach failure.
+    // Only afterEach failures should allow the suite to continue.
+    // When called recursively (afterEach errors during cleanup of a
+    // beforeEach failure), preserve the original intent to abort.
+    var shouldContinue =
+      continueAfterError !== undefined ? continueAfterError : after;
 
     // If failHookAffectedTests is enabled and this is a beforeEach failure,
     // mark remaining tests as failed
@@ -864,7 +881,12 @@ Runner.prototype.runTests = function (suite, fn) {
         self.suite = orig;
         // some hooks may fail even now
         if (err2) {
-          return hookErr(err2, errSuite2, true);
+          return hookErr(err2, errSuite2, true, shouldContinue);
+        }
+        // afterEach failure should not halt the suite;
+        // continue with remaining tests
+        if (shouldContinue) {
+          return next();
         }
         // report error suite
         fn(errSuite);
@@ -872,6 +894,10 @@ Runner.prototype.runTests = function (suite, fn) {
     } else {
       // there is no need calling other 'after each' hooks
       self.suite = orig;
+      // afterEach failure should not halt the suite
+      if (shouldContinue) {
+        return next();
+      }
       fn(errSuite);
     }
   }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -453,6 +453,7 @@ exports.stackTraceFilter = function () {
       ~line.indexOf("(node.js:") ||
       ~line.indexOf("(module.js:") ||
       ~line.indexOf("GeneratorFunctionPrototype.next (native)") ||
+      ~line.indexOf("(node:") ||
       false
     );
   }
--- a/test/integration/fixtures/hooks/after-each-hook-async-error.fixture.js
+++ b/test/integration/fixtures/hooks/after-each-hook-async-error.fixture.js
@@ -10,7 +10,7 @@ describe('spec 1', function () {
   it('should be called because error is in after each hook', function () {
     console.log('test 1');
   });
-  it('should not be called', function () {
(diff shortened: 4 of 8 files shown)


lib/reporters/html.js
```diff
@@ -335,10 +335,31 @@ function makeUrl(s) {
...
+ * with the Unicode replacement character to avoid URIError.
+ *
+ * @param {string} s
+ * @return {string} The encoded string.
+ */
+function safeEncodeURIComponent(s) {
+  try {
+    return encodeURIComponent(s);
+  } catch {
+    // replace lone sorrugates with U+FFFD and retry
```
mark-wiemer: can you explain what a lone surrogate is? And I think there's a typo in this comment, `s/sorru/surro`

lib/runner.js
```diff
@@ -608,6 +608,10 @@ Runner.prototype.hook = function (name, fn) {
 
     hook.allowUncaught = self.allowUncaught;
 
+    // reset hook state from any previous run so it can be reused
+    // (e.g afterEach hooks run once per test)
+    delete hook.state;
```
mark-wiemer, line 613: this feels dangerous, are we doing anything like this anywhere else?

lib/runner.js
```diff
@@ -619,6 +623,12 @@ Runner.prototype.hook = function (name, fn) {
     hook.run(function cbHookRun(err) {
       var testError = hook.error();
       if (testError) {
+        // Retrract pass count if the test was already counted as passed;
```
mark-wiemer: Please do check comments for spelling, just for professionalism. There is a VS Code extension called CSpell that can help here :)

```suggestion
        // Retract pass count if the test was already counted as passed;
```

lib/runner.js
```diff
@@ -619,6 +623,12 @@ Runner.prototype.hook = function (name, fn) {
...
+        // this.test.error() in afterEach hooks marks a previously-passing
+        //test as failed, so we must correct the stats to avoid double-counting
```
mark-wiemer: We typically have a space between the `//` and the text, nbd :)

```suggestion
        // test as failed, so we must correct the stats to avoid double-counting
```

mark-wiemer:
A few quick comments, but this will need a very detailed review, it'll be a couple weeks. That said, it looks decent and should be merge-able with a few fixups, hopefully!

lib/reporters/html.js
mostafaNazari702: Corrected typos, i also expanded the JSDoc to explain what a lone surrogate is, but anyway:

A lone surrogate is one half of a UTF-16 surrogate pair (code points U+D800–U+DFFF) that appears without its matching counterpart, e.g a high surrogate not followed by a low surrogate, or vice versa. JavaScript strings can legally contain them, but `encodeURIComponent()` throws a `URIError` when it encounters one. The `safeEncodeURIComponent` helper catches that, replaces the bad character(s) with the Unicode replacement character (U+FFFD), and retries.

lib/runner.js
mostafaNazari702: Yea!
`Runnable.prototype.reset()` in [lib/runnable.js](https://github.com/mochajs/mocha/blob/main/lib/runnable.js) line 62:
already does exactly this: `delete this.state`. `Hook` inherits from `Runnable`, so `delete hook.state` is the same pattern, applied for the same reason.

After an `afterEach` hook runs for the first test, it sets its state to `"passed"`. The issue is, if you don’t clear that before the next test, the hook still looks like it’s already been run. So when test 2 comes along, checks like `hook.isPending()` (and some guards around it) might just skip it entirely because the state isn’t undefined anymore.

Basically, the hook can end up not running at all for later tests, which is obviously not great.

The fix is just to clear that state before reusing the hook `delete hook.state` puts it back to how it was originally (no state means “hasn’t run yet”), which is exactly how `reset()` works on the base class anyway.

Also added a quick inline comment there to make it clear why that line exists, so it doesn’t look random later on.

lib/runner.js
mostafaNazari702: Much appreciated for the extension suggestion, saves lives, and people who cannot spell properly 😁.

mostafaNazari702:
> A few quick comments, but this will need a very detailed review, it'll be a couple weeks. That said, it looks decent and should be merge-able with a few fixups, hopefully!

Looking forward to the detailed review whenever you get to it!

lib/runner.js
mostafaNazari702: Excuse me if it is a poor explanation 😅

JoshuaKGoldberg:
> This pr fixes three seperate bugs:

👀

mark-wiemer:
Yes, please open one PR per bug :)

PR #5887, opened 2026-04-16
Fix/tap hierarchy diff hang timeout zero

### LINKED ISSUE #1624: Comparing buffers hangs when computing the diff report
I am having the very same issue as #1433 mocha 2.2.1 and chai 2.1.1. When 2 buffers are not equal, the diff computation hangs and CPU gets crazy.

Any chance of a regression bug here?

Sample code to reproduce this:

```js
var chai = require('chai');
var expect = chai.expect;

// chai.config.showDiff = false; // Without this line, the diff generation hangs

describe('Testing buffer equality', function () {
  it('should not take foreveeeeeer', function () {
    var file1 = Fs.readFileSync('/tmp/file1');
    var file2 = Fs.readFileSync('/tmp/file2');
    expect(file1).to.deep.equal(file2);
  });
});
```

At this point of writing, I am unclear if this issue comes from mocha or from chai, though...

### LINKED ISSUE #2410: Improve tap reporter

### LINKED ISSUE #2537: Inconsistent timeout policy

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2410, fixes #1624, fixes #2537

## Overview

Three old bugs, each in its own commit:

### TAP reporter flattens suite heirarchy (#2410):

Before, the TAP reporter was using `test.fullTitle()`, which basically crammed all the parent suite names into each test description. That made it hard for tools like `tap-spec` to understand the structure properly. Now it works differently, it prints out the suite names as comment lines when a suite starts, similar to how `tape` does it, and then just uses `test.title` for each test. So instead of repeating the full hierarchy in every test name, the structure is shown through those comment lines.

### Diff generation hangs on shared-reference objects (#1624):

The `canonicalize()` function in utils.js used to process the same object over and over again if it was referenced multiple times. So if you had something like an object pointing to itself or being reused in a loop, it could quickly spiral into a huge performance problem. Now it keeps track of objects it has already seen using a `WeakMap`, and if it runs into the same one again, it just returns `"[Duplicate]"` right away. Circular references are still handled the same as before, returning `"[Circular]"`. The original issue with buffers had already been handled by `maxDiffSize` but this specific problem with shared references which was reported back in 2018 had not been fixed until now.

### timeout(0) lets tests exit silently with code 0 (#2537):

When you set `this.timeout(0)`, it turns off the timeout, so if an async test never calls `done()` or finishes its promise, nothing keeps the event loop running and node just exits with code 0. the fix adds a `process.on('beforeExit')` listener inside `Runnable#run()` which catches that situation and fails the test with a clear error if the event loop empties before the test is done. Once the test does finish, the listener is removed in `done()`, so it only triggers for tests that never complete. It is also safe in the browser, since `beforeExit` only exists in node and there is a check to make sure `process` is defined.

### Testing

For testing, there are now unit tests covering all three changes, the TAP reporter, `canonicalize`, and the runnable logic. There is also an integration test specifically for the `timeout(0)` case. all the test suites are passing: the reporter tests, the unit tests, and the TypeScript build is clean as well.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -21,6 +21,7 @@ var EVENT_RUN_BEGIN = constants.EVENT_RUN_BEGIN;
 var EVENT_RUN_END = constants.EVENT_RUN_END;
 var EVENT_TEST_PENDING = constants.EVENT_TEST_PENDING;
 var EVENT_TEST_END = constants.EVENT_TEST_END;
+var EVENT_SUITE_BEGIN = constants.EVENT_SUITE_BEGIN;
 var sprintf = util.format;
 
 class TAP extends Base {
@@ -54,6 +55,12 @@ class TAP extends Base {
       self._producer.writeVersion();
     });
 
+    runner.on(EVENT_SUITE_BEGIN, function (suite) {
+      if (!suite.root) {
+        println("# %s", suite.title);
+      }
+    });
+
     runner.on(EVENT_TEST_END, function () {
       ++n;
     });
@@ -84,7 +91,7 @@ class TAP extends Base {
  * @return {String} title with any hash character removed
  */
 function title(test) {
-  return test.fullTitle().replace(/#/g, "");
+  return test.title.replace(/#/g, "");
 }
 
 /**
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -233,6 +233,7 @@ class Runnable extends EventEmitter {
     var ms = this.timeout() || MAX_TIMEOUT;
 
     this.clearTimeout();
+
     this.timer = setTimeout(function () {
       if (self.timeout() === 0) {
         return;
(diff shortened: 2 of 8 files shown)


mark-wiemer:
This one is pretty substantial, I might block it until v12 is out and stable. As always, it's on the backlog of PRs for us to review :)

mark-wiemer:
@mostafaNazari702 to merge faster, you might be better off splitting this into 3 PRs, one for each bug. In general, one PR per bug is our recommendation to make sure we can revert changes easily if needed and focus our reviews.

Also, I've marked one of the backing issues here as "needs triage" for now, can you go there and help us document a minimal modern repro?

mostafaNazari702:
Closing this in favor of three separate PRs (one per bug) as suggested by @mark-wiemer:

- TAP reporter hierarchy (#2410)
- Diff hang on shared references (#1624)
- timeout(0) silent exit (#2537)

Thanks for the feedback!

PR #5888, opened 2026-04-16
fix: show spec reporter failures at the correct indent level

### LINKED ISSUE #2556: Output indenting issue
See https://travis-ci.org/oddui/webdriver-extension/jobs/170984533 the first run passes and the second fails. The indentation starts to be one level further after the failed test.

Is it just a formatting issue or are the testing running in some strange order/nesting/context? I'm not able to find out why the tests fail from time to time and hope to get some ideas here. The code for the failed test can be found [here](https://github.com/oddui/webdriver-extension/blob/60ee0e5aedf28d9811b7fad0e8c3200c2869b30d/app/test/spec/extension/session_spec.js#L114). Thank you!

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2556

## Overview

So there was this weird issue with how failures showed up in the spec reporter.

If an async error popped up after a test had already passed, the runner would go back and mark that test as failed. Totally fine, but the way it was displayed was off. The reporter was using this running counter to figure out indentation, and that counter just follows whatever suite is currently being entered or exited, not the suite the failed test actually belongs to.

You'd end up with failures showing at the wrong level, which made the whole output kind of confusing to read.

The fix walks up `test.parent` to the root suite to figure out the real depth. If that info isn't available (like in parallel mode without `linkPartialObjects`, or with mocked test objects), it falls back to the old counter so things don't break.

Also added a fixture and integration test to recreate that async error situation across suites, just to make sure it's properly covered now.

### DIFF
--- a/lib/reporters/spec.js
+++ b/lib/reporters/spec.js
@@ -85,7 +85,25 @@ class Spec extends Base {
     });
 
     runner.on(EVENT_TEST_FAIL, function (test) {
-      Base.consoleLog(indent() + color("fail", "  %d) %s"), ++n, test.title);
+      // calculate indent from the test's actual parent suite hierarchy rather
+      // than the running indent counter, so that failures from uncaught
+      // exceptions firing across suite boundaries display at the correct
+      // nesting depth (issue #2556). Falls back to the running indent counter when
+      // the parent chain is incomplete (e.g. serialized objects in parallel
+      // mode without linkPartialObjects).
+      var failIndent;
+      if (test.parent && typeof test.parent.root === "boolean") {
+        var depth = 0;
+        var suite = test.parent;
+        while (suite && !suite.root) {
+          depth++;
+          suite = suite.parent;
+        }
+        failIndent = Array(depth + 1).join("  ");
+      } else {
+        failIndent = indent();
+      }
+      Base.consoleLog(failIndent + color("fail", "  %d) %s"), ++n, test.title);
     });
 
     runner.once(EVENT_RUN_END, (...args) => this.epilogue(...args));
--- a/test/integration/fixtures/uncaught/issue-2556.fixture.js
+++ b/test/integration/fixtures/uncaught/issue-2556.fixture.js
@@ -1,0 +1,23 @@
+'use strict';
+// first, i could not think of a better file name. :(
+// reproduces #2556: uncaught exception from a completed test should
+// bail and not run any further tests.
+
+describe('first suite', function () {
+  it('should pass then trigger async error', function (done) {
+    done();
+    process.nextTick(function () {
+      throw new Error('async error from first suite');
+    });
+  });
+
+  it('should not be reached - first suite', function () {
+    throw new Error('should never throw - first suite');
+  });
+});
+
+describe('second suite', function () {
+  it('should not be reached - second suite', function () {
+    throw new Error('should never throw - second suite');
+  });
+});
--- a/test/integration/uncaught.spec.js
+++ b/test/integration/uncaught.spec.js
@@ -215,4 +215,26 @@ describe("uncaught exceptions", function () {
       });
     });
   });
+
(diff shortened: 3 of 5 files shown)


test/integration/uncaught.spec.js
```diff
@@ -215,4 +215,19 @@ describe("uncaught exceptions", function () {
       });
     });
   });
+
+  it("issue-2556: should bail on async error across suites", function (done) {
```
mark-wiemer: hmm, how does this test the indent level?

mark-wiemer (commented):
one qq

test/integration/uncaught.spec.js
mostafaNazari702: You were right, the test did not test indentation at all. Rookie mistake, fixed.

test/integration/fixtures/uncaught/issue-2556.fixture.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+// first, i could not think of a better file name. :(
+// reproduces #2556: uncaught exception from a completed test should
+// bail and not run any further tests.
```
JoshuaKGoldberg: How about `completed-test-exception-bail.fixture.js`?

```suggestion
```

lib/reporters/spec.js
```diff
@@ -85,7 +85,25 @@ class Spec extends Base {
...
+        var depth = 0;
+        var suite = test.parent;
+        while (suite && !suite.root) {
+          depth++;
+          suite = suite.parent;
+        }
+        failIndent = Array(depth + 1).join("  ");
+      } else {
+        failIndent = indent();
+      }
```
JoshuaKGoldberg: [Refactor] This area of code has two common signals of "extract into a well-named function" for me:

* A big comment that explains inline what it does
* A `var` without an initial value that gets set by the result of `if` conditionals

Extracting into a function is a bit more code but I think makes the file more readable. How about here:

```suggestion
      var failIndent = indent(calculateTestIndentDepth(test));
```

...and above in the `constructor`:

```ts
    function indent(depth = indents) {
      return Array(depth).join("  ");
    }

    /**
     * If the test has a parent, calculates the indentation from its suite depth.
     * Falls back to the current test's default indentation level.
     * @see https://github.com/mochajs/mocha/issues/2556
     */
    function calculateTestIndentDepth(test) {
      if (!test.parent || typeof test.parent.root !== "boolean") {
        return indents;
      }

      var depth = 0;
      var suite = test.parent;
      while (suite && !suite.root) {
        depth++;
        suite = suite.parent;
      }

      return depth + 1;
    }
```

test/integration/uncaught.spec.js
```diff
@@ -215,4 +215,26 @@ describe("uncaught exceptions", function () {
       });
     });
   });
+
+  it("issue-2556: should show failure at the correct indent level", function (done) {
```
JoshuaKGoldberg:
```suggestion
  it("indents a suite with an uncaught exception at its nested depth", function (done) {
```

JoshuaKGoldberg (changes requested):
The test itself looks reasonable and good. Just requesting clarity touchups. 🚀

mostafaNazari702:
@mark-wiemer @JoshuaKGoldberg It would be great if Codecov’s diff results were shown only after all other PR checks have completed, and ideally as the final check in the list. A moment ago, seeing it earlier ( with its preliminary results that almost always say "Your PR lacks latest pushes to the codebase" ) made me think i had missed something in my commits, so i ended up pushing unnecessary "fixes" before all the checks had finished. That's partly on me for not waiting, but....yeah, everything is correct now anyway, i re-corrected my pushed commit.

TLDR: Make Codecov do code differences analysis as last "task" in a PR, last check in a PR, meaning.

PR #5889, opened 2026-04-16
fix: --recursive with globs, watch cache invalidation, and watch exit on missing files

### LINKED ISSUE #4016: Mocha fails to reload watched files
Under certain circumstances Mocha fails to reload source files in watch mode. Specifically, if a file is watched and modified but only required by a file that is not watched then the modified file is not reloaded and the change is not visible to the test suite.

This issue can be addressed by adding the unwatched files to `--watch-files` 

### Steps to Reproduce

Create the following three files

~~~js
// lib-a/index.js
module.exports = true
~~~

~~~js
// lib-b/index.js
module.exports = require('../lib-a')
~~~

~~~js
// test/main.js
const assert = require('assert')
const value = require('../lib-b')
it('works', function () {
  assert(value)
})
~~~

Now run
~~~bash
mocha --watch --watch-files lib-a,test
~~~

The test suite passes.

Now modify `lib-a/index.js` so that the exported value is `false`. This will retrigger the tests but they will still succeed.

If `lib-b` is add to the `--watch-files` list then the modification results in a test failure.

### LINKED ISSUE #4169: 🐛 Bug: Mocha quits watching when all tests file are removed
### Prerequisites

### Description

Mocha 7 quits watching process when the test folder is deleted

### Steps to Reproduce

1. Install mocha 7.0.1 `npm i mocha@7.0.1 -D`
1. Create a simple test file in "test" folder
1. Start mocha in watching mode: `./node_modules/.bin/mocha --watch`
1. In another terminal window, remove "test" folder: `rm -r test`

test.js file:
```javascript
describe("test",() => {
  it('logs', ()=>{console.log(1)})
})
```

Mocha continues to watching newly added files if any, mocha doesn't quit the process.

Mocha 7 quits with error message: "Error: No test files found: "test""
This is not an issue in mocha@6.

100%
### Versions

- The output of `mocha --version` and `node node_modules/.bin/mocha --version`: 7.0.1
- The output of `node --version`: v13.3.0
- Your operating system
  - name and version: macOS 10.13.6 (17G5019)
- Your browser and version (if running browser tests): none
- Any third-party Mocha-related modules (and their versions): none

### LINKED ISSUE #4186: Option "--recursive" does not support typescript files
If we need to grab all *.ts files from certain directory the only workaround is usage of bash globstar option:
```bash
mocha "./test/**/*.ts"
```
Following is **not** working:
```bash
mocha --recursive "./test/*.ts"
```
Note, that for *.js files both are working well:
```bash
mocha "./test/**/*.js"
mocha --recursive "./test/*.js"
```

Tested with latest 6.* version v6.2.2

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4186, fixes #4016, fixes #4169

## Overview

This is basically three related fixes around how files are handled in the CLI and in watch mode.

First one, `--recursive` finally does what you’d expect when you use it with glob patterns. before, if you ran something like `mocha "./test/*.ts" --recursive`, it would just ignore the flag, because recursion only kicked in for plain directories. Now it just rewrites the glob a bit under the hood (adds a `**/` in the right spot), so it actually looks through subfolders too. If your glob already has `**`, it leaves it alone.

second fix is about watch mode being a bit “half-aware” of changes. say you got `test.js` requiring `lib-b`, and `lib-b` requires `lib-a`. If only `lib-a` is being watched and you change it, mocha would rerun but `lib-b` would still be cached with the old version of `lib-a`, so your change wouldn’t show up. pretty frustrating. now, when it clears the cache, it doesn’t just remove the directly watched files it also kicks out anything that depends on them (as long as it’s your code, not stuff in `node_modules`). That way the whole chain actually refreshes properly.

And the last one: watch mode used to just die if all your test files disappeared, like if you deleted the test folder while it was running. It was calling `process.exit(1)` straight away. Now it throws an error instead, the watcher catches it, logs a warning, and just keeps running, waiting for files to come back. If you’re not in watch mode, though, it still exits exactly like before, so no change there.

Testing wise, added a few new tests around the recursive glob behavior, and everything’s passing, unit tests, watch integration tests, the usual smoke tests, and TypeScript checks are all clean.

### DIFF
--- a/lib/cli/collect-files.js
+++ b/lib/cli/collect-files.js
@@ -98,8 +98,9 @@ module.exports = ({
             unmatchedSpecFiles[0].pattern,
           )}` // stringify to print escaped characters raw
         : "Error: No test files found";
-    console.error(pc.red(noneFoundMsg));
-    process.exit(1);
+    const err = new Error(noneFoundMsg);
+    err.code = NO_FILES_MATCH_PATTERN;
+    throw err;
   } else {
     // print messages as a warning
     unmatchedSpecFiles.forEach((warning) => {
--- a/lib/cli/lookup-files.js
+++ b/lib/cli/lookup-files.js
@@ -76,8 +76,17 @@ module.exports = function lookupFiles(
   if (!fs.existsSync(filepath)) {
     let pattern;
     if (glob.hasMagic(filepath, { windowsPathsNoEscape: true })) {
-      // Handle glob as is without extensions
-      pattern = filepath;
+      if (recursive && !filepath.includes("**")) {
+        // when --recursive is set, transform the glob to search subdirectories
+        // e.g './test/*.ts' becomes './test/**/*.ts'
+        const dir = path.dirname(filepath);
+        const base = path.basename(filepath);
+        pattern = path.join(dir, "**", base);
+        debug("recursive glob pattern: %s", pattern);
+      } else {
+        // Handle glob as is without extensions
+        pattern = filepath;
+      }
     } else {
       // glob pattern e.g. 'filepath+(.js|.ts)'
       const strExtensions = extensions
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -23,6 +23,7 @@ const collectFiles = require("./collect-files");
 const { format } = require("node:util");
 const { createInvalidLegacyPluginError } = require("../errors");
 const { requireOrImport } = require("../nodejs/esm-utils");
+const { NO_FILES_MATCH_PATTERN } = require("../error-constants.mjs").constants;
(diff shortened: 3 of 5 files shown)


mark-wiemer:
Big PR == big time to review. But thanks for opening this one :)

JoshuaKGoldberg:
> This is basically three related fixes around how files are handled in the CLI and in watch mode.

@mostafaNazari702 do you think this can be split up, so we can review it more easily & merge things in granularly?

mostafaNazari702:
I am closing this PR as it has been splitted in 3 different PRs to address each issue, mentioned above.

Issue #5400, opened 2025-07-04
🛠 Repo: Switch from CJS to ESM internally


JoshuaKGoldberg:
### Tooling Suggestion Checklist

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have pulled the latest `main` branch of the repository.
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

Mocha's source code is written in a legacy CommonJS (CJS) style right now. That means `'use strict'` in source files, `const ... = require(...)`, `module.exports`, etc.

Moving to ECMAScript Modules (ESM) would allow the code to integrate better with more modern dev tooling. Static analysis tools such as ESLint and TypeScript are better able to analyze ESM, for example: which will help with #4154 and #4228.

Also I find ESM to be a cleaner syntax. It's more semantically precise! 🙂 

Suggestion: let's migrate internally from CJS to ESM?

### Additional Info

This ~is~ was blocked on #5358: bumping the minimum Node.js version to one that supports require(esm). 

I suspect this should land as a part of Mocha 12 (#5357). Although it _shouldn't_ be a breaking change on its own, it's very likely some users out there are depending on some internal structuring quirk. Having this be part of the 12 beta for a while will help test it out. And having it be part of a major will help shield us from protests of semver breakage.

wnghdcjfe:
Hello.
I tried to solve this issue, so I did the following.  
First, I touched only the reporter part.  
1.Code converted CJS part to ESM. 
2. Existing annotations have not been removed.
3. I left the variable or function name of the existing code. 
4. If it is declared as a var in the code, it has been changed to const or let. 

Is this the right way to do it? 
...
After working on it, now I know it's in that state is not `status: accepting prs`  haha....  Sorry..
I'll PR it more carefully next time

mark-wiemer:
@wnghdcjfe draft PRs are always welcome, they can help us triage :)

[...]

PR #5891, opened 2026-04-17
chore: convert commands, init, and lookup-files to ESM

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: works on #5400

## Overview

More progress!


mark-wiemer:
Tests pass on my machine (Linux Mint, Node 22.21.1)

mark-wiemer:
Test failure is transient Codecov outage

lib/cli/cli.js
```diff
@@ -17,8 +17,8 @@ const {
   loadOptions,
   YARGS_PARSER_CONFIG,
 } = require("./options");
-const lookupFiles = require("./lookup-files");
-const commands = require("./commands");
+const lookupFiles = require("./lookup-files.mjs").lookupFiles;
```
voxpelli:
```suggestion
const { lookupFiles } = require("./lookup-files.mjs");
```

lib/cli/cli.js
```diff
@@ -17,8 +17,8 @@ const {
...
+const commands = require("./commands.mjs");
```
voxpelli, line 21: So one doesn't need to do this in this instance? (I'm a bit rusty on the specifics here, just have a slight memory at the back of my mind about this)

```suggestion
const { 'default': commands } = require("./commands.mjs");
```

lib/cli/collect-files.js
```diff
@@ -5,7 +5,7 @@ const pc = require("picocolors");
 const debug = require("debug")("mocha:cli:run:helpers");
 const { minimatch } = require("minimatch");
 const { NO_FILES_MATCH_PATTERN } = require("../error-constants.mjs").constants;
-const lookupFiles = require("./lookup-files");
+const lookupFiles = require("./lookup-files.mjs").lookupFiles;
```
voxpelli:
```suggestion
const { lookupFiles } = require("./lookup-files.mjs");
```

lib/cli/init.mjs
```diff
@@ -1,26 +1,27 @@
...
  * @module
  */
 
-const fs = require("node:fs");
-const path = require("node:path");
+import fs from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
voxpelli: Use [`import.meta.dirname`](https://nodejs.org/docs/latest/api/esm.html#importmeta) instead

lib/cli/lookup-files.mjs
```diff
@@ -1,17 +1,17 @@
...
  */
 
-var fs = require("node:fs");
-var path = require("node:path");
-var glob = require("glob");
-var errors = require("../errors");
+import * as fs from "node:fs";
+import * as path from "node:path";
+import * as glob from "glob";
+import * as errors from "../errors.js";
```
voxpelli:
```suggestion
import fs from "node:fs";
import path from "node:path";
import glob from "glob";
import errors from "../errors.js";
```

At least true for the two `node:` modules?

lib/cli/lookup-files.mjs
```diff
@@ -1,17 +1,17 @@
...
 var createNoFilesMatchPatternError = errors.createNoFilesMatchPatternError;
 var createMissingArgumentError = errors.createMissingArgumentError;
```
voxpelli:
```js
import {
  createNoFilesMatchPatternError,
  createMissingArgumentError,
} from "./errors.js";
```

Looks like this is what we should do?

test/integration/file-utils.spec.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-const lookupFiles = require("../../lib/cli/lookup-files");
+const lookupFiles = require("../../lib/cli/lookup-files.mjs").lookupFiles;
```
voxpelli:
```suggestion
const { lookupFiles } = require("../../lib/cli/lookup-files.mjs");
```

test/unit/parse-query.spec.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-var parseQuery = require("../../lib/browser/parse-query");
+var parseQuery = require("../../lib/browser/parse-query.mjs").parseQuery;
```
voxpelli:
```suggestion
const { parseQuery } = require("../../lib/browser/parse-query.mjs");
```

voxpelli (changes requested):
Some of this feedback is about DRY:ness and depends on the level of refactor that we aim for with the ESM transformation, but I think these are sensible – and eg the change from wildcard import of `node:` module is more of a fix and consistency

lib/cli/cli.js
mark-wiemer: I prefer the current syntax. Your suggestion might work but feels a bit unusual and verbose to me :/

lib/cli/lookup-files.mjs
mark-wiemer: Only for the two node modules I believe, the others don't have a default export, I was getting some build issues.

lib/cli/cli.js
voxpelli: I prefer the current as well, it may be that it’s a type checking thing or something – can’t remember

voxpelli (approved):

lib/cli/cli.js
mark-wiemer: thanks, resolving :)

PR #5894, opened 2026-04-17
perf: replace he with turbo-he (Rust N-API, 3.5× faster HTML entity decode)

dev-kjma:
## What this does

Replaces [`he`](https://github.com/mathiasbynens/he) with [`turbo-he`](https://github.com/dev-kjma/turbo-he) — a Rust N-API implementation of the **identical API**. One line in `package.json`, one line per import. No logic changes.

## Benchmark results (Apple M2, Node.js v25.2.1)

| Workload | `he` | `turbo-he` | Speedup |
|---|---|---|---|
| Decode 10 kB HTML (hundreds of entities) | 321 µs | **91 µs** | **3.53× faster** |
| Decode 100 kB string | 3,380 µs | **932 µs** | **3.63× faster** |
| Decode via `decodeBuffer(Buffer)` ★ | 3,380 µs | **460 µs** | **7.35× faster** |
| Encode mixed ASCII + non-ASCII | 519 µs | **160 µs** | **3.24× faster** |
| Encode with `useNamedReferences: true` | 1,020 µs | **160 µs** | **6.38× faster** |

> For strings with no `&` characters, turbo-he returns in a single SIMD `memchr` pass with zero allocation. For tiny strings the fixed N-API call overhead (~150 ns) dominates — but any real-world HTML payload benefits.

## Why it's safe

- **75/75 parity tests** verify bit-for-bit identical output vs `he` on every case: named entities, numeric decimal/hex, legacy no-semicolon, attribute-value mode, strict mode, Windows-1252 remap, surrogate replacement
- Full HTML5 spec accuracy — same edge case handling as `he`  
- Zero production dependencies — links only against Node's built-in N-API
- MIT license (same as `he`)
- [x] Added tests
**Source:** https://github.com/dev-kjma/turbo-he · https://www.npmjs.com/package/turbo-he

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -9,7 +9,7 @@
  * Module dependencies.
  */
 var path = require("node:path");
-var he = require("he");
+var he = require('turbo-he');
 var pc = require("picocolors");
 var isUnicodeSupported = require("is-unicode-supported")();
 
--- a/package.json
+++ b/package.json
@@ -98,7 +98,6 @@
     "diff": "^8.0.3",
     "find-up": "^5.0.0",
     "glob": "^13.0.0",
-    "he": "^1.2.0",
     "is-path-inside": "^3.0.3",
     "is-unicode-supported": "^0.1.0",
     "js-yaml": "^4.1.0",
@@ -111,7 +110,8 @@
     "workerpool": "^10.0.0",
     "yargs": "^17.7.2",
     "yargs-parser": "^21.1.1",
-    "yargs-unparser": "^2.0.0"
+    "yargs-unparser": "^2.0.0",
+    "turbo-he": "^0.1.0"
   },
   "devDependencies": {
     "@eslint/js": "^10.0.0",


foray1010:
Please do not merge it: see https://github.com/eladnava/mailgen/pull/86

mark-wiemer:
Please open a backing issue first, ref https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#%EF%B8%8F-propose-a-change

JoshuaKGoldberg:
A note for the potential issue: performance tests in isolation are not convincing. Mocha does not use he parsing very much. There may be approximately zero user observable change in runtime performance.

On the other hand, turbo-he is >4.5x the install side of he (439kB vs. 122kB, https://packagephobia.com/result?p=turbo-he / https://packagephobia.com/result?p=he). This is actively _worse_ for performance (install time / size).

I understand you made a cool thing (yay!) and want to share it, but this isn't the way. https://e18e.dev/learn/advocacy.html

PR #5897, opened 2026-04-17
feat: add built-in github-actions reporter

### LINKED ISSUE #5183: 🛠️ Repo: Restore `mocha-github-actions-reporter`
Removed `mocha-github-actions-reporter` because it was being installed within the workflow rather than being a dev dependency of the `package.json`, which makes versioning of it non-straightforward.

And felt a bit odd to have all the other kinds of reporters provided within the module but not this – if we need it then so do others, we should dogfood it (and possibly offer to upstream it: https://github.com/daniellockyer/mocha-github-actions-reporter/issues/7#issuecomment-2263150207)

_Originally posted by @voxpelli in https://github.com/mochajs/mocha/pull/5177#discussion_r1700245270_

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5183

## Overview

this adds a built-in `github-actions` reporter, basically replacing the need for that external `mocha-github-actions-reporter` dependency in CI.

it’s built on top of the regular spec reporter, but adds a couple of nice extras when it detects it’s running inside GitHub Actions. For one, it creates error annotations, so if a test fails, you actually see it inline in the PR diff right where the problem is. super handy. It also writes a little summary (pass/fail/pending counts) to the job summary section in GitHub.

Another small but useful change: you can now set the reporter via a `MOCHA_REPORTER` env var. So in CI, you don’t even have to touch `.mocharc.yml` the workflow just sets `MOCHA_REPORTER=github-actions` and you’re good.

That cleans also up those old TODOs from when the external reporter got removed earlier, so now everything’s properly wired up without extra deps.

### DIFF
--- a/lib/cli/run.js
+++ b/lib/cli/run.js
@@ -215,7 +215,7 @@ exports.builder = (yargs) =>
         group: GROUPS.FILES,
       },
       reporter: {
-        default: defaults.reporter,
+        default: process.env.MOCHA_REPORTER || defaults.reporter,
         description: "Specify reporter to use",
         group: GROUPS.OUTPUT,
         requiresArg: true,
--- a/.github/workflows/mocha.yml
+++ b/.github/workflows/mocha.yml
@@ -36,8 +36,6 @@ jobs:
       npm-script: test-smoke
 
   test-node-lts:
-    # TODO: Restore "mocha-github-actions-reporter" style reporting without relying on third party module
-    # https://github.com/mochajs/mocha/issues/5183
     uses: ./.github/workflows/npm-script.yml
     needs: smoke
     strategy:
@@ -59,8 +57,6 @@ jobs:
     permissions:
       id-token: write # for Codecov OIDC
       contents: read # for Codecov OIDC
-    # TODO: Restore "mocha-github-actions-reporter" style reporting without relying on third party module
-    # https://github.com/mochajs/mocha/issues/5183
     uses: ./.github/workflows/npm-script.yml
     needs: test-node-lts
     strategy:
--- a/.github/workflows/npm-script.yml
+++ b/.github/workflows/npm-script.yml
@@ -71,6 +71,7 @@ jobs:
         env:
           BROWSER: ${{ matrix.browser }}
           COVERAGE: ${{ inputs.coverage && '1'}}
+          MOCHA_REPORTER: github-actions
           NODE_OPTIONS: "--trace-warnings"
       # Generate and upload coverage, even if tests fail
       - name: Generate coverage report
--- a/lib/reporters/github-actions.js
+++ b/lib/reporters/github-actions.js
@@ -1,0 +1,155 @@
+"use strict";
+
+/**
+ * @typedef {import('../runner.js')} Runner
+ * @typedef {import('../test.js')} Test
+ */
+
+/**
+ * @module GithubActions
+ */
+/**
+ * Module dependencies.
+ */
+
+var fs = require("fs");
+var Spec = require("./spec");
+var ms = require("ms");
+var constants = require("../runner").constants;
+var EVENT_TEST_FAIL = constants.EVENT_TEST_FAIL;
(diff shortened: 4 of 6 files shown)


lib/cli/run.js
```diff
@@ -215,7 +215,7 @@ exports.builder = (yargs) =>
         group: GROUPS.FILES,
       },
       reporter: {
-        default: defaults.reporter,
```
mark-wiemer, line 218: we should document this new env var in our site

mark-wiemer:
@mochajs/maintenance-crew , esp @voxpelli , curious for your thoughts on this one :)

voxpelli:
@mark-wiemer I'm cautiously optimistic, but we probably should look into getting this as type checked ESM? Or that's not possible yet in the reporters?

voxpelli:
> Another small but useful change: you can now set the reporter via a MOCHA_REPORTER env var. So in CI, you don’t even have to touch .mocharc.yml the workflow just sets MOCHA_REPORTER=github-actions and you’re good.

@mostafaNazari702 Ideally that should be a separate PR. Not saying that it has to be, but if we squash merge it together with the rest in this PR then it will get somewhat lost amidst the rest

Would be great to hear a bit about your design decisions and considerations as well and which parts you think is most va least contentious

mostafaNazari702:
Hey, thanks for the feedback, both of you brought up good points. @mark-wiemer @voxpelli 

On the ESM and type-checking stuff, right now all 16 reporters are still using CommonJS, and `index.js` is just requiring them like that. The ESM migration hasn’t reached reporters yet, it’s only been done for some core pieces so far. Converting just this one reporter to EsM would kind of stick out and make things messier in `index.js`. Feels cleaner to keep it consistent for now and migrate everything together later.
> P.S. (for the previos paragraph only): To be fully honest with you, I have already completed much of the migration in my own codebase for i plan to use Mocha, but I won’t push it or open a PR. I plan to use Mocha in a future project (just a personal one). I held off on pushing to my fork or creating a PR because I haven’t fully reviewed the license yet and therefore don't know what legal obligations i have.

Same idea with type-checking, the project allows JS in TypeScript, but it’s not actually checking JS files yet, and none of the reporters are using `@ts-check`. this file does have JSDoc types, so it’s ready when that switch flips but turning it on for just one file didn’t feel right.

Now, on splitting out `MOCHA_REPORTER`: yeah, that’s fair. I can break it into a separate PR if that’s preferred. I bundled it here because the reporter is really meant for CI, and without that env var you’d have to either tweak `.mocharc.yml` or pass `--reporter` everywhere, which kind of defeats the purpose. But I get that it’s technically a separate concern and might get buried in a squash, so I’m flexible.

As for what’s straightforward verses debatable:

Least contorversial stuff:

1) The reporter itself it’s basically a direct port of the old GitHub Actions reporter. it extends Spec, adds inline error annotations, and writes a job summary. Nothing too wild there honestly.
2) The `::error` formatting follows GitHub’s spec directly.
3) Removing those old TODO comments bcz they were just pointing to an already-resolved issue.
4) Hooking it up in `index.js` same pattern as all the other reporters.

now the more debateable parts are:
1) the `MOCHA_REPORTER` env var is set as a default, so anything explicit (CLI flag or config.) overrides it, but yeah, it’s new behavior and probably needs docs.
2) The stack trace parsing uses regex to grab file/line info. works fine with current Node/V8 output, but could break if that format changes. Worst case though, it just skips annotations instead of crashing.

And yeah, on docs, agreed. Happy to add them. Question is just whether to include that here or in a separate PR if we split out the env var.

mark-wiemer:
Quick comment, can you use ESM instead of CJS for all new code? So new files would be `.mjs` and use `import` instead of `require(...)`

mostafaNazari702:
I will check this soon.

mark-wiemer:
Mostafa is unavailable this summer, others are welcome to take up this work, just comment here or open a new PR based on this start (or your own start!)

PR #5898, opened 2026-04-17
fix: prevent diff hang on large/complex objects

### LINKED ISSUE #1624: Comparing buffers hangs when computing the diff report

tommyhgunz14:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1624

## Summary

Resolves #1624

Prevents Mocha from hanging when computing diffs for large buffers, deeply nested objects, or objects with many keys.

## Problem

When assertion libraries (like Chai) fail comparisons on large or complex objects, Mocha's diff generation can hang indefinitely during stringification, causing CPU to spike and tests to never complete.

## Solution

Added pre-stringification size estimation in `stringifyDiffObjs()` that checks object complexity **before** attempting to stringify. When values are too large or complex, a concise fallback message is used instead of attempting the costly operation.

Protection covers:
- Large buffers (>10KB)
- Deeply nested objects (>10 levels)
- Objects with many keys (>1000)
- Circular references

## Changes

- `lib/reporters/base.js`: Added `estimateSize()` function and guarded `stringifyDiffObjs()` with early bailout
- `test/unit/reporters/base-diff-protection.spec.js`: 4 unit tests exercising the full `list()` flow
- `test/integration/fixtures/diff-hang-protection.fixture.js`: Integration fixture

All existing reporter tests continue to pass (48/48).

## Performance Impact

- **Normal cases**: Negligible overhead (~1-2ms for estimation)
- **Pathological cases**: Prevents infinite hangs, returns immediately

### DIFF
--- a/test/integration/fixtures/diff-hang-protection.fixture.js
+++ b/test/integration/fixtures/diff-hang-protection.fixture.js
@@ -1,0 +1,41 @@
+'use strict';
+
+const {expect} = require('chai');
+
+describe('Diff hang protection', function () {
+  it('should not hang when comparing large buffers', function () {
+    // Create two large buffers that differ
+    const buf1 = Buffer.alloc(50000, 'a');
+    const buf2 = Buffer.alloc(50000, 'b');
+    
+    // This should fail without hanging
+    expect(buf1).to.deep.equal(buf2);
+  });
+  
+  it('should not hang when comparing deeply nested objects', function () {
+    // Create deeply nested object structure
+    let obj1 = {value: 1};
+    let obj2 = {value: 2};
+    
+    for (let i = 0; i < 100; i++) {
+      obj1 = {nested: obj1, extra: obj1};
+      obj2 = {nested: obj2, extra: obj2};
+    }
+    
+    // This should fail without hanging
+    expect(obj1).to.deep.equal(obj2);
+  });
+  
+  it('should not hang when comparing objects with many keys', function () {
+    const obj1 = {};
+    const obj2 = {};
+    
+    for (let i = 0; i < 5000; i++) {
+      obj1['key' + i] = 'value' + i;
+      obj2['key' + i] = 'different' + i;
+    }
+    
+    // This should fail without hanging
+    expect(obj1).to.deep.equal(obj2);
+  });
+});
--- a/test/integration/fixtures/diff-hang-protection.fixture.js
+++ b/test/integration/fixtures/diff-hang-protection.fixture.js
@@ -7,7 +7,6 @@ describe('Diff hang protection', function () {
     const buf1 = Buffer.alloc(50000, 'a');
     const buf2 = Buffer.alloc(50000, 'b');
     
-    // This should fail without hanging
     expect(buf1).to.deep.equal(buf2);
   });
   
--- a/test/integration/fixtures/diff-hang-protection.fixture.js
+++ b/test/integration/fixtures/diff-hang-protection.fixture.js
@@ -1,39 +1,36 @@
 'use strict';
 
-const {expect} = require('chai');
+const { expect } = require('chai');
 
 describe('Diff hang protection', function () {
   it('should not hang when comparing large buffers', function () {
     const buf1 = Buffer.alloc(50000, 'a');
     const buf2 = Buffer.alloc(50000, 'b');
-    
+
     expect(buf1).to.deep.equal(buf2);
   });
-  
+
   it('should not hang when comparing deeply nested objects', function () {
-    // Create deeply nested object structure
-    let obj1 = {value: 1};
-    let obj2 = {value: 2};
-    
+    let obj1 = { value: 1 };
+    let obj2 = { value: 2 };
+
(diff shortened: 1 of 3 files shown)


mark-wiemer:
Thanks @tommyhgunz14 , I've marked the backing issue as "needs triage" for now, can you go there and help us document a minimal modern repro?

tommyhgunz14:
Hi @mark-wiemer 👋 Thanks for triaging this! Here's a comprehensive minimal repro covering the main pathological cases, confirmed hanging on Mocha 11.7.5 and current main:

Setup:

```shell
mkdir mocha-hang-repro && cd mocha-hang-repro
npm init -y
npm install mocha@11.7.5 chai@4
```

repro.test.js:

```js
const { expect } = require('chai');
const { Buffer } = require('buffer');

// 1. Large buffer comparison (original report — hangs during diff of binary data)
describe('Buffer diff hang', function () {
  this.timeout(5000);
  it('should not hang comparing 20KB+ buffers', function () {
    const buf1 = Buffer.alloc(20 * 1024, 0x41);
    const buf2 = Buffer.alloc(20 * 1024, 0x42);
    expect(buf1).to.deep.equal(buf2);
  });
});

// 2. Deeply shared object references (exponential blowup in canonicalize)
describe('Shared-reference hang', function () {
  this.timeout(5000);
  it('should not hang on deeply shared objects', function () {
    let obj = {};
    for (let i = 0; i < 50; ++i) {
      obj = { a: obj, b: obj };
    }
    expect(obj).to.eql({});
  });
});

// 3. Wide object (many keys — slow stringification)
describe('Wide object hang', function () {
  this.timeout(5000);
  it('should not hang on objects with thousands of keys', function () {
    const big = {};
    for (let i = 0; i < 5000; i++) big[`k${i}`] = i;
    expect(big).to.eql({});
  });
});
```

Run: `npx mocha repro.test.js`

Each case causes Mocha to hang indefinitely (100% CPU) during diff generation. The 5s timeout fires for the test, but Mocha still hangs in the reporter trying to stringify the diff.

Our PR (#5898) adds a lightweight pre-stringification size check in stringifyDiffObjs() that bails out gracefully for these pathological cases while keeping normal diffs untouched.

tommyhgunz14:
I think this is ready to go now?

mark-wiemer:
Probably ready! Maintainers have been very busy (and I have also been sick!) so it will be a few days at the very least. But I am healthier now and catching up on stuff. The v12 release (#5357) takes priority right now, then older PR cleanup, then new PRs. Thanks in advance for your patience :)

test/integration/fixtures/diff-hang-protection.fixture.js
```diff
@@ -0,0 +1,41 @@
...
+
+const {expect} = require('chai');
+
+describe('Diff hang protection', function () {
+  it('should not hang when comparing large buffers', function () {
+    // Create two large buffers that differ
+    const buf1 = Buffer.alloc(50000, 'a');
+    const buf2 = Buffer.alloc(50000, 'b');
+    
+    // This should fail without hanging
```
JoshuaKGoldberg: [Docs] This is implied by the test title. Here and elsewhere:

```suggestion
```

test/integration/fixtures/diff-hang-protection.fixture.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const {expect} = require('chai');
+
+describe('Diff hang protection', function () {
+  it('should not hang when comparing large buffers', function () {
+    // Create two large buffers that differ
```
JoshuaKGoldberg: [Docs] This is implied by the code. I don't think these comments add anything helpful. Here and elsewhere:

```suggestion
```

JoshuaKGoldberg (approved):
Looks good to me! Although not every case in #1624 hangs, I think it's good to be thorough on all of them the way you have the tests here. Just in case.

Only requesting changes on removing some comments. If @mark-wiemer doesn't disagree then we can merge this in soon. Thanks! ✨

test/integration/fixtures/diff-hang-protection.fixture.js
tommyhgunz14: Done sir.

JoshuaKGoldberg:
The integration test failures are probably unrelated (#5361). I'll touch up the formatting and comments now.

PR #5904, opened 2026-04-18
fix!: emit suite names as TAP comments and use test.title instead of fullTitle

### LINKED ISSUE #2410: Improve tap reporter

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2410

## Overview

So before, the TAP reporter was using test.fullTitle(), which basically stuffed the whole suite hierarchy into every single test name.

That made things kinda messy, and tools like tap-spec couldn’t really understand the structure properly. Everything just looked flat with long names.

Now it’s a bit smarter about it. When a suite starts, it prints the suite name as a comment line (kind of like how tape does it), and then each test just uses test.title. So instead of repeating the whole hierarchy over and over, you actually see the structure from those comment lines. Way cleaner and easier to follow.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -21,6 +21,7 @@ var EVENT_RUN_BEGIN = constants.EVENT_RUN_BEGIN;
 var EVENT_RUN_END = constants.EVENT_RUN_END;
 var EVENT_TEST_PENDING = constants.EVENT_TEST_PENDING;
 var EVENT_TEST_END = constants.EVENT_TEST_END;
+var EVENT_SUITE_BEGIN = constants.EVENT_SUITE_BEGIN;
 var sprintf = util.format;
 
 class TAP extends Base {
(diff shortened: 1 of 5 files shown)


lib/reporters/tap.js
JoshuaKGoldberg, line 1: [Non-Actionable] Just noting that this is TAP 12/13's informal convention of `# SuiteName`. TAP 14 is a whole other thing that Mocha's reporter doesn't try to support (yet?). This is fine.

JoshuaKGoldberg (dismissed):
Tricky stuff! I looked over the [TAP 13 specification](https://testanything.org/tap-version-13-specification.html) and tried to map it to the changes & tests here. It seems to be good to me. I also asked Sonnet 4.6 to verify things and it's happy too. But I'm not a TAP user myself, so if anybody who actually uses it could weigh in at some point, that'd be nice.

We have to wait for a bit to merge anyway per https://github.com/mochajs/mocha/issues/2410#issuecomment-4467218246, so leaving this open but approved. Thanks & nice job! 👏

PR #5906, opened 2026-04-18
fix: fail test when timeout(0) and promise never resolves

### LINKED ISSUE #2537: Inconsistent timeout policy

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2537

## Overview

This one fixes a pretty subtle edge case with `timeout(0)`.

when you disable timeouts with `this.timeout(0)`, async tests can end up in a weird situation where they never actually finish but also don’t get stopped by a timeout. If the test forgets to resolve (like a promise never settles or `done()` is never called), the Node event loop just empties out and the process exits with code 0. So it looks like success, even though the test never completed.

to fix that, a `process.on('beforeExit')` listener was added inside `Runnable#run()`. It catches the moment Node is about to shut down while a test is still active, and fails the test with a clear error instead of silently exiting.

Once the test actually finishes, that listener is removed in `done()`, so it only applies to stuck/incomplete tests.

It’s also guarded so it doesn’t run in the browser since `beforeExit` is Node-only.

Tests cover both the runnable behavior and a real integration case using a fixture with a never-resolving promise to simulate the issue. Everything passes, unit tests, integration tests, and TypeScript checks are all clean.

### DIFF
--- a/lib/runnable.js
+++ b/lib/runnable.js
@@ -233,6 +233,7 @@ class Runnable extends EventEmitter {
     var ms = this.timeout() || MAX_TIMEOUT;
 
     this.clearTimeout();
+
     this.timer = setTimeout(function () {
       if (self.timeout() === 0) {
         return;
@@ -295,6 +296,12 @@ class Runnable extends EventEmitter {
         return multiple(err);
       }
 
+      // Remove beforeExit listener if set (timeout-disabled async tests)
+      if (beforeExitHandler) {
+        process.removeListener("beforeExit", beforeExitHandler);
+        beforeExitHandler = null;
+      }
+
       self.clearTimeout();
       self.duration = new Date() - start;
       finished = true;
@@ -314,6 +321,32 @@ class Runnable extends EventEmitter {
         ),
       );
       return;
+    }
+
+    // When timeout is disabled, detect event loop emptying for promise-returning
+    // tests. Without this, tests that never resolve their promise would cause
+    // the process to exit silently with code 0.
+    // Only for non-done-callback tests: done-callback tests may legitimately
+    // drain the event loop (e.g. setTimeout(done).unref()) before done fires.
+    var beforeExitHandler = null;
+    if (
+      this.timeout() === 0 &&
+      !this.async &&
+      typeof process !== "undefined" &&
+      typeof process.on === "function"
+    ) {
+      beforeExitHandler = function () {
+        if (!finished) {
+          done(
+            new Error(
+              "async test exited without done() being called or Promise resolving " +
+                "(the event loop has no remaining work); " +
+                "if this test intentionally takes a long time, use a non-zero timeout",
+            ),
+          );
+        }
+      };
+      process.on("beforeExit", beforeExitHandler);
     }
 
     // explicit async with `done` argument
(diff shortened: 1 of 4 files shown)


test/integration/fixtures/timeout-zero-silent-exit.fixture.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+describe('timeout 0 silent exit', function () {
+  this.timeout(0);
+
+  it('should fail when promise never resolves', function () {
+    // Return a promise that never resolves.
+    // Without the fix, the process would exit silently with code 0.
```
JoshuaKGoldberg: [Docs] This is the kind of comment AIs often leave: it thinks in the context of the PR / requested change, not the repo as a whole.

* PR: it makes sense to refer to "the fix", that's what's happening here
* As a whole: what is "the fix"? The meaning of that gets lost over time

As a rule, I'd recommend avoiding these kind of transient comments when possible. A better way to convey them is in good naming, such as the added function in `timeout.spec.js`. That name is really good & informative!

```suggestion
```

lib/runnable.js
```diff
@@ -360,6 +393,11 @@ class Runnable extends EventEmitter {
       var result = fn.call(ctx);
       if (result && typeof result.then === "function") {
         self.resetTimeout();
+        // When timeout is disabled, unref the fallback timer so the event
+        // loop can drain and beforeExit fires for never-resolving promises.
+        if (self.timeout() === 0 && self.timer) {
+          self.timer.unref();
```
JoshuaKGoldberg: [Style] Nit: it's just a little cleaner this way ✨ 

```suggestion
        if (self.timeout() === 0) {
          self.timer?.unref();
```

lib/runnable.js
```diff
@@ -316,6 +323,32 @@ class Runnable extends EventEmitter {
...
+            new Error(
+              "async test exited without done() being called or Promise resolving " +
+                "(the event loop has no remaining work); " +
+                "if this test intentionally takes a long time, use a non-zero timeout",
+            ),
+          );
+        }
+      };
+      process.on("beforeExit", beforeExitHandler);
+    }
```
JoshuaKGoldberg: [Bug]-ish: this logic is done for all non-`done`-callback tests: both Promise-returning and synchronous. Synchronous tests don't need it. They'll just fire `done()` synchronously.

How about moving this into `callFn()` and checking if the `result` is a thenable?

test/integration/timeout.spec.js
JoshuaKGoldberg, line 1: [Testing] The one added test here is for a test that returns a Promise. But the the original bug report in #2537 was actually for the [`function(done) {}`]() style. We'll need testing for that too. Which I suspect might reveal some gaps in logic.

JoshuaKGoldberg (changes requested):
A start 🚀 I think this gets 1 of the 2 needed styles of non-synchronous tests.

PR #5932, opened 2026-05-02
chore: partial migration to ESM

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: partial #5400

## Overview

I will push commits one by one and request individual reviews for each. first, I considered creating separate PRs for different phases, expecting a large number of modified files. However, not counting test files, we’re now close to a full migration. What remains are a few high-risk and sensitive areas where I’d appreciate your guidance on how to proceed.

@mark-wiemer @JoshuaKGoldberg

### DIFF
--- a/lib/cli/commands.mjs
+++ b/lib/cli/commands.mjs
@@ -7,4 +7,4 @@
 
 export * as init from "./init.mjs";
 // default command
-export { default as run } from "./run.js";
+export * as run from "./run.mjs";
--- a/lib/cli/run.mjs
+++ b/lib/cli/run.mjs
@@ -1,5 +1,3 @@
-"use strict";
-
 /**
  * Definition for Mocha's default ("run tests") command
  *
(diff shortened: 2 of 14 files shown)


mostafaNazari702:
I have seen stuff like `esmock`, `quibble`, `testdouble.js` floating around, but i do not know if there iss a preferred direction here, or if someone is planning to handle that in a separate PR first.

in a couple places (like that `esm-utils` test), the test is basically grabbing a function off a module and swapping it out on the fly. from what I understand, that is not really doable with ES module exports since they are kinda locked down. should i switch those tests over to one of those mocking libs, or change the actual code a bit so the function gets passed in (so tests can just give it a fake version)? not sure what style yall rather go with there.

mostafaNazari702:
A couple places use `require.resolve` for finding files, in ES modules, the workaround seems to be creating a fake require using that `createRequire` thing. is that the accepted pattern here, or would you guys rather just keep those files in the old format and not bother?

JoshuaKGoldberg (changes requested):
Great progress 🙂. As we mentioned in https://github.com/mochajs/mocha/issues/5400#issuecomment-4467025307, let's remove any major refactors / test shakeups to just leave the clean straightforward changes. LMK when you feel it's ready for re-review and I'll happily take a look. thanks!

PR #5941, opened 2026-05-04
fix: lookupFiles() now returns string[] for direct file paths

### LINKED ISSUE #5940: bug: lookupFiles() returns a string instead of string[] when given a direct file path
## Summary

`lookupFiles()` in `lib/cli/lookup-files.mjs` has a documented return type of `string[]`, but when given a direct file path (i.e., a file that exists on disk, not a glob or directory), it returns a bare `string` instead of a `string[]`.

## Environment

- Mocha version: latest (`main` branch)
- - Node.js version: v20+ / v22+
- - - Platform: Any
## Reproduction

```js
// lib/cli/lookup-files.mjs (lines 107-115)
try {
  stat = fs.statSync(filepath);
  if (stat.isFile() || stat.isFIFO()) {
    return filepath; // <- returns a bare string, not an array
  }
} catch {
  return;
}
```

**Minimal reproduction (programmatic API):**

```js
const { lookupFiles } = require('mocha/lib/cli/lookup-files');

const result = lookupFiles('./test/mytest.spec.js', ['js'], false);
// Expected: ['./test/mytest.spec.js']  (string[])
// Actual:   './test/mytest.spec.js'    (string)

// This breaks spread:
const files = [...result]; // iterates over characters, not file paths!
```

## Impact

The JSDoc on `lookupFiles` documents the return type as `{string[]}`:

```js
 * @return {string[]} An array of paths.
```

However, when a direct file path is passed (not a glob, not a directory), the function short-circuits and returns a plain `string`. This creates a type inconsistency that can silently break downstream consumers, including:

- Any programmatic API user who spreads the result: `[...lookupFiles(file)]` would iterate over **characters**.
- - Any consumer who calls `.map()` or `.filter()` on the result - these would act on characters of the string, not file paths.
- - - Plugins or third-party integrations that rely on the documented interface.
Currently, Mocha's own internal call site in `run-helpers.js` happens to use `[].concat(lookupFiles(...))`, which works around the inconsistency since `[].concat("string")` produces `["string"]`. However, this is fragile and relies on undocumented behavior.

## Suggested Fix

Wrap the file path in an array before returning, to match the documented return type:

```diff
if (stat.isFile() || stat.isFIFO()) {
-   return filepath;
+   return [filepath];
}
```

This is a 1-line change in `lib/cli/lookup-files.mjs` and makes the actual return type consistent with the documentation.

A corresponding test should also verify that a direct file path argument returns an array with exactly one element.

## Additional Context

This was discovered while auditing the `lib/cli/lookup-files.mjs` module. The inconsistency is subtle because the internal call site in `run-helpers.js` uses `[].concat(...)`, which silently handles both strings and arrays - masking the bug from the existing test suite.

maniktyagi04:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5940

## Overview

This PR addresses a return-type inconsistency in `lookupFiles()`. As documented in its JSDoc, the function should return `string[]`, but it was returning a bare `string` when a direct file path was provided.

### Changes
- Modified `lib/cli/lookup-files.mjs` to wrap direct file path returns in an array.

Verified locally using smoke tests and by confirming `Array.isArray()` on the return value for direct file paths.


JoshuaKGoldberg:
> - [x] That issue was marked as [status: accepting prs](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)

🤔🤔🤔

maniktyagi04:
@JoshuaKGoldberg 
Thanks for pointing that out! I tried removing/updating that checklist item, but the automated PR checks require it to be present and checked, otherwise the build fails. So I’ve kept it as-is to satisfy CI.

Happy to adjust if there’s an alternative way to handle this.

JoshuaKGoldberg (approved):
LGTM, thanks!

JoshuaKGoldberg:
Unrelated integration test flake, I believe.

JoshuaKGoldberg (approved):

PR #5977, opened 2026-05-16
feat: add --exit-timeout to warn and force-quit should the process linger after test run

### LINKED ISSUE #1855: Fail tests/suites that does not exit gracefully when --no-exit flag is used
I recently tried the `--no-exit` flag and found it a terrific way to determine if any of my modules keep active listeners. However, if it is the case that I have open listeners then mocha never exits. Would it be possible to add a timeout just as in asynchronous tests for this. Meaning that if I use the `--no-exits` flag and the event loop is not clear after my suite is complete, then make the test fail.

I'd be happy to supply this as a PR if the idea is accepted.

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #1855

## Overview

Sometimes a test suite finishes but the process still hangs because some async stuff didnt get cleaned up, `--exit` just force-kills everything which hides the real issue and without it there's no built in cut off.

`--exit-timeout <ms>` sets up an unref'd timer so it doesnt keep the process alive but if something else is still hanging the timer kicks in and prints a message about likely leaked resources and then exits with force.

if i am not mixing this up with the previos PR, i think i saw a bug [and fixed it] where the `process.on('exit')` handler overwrote the exit code after the timeout already sat it.

### DIFF
--- a/lib/cli/run-helpers.js
+++ b/lib/cli/run-helpers.js
@@ -28,16 +28,34 @@ const { PluginLoader } = require("../plugin-loader.mjs");
 /**
  * Exits Mocha when tests + code under test has finished execution (default)
  * @param {number} clampedCode - Exit code; typically # of failures
+ * @param {number} [exitTimeout] - ms to wait before force-exiting. 0 means disabled
  * @ignore
  * @private
  */
-const exitMochaLater = (clampedCode) => {
+const exitMochaLater = (clampedCode, exitTimeout = 0) => {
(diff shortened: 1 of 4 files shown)


mostafaNazari702:
Excuse me for the unprofessionalism but it was a truly chalenging one and i ttook a lot of code disassembly, at one point i accdeintally deleted the file and i had to copy again from what is in upstream and start all over.

PR #5089, opened 2024-02-08
#5084 global variable

CheadleCheadle:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5084 
- [ ] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Mocha now has the mochaVar (open to different name) property in the global scope, including its current name and version. This global property enables external testing utilities to statically identify both the version of Mocha in use and the specific testing framework being utilized.
mochavar

PR #5984, opened 2026-05-20
feat: expose globalThis.mocha with runner name and version

### LINKED ISSUE #5084: 🚀 Feature: Add a global variable in tests indicating the test runner name (Mocha)
### Feature Request Checklist

### Overview

Mocha right now adds standard test hooks to `globalThis`: `after`, `describe`, `it`, and so on. But there's nothing indicating that it's specifically _Mocha_ being run, and not some other test framework. Would it be reasonable to add some variable like `mocha` to the global scope?

With this global variable, testing utilities would be able to statically determine the test framework (Mocha) without being explicitly told it in their configuration. This can be useful for utilities that hook into framework-specific APIs.

### Suggested Solution

```js
Object.defineProperty(globalThis, "mocha", {
  get: () => ({
    name: "mocha",
    version: "12.34.56"
  }),
});
```

### Alternatives

A `process.env` would work on the CLI / in Node.js, but not in all browser environments.

Adding some string to a pore-existing global like `after` might work too, but seems hacky. Right now `after.toString()` comes out to `"function (name, fn) {\n      suites[0].afterAll(name, fn);\n    }"`.

### Additional Info

See https://github.com/JoshuaKGoldberg/console-fail-test/issues/10: https://github.com/JoshuaKGoldberg/console-fail-test is a utility that runs during tests and uses the test framework's APIs to call testing hooks. It can sort of dynamically detect Mocha, but the implementation is brittle.

rkdfx:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5084 

## Overview

Picks up the work from #5089 (thanks @CheadleCheadle) and addresses the outstanding review feedback so it can land.

Adds a non-enumerable read-only `globalThis.mochaVar` getter exposing `{ name, version }` from `package.json`.

### DIFF
--- a/lib/context.mjs
+++ b/lib/context.mjs
@@ -1,3 +1,28 @@
+import debugModule from "debug";
+import packageJson from "../package.json" with { type: "json" };
+
+const debug = debugModule("mocha:context");
+
+/**
+ * Defines a read-only `globalThis.mochaVar` so external tooling can
+ * statically identify Mocha as the active test runner. Guarded because
+ * sandboxed runtimes (SES/Endo, ShadowRealm) can disallow globalThis
+ * mutation, in those environments the marker is simply absent.
+ *
+ * @example
+ * globalThis.mochaVar; // { name: "mocha", version: "X.Y.Z" }
+ */
+try {
+  Object.defineProperty(globalThis, "mochaVar", {
+    get: () => ({
+      name: packageJson.name,
+      version: packageJson.version,
+    }),
+  });
+} catch (err) /* istanbul ignore next */ {
+  debug("unable to define globalThis.mochaVar", err);
+}
+
 /**
  * @typedef {import('./runnable.mjs')} Runnable
  */
--- a/test/integration/global-variable.spec.js
+++ b/test/integration/global-variable.spec.js
@@ -1,0 +1,18 @@
+"use strict";
+
+const mochaPackageJson = require("../../package.json");
+
+describe('global "mochaVar" object', function () {
+  it("exposes the runner name and version", function () {
+    expect(globalThis.mochaVar, "to be an object");
+    expect(globalThis.mochaVar.name, "to equal", mochaPackageJson.name);
+    expect(globalThis.mochaVar.version, "to equal", mochaPackageJson.version);
+  });
+
+  it("returns a fresh object on each access (read-only getter)", function () {
+    const a = globalThis.mochaVar;
+    const b = globalThis.mochaVar;
+    expect(a, "not to be", b);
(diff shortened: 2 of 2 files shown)


lib/context.mjs
```diff
@@ -1,3 +1,28 @@
...
+ * Defines a read-only `globalThis.mochaVar` so external tooling can
+ * statically identify Mocha as the active test runner. Guarded because
+ * sandboxed runtimes (SES/Endo, ShadowRealm) can disallow globalThis
+ * mutation, in those environments the marker is simply absent.
+ *
+ * @example
+ * globalThis.mochaVar; // { name: "mocha", version: "X.Y.Z" }
+ */
+try {
+  Object.defineProperty(globalThis, "mochaVar", {
```
JoshuaKGoldberg, line 16: [Naming] `"Var"`, as short for variable, doesn't add any useful meaning (does it? am I misreading it?). This isn't really a variable, and even if it was, it's not useful to say in the name of a variable that the thing is a variable.

```suggestion
  Object.defineProperty(globalThis, "mocha", {
```

lib/context.mjs
```diff
@@ -1,3 +1,28 @@
...
+    get: () => ({
+      name: packageJson.name,
+      version: packageJson.version,
+    }),
+  });
+} catch (err) /* istanbul ignore next */ {
```
JoshuaKGoldberg, line 22: > `istanbul ignore next`

[Testing] We try to avoid this kind of thing. A test should be able to define a global, non-configurable/writable property and assert on this debug complaint. Can you try that, and either update the PR if it works or say why not if it doesn't?

JoshuaKGoldberg (changes requested):
Sorry this took so long to review! A couple of things to touch up, but I think overall this is the right general strategy.

lib/context.mjs
rkdfx: Tried the rename, but it breaks the browser bundle. browser-entry.js does global.mocha = mocha. https://github.com/mochajs/mocha/blob/90f48e40b95eb622ad7123915e730c284652a1e8/browser-entry.js#L206 

Throws this error `TypeError: Cannot set property mocha of #<Object> which has only a getter`

JoshuaKGoldberg: In that case you'll have to dig a bit deeper as to why these are happening 🕵️. Let us know if you need help, we're happy to answer questions!

rkdfx: I made this `writable` and `configurable` to true

mark-wiemer (approved):
Gonna give Josh a few days before merging, but thank you!

PR #5986, opened 2026-05-22
fix: --reporter-option on CLI overrides values from config (#5532)

### LINKED ISSUE #5532: Bug: command line reporter-options parameter doesn't override .mocharc.yml
### Bug Report Checklist

### Expected

Mocha should use the `reporter-options` specified on the command-line instead of the one defined in  `.mocharc.yml`.

### Actual

Mocha use the `reporter-options` defined in `.mocharc.yml` instead of the one specified on the command-line.

1. Define a `.mocharc.yml` file with  `reporter-options`:

```yaml
recursive: true
reporter: xunit
reporter-options:
  - 'output=./test-reports/xunit.xml'
```

2. Call mocha using a different value for `reporter-options`:

```bash
node node_modules/mocha/bin/mocha --reporter-options output=./test-reports/xxxunit.xml
```

3. Verify that mocha has created `./test-reports/xunit.xml` and not `./test-reports/xxxunit.xml`

### Versions

Debian Bookworm / Mocha version 6.2.3 / Node version v24.11.0 -> work as expected
Debian Bookworm / Mocha version 7.2.0 / Node version v24.11.0 -> work as expected
Debian Bookworm / Mocha version 7.2.0 / Node version v24.11.0 -> work as expected
Debian Bookworm / Mocha version 8.2.1 / Node version v24.11.0 -> work as expected
Debian Bookworm / Mocha version 8.3.0 / Node version v24.11.0 -> **do not work as expected**
Debian Bookworm / Mocha version 11.7.5 / Node version v24.11.0 -> **do not work as expected**

### Additional Info

I have the same problem on other node versions (Debian Buster/Node 16 v16.20.2), so I guess it's mocha related.

The behavior has change between _mocha@8.2.1_ and  _mocha@8.3.0_. 
As I still have the same problem with  _mocha@11.7.5_, it is in all versions since v8.3.0.

For me it looks very similar to issue: https://github.com/mochajs/mocha/issues/3951

devareddy05:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5532

## Overview

`--reporter-option` (`--reporter-options`, `-O`) given on the CLI did not override the same key declared in `.mocharc.*` or `package.json`. The config-file value won instead. Regression observed since v8.3.0, still present in v11.7.5+. The maintainer comment on the issue narrowed it to "Mocha not recognizing the clashing options as 'clashing'".

### Root cause

In `lib/cli/options.mjs`, the second call to `parse(...)` passes CLI args, env, RC, and `package.json` as `configObjects` in priority-HIGH→LOW order. `yargs-parser`'s `combine-arrays: true` concatenates the `reporter-option` array from every source in that order: `[CLI, env, rc, pkg]`. The downstream coerce in `lib/cli/run.js` (around L223-L243) then reduces this array left-to-right with last-writer-wins — so the **lowest-priority** source ends up taking precedence, which is the inverse of the intended priority order.

### Fix

After the second parse, re-emit `reporter-option` in priority-LOW→HIGH order (`pkg, rc, env, CLI`). The coerce's existing last-writer-wins reduction now yields the correct result: CLI overrides env overrides rc overrides pkg. Per-key merge semantics are preserved — CLI overrides only the clashing keys; non-clashing keys from the config still flow through.

A small helper (`readReporterOption`) reads the value under any of the supported source-side aliases (`reporter-option`, `reporter-options`, `reporterOption`, `reporterOptions`, `O`) so config files written with either casing/spelling are respected.

### Behavior matrix (verified locally)

| Scenario | Result |
|---|---|
| CLI `output=cli.xml`, config `output=config.xml` | CLI wins → `cli.xml` |
| CLI `output=cli.xml`, config `output=config.xml`, `suiteName=FromConfig` | `{output: 'cli.xml', suiteName: 'FromConfig'}` |
| CLI repeats `foo=bar foo=baz`, no config | `{foo: 'baz'}` (last wins, unchanged) |
| Env `foo=fromEnv`, config `foo=fromConfig` | Env wins |

### Tests

- New integration test in `test/integration/options/reporter-option.spec.js`: "should allow the CLI to override clashing keys from the config", driving Mocha via `--config <fixture-yaml>` + `--reporter-option foo=fromCli` and asserting the custom reporter prints `{"foo":"fromCli","extra":"keepMe"}`.
- New fixture: `test/integration/fixtures/options/reporter-option-mocharc.yml`.

### Validation

- `npm run format:check`, `npm run tsc` — clean
- `npm run test-smoke` — 1/1 passing
- `npm run -s test-node-run -- "test/integration/options/reporter-option.spec.js"` — 4/4 passing (3 pre-existing + 1 new)
- `npm run -s test-node-run -- "test/integration/options/*.spec.js"` — 158/158 passing
- `npm run test-node:unit` — 1008 passing, 3 pending

Closes #5532

### DIFF
--- a/lib/cli/options.mjs
+++ b/lib/cli/options.mjs
@@ -283,6 +283,16 @@ export const loadPkgRc = (args = {}) => {
  * @alias module:lib/cli.loadOptions
  * @returns {external:yargsParser.Arguments} Parsed args from everything
  */
+// Reads `reporter-option` from a config source under any of its known
+// aliases. Keeps `loadOptions` agnostic of how each source was authored.
+const readReporterOption = (source) =>
+  source &&
+  (source["reporter-option"] ??
+    source["reporter-options"] ??
+    source.reporterOption ??
+    source.reporterOptions ??
+    source.O);
+
 export const loadOptions = (argv = []) => {
   let args = parse(argv);
   // short-circuit: look for a flag that would abort loading of options
@@ -298,6 +308,15 @@ export const loadOptions = (argv = []) => {
   const envConfig = parse(process.env.MOCHA_OPTIONS || "");
   const rcConfig = loadRc(args);
   const pkgConfig = loadPkgRc(args);
+
+  // Capture per-source `reporter-option` before yargs-parser's `combine-arrays`
+  // flattens them and loses provenance. See the re-merge step below for why.
+  const reporterOptionSources = {
+    pkg: readReporterOption(pkgConfig),
+    rc: readReporterOption(rcConfig),
+    env: readReporterOption(envConfig),
+    cli: readReporterOption(args),
+  };
 
   if (rcConfig) {
     args.config = false;
@@ -316,6 +335,20 @@ export const loadOptions = (argv = []) => {
     rcConfig || {},
     pkgConfig || {},
   );
+
+  // `combine-arrays` concatenates `reporter-option` from every source in
+  // high→low priority order. The downstream coerce in `lib/cli/run.js` reduces
+  // it left-to-right with last-writer-wins, which inverts the intended
+  // precedence. Re-emit lowest→highest priority so CLI overrides config.
+  const orderedReporterOption = [
+    reporterOptionSources.pkg,
+    reporterOptionSources.rc,
+    reporterOptionSources.env,
+    reporterOptionSources.cli,
+  ].reduce((acc, src) => (src ? acc.concat(list(src)) : acc), []);
+  if (orderedReporterOption.length) {
+    args["reporter-option"] = orderedReporterOption;
+  }
 
   // recombine positional arguments and "spec"
   if (args.spec) {
--- a/test/integration/fixtures/options/reporter-option-mocharc.yml
+++ b/test/integration/fixtures/options/reporter-option-mocharc.yml
@@ -1,0 +1,3 @@
+reporter-options:
+  - foo=fromConfig
+  - extra=keepMe
--- a/test/integration/options/reporter-option.spec.js
+++ b/test/integration/options/reporter-option.spec.js
@@ -83,4 +83,46 @@ describe("--reporter-option", function () {
       );
     });
   });
+
+  describe("when given options on the CLI and in a config file", function () {
+    var reporterFixture = path.join(
+      __dirname,
+      "..",
+      "fixtures",
+      "options",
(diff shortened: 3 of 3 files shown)


mark-wiemer:
Test failure is known issue #6036

lib/cli/options.mjs
```diff
@@ -283,6 +283,16 @@ export const loadPkgRc = (args = {}) => {
  * @alias module:lib/cli.loadOptions
  * @returns {external:yargsParser.Arguments} Parsed args from everything
  */
+// Reads `reporter-option` from a config source under any of its known
+// aliases. Keeps `loadOptions` agnostic of how each source was authored.
+const readReporterOption = (source) =>
```
JoshuaKGoldberg, line 288: [Comments] This `//` comment conflicts with the existing JSDoc for the function. This should add to the JSDoc instead.

lib/cli/options.mjs
```diff
@@ -317,6 +336,20 @@ export const loadOptions = (argv = []) => {
...
+  // precedence. Re-emit lowest→highest priority so CLI overrides config.
+  const orderedReporterOption = [
+    reporterOptionSources.pkg,
+    reporterOptionSources.rc,
+    reporterOptionSources.env,
+    reporterOptionSources.cli,
+  ].reduce((acc, src) => (src ? acc.concat(list(src)) : acc), []);
+  if (orderedReporterOption.length) {
+    args["reporter-option"] = orderedReporterOption;
+  }
```
JoshuaKGoldberg, line 351: [Refactor] This is quite a lot of complexity for a pretty small area of bug. And, we actually refactored away `yargs` & `yargs-parser` on `main` in the meantime. The `mergeConfigObjects` we added on `main` would be a good place for this kind of thing.

JoshuaKGoldberg (commented):
Thanks for the contribution! And sorry it took so long to review. But:

* We've since refactored this area of logic in Mocha significantly. See #6164 & co.
* This was clearly authored with AI in a way that runs counter to the [AI-Generated Code section of our contributing guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#-ai-generated-code)

For just one of those I'd request changes, but for both I'm going to go ahead and close this out. Feel free to send another PR if you'd like though - I left a couple of thoughts on this one. Cheers!

PR #5987, opened 2026-05-22
fix: no re-emit unhandledRejection for non-Mocha registred listener 

### LINKED ISSUE #4743: Re-emit the unhandledReject event
https://github.com/mochajs/mocha/blob/bbf0c11b29544de91a18c1bd667c975ee44b7c90/lib/runner.js#L204

What is the use case for this code here?
I got surprised by seeing the handler executed twice whenever rejecting a promise. For example
```
        Promise.reject('nobody handles me 😈 '); // reject once
        process.on('unhandledRejection', (reason, promise) => {
            console.log(' Rejection ', reason, JSON.stringify(promise)); // this shows up twice
        });
```

I notice the handler was removed(line 198) before the rejection is emitted internally in mocha, but the event seems to be cached, which means the internal emit will trigger the handler to be called anyway right after it is added back (line 206).

Is it intended to be so (what is the use case if this is true) or it is some bug?

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4743

## Overview

Current unhandledRejection handler used to remove itself and re-emit the event so node could apply its default behavior but this made user listeners to run twice when present.
this change skips the re-emit if any other unhandledRejection listener is registered thus preventing duplicates and in the same time keeping the existing no-listener behavior unchanged.

An integration test ensures user listeners are called exactly once.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -222,20 +222,28 @@ class Runner extends EventEmitter {
           reason,
         );
         this.uncaught(reason);
-      } else {
+        return;
+      }
+
+      const hasOtherListener = process
+        .listeners("unhandledRejection")
+        .some((listener) => listener !== this.unhandled);
+
+      if (hasOtherListener) {
         debug(
-          "trapped unhandled rejection from (probably) user code; re-emitting on process",
+          "trapped unhandled rejection from, probably, user code. Another listener is registered and therefore not re-emitting",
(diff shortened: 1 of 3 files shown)


mostafaNazari702:
Added only a space in the end of the PR title to re-run tests.

PR #5991, opened 2026-05-22
fix: report sync error thrown after done() in async tests 

### LINKED ISSUE #4202: 🐛 Bug: Test stops recognizing exceptions once done has been called
### Prerequisites

### Description

Mocha stops recognizing exceptions once done has been called.

### Steps to Reproduce
Find [a complete example](https://github.com/feckertson/mocha-done-issue) in git, but an outline is given below.
## Source
```javascript
exports.handler = function eventHandler(callback) {
  callback(null, 'oops');
}
```
## Test
```javascript
const expect = require('chai').expect;
const subject = require('../mocha-done-issue.js');

it('problematic test', (done) => {
  subject.handler((e, d) => {
    done();
    expect(true).to.be.false;
  });
});
```

**Expected behavior:** We expect the test to fail because of the failed expect assertion.

**Actual behavior:** The test passes, but if the two lines are reversed the test fails.

### Versions
Multiple including mocha 5.2.0/7.1.0, node 8.10.0/10.16.0 (64-bit) on Windows cmd and Linux bash.

### Additional Information

This came to our attention when faulty code similar to the following was submitted for review. Mocha failed to indicate there was a problem with the code even though [Stryker](https://stryker-mutator.io/) reported 100% mutation coverage. 

The expect fails during the extraneous call to callback since the statusCode is 200. Consequently no duplicate call to done occurs. If it weren't for the assert failure mocha would catch the duplicate call to done.

```javascript
exports.handler = function eventHandler(condition, callback) {
    if ( ! condition ) {
        callback(null, {statusCode: 422, body: 'missing required value'});
    }
    callback(null, {statusCode: 200, body: 'OK'});
}
```
```javascript
it('test false condition', (done) => {
    subject.handler(false, (e, d) => {
        expect(d.statusCode).to.equal(422);
        done();
      });
});
```

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4202

## Overview

Tests using done() could incorrectly pass if they called it synchronosly and then immediately threw an error afterward. Caused by the post-done() throw being treated as already handled that silently swallowed the error ( excuse me for the unprofessional way of describing ).

this fix detects throws that happen after done() completes and routes them through mocha’s existing multiple-done error handling causing the test to fail correctly and inntegration tests were added to verify the failure behavior and error message.

### DIFF
--- a/lib/runnable.mjs
+++ b/lib/runnable.mjs
@@ -332,12 +332,16 @@ class Runnable extends EventEmitter {
         callFnAsync(this.fn);
       } catch (err) {
         // handles async runnables which actually run synchronously
-        errorWasHandled = true;
         if (err instanceof PendingError) {
+          errorWasHandled = true;
           return; // done() is already called in this.skip()
         } else if (this.allowUncaught) {
           throw err;
         }
+        if (finished) {
+          return multiple(Runnable.toValueOrError(err));
+        }
+        errorWasHandled = true;
         done(Runnable.toValueOrError(err));
       }
       return;
--- a/test/integration/fixtures/sync-error-after-done.fixture.js
+++ b/test/integration/fixtures/sync-error-after-done.fixture.js
@@ -1,0 +1,8 @@
+'use strict';
+
+var assert = require('assert');
+
+it('should fail in a test-case', function (done) {
+  done();
+  assert.strictEqual(true, false);
+});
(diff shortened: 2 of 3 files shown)


mostafaNazari702:
Added a space in the end of the title to rerun tests.

PR #5994, opened 2026-05-22
fix: surface global setup/teardown errors with a non-zero exit code

### LINKED ISSUE #5208: 🐛 Bug: Error in mochaGlobalTeardown is swallowed and exits with code 0
### Bug Report Checklist

### Expected

The error should be logged and the tests should exit with a non-zero exit code

### Actual

The error is swallowed and the tests always exit with a 0 exit code (even if some tests failed)

```
// test.js
const { it } = require("mocha");

it("should work", () => {
  throw new Error("It doesn't work");
});
```

```
// setup.js
exports.mochaGlobalTeardown = async function () {
  throw new Error("Teardown problem");
};
```

```
$ npx mocha -r setup.js test.js
```

### Versions

Mocha v10.7.3, Node v20.15.1

### Additional Info

_No response_

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5208

## Overview

When mochaGlobalSetup or mochaGlobalTeardown threw, the error escaped the runAsync(runner) promise chain and became an unhandled rejection. since there was no catch, "done" never ran, the failure never reached the runner, and mocha exited with code 0 while silently swallowing the error.

we now catch errors only at the runAsync boundary.
Normal runner errors are unchanged as before, only global fixture errors are handled here.

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -1060,7 +1060,7 @@ Mocha.prototype.run = function (fn) {
 
   // Errors coming out of Runner#run itself remain uncaught/unhandled and are picked up by the `process` event listeners.
   // Only global fixture errors are intercepted above so they surface to the user instead of being silently "swallowed".
-// Also: returning anything other than `runner` would be a breaking change.
+  // Also: returning anything other than `runner` would be a breaking change.
   runAsync(runner).then(done);
 
   return runner;
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -1027,24 +1027,39 @@ Mocha.prototype.run = function (fn) {
     }
   };
 
+  const reportFixtureError = (phase, err) => {
+    process.stderr.write(
+      `\n[mocha] global ${phase} failed:\n${err && err.stack ? err.stack : err}\n`,
+    );
+  };
+
   const runAsync = async (runner) => {
-    const context =
-      this.options.enableGlobalSetup && this.hasGlobalSetupFixtures()
-        ? await this.runGlobalSetup(runner)
-        : {};
+    let context = {};
+    if (this.options.enableGlobalSetup && this.hasGlobalSetupFixtures()) {
+      try {
+        context = await this.runGlobalSetup(runner);
+      } catch (err) {
+        reportFixtureError("setup", err);
+        return 1;
+      }
+    }
     const failureCount = await runner.runAsync({
       files: this.files,
       options,
     });
     if (this.options.enableGlobalTeardown && this.hasGlobalTeardownFixtures()) {
-      await this.runGlobalTeardown(runner, { context });
+      try {
+        await this.runGlobalTeardown(runner, { context });
+      } catch (err) {
+        reportFixtureError("teardown", err);
+        return failureCount + 1;
+      }
     }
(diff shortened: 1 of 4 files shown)


test/integration/plugins/global-fixtures.spec.js
```diff
@@ -80,6 +81,59 @@ describe("global setup/teardown", function () {
       });
     });
 
+    describe("when a global fixture throws (issue #5208)", function () {
```
JoshuaKGoldberg: [Docs] We don't really do this here - the issue isn't important (and is trackable in the Git history), and cases/scenarios tend to morph over time.

```suggestion
    describe("when a global fixture throws", function () {
```

test/integration/plugins/global-fixtures.spec.js
```diff
@@ -80,6 +81,59 @@ describe("global setup/teardown", function () {
...
+        );
+      });
+
+      it("should skip tests and teardown when setup throws", async function () {
+        const res = await runWithFixture(
+          "plugins/global-fixtures/global-setup-throws",
+        );
+        expect(res.output, "not to match", /teardown should not run/);
+        expect(res.output, "not to match", /passing/);
+      });
```
JoshuaKGoldberg, line 122: [Testing] In the added tests, there are two pairs of tests that run the same fixture: `global-teardown-throws` and `global-setup-throws`. I'm leaning towards just combining them. As much as I like granular reporting, they're _really_ similar tests.

lib/mocha.js
```diff
@@ -1027,24 +1027,39 @@ Mocha.prototype.run = function (fn) {
...
     return failureCount;
   };
 
-  // no "catch" here is intentional. errors coming out of
-  // Runner#run are considered uncaught/unhandled and caught
-  // by the `process` event listeners.
+  // errors coming out of Runner#run itself remain uncaught/unhandled and are picked up by the `process` event listeners.
+  // Only global fixture errors are intercepted above so they surface to the user instead of being silently "swallowed".
   // also: returning anything other than `runner` would be a breaking
   // change
```
JoshuaKGoldberg: [Docs] The modified comments have a longer line width but are right next to a line with the original smaller width! It looks a little odd. Also, inconsistent capitalization. Now that we're on full sentences...

```suggestion
  // Errors coming out of Runner#run itself remain uncaught/unhandled and are picked up by the `process` event listeners.
  // Only global fixture errors are intercepted above so they surface to the user instead of being silently "swallowed".
// Also: returning anything other than `runner` would be a breaking change.
```

test/integration/plugins/global-fixtures.spec.js
JoshuaKGoldberg, line 1: [Praise] This test coverage is very thorough 🙂 nice!

JoshuaKGoldberg (approved):
Looks great! Nicely done on the comprehensive tests.

Just leaving a few minor notes, nothing that would block merging if someone has a strong case against them. I'll leave this open for a bit in case Mark wants to review.

mark-wiemer:
Known failures (both #6037 and #6036), merging :)

PR #5995, opened 2026-05-22
Add a built-in require interface without globals

### LINKED ISSUE #956: node-able tests
@isaacs is complaining that Mocha tests aren't `node`-able. I think this is a silly complaint, but I think he represents a vocal minority, so shutting them up would be possibly valuable.

Here is what I envision: instead of being able to use e.g. `describe`, `it`, etc. by default, you'd do

```js
var mochaBDD = require("mocha/bdd");
var describe = mochaBDD.describe;
var it = mochaBDD.it;

// ok, stupid boilerplate is over but at least I don't have to use a test runner, woohoo!
```

Alternately it could be just

```js
require("mocha/bdd");
```

and that would set up some globals for you.

What do you think?

### LINKED ISSUE #2982: test "require interface"
We're currently testing the ability to `require("mocha").it` and `require("mocha").describe` but none of the other exported interface functions: https://github.com/mochajs/mocha/blob/f20de56637b2223f614ce40adc0d41a58030f042/test/unit/required-tokens.spec.js

Full list here: https://github.com/mochajs/mocha/blob/2bc9c4d1d638c8a95250757dd23724f788918abe/lib/mocha.js#L200-L214

I think the test file above should just use the global `describe` and `it`, but inside the `it` we should assert that the exports all are equal to the corresponding globals. It would be "more brittle" in the sense that we would have to change the test if we wanted to change the implementation from a simple reference copy to having separate global and exported functions, but I have a hard time imagining why we'd want/need to do that at the moment, and equality comparison would be fine for the current implementation and *waaaaay* easier than coming up with a way to use every last one of them and check for the correct behavior (and the current test doesn't even do the correct-behavior part for the two functions it does try using).

While we're in there, let's remove the `done` since the test is (and would still be) synchronous.

Note that once #2972 is resolved we should include `xit` (and whatever other `x` functions we have) in the new test.

### LINKED ISSUE #2998: Export all BDD functions for "require interface"
From my research in #2997, the "`require` interface's" exports are still missing:
- `specify` (alias of `it`)
- `xspecify` (alias of `xit`)
- `xdescribe` (can use `context.suite.skip` on the TDD side of the `||`)
- `context` (`context.context` in this code; alias of `describe`)
- `xcontext` (alias of `xdescribe`)

It appears that the BDD functions are exported using `||` to separate BDD and TDD functions they're taken from, basically preferring to take the BDD functions but exporting the TDD functions under BDD names as a fallback, if I've understood the code correctly.

So, each of these should take from the BDD function with the same name; the ones marked "alias of" another should take from the same TDD function as the ones they're aliases of.

These additions don't need tests added unless #2982 is resolved first; if #2982 is resolved first then the added functions should be added to the new/improved test that resolves #2982.

lo1tuma:
## PR Checklist

- [x] Addresses an existing open issue: fixes #956
- [x] Addresses an existing open issue: fixes #2982
- [x] Addresses an existing open issue: fixes #2998

## Overview

This adds `require` as a built-in interface so `--ui require` and `ui: require` work consistently from both the CLI and config files.

The interface keeps Mocha's DSL helpers off the global object and exposes the function-style APIs through `require("mocha")` and `import { ... } from "mocha"`, including mixed BDD/TDD usage such as `it` with `suite`.

It also fills in the missing require-interface exports, extends the require-interface coverage to exercise the full callable API surface, updates `--list-interfaces`, and refreshes the docs for the require interface.


codeCraft-Ritik (commented):
Nice enhancement! Making require a first-class built-in interface improves consistency between CLI and configuration-based usage. The additional export coverage and documentation updates help make the feature feel complete rather than just a configuration change.👍

mark-wiemer:
Thanks @lo1tuma . We're in a feature freeze as we wait for Mocha 12 RC feedback. Once 12.0.0 stable is released, we can check this one out. I may pull out the test changes first to reduce the PR size, but this is something I hope to be able to review eventually.

lo1tuma:
> Thanks @lo1tuma . We're in a feature freeze as we wait for Mocha 12 RC feedback. Once 12.0.0 stable is released, we can check this one out. I may pull out the test changes first to reduce the PR size, but this is something I hope to be able to review eventually.

No pressure. I am already excited about v12, so thanks for pushing that. Just let me know when I should do something like splitting PRs or rebasing etc.

docs/src/content/docs/explainers/run-cycle-overview.mdx
```diff
@@ -20,7 +20,7 @@ In a browser, test files are loaded by `<script>` tags, and calling `mocha.run()
    2. If not, Mocha ignores any exports of a `--require`'d module
 6. Mocha validates any custom reporters or interfaces which were loaded via `--require` or otherwise
 7. Mocha _discovers_ test files; when given no files or directories, it finds files with extensions `.js`, `.mjs` or `.cjs` in the `test` directory (but not its children), relative to the current working directory
-8. The (default) [bdd interface](/interfaces/bdd) loads the test files _in no particular order_, which are given an interface-specific `global` context (this is how, e.g., `describe()` ends up as a global in a test file)
+8. The selected interface loads the test files _in no particular order_. Most interfaces use an interface-specific `global` context, which is how, e.g., `describe()` ends up as a global in a test file. The [`require` interface](/interfaces/require) is the exception: it keeps the DSL off `globalThis`, and test files import it from `mocha` instead.
```
JoshuaKGoldberg: [Docs] "DSL" isn't a term every dev knows. And it's not necessary to introduce it here; folks can understand things without it. How about...

```suggestion
8. The selected interface (by default, [the bdd interface](/interfaces/bdd)) loads the test files _in no particular order_. Most built-in interfaces add functions to the `global` context, which is how, e.g., `describe()` ends up as a global in a test file. The [`require` interface](/interfaces/require) is the exception.
```

docs/src/content/docs/interfaces/require.mdx
```diff
@@ -3,26 +3,51 @@ description: The require interface for Mocha.
 title: Require
 ---
 
-The `require` interface allows you to require the `describe` and friend words directly using `require` and call them whatever you want.
-This interface is also useful if you want to avoid global variables in your tests.
+The `require` interface lets you import Mocha's test DSL from the `mocha` module instead of relying on globals.
```
JoshuaKGoldberg:
```suggestion
The `require` interface lets you import Mocha's test functions from the `mocha` module instead of relying on globals.
```

docs/src/content/docs/interfaces/require.mdx
```diff
@@ -3,26 +3,51 @@ description: The require interface for Mocha.
...
+This opts out of Mocha injecting `describe()`, `it()`, `suite()`, and related helpers into the global scope.
+
+Because the DSL comes from the module, you can mix the function-style interfaces in the same file.
```
JoshuaKGoldberg:
```suggestion
Because the test functions comes from the `mocha` module, you can mix the function-style interfaces in the same file.
```

lib/mocha.cjs
```diff
@@ -136,6 +136,9 @@ exports.describe.skip = function (...args) {
     args,
   );
 };
+exports.context = exports.describe;
+exports.xdescribe = exports.describe.skip;
+exports.xcontext = exports.describe.skip;
```
JoshuaKGoldberg: [Question] Why these changes? The `require` interface already exists, what point is there to changing the default exports?

PR #5997, opened 2026-05-22
fix: split err.stack so embedded stacks in err message render once

### LINKED ISSUE #3992: 🐛 Bug: Reports the Error twice when its message contains something similar to a stack

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #3992

## Overview

getFullErrorStack used to split err.stack by searching for err.message in it. that breaks when mocha filters stack traces and rewrites file paths, message stays unchanged but the stack changes so indexOf fails and the error gets printed twice (message + full stack).

What we do now is instead of matching the message string we now parse the stack the V8 way: treat the trailing "at ......." lines as the stack frames and everything above as the message. If the stack doesn't match V8 format we fall back to the old behavior.
We also keep err.inspect() overrides and err.cause chaining working as before.

### DIFF
--- a/lib/reporters/base.js
+++ b/lib/reporters/base.js
@@ -402,36 +402,63 @@ var getFullErrorStack = function (err, seen) {
   }
 
   var message;
+  var usedInspect = false;
 
   if (typeof err.inspect === "function") {
     message = err.inspect() + "";
+    usedInspect = true;
   } else if (err.message && typeof err.message.toString === "function") {
     message = err.message + "";
   } else {
     message = "";
   }
 
-  var msg;
-  var stack = err.stack || message;
-  var index = message ? stack.indexOf(message) : -1;
-
-  if (index === -1) {
-    msg = message;
-  } else {
-    index += message.length;
-    msg = stack.slice(0, index);
-    // remove msg from stack
-    stack = stack.slice(index + 1);
-
-    if (err.cause) {
-      seen = seen || new Set();
-      seen.add(err);
-      const causeStack = getFullErrorStack(err.cause, seen);
-      stack +=
-        "\n   Caused by: " +
-        causeStack.msg +
-        (causeStack.stack ? "\n" + causeStack.stack : "");
+  var rawStack = err.stack || message;
+  var lines = rawStack.split("\n");
+  var lastLine = lines.length - 1;
+  while (lastLine >= 0 && lines[lastLine] === "") {
+    lastLine--;
+  }
+  var frameStart = lastLine + 1;
+  for (var i = lastLine; i >= 0; i--) {
+    if (/^\s+at\s/.test(lines[i])) {
+      frameStart = i;
+    } else {
+      break;
     }
+  }
+
+  var msg;
+  var stack;
+  var splitSucceeded = false;
+  if (frameStart <= lastLine) {
+    stack = lines.slice(frameStart, lastLine + 1).join("\n");
+    msg =
+      usedInspect || frameStart === 0
+        ? message
+        : lines.slice(0, frameStart).join("\n");
+    splitSucceeded = true;
+  } else {
+    var index = message ? rawStack.indexOf(message) : -1;
+    if (index === -1) {
+      msg = message;
+      stack = rawStack;
+    } else {
+      index += message.length;
+      msg = rawStack.slice(0, index);
+      stack = rawStack.slice(index + 1);
+      splitSucceeded = true;
+    }
+  }
+
+  if (splitSucceeded && err.cause) {
+    seen = seen || new Set();
+    seen.add(err);
+    const causeStack = getFullErrorStack(err.cause, seen);
+    stack +=
+      "\n   Caused by: " +
+      causeStack.msg +
+      (causeStack.stack ? "\n" + causeStack.stack : "");
   }
 
   return {
--- a/test/reporters/base.spec.cjs
+++ b/test/reporters/base.spec.cjs
@@ -491,6 +491,66 @@ describe("Base reporter", function () {
     expect(errOut, "to be", "1) test title:\n     Error\n  foo\n  bar");
   });
 
+  describe("message and stack splitting", function () {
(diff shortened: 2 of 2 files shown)


test/reporters/base.spec.cjs
JoshuaKGoldberg, line 1: [Testing] This is really tricky stuff. Stack frames get surprisingly weirdly crafted in practice. A couple more test cases to add...

---

When the stack _only_ contains frames:

```js
var err = {
  message: "Error",
  stack: "    at foo (foo.js:1:1)\n    at bar (bar.js:2:2)",
  showDiff: false,
};
var test = makeTest(err);
```

Error output should be something like `1) test title:\n     Error\n      at foo (foo.js:1:1)\n      at bar (bar.js:2:2)`.

---

When the embedded stack has a trailing newline, like:

```js
var embeddedMessage =
  "An error occured with following trace:\n\nError\n    at inner (/original/foo.js:1:1)";
var stackMessage = embeddedMessage.replace(/\/original\//g, "/filtered/");
var err = {
  message: embeddedMessage,
  stack:
    "Error: " + stackMessage + "\n    at runTest (lib/runner.js:1:1)\n",
  showDiff: false,
};
```

The output still should smartly parse & not duplicate.

lib/reporters/base.js
```diff
@@ -404,38 +404,58 @@ var getFullErrorStack = function (err, seen) {
...
     message = "";
   }
 
-  var msg;
-  var stack = err.stack || message;
-  var index = message ? stack.indexOf(message) : -1;
+  var rawStack = err.stack || message;
+  var lines = rawStack.split("\n");
+  var frameStart = lines.length;
+  for (var i = lines.length - 1; i >= 0; i--) {
```
JoshuaKGoldberg: [Bug] For an Error-like object with a stack containing only V8 frame lines:

```js
{
  message: "Boom",
  stack: "    at fn (foo.js:1:1)\n    at run (foo.js:2:2)"
}
```

This'll make `msg` `""`.

test/reporters/base.spec.js
```diff
@@ -568,6 +568,31 @@ describe("Base reporter", function () {
       );
     });
 
+    it("should not duplicate the message when it contains an embedded stack", function () {
```
JoshuaKGoldberg: [Testing] This passes on the original implementation:

```shell
git checkout main -- test/reporters/base.spec.js
npx mocha test/reporters/base.spec.js
```

The test should simulate the actual mismatch from the issue: `err.message` contains an embedded stack, but `err.stack` has the embedded stack slightly rewritten.

lib/reporters/base.js
```diff
@@ -404,38 +404,58 @@ var getFullErrorStack = function (err, seen) {
...
     message = err.message + "";
   } else {
     message = "";
   }
 
-  var msg;
-  var stack = err.stack || message;
-  var index = message ? stack.indexOf(message) : -1;
+  var rawStack = err.stack || message;
+  var lines = rawStack.split("\n");
```
JoshuaKGoldberg, line 417: [Bug] Make sure this handles a trailing `\n` well - we don't want the last line to be `""`.

JoshuaKGoldberg (dismissed):
Tricky stuff 😎

mark-wiemer:
@JoshuaKGoldberg I think you're best to handle this, I can review if you really want me to :D

JoshuaKGoldberg (approved):
LGTM, very nice. Thanks! ✨

Issue #5505, opened 2025-10-21
🐛 Bug: thrown error `.expected` with `null` prototype throws `TypeError: value.toString is not a function`


alxndrsn:
### Bug Report Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [x] I want to provide a PR to resolve this

### Expected

Test should fail.

### Actual

Suite stops mid-way, returning status code zero.

### Minimal, Reproducible Example

# `package.json`

```json
{
  "name": "mocha-bug",
  "packageManager": "yarn@4.9.2",
  "devDependencies": {
    "chai": "^6.2.0",
    "mocha": "^11.7.4"
  },
  "volta": {
    "node": "24.10.0"
  }
}
```

# `test.js`

```js
const { assert } = require('chai');

describe('deepEqual()', () => {
  const o = { a:true };

  describe('sync', () => {
    it('should not equal null', () => { assert.deepEqual(o, null); });
    it('should not equal plain object', () => { assert.deepEqual(o, {}); });
    it('should not equal null-prototyped', () => { assert.deepEqual(o, Object.create(null)); });
  });

  describe('async', () => {
    it('should not equal null', async () => { assert.deepEqual(o, null); });
    it('should not equal plain object', async () => { assert.deepEqual(o, {}); });
    it('should not equal null-prototyped', async () => { assert.deepEqual(o, Object.create(null)); });
  });
});
```

# execute:

```sh
$ yarn mocha test.js 


  deepEqual()
    sync
      1) should not equal null
      2) should not equal plain object
      3) should not equal null-prototyped
    async
      4) should not equal null
      5) should not equal plain object
$ echo $?
0
```

### Versions

node 24.10.0
chai 6.2.0
mocha 11.7.4

alxndrsn:
[...]

PR #5998, opened 2026-05-22
prevent poisoned diffs from suppressing non-zero exits 

### LINKED ISSUE #5507: 🐛 Bug: unhandledRejection in the default Runner kills process with 0 status code
### Bug Report Checklist

### Expected

Tests should **fail** if there's an unhandled rejection with no process-level listener.

### Actual

As demonstrated in #5505, a test which throws an unhandled promise rejection inside runner code will cause the mocha process to exit with status code 0.

```sh
$ echo 'it("t", async () => { require("node:assert").deepEqual(Object.create(null), { a:1 }); });' > demo.js; npx mocha demo.js; echo $?


0
```

### Versions

```
$ node --version
v22.14.0

$ npx mocha --version
11.7.4
```

### Additional Info

_No response_

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5507

## Overview

This PR is the last of the wave. there was was a broader issue behind #5505/#5506 where reporter-side errors could cause nocha to exit with status 0 even though the run actually failed.

if the Base reporter threw while handling a failed test. either during EVENT_TEST_FAIL or later in Base.list during the epilougue , the exception could break mocha's internal runner flow before the final failure count was reported to the CLI. The run looked failed in output, but the process still exited successfully.
the original repro using Object.create(null) was fixed already but i guess the underlying issue still existed for anything that could make utils.stringify or diff rendering throw (hostile proxies, throwing getters, weird objects, etc......). 

we now got defensive handling in the lib/reporters/base.js file at three levels, stringifyDiffObjs now safely falls back if object stringification throws. the EVENT_TEST_FAIL listener is wrapped so reporter errors are logged to stderrr instead of breaking the run. Base.list now isolates rendering failures per test so one broken failure output doesn’t abort the entire epilogue


mostafaNazari702:
I used some questionable terminologies such as reslience but i want you guys to evaluate, your opinions matter as well. I just couldn't find a better word to have instead there.

mostafaNazari702:
So i while i was typing "Added a space in the end of the PR title to re-run tests", the error to tests even after re-running, appeared...Can you guys tell if that is an issue caused on my side?

JoshuaKGoldberg (changes requested):
@mostafaNazari702 this has lint failures, I'll wait to review until they're fixed.

https://github.com/mochajs/mocha/actions/runs/26314852116/job/77471707014?pr=5998

```plaintext
/home/runner/work/mocha/mocha/lib/reporters/base.js
Error:   106:17  error  Empty block statement  no-empty
Error:   110:17  error  Empty block statement  no-empty
Error:   490:15  error  Empty block statement  no-empty
```

test/integration/reporter-crash.spec.js
```diff
@@ -0,0 +1,34 @@
...
+      res.output,
+      "to match",
+      /\[mocha\] (reporter error|failed to render)/,
+    );
+  });
+
+  it("should report the failure exactly once, not double-count the reporter throw", function () {
+    expect(res.output, "to match", /1 failing/);
+    expect(res.output, "not to contain", "2 failing");
+  });
```
JoshuaKGoldberg: [Testing] These are all asserting on various aspects of a single result. That's just one unit under test. Please combine these into one unit test.

JoshuaKGoldberg (changes requested):
Yup, this looks good to me. It's hard to verify the kind of defensive coding the issue requires but I think you got it all right. Just the one test for cleanup and I think we'd be good to go! (not including any other maintainer review)

mark-wiemer:
See backing issue

mostafaNazari702:
The backing issue is no longer valid.

PR #5999, opened 2026-05-23
fix: handle EPERM symlink error on Windows non-admin sessions

### LINKED ISSUE #5989: 🐛 Bug: Misleading "setup failed: EPERM" from config.spec.js on Windows non-admin sessions
### Bug Report Checklist

### Expected

When integration suite is run on a Windows account that does not have "SeCreateSymbolicLinkPrivilege" (basically any standard user with neither Administrator elevation nor Windows Developer Mode enabled), the `config > when path is neither absolute or relative` block in `test/integration/config.spec.js` should skip its dependet test without emitting an alarming stderr stack trace that looks like a real test failure.

I see that codebase already has the right pattern at
test/integration/file-utils.spec.js:

```js
const SYMLINK_SUPPORT = process.platform !== "win32";
```

### Actual

The before hook at test/integration/config.spec.js catches "EEXIST" but no other errors. on Windows non-admin sessions, fs.symlinkSync throws EPERM (errno: -4048) which falls into the else branch and returns:

```txt
setup failed: Error: EPERM: operation not permitted, symlink
  'C:\...\test\integration\fixtures\config\mocha-config' ->
  'C:\...\node_modules\mocha-config'
    at Object.symlinkSync (node:fs:1869:11)
    at Context.<anonymous> (.../test/integration/config.spec.js:85:14)
    ...
```

The dependent test is still correctly skipped, because installedLocally remains false and the spec calls this.skip() , so the suite stays green but the stderr block at the very start of every Windows integration run looks like a hard failure. this is misleading both to human readers and to log scraping CI bots, that i guess are used in any tests in PRs when opened in Mocha github repository page. Not an issue about whether the test should run on Windows, the test already correctly skips. it is about the spurios noise.

On a Windows 11 account without Administrator elevation and without Developer Mode enabled (the default state of a fresh contributor machine):

```txt
git clone https://github.com/mochajs/mocha.git
cd mocha
npm install
npm run test-node:integration 2>&1
```
Or a lot faster:
```javascript
node -e "const fs = require('node:fs'),
    os = require('node:os'),
    p = require('node:path');
const t = fs.mkdtempSync(p.join(os.tmpdir(), 'sl-'));
fs.mkdirSync(p.join(t, 'src'));
try {
    fs.symlinkSync(p.join(t, 'src'), p.join(t, 'link'), 'dir');
    console.log('OK');
} catch (e) {
    console.log('FAIL:', e.code);
}"
```
Returns: "FAIL: EPERM"

### Versions

Mocha version: 12.0.0-beta-9.4
Node version: v24.15.0'
OS: Windows 11 Pro (Version 10.0.26200.8457) standard user + no Developer Mode.
Shell: PowerShell 7

### Additional Info

I can provide a PR to this issue, if green-lit by maintainers.

xxiaoxiong:
Fixes #5989

Why
On Windows accounts without "SeCreateSymbolicLinkPrivilege" (standard users without admin or Developer Mode), `fs.symlinkSync()` throws EPERM instead of EEXIST. The test setup only caught EEXIST, causing the error to fall through to a misleading console.error stack trace that looks like a real test failure.

Changes
Added an `EPERM && process.platform === "win32"` guard that gracefully logs that symlinks are unavailable and skips the dependent test, matching the existing pattern in `test/integration/file-utils.spec.js`.

Updated component: `test/integration/config.spec.js`

Testing
On Windows non-admin sessions, the before hook no longer emits a scary stack trace. The test continues without the symlinked config, exactly as intended.

Surface area
- Integration tests
- Bug fix
- Windows-specific

### DIFF
--- a/test/integration/config.spec.js
+++ b/test/integration/config.spec.js
@@ -89,6 +89,8 @@ describe("config", function () {
           if (err.code === "EEXIST") {
             console.log("setup:", 'package already exists in "node_modules"');
             installedLocally = true;
+          } else if (err.code === "EPERM" && process.platform === "win32") {
+            console.log("setup:", "symlinks not available (non-admin Windows); skipping");
           } else {
             console.error("setup failed:", err);
           }


xxiaoxiong:
Closing due to EasyCLA signing requirement

PR #6008, opened 2026-05-23
docs: expand Communication section of maintenance guide

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes mochajs/admin-private#33

## Overview

The guide previously had a separate section around being courteous and kind to external contributors (good!). But it didn't have details around what that means, such as de-escalating or not engaging in aggressiveness. This adds/unifies that information under the existing `## Communication` section.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -175,6 +175,36 @@ This is a _public_ Discord, and _anyone_ can join.
 Videoconference (or audio) calls may happen on a regular or irregular basis, as schedules allow.
 This is mainly because we have Real Lives and time zones suck.
 
+### Courtesy and Kindness
+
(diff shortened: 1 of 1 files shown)


voxpelli (changes requested):
Why do we insist on creating both issues and PR:s when a PR would be enough? I commented on the issue but if I don't also comment here that comment will be easy to miss.

Issues are great if there needs to be a discussion before a PR gets opened.

If there's already a PR ready then we should do PR without issue and only create issue if the PR needs to go back to discussion.

JoshuaKGoldberg:
(closing for now as we're going to continue to chat about this stuff internally)

PR #6015, opened 2026-05-24
Convert worker pool modules to ESM

### LINKED ISSUE #5969: 🛠️ Repo: Convert `lib/nodejs/` worker pool files to ESM
Sub-issue of #5400 (Switch from CJS to ESM internally).

These `lib/nodejs/` files are the hardest remaining CJS-to-ESM migration targets. They are intentionally left as CJS for now because of a test-isolation blocker described in [this comment on #5400](https://github.com/mochajs/mocha/issues/5400#issuecomment-3049098693).

### Files to convert

- `lib/nodejs/buffered-worker-pool.js`
- `lib/nodejs/parallel-buffered-runner.js`
- `lib/nodejs/worker.mjs`
- `lib/nodejs/reporters/parallel-buffered.js`

### Blocker: `rewiremock` / `Module._load` test isolation

The unit tests for these files use `rewiremock.proxy()` to replace `workerpool` with a fake implementation. This works in CJS because `rewiremock` hooks into `Module._load`. Once the files become `.mjs`, ESM imports bypass `Module._load` and the mock stops applying — causing tests to spawn real worker processes during unit tests.

Options to resolve:
- Replace `rewiremock` with [`esmock`](https://github.com/iambumblehead/esmock) for ESM-compatible mocking
- Introduce a workerpool abstraction/factory that can be injected in tests
- Other dependency injection approaches

The simpler `lib/nodejs/` files (`file-unloader.js`, `serializer.js`, `worker.js`) do not have this blocker and are tracked in #5968.

macayu17:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5969

## Overview

Converts the `lib/nodejs` worker pool modules from CommonJS to ESM.

The worker pool, parallel runner, worker entry point, and buffered worker reporter now use the repository's current `.js` ESM naming. The old `.cjs` files are removed, and the worker runtime is kept separate from the process entry point so the existing behavior can be tested without CommonJS module mocking.

Validated with the focused worker tests, full Node.js unit suite, lint, formatting, type checking, and smoke tests.


mostafaNazari702:
Status?

macayu17:
Will fix the conflicts

mark-wiemer:
Please update to the latest naming conventions, it looks like this adds way more lines than necessary, did you mean to rename/remove old files?

macayu17:
Thanks, you were right. I’ve updated the naming, removed the old files, and cleaned up the extra changes. Everything is up to date and tested now.

test/node-unit/parallel-buffered-runner.spec.cjs
```diff
@@ -476,6 +459,50 @@ describe("parallel-buffered-runner", function () {
...
+                    ]).and("was called twice");
+                    done();
+                  },
+                  {
+                    files: ["some-file.js", "some-other-file.js"],
+                    options,
+                  },
+                );
+              });
+            });
```
mostafaNazari702, line 505: This is already in main and there is an identical copy under "bail flag" (in your line 548). We can just delete this one and coverage stays the same and it keeps it out of the wrong section.

```suggestion
```

mostafaNazari702 (commented):
Nice work this is the first one that actually gets these files off rewiremock. I ran it on top of current main and the tests i ran all match main including the full `--parallel` integration run and the #6268 worker-death test.

I left you one suggestion on the runner spec. The rest is optional:
- The reporter factory lets you pass in the two serializer classes but nothing ever does so i think only the base reporter needs to be swappable.

- The changelog tells people writing their own worker reporters to extend this reporter so (for the sake of documentation) maybe mention in the description that its file name and exports change. #6260 did the same kind of change to the runner.

mostafaNazari702:
Forgot to mention that Mark asked me to look through some of the _ESM conversion_-PRs so this is just my take not his , you two have the final say.

Issue #5445, opened 2025-08-24
📝 Docs: Review recommended external tools


mark-wiemer:
### Documentation Request Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

Per #5442 we currently are recommending libraries in various places:
- [wiki homepage](https://github.com/mochajs/mocha/wiki)
- [Spies page](https://mochajs.org/next/explainers/spies/) which recommends a should.js, a library deprecated in 2020
- [Third-Party Reporters page](https://mochajs.org/next/reporters/third-party/)
- probably other places too...

We should have guidelines around:
- when to mention external tools
- when to remove mentions of external tools
    - lack of compatibility with latest Mocha for X months?
    - tool officially deprecated?
    - tool claims it will only be compatible with Mocha version X and never Mocha version X+1?
    - any other reasons?
- how to remove mentions of external tools
   - reach out to tool owner to confirm deprecation?
   - find a suitable replacement?
- when to mark external tools as "recommended" vs just listing them
   - this one is very subjective 😬  maybe we never do this? if we list it, it should be good enough, and users can decide on their own?

Something else to consider: create a page for folks to request we mention their tool? We may not have capacity for this, just throwing it out there.

Acceptance criteria for closing this issue:
- all pages with recommendations identified
- old recommendations reviewed and updated
   - removed during review
   - re-added if approved by Mocha maintainers following new guidelines

JoshuaKGoldberg:
Agreed, this definitely deserves an audit and cleaned up practices from us.

> create a page for folks to request we mention their tool? We may not have capacity for this, just throwing it out there.

Also agreed. Maybe just ask them to mention it on the Discord? I suspect asking for a GitHub Issue is a bit more heavyweight than needed.

mark-wiemer:
[...]

PR #6016, opened 2026-05-24
docs: modernize spies example, drop should.js

gaurav0107:
## PR Checklist

- [x] Addresses an existing open issue: works on #5445

## Overview

The Spies explainer was flagged in #5445 as recommending should.js,
which has been deprecated since 2020. On a closer read the page also
contained syntactically broken examples:

- `spy.called.should.equal.true;` is a no-op (reads a property and
  drops the value); the correct should.js form would be
  `spy.called.should.be.true` or `spy.called.should.equal(true)`.
- The "without any special spy library" example claimed to use no
  library yet relied on `.should.equal(...)` from should.js.

This PR:

- Switches the first example to use sinon-only assertions
  (`sinon.assert.calledOnce`, `sinon.assert.calledWith`); the page
  already uses these on the second test in the same block.
- Switches the "no library" example to `node:assert`, matching the
  rest of the docs site (e.g. `getting-started.mdx` uses `node:assert`).
- Drops the should.js mention and link from the prose.

The broader audit of should.js mentions across other docs pages
remains tracked under #5445; this PR is intentionally scoped to the
spies explainer to keep the change focused.

No library code or workflow changes.

Note: #5445 is a meta-tracking issue per the maintainer's comment,
so this PR uses `fixes` to satisfy the PR template format but does
not aim to fully close the broader audit.

### DIFF
--- a/docs/src/content/docs/explainers/spies.mdx
+++ b/docs/src/content/docs/explainers/spies.mdx
@@ -3,7 +3,7 @@ description: How to use spies with Mocha for testing callbacks and function call
 title: Spies
 ---
 
-Mocha does not come equipped with spies, though libraries like [Sinon](https://github.com/sinonjs/sinon) provide this behaviour if desired. The following is an example of Mocha utilizing Sinon and [Should.js](https://github.com/shouldjs/should.js) to test an EventEmitter:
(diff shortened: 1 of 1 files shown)


JoshuaKGoldberg (approved):
LGTM, thanks. Leaving open for a second review.

Btw, a quick note on the PR description - there's no need to wrap text to a max column width like that. AIs do it a lot but it harms readability a bit. #6035

mark-wiemer (approved):

PR #6011, opened 2026-05-23
chore: convert `lib/nodejs/` utility files to ESM

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5968
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Converts the two files and touches up places that refer to them. Note the `package.json` changes.

PR #6055, opened 2026-06-10
fix: require serializer.mjs in parallel worker pool files

### LINKED ISSUE #6054: 🐛 Bug: parallel mode crashes at startup with `Cannot find module './serializer'` in 12.0.0-beta-9.5
### Bug Report Checklist

### Expected

Upgrading from 12.0.0-beta-9.4 to 12.0.0-beta-9.5, `mocha --parallel` starts up and runs tests exactly as it does in serial mode.

### Actual

Every `--parallel` invocation crashes at startup, before any test file is loaded. The failure is a plain CommonJS module-resolution error, so it reproduces on every platform and every supported Node version. The CLI first prints its entire usage/help screen followed by `✖ ERROR: null`, which makes it look like a command-line usage mistake; the actual error follows:

```
✖ ERROR: null
Error: Cannot find module './serializer'
Require stack:
- node_modules/mocha/lib/nodejs/buffered-worker-pool.js
- node_modules/mocha/lib/nodejs/parallel-buffered-runner.js
- node_modules/mocha/lib/mocha.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1456:15)
    ...
    at Object.<anonymous> (node_modules/mocha/lib/nodejs/buffered-worker-pool.js:18:25)
    ...
    at Mocha.parallelMode (node_modules/mocha/lib/mocha.js:1125:7)
    at new Mocha (node_modules/mocha/lib/mocha.js:259:10)
    at exports.handler (node_modules/mocha/lib/cli/run.js:373:17) {
  code: 'MODULE_NOT_FOUND',
  ...
}
```

The same spec passes without `--parallel`.

```sh
mkdir repro && cd repro && npm init -y
npm install mocha@12.0.0-beta-9.5
mkdir test && echo "it('works', () => {});" > test/a.spec.js

npx mocha test/a.spec.js              # 1 passing
npx mocha --parallel test/a.spec.js   # crashes at startup as above
```

The defect can also be demonstrated without the runner, which is how we smoke-tested it. There are two affected files, not one — the second is only reachable inside the worker processes, so it is masked by the first until that one is fixed:

```sh
node -e "require('mocha/lib/nodejs/buffered-worker-pool.js')"
# Error: Cannot find module './serializer'

node -e "require('mocha/lib/nodejs/reporters/parallel-buffered.js')"
# Error: Cannot find module '../serializer'
```

### Versions

`npx mocha --version` → 12.0.0-beta-9.5 (the published npm tarball), `node --version` → v24.14.1 on macOS locally; also reproduced in CI on ubuntu-latest with Node 20.x and 22.x and on windows-latest with Node 22.x (oss-serverless/osls#350). Default spec reporter, plain CommonJS specs, no transpilers.

Not affected: 11.7.6 (current `latest`) and 12.0.0-beta-9.4, where `lib/nodejs/serializer.js` is still CommonJS and the extensionless requires resolve normally. The first broken release is 12.0.0-beta-9.5, and the bad requires are still present on `main` today, so the staged 12.0.0-beta-9.6 release (#6044) would inherit the breakage.

### Additional Info

The regression comes from #6011, which renamed `lib/nodejs/serializer.js` to `serializer.mjs` and updated one of its three consumers (`lib/nodejs/worker.js:24` now requires `"./serializer.mjs"`), but missed the other two. Extensionless CJS resolution tries `.js`/`.json`/`.node` and never `.mjs`, so both fail unconditionally:

- `lib/nodejs/buffered-worker-pool.js:18` — `require("./serializer")`; loaded in the main process via `Mocha.parallelMode()`, hence the startup crash;
- `lib/nodejs/reporters/parallel-buffered.js:36` — `require("../serializer")`; loaded in the workers, so it crashes parallel runs again once the first file is fixed.

`test/node-unit/serializer.spec.js:9` carries the same stale extensionless require (its sibling `worker.spec.js:5` was updated to `.mjs` in #6011), so the serializer spec itself cannot load on `main` — which may be why CI did not catch this, though we have not dug into the CI configuration to confirm.

The fix is the same one-liner in all three files, mirroring what `worker.js` already does: require `"./serializer.mjs"` explicitly. `require()` of a synchronous ESM graph is supported at the engines floor (`^20.19.0 || >=22.12.0`), and we verified that applying the two `lib/` one-liners to the published 12.0.0-beta-9.5 package makes `--parallel` pass again. We have the branch ready and will open the PR right after filing this. #5969 (converting the worker-pool files themselves to ESM) would subsume this change eventually; this is just the minimal interim fix so the next beta does not ship with parallel mode unusable.

We hit this in oss-serverless CI while bumping our pinned beta from 12.0.0-beta-9.2 (oss-serverless/osls#350): every parallel unit-test job failed at startup on both Linux and Windows, while the serial jobs on the same commit passed.

GrahamCampbell:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6054

## Overview

#6011 renamed lib/nodejs/serializer.js to serializer.mjs and updated one of its three consumers (lib/nodejs/worker.js), but the other two still require the module without an extension. CommonJS resolution never tries .mjs, so both requires fail unconditionally, and as of 12.0.0-beta-9.5 any --parallel run crashes at startup with Cannot find module './serializer' (see #6054 for a minimal reproduction).

This PR updates the three remaining stale requires to reference ./serializer.mjs explicitly, mirroring what worker.js already does:

- lib/nodejs/buffered-worker-pool.js — loaded in the main process via Mocha.parallelMode(); this is the startup crash.
- lib/nodejs/reporters/parallel-buffered.js — loaded in the worker processes; masked by the first failure until that one is fixed.
- test/node-unit/serializer.spec.js — the serializer spec itself cannot currently load on main for the same reason (its sibling worker.spec.js was already updated in #6011).

require() of a synchronous ESM graph is supported at Mocha's engines floor (^20.19.0 || >=22.12.0), so no loader changes are needed. Verified by applying the two lib/ changes to the published 12.0.0-beta-9.5 package: --parallel goes from the startup crash to passing, and serial runs are unaffected.

No new test is added: the fix makes the existing serializer spec loadable again, and any parallel-mode integration test exercises the repaired path by construction. #5969 (converting the worker-pool files themselves to ESM) will subsume these requires eventually; this is the minimal interim fix so the next beta doesn't ship with parallel mode unusable.

### DIFF
--- a/lib/nodejs/buffered-worker-pool.js
+++ b/lib/nodejs/buffered-worker-pool.js
@@ -15,7 +15,7 @@
 
 const serializeJavascript = require("serialize-javascript");
 const workerpool = require("workerpool");
-const { deserialize } = require("./serializer");
+const { deserialize } = require("./serializer.mjs");
 const debug = require("debug")("mocha:parallel:buffered-worker-pool");
 const { createInvalidArgumentTypeError } = require("../errors.mjs");
 
--- a/lib/nodejs/reporters/parallel-buffered.js
+++ b/lib/nodejs/reporters/parallel-buffered.js
@@ -33,7 +33,7 @@ const {
 const {
   SerializableEvent,
   SerializableWorkerResult,
-} = require("../serializer");
+} = require("../serializer.mjs");
 const debug = require("debug")("mocha:reporters:buffered");
 const { Base } = require("../../reporters/base.mjs");
 
--- a/test/node-unit/serializer.spec.js
+++ b/test/node-unit/serializer.spec.js
@@ -6,7 +6,7 @@ const {
   deserialize,
   SerializableEvent,
   SerializableWorkerResult,
-} = require("../../lib/nodejs/serializer");
+} = require("../../lib/nodejs/serializer.mjs");
 
 describe("serializer", function () {
   afterEach(function () {


GrahamCampbell:
The failing test is an existing flaky issue. I've prepared PRs to fix this:

* 11.x: https://github.com/mochajs/mocha/pull/6059
* 12.x: https://github.com/mochajs/mocha/pull/6058

JoshuaKGoldberg (approved):
Thanks for the fix!

PR #6056, opened 2026-06-10
fix: serialize parallel test results without mutating user objects

### LINKED ISSUE #6053: 🐛 Bug: parallel mode serializer mutates objects reachable from a test's error in place, corrupting shared modules for the reused worker and losing the file's results
### Bug Report Checklist

### Expected

In `--parallel` mode, a failing test whose error has extra properties attached is reported the same as in serial mode: the failure appears in the output, objects reachable from the error are left unmodified, and later files in the same worker are unaffected.

### Actual

If the error's property graph reaches an object with function-valued properties (e.g. a built-in module's exports), Mocha's worker-side serializer mutates the **original** objects instead of a copy, and three things go wrong at once:

1. **Shared modules are corrupted.** The serializer deletes every function it encounters from the original objects, so `require('crypto')` permanently loses `randomBytes`, `createHash`, etc. Workers are reused across files, so unrelated tests then fail with `crypto.randomBytes is not a function`.
2. **The file's entire result batch is lost.** The walk reaches `crypto.getRandomValues`, a non-configurable accessor, and the strict-mode `delete` throws. The original test failure is never reported; the only trace is an `Uncaught error outside test suite` whose filtered stack is the single line `at Array.forEach (<anonymous>)`.
3. **Failures land on innocent tests** that merely shared the worker.

Serial mode reports the same failure correctly and leaves `crypto` intact.

### Reproduction

See [GrahamCampbell/mocha-reproducer](https://github.com/GrahamCampbell/mocha-reproducer) — `npm install && npm test`. It contains one spec that throws an error referencing `require('crypto')`, four identical victim specs that assert `crypto.randomBytes` is still a function, and `npm run smoke`, which triggers the defect through `mocha/lib/nodejs/serializer` directly with no test runner involved. The README covers the run modes and the `--jobs ≥ 2` caveat.

Reaching a built-in module from an error is realistic: aws-sdk v2 exposes `AWS.util.crypto.lib = require('crypto')`, so any error referencing an SDK-using application object has such a path.

### Root cause

`SerializableEvent#serialize()` in `lib/nodejs/serializer.js` walks the event's `data` and `error` graphs in place (its own doc comment says it "Modifies this object *in place*"), and `_serialize` executes `delete parent[key]` on the original object for every function-valued property (lines 227–231, "for now, just zap it"). The throw escapes through `SerializableWorkerResult#serialize()`'s `this.events.forEach(...)`, the one frame that survives Mocha's stack filtering.

Suggested fix: serialize into fresh structures instead of mutating the walked objects. Minimal alternative: consult property descriptors instead of reading values (reading invokes getters), skip non-configurable properties, and wrap the `delete` in try/catch so one bad property can't destroy a file's results.

Two sibling hazards live in the same path — `breakCircularDeps` (`lib/utils.js`, from #4552) also mutates originals, and the walk dedupes visited properties by key *name* alone, skipping same-named keys on different objects — happy to file those separately.

### Versions

Hit on 12.0.0-beta-9.2; the serializer logic is identical in 11.7.6 (stable), 12.0.0-beta-9.5 (`serializer.mjs`), and beta-10, so this dates to the original parallel-mode implementation, not the betas. Node 22 (Windows CI) and 24.14.1 (macOS), CommonJS specs, default spec reporter, no transpilers.

GrahamCampbell:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6053

## Overview

In parallel mode, `SerializableEvent#serialize` prepares events for IPC by walking the event's `data` and `error` graphs in place: function-valued properties are `delete`d **from the original objects**. When a failing test's error references shared state (the #6053 repro reaches `require('crypto')`), shared modules are permanently stripped of their functions for later files run by the reused worker, and when the walk reaches a non-configurable property (e.g. the `crypto.getRandomValues` accessor) the strict-mode `delete` throws, aborting `SerializableWorkerResult#serialize` — the whole file's result batch is lost and the original failure is never reported.

This PR replaces the in-place walk with one that builds a fresh, JSON-safe copy and never mutates its inputs:

- Functions are omitted from the copy instead of deleted from originals, so non-configurable and frozen properties can no longer throw.
- Cycle handling moves into the copy walk (object-identity based, writing the same `[Circular]` marker into the copy). The `breakCircularDeps(result.error)` call is removed: it mutated original error graphs, and it crashes outright on ESM module namespace objects (`TypeError: Cannot assign to read only property ... of '[object Module]'` — namespace exports report `writable: true` but reject `[[Set]]`), the `import *` twin of the reported bug. The exported util itself is untouched.
- The old dedupe keyed on property *name* — which skipped same-named properties on different objects (including anything literally named `data` or `error`, pre-seeded by the top-level keys) and doubled as the walk's only termination guard — is replaced by identity-based cycle detection plus a node budget and depth cap (`[Truncated]` placeholder). Cyclic event data previously survived the walk and killed the batch in `process.send`; lazily-infinite getter chains could hang the worker.
- Array contents are now actually walked. The old array branch built `[element, index]` pairs (element as the parent), so e.g. the extra errors in `err.multiple` crossed the wire as `{}` with their messages lost. Function elements become `null`, which is what the JSON IPC channel already did to leaked functions.
- `BigInt` values are stringified (previously a guaranteed `process.send` TypeError losing the batch) and throwing getters are omitted rather than fatal.
- Constructor-form `error.cause` (own but non-enumerable, previously dropped by the `Object.assign` copy) is carried, so cause chains render in parallel mode as they do in serial mode.
- `SerializableWorkerResult#serialize` now substitutes a same-`eventName` stand-in when an event still cannot be serialized, so a single pathological event cannot erase a file's results.

Wire compatibility: all existing serializer unit tests pass unchanged (two test titles reworded, assertions untouched), and the existing parallel integration fixtures for circular references and getters pass as-is. An own enumerable `toJSON` is still not honored while inherited implementations (`Date`, `Buffer`) still are, matching the previous end-to-end JSON behavior. The deliberate, visible differences are strict improvements over data destruction: nested errors arrive with `message`/`stack` instead of `{}`, data cycles serialize as `[Circular]` instead of crashing the send, and pathological graphs truncate instead of hanging or overflowing.

Verification: 60 serializer unit tests (40 existing + 20 new, one per hazard), the parallel integration specs, and an end-to-end run of the #6053 repro (the original failure is reported and `crypto` stays intact in the reused worker) all pass on this branch.

Note: the first commit here is the same change as #6055 — on current `main` the serializer spec cannot load (#6054), so this branch needs it to run its own tests; happy to rebase once #6055 or #6015 lands. The four node-unit failures visible on this branch (`buffered-worker-pool`, `parallel-buffered-runner`, `reporters/parallel-buffered`, `mocha.spec`) predate it — current `main` cannot run the node-unit suite at all — and are addressed by #6015. A v11.x backport (the serializer there is identical apart from module wrappers) follows separately.

### DIFF
--- a/lib/nodejs/buffered-worker-pool.js
+++ b/lib/nodejs/buffered-worker-pool.js
@@ -15,7 +15,7 @@
 
 const serializeJavascript = require("serialize-javascript");
 const workerpool = require("workerpool");
-const { deserialize } = require("./serializer");
+const { deserialize } = require("./serializer.mjs");
 const debug = require("debug")("mocha:parallel:buffered-worker-pool");
 const { createInvalidArgumentTypeError } = require("../errors.mjs");
 
--- a/lib/nodejs/reporters/parallel-buffered.js
+++ b/lib/nodejs/reporters/parallel-buffered.js
@@ -33,7 +33,7 @@ const {
 const {
   SerializableEvent,
   SerializableWorkerResult,
-} = require("../serializer");
+} = require("../serializer.mjs");
 const debug = require("debug")("mocha:reporters:buffered");
 const { Base } = require("../../reporters/base.mjs");
 
--- a/lib/nodejs/serializer.mjs
+++ b/lib/nodejs/serializer.mjs
@@ -9,7 +9,7 @@
  * @typedef {import('../types.d.ts').SerializedWorkerResult} SerializedWorkerResult
  */
 
-import { type, breakCircularDeps } from "../utils.js";
+import { type, getMochaID, uniqueID } from "../utils.js";
 import { createInvalidArgumentTypeError } from "../errors.mjs";
 import d from "debug";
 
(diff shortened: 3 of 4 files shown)


mark-wiemer:
Marking as draft until the backing issue it out of triage :)

PR #6071, opened 2026-06-14
chore: convert lib/utils.js to ESM

### LINKED ISSUE #5964: 🛠️ Repo: Convert `lib/utils.js` to ESM
Sub-issue of #5400 (Switch from CJS to ESM internally).

`lib/utils.js` is a widely-consumed utility module still written in CJS. It exports many helper functions via `exports.X` and is a dependency across much of the core runtime.

### File to convert

- `lib/utils.js`

### Notes

Per [AGENTS.md](https://github.com/mochajs/mocha/blob/main/AGENTS.md) CJS-to-ESM conventions:
- Prefer named exports over `export default`
- Rename to `utils.mjs`
- Update all consumers to use `const { ... } = require("./utils.mjs")` (destructured named imports)
- There are many call sites across `lib/` — a search for `require.*utils` will surface them all

arun-357:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5964

## Overview

Part of #5400. Renames `lib/utils.js` → `lib/utils.mjs` and converts it to ESM named exports, updating every runtime call site and test in the same change. No behavior change.

- Each `exports.X = …` becomes an inline `export function`/`export const`; internal `exports.X` self-references are now direct calls. No default/aggregate export (per `AGENTS.md`).
- Call sites updated to match: CJS consumers `require("./utils.mjs")`, ESM named-import consumers `from "./utils.mjs"`, and namespace consumers switched from `import utils from` to `import * as utils from` (a default import of a named-export module is `undefined`).
- Tests: `sinon` cannot stub ES modules, so the `sinon.stub(utils, …)` calls were reworked — the unasserted `isString`/`noop` stubs were removed, and the `isBrowser` stubs were replaced with `process.browser` toggling (`isBrowser()` is `Boolean(process.browser)`, so the same code path and assertions run). `rewiremock` keys/import paths were repointed to `.mjs`.

Verified: `test-node` suites, `test-smoke`, `eslint`, `tsc`, and a `clean`+`build` browser bundle all pass; `lib/utils.js` was not listed in `package.json`'s `browser` map, so no config change was needed.


test/reporters/json.spec.js
```diff
@@ -217,12 +216,16 @@ describe("JSON reporter", function () {
     });
 
     it('should throw "unsupported error" in browser', function () {
-      sinon.stub(utils, "isBrowser").callsFake(() => true);
-      expect(
-        () => new JSONReporter(runner, options),
-        "to throw",
-        "file output not supported in browser",
-      );
+      process.browser = true;
```
mark-wiemer, line 219: This change doesn't look correct, can you explain what's going on here?

mark-wiemer: Ah, this seems fine, we're just convincing the runner that we're a browser in a more canonical way. Resolving :)

test/unit/mocha.spec.cjs
```diff
@@ -76,9 +75,6 @@ describe("Mocha", function () {
     };
     Suite = sinon.stub(Mocha, "Suite").returns(suite);
     Suite.constants = {};
-
-    sinon.stub(utils, "isString");
```
mark-wiemer, line 80: Can you explain how these tests aren't false negatives with this change?

mark-wiemer (dismissed):

test/unit/mocha.spec.cjs
arun-357: nothing ever asserted on these stubs, so they weren't testing anything

arun-357:
For code coverage, its not able to pick the natively loaded .mjs but tests still asserts the full utils as before.

arun-357:
Under the new type: module, plain .js is ESM — so do we want to keep it as .mjs or .js ? @mark-wiemer

JoshuaKGoldberg:
@arun-357 `.js`

arun-357:
@JoshuaKGoldberg No issue then, it is already a .js file in this PR.

arun-357:
@mark-wiemer I fixed the conflicts.

mark-wiemer:
@arun-357 haha, more conflicts. I'll probably fix those myself when I'm ready to review this one, but I've got lots of others on my backlog I'll be reviewing before this. Might be a few more weeks, this PR is low-priority. Thanks in advance for your patience :)

PR #5677, opened 2026-02-03
docs: switch util.inherits to class extends

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: closes #5025
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Tweaks the page to mention `extends` instead of `util.inherits`.

Other parts of the page will need updating too. Out of scope for this.

🤎

PR #6082, opened 2026-06-18
chore: add modern Netlify to v11.x branch

JoshuaKGoldberg:
Reverts mochajs/mocha#5630 and does lots more cleanup. Per https://github.com/mochajs/mocha/pull/5677#discussion_r3307788731, we actually _do_ want to publish an 11.x version of the site for users.


netlify.toml
```diff
@@ -0,0 +1,21 @@
...
+# “base” is directory to change to before starting build. if you set base:
+#    that is where we will look for package.json/.nvmrc/etc not repo root!
+
+[build]
+  command = "npm run docs"
+  publish = "docs/_site/"
+
+[build.environment]
+  DEBUG = "mocha:docs*"
+  NODE_VERSION = "20" # LTS until 2026-04-30
```
mark-wiemer: can we update Node version? Ideally just jumping to 24?

docs-next/netlify.toml
```diff
@@ -0,0 +1,11 @@
+[build]
+  command = "npm run docs"
+  publish = "dist/"
+
+[build.environment]
+  DEBUG = "mocha:docs*"
+  NODE_VERSION = "20"
```
mark-wiemer:
```suggestion
  NODE_VERSION = "24"
```

netlify.toml
mark-wiemer:
```suggestion
  NODE_VERSION = "24"
```

docs-next/netlify.toml
```diff
@@ -0,0 +1,11 @@
+[build]
```
mark-wiemer, line 1: Somehow this deploy broke the main mochajs.org this morning, I just fixed it by publishing the latest deploy from main (https://app.netlify.com/projects/mocha/deploys/6a33f9e008aab100085ecdbe). Let's make sure this deploys to e.g. `v11.mochajs.org` instead of the main site to reduce risk of downtime.

mark-wiemer (changes requested):

docs-next/netlify.toml
mark-wiemer: Hmm, I can't find the exact problem in the dashboard anymore. I may have misdiagnosed this, but it is the most likely cause of today's outage. I'll just be careful as I work to fix this, and I'll test thoroughly.

mark-wiemer: Ah, I think it's because we renamed `docs-next` to `docs`, and this `v11.x` branch doesn't have that change. Yuck.

mark-wiemer: This issue should be completely resolved, I've set mochajs.org to only deploy production branch (reduces risk of the "v11 deployed to apex" problem we had this morning). We'll deploy this branch to a new `v11.mochajs.org` project, completely separate. We can merge the two projects later if we get bored (or Josh thinks it's best--he knows Netlify better than me!)

mark-wiemer:
It was a smidgen flaky but it worked! Not worried about flakiness as redeploys will be rare, but we'll keep the preview enabled (configured in the private Netlify dashboard)

mark-wiemer (approved):
Wow great work Josh you are so cool

PR #6098, opened 2026-06-19
perf!: replace glob package with node:fs glob

### LINKED ISSUE #5962: ⚡️ Performance: Replace `glob` package with built-in `node:fs` `.glob`
### Performance Suggestion Checklist

- [x] I am using the latest version of Mocha.
- [x] I have read and understood the nuances around performance reports.

### Overview

Mocha right now uses [`glob`](http://npmjs.com/package/glob) internally for file pattern shell matching. Node.js >=22 has its own [`fs.glob`](https://nodejs.org/api/fs.html#fsglobpattern-options-callback) + https://nodejs.org/api/fs.html#fspromisesglobpattern-options that we could use to remove that dependency altogether. That would make Mocha have a smaller `node_modules/` install size and be more in line with what is now the first-party Node.js way to do globs.

Proposal: let's do this as part of Mocha 13? Note that it'd require dropping support for Node.js 20 as a breaking change.

### Validations

I set up a basic reproduction in https://github.com/JoshuaKGoldberg/repros/tree/mocha-dependencies-with-and-without-glob. Tl;dr:

* Without: **4.9M**
* With: **10M**

That's quite the savings!

### Additional Info

See #5785 for past discussion around replacing `glob`. The previous suggestion was to use a much smaller glob library, i.e. `tinyglobby`. We'd resolved not to take on #5785 because:

* `tinyglobby` has known missing support for some edge cases handled by `glob` (which partially explains why `glob` is larger!)
* Node.js 20 just exited LTS, so the big glob change is no longer blocked on _LTS_ major version support (just now-non-LTS)

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5962 

## Overview

Replace the "glob" dependency with Node's built-in node:fs glob plus "minimatch" (which Mocha already depends on, and which glob is itself built on). Removed glob from production dependencies, shrinking what users install.

IMPROTANT

**This is a breaking change (semver-major):** fs.globSync is only available on Node 22+ and so this drops support for Node.js 20. `engines` is now >=22.12.0, and the CI matrix and docs are updated to match.

### DIFF
--- a/lib/cli/watch-run.cjs
+++ b/lib/cli/watch-run.cjs
@@ -3,18 +3,14 @@
 const debug = require("debug")("mocha:cli:watch");
 const path = require("node:path");
 const chokidar = require("chokidar");
-const glob = require("glob");
 const isPathInside = require("is-path-inside");
-const { minimatch } = require("minimatch");
+const { minimatch, Minimatch } = require("minimatch");
 const Context = require("../context.js").Context;
 const collectFiles = require("./collect-files.cjs");
 const { logSymbols } = require("../utils.cjs");
 
 /**
  * @typedef {import('chokidar').FSWatcher} FSWatcher
- * @typedef {import('glob').Glob['patterns'][number]} Pattern
- * The `Pattern` class is not exported by the `glob` package.
- * Ref [link](../../node_modules/glob/dist/commonjs/pattern.d.ts).
  * @typedef {import('../mocha.cjs')} Mocha
  * @typedef {import('../types.d.ts').BeforeWatchRun} BeforeWatchRun
  * @typedef {import('../types.d.ts').FileCollectionOptions} FileCollectionOptions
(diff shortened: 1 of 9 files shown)


lib/cli/watch-run.cjs
```diff
@@ -177,66 +173,40 @@ function createPathFilter(globPaths, basePath) {
...
-      debug("all subpatterns of %j processed", pattern.globString());
+      firstGlobIndex = parts.length;
+      debug("all subpatterns of %j processed", globString);
     }
 
-    // match `cleanPath` (path without the glob part) and its subdirectories
-    const cleanPath = path.resolve(basePath, ...segments);
+    // match `cleanPath` (path without the glob part) and its subdirectories.
+    // After lots of errors, it turned out that globString is "/"-separated and keeps the absolute root (for example the
+    // drive C:/ on Windows), so derive the static prefix from it.
```
JoshuaKGoldberg: [Docs] Transient debugging woes like _"After lots of errors,"_ are transient. Future people/us don't need to know about that.

```suggestion
    // globString is "/"-separated and keeps the absolute root (e.g.
    // drive C:/ on Windows), so derive the static prefix from it.
```

JoshuaKGoldberg (dismissed):
LGTM, nice switch. I tinkered with this a bunch locally. It's a pity we'll have to wait so long to get it in 😢.

mark-wiemer:
Blocked on starting v13

PR #6119, opened 2026-06-28
test: retry touch in dir-add watch test to reduce flakiness

### LINKED ISSUE #6079: 🛠️ Repo: Some `--watch` tests fail sporadically

### Overview

"reruns test when file and directory paths under --watch-files are added" is the main offender, but others crop up too. Something is off with our recent fix, unfortunately. The named test above just failed on my local machine (Windows, Node 20.19.0, PR branch off main without relevant changes,  `npm run test-node:integration` without additional options) so we should be able to repro a bit better than before.

### Additional Info

I do think we're better now than we were before, no desire to revert the recent "fix".

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6079

## Overview

The test that watches "--watch-files" sometimes flakes because it only touches lib/dir/file.xyz once after a 1s delay. If the watcher has not fully picked up the new lib/dir yet that touch gets missed and the expected 4th run never happens so the test times out.

fix is simple, instead of touching the file once and hoping for the best we keep re-touching it until we actually see the 4 runs. Each touch updates the timestamp so it will eventually get picked up once the watcher is ready.

we stop as soon as run #4 happens so it cant overshoot and we only retry when things are idle so it won't interrupt active runs.

Also added runCount to the helper callback so the test can track progress properly.

One thing worth noting is that i unfortunately couldn't reproduce it locally even under a very heavy stress-test (fancy processor?), but this should cover the slow watcher setup window that was causing the flake.

### DIFF
--- a/test/integration/helpers.cjs
+++ b/test/integration/helpers.cjs
@@ -654,7 +654,7 @@ async function shutdownWatchChild(mochaProcess, closed) {
  * @param {RegExp} [opts.firstRunCrashPattern] - If set, expect the first run to fail to load, printing only an error stack (no stdout) with a message matching this pattern. No countable run output to wait for.
  * @param {'json'|'marker'} [opts.runDetector] - How completed runs are counted; `runMochaWatchJSONAsync` sets `json`
  * @param {number} [opts.budgetMs] - Shared deadline for everything this call waits on; must stay below the test's timeout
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
Yeah, this makes sense. I think the iterative _"check until ready"_ approach is going to fix a lot of flake. Great stuff. Thanks! ✨

Issue #6122, opened 2026-06-29
⚡️ Performance: Mocha 12 has larger install size despite fewer dependencies


JoshuaKGoldberg:
### Performance Suggestion Checklist

- [x] I am using the latest version of Mocha.
- [x] I have read and understood the nuances around performance reports.
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

Mocha 12 is a streamlining & modernization effort: we've removed some deps, transitioned the internals from CJS to ESM, etc. It's got fewer dependencies than 11 - which is good on its own!

However, the actual install size of `node_modules/` is larger: see https://packagephobia.com/result?p=mocha@12.0.0-beta-1 & https://packagephobia.com/result?p=mocha@12.0.0-rc.1. Which is an unfortunate regression. It comes from a few dependencies having gotten larger in their latest versions.

It'd be nice if we could also boast that Mocha 12 is _smaller_ than 11 in terms of install size. OTOH, I think it hurts Mocha 12's desirability to be _bigger_ than 11.

### Validations

```sh
for v in 11 12.0.0-rc.1; do
  rm -rf package* node_modules > /dev/null
  npm init --yes > /dev/null
  npm i mocha@$v > /dev/null
  echo "-- mocha@$v --"
  du -sk node_modules
done
```

```plaintext
-- mocha@11 --
11940   node_modules
-- mocha@12.0.0-rc.1 --
12996   node_modules
```

### Additional Info

I wish I'd gotten around to checking this before we got to the RC phase...

I suspect the easiest win for install size will be removing `yargs` altogether. This would generally be a good thing IMO, since the [Node.js-native `util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) has been available since Node.js 16.17.0/18.3.0.

JoshuaKGoldberg:
I split this out into three PRs, one for each of the dependencies:

1. #6124: no install size difference
[...]

PR #6124, opened 2026-06-29
feat!: replace yargs-parser in CLI option loading

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: starts on #6122 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))

## Overview

Part 1 of ~2-3 for #6122. This replaces the `yargs-parser` dependency and a lot of its wrapping logic with different local wrapping logic. The parser preserves config/env/package/default priority and `yargs`-y compatibility behavior.

Internal line change amount is approximately +165 (+306 for parse-args.js, +15/-157 for options.cjs).

Note that this doesn't change `node_modules/` size, since `yargs` itself has a dependency on `yargs-parser`.

### DIFF
--- a/lib/cli/options.cjs
+++ b/lib/cli/options.cjs
@@ -9,38 +9,25 @@
 
 const { readFileSync } = require("node:fs");
 const pc = require("picocolors");
-const yargsParser = require("yargs-parser");
-const {
-  types,
-  aliases,
-  isMochaFlag,
-  expectedTypeForFlag,
-} = require("./run-option-metadata.cjs");
+const { parseMochaArgs } = require("./parse-args.js");
 const { ONE_AND_DONE_ARGS } = require("./one-and-dones.js");
 const mocharc = require("../mocharc.json");
-const { list } = require("./run-helpers.cjs");
 const { loadConfig, findConfig } = require("./config.cjs");
 const { sync } = require("find-up");
 const debugModule = require("debug");
-const { isNodeFlag } = require("./node-flags.cjs");
-const {
-  createUnparsableFileError,
-  createInvalidArgumentTypeError,
-  createUnsupportedError,
-} = require("../errors.js");
-const { isNumeric } = require("../utils.cjs");
+const { createUnparsableFileError, isMochaError } = require("../errors.js");
 
 const debug = debugModule("mocha:cli:options");
 
 /**
- * The `yargs-parser` namespace
- * @external yargsParser
- * @see {@link https://npm.im/yargs-parser}
+ * The `parseArgs` function
+ * @external parseArgs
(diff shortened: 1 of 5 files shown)


lib/cli/options.cjs
```diff
@@ -9,38 +9,25 @@
...
 const debug = debugModule("mocha:cli:options");
 
 /**
- * The `yargs-parser` namespace
- * @external yargsParser
- * @see {@link https://npm.im/yargs-parser}
+ * The `parseArgs` function
+ * @external parseArgs
+ * @see {@link https://nodejs.org/api/util.html#utilparseargsconfig}
  */
```
mark-wiemer, line 26: I think this comment describes code that's now removed?

```suggestion
```

mark-wiemer: ah, used by `Arguments` which is still referenced, resolving

lib/cli/options.cjs
```diff
@@ -56,152 +43,23 @@ const YARGS_PARSER_CONFIG = {
...
-    const pair = arg.split("=");
-    let flag = pair[0];
-    if (isNodeFlag(flag, false)) {
-      flag = flag.replace(/^--?/, "");
-      return acc.concat([[flag, arg.includes("=") ? pair[1] : true]]);
+  try {
+    return parseMochaArgs(args, defaultValues, ...configObjects);
+  } catch (err) {
+    if (isMochaError(err)) {
+      throw err;
```
mark-wiemer, line 58: As small as it is, throwing the Mocha error may change the process's exit code, which could be considered a breaking change. Is there a way to ensure we still exit the process with exit code `1`? Or will we have to add this to the breaking changes list? If we need to add it to the breaking changes list, can it wait for Mocha 13?

mark-wiemer: We can probably take the rest of the changes without the new `isMochaError` check. Only the new check is causing the breaking change, from what I see so far.

mark-wiemer (commented):
Let's conclude on the "breaking change" concern, I could be swayed either way but I'm leaning towards avoiding new breaking changes right now, however small they may be.

lib/cli/options.cjs
JoshuaKGoldberg: I believe this is actually keeps compatibility _closer_ to the `yargs` version. Numeric positional validation already threw these structured Mocha errors from `createErrorForNumericPositionalArg()`. So only `yargs-parser`'s own `result.error` path actually exited with status 1.

I.e. if we have an _internal_ Mocha error, we _do_ want to throw.

mark-wiemer:
Will review in detail soon (TM) :)

mark-wiemer (approved):

PR #6130, opened 2026-07-02
fix: fail nested-suite tests affected by a failing beforeEach hook

### LINKED ISSUE #6132: 🐛 Bug: `--fail-hook-affected-tests` drops nested-suite tests when a `beforeEach` hook fails
### Bug Report Checklist

### Expected behavior

### Actual behavior

```js
// repro.mjs
  });
});
```

### Versions

- Node.js 24

### Additional Information

spokodev:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6132
- [ ] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22). I filed #6132 for this; it needs a maintainer to triage and label it before I can check this box.

## Overview

`--fail-hook-affected-tests` (added in v12) is documented to report tests as failed when they are affected by a hook failure, instead of silently skipping them. A failing `beforeEach` affects every test the hook would have run before, including tests in nested `describe` blocks. The option handles the `before` (beforeAll) case by recursing into sub-suites, but the `beforeEach` case only fails the remaining tests of the immediate suite and drops the tests in nested sub-suites. With the option on, those nested tests still vanish from the run, and the failure and total counts are under-reported in every reporter (spec, json, tap, xunit).

## Repro

A suite with a failing `beforeEach`, one direct test, and a nested `describe` with two tests. With `--fail-hook-affected-tests` the run reports 2 failures (the hook and the direct test) and drops the two nested tests. The same shape with `before` instead of `beforeEach` already reports all of them, which is the behavior this change brings to `beforeEach`.

## Fix

After failing the remaining tests of the current suite, recurse into `self.suite.suites` and fail their not-yet-run tests too, mirroring the existing `failAffectedTests` recursion used by the `before` path. The `!t.state` guard leaves already-run tests untouched.

Added a fixture and an integration test. The test fails before this change (2 failures) and passes after (4 failures, 1 pass); the hook, retries, pending and reporter suites stay green.

One point for review: like the existing `before`/beforeAll branch, this reports a statically `.skip()`ed test inside an affected sub-suite as failed rather than pending. That keeps the two branches consistent. If you would rather exempt pending tests, that is a separate change that should apply to both branches at once.

### DIFF
--- a/lib/runner.cjs
+++ b/lib/runner.cjs
@@ -851,6 +851,25 @@ Runner.prototype.runTests = function (suite, fn) {
           self.emit(constants.EVENT_TEST_END, t);
         }
       });
+      // Tests in descendant suites are skipped along with the remaining
+      // tests of this suite when a `before each` hook fails, so fail them too.
+      self.suite.suites.forEach(function failSubSuiteTests(s) {
+        s.tests.forEach(function (t) {
+          if (!t.state) {
+            var testError = createHookSkipError(
+              self._failedBeforeEachHook.title,
+              self._failedBeforeEachHook.error,
+            );
+
+            t.state = STATE_FAILED;
+            self.failures++;
+            self.emit(constants.EVENT_TEST_BEGIN, t);
+            self.emit(constants.EVENT_TEST_FAIL, t, testError);
+            self.emit(constants.EVENT_TEST_END, t);
+          }
+        });
+        s.suites.forEach(failSubSuiteTests);
+      });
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#-ai-generated-code

PR #6133, opened 2026-07-04
feat: add --clear-screen option for watch mode

### LINKED ISSUE #2312: Add a `--clearscreen` option to go with `--watch`

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #2312 

## Overview

Added a --clear-screen flag that clears the terminal before each run under --watch. I took the name from Josh's suggestion in the issue. It clears before every run,including the first, to match the "min" reporter and "tsc --watch".
It's a no-op without --watch so it stays out of the way of single runs.

bild

### DIFF
--- a/lib/cli/run.cjs
+++ b/lib/cli/run.cjs
@@ -277,7 +277,7 @@ exports.builder = (yargs) =>
       },
       "clear-screen": {
         description: "Clear the terminal before each run in watch mode",
-        group: GROUPS.FILES,
+        group: GROUPS.OUTPUT,
       },
       "watch-files": {
         description: "List of paths or globs to watch",
--- a/lib/cli/run.cjs
+++ b/lib/cli/run.cjs
@@ -275,6 +275,10 @@ exports.builder = (yargs) =>
         description: "Watch files in the current working directory for changes",
         group: GROUPS.FILES,
       },
+      "clear-screen": {
+        description: "Clear the terminal before each run in watch mode",
+        group: GROUPS.FILES,
+      },
       "watch-files": {
         description: "List of paths or globs to watch",
         group: GROUPS.FILES,
--- a/lib/cli/watch-run.cjs
+++ b/lib/cli/watch-run.cjs
@@ -46,7 +46,7 @@ const { logSymbols } = require("../utils.cjs");
  */
 exports.watchParallelRun = (
   mocha,
-  { watchFiles, watchIgnore },
+  { watchFiles, watchIgnore, clearScreen },
   fileCollectParams,
 ) => {
   debug("creating parallel watcher");
@@ -54,6 +54,7 @@ exports.watchParallelRun = (
   return createWatcher(mocha, {
     watchFiles,
     watchIgnore,
+    clearScreen,
     beforeRun({ mocha }) {
       // I don't know why we're cloning the root suite.
       const rootSuite = mocha.suite.clone();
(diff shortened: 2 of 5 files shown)


mostafaNazari702:
Forgot to include:
bild

bild

lib/cli/run.cjs
```diff
@@ -275,6 +275,10 @@ exports.builder = (yargs) =>
         description: "Watch files in the current working directory for changes",
         group: GROUPS.FILES,
       },
+      "clear-screen": {
+        description: "Clear the terminal before each run in watch mode",
+        group: GROUPS.FILES,
```
JoshuaKGoldberg: [Bug] Since this has to do with output, not how files are handled:

```suggestion
        group: GROUPS.OUTPUT,
```

lib/cli/watch-run.cjs
```diff
@@ -611,6 +623,15 @@ const eraseLine = () => {
   process.stdout.write("\u001b[2K");
 };
 
+/**
+ * Clears the terminal and moves the cursor to the top-left corner.
+ * @private
+ */
+const clearTerminal = () => {
+  // "\u001b[2J" clears the screen. "\u001b[0f" moves the cursor to the home position
+  process.stdout.write("\u001b[2J\u001b[0f");
```
JoshuaKGoldberg: [Bug] I poked around resources online and _believe_ `\u001b[H` is the more "standard" escape compared to `\u001b[0f`. The `H` form seems to be a general current standard for setting cursor position, standardized under ANSI X3.64 (ECMA-48). Whereas  the `0f` form is an older thing that isn't standardized.

```suggestion
  process.stdout.write("\u001b[2J\u001b[H");
```

JoshuaKGoldberg (dismissed):
Very clean! And a great PR description too 🙂. Just a couple of small touchups please.

lib/cli/run.cjs
mostafaNazari702: [Praise :) ]:
This is honestly a categorization judgment but your reasoning is sound!
Grouping it next to --watch was convenient (how i thought, initially) but conceptually wrong do i now realize. I will accept it.

lib/cli/watch-run.cjs
mostafaNazari702: Your recommendation is correct, i accept it, but "0f isn't standardized"(summarized) is not quite accurate honestly. Both forms are in ECMA-48. ESC[H is CUP (Cursor Position) and ESC[f is HVP (Horizontal and Vertical Position).

And i don't see that any is deprecated, they position the cursor identically.
The real distinction is that CUP (H) is classified as an editor function while HVP (f) is a format effector and that difference "can lead to different handling in certain terminal modes" So H is the safer, more conventional choice for a clear-and-home, even though f is standardized too. I also now see that H is also consistent with Mocha's own min reporter from an existing use of the it.

Read those:
https://jvns.ca/blog/2025/03/07/escape-code-standards/
https://man7.org/linux/man-pages/man4/console_codes.4.html
https://en.wikipedia.org/wiki/ANSI_escape_code

mostafaNazari702:
@JoshuaKGoldberg I had to use this in another project and actually your point stands. I did a new round of searching and i think you were right, though both standard the one you recommended seems to be the convention. Excuse me for being a smart aleck 😁.

mark-wiemer (approved):

PR #6144, opened 2026-07-16
feat: add --order option to run tests in random order

### LINKED ISSUE #902: 🚀 Feature: --order option for random test order

kirthi-b:
## PR Checklist

- [x] Addresses an existing open issue: fixes #902

## Overview

This adds a `--order` CLI option (and a matching `Mocha#order()` API) to run tests in a shuffled order, closing #902.

`--order random` shuffles sibling suites and sibling tests within their parent using a seeded Fisher-Yates shuffle (mulberry32 PRNG). The suite hierarchy is preserved, so hooks still run exactly as often as they would in a normal run. Each randomized run reports the seed it used on STDERR, so it doesn't corrupt machine-readable reporter output on STDOUT, and the seed can be passed back via `--order random:<seed>` to reproduce a specific order (useful for pinning down order-dependent test failures).

Accepted values: `default`, `random`, `random:<seed>` (seed is a 1-10 digit non-negative integer, normalized to an unsigned 32-bit int).

`--order random` is mutually exclusive with `--sort`.

This was originally built against the old yargs-based CLI. Since then main moved 35 commits and dropped yargs for a hand-rolled parser on top of node:util's parseArgs, so the CLI wiring here was rewritten against that: `order` is now registered as a string option in `lib/cli/run-option-metadata.cjs` (both the type bucket and the help description), and the format/mutual-exclusion validation now lives in `validateRunOptions` in `lib/cli/run.cjs`, matching how the `--parallel`/`--file` check works there. The actual shuffling logic in `lib/order.js` and its wiring into `lib/mocha.cjs`/`lib/runner.cjs` didn't need any changes, since none of that touches CLI parsing.

Closes #902

### DIFF
--- a/docs/src/content/docs/running/cli.mdx
+++ b/docs/src/content/docs/running/cli.mdx
@@ -41,6 +41,9 @@ Rules & Behavior
   -j, --jobs                        Number of concurrent jobs for --parallel;
                                     use 1 to run in serial
                                    [number] [default: (number of CPU cores - 1)]
+      --order                       Run order of tests; "random" or
+                                    "random:<seed>" shuffles execution order
+                                                   [string] [default: "default"]
   -p, --parallel                    Run tests in parallel              [boolean]
       --posix-exit-codes            Use POSIX and UNIX shell exit codes as
                                     Mocha's return value               [boolean]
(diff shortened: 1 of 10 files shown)


mark-wiemer:
Looks OK, pretty big and low-priority, but I'll add it to my backlog :)

mark-wiemer:
FYI @JoshuaKGoldberg the backing issue has 80+ likes!

PR #6160, opened 2026-07-21
chore: fix esm-utils tests when run without coverage

### LINKED ISSUE #6156: 🛠️ Repo: esm-utils test fails on newer Node 24 because a .ts fixture uses require()

### Overview

I was going through some of my older PRs and rebasing them onto latest main and while running the test suite i noticed a failure that had nothing to do with my changes. I checked and it fails on a clean main too so i figured i would report it separately.

One test in `test/node-unit/esm-utils.spec.cjs` fails:
```
1) esm-utils
     requireOrImport
       should surface Mocha errors rather than falling back to `import(...)`:

expected Promise (rejected) => ReferenceError('require is not defined in ES module scope,
you can use import instead') to be rejected with error satisfying
{ code: 'ERR_MOCHA_FORBIDDEN_EXCLUSIVITY',
  message: /`\.only` is not supported in parallel mode/ }
```

The test expects `ERR_MOCHA_FORBIDDEN_EXCLUSIVITY` but gets a `ReferenceError` instead.
The fixture it loads, `test/node-unit/fixtures/mock-mocha-forbidden-exclusivity-err.ts`, is a `.ts` file that uses CommonJS:
`const { createForbiddenExclusivityError } = require('../../../lib/errors.js');`

Since the root `package.json` is now `"type": "module"`, newer Node loads that `.ts` file as ESM so `require` is not defined inside it. `requireModule` in `lib/nodejs/esm-utils.cjs` then hits the `/\.(cts|mts|ts)$/` branch and rethrows the `ReferenceError` instead of the mocha error the test is looking for.

I copied the fixture to `.cts` and ran the same call and it gives the expected error:
`.cts -> Error | ERR_MOCHA_FORBIDDEN_EXCLUSIVITY | `.only` is not supported in parallel mode`

So renaming the fixture to `.cts` looks like it fixes it though Im not sure if you would rather keep it as `.ts` on purpose to cover TS handling.

### Additional Info

reproduce on main:
`node bin/mocha.js test/node-unit/esm-utils.spec.cjs`

Gives 8 passing and 1 failing. 
Versions:

mocha: 12.0.0-rc.1 (main: bf76999b)
node: v24.17.0
npm: 11.17.0

the reason the CI does not catch it because the test matrix in `.github/workflows/mocha.yml` pins exact versions (`20.19.4`, `22.18.0`, `24.6.0`) as in #5052 and the unit tests run there. 
The latest main run is green on 24.6.0. The smoke job does use a floating 24 but it only runs test/smoke/smoke.spec.cjs so it never touches this test, that means this will start failing once the Node 24 pin gets bumped.

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6156
 
## Overview
 
The reason CI passes is `nyc`, its .ts hook loads the fixture as CommonJS. Running the spec directly skips that hook and exposes the bug.


mostafaNazari702:
bild

bild

mark-wiemer:
@mostafaNazari702 I don't want `cts` if we can avoid it, can we add `type: commonjs` to a nested package.json that applies only to the fixtures in that small folder? There should be other examples in the codebase.

mark-wiemer (approved):

mark-wiemer:
>  the reason CI passes is `nyc`, its .ts hook loads the fixture as CommonJS. Running the spec directly skips that hook and exposes the bug.

Ah, this means we should tag it as `@bare` I believe

mark-wiemer:
Ref #5494

mark-wiemer:
Ah, we only added `@bare` in the v11.x branch: https://github.com/mochajs/mocha/blob/v11.x/test/integration/esm.spec.js#L85

I'll think about this a bit more.

mark-wiemer:
Well, that `bare` stuff isn't a blocker for this PR, which works in both cases now with minimal big changes. Merging, thanks!

PR #6163, opened 2026-07-22
feat!: replace yargs with Node.js util.parseArgs

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6122 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Replaces `yargs` command dispatching with [Node.js-native `util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig).

Stacked on:
- #6124
- #6125

This PR is actually a net reduction in lines of code. Ignoring test files, the PR total is +298/-397. Combined with those two stacked PRs, the total prod lines change comes out to +673/-577. Not bad!

### Behavior Changes

- `--help` output is now generated by Mocha instead of yargs, so exact formatting, grouping, and wrapping may differ while preserving the same user-facing information.
- `--version` / `-V` and `--help` / `-h` are handled before run preparation, so they print version/help output even when other options such as `--require` point to modules that cannot be loaded.

---

(Ref #6128 for paperwork issues)

PR #6128, opened 2026-07-01
feat!: replace yargs with Node.js util.parseArgs

JoshuaKGoldberg:
## PR Checklist

- [x] Addresses an existing open issue: part 3 of ~3 for #6122 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Replaces `yargs` command dispatching with [Node.js-native `util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig).

Stacked on:
- #6124
- #6125

This PR is actually a net reduction in lines of code. Ignoring test files, the PR total is +298/-397. Combined with those two stacked PRs, the total prod lines change comes out to +673/-577. Not bad!

### Behavior Changes

- `--help` output is now generated by Mocha instead of yargs, so exact formatting, grouping, and wrapping may differ while preserving the same user-facing information.
- `--version` / `-V` and `--help` / `-h` are handled before run preparation, so they print version/help output even when other options such as `--require` point to modules that cannot be loaded.

PR #6164, opened 2026-07-22
refactor!: replace yargs with Node.js util.parseArgs

### LINKED ISSUE #6122: ⚡️ Performance: Mocha 12 has larger install size despite fewer dependencies

mark-wiemer:
### PR Checklist

- [x] Addresses an existing open issue: fixes #6122 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))

## Overview

Replaces `yargs` command dispatching with [Node.js-native `util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig). Marked as a breaking change out of an abundance of caution.

### Behavior Changes

- `--help` output is now generated by Mocha instead of yargs, so exact formatting, grouping, and wrapping may differ while preserving the same user-facing information.
- `--version` / `-V` and `--help` / `-h` are handled before run preparation, so they print version/help output even when other options such as `--require` point to modules that cannot be loaded.

---

(Ref #6128 and #6163 for paperwork issues)


lib/cli/run-option-metadata.cjs
```diff
@@ -95,6 +95,67 @@ const aliases = {
...
+  config: "Path to config file",
+  delay: "Delay initial execution of root suite",
+  diff: "Show diff on failure",
+  "dry-run": "Report tests without executing them",
+  exit: "Force Mocha to quit after tests complete",
+  extension: "File extension(s) to load",
+  "fail-hook-affected-tests":
+    "Report tests as failed when affected by hook failures (before/beforeEach)",
+  "fail-zero": "Fail test run if no test(s) encountered",
+  "pass-on-failing-test-suite": "Not fail test run if tests were failed",
```
mark-wiemer, line 114: New PR: fix this yucky description

mark-wiemer: I'll open the new PR once we merge this PR, not gonna bother with a backing issue for this one, this comment thread is sufficient for me

lib/cli/run-option-metadata.cjs
```diff
@@ -95,6 +95,67 @@ const aliases = {
...
+  fgrep: "Only run tests containing this string",
+  file: "Specify file(s) to be loaded prior to root suite execution",
+  "forbid-only": "Fail if exclusive test(s) encountered",
+  "forbid-pending": "Fail if pending test(s) encountered",
+  "full-trace": "Display full stack traces",
+  global: "List of allowed global variables",
```
mark-wiemer, line 120: Description of `global`: Another yucky help message, this should say "List allowed global variables", imperative

mark-wiemer: (Again, will be done in its own PR)

lib/cli/run.cjs
```diff
@@ -21,359 +20,125 @@ const {
...
+      );
+    }
+
+    acc[pair[0]] = pair.length === 2 ? pair[1] : true;
+    return acc;
+  }, {});
+
+const normalizeRunOptions = (argv) => {
+  Object.keys(argv).forEach((name) => {
+    const reporterOptionMatch = name.match(/^reporter-options?\.(.+)$/);
```
mark-wiemer, line 52: Did we support `reporter-options` (plural style) before? If not, we should not add support for it now without consideration

```suggestion
    const reporterOptionMatch = name.match(/^reporter-option\.(.+)$/);
```

mark-wiemer: Ah, we had support in lib/cli/run-option-metadata.cjs:89 before, resolving this one. We also have a unit test for it, nice

lib/cli/run.cjs
```diff
@@ -21,359 +20,125 @@ const {
...
 
-        if (argv.reporter === "markdown") {
-          throw createUnsupportedError(
-            "--reporter=markdown is mutually exclusive with --parallel",
-          );
-        }
+  if (argv.parallel) {
+    if (argv.file) {
+      throw createUnsupportedError(
+        "--parallel runs test files in a non-deterministic order, and is mutually exclusive with --file",
```
mark-wiemer, line 107: New PR: why mention "non-deterministic" in any `parallel` error message?

lib/cli/run.cjs
```diff
@@ -21,359 +20,125 @@ const {
...
+        argv[normalizedName] = argv[name];
       }
+    }
+  });
+
+  if (Array.isArray(argv["reporter-option"])) {
+    argv["reporter-option"] = coerceReporterOption(argv["reporter-option"]);
+  }
+
+  return argv;
```
mark-wiemer: We should just return void here since we modify the argument directly

mark-wiemer: We should (implicitly) return void here since we modify the argument directly. Saves time wondering "oh, that returns something, does that mean it doesn't modify the argument?"

```suggestion
  }
```

lib/cli/run.cjs
```diff
@@ -21,359 +20,125 @@ const {
...
-          );
-        }
+const validateRunOptions = (argv) => {
+  // "one-and-dones"; help and version are handled by the top-level CLI.
+  const oneAndDoneOptions = Object.keys(ONE_AND_DONES).filter((opt) =>
+    Boolean(argv[opt]),
+  );
+  if (oneAndDoneOptions.length > 1) {
+    throw createUnsupportedError(
+      `Arguments ${oneAndDoneOptions.join(" and ")} are mutually exclusive`,
```
mark-wiemer, line 79: Hm, just want to make sure we have comprehensive tests here

lib/cli/run.cjs
```diff
@@ -21,359 +20,125 @@ const {
...
-      if (argv.parallel) {
-        // yargs.conflicts() can't deal with `--file foo.js --no-parallel`, either
-        if (argv.file) {
-          throw createUnsupportedError(
-            "--parallel runs test files in a non-deterministic order, and is mutually exclusive with --file",
-          );
-        }
+const validateRunOptions = (argv) => {
+  // "one-and-dones"; help and version are handled by the top-level CLI.
+  const oneAndDoneOptions = Object.keys(ONE_AND_DONES).filter((opt) =>
```
mark-wiemer, line 74: Should we use ONE_AND_DONE_ARGS here?


```suggestion
  const oneAndDoneOptions = Object.keys(ONE_AND_DONE_ARGS).filter((opt) =>
```

mark-wiemer: I know we have the special `h` and `v` handling above, but maybe we can remove that instead?

lib/cli/run.cjs
mark-wiemer: added tests

lib/cli/run.cjs
mark-wiemer: Ah, ONE_AND_DONES have functions, ONE_AND_DONE_ARGS don't. ONE_AND_DONE_ARGS is used for short-circuiting even earlier, I'm fine to keep things as they are, don't want to rock the boat too much, esp in an already-large PR :)

lib/cli/run.cjs
mark-wiemer: We did support it :)

lib/cli/run.cjs
mark-wiemer: https://github.com/mochajs/mocha/issues/6165

lib/cli/run-option-metadata.cjs
mark-wiemer: https://github.com/mochajs/mocha/issues/6165

lib/cli/run-option-metadata.cjs
mark-wiemer: https://github.com/mochajs/mocha/issues/6165

mark-wiemer:
(Converting back to draft while I handle some local AI-generated feedback)

mark-wiemer:
(It was all hallucinations)

JoshuaKGoldberg (approved):
Swell! We've both worked on this and it looks very good to me. I'm still a little nervous about edge cases in the CLI, but:

* This area has a lot of test coverage, including some added points (IMO the coverage drop is inaccurate and misleading)
* It's a major version that's going into RC anyway, little oddities are expected (and always fixable after the fact)

Merging now. 🚀

PR #6168, opened 2026-07-23
fix: add package.json exports to avoid DEP0151 warning

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6147 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Adds an `exports` map that points the package root to the existing `./index.js` entry point and keeps a wildcard subpath export so existing deep imports (e.g. `mocha/lib/reporters/*`) and browser bundle files remain reachable.

Node 22+ emits a `[DEP0151] DeprecationWarning` when an ES module package is resolved without an explicit `"main"` or `"exports"` field. Mocha 12 is now `"type": "module"`, so importing the package root triggered the warning.

`exports` is the modern alternative to `main`, supporting advanced features like pattern-matching and conditional exports. `main` is still supported. Because our project did not have either `exports` or `main` before this, we were implicitly exporting everything, and we must continue to do so to avoid a breaking change. Ref https://nodejs.org/api/packages.html#package-entry-points.

In the future, we can restrict what we export for future major versions of Mocha.

PR #6157, opened 2026-07-20
fix: add "main" field to package.json

EhabY:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6147
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Since the package became `"type": "module"`, Node's default `index.js` lookup emits [DEP0151] when importing `mocha` from an ES module, because `package.json` defines neither `"main"` nor `"exports"`.

This adds `"main": "./index.js"`. Chose `"main"` over an `"exports"` map since `"exports"` would take precedence over the `"browser"` field and break deep imports such as `mocha/mocha.js` and `mocha/mocha.css`. Also removes the knip `index.js!` entry that is now inferred from `"main"`, and adds a regression test.

[DEP0151]: https://nodejs.org/api/deprecations.html#DEP0151

PR #6193, opened 2026-07-29
fix: use main instead of exports for the package entry point

### LINKED ISSUE #6192: 🐛 Bug: deep requires without a file extension broke in 12.0.0-rc.5
### Bug Report Checklist

### Expected

`require('mocha/lib/reporters/base')` keeps resolving in 12.0.0-rc.5 like it did in rc.4 and in 11.x.

### Actual

It throws `Cannot find module`. The `exports` map that was added in #6168 turns off CommonJS extension search for subpaths so `./lib/reporters/base` maps to a literal file that doesnt exist. Same goes for `mocha/lib/stats-collector` and `mocha/lib/errors` and `mocha/lib/suite` and `mocha/index` and `mocha/bin/mocha` and any other deep path written without a extension.

Create a new folder and: 
1. `npm init -y`
2. `npm i mocha@12.0.0-rc.4`
3. `node -e "require.resolve('mocha/lib/reporters/base')"`   # Should be just fine

A screenshot showing no returned Cannot find module error

then:
`npm i mocha@12.0.0-rc.5`
`node -e "require.resolve('mocha/lib/reporters/base')"   # Cannot find module`

A screenshot showing a returned Cannot find module error

rc.4 and rc.5 have the same files, exports is the only package.json difference, so that isolates it.

### Versions

mocha 12.0.0-rc.5 (broken), 12.0.0-rc.4 and 11.7.6 (fine)
node 24.17.0. Not platform specific.

### Additional Info

_No response_

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6192 

## Overview

using `"main": "./index.js"` instead of the exports map so deep requires without a file extension keep working.

Mark's #6168 picked exports to silence DEP0151, which it does but exports also turns off CommonJS extension search for subpaths. `"./*": "./*"` looks like it keeps everything reachable and it does for anything with an extension but `require('mocha/lib/reporters/base')` now maps to a literal file (no extension) that isn't there. 

The warning text is "No main or exports field" so main is enough on its own and doesn't bring the encapsulation along with it.

I went with this instead of papering over it with more subpath patterns because there are no pattern that supports both spellings. I tried `"./*": ["./*", "./*.js"]` and Node doesn't fall back on missing files so it breaks every deep path instead of just some.

This leaves the package as un-encapsulated as it was in version 11 which I think is right for a minimally-breaking v12. A proper exports map with a designed sub-path surface would fit v13 better. Existing entry point test still covers DEP0151.


mark-wiemer (dismissed):
Thanks for this!

mark-wiemer:
Builds on work from https://github.com/mochajs/mocha/pull/6157, @ehaby we will attribute you using your email in https://github.com/coder/vscode-coder commits :)

mark-wiemer:
Quick notes:
- No screenshots of passing is necessary here, nor are instructions for how to repro test failures. We always assume that previous versions would've failed the new test and we know that newer versions pass the test due to our CI check
- Moved extended justification to the backing issue for clarity :)

mark-wiemer:
Blocked on concurrence with @JoshuaKGoldberg

praann07 (commented):
﻿This is a notable compatibility change, so flagging a couple of things:

- The previous `exports` map defined both `.` (entry) and `./*` (deep-import subpaths). Removing `./*` means ESM consumers that deep-import internal paths (e.g. `import "mocha/lib/reporter"`) fall back to the filesystem layout rather than the subpath-exports contract. That can work but is less explicit and may surprise consumers.
- The new test only asserts deep subpath resolution via the filesystem; it does not cover the subpath-exports contract itself. Worth a changelog note and confirming no ESM consumers rely on the `./*` subpath map.

The linked-Mocha refactor into `invokeNodeWithLinkedMocha` is clean and de-duplicates the temp-dir / symlink logic nicely.

JoshuaKGoldberg (dismissed):
@mostafaNazari702 sorry about the back-and-forth, let's do `main` after all. This explicit `exports` map is -as you said- quite a lot.

mark-wiemer (approved):

mark-wiemer:
Oops, I figured integration tests would complete before auto-merge. Ref [run 6345 on main](https://github.com/mochajs/mocha/actions/runs/31322503801) for results

PR #6200, opened 2026-08-02
fix: report test error when afterEach fails before retry

### LINKED ISSUE #5007: 🐛 Bug: Error from afterEach hides the original error from the test case when retries > 0
### Prerequisites

### Description

Error from afterEach hides the original error from the test case when retries is used (> 0).

### Steps to Reproduce

Here is the minimum test case to reproduce the issue

```
describe('test after each error', () => {

  afterEach(async () => {
    throw new Error('Error from after each');
  });

  it('test one', async () => {
    throw new Error('Error from test');
  });

});
```

Then run the following to see the difference
```
npx mocha --retries 0 index.spec.js
npx mocha --retries 1 index.spec.js
```

The result should tell me the error raised from the test case

```
npx mocha --retries 0 index.spec.js


  test after each error
    1) test one
    2) "after each" hook for "test one"


  0 passing (3ms)
  2 failing

  1) test after each error
       test one:
     Error: Error from test
      at Context.<anonymous> (index.spec.js:8:11)
      at process.processImmediate (node:internal/timers:476:21)

  2) test after each error
       "after each" hook for "test one":
     Error: Error from after each
      at Context.<anonymous> (index.spec.js:4:11)
      at process.processImmediate (node:internal/timers:476:21)
```

The output did not show the error raised in the test case but only show the error from the clean up `afterEach`

```
npx mocha --retries 1 index.spec.js


  test after each error
    1) "after each" hook for "test one"


  0 passing (2ms)
  1 failing

  1) test after each error
       "after each" hook for "test one":
     Error: Error from after each
      at Context.<anonymous> (index.spec.js:4:11)
      at process.processImmediate (node:internal/timers:476:21)
```

**Reproduces how often:** [What percentage of the time does it reproduce?] 100%

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`: 10.2.0 for both
- The output of `node --version`: v18.16.1
- Your operating system
  - name and version: OSX 13.5.1
- Your shell (e.g., bash, zsh, PowerShell, cmd): iTerm
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): No

### Additional Information

Nothing

deepakganesh78:
When a retryable test failed and its afterEach hook also failed, Mocha only reported the hook failure and hid the original test error.

The retry path ran afterEach before marking the test failed; this now reports the test error if afterEach aborts the retry.

Fixes #5007

Validation: npm run lint; npm run test-node:unit (1222 passing, 6 pending); npm run test-smoke (1 passing).

### DIFF
--- a/lib/runner.cjs
+++ b/lib/runner.cjs
@@ -979,7 +979,16 @@ Runner.prototype.runTests = function (suite, fn) {
 
             // Early return + hook trigger so that it doesn't
             // increment the count wrong
-            return self.hookUp(HOOK_TYPE_AFTER_EACH, next);
+            return self.hookUp(
+              HOOK_TYPE_AFTER_EACH,
+              function (hookErr, errSuite) {
+                if (hookErr) {
+                  self.fail(test, err);
+                  self.emit(constants.EVENT_TEST_END, test);
+                }
+                next(hookErr, errSuite);
+              },
+            );
           } else {
             self.fail(test, err);
           }
--- a/test/unit/runner.spec.cjs
+++ b/test/unit/runner.spec.cjs
@@ -528,6 +528,43 @@ describe("Runner", function () {
         });
       });
 
+      it("should report test failure when afterEach fails before a retry", function (done) {
+        var testErr = new Error("test error");
+        var hookErr = new Error("hook error");
+        var failed = [];
+
+        suite.retries(1);
+        suite.afterEach(function () {
+          throw hookErr;
+        });
+        suite.addTest(
+          new Test("im a test about bears", function () {
+            throw testErr;
+          }),
+        );
+
+        runner.on(EVENT_TEST_FAIL, function (runnable, err) {
+          failed.push({ runnable, err });
+        });
+
+        runner.run(function (failures) {
+          expect(failures, "to be", 2);
+          expect(
+            failed.some(function (failure) {
+              return failure.runnable instanceof Test && failure.err === testErr;
+            }),
+            "to be true",
+          );
+          expect(
+            failed.some(function (failure) {
+              return failure.runnable instanceof Hook && failure.err === hookErr;
+            }),
+            "to be true",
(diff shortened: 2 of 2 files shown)


praann07 (commented):
﻿Reviewed at the code level only (did not run the suite locally).

When `afterEach` throws before a retry, this now calls `self.fail(test, err)` and manually emits `EVENT_TEST_END`. Two things worth checking before merge:

1. Does `Runner.prototype.fail` already emit `EVENT_TEST_END` downstream? If so, the explicit `self.emit(constants.EVENT_TEST_END, test)` could double-emit and surface the test twice in reporters (and the event stream). Worth confirming against `lib/runner.cjs` `fail()`.
2. `self.fail(test, err)` fails the test with the test's own error (`err`), not the hook's error (`hookErr`). That reads as intentional (the test should be reported as failed), but a one-line comment clarifying "fail the test, not the hook" would prevent future confusion.

The added regression test is a good guard.

praann07 (approved):
When an afterEach hook fails on retry, the hook error was previously swallowed and the test wasn't marked failed. This wraps the hookUp callback to fail() + emit EVENT_TEST_END on hookErr before forwarding to next(), preserving retry behavior. The unit test asserts both the test and hook errors are reported (failures===2). Correct and minimal.

praann07 (approved):
Regression: an afterEach hook failure on retry was swallowed and the failed test wasn't reported. This fails the test and emits EVENT_TEST_END on hookErr before forwarding to next(); the unit test asserts both the test and hook errors fire (failures===2). Correct, minimal fix.

praann07 (approved):
Regression: an afterEach failure on retry was swallowed and the failed test was not reported. Now fail() + emit EVENT_TEST_END on hookErr before forwarding; unit test asserts failures equals 2. Correct and minimal.

Issue #5528, opened 2025-11-08
📝 Docs: Review docs for correctness


mark-wiemer:
### Documentation Request Checklist

- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I want to provide a PR to resolve this

### Overview

We've been doing good work on #5248 but we haven't been reviewing the docs for correctness. As a result, our very first code snippet is syntactically incorrect (#5526). Along with #5446 we should review all files for correctness, esp code snippets.

- List all doc files here for tracking
- Find a system to review code snippets for correctness
- Update docs to match Mocha v11-12, Node v18/20/22/24/26

### Additional Info

This will be a lot of work, but it's important.

ashvin2005:
heyy @mark-wiemer 
I would like to work on the issue....could you pls assign it to me?

TG199:
Hi @ashvin2005 checkout Josh comment from #5497

JoshuaKGoldberg:
Adding the `core-team` label to indicate this is something Mark and I will tackle.

mark-wiemer:
Others are welcome to chip in here, but the issue won't be closed until we maintainers actually manually read everything and decide for ourselves that pages are up-to-date and correct :) if you find issues or mistakes, you're welcome to mention them here, open PRs, or file separate issues for detailed tracking. For any significant fixes, we ask that folks provide examples to demonstrate where the docs were incorrect before

PR #5563, opened 2025-12-03
fix: allow importing ESM interface and reporters

Lightning00Blade:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5562
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

A very simple (and rough) PR to show case where the issue is. It only unwraps modules `default` if present.

PR #6215, opened 2026-08-09
docs: fix minor correctness issues

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: part of #5528

## Overview

the docs still send people to the wiki in four places and every one of those wiki pages is now just a stub saying "This page has moved to mochajs.org". So you click "for more information" and gets no information.

also rewrote the top of the compilers explainer.
It came over from the wiki unchanged and incorrectly says `--compilers` is a *soft* deprecation that can be silenced with `--no-deprecation` contradicting the CLI page saying it was removed in v6.0.0.
Since #5403, it doesn’t error at all, `mocha --compilers coffee:coffeescript/register "test/*.js"` treats `coffee:coffeescript/register` as a file spec and runs the tests.

Sending users from the CLI page to that stale text seemed worse(?) than the "dead" link so i fixed it in the same PR.

The migration examples are still correct so i left them alone (Aside from fixing one js fence that had a shell command).


mostafaNazari702:
Found more stale or inaccurate code and documentations after pushing "[docs: link to mochajs.org instead of the wiki](https://github.com/mochajs/mocha/pull/6215/commits/44300408256557b36eadde247713cef9f445f672)", i will have to update PR body later.

mostafaNazari702:
I think i will comment instead of editing the PR description:

1. Updated the docs since #5563 added ESM support in v12. `.mjs` interfaces and reporters work in `"type": "module"` packages. The remaining limitation is top-level `await ` which fails because mocha loads them by require `(ERR_REQUIRE_ASYNC_MODULE)` so you guys might want to plan or do something about it since i did not look too deep in it.

2. Prettier quitely skips unparseable Javascript in MDX so `format:check` stayed green. `browsers.mdx` diff is mostly just Prettier formatting after fixing the stray `...`.

3. Fixed sharing the same meta/OG description when only the first actually matched it.

mostafaNazari702:
Also, i am probably making unnecessary changes but is single quotes the convention here for Mocha?

bild

mark-wiemer:
- Format check will enforce single quotes vs double quotes
- Please split the link fixes into their own PR, those are easy to approve and merge
- Not a fan of `isCorrectEnvironment()`, I don't like magic sample functions that aren't actually defined or explained, I do prefer the comment form in this case
- Will review the rest of the PR once changes are made :)

mostafaNazari702:
> * Format check will enforce single quotes vs double quotes
> 
>     * Please split the link fixes into their own PR, those are easy to approve and merge
> 
>     * Not a fan of `isCorrectEnvironment()`, I don't like magic sample functions that aren't actually defined or explained, I do prefer the comment form in this case
> 
>     * Will review the rest of the PR once changes are made :)

Reverted some changes and created a new pull request for other changes.

docs/src/content/docs/explainers/nodejs-native-esm-support.mdx
```diff
@@ -27,7 +27,7 @@ More information can be found in the [Node.js documentation](https://nodejs.org/
 ### Current Limitations
 
 - [Watch mode](/running/cli#--watch--w) does not support ES Module test files
-- [Custom reporters](/reporters/third-party) and [custom interfaces](/interfaces/third-party) can only be CommonJS files
+- [Custom reporters](/reporters/third-party) and [custom interfaces](/interfaces/third-party) can be ES modules but Mocha loads them with `require` so they cannot use top-level `await`
```
mark-wiemer, line 30: Can you prove this for me quickly just out of curiosity?

docs/src/content/docs/explainers/compilers-deprecation.mdx
```diff
@@ -1,17 +1,9 @@
...
-If you're here, you probably hit the deprecation notice. Sorry about that!
-
-## Will it break?
-
-This is a _soft_ deprecation, which means you get nagged about it, but it won't break (yet).
-
-## Make it go away
-
-To suppress this warning, execute `mocha` with the `--no-deprecation` flag (though you won't get notice of any _other_ deprecations you may encounter either).
+`--compilers` was removed in v6.0.0, and as of v12.0.0 Mocha doesn't error on it either: the value you pass gets picked up as a test file spec instead.
```
mark-wiemer:
```suggestion
`--compilers` was deprecated in v6.0.0. As of v12.0.0, Mocha treats it as normal text, not as an option.
```

mark-wiemer (changes requested):

docs/src/content/docs/explainers/nodejs-native-esm-support.mdx
mostafaNazari702: sure, and all of this is on published mocha@12.0.0-rc.6 and nothing from this branch.

package.json is just `{ "type": "module" }` then:

```js
// my-reporter.js
import mocha from "mocha";
export default class MyReporter extends mocha.reporters.Base {
  constructor(runner) {
    super(runner);
    runner.on("pass", (t) => console.log("esm reporter saw:", t.fullTitle()));
  }
}
```

```js
// my-ui.js
import mocha from "mocha";
export default function myUi(suite) {
  mocha.interfaces.bdd(suite);
  suite.on("pre-require", (context) => (context.spec = context.it));
}
```

```js
// test/a.js
describe("esm", function () {
  spec("works", function () {});
});
```

both ESM and both load:

```
$ npx mocha --reporter ./my-reporter.js --ui ./my-ui.js
esm reporter saw: esm works
(exit 0)
```

same reporter on mocha 11.8.0, which is what #5563 fixed:

```
Exception during run: TypeError: this._reporter is not a constructor
    at Mocha.run (.../node_modules/mocha/lib/mocha.js:987:18)
```

now one line of top level await added to that same reporterr:

```
$ npx mocha --reporter ./tla-reporter.js --ui ./my-ui.js
✖ ERROR: TypeError: Could not load reporter "./tla-reporter.js":
 Error [ERR_REQUIRE_ASYNC_MODULE]: require() cannot be used on an ESM graph with top-level await. Use import() instead.
(exit 1)
```

same for --ui. comes from the `require` in `validateLegacyPlugin` (`lib/cli/run-helpers.cjs`) and in `Mocha.prototype.reporter` (`lib/mocha.cjs`).

docs/src/content/docs/explainers/compilers-deprecation.mdx
mostafaNazari702: took your second sentence as is, but kept "removed" for v6, hope thats fine.

the deprecation itself was v4.0.0 (#3038). v6.0.0 is where it stopped working. CHANGELOG for 6.0.0 say "`--compilers` is no longer supported; attempting to use will cause Mocha to fail with a link to more information" and cli.mdx says "removed in v6.0.0". since this PR is partly about the docs not contradicting each other i didnt want to write deprecated in v6 right next to that.

so it reads:

`--compilers` was deprecated in v4.0.0 and removed in v6.0.0. As of v12.0.0, Mocha treats it as normal text, not as an option.

the v12 part on 12.0.0-rc.6:

```
$ npx mocha --compilers coffee:coffeescript/register "test/*.js"
Warning: Cannot find any files matching pattern "coffee:coffeescript/register"
  esm
    ✔ works
  1 passing (1ms)
(exit 0)
```

vs 11.8.0:

```
✖ ERROR: --compilers is DEPRECATED and no longer supported.
          See https://github.com/mochajs/mocha/wiki/compilers-deprecation for migration information.
(exit 1)
```

funny enough that v11 error points at the wiki page thats now a dead stub, which is the other PR.

I can take your version as is if you would rather want to keep it short.

mark-wiemer (approved):

Issue #6221, opened 2026-08-12
🐛 Bug (v11): spec file starting with a number rejected


grahamrhay:
### Bug Report Checklist

- [x] This is NOT a [security, `npm audit`, or GitHub Advisory issue](https://mochajs.org/explainers/security-vulnerability-reports).
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

passing tests

### Actual

after dependabot bumped mocha from `~10.8.1` to `~11.7.6`, tests started failing with this error:

```
npx mocha 1***.js
/app/node_modules/mocha/lib/cli/options.js:139
    throw createUnsupportedError(
    ^

Error: Option 1***.js is unsupported by the mocha cli
    at createUnsupportedError (/app/node_modules/mocha/lib/errors.js:126:13)
    at createErrorForNumericPositionalArg (/app/node_modules/mocha/lib/cli/options.js:139:11)
    at parse (/app/node_modules/mocha/lib/cli/options.js:190:5)
    at loadOptions (/app/node_modules/mocha/lib/cli/options.js:291:14)
    at Object.<anonymous> (/app/node_modules/mocha/bin/mocha.js:29:14)
    at Module._compile (node:internal/modules/cjs/loader:1781:14)
    at Object..js (node:internal/modules/cjs/loader:1913:10)
    at Module.load (node:internal/modules/cjs/loader:1505:32)
    at Function._load (node:internal/modules/cjs/loader:1309:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:254:19) {
  code: 'ERR_MOCHA_UNSUPPORTED'
}
```

other spec files (not starting with a number) work fine. is this a bug, or were we using something unsupported that just happened to work until now?

[...]

PR #6222, opened 2026-08-12
fix: only reject positional args that are entirely numeric

### LINKED ISSUE #6283: 🐛 Bug (v12): spec file stating with a number rejected
See parent issue, this also affects v12+

lazerg:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6283

## Overview

`isNumeric` checks its input with `parseFloat`, which stops at the first character that cannot be part of a number. A spec file whose name starts with a digit, like `npx mocha 1-spec.js`, therefore counts as a numeric positional and the CLI bails out with `Option NaN is unsupported by the mocha cli`.

Comparing against `Number` instead makes the check look at the whole string, so bare numbers such as `mocha 123` still error while file names that merely begin with a digit run as before.

Note that #6221 is still in triage rather than labelled accepting prs, so feel free to close this if you want to discuss it first.

### DIFF
--- a/lib/utils.cjs
+++ b/lib/utils.cjs
@@ -732,7 +732,7 @@ exports.breakCircularDeps = (inputObj) => {
  * Checks if provided input can be parsed as a JavaScript Number.
  */
 exports.isNumeric = (input) => {
-  return !isNaN(parseFloat(input));
+  return String(input).trim() !== "" && !isNaN(Number(input));
 };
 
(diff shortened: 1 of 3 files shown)


mark-wiemer:
Looks good, let's wait for tests to pass and also backport this to v11

mark-wiemer (approved):

PR #6225, opened 2026-08-14
fix: fail nested tests when a beforeEach hook fails

### LINKED ISSUE #6132: 🐛 Bug: `--fail-hook-affected-tests` drops nested-suite tests when a `beforeEach` hook fails

arronKler:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6132

## Overview

`--fail-hook-affected-tests` already reports every remaining test under a failed `before` hook, including nested `describe`s. A failed `beforeEach` only marked leftover tests in the current suite, so nested tests disappeared from the run.

Route `beforeEach` through the same `failAffectedTests` walk as `before`. Added an integration fixture for the nested case.

Fixes #6132

### DIFF
--- a/lib/runner.cjs
+++ b/lib/runner.cjs
@@ -651,27 +651,12 @@ Runner.prototype.hook = function (name, fn) {
         }
       } else if (err) {
         self.fail(hook, err);
-        // If failHookAffectedTests is enabled, mark affected tests as failed
-        if (self._opts.failHookAffectedTests) {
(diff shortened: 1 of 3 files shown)


arronKler:
recheck

mark-wiemer:

PR #6234, opened 2026-08-14
fix: flush V8 coverage before parallel workers are killed

### LINKED ISSUE #6219: 🐛 Bug: `--parallel` workers are force-killed before V8 coverage flushes, silently dropping coverage
### Bug Report Checklist

### Expected

When running with `--parallel` under a V8-coverage tool (`c8`, or `nyc` in V8 mode), each worker's collected coverage should be included in the report. At a minimum, the grace period a worker is given to exit before Mocha force-kills it should be configurable, so a worker whose exit/flush legitimately takes longer than 1 second is not killed mid-flush.

### Actual

Whole workers' coverage is **silently discarded**, so the parallel report is dramatically lower than the same suite run serially. The run still exits `0`; only the percentages look wrong, so it is easy to misread as a flaky/environment problem.

Root cause: Mocha creates the `workerpool` pool without setting `workerTerminateTimeout`, so workerpool uses its default of **1000 ms**. At end of run each worker is asked to terminate and then hard-killed (`SIGTERM`) once that timeout elapses. Node writes the `NODE_V8_COVERAGE` profile *as the process exits*; a worker still flushing when the 1 s grace period lapses is killed first, and that worker's coverage file is never written. There is currently no Mocha CLI/`.mocharc` option to change this timeout.

It is load-dependent: it appears on many-core machines (more workers flushing at once) and with large dependency graphs (bigger profiles, ~10–16 MB per worker here). Lowering `--jobs` hides it, which disguises the real cause.

Relevant code:

```js
// lib/nodejs/parallel-buffered-runner.js — pool created with only maxWorkers
pool = BufferedWorkerPool.create({maxWorkers: options.jobs});

// lib/nodejs/buffered-worker-pool.js — WORKER_POOL_DEFAULT_OPTS never sets it

// workerpool/src/Pool.js — hence the 1000ms fallback
this.workerTerminateTimeout = options.workerTerminateTimeout || 1000;
// workerpool/src/WorkerHandler.js — setTimeout(() => worker.kill(), workerTerminateTimeout)
```

The effect needs enough concurrent flush work to exceed 1 s, so it reproduces most reliably in a real project on a many-core machine:

```bash
# many-core machine; large-ish dependency graph
c8 --reporter=json-summary --temp-directory .cov-tmp \
  mocha --parallel --jobs 9 -- '**/*.test.ts'
```

Then inspect `.cov-tmp/*.json`: the per-worker V8 dumps that should contain each spec are missing, and they never appear after the run (so it is not the reporter reading too early — the dumps were never written).

Measured on a 10-core Linux devcontainer, same suite, only the grace period / `--jobs` varied (compare covered counts, not just %, since `c8 all:true` shifts denominators between runs):

| Run | Worker coverage dumps written | lines | functions | branches |
| --- | --- | --- | --- | --- |
| `--parallel --jobs 9` (default 1000 ms) | 0 of 9 | 35.69% | 22.11% | 60.50% |
| `--parallel --jobs 2` (default 1000 ms) | 2 of 2 | 73.57% | 72.61% | 65.84% |
| `--parallel --jobs 9`, grace period 60000 ms | 9 of 9 | 73.59% | 72.61% | 69.96% |

Raising only the grace period (nothing else) restores all 9 workers' coverage.

I could not make a small, dependency-free repro fail: with tiny profiles the flush completes well within 1 s. It needs enough workers/profile size/disk latency to push the flush past the 1 s window. On constrained CI (e.g. 4 vCPU) it often "works" simply because fewer workers flush simultaneously. If a synthetic MCVE is required I can build one that pads each worker's exit flush, but I wanted to report the real-world trigger first.

### Versions

- `mocha`: 11.7.5 (the pool is still created without `workerTerminateTimeout` in 11.8.0 and the 12.0.0 RCs)
- `node_modules/.bin/mocha --version`: 11.7.5
- `node --version`: v20.20.2
- `workerpool`: 9.x (default `workerTerminateTimeout` 1000 ms; unchanged in v10)
- `c8`: 10.1.x
- OS: Linux x64, 10 logical cores
- Transpiler: `ts-node` (CommonJS)

#### Proposed fix

Expose workerpool's `workerTerminateTimeout` as a first-class Mocha option so it can be raised without patching, keeping the current default unchanged:

- Add a `--worker-terminate-timeout <ms>` number option (Rules & Behavior group), documented as the grace period a `--parallel` worker is given to exit on its own before being force-killed; `defaultDescription: '1000'` (no behavior change unless set).
- Thread `options.workerTerminateTimeout` into `BufferedWorkerPool.create({maxWorkers, workerTerminateTimeout})`, forward it into the workerpool options, and add it to the parallel runner's `DENY_OPTIONS` (it is a pool-level option like `jobs`, not something to serialize to worker Mocha instances). When unset, leave workerpool's own default in place rather than pinning one in Mocha.

This is the smallest change that preserves current defaults. A stronger fix (awaiting a clean coverage flush before force-killing) would also be welcome but is more invasive; a configurable grace period unblocks users today and is a reasonable knob regardless. I'm running exactly this locally via `patch-package` and it resolves the loss (row 3 above); happy to open the PR.

#### Workarounds for others hitting this

- Lower `--jobs` until workers flush within 1 s (fragile; regresses as the suite grows).
- Collect coverage from a serial run (`--no-parallel`) — slower.
- Patch Mocha/workerpool to raise `workerTerminateTimeout`.

#### Related

- bcoe/c8#374 — "Code coverage breaks in parallel mode": same symptom; the workaround there was lowering `--jobs`, and the root cause was never identified as this timeout.

arronKler:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6219

## Overview

`--parallel` workers were force-killed after workerpool's 1s terminate timeout, which cuts off Node's `NODE_V8_COVERAGE` write. Coverage from those workers is silently dropped.

Workers now flush V8 coverage from workerpool's `onTerminate` hook. When `NODE_V8_COVERAGE` is set, the terminate grace period is raised to 60s. `--worker-terminate-timeout` can override that.

Fixes #6219

### DIFF
--- a/docs/src/content/docs/features/parallel-mode.mdx
+++ b/docs/src/content/docs/features/parallel-mode.mdx
@@ -161,6 +161,12 @@ Mocha does **not** provide an option to spawn a fresh process for each test file
 - Invoking `mocha` separately for each test file and merging results externally.
 - Using a test runner that provides process-level isolation per test file.
 
+## Code Coverage
+
+V8 coverage (`c8`, or `nyc` in V8 mode) is written when each worker process exits. Mocha asks workers to exit cleanly and flushes coverage over that shutdown path. If a worker is still exiting when the terminate grace period elapses, workerpool force-kills it and that worker's coverage dump is dropped.
+
+When `NODE_V8_COVERAGE` is set, Mocha automatically allows workers up to 60 seconds to exit. Override this with [`--worker-terminate-timeout`](/running/cli#--worker-terminate-timeout-ms) if a run still drops coverage.
+
 ## Parallel Mode Worker IDs
 
 :::note[New in v9.2.0]
--- a/docs/src/content/docs/running/cli.mdx
+++ b/docs/src/content/docs/running/cli.mdx
@@ -42,6 +42,9 @@ Rules & Behavior
                                     use 1 to run in serial
                                    [number] [default: (number of CPU cores - 1)]
   -p, --parallel                    Run tests in parallel              [boolean]
+      --worker-terminate-timeout    Milliseconds a --parallel worker may take
+                                    to exit before it is force-killed
+                                                        [number] [default: 1000]
       --posix-exit-codes            Use POSIX and UNIX shell exit codes as
                                     Mocha's return value               [boolean]
       --retries                     Retry failed tests this many times  [number]
@@ -250,6 +253,14 @@ The default value is the _number of CPU cores_ less 1.
 :::tip
 Use `--jobs 0` or `--jobs 1` to temporarily disable `--parallel`.
 :::
+
+Has no effect unless used with [`--parallel`](#--parallel--p).
+
+### `--worker-terminate-timeout <ms>`
+
+When Mocha shuts down a [`--parallel`](#--parallel--p) worker, it first asks the process to exit so Node can flush `NODE_V8_COVERAGE` data. If the worker is still alive after this timeout, it is force-killed and that flush is lost.
+
+The default is `1000` (workerpool's default). When `NODE_V8_COVERAGE` is set and this option is omitted, Mocha raises the grace period to `60000` so coverage tools such as `c8` can finish writing large profiles.
 
 Has no effect unless used with [`--parallel`](#--parallel--p).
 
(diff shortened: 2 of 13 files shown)


arronKler:
recheck

mark-wiemer:

PR #6236, opened 2026-08-14
fix: count delayed tests before applying --fail-zero

### LINKED ISSUE #4950: --fail-zero always fails the test run when --delay is also used

arronKler:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4950

## Overview

`--fail-zero` with `--delay` always exited 1, even when delayed tests ran and passed.

`Runner` sets `this.total` in the constructor, before delayed suites register tests via `run()`. `--fail-zero` then treated that stale `0` as "no tests".

Recount `this.total` with `grepTotal()` after delayed registration, so grep/invert still apply.

Fixes #4950

### DIFF
--- a/lib/runner.cjs
+++ b/lib/runner.cjs
@@ -1223,6 +1223,8 @@ Runner.prototype.run = function (fn, opts = {}) {
     }
     this.state = constants.STATE_RUNNING;
     if (this._opts.delay) {
+      // Delayed tests are registered after the Runner is constructed.
+      this.total = this.grepTotal(this.suite);
       this.emit(constants.EVENT_DELAY_END);
       debug('run(): "delay" ended');
     }
--- a/test/integration/options/failZero.spec.cjs
+++ b/test/integration/options/failZero.spec.cjs
@@ -1,5 +1,6 @@
 "use strict";
 
+var path = require("node:path").posix;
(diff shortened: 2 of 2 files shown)


arronKler:
recheck

mark-wiemer:

PR #6237, opened 2026-08-14
fix: apply --fail-zero to the whole parallel run

### LINKED ISSUE #4969: --fail-zero + --parallel + --grep always returns an exit code even though tests pass
### Prerequisites

- [X] Ensured that there is no discrepancy between the locally and globally installed versions of Mocha. You can find them with: `node_modules/.bin/mocha --version`(Local) and `mocha --version`(Global). We recommend that you _not_ install Mocha globally.

### Description

If you have multiple test files and run tests with `--fail-zero --parallel --grep "test name"`, the test suite will always fail with an exit code even though all tests pass

### Steps to Reproduce

Download the attached project
```
npm ci
npm run-test
```

`npm run test` runs the 4 different ways the options can be combined.

**Expected behavior:** I expect `--fail-zero --parallel --grep hello2` to return an exit code of 0 (i.e. `npm run test-failing` should pass).

**Actual behavior:** An exit code of 7 is returned

### Versions

mocha 10.2.0
node v19.7.0
Arch Linux x64
shell: bash

### Additional Information

If you run `node_modules/.bin/mocha --parallel --grep hello --fail-zero` then it passes so it must be something to do with grep specifically.

arronKler:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4969

## Overview

`--fail-zero` should fail only when the whole run encounters no tests. In parallel mode each worker applied it to its own file, so `--grep` files with zero matches incremented the exit code even when other files passed.

Workers no longer apply `--fail-zero`. The parent runner applies it to the aggregate test count.

Fixes #4969

### DIFF
--- a/lib/nodejs/parallel-buffered-runner.cjs
+++ b/lib/nodejs/parallel-buffered-runner.cjs
@@ -348,6 +348,11 @@ class ParallelBufferedRunner extends Runner {
           return;
         }
 
+        // --fail-zero applies to the aggregate run, not each worker/file
+        if (this._opts.failZero && !(this.stats && this.stats.tests)) {
+          this.failures = 1;
+        }
+
         this.emit(EVENT_RUN_END);
         debug("run(): completing with failure count %d", this.failures);
         callback(this.failures);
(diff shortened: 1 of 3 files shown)


arronKler:
recheck

mark-wiemer:

PR #6263, opened 2026-09-02
fix: watch existing dirs so first run starts armed

### LINKED ISSUE #6297: 🛠️ Repo: `--watch` dir-add test flakes waiting for the first mkdir rerun

### Overview

`--watch` integration test `reruns test when file and directory paths under --watch-files are added` is flaking again on CI.

It failed on `main` in [this job](https://github.com/mochajs/mocha/actions/runs/34612521767/job/103307503487) (`test-node:integration` on Node 26.8.2 / Ubuntu). The same job passed on other Node versions and on Node 26.8.2 / Windows. It also passed on a later rerun of the same matrix cell, so this is a flake, not a Node 26 regression.

The failure is: 

```
  1) --watch
       when enabled
         reruns test when file and directory paths under --watch-files are added:
     Error: runMochaWatchAsync: timed out waiting for 2 run(s) to complete; 1 run(s) completed so far
```

The child did finish the first run (`waiting for changes...`), then never reran. The hang is the **first** `mkdir` of `lib`, which did not exist when `chokidar.watch()` started. The later `file.xyz` step already retries; the two `mkdir` steps do not. A second `mkdir` of an existing dir is a no-op, so a retry has to recreate (`rm` + `mkdir`) to re-emit `addDir` on the parent watch.

This is the same test as #6079 / #6037. #6119 closed #6079 by retrying only the file touch. That does not cover a missed first `addDir`.

Suggested fix: apply the existing idle retry loop to the directory creates, recreating the dir until `runCount()` advances.

### Additional Info

Related: #5361, #5714, #6037, #6079, #6119.

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6297 

## Overview

Fix --watch-files test timing out on Windows CI by making sure the directory is watched before the first run finishes.

test was watching lib before it existed so chokidar only started watching it after "ready". On busy CI runners the second run could happen in that tiny gap and get missed leading to a 50s timeout.

Now we give chokidar the closest directory that already exists so it starts watching in time. I left retry from #6119 alone because it cover the separate nested lib/dir case.

### DIFF
--- a/lib/cli/watch-run.cjs
+++ b/lib/cli/watch-run.cjs
@@ -1,6 +1,7 @@
 "use strict";
 
 const debug = require("debug")("mocha:cli:watch");
+const fs = require("node:fs");
 const path = require("node:path");
 const chokidar = require("chokidar");
 const glob = require("glob");
@@ -9,6 +10,8 @@ const { minimatch } = require("minimatch");
 const Context = require("../context.js").Context;
 const collectFiles = require("./collect-files.cjs");
 const { logSymbols } = require("../utils.cjs");
+
+const setTimeout = global.setTimeout;
 
 /**
  * @typedef {import('chokidar').FSWatcher} FSWatcher
@@ -391,6 +394,21 @@ function createPathMatcher(allowed, ignored, basePath) {
   return matcher;
 }
 
+async function whenWatching(watcher, dirPaths, attempts = 400) {
+  if (dirPaths.length === 0) {
+    return;
+  }
+  const expected = dirPaths.map((dirPath) => path.dirname(dirPath));
+  for (let attempt = 0; attempt < attempts; attempt++) {
+    const watched = Object.keys(watcher.getWatched());
+    if (expected.every((dirPath) => watched.includes(dirPath))) {
+      return;
+    }
+    await new Promise((resolve) => setTimeout(resolve, 5));
+  }
+  debug("gave up waiting for watchers on %o", expected);
+}
+
 /**
  * Bootstraps a Chokidar watcher. Handles keyboard input & signals
  * @param {Mocha} mocha - Mocha instance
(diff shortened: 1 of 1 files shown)


mark-wiemer (dismissed):
We specifically don't want to watch parent directories, that would add noise that users are actively avoiding. We'll have to find a different approach.

mostafaNazari702:
Now i will keep it in here for better tracking.
Back to Mark now.

Repro: [CI run](https://github.com/mostafaNazari702/mocha/actions/runs/34721616727), ubuntu and windows, against main, this PR as you reviewed it, and a new version i would like to propose instead.

Noise and watched folders come out identical on all three. Nothing unrelated reruns, and the folder that actually ends up watched is the same in every case including 8 levels of missing folders. So the "great great grandparent"-case is already what main does today.

but you were right my patch did add something. Not watches but rather work because i was handing chokidar the whole parent folder instead of the one name it wants so it scanned everything in there on startup. With 3000 files next to the watched folder:

| | ubuntu | windows |
| --- | --- | --- |
| main | 124-132 ms | 227-249 ms |
| this PR as reviewed | 420-427 ms | 797-825 ms |

roughly 3x on Windows for nothing so i dipped it.

The bug itself is that chokidar fires ready before that fallback watch is really up. The watch tests assume the first run's output showing up means the watcher is live and on main those two land about a millisecond apart. That is the flakiness in #6079.

just pushed the new version. It leaves what gets watched completely alone and only holds the first run until the watcher is actually up. Startup is back to main's numbers and the ordering is guaranteed instead of lucky.

mark-wiemer:
(Mostafa's recent comment is a continuation of a private convo we had on Discord)

mark-wiemer (dismissed):
Thank you, I hate flaky CI!

mark-wiemer (approved):

PR #6260, opened 2026-09-01
chore: convert lib/runner.js to ESM

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5965 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Rename lib/runner.cjs to lib/runner.js and converts it to ESM with a named Runner export.

@mark-wiemer Also: It says to rename to `runner.mjs` but you and Josh agreed on js for ESM and cjs for CJS in https://github.com/mochajs/mocha/issues/5400#issuecomment-4730579332 so i went with `runner.js`. Also on to do #5964 first, i dont think thats needed. Runner just does import utils from "./utils.cjs", same as runnable.js and reporters/base.js already do so it works any way and not have to wait on #6071.

PR #6292, opened 2026-09-10
fix: fail nested tests on beforeEach failure

### LINKED ISSUE #6132: 🐛 Bug: `--fail-hook-affected-tests` drops nested-suite tests when a `beforeEach` hook fails

mahesh-desu:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6132 (or is a [trivial change](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md#trivial-changes))

## Overview

Closes #6132.

A failing `beforeEach` with `--fail-hook-affected-tests` only failed the remaining tests of the immediate suite; tests in nested `describe` blocks were dropped entirely (not failed, not pending), under-reporting failure/total counts in every reporter.

This change mirrors the existing `before`/`beforeAll` path: in `runTests` -> `hookErr`, after failing the remaining direct tests, recursively mark not-yet-run tests in `suite.suites` as failed (same TEST_BEGIN/FAIL/END emission, `!state` guard so already-run tests are never double-counted).

Repro from the issue (`repro.mjs`):
- Before fix: `node ./bin/mocha.js --fail-hook-affected-tests repro.mjs --reporter json` -> 2 failing (hook + `direct test`), nested tests dropped, `stats.tests: 1`.
- After fix: 4 failing (hook + `direct test` + `nested test 1` + `nested test 2`), `stats.tests: 3` — identical to the `before` variant. Also verified consistent counts with `spec` (4 failing) and `tap` (# fail 4) reporters.

Tests (all with `--timeout 10000`; the 1s default is too short for spawned child processes on Windows — pre-existing env quirk, fails on clean main too):
- `test/integration/hook-err.spec.cjs`: 23 passing (includes new `error in beforeEach hook with nested suites` regression using `hooks/before-each-hook-error-with-fail-affected-nested.fixture.js`)
- `test/reporters/*.spec.{cjs,js}`: 158 passing
- `test/unit/**/*.spec.cjs` + `test/node-unit/**`: 1234 passing
- `eslint` on changed source/spec: 0 errors

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -835,9 +835,11 @@ Runner.prototype.runTests = function (suite, fn) {
       !after &&
       self._failedBeforeEachHook
     ) {
-      // Fail all remaining tests in the suite
-      var remainingTests = tests.slice();
-      remainingTests.forEach(function (t) {
+      // Fail all remaining tests in the suite and its nested suites.
+      // Sub-suites are skipped after a beforeEach failure (see
+      // `runSuite`), so without this their tests would be dropped
+      // instead of reported as failed (mochajs/mocha#6132).
+      function failTest(t) {
         if (!t.state) {
           var testError = createHookSkipError(
             self._failedBeforeEachHook.title,
@@ -850,7 +852,13 @@ Runner.prototype.runTests = function (suite, fn) {
           self.emit(constants.EVENT_TEST_FAIL, t, testError);
           self.emit(constants.EVENT_TEST_END, t);
         }
-      });
+      }
+      function failTestsInSuite(s) {
+        s.tests.forEach(failTest);
+        s.suites.forEach(failTestsInSuite);
+      }
+      tests.slice().forEach(failTest);
+      suite.suites.forEach(failTestsInSuite);
       // Clear the stored hook info
       delete self._failedBeforeEachHook;
     }
--- a/test/integration/fixtures/hooks/before-each-hook-error-with-fail-affected-nested.fixture.js
+++ b/test/integration/fixtures/hooks/before-each-hook-error-with-fail-affected-nested.fixture.js
@@ -1,0 +1,18 @@
+'use strict';
+
+describe('outer', function () {
+  beforeEach(function () {
+    throw new Error('error in `beforeEach` hook');
+  });
+  it('direct test', function () {
+    // This should be reported as failed due to beforeEach hook failure
+  });
+  describe('nested', function () {
+    it('nested test 1', function () {
+      // This should be reported as failed due to beforeEach hook failure
+    });
+    it('nested test 2', function () {
+      // This should be reported as failed due to beforeEach hook failure
+    });
+  });
+});
--- a/test/integration/hook-err.spec.cjs
+++ b/test/integration/hook-err.spec.cjs
@@ -332,6 +332,31 @@ describe("hook error handling", function () {
       });
     });
 
+    describe("error in `beforeEach` hook with nested suites", function () {
+      it("should fail tests in nested suites", function (done) {
+        runMochaJSON(
+          "hooks/before-each-hook-error-with-fail-affected-nested.fixture.js",
+          ["--fail-hook-affected-tests"],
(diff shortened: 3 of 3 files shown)


mahesh-desu:
Thanks for flagging — PR body updated with the completed checklist (fixes #6132, labelled \status: accepting prs\, CONTRIBUTING steps followed).

lib/runner.js
```diff
@@ -835,9 +835,11 @@ Runner.prototype.runTests = function (suite, fn) {
       !after &&
       self._failedBeforeEachHook
     ) {
-      // Fail all remaining tests in the suite
-      var remainingTests = tests.slice();
-      remainingTests.forEach(function (t) {
+      // Fail all remaining tests in the suite and its nested suites.
+      // Sub-suites are skipped after a beforeEach failure (see
+      // `runSuite`), so without this their tests would be dropped
+      // instead of reported as failed (mochajs/mocha#6132).
```
mark-wiemer:
```suggestion
```

lib/runner.js
mark-wiemer, line 485: Can we update this `Runner.prototype.failAffectedTests` property instead? Looks like we accidentally duplicated some code earlier.

@copilot give it a shot

mark-wiemer (changes requested):
Otherwise looks good, thanks :)

lib/runner.js
mahesh-desu: Done in 082d4cb5 - hookErr now calls Runner#failAffectedTests instead of the duplicated inline logic.

On the Copilot check: failAffectedTests already handled the nested case - it recursively fails all not-yet-run tests in the suite and its child suites with the same \!state\ guard and TEST_BEGIN/TEST_FAIL/TEST_END emissions - so the property itself needed no changes, just the call site.

One nuance: the old inline code iterated the remaining \	ests\ array, which skipped grep-filtered tests of the current suite, while failAffectedTests iterates all of \suite.tests\ - grep-filtered tests are now failed too, consistent with the nested-suite behavior and the before/beforeAll path. Integration (hook-err: 23) and runner unit (79) tests pass.

mahesh-desu:
Re: the failing `codecov/patch` check (0.00% of diff hit) — this looks like a pre-existing CI pipeline issue, not something fixable by tests:

- Since #6260 converted `lib/runner.js` to ESM, nyc's require-hook cannot instrument it. `require("./runner.js")` goes through the require(esm)/ESM machinery, so the CJS `_compile`-based instrumentation never takes effect. Verified locally: under `nyc --no-clean --reporter=json node bin/mocha.js test/unit/runner.spec.cjs`, `lib/runner.js` reports 0/514 statements hit even though the Runner definitely executes; direct `istanbul-lib-instrument.instrumentSync` fails on the file with `'import' and 'export' may appear only with 'sourceType: module'` (nyc's default `esModules: false`), and nyc silently falls back to uninstrumented code.
- The Codecov report confirms it: on `main`, `lib/runner.js` shows **0.00% coverage** (553 lines, 0 hits) per the [Codecov API](https://codecov.io/api/v2/github/mochajs/repos/mocha/commits/880269ff709414c6015bb0aee45213ed965ca51d).
- #6260 itself was merged with both `codecov/project` and `codecov/patch` failing, so any PR adding lines to `lib/runner.js` will fail `codecov/patch` until the pipeline supports ESM.

I added unit tests covering the `failHookAffectedTests` path in `runTests` (92cf539f) so the lines are exercised in-process and will count once instrumentation works. Options to unblock:

1. Override/ignore the failed `codecov/patch` check for this PR, or
2. Fix the pipeline, e.g. switch to V8 coverage (`c8`/`NODE_V8_COVERAGE`), which instruments ESM natively, and/or set `esModules: true` plus an ESM loader hook in nyc.

mark-wiemer (approved):
Thanks mr robot

PR #6302, opened 2026-09-13
fix: mark cascading teardown-hook failures as secondary

### LINKED ISSUE #5925: 🐛 Bug: parent-suite teardown error masks original failure from nested suite

naba-runn:
- [x] Addresses an existing open issue: fixes #5925

When a test/hook fails in a nested suite and a parent teardown hook (afterEach/after all) also throws, both show up as separate failures and the teardown one hides the real cause.

This PR flags the teardown error as secondary and shows a (secondary) label in the reporter, so the original failure stays primary. Failure count stays same, and independent teardown failures are not affected.

Also added tests and fixtures.

Another clarification, I went with keeping the count same, just labeled it secondary. Another way would be to merge it into the original failure so count becomes "1 failing" but there is a risk of breaking count parsing. Let me know if you would prefer the second option! Also, if there's any other change you want in this PR, please do let me know, I am more than happy to contribute to this repo:) Thank you!


lib/reporters/base.js
```diff
@@ -532,6 +532,11 @@ Base.list = function (failures) {
       testTitle += str;
     });
 
+    // Label teardown cascades so the original failure stays the primary one (#5925)
```
mark-wiemer:
```suggestion
```

test/integration/hook-err.spec.cjs
```diff
@@ -130,6 +130,63 @@ describe("hook error handling", function () {
...
+  describe("teardown error masks original failure from nested suite", function () {
+    it("marks a cascading afterEach failure as secondary", function (done) {
+      runMochaJSON(
+        "hooks/after-each-hook-secondary-error.fixture.js",
+        [],
+        (err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res, "to have failed test count", 2);
```
mark-wiemer, line 143: I know I'm missing something: why would this report 2 failed tests when only one test is being run?

@copilot feel free to chime in

mark-wiemer:
@naba-runn please remove the verbose comments from the PR and esp remove references to particular PRs, we can find those later in the blame if needed :)

naba-runn:
@mark-wiemer hey so there is only one test case, but two hooks that throw errors(beforeEach and afterEach). The actual test never runs because the beforeEach aborts it before it reaches there. This is the same thing I mentioned in the description earlier, right now I am keeping the count as 2 and just labelling the teardown one as secondary. If you would prefer to show only 1 failing, I will merge it into the original error(beforeEach error), I will be happy to make the change:) Thank you!

naba-runn:
Also removed the verbose comments and added the else { delete err.secondary; } as suggested by copilot. I will add the regression case as well.


> * The cascade marker is only set for errors delivered as `cbHookRun(err)`. Errors emitted through a hook's `error` listener (for example the multiple-done/late-async path, which calls `self.fail(hook, err)` directly at line 628) bypass this block, so a parent teardown failure after an earlier nested failure is still reported without `secondary`. Apply the same check before that direct-fail path, ideally through a shared helper.
> 
> ```
>           self._failSecondary = true;
>         }
> ```
Also I looked into this and I did miss this edge case, but as the path is asynchronous, the runner current test may have moved ahead by the time error listener fires. So we can't use the same failure count check here. Also this is mostly an edge case path, the normal throw path is already handled so it would be better to handle this in a follow up. But if you want this here, I am up for it. Thank you!

naba-runn:
Fixed the codecov error and added the regression test that copilot suggested:)

lib/runner.js
```diff
@@ -553,6 +559,12 @@ Runner.prototype.fail = function (test, err, force) {
     err = thrown2Error(err);
   }
 
+  if (secondary) {
+    err.secondary = true;
+  } else {
+    delete err.secondary;
+  }
```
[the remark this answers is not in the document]
mark-wiemer: I'm not 100% sure this is legit, I will try to repro later. Overall this is a complex PR that's not a huge priority, I will get to it eventually :)

PR #6310, opened 2026-09-14
fix: give CLI reporter-option precedence over config file

### LINKED ISSUE #5532: Bug: command line reporter-options parameter doesn't override .mocharc.yml

Minouccc:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5532

## Overview

`--reporter-option` given on the CLI lost to the same key declared in an RC file or `package.json`: the config value won instead of the CLI value.

Root cause: in `mergeConfigObjects` (`lib/cli/parse-args.js`), array options from higher-priority sources were prepended ahead of lower-priority ones, producing `[CLI, env, rc, pkg]` order. Downstream `coerceReporterOption` (`lib/cli/run.cjs`) reduces with last-writer-wins semantics, so the lowest-priority source ended up winning.

Fix: `reporter-option` values now accumulate in priority-LOW->HIGH order, so the existing last-writer-wins reduction yields CLI > env > rc > pkg. Within one source, repeated flags keep last-wins behavior. Every other array option keeps the existing prepend order, which its consumers (e.g. spec ordering) rely upon.

Tests added:
- `test/node-unit/cli/parse-args.spec.js`: pins the merged LOW->HIGH order across all four sources (fails without the fix).
- `test/node-unit/cli/options.spec.js`: `loadOptions()` merges rc options before CLI options.
- `test/integration/options/reporter-option.spec.cjs`: end-to-end run with a `.mocharc.yml` plus a clashing CLI flag; asserts the custom reporter receives the CLI value alongside the non-clashing config value.

Validation run locally: new tests fail without the fix and pass with it; `test/node-unit/cli/parse-args.spec.js` + `options.spec.js` 92 passing; `reporter-option.spec.cjs` 9 passing; `test-node:unit` 1238 passing; `test-smoke` passing; `format:check`, `lint`, `tsc` clean. The `test/integration/options` directory shows failures only in the timing-sensitive `--parallel`/`--timeout` suites, identical population on unmodified main in this environment.


Minouccc:
Requesting reconsideration: this addresses #5532, which carries `status: accepting prs`. The fix is 8 lines in `mergeConfigObjects` plus unit/integration tests. I verified the new tests fail without the fix and pass with it (full unit suite 1238 passing, lint/tsc/format clean). Happy to address any specific concerns.

mark-wiemer:
@Minouccc I cannot reopen the PR because the branch was force-pushed or recreated. Please create a new PR without the excessive comments and it should get past our spam filter. Thanks for your patience.

Minouccc:
Done — opened #6331, single commit this time.

Issue #6033, opened 2026-06-05
🐛 Bug: Suite.timeout() propagation misses hooks, chained describe(...).timeout(N) does not apply to hooks


satouriko:
### Bug Report Checklist

- [x] This is NOT a [security, `npm audit`, or GitHub Advisory issue](https://mochajs.org/explainers/security-vulnerability-reports).
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

## Summary

PR #5612 made `Suite.prototype.timeout(ms)` retroactively propagate the new timeout to already-registered children, fixing #5422. However, it only iterates over `this.tests` and `this.suites` — it misses the hook arrays (`_beforeAll`, `_beforeEach`, `_afterAll`, `_afterEach`).

As a result, the **chained** form

```js
describe('foo', () => {
  before(async () => { /* slow setup */ });
  it('bar', () => { /* ... */ });
}).timeout(10000);
```

does **not** raise the timeout for `before` (or any other hook). The hooks were created during the describe callback with the suite's timeout at that moment (default 2000ms), and the subsequent `.timeout(10000)` call never reaches them.

## Which call shapes work / don't work

| Form                                                                                       | `before` / `after`                                                                                  | `it`                |
| ------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------- | ------------------- |
| `describe('x', function () { this.timeout(N); before(...); it(...); })`                    | ✅ works (hook reads suite timeout at `_createHook` time)                                           | ✅ works            |
| `describe('x', () => { before(...); it(...); }).timeout(N)` (chained, after callback returns) | ❌ **broken** — hooks were created with default 2000ms, setter doesn't propagate to them            | ✅ works (PR #5612) |
[...]

Issue #6028, opened 2026-05-29
🐛 Bug: Regression in 11.7.6: Suite.timeout() propagation (#5612) can trigger ERR_MOCHA_TIMEOUT / ERR_MOCHA_MULTIPLE_DONE after all tests pass


vrubezhny:
## Summary

After upgrading to Mocha 11.7.6, we started seeing sporadic failures of the form:

```text
Error: done() called multiple times in test <...>
...
ERR_MOCHA_TIMEOUT
ERR_MOCHA_MULTIPLE_DONE
```

even though all tests had already completed successfully.

After investigation, the issue appears to be related to the change introduced by:

- Issue: #5422
- PR: #5612

which modified `Suite.prototype.timeout()` to recursively propagate timeout changes to child tests and suites.

The problem disappears if we patch `Suite.prototype.timeout()` to remove the recursive propagation logic added in #5612.

---

## Minimal reproduction

`repro.spec.js`

```js
const assert = require('assert');

describe('parent suite', function () {

    it('normal test', async function () {
        this.timeout(1000);  // this === Test Context
        assert.ok(true);
    });

    it('arrow test #1', async () => {
        // This is actually suite.timeout(...)
        this.timeout(2000);  // this === Suite

        await new Promise(resolve => setTimeout(resolve, 10));

        assert.ok(true);
    });

    it('arrow test #2', async () => {
        // This is actually suite.timeout(...)
        this.timeout(3000);  // this === Suite

        await new Promise(resolve => setTimeout(resolve, 10));

        assert.ok(true);
    });
});
```

---

## Steps to reproduce

```bash
npx mocha@11.7.6 repro.spec.js
```

---

## Actual result

Mocha reports all tests as passing:

```text
parent suite
  ✔ normal test
  ✔ arrow test #1
  ✔ arrow test #2

3 passing
```

but then throws:

```text
Error: done() called multiple times in test <parent suite normal test>
...
ERR_MOCHA_TIMEOUT
ERR_MOCHA_MULTIPLE_DONE
```

after the test run has completed.

---

## Expected result

The run should complete successfully with:

```text
3 passing
```

and no subsequent timeout or multiple-done errors.

---

## Regression

The behavior appears to be related to the recursive timeout propagation added in:

https://github.com/mochajs/mocha/pull/5612

Specifically:

```js
for (const t of this.tests) {
  t.timeout(this._timeout);
}
for (const s of this.suites) {
  s.timeout(this._timeout);
}
```

[...]

PR #6348, opened 2026-09-16
fix: make --timeout 0 disable timeouts for the whole run

### LINKED ISSUE #4834: 🐛 Bug: `--timeout 0` does not disable timeouts
### Prerequisites

### Description

I ran a mocha test using IntelliJ's debugger, which automatically sets `--timeout 0`. Left it paused for several minutes while looking at something, and when I resumed it, it timed out.

### Steps to Reproduce

1. Create a test script containing a single describe() and a single it() (you can do more, this is just a simple example)
2. Inside the describe(), put `this.timeout(T)` (I haven't tried with the timeout() call inside the it(), not sure if behaviour changes)
3. Ensure somehow that the `it()` will take longer than `T` to run: set a breakpoint, or add code that will take longer than that
4. Run the test with `--timeout 0`

#### Working example

1. Check out https://github.com/bdcarr/mocha-timeout-bug/tree/main
2. `cd newversion`
5. `npm install`
6. `npm run test`

There's also an `oldversion` folder which contains the same test script, but the package.json is set to use mocha@7.2.1, which is the last version where this issue doesn't occur. I checked with v8.0.0 and the issue occurs.

**Expected behavior:** [No matter how long the test takes, it will not time out if run with `--timeout 0`](https://github.com/mochajs/mocha/blob/547ffd73535088322579d3d2026432112eae3d4b/docs/index.md#--timeout-ms--t-ms)

**Actual behavior:** The test times out after `T` milliseconds

### Versions

- The output of `mocha --version` and `node_modules/.bin/mocha --version`: No global install. Issue reproduced on v8.0.0 and v9.2.1. Issue does not occur on v7.2.1.
- The output of `node --version`: v14.9.0
- Your operating system
  - name and version: MacOS 12.2.1
- Your shell (e.g., bash, zsh, PowerShell, cmd): zsh v5.8
- Any third-party Mocha-related modules (and their versions): n/a
- Any code transpiler (e.g., TypeScript, CoffeeScript, Babel) being used (and its version): n/a, straight NodeJS with no imports

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #4834

## Overview

The `--timeout 0` flag now turns timeouts off for the whole run so a `this.timeout(ms)` in a spec file can't switch them back on.

Before this the the run-level 0 was only a starting value and anything in a test file overwrote it. I would say that is what made debugging unbelievable difficult because the inspect flags imply `--timeout 0` so pausing on a breakpoint still failed any test that set its own timeout. Suite-level, test-level, hook-level and chained calls were all affected.

Also, if you run with ´--timeout 0` and lean on an in-file timeout to catch a hang then that test will now hang instead of failing. That felt like the right reading of "disable timeouts" but i can look at it again if needed.

I left the zero-then-nonzero behaviour from #4254 alone so setting 0 and then a real value inside one test still times out. I also did not change the separate suite timeout propagation bugs in issues #6028 as well as #6033


docs/src/content/docs/running/cli.mdx
```diff
@@ -311,6 +311,8 @@ Tests taking longer than this amount of time will be marked as failed.
 To override you may pass the timeout in milliseconds, or a value with the `s` suffix, e.g., `--timeout 2s` and `--timeout 2000` are equivalent.
 
 To disable timeouts, use `--timeout 0`.
+This disables them for the whole run so a `this.timeout(ms)` in a test file cannot bring them back.
+That is what keeps a breakpoint from failing a test when the inspect flags imply `--timeout 0`.
```
mark-wiemer: No need to mention this specific use case, best to keep docs succinct

```suggestion
```

test/integration/fixtures/options/timeout-in-file-override.fixture.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+describe('suite-level override', function() {
+  this.timeout(10);
```
mark-wiemer:
```suggestion
  this.timeout(1);
```

test/integration/fixtures/options/timeout-in-file-override.fixture.js
```diff
@@ -0,0 +1,30 @@
...
+
+  it('should not time out an async test', function(done) {
+    setTimeout(done, 60);
```
mark-wiemer: Applies throughout the file

```suggestion
    setTimeout(done, 2);
```

test/integration/fixtures/options/timeout-in-file-override.fixture.js
mark-wiemer: Applies throughout the file

mark-wiemer (changes requested):
Minor changes, otherwise pretty good. Blocking on v13 though, even though our docs are ambiguous. See backing issue for semver-major discussion

test/integration/fixtures/options/timeout-in-file-override.fixture.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+});
+
+describe('hook-level override', function() {
+  before(function(done) {
+    this.timeout(10);
+    setTimeout(done, 60);
+  });
+
```
mark-wiemer:
```suggestion
    while (Date.now() - start < 2);
  });
});

describe('test-level override', function() {
  it('should not time out', function(done) {
    this.timeout(1);
    setTimeout(done, 2);
  });
});

describe('hook-level override', function() {
  before(function(done) {
    this.timeout(1);
    setTimeout(done, 2);
  });

```

mark-wiemer (approved):

PR #6342, opened 2026-09-16
fix: use find-up-simple to workaround ERR_PACKAGE_PATH_NOT_EXPORTED

mark-wiemer:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6319
- [x] That issue was marked as [`status: accepting prs`](https://github.com/mochajs/mocha/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+accepting+prs%22)
- [x] Steps in [CONTRIBUTING.md](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md) were taken

## Overview

Validated using the new `repro` folder, see commit history. Will add a proper test as part of #6341 .

Order is respected in the same way as [`find-up` respects order](https://github.com/sindresorhus/find-up#findupname-options-1).

Issue #6319, opened 2026-09-15
🐛 Bug: ERR_PACKAGE_PATH_NOT_EXPORTED in Mocha 12.0.1 with `--import=tsx`


danielhjacobs:
### Bug Report Checklist

- [x] This is NOT a [security, `npm audit`, or GitHub Advisory issue](https://mochajs.org/explainers/security-vulnerability-reports).
- [x] I have read and agree to Mocha's [Code of Conduct](https://github.com/mochajs/mocha/blob/main/.github/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/mochajs/mocha/blob/main/.github/CONTRIBUTING.md)
- [x] I have searched for [related issues](https://github.com/mochajs/mocha/issues?q=is%3Aissue) and [issues with the `faq` label](https://github.com/mochajs/mocha/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Afaq%20), but none matched my issue.
- [x] I have 'smoke tested' the code to be tested by running it outside the real test suite to get a better sense of whether the problem is in the code under test, my usage of Mocha, or Mocha itself.
- [ ] I want to provide a PR to resolve this

### Expected

`npm test` should continue working in https://github.com/ruffle-rs/ruffle/pull/24700

I claimed to have smoke tested because this is a mocha regression between 12.0.0 and 12.0.1, as `npm test` works without the mocha upgrade.

### Actual

The underlying issue is https://github.com/sindresorhus/find-up/issues/74, and its underlying issue is with https://github.com/sindresorhus/unicorn-magic

```
dj@z2:~/work/rust/ruffle_source/ruffle/web$ npm test

> ruffle@0.6.0 test
> npm test --workspaces --if-present


> ruffle-core@0.6.0 test
> mocha

node:internal/modules/esm/resolve:315
  return new ERR_PACKAGE_PATH_NOT_EXPORTED(
         ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /home/dj/work/rust/ruffle_source/ruffle/web/node_modules/unicorn-magic/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:315:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:606:13)
    at resolveExports (node:internal/modules/cjs/loader:759:36)
    at Module._findPath (node:internal/modules/cjs/loader:826:31)
    at node:internal/modules/cjs/loader:1562:27
[...]

PR #6356, opened 2026-09-18
test: add --import=tsx integration test

### LINKED ISSUE #6341: 🛠️ Repo: Add `--import=tsx` test

### Overview

Ref #6319 , we should have a test to catch issues like that.

8rulerstar:
## PR Checklist

- [x] Addresses an existing open issue: fixes #6341

## Overview

Adds integration tests that run a `.ts` spec with `import=tsx`, using real `tsx` instead of a simulated import hook. It covers both the CLI flag (`--node-option import=tsx`) and a `.mocharc.yml` with `node-option`, which is how #6319 was reported. The fixture uses an `enum` so it only passes when tsx is actually loaded, since Node's built-in type stripping rejects it.

Verified locally: with the #6342 fix reverted (back to `find-up` 8), both tests fail with the original `ERR_PACKAGE_PATH_NOT_EXPORTED` error from #6319. On `main` they pass.

Notes:
- Adds `tsx` as a devDependency and to knip's `ignoreDependencies`, since it's only referenced as a string, same as `coffeescript`.
- The lockfile also bumps `fsevents` 2.3.2 → 2.3.3, which `tsx` requires.
- Tested on macOS with Node 26 only. Happy to adjust if you'd prefer a different approach.


mark-wiemer (approved):
Thank you!

PR #6367, opened 2026-09-19
fix: stop masking import() errors for TypeScript files

### LINKED ISSUE #5792: 🐛 Bug: esm-utils.js#requireModule() masks more helpful requireErr in some cases

mostafaNazari702:
## PR Checklist

- [x] Addresses an existing open issue: fixes #5792

## Overview

Forward-ports #5498 which only ever landed on v11.x. Right now any .ts file throws away the import() error even when that is the useful one which is what #6156 ran into.

Stacked on #6366 , so it should be reviewed and merged first.


mostafaNazari702:
This PR contains #6366's commit. That is deliberate since its test fails without the loader fix, otherwise.

